PR #4, opened 2021-04-02
ACI dynamic compilation & kryo deserialization

tbekas:


src/main/scala/org/tessellation/aci/ACIRepository.scala
```diff
@@ -0,0 +1,45 @@
...
+                 t.id, t.address, t.definition
+          from aci_mapping m 
+          join aci_type t
+          on m.dst_type_id = t.id
+         where m.src_type_id = $srcTypeId"""
+      .query[(ACITypeMapping, ACIType)]
+      .to[List]
+      .transact(xa)
+
+  def saveAciType(aciType: ACIType): F[Int] =
```
buckysballs, line 32: Here (or in the endpoint) we will want to do a validation step which checks if the type inherits from Omega

buckysballs: is it of type A <: \Omega

kpudlik: Ω

src/main/scala/org/tessellation/aci/ExplicitKryoRegistrar.scala
```diff
@@ -0,0 +1,26 @@
...
+    classes.foreach {
+      case (registryClass, id) => k.register(registryClass, id)
+    }
+}
+
+object ExplicitKryoRegistrar {
+  implicit val monoid: Monoid[ExplicitKryoRegistrar] =
+    new Monoid[ExplicitKryoRegistrar] {
+
+      def combine(x: ExplicitKryoRegistrar,
```
buckysballs, line 19: When we are recompiling the registrar, we want to include all algebras/coalgebras and implicit methods into the Home object

tbekas:
Closing in favour of a new PR

PR #7, opened 2021-05-21
Monoid

kpudlik:


src/main/scala/org/tessellation/schema/Hom.scala
```diff
@@ -177,28 +175,76 @@ trait Hom[+A, +B] extends Ω {}
...
+        case (Cell(NullTerminal(), _), Cell(NullTerminal(), _)) => empty
+        case (cella@Cell(a, _), cellb@Cell(b, _)) => {
+
+          val input: ΩList = a :: b
+
+          val combinedHyloM: Ω => M[Either[CellError, Ω]] = (a: Ω) => {
+            for {
+              aOutput <- cella.run()
+              bOutput <- cellb.run()
+              combinedOutput = aOutput.flatMap(aΩ => bOutput.map(bΩ => aΩ :: bΩ))
```
marcinwadon, line 236: hmm. is it exactly the same `cella.run().flatMap(a => cellb.run()`?
I'm thinking about a fast-fail but for comp is a syntax sugar so it should also not compute b output when a fails, right?

src/main/scala/org/tessellation/schema/ΩList.scala
```diff
@@ -0,0 +1,11 @@
+package org.tessellation.schema
+
+abstract class ΩList extends Ω {
+  def ::(elem: Ω): ΩList = new ::(elem, this)
+}
+case class ::(h: Ω, t: ΩList) extends ΩList {
+  override def toString: String = s"$h :: $t"
```
marcinwadon, line 7: instead of `toString` you should use `Show` class

src/main/scala/org/tessellation/schema/Hom.scala
kpudlik: Yes, it should fast-fail and put the very first error as `Left`. Actually that is open question how to call both cells (one after another? parallel?) so that fast-fail may be deterministic or not :D

src/main/scala/org/tessellation/schema/ΩList.scala
kpudlik: I like the idea but in the end, I need to override toString anyway to use such printing syntax everywhere.

Similar pattern as HList uses: https://github.com/milessabin/shapeless/blob/main/core/src/main/scala/shapeless/hlists.scala#L38

src/main/scala/org/tessellation/schema/Hom.scala
```diff
@@ -177,28 +174,78 @@ trait Hom[+A, +B] extends Ω {}
...
+  case class NullTerminal() extends Ω
+
+  def unapply[M[_], F[_], A, B, S](cell: Cell[M, F, A, B, S]): Some[(A, S => M[B])] = Some(cell.data, cell.hylo)
+
+  implicit def toCell[M[_], F[_], A <: Cell[M, F, Ω, Either[CellError, Ω], Ω]](
+    a: A
+  ): Cell[M, F, Ω, Either[CellError, Ω], Ω] = a.asInstanceOf[Cell[M, F, Ω, Either[CellError, Ω], Ω]]
+
+  implicit def cellMonoid[M[_]: Applicative, F[_]: Applicative](
+    implicit M: Monad[M],
```
tbekas, line 197: I think that the `Applicative` constraint on `M` is unnecessary since you have already `Monad` on `M`.

src/main/scala/org/tessellation/schema/Hom.scala
```diff
@@ -177,28 +174,78 @@ trait Hom[+A, +B] extends Ω {}
...
+    F: Traverse[F]
+  ): Monoid[Cell[M, F, Ω, Either[CellError, Ω], Ω]] =
+    new Monoid[Cell[M, F, Ω, Either[CellError, Ω], Ω]] {
```
tbekas, line 200: I thought that we need `MonoidK` for Cell since we're combining the structure, not the values? But if that's not the case I think we'll need a `Monoid` or at least `Semigroup` for the value type, since a regular `Monoid` should also combine values.

src/test/scala/org/tessellation/consensus/CellMonoidTest.scala
```diff
@@ -0,0 +1,132 @@
...
+        )
+      )
+      val l1Cell = L1StartConsensusCell(L1Edge(Set.empty[L1Transaction]), metadata)
+      val l0Cell = L0Cell(L0Edge(Set.empty[L1Block]))
+
+      val l1Result = l1Cell.run().unsafeRunSync()
+      val l0Result = l0Cell.run().unsafeRunSync()
+      val expectedResult = l1Result.flatMap(a => l0Result.map(b => a :: b))
+
+      // TODO: Investigate why syntax doesn't work
```
tbekas, line 120: The import should be `cats.syntax.all._` instead of `cats.syntax._`.

src/main/scala/org/tessellation/schema/Hom.scala
kpudlik: Fair point. It must have been a leftover :+1:

src/main/scala/org/tessellation/schema/Hom.scala
kpudlik: The question is then which type of `Cell` needs to be `*` to implement `MonoidK`. The implementation under the hood will be the same I guess for both `MonoidK` and `Monoid`. Such situation happens for example for `Monoid[List[Int]]` and `MonoidK[List].empty[Int]`. If we use `MonoidK` them we need to fight more with casting as  `A` type passed to `empty[A]` and `combineK[A]` need to be the same for all the cells (sooo there must be a common type so we can cast both As to same type...). So in the end we will just do `empty[Ω]` and `comineK[Ω]` which won't differ from `Monoid`. We use `ΩList extends Ω` as common type for all the `Ω`.

src/test/scala/org/tessellation/consensus/CellMonoidTest.scala
kpudlik: Didn't help, unfortunately.

buckysballs (approved):

src/test/scala/org/tessellation/consensus/CellMonoidTest.scala
```diff
@@ -0,0 +1,132 @@
...
+import org.scalatest.matchers.should.Matchers
+import org.tessellation.Node
+import org.tessellation.consensus.L1ConsensusStep.{L1ConsensusContext, L1ConsensusMetadata}
+import org.tessellation.consensus.transaction.RandomTransactionGenerator
+import org.tessellation.schema.Cell.{NullTerminal, cellMonoid}
+import org.tessellation.schema.{Cell, CellError, StackF, Ω, ΩList}
+import org.tessellation.snapshot.{L0Cell, L0Edge, Snapshot}
+import cats.syntax._
+
+class CellMonoidTest
```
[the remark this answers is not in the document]
kpudlik: Hmm I think we have similar unit tests there:

"Combine pipelines of same types": Lines 95-98: `val combined = M.combine(l1CellA, l1CellB)`
"Combine pipelines of different types": Line 122: `val combined = M.combine(l1Cell, l0Cell)`

>In both cases we want to merge the pipelines by calling combine on their head elements and choosing sequential Cells recursively

Ok I think I get it. So you just want to continuously combine whole stream pipelines instead of testing single combine result 🤔 .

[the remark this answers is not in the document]
buckysballs: Yes exactly. We can just merge two `L1 >>> L2` pipelines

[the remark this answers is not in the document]
kpudlik: Isn't it then a unit test for `Arrow` instead of for `Monoid`? 🤔

PR #16, opened 2021-09-17
P2P Client + Tomek's serialization

marcinwadon:


modules/shared/src/main/scala/org/tesselation/schema/cluster.scala
```diff
@@ -19,6 +19,11 @@ object cluster {
   @derive(decoder, encoder, show)
   case class PeerToJoin(id: PeerId, ip: Host, p2pPort: Port)
 
+  object PeerToJoin {
+    implicit def toPeer(peer: PeerToJoin): Peer =
+      Peer(peer.id, peer.ip, peer.p2pPort, peer.p2pPort)
```
TheMMaciek, line 24: Is that correct? The same port used twice?

TheMMaciek (approved):

modules/shared/src/main/scala/org/tesselation/schema/cluster.scala
marcinwadon: Yes, at least for now.

PR #17, opened 2021-09-17
Majority picking cell implementations

TheMMaciek:


tbekas (approved):
Shouldn't we merge it into `feature/node2.0` branch instead of `develop`?

PR #18, opened 2021-09-22
Joining handshake

marcinwadon:


build.sbt
```diff
@@ -125,8 +139,8 @@ lazy val core = (project in file("modules/core"))
   .settings(
     name := "tesselation-core",
     scalacOptions ++= List("-Ymacro-annotations", "-Yrangepos", "-Wconf:cat=unused:info", "-language:reflectiveCalls"),
-    scalafmtOnCompile := true,
-    scalafixOnCompile := true,
+    scalafmtOnCompile := false,
+    scalafixOnCompile := false,
```
tbekas, line 143: What was the reason for this change? Rules were too restrictive?

modules/core/src/main/scala/org/tesselation/config/Config.scala
```diff
@@ -26,22 +26,23 @@ object Config {
 
   def default[F[_]](environment: AppEnvironment): ConfigValue[F, AppConfig] =
     (
-//      env("CL_DUMMY_SECRET").default("foo").secret,
+      env("CL_EXTERNAL_IP").default("127.0.0.1"),
```
tbekas, line 29: Maybe we can use regular config for that? Env variables with `CL_` prefix would just override config properties resulting in the same behavior but in a uniform way.  https://github.com/lightbend/config#optional-system-or-env-variable-overrides

modules/core/src/main/scala/org/tesselation/Main.scala
```diff
@@ -25,26 +27,29 @@ object Main extends IOApp {
...
+                  storages <- Resource.eval {
+                    Storages.make[IO]
+                  }
+                  services <- Resource.eval {
+                    Services.make[IO](cfg, nodeId, storages, p2pClient)
+                  }
+                  api = HttpApi.make[IO](storages, services, cfg.environment)
+                  _ <- MkHttpServer[IO].newEmber(ServerName("public"), cfg.publicHttp, api.publicApp)
+                  _ <- MkHttpServer[IO].newEmber(ServerName("p2p"), cfg.p2pHttp, api.p2pApp)
+                  _ <- MkHttpServer[IO].newEmber(ServerName("cli"), cfg.cliHttp, api.cliApp)
```
tbekas, line 47: I like the naming `"cli"` endpoints instead of the `"owner"` endpoints 👍

modules/core/src/main/scala/org/tesselation/http/routes/ClusterRoutes.scala
```diff
@@ -38,13 +36,26 @@ final case class ClusterRoutes[F[_]: Async](
             case PeerIdInUse(id) => Conflict(s"Peer id=${id} already in use.")
             case PeerHostPortInUse(host, port) =>
               Conflict(s"Peer host=${host.toString} port=${port.value} already in use.")
-            case SessionAlreadyExists =>
-              Conflict(s"Session already exists.")
+            case SessionDoesNotExist =>
+              Conflict(s"Session does not exist.")
```
tbekas, line 40: Why does it result in a 409 error? Isn't this an expected behavior when joining, that the session will not exist?

modules/core/src/main/scala/org/tesselation/http/routes/RegistrationRoutes.scala
```diff
@@ -0,0 +1,27 @@
...
+
+final case class RegistrationRoutes[F[_]: Async](cluster: Cluster[F]) extends Http4sDsl[F] {
+
+  implicit val logger = Slf4jLogger.getLogger[F]
+
+  private[routes] val prefixPath = "/registration"
+
+  private val httpRoutes: HttpRoutes[F] = HttpRoutes.of[F] {
+    case GET -> Root / "request" =>
+      Ok(cluster.getRegistrationRequest)
```
tbekas, line 21: Is this a debug-only endpoint? Should it be committed?

modules/shared/src/main/scala/org/tesselation/ext/crypto.scala
```diff
@@ -0,0 +1,30 @@
...
+  implicit class RefinedSigned[F[_]: KryoSerializer, A <: AnyRef](data: A) {
+
+    def sign(): Either[Throwable, Signed[A]] = Signed.forKryo[F, A](data)
+  }
+
+  implicit class RefinedSignedF[F[_]: KryoSerializer, A <: AnyRef](data: A) {
+
+    def signF[M[_]: MonadThrow](): M[Signed[A]] = Signed.forKryo[F, A](data).liftTo[M]
+  }
+}
```
tbekas, line 30: Nice, I like that you're adding syntax for core functionalities 👍

build.sbt
marcinwadon: It was changing the code just before compiling it, leading to non-compilable code.

modules/core/src/main/scala/org/tesselation/config/Config.scala
marcinwadon: We can discuss this and change if needed

modules/core/src/main/scala/org/tesselation/http/routes/ClusterRoutes.scala
marcinwadon: No, when you try to join to the node that does not have a session, you shouldn't be able to join.

modules/core/src/main/scala/org/tesselation/http/routes/RegistrationRoutes.scala
marcinwadon: No, debug endpoints are in `DebugRoutes`. Here is the proper endpoint for asking for a registration request.

build.sbt
marcinwadon: bt actually that's wrong change. I will revert it

modules/core/src/main/scala/org/tesselation/http/routes/ClusterRoutes.scala
marcinwadon: @tbekas I rephrased the error message to be more verbose

marcinwadon: you were right.

modules/core/src/main/scala/org/tesselation/config/Config.scala
tbekas: Ok

modules/core/src/main/scala/org/tesselation/http/routes/ClusterRoutes.scala
tbekas: Cool, thanks for checking it.

tbekas (approved):
LGTM

modules/shared/src/main/scala/org/tesselation/ext/crypto.scala
TheMMaciek: Nice. Did you check if the possibility of mixup exists? So the situation where I expected that I'm fetching this hash, but I'm actually getting some scala default implementation of a hash?

marcinwadon: @TheMMaciek not yet

marcinwadon: but actually you would want `Hash` or `Signed` type classes. default implementation of `.hash` returns Int probably

TheMMaciek: Yep, relying on the Hash and Signed should give us a lot of safety.

TheMMaciek (approved):

PR #19, opened 2021-09-23
Migrations support

tbekas:


tbekas:
Tests are migrated to weaver, check PR #20

PR #20, opened 2021-09-24
Migrations support (weaver tests)

tbekas:


modules/shared/src/test/scala/org/tesselation/kryo/KryoSerializerSuite.scala
```diff
@@ -0,0 +1,50 @@
...
+
+    for {
+      bytes <- serializerV2.use { implicit kryo =>
+        kryo.serialize(v2).liftTo[IO]
+      }
+      obj <- serializerV1.use { implicit kryo =>
+        kryo.deserialize[NoChangesV1](bytes).liftTo[IO]
+      }
+      expectedV1 = NoChangesV1(amount = 15, address = "anyAddress")
+    } yield expect(obj == expectedV1)
```
marcinwadon, line 47: you can use `expect.same(obj, expectedV1)` probably

modules/shared/src/main/scala/org/tesselation/kryo/KryoSerializer.scala
```diff
@@ -31,18 +34,23 @@ object KryoSerializer {
...
-        def deserialize[T](bytes: Array[Byte]): Either[Throwable, T] =
-          Either.catchNonFatal {
-            kryoPool.fromBytes(bytes).asInstanceOf[T]
-          }
-      }
+  def forAsync[F[_]: Async](
+    registrar: Map[Class[_], Int],
+    migrations: List[Migration[AnyRef, AnyRef]] = List.empty
+  ): Resource[F, KryoSerializer[F]] = make[F](registrar).map { kryoPool =>
+    val migrationsMap = migrations.map(_.toPair).toMap
```
marcinwadon, line 41: it's used once only, so you can use it there explicitly:
```
migrations
  .map(_.toPair)
  .toMap
  .getOrElse...
```

tbekas: I think it might recalculate the map every time we deserialize something. I think that compiler isn't that smart to cut it off after `.toMap`

marcinwadon: Both ways it will recalculate IMO

tbekas: I checked by adding stdout logs:
```
.map { m =>
  val p = m.toPair
  println(s"mapping ${p._1}")
  p
}.toMap
```
and it recalculates the whole map during each deserialization. If we keep it as a variable outside of the `deserialize` function (as it is), it doesn't recalculate it.

marcinwadon: Yep, you're right. I forget we're using a Resource here that is initialized once only

marcinwadon (approved):

PR #21, opened 2021-09-27
Cluster structure and HashSignature with Signed class

marcinwadon:


modules/keytool/src/main/scala/org/tesselation/keytool/security/Signing.scala
```diff
@@ -1,5 +1,40 @@
...
+      _ <- WithSecureRandom.get[F].flatMap { secureRandom =>
+        Async[F].delay { s.initSign(privateKey, secureRandom) }
+      }
+      _ <- Async[F].delay { s.update(bytes) }
+      signed <- Async[F].delay { s.sign() }
+    } yield signed
+
+  def verifySignature[F[_]: Async: SecurityProvider](
+    originalInput: Array[Byte],
+    signedOutput: Array[Byte],
```
tbekas, line 29: Isn't `signedOutput` just a signature? If this is the case I would rename the first two parameters of this function to just `inputData` (or just `bytes` like in function above) and `signature`.

tbekas (approved):
LGTM

modules/keytool/src/main/scala/org/tesselation/keytool/security/Signing.scala
marcinwadon: I left it as it is in the original code

PR #23, opened 2021-09-28
Introduce kernel module with basic types

TheMMaciek:


modules/kernel/src/main/scala/org/tesselation/kernel/Cell.scala
```diff
@@ -0,0 +1,74 @@
+package org.tesselation.kernel
+
+import cats._
+import cats.implicits._
```
marcinwadon, line 4: Please use new syntax here, `import cats.implicits._` should not be used anymore, instead:
`import cats.syntax.all._` or `import cats.syntax.functor._` and so on for every feature you want

modules/kernel/src/main/scala/org/tesselation/kernel/PipeArrow.scala
```diff
@@ -0,0 +1,21 @@
+package org.tesselation.kernel
+
+import cats.arrow.Arrow
+import cats.effect.IO
+
+import fs2.{Pipe, Stream}
+
+object PipeArrow {
+  implicit val arrowInstance: Arrow[Pipe[IO, *, *]] = new Arrow[Pipe[IO, *, *]] {
```
marcinwadon, line 9: I'd name it `arrowInstanceForIO` or `arrowIOInstance`

modules/kernel/src/main/scala/org/tesselation/kernel/Stack.scala
```diff
@@ -0,0 +1,38 @@
+package org.tesselation.kernel
+
+import cats.syntax.all._
```
marcinwadon, line 3: If possible, please import explicitly what's needed.

modules/kernel/src/test/scala/org/tesselation/kernel/CellMonoidTest.scala
```diff
@@ -0,0 +1,137 @@
...
+import cats.effect.IO
+import cats.syntax.all._
+
+import org.tesselation.kernel.Cell.NullTerminal
+
+import higherkindness.droste.{AlgebraM, CoalgebraM, scheme}
+import weaver.SimpleIOSuite
+import weaver.scalacheck.Checkers
+
+object CellMonoidTest extends SimpleIOSuite with Checkers {
```
marcinwadon, line 13: imo it should be tested using Discipline or cats-laws

marcinwadon (changes requested):

modules/kernel/src/main/scala/org/tesselation/kernel/Cell.scala
TheMMaciek: Right!

modules/kernel/src/main/scala/org/tesselation/kernel/Stack.scala
TheMMaciek: Ok, so we are changing the approach right? So far we were importing everything.

modules/kernel/src/test/scala/org/tesselation/kernel/CellMonoidLawsSuite.scala
```diff
@@ -0,0 +1,34 @@
...
+  // WARN: By comparing cell's run method result we can consider monoid for cells fulfilling the monoid laws.
+  //       Because two functions with the same implementation are never equal to each other we can't compare
+  //       the cells to each other by comparing the cell objects (neither convert nor hylo functions will be ever equal).
+  implicit val eq: Eq[Cell[IO, StackF, Ω, Either[CellError, Ω], Ω]] =
+    (x: Cell[IO, StackF, Ω, Either[CellError, Ω], Ω], y: Cell[IO, StackF, Ω, Either[CellError, Ω], Ω]) => {
+      for {
+        xResult <- x.run()
+        yResult <- y.run()
+      } yield xResult == yResult
+    }.unsafeRunSync()
```
marcinwadon, line 27: IMO we should write it without this unsafe run sync

marcinwadon (commented):
Looks solid :+1:

modules/kernel/src/test/scala/org/tesselation/kernel/CellMonoidLawsSuite.scala
TheMMaciek: Maybe I could test it with a different monad 🤔

TheMMaciek: Switched to Id, will push the update in a second.

marcinwadon (approved):

PR #29, opened 2021-10-06
Quill, Doobie setup with Address DB layer

kpudlik:


modules/core/src/main/scala/org/tesselation/config/Config.scala
```diff
@@ -1,19 +1,53 @@
 package org.tesselation.config
 
 import cats.effect.Async
-import cats.syntax.parallel._
+import cats.implicits._
```
marcinwadon, line 4: Please use syntax

modules/core/src/main/scala/org/tesselation/domain/db/AddressStorage.scala
```diff
@@ -0,0 +1,11 @@
+package org.tesselation.domain.db
+
+import org.tesselation.schema.address.{Address, Balance}
+
+trait AddressStorage[F[_]] {
+  def getBalance(address: Address): F[Balance]
+
+  def updateBalance(address: Address, balance: Balance): F[schema.StoredAddress]
```
marcinwadon, line 8: Import schema directly

marcinwadon: Yet, I think it shouldn't be named "db" because it leaks the implementation detail

modules/core/src/main/scala/org/tesselation/infrastructure/db/AddressStorage.scala
```diff
@@ -0,0 +1,33 @@
...
+  ): AddressStorage[F] = new AddressStorage[F] {
+
+    val xa = DoobieTransactor[F].xa
+
+    import ctx._
+
+    override def getBalance(address: Address): F[Balance] =
+      run(getAddressBalance(lift(address))).map(_.headOption.getOrElse(Balance(0))).transact(xa)
+
+    override def updateBalance(address: Address, balance: Balance): F[schema.StoredAddress] =
```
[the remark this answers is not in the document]
marcinwadon: Yes, you need to run it in a single transact

marcinwadon (changes requested):

modules/core/src/main/scala/org/tesselation/config/types.scala
```diff
@@ -4,17 +4,22 @@ import scala.concurrent.duration.FiniteDuration
...
-    externalIp: Host,
-    publicHttp: HttpServerConfig,
-    p2pHttp: HttpServerConfig,
-    cliHttp: HttpServerConfig
+    httpConfig: HttpConfig,
+    dbConfig: DBConfig
+  )
+
+  case class DBConfig(
+    driver: NonEmptyString, // TODO: Make it Regex refined type?
```
tbekas, line 19: I'm in favor of not putting TODOs into the `develop` or `master` branches. I think that if something is important to do then it should be either done as a part of the current change or a task on JIRA should be created for it.

marcinwadon: +1

marcinwadon (approved):

PR #30, opened 2021-10-12
Genesis parser with initialization flow

marcinwadon:


modules/core/src/main/scala/org/tesselation/Main.scala
```diff
@@ -27,31 +32,39 @@ object Main extends IOApp {
...
+                      nodeId = PeerId.fromPublic(keyPair.getPublic)
+                      p2pClient = P2PClient.make[IO](res.client)
+                      storages <- Resource.eval(Storages.make[IO])
+                      services <- Resource.eval(Services.make[IO](cfg, nodeId, keyPair, storages))
+                      programs <- Resource.eval(Programs.make[IO](storages, services, p2pClient, nodeId))
+                      _ <- Resource.eval {
+                        cli.method match {
+                          case CliMethod.RunValidator => IO.unit
+                          case CliMethod.RunGenesis =>
+                            GenesisLoader.make[IO].load(Path(cli.genesisPath)).flatMap { accounts =>
```
kpudlik, line 52: What is the type of `genesisPath`? Maybe you could cast it to `Path` because it probably won't be used as regular string. The same we do with `Url` itd.

modules/core/src/main/scala/org/tesselation/cli/config.scala
```diff
@@ -0,0 +1,18 @@
...
+
+  object CliMethod extends Enumeration {
+    type CliMethod = Value
+
+    val RunValidator, RunGenesis = Value
+  }
+
+  case class CliConfig(
+    method: CliMethod = null,
+    genesisPath: String = null
```
kpudlik, line 15: How about `genesisPath: Path`? Imo we should use refined/new/concrete types in both configs

marcinwadon: good catch :+1: done

modules/core/src/main/scala/org/tesselation/Main.scala
marcinwadon: :+1:

PR #34, opened 2021-10-14
Node state transitions

marcinwadon:


modules/core/src/main/scala/org/tesselation/domain/node/NodeStorage.scala
```diff
@@ -0,0 +1,24 @@
...
+import cats.Applicative
+
+import org.tesselation.schema.node.NodeState
+
+trait NodeStorage[F[_]] {
+  def getNodeState: F[NodeState]
+
+  def setNodeState(nodeState: NodeState): F[Unit]
+
+  def tryModifyState[A](from: Set[NodeState], onStart: NodeState, onFinish: NodeState, fn: => F[A]): F[A]
```
kpudlik, line 12: I am wondering if it would be useful or cleaner to curry 🤔 

`def tryModifyState[A](from, onStart, onFinish)(fn): F[A]`

kpudlik (dismissed):

modules/core/src/main/scala/org/tesselation/domain/node/NodeStorage.scala
marcinwadon: yep, great idea. let me try

PR #38, opened 2021-10-18
Prometheus metrics

kpudlik:


modules/core/src/main/scala/org/tesselation/http/routes/MetricRoutes.scala
```diff
@@ -0,0 +1,22 @@
+package org.tesselation.http.routes
+import cats.effect.Async
```
marcinwadon, line 2: isn't a gap needed there?

modules/core/src/main/scala/org/tesselation/infrastructure/metrics/Atomic.scala
```diff
@@ -0,0 +1,31 @@
+package org.tesselation.infrastructure.metrics
+import java.util.concurrent.atomic.{AtomicInteger => JAtomicInteger, AtomicLong => JAtomicLong}
```
marcinwadon, line 2: aa

PR #41, opened 2021-10-19
Cluster state

marcinwadon:


modules/core/src/main/scala/org/tesselation/modules/Daemons.scala
```diff
@@ -3,32 +3,30 @@ package org.tesselation.modules
...
     cfg: AppConfig
-  ): F[Daemons[F]] =
-    for {
-      _ <- Async[F].unit
-      gossipDaemon = GossipDaemon
-        .make[F](storages.rumor, queues.rumor, storages.cluster, p2pClient.gossip, handler, cfg.gossipConfig.daemon)
-      _ <- gossipDaemon.startDaemon
-    } yield new Daemons[F](gossip = gossipDaemon) {}
+  ): F[Unit] =
+    Set[Daemon[F]](
```
tbekas, line 26: Instead of `Set().toList` I would make it a list from the beginning to control the order of the startup.

modules/core/src/main/scala/org/tesselation/infrastructure/cluster/storage/ClusterStorage.scala
```diff
@@ -28,6 +30,18 @@ object ClusterStorage {
 
       def hasPeerHostPort(host: Host, p2pPort: Port): F[Boolean] =
         peers.get.map(_.exists(peer => peer.ip == host && peer.p2pPort == p2pPort))
+
+      def setPeerState(id: PeerId, state: node.NodeState): F[Unit] =
+        hasPeerId(id).ifM(
```
tbekas, line 35: Clearly, the data structure that you need is `Map` of peers instead of a `Set`.

PR #42, opened 2021-10-20
Port over iterative SAW + required classes. Minor changes for languge update.

ryle-goehausen:
Had to remove data generator due to cats effects issues will re-add after resolution. Ref.unsafe produces this error:

```
/Users/ryle/projects/tessellation/modules/core/src/main/scala/org/tesselation/trust/DataGenerator.scala:46:62
could not find implicit value for parameter F: cats.effect.kernel.Sync[[_]F[_]] (Could not find an instance of Monad for [_]F[_])
  private final val randomEffect: Ref[F, Random] = Ref.unsafe(new Random())
```
This is due to the upgrade of cats library. Can be fixed but this whole class needs to be updated anyways and put in the test context. 

I am splitting up the PRs into incremental updates to make this easier to review. This is based on PR on constellation repo for SAW updates to fix the stack overflow -- otherwise it should be identical.

### DIFF
--- a/modules/core/src/main/scala/org/tesselation/trust/DataGenerator.scala
+++ b/modules/core/src/main/scala/org/tesselation/trust/DataGenerator.scala
@@ -1,0 +1,37 @@
+package org.tesselation.trust
+
+case class TrustEdge(src: Int, dst: Int, trust: Double, isLabel: Boolean = false) {
+  def other(id: Int): Int = Seq(src, dst).filterNot(_ == id).head
+}
+
+// Simple way to simulate modularity of connections / generate a topology different from random
+case class TrustNode(id: Int, xCoordinate: Double, yCoordinate: Double, edges: Seq[TrustEdge] = Seq()) {
+
+  def distance(other: TrustNode): Double =
+    Math.sqrt {
+      Math.pow(xCoordinate - other.xCoordinate, 2) +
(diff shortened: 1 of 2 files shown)


marcinwadon (commented):
One thing - can you move the trust package into either `domain` or `infrastructure`? Finally it will live both here and here.
The idea behind that is that `domain` is clear, free of knowledge about specific implementations. Here is where all the traits and interfaces are, as well as things depending only on the domain part (i.e. on the traits). It's a separation of concerns. The `infrastructure` part knows about specific implementations, libraries and tools, etc. We often use `strategy` pattern here. It can also be described as algebra/interpreter pattern (where algebra is free/clean and interpreter says how something should be executed or calculated).

In the context of trust, imo:
TrustEdge should be in the `schema.trust`
TrustNode should be in the `schema.trust` as well.
and the self avoiding walk should be divided into two parts,
one should be generic in the domain, the trait with `calculateTrust` function, or whatever it needs. And SelfAvoidingWalk should be in the infrastructure, implementing the trait and the `calculateTrust` fn that will use self avoiding walk algorithm under the hood.
That way, the implementation is separated from the domain, and other services does not know how trust recalculation works under the hood, they just call `calculateTrust` and that's it.

marcinwadon (approved):

PR #43, opened 2021-10-20
Add CLI route for updating trust labels and cluster storage

ryle-goehausen:
I considered just putting this info directly on Peer class, but it seems like that gets pulled around all over the place, and this doesn't really need to be accessed as much.

Probably a longer discussion about what the correct way to map this information to proper sqlite tables is, but this is sufficient for now for staging future changes.

### DIFF
--- a/modules/core/src/main/scala/org/tesselation/domain/cluster/storage/ClusterStorage.scala
+++ b/modules/core/src/main/scala/org/tesselation/domain/cluster/storage/ClusterStorage.scala
@@ -1,14 +1,10 @@
 package org.tesselation.domain.cluster.storage
 
-import org.tesselation.schema.cluster
-import org.tesselation.schema.cluster.TrustInfo
 import org.tesselation.schema.peer.{Peer, PeerId}
 
 import com.comcast.ip4s.{Host, Port}
 
 trait ClusterStorage[F[_]] {
-  def updateTrust(trustUpdates: cluster.InternalTrustUpdateBatch): F[Unit]
-  def getTrust(): F[Map[PeerId, TrustInfo]]
   def getPeers: F[Set[Peer]]
   def addPeer(peer: Peer): F[Unit]
   def hasPeerId(id: PeerId): F[Boolean]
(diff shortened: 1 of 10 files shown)


modules/core/src/main/scala/org/tesselation/domain/cluster/storage/ClusterStorage.scala
```diff
@@ -1,10 +1,14 @@
...
 
+import org.tesselation.schema.cluster
+import org.tesselation.schema.cluster.TrustInfo
 import org.tesselation.schema.peer.{Peer, PeerId}
 
 import com.comcast.ip4s.{Host, Port}
 
 trait ClusterStorage[F[_]] {
+  def updateTrust(trustUpdates: cluster.InternalTrustUpdateBatch): F[Unit]
+  def getTrust(): F[Map[PeerId, TrustInfo]]
```
marcinwadon, line 11: if that's disjoint to Peer, imo you should create a `trust` or `reputation` package with the `TrustStorage` inside. We'd like to avoid mixing things when not necessary

modules/shared/src/main/scala/org/tesselation/schema/cluster.scala
```diff
@@ -19,6 +19,20 @@ object cluster {
   @derive(decoder, encoder, show)
   case class PeerToJoin(id: PeerId, ip: Host, p2pPort: Port)
 
+  @derive(decoder, encoder, show)
+  case class InternalTrustUpdate(id: PeerId, trust: Double)
+
+  @derive(decoder, encoder, show)
+  case class InternalTrustUpdateBatch(updates: Seq[InternalTrustUpdate])
```
marcinwadon, line 26: please use `List` or `Set`, instead of `Seq`. We're using `List` all over the codebase, firstly to avoid ton of conversions, secondly, it nicely interops with cats and allows for more operations.
I think `List` is more proper here, if there can be at least two updates with the same id and trust

modules/core/src/main/scala/org/tesselation/domain/cluster/storage/ClusterStorage.scala
ryle-goehausen: Done

PR #44, opened 2021-10-20
Gossip signature validation

tbekas:


modules/core/src/main/scala/org/tesselation/infrastructure/gossip/GossipDaemon.scala
```diff
@@ -49,44 +51,57 @@ object GossipDaemon {
...
+          for {
+            computedHash <- signedRumor.value.hashF
+            valid = computedHash == hash
+            _ <- if (valid) Applicative[F].unit
+            else
+              logger.warn(
+                s"Discarding rumor of type ${signedRumor.value.tpe} with hash ${hash.show} due to invalid hash. " +
+                  s"Actual hash was ${computedHash.show}."
+              )
+          } yield valid
```
marcinwadon, line 75: it can be simplified to this:

```scala
signedRumor.value.hashF
  .map(_ == hash)
  .ifM(
    Applicative[F].unit,
    logger.warn(...)
  )
```

tbekas: We need to return an `F[Boolean]` from this closure. We could do it like this
```
signedRumor.value.hashF
  .map(_ == hash)
  .ifM(
    Applicative[F].pure(true),
    logger.warn(...) >> Applicative[F].pure(false)
  )
```
but I'm not sure if this is actually better. It's easy to make a mistake this way, because we're not returning the original result of comparison, but arbitrary `true/false` values. What do you think?

marcinwadon: ah, okay. didn't see that. if so, this would work:
```scala
signedRumor.value.hashF
  .map(_ == hash)
  .flatTap { valid =>
    if (valid) Applicative[F].unit else logger.warn(...)
  }
```
It's actually the same as the for comprehension, just without intermediate names. so up to you :)

PR #49, opened 2021-10-21
[WIP] Making it easy to startup a fresh nodes cluster

esomore:
This PR is trying to make it easy to start a new node cluster; it aims to capture the needed steps for bootstrapping a cluster for others to deploy test environments and experiment with the project as it moves towards mainnet 2.0.

- [ ] Add workflow to generate the keystore so it won't happen in the dockerfile
- [ ] Add multiple instances to garden file and figure out how to configure peer discovery
- [ ] Add tests section in garden file
- [ ] Add workflow to generate a genesis from parameters
- [ ] Add prometheus and grafana dashboards

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,6 @@ COPY build.sbt .
 # Then build
 RUN sbt 'scalafixAll --check --rules OrganizeImports;scalafmtCheckAll;test;it:test;keytool/assembly;core/assembly'
 
-
 FROM lwieske/java-8 as runner
 
 ENV CL_KEYSTORE=keystore.p12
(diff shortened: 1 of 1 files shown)


marcinwadon:
Hey @esomore. Thank you for your contribution.
I'd like to inform you that we're not going to merge your pull request, because of following reasons:
1. First of all, the node deployment should be separated from the core protocol code. We're going to support many cloud providers, as well as docker/k8s. The work is [there](https://github.com/Constellation-Labs/constellation-node-aws-ami) and [there](https://github.com/Constellation-Labs/constellation-cli).
2. Due to security policies, we cannot make a dependency on 3rd party like `mozilla/sbt:latest`.
3. The `run` script you provided is obsolete and already deprecated. We are using different configurations (env variables, file config), as well as opening jmxremote is a security leak. We can do that on our managed environments but it would be a serious issue once launched on mainnet.
4. In the roadmap there is a plan for docker support and orchestration, but we cannot do it right now, because the code is quickly evolving. The bouncy-castle related things has been changed today, the "run" script is also in progress, and we have no time for fixing it over and over during the first phase. We will provide the scripts and documentation when that part will be ready.

I really appreciate your effort, but I need to close this pull request. If you want to contribute to the project, we can find something out that can be done at this stage of the project.

esomore:
Thank you for the fast response. I will take a look on the other reps you linked.

PR #53, opened 2021-10-28
Trust implemented as Daemon using gossip.

ryle-goehausen:
Model code not yet enabled, so mostly a stub for now.
Add a configuration for trust with an interval for daemon. Daemon is structured
same as Gossip daemon in terms of classes used. Relies on trustStorage data assumed to be updated by gossip and updates internal storage with predictions.

This is for feature parity with existing codebase. Long term requirements will need model updates.

### DIFF
--- a/modules/core/src/main/scala/org/tesselation/infrastructure/trust/TrustDaemon.scala
+++ b/modules/core/src/main/scala/org/tesselation/infrastructure/trust/TrustDaemon.scala
@@ -1,0 +1,66 @@
+package org.tesselation.infrastructure.trust
+
+import cats.Parallel
+import cats.effect.std.Random
+import cats.effect.{Async, Spawn, Temporal}
+import cats.syntax.flatMap._
+import cats.syntax.functor._
+
+import org.tesselation.config.types.TrustDaemonConfig
+import org.tesselation.domain.Daemon
+import org.tesselation.domain.trust.storage.TrustStorage
+import org.tesselation.kryo.KryoSerializer
+import org.tesselation.schema.peer.PeerId
+import org.tesselation.schema.trust.TrustInfo
+import org.tesselation.security.SecurityProvider
+
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+
+trait TrustDaemon[F[_]] extends Daemon[F]
+
+object TrustDaemon {
+
+  def make[F[_]: Async: SecurityProvider: KryoSerializer: Random: Parallel](
+    cfg: TrustDaemonConfig,
+    trustStorage: TrustStorage[F],
+    selfPeerId: PeerId
+  ): TrustDaemon[F] = new TrustDaemon[F] {
+
+    private val logger = Slf4jLogger.getLogger[F]
+
+    def start: F[Unit] =
+      for {
+        _ <- Spawn[F].start(modelUpdate.foreverM).void
+      } yield ()
+
+    private def calculatePredictedTrust(trust: Map[PeerId, TrustInfo]): Map[PeerId, Double] = {
+      val selfTrustLabels = trust.flatMap { case (peerId, trustInfo) => trustInfo.publicTrust.map(peerId -> _) }
+      val allNodesTrustLabels = trust.view.mapValues(_.peerLabels).toMap + (selfPeerId -> selfTrustLabels)
+
+      val peerIdToIdx = allNodesTrustLabels.keys.zipWithIndex.toMap
+      val idxToPeerId = peerIdToIdx.map(_.swap)
+
+      val trustNodes = allNodesTrustLabels.map {
+        case (peerId, labels) =>
+          TrustNode(peerIdToIdx(peerId), 0, 0, labels.map {
+            case (pid, label) =>
+              TrustEdge(peerIdToIdx(peerId), peerIdToIdx(pid), label, peerId == selfPeerId)
+          }.toList)
+      }.toList
+
+      SelfAvoidingWalk
+        .runWalkFeedbackUpdateSingleNode(peerIdToIdx(selfPeerId), trustNodes)
+        .edges
+        .map(e => idxToPeerId(e.dst) -> e.trust)
+        .toMap
+    }
+
+    private def modelUpdate: F[Unit] =
+      for {
+        _ <- Temporal[F].sleep(cfg.interval)
+        predictedTrust <- trustStorage.getTrust.map(calculatePredictedTrust)
+        _ <- trustStorage.updatePredictedTrust(predictedTrust)
+      } yield ()
+
+  }
+}
--- a/modules/core/src/main/scala/org/tesselation/infrastructure/trust/storage/TrustStorage.scala
+++ b/modules/core/src/main/scala/org/tesselation/infrastructure/trust/storage/TrustStorage.scala
@@ -33,6 +33,21 @@ object TrustStorage {
       def getPublicTrust: F[PublicTrust] = getTrust.map { trust =>
         PublicTrust(trust.mapFilter { _.publicTrust })
       }
+
+      def updatePeerPublicTrustInfo(peerId: PeerId, publicTrust: PublicTrust): F[Unit] = trust.update { t =>
+        t.updatedWith(peerId) { trustInfo =>
+          trustInfo
+            .orElse(Some(TrustInfo()))
+            .map(_.copy(peerLabels = publicTrust.labels))
+        }
+      }
+
+      override def updatePredictedTrust(trustUpdates: Map[PeerId, Double]): F[Unit] = trust.update { t =>
+        t ++ trustUpdates.map {
+          case (k, v) =>
+            k -> t.getOrElse(k, TrustInfo()).copy(predictedTrust = Some(v))
+        }
+      }
     }
 
 }
--- a/modules/core/src/main/scala/org/tesselation/config/Config.scala
+++ b/modules/core/src/main/scala/org/tesselation/config/Config.scala
@@ -61,6 +61,14 @@ object Config {
     )
   )
 
+  val trustConfig = ConfigValue.default(
+    TrustConfig(
+      TrustDaemonConfig(
+        10.minutes
+      )
+    )
+  )
+
   def load[F[_]: Async]: F[AppConfig] =
     env("CL_APP_ENV")
       .default("testnet")
@@ -72,8 +80,8 @@ object Config {
       .load[F]
 
   def default[F[_]](environment: AppEnvironment): ConfigValue[F, AppConfig] =
-    (keyConfig, httpConfig, dbConfig, gossipConfig).parMapN { (key, http, db, gossip) =>
-      AppConfig(environment, key, http, db, gossip)
+    (keyConfig, httpConfig, dbConfig, gossipConfig, trustConfig).parMapN { (key, http, db, gossip, trust) =>
+      AppConfig(environment, key, http, db, gossip, trust)
     }
 
 }
--- a/modules/core/src/main/scala/org/tesselation/config/types.scala
+++ b/modules/core/src/main/scala/org/tesselation/config/types.scala
@@ -13,7 +13,8 @@ object types {
     keyConfig: KeyConfig,
     httpConfig: HttpConfig,
     dbConfig: DBConfig,
-    gossipConfig: GossipConfig
+    gossipConfig: GossipConfig,
+    trustConfig: TrustConfig
   )
 
   case class DBConfig(
@@ -49,6 +50,14 @@ object types {
   case class GossipConfig(
     storage: RumorStorageConfig,
     daemon: GossipDaemonConfig
+  )
+
+  case class TrustDaemonConfig(
+    interval: FiniteDuration
+  )
+
+  case class TrustConfig(
+    daemon: TrustDaemonConfig
   )
 
   case class HttpServerConfig(
--- a/modules/core/src/main/scala/org/tesselation/domain/trust/storage/TrustStorage.scala
+++ b/modules/core/src/main/scala/org/tesselation/domain/trust/storage/TrustStorage.scala
@@ -5,6 +5,8 @@ import org.tesselation.schema.trust.{InternalTrustUpdateBatch, PublicTrust, Trus
 
 trait TrustStorage[F[_]] {
   def updateTrust(trustUpdates: InternalTrustUpdateBatch): F[Unit]
+  def updatePredictedTrust(trustUpdates: Map[PeerId, Double]): F[Unit]
   def getTrust: F[Map[PeerId, TrustInfo]]
   def getPublicTrust: F[PublicTrust]
+  def updatePeerPublicTrustInfo(peerId: PeerId, publicTrust: PublicTrust): F[Unit]
 }
--- a/modules/core/src/main/scala/org/tesselation/http/p2p/P2PClient.scala
+++ b/modules/core/src/main/scala/org/tesselation/http/p2p/P2PClient.scala
@@ -2,7 +2,7 @@ package org.tesselation.http.p2p
 
 import cats.effect.Concurrent
 
-import org.tesselation.http.p2p.clients.{ClusterClient, GossipClient, SignClient}
+import org.tesselation.http.p2p.clients._
 import org.tesselation.kryo.KryoSerializer
 
 import org.http4s.client._
@@ -11,6 +11,7 @@ trait P2PClient[F[_]] {
   val sign: SignClient[F]
   val cluster: ClusterClient[F]
   val gossip: GossipClient[F]
+  val trust: TrustClient[F]
 }
 
 object P2PClient {
@@ -22,5 +23,6 @@ object P2PClient {
       val sign: SignClient[F] = SignClient.make[F](client)
       val cluster: ClusterClient[F] = ClusterClient.make[F](client)
       val gossip: GossipClient[F] = GossipClient.make[F](client)
+      val trust: TrustClient[F] = TrustClient.make[F](client)
     }
 }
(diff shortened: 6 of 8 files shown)


tbekas (dismissed):

modules/core/src/main/scala/org/tesselation/infrastructure/trust/TrustDaemon.scala
```diff
@@ -0,0 +1,53 @@
...
+    private def calculateTrust(): F[Unit] =
+      Async[F].pure(())
+
+    private def pollTrust: F[Unit] =
+      for {
+        _ <- Temporal[F].sleep(cfg.interval)
+        peers <- clusterStorage.getPeers
+        _ <- peers.toList.parTraverse { peer =>
+          p2pClient.trust.getPublicTrust.run(peer).map(pt => trustStorage.updatePeerPublicTrustInfo(peer.id, pt))
+        }
```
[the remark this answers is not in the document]
marcinwadon: > To clarify: what we need here is a chain of signatures

Chain of signatures is a thing required at a totally different layer and for different purpose. Here, it’s not needed at all, and the result will be the same, it’s still secured by the algorithms I’ve described above (replay attack not possible)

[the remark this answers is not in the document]
ryle-goehausen: @marcinwadon 

Re: signatures in gossip. Reviewed the code and I see it now, so that rules out some of the attacks I mentioned. 

>no, because we track each message and handle it exactly once only. and there is always a way to re-check and track it in the handler, when more security is needed.

Regarding this -- can you provide more detail exactly on where this is implemented (I am unfamiliar with this section of the code -- and I don't see it at a glance. I just see hash and signature validators, is this on a different branch?) Also -- can you verify if the hashes are based purely on the value of the data? Because if so, trust updates can actually have the same data multiple times, e.g. saying node_a 50%, node_a 40%, node_a 50% as three separate updates would have a duplicate for message 1 and 3. You mention the counters later on -- that's what I and Wyatt mean by a dependency hash chain. (Irrelevant whether it's a counter or a hash reference -- hash references are easier for verifying against storage)

I'm not arguing against the scalability problems here. I'd prefer to use a gossip based solution as long as we have an appropriate guarantee of defense against replays -- but the model as implemented is not complete, and should be prioritized. I imagine the broadcast is quite terrible for messages that need to be repeated much more often -- but trust updates can be incredibly slow and still provide security.

A partial solution for dealing with trust updates violating the state is simply to ban any node that signs an invalid chain -- that avoids dealing with any consensus at all. If there are two signatures of the same state it's relatively easy to prove that there's something wrong -- no agreement is necessary, which means this doesn't require a full implementation that connects to the rest of the code. Any update can simply check the previous update. 

Again, right now I'm still reviewing the replay attack protection to try to verify what you said above. I'll discuss prioritization of this with Wyatt. I was originally trying to just get feature parity so I can work on improvements to the core model. If the long delay on trust updates is insufficient, I can work on the final implementation of gossiped trust -- it doesn't require any other components if I implement 'lastHash' update essentially, but it's more work than feature parity.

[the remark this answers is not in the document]
tbekas: @ryle-goehausen 
>can you provide more detail exactly on where this is implemented (I am unfamiliar with this section of the code -- and I don't see it at a glance. I just see hash and signature validators, is this on a different branch?)

We track `seen` rumors in `RumorStorage`. If we receive a rumor we've already seen, we're just going to ignore it, the implementation of this filtering can be found in `setRumorsSeenAndGetUnseen` method.

It's worth mentioning that there's a retention time on hashes of seen rumors, currently set to 2 minutes - so currently in practice after 2 minutes it will be possible to replay the message and the receiver will process it as it was never seen before. We can solve this issue however with retention based on message counters instead of a time @marcinwadon is working on adding counters to the message.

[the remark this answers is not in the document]
marcinwadon: I'd like to add that having a chain of messages would require us to keep all the historical data. To avoid this, the Transaction and the Fiber forms a chain using counters as well (ordinal). That way we can quickly verify whether the transaction or fiber is correct in order, by comparing the ordinal number.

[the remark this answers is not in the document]
marcinwadon: @ryle-goehausen see: #56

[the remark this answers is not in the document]
ryle-goehausen: Hey guys, talked with Wyatt -- I'm going to change this PR to use the gossip since you are handling the sequence numbers (the replay defense.)

Re: retention time on hashes -- I'm going to just add a randomization factor to the trust message to get around the issue of a legitimate user changing a score back and forth -- so it doesn't accidentally generate the same legitimate hash twice and ignore the message.

>I'd like to add that having a chain of messages would require us to keep all the historical data. To avoid this, the Transaction and the Fiber forms a chain using counters as well (ordinal). 

For the other gossip messages with arbitrary complexity yes -- for a trust update we're already storing the latest version of it (so storing a hash is easy,) and it only ever compares one value. I fully agree though that this is the equivalent of a sequence number on gossip messages (which you have implemented.) I was thinking I was going to need to implement this logic myself -- which is why hash is easy to do as just 'lastTrustUpdate'. Sequence number works exactly the same in this case, since you're only comparing one value.

Libp2p has the exact same mechanism -- see here: https://github.com/libp2p/specs/pull/217/files so this has already been reviewed elsewhere and is valid.

ryle-goehausen:
@marcinwadon Updated with removal of poll and using gossip. Model still isn't turned on but I'll merge other changes next -- this is just covering the model daemon.

ryle-goehausen:
@marcinwadon ready for approval again, merged changes from dev.

modules/core/src/main/scala/org/tesselation/infrastructure/trust/TrustDaemon.scala
```diff
@@ -0,0 +1,74 @@
...
+  ): TrustDaemon[F] = new TrustDaemon[F] {
+
+    private val logger = Slf4jLogger.getLogger[F]
+
+    def start: F[Unit] =
+      for {
+        _ <- Spawn[F].start(modelUpdate.foreverM).void
+      } yield ()
+
+    private def calculateTrust(): F[Unit] = {
```
kpudlik, line 35: That method breaks the single responsibility principle because it not only calculates trust but also updates the trust.

My suggestion:

```scala
// Keep that one pure so we can unit test that. Also rename it so we know which trust it calculates.
private def calculatePredictedTrust(trustInfo: Map[PeerId, TrustInfo]): Map[PeerId, Double] { ... }

private def modelUpdate: F[Unit] = 
  for {
     _ <- Temporal[F].sleep(cfg.interval)
     predictedTrust <- trustStorage.getTrust.map(calculatePredictedTrust)
     _ <- trustStorage.updatePredictedTrust(predictedTrust)
  } yield ()
```

kpudlik (changes requested):

modules/core/src/main/scala/org/tesselation/infrastructure/trust/TrustDaemon.scala
```diff
@@ -0,0 +1,74 @@
...
+      val predicted = trustStorage.getTrust.map { trust =>
+        val fixed = trust.toList
+        val selfTrust = fixed.flatMap { case (peerId, trustInfo) => trustInfo.publicTrust.map(t => peerId -> t) }
+        val allNodes = fixed.map {
+          case (peerId, trustInfo) =>
+            peerId -> trustInfo.peerLabels
+        } :+ (selfPeerId -> selfTrust.toMap)
+        val peerIdToIdx = allNodes.zipWithIndex.map { case ((peerId, _), idx) => peerId -> idx }.toMap
+        val idxToPeerId = peerIdToIdx.map { case (k, v) => v -> k }
```
kpudlik, line 44: A little bit more elegant way ;)

```scala
val idxToPeerId = peerIdToIdx.map(_.swap)
```

kpudlik (changes requested):

modules/core/src/main/scala/org/tesselation/infrastructure/trust/TrustDaemon.scala
kpudlik, line 44: Converting `Map` to `List` seems redundant because you do this in the calculation flow:

`Map[k, v]` -> `List[(k, v)]` -> `Map[k, v]` -> `List[k, v]` -> `Map[k,v]`

but we could do it without intermediate conversions:

`Map[k,v]` -> List[(k, v)] -> `Map[k, v]`

Also, you could use `mapValues` in some places as key (`peerId`) stays the same. That being said whole pure function for calculating predicted trust (see my other comments) could look like this:

```scala
        val selfTrustLabels = trust.flatMap { case (peerId, trustInfo) => trustInfo.publicTrust.map(peerId -> _) } // Map
        val allNodesTrustLabels = trust.view.mapValues(_.peerLabels).toMap + (selfPeerId -> selfTrustLabels) // Map

        val peerIdToIdx = allNodesTrustLabels.keys.zipWithIndex.toMap // Map
        val idxToPeerId = peerIdToIdx.map(_.swap) // Map

        val trustNodes = allNodesTrustLabels.map {
          case (peerId, labels) =>
            TrustNode(peerIdToIdx(peerId), 0, 0, labels.map {
              case (pid, label) =>
                TrustEdge(peerIdToIdx(peerId), peerIdToIdx(pid), label, peerId == selfPeerId)
            }.toList)
        }.toList // Map -> List because SAW needs List
        
        SelfAvoidingWalk
          .runWalkFeedbackUpdateSingleNode(peerIdToIdx(selfPeerId), trustNodes)
          .edges
          .map(e => idxToPeerId(e.dst) -> e.trust)
          .toMap // List -> Map
```

modules/core/src/main/scala/org/tesselation/infrastructure/trust/TrustDaemon.scala
```diff
@@ -0,0 +1,74 @@
...
+        val idMappedScores: Map[PeerId, Double] = SelfAvoidingWalk
```
kpudlik, line 45: `val idMappedScored` is redundant because you immediately return it in line 59

kpudlik (changes requested):

modules/core/src/main/scala/org/tesselation/infrastructure/trust/TrustDaemon.scala
ryle-goehausen: Fixed. Going to start adding tests for this in next PRs as well.

modules/core/src/main/scala/org/tesselation/infrastructure/trust/TrustDaemon.scala
ryle-goehausen: Fixed

ryle-goehausen: Fixed

modules/core/src/main/scala/org/tesselation/infrastructure/trust/TrustDaemon.scala
ryle-goehausen: Fixed

kpudlik (approved):

modules/core/src/main/scala/org/tesselation/infrastructure/trust/TrustDaemon.scala
```diff
@@ -0,0 +1,66 @@
...
+
+    private val logger = Slf4jLogger.getLogger[F]
+
+    def start: F[Unit] =
+      for {
+        _ <- Spawn[F].start(modelUpdate.foreverM).void
+      } yield ()
+
+    private def calculatePredictedTrust(trust: Map[PeerId, TrustInfo]): Map[PeerId, Double] = {
+      val selfTrustLabels = trust.flatMap { case (peerId, trustInfo) => trustInfo.publicTrust.map(peerId -> _) } // Map
```
kpudlik, line 37: Remove these `// Map` comments please :) These were just for code review purposes

modules/core/src/main/scala/org/tesselation/infrastructure/trust/TrustDaemon.scala
```diff
@@ -0,0 +1,66 @@
...
+      val allNodesTrustLabels = trust.view.mapValues(_.peerLabels).toMap + (selfPeerId -> selfTrustLabels) // Map
```
kpudlik, line 38: Remove these `// Map` comments please :) These were just for code review purposes

modules/core/src/main/scala/org/tesselation/infrastructure/trust/TrustDaemon.scala
```diff
@@ -0,0 +1,66 @@
...
+
+      val peerIdToIdx = allNodesTrustLabels.keys.zipWithIndex.toMap // Map
```
kpudlik, line 40: Remove these `// Map` comments please :) These were just for code review purposes

modules/core/src/main/scala/org/tesselation/infrastructure/trust/TrustDaemon.scala
```diff
@@ -0,0 +1,66 @@
...
+      val idxToPeerId = peerIdToIdx.map(_.swap) // Map
```
kpudlik, line 41: Remove these `// Map` comments please :) These were just for code review purposes

modules/core/src/main/scala/org/tesselation/infrastructure/trust/TrustDaemon.scala
```diff
@@ -0,0 +1,66 @@
...
+
+      val trustNodes = allNodesTrustLabels.map {
+        case (peerId, labels) =>
+          TrustNode(peerIdToIdx(peerId), 0, 0, labels.map {
+            case (pid, label) =>
+              TrustEdge(peerIdToIdx(peerId), peerIdToIdx(pid), label, peerId == selfPeerId)
+          }.toList)
+      }.toList // Map -> List because SAW needs List
```
kpudlik, line 49: Remove these `// Map` comments please :) These were just for code review purposes

modules/core/src/main/scala/org/tesselation/infrastructure/trust/TrustDaemon.scala
```diff
@@ -0,0 +1,66 @@
...
+
+      SelfAvoidingWalk
+        .runWalkFeedbackUpdateSingleNode(peerIdToIdx(selfPeerId), trustNodes)
+        .edges
+        .map(e => idxToPeerId(e.dst) -> e.trust)
+        .toMap // List -> Map
```
kpudlik, line 55: Remove these `// Map` comments please :) These were just for code review purposes

kpudlik (changes requested):
Everything is ok but these comments which I put in sample code could be removed :)

modules/core/src/main/scala/org/tesselation/infrastructure/trust/storage/TrustStorage.scala
```diff
@@ -33,6 +33,17 @@ object TrustStorage {
...
+
+      def updatePeerPublicTrustInfo(peerId: PeerId, publicTrust: PublicTrust): F[Unit] = trust.update { t =>
+        t ++ Map(peerId -> t.getOrElse(peerId, TrustInfo()).copy(peerLabels = publicTrust.labels))
+      }
+
+      override def updatePredictedTrust(trustUpdates: Map[PeerId, Double]): F[Unit] = trust.update { t =>
+        t ++ trustUpdates.map {
+          case (k, v) =>
+            k -> t.getOrElse(k, TrustInfo()).copy(predictedTrust = Some(v))
+        }
```
kpudlik, line 45: We could use `mapValues` here as key stays the same

```scala
t ++ trustUpdates.view.mapValues { predictedTrust =>
  TrustInfo(predictedTrust = Some(predictedTrust))
}
```

modules/core/src/main/scala/org/tesselation/infrastructure/trust/storage/TrustStorage.scala
```diff
@@ -33,6 +33,17 @@ object TrustStorage {
       def getPublicTrust: F[PublicTrust] = getTrust.map { trust =>
         PublicTrust(trust.mapFilter { _.publicTrust })
       }
+
+      def updatePeerPublicTrustInfo(peerId: PeerId, publicTrust: PublicTrust): F[Unit] = trust.update { t =>
+        t ++ Map(peerId -> t.getOrElse(peerId, TrustInfo()).copy(peerLabels = publicTrust.labels))
```
kpudlik, line 38: There is a nice function `updatedWith` for `put or update` case :)
```scala
t.updatedWith(peerId) { trustInfo =>
  Some(trustInfo.getOrElse(TrustInfo()).copy(peerLabels = publicTrust.labels))
}
```

kpudlik (changes requested):

modules/core/src/main/scala/org/tesselation/infrastructure/trust/storage/TrustStorage.scala
ryle-goehausen: Fixed

modules/core/src/main/scala/org/tesselation/infrastructure/trust/TrustDaemon.scala
ryle-goehausen: Fixed all these.

kpudlik (approved):

PR #55, opened 2021-11-02
ACI Classpath loading

tbekas:


modules/core/src/main/scala/org/tesselation/domain/aci/StateChannelContext.scala
```diff
@@ -0,0 +1,8 @@
+package org.tesselation.domain.aci
+
+import org.tesselation.schema.address.Address
+
+trait StateChannelContext[F[_]] {
+  val address: Address
+  def createCell(input: Array[Byte]): F[StdCell[F]]
+}
```
marcinwadon, line 8: IMO it shouldn't be in the `core` module. probably `kernel` is a better place for this
@kpudlik @TheMMaciek ?

modules/core/src/main/scala/org/tesselation/infrastructure/aci/StateChannelInputPipeline.scala
```diff
@@ -0,0 +1,34 @@
...
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+import cats.syntax.flatMap._
+
+class StateChannelInputPipeline[F[_]: Async](
+  stateChannelInputQueue: Queue[F, StateChannelInput],
+  stateChannelContexts: Map[Address, StateChannelContext[F]]
+) {
+  val logger = Slf4jLogger.getLogger[F]
+
+  Stream.fromQueueUnterminated(stateChannelInputQueue)
```
marcinwadon, line 19: ?

modules/core/src/main/scala/org/tesselation/domain/aci/StateChannelContext.scala
tbekas: I think that the `kernel` module should contain only the stuff necessary for the users to create the StateChannel. This context is an internal thing, that gets created when loading a jar.

modules/core/src/main/scala/org/tesselation/infrastructure/aci/StateChannelInputPipeline.scala
tbekas: Yep, I'm addressing it :)

modules/core/src/main/scala/org/tesselation/domain/aci/StateChannelContext.scala
marcinwadon: but related to state channels 🤔 let's discuss it on slack

marcinwadon (approved):

PR #56, opened 2021-11-02
Gossip counter and session verification

marcinwadon:


modules/core/src/main/scala/org/tesselation/infrastructure/gossip/GossipDaemon.scala
```diff
@@ -86,6 +92,27 @@ object GossipDaemon {
...
+    private def validateSession(batch: RumorBatch): F[RumorBatch] =
+      batch.filterA {
+        case (hash, signedRumor) =>
+          signedRumor.value.origin match {
+            case ownId if ownId === nodeId => true.pure[F]
+            case peerId =>
+              clusterStorage
+                .getPeer(peerId)
+                .map {
+                  _.fold(false) { _.session === signedRumor.value.session }
```
kpudlik, line 104: Or 'forall'

kpudlik (approved):

modules/shared/src/main/scala/org/tesselation/schema/gossip.scala
```diff
@@ -20,9 +23,18 @@ object gossip {
   case class Rumor(
     tpe: String,
     origin: PeerId,
+    counter: Long,
```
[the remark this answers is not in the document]
marcinwadon: I will use `PosLong`

modules/shared/src/main/scala/org/tesselation/crypto/Signed.scala
```diff
@@ -22,6 +23,8 @@ case class Signed[A](value: A, proofs: NonEmptyList[SignatureProof])
 
 object Signed {
 
+  implicit def order[A: Order]: Order[Signed[A]] = (x: Signed[A], y: Signed[A]) => x.value.compare(y.value)
```
tbekas, line 26: Nice

modules/core/src/main/scala/org/tesselation/infrastructure/gossip/RumorHandler.scala
```diff
@@ -17,30 +19,44 @@ import org.tesselation.schema.peer.PeerId
...
-  ): RumorHandler[F] = fromBiFn((id: PeerId, a: A) => f(ReceivedRumor(id, a)))
-
-  def fromFn[F[_]: Async: KryoSerializer, A <: AnyRef: TypeTag: ClassTag](
-    f: A => F[Unit]
-  ): RumorHandler[F] = fromBiFn((_: PeerId, a: A) => f(a))
-
-  def fromBiFn[F[_]: Async: KryoSerializer, A <: AnyRef: TypeTag: ClassTag](
-    f: (PeerId, A) => F[Unit]
-  ): RumorHandler[F] = {
+    rumorStorage: RumorStorage[F],
```
tbekas, line 22: I dont' think that `RumorStorage` should be an argument of any handler creating function. I think that we should use kleisli functionality and pass the storage along with the rumor, so that in places where you're creating handlers you don't need to depend on the `RumorStorage`.

Basically we would replace this:
`type RumorHandler[F[_]] = Kleisli[OptionT[F, *], Rumor, Unit]`
with this:
`type RumorHandler[F[_]] = Kleisli[OptionT[F, *], (Rumor, RumorStorage), Unit]`

modules/core/src/main/scala/org/tesselation/domain/gossip/RumorStorage.scala
```diff
@@ -13,4 +14,8 @@ trait RumorStorage[F[_]] {
 
   def getSeenHashes: F[List[Hash]]
 
+  def tryUpdateCounter(rumor: Rumor): F[Option[Long]]
```
tbekas, line 17: I think it should just return `F[Boolean]`

tbekas (changes requested):

modules/core/src/main/scala/org/tesselation/infrastructure/gossip/GossipDaemon.scala
```diff
@@ -53,10 +57,12 @@ object GossipDaemon {
         .fromQueueUnterminated(rumorQueue)
         .evalMap(validateHash)
         .evalMap(validateSignature)
+        .evalMap(validateSession)
```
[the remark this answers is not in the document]
marcinwadon: IMO we should filter it out. Let's say - there is a node state being gossiped, node rejoined, I received new node state gossip first (let's say Ready state) and then I receive old one from the other session. It must be ignored.

Hmm 🤔 I'm trying to think about possible scenario when it fails

modules/core/src/main/scala/org/tesselation/infrastructure/cluster/storage/ClusterStorage.scala
```diff
@@ -27,6 +27,9 @@ object ClusterStorage {
       def getPeers: F[Set[Peer]] =
         peers.keys.flatMap(_.map(peers(_).get).sequence).map(_.flatten.toSet)
 
+      def getPeer(id: PeerId): F[Option[Peer]] =
```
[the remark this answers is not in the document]
marcinwadon: It's not always useful. When you keep all your outputs like this, then fine. (like EitherT in our previous SnapshotService class). When it's for a single usage only, it's pointless to wrap it there into monad transformer, because you need to unwrap it anyway (because we're not using OptionT flows, but F only)

modules/core/src/main/scala/org/tesselation/infrastructure/gossip/RumorStorage.scala
```diff
@@ -21,11 +22,13 @@ object RumorStorage {
     for {
       active <- MapRef.ofConcurrentHashMap[F, Hash, Signed[Rumor]]()
       seen <- MapRef.ofConcurrentHashMap[F, Hash, Unit]()
-    } yield make(active, seen, cfg)
+      seenCounter <- MapRef.ofConcurrentHashMap[F, PeerId, Map[String, Long]]()
```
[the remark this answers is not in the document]
marcinwadon: Why `(PeerId, String)`? There will always be the same peer id

modules/core/src/main/scala/org/tesselation/domain/gossip/RumorStorage.scala
marcinwadon: `tryGetAndUpdateCounter` then?

modules/core/src/main/scala/org/tesselation/infrastructure/gossip/GossipDaemon.scala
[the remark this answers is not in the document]
tbekas: Snapshot proposals for example. Node gossips proposal at height K, rejoins at K+1 and proposals at K are ignored.

modules/core/src/main/scala/org/tesselation/domain/gossip/RumorStorage.scala
tbekas: Well if the method will return `Boolean` instead of an `Option[Long]` the `Get` verb doesn't fit, so I would stay with the old name.

marcinwadon: No, I need an `Option[Long]`. Same as original method `getAndUpdate` does

modules/core/src/main/scala/org/tesselation/infrastructure/cluster/storage/ClusterStorage.scala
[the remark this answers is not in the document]
tbekas: Hm, ok let's keep it in `F` then.

modules/core/src/main/scala/org/tesselation/infrastructure/gossip/GossipDaemon.scala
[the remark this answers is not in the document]
marcinwadon: To sum up the discussion: JIRA task created, scheduled for the next sprint.

kpudlik (dismissed):

kpudlik (approved):

kpudlik (dismissed):

modules/core/src/main/scala/org/tesselation/infrastructure/gossip/GossipDaemon.scala
```diff
@@ -53,10 +57,12 @@ object GossipDaemon {
...
         .evalMap(rumorStorage.addRumors)
-        .flatMap(batch => Stream.iterable(batch))
+        .map(_.sortBy(_._2))
+        .flatMap(Stream.iterable)
         .filter { case (_, signedRumor) => signedRumor.value.origin =!= nodeId }
-        .parEvalMapUnordered(cfg.maxConcurrentHandlers)(handleRumor)
+        .parEvalMap(cfg.maxConcurrentHandlers)(handleRumor)
```
tbekas, line 65: If we don't group by type we cannot `parEvalMap`, because the ordering of handled messages won't be preserved. Let's revert it to a regular `evalMap` and we will figure it out later on how to parallelize message types.

tbekas (changes requested):

tbekas (approved):
LGTM

kpudlik (approved):

PR #57, opened 2021-11-03
Create transaction method for wallet

kpudlik:


modules/wallet/src/main/scala/org/tesselation/wallet/Main.scala
```diff
@@ -28,22 +41,26 @@ object Main
...
+                  showAddress[IO](keyPair)
+                    .handleErrorWith(err => logger.error(err)(s"Error while showing address."))
+                    .as(ExitCode.Success)
+                case ShowId() =>
+                  showId[IO](keyPair)
+                    .handleErrorWith(err => logger.error(err)(s"Error while showing id."))
+                    .as(ExitCode.Success)
+                case ShowPublicKey() =>
+                  showPublicKey[IO](keyPair)
+                    .handleErrorWith(err => logger.error(err)(s"Error while showing id."))
```
marcinwadon, line 57: Error while showing public key.

modules/wallet/src/main/scala/org/tesselation/wallet/Main.scala
```diff
@@ -55,6 +72,33 @@ object Main
...
+  private def createAndStoreTransaction[F[_]: Async: SecurityProvider: KryoSerializer](
+    keyPair: KeyPair,
+    destination: Address,
+    fee: TransactionFee,
+    amount: TransactionAmount,
+    prevTxPath: Option[Path],
+    nextTxPath: Path
+  ): F[Unit] =
+    for {
+      logger <- Slf4jLogger.getLogger[F].pure[F]
```
marcinwadon, line 84: You should use `createLogger` which is safe and already delayed in F

modules/wallet/src/main/scala/org/tesselation/wallet/transaction/package.scala
```diff
@@ -0,0 +1,51 @@
...
+import eu.timepit.refined.auto._
+import eu.timepit.refined.types.all.NonNegBigInt
+import io.estatico.newtype.ops._
+
+package object transaction {
+
+  implicit class TransactionOrdinalOps(ordinal: TransactionOrdinal) {
+
+    def next: TransactionOrdinal =
+      TransactionOrdinal(NonNegBigInt.unsafeFrom(ordinal.coerce + 1))
```
marcinwadon, line 27: have you tried `NonNegBitInt(ordinal.coerce + 1)` or `ordinal |+| 1`?

modules/wallet/src/main/scala/org/tesselation/wallet/transaction/package.scala
```diff
@@ -0,0 +1,51 @@
...
+
+  def createTransaction[F[_]: Async: KryoSerializer: SecurityProvider](
+    keyPair: KeyPair,
+    destination: Address,
+    prevTx: Option[Signed[Transaction]],
+    fee: TransactionFee,
+    amount: TransactionAmount
+  ): F[Signed[Transaction]] =
+    for {
+      source <- Async[F].delay(keyPair.getPublic.toAddress)
```
marcinwadon, line 38: is it a side effect? if no, you can use `pure`

marcinwadon (changes requested):

modules/wallet/src/main/scala/org/tesselation/wallet/transaction/package.scala
kpudlik: There is no `Semigroup[NonNegBitInt]` in refined.cats :<

The first one will obviously throw at compile time because it is not literal (standard apply method works only for literals)

kpudlik: Oh wait there is a `Monoid` ! :)

TheMMaciek: If there is please remove the `semigroup` for `NonNegBigInt` from `OrphanInstances` trait.

kpudlik: Ok there is not. It was for standard int

marcinwadon (approved):

PR #59, opened 2021-11-03
Switch from scopt to decline in keytool

kpudlik:


modules/shared/src/main/scala/org/tesselation/cli/env.scala
```diff
@@ -0,0 +1,77 @@
...
+
+    val opts: Opts[StorePath] = Opts
+      .env[Path]("CL_KEYSTORE", help = "Keystore path")
+      .map(_.coerce)
+  }
+
+  case class EnvConfig(
+    keystore: StorePath,
+    storepass: StorePass,
+    keypass: KeyPass,
```
marcinwadon, line 66: can we unify it into `password` or we need it to support legacy keys?

marcinwadon (dismissed):

modules/shared/src/main/scala/org/tesselation/cli/env.scala
kpudlik: We need to support legacy keys in `wallet` but we force `password` in `keytool` already

marcinwadon (approved):

PR #61, opened 2021-11-04
Add TrustPush program connecting trust update with gossip

ryle-goehausen:
Simple program takes internal updates and publishes latest PublicTrust
snapshot. Avoided incremental updates to keep it simpler as data is
extremely small (labels and observations only.)

Triggered from CLI trust update after storage is updated. Spreads
gossip and add an empty handler which will be used in subsequent PR.

Adds randomization factor to handle issues with reversions in scores
to avoid problem with seenHashes. Removed @newtype due to addition
of UUID for randomness.

### DIFF
--- a/modules/core/src/main/scala/org/tesselation/domain/cluster/programs/TrustPush.scala
+++ b/modules/core/src/main/scala/org/tesselation/domain/cluster/programs/TrustPush.scala
@@ -1,0 +1,30 @@
+package org.tesselation.domain.cluster.programs
+
+import cats.effect.Async
+import cats.syntax.flatMap._
+import cats.syntax.functor._
+
+import org.tesselation.domain.gossip.Gossip
+import org.tesselation.domain.trust.storage.TrustStorage
+
+object TrustPush {
+
+  def make[F[_]: Async](
+    trustStorage: TrustStorage[F],
+    gossip: Gossip[F]
+  ): TrustPush[F] = new TrustPush[F](trustStorage, gossip) {}
+
+}
+
+sealed abstract class TrustPush[F[_]: Async] private (
+  trustStorage: TrustStorage[F],
+  gossip: Gossip[F]
+) {
+
+  def publishUpdated(): F[Unit] =
+    for {
+      trust <- trustStorage.getPublicTrust
+      _ <- gossip.spread(trust)
+    } yield {}
+
+}
--- a/modules/core/src/main/scala/org/tesselation/infrastructure/trust/handler.scala
+++ b/modules/core/src/main/scala/org/tesselation/infrastructure/trust/handler.scala
@@ -1,0 +1,31 @@
+package org.tesselation.infrastructure.trust
+
+import cats.Applicative
+import cats.effect.Async
+import cats.syntax.flatMap._
+import cats.syntax.show._
+
+import org.tesselation.domain.trust.storage.TrustStorage
+import org.tesselation.infrastructure.gossip.RumorHandler
+import org.tesselation.kryo.KryoSerializer
+import org.tesselation.schema.gossip.ReceivedRumor
+import org.tesselation.schema.trust.PublicTrust
+
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+
+object handler {
+
+  def trustHandler[F[_]: Async: KryoSerializer](
+    trustStorage: TrustStorage[F]
+  ): RumorHandler[F] = {
+    val logger = Slf4jLogger.getLogger[F]
+
+    RumorHandler.fromReceivedRumorFn[F, PublicTrust](latestOnly = true) {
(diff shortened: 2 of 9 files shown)


modules/core/src/main/scala/org/tesselation/domain/cluster/programs/TrustPush.scala
```diff
@@ -0,0 +1,34 @@
...
+import org.tesselation.effects.GenUUID
+
+object TrustPush {
+
+  def make[F[_]: Async: GenUUID](
+    trustStorage: TrustStorage[F],
+    gossip: Gossip[F]
+  ): F[TrustPush[F]] = {
+    val push = new TrustPush[F](trustStorage, gossip) {}
+    Async[F].delay(push)
```
marcinwadon, line 18: It's pure, so it's pointless to wrap it into delay

modules/shared/src/main/scala/org/tesselation/schema/trust.scala
```diff
@@ -30,7 +31,14 @@ object trust {
...
   @derive(decoder, encoder, show)
-  @newtype
-  case class PublicTrust(labels: Map[PeerId, Double])
+  case class PublicTrust(
+    labels: Map[PeerId, Double],
+    /**
+      * Randomization is only required in the event of gossiped and signed data to prevent
+      * reversions from being filtered out. When requested from a node directly it is
+      * unnecessary.
+      */
```
marcinwadon, line 40: Can you explain it more? It sounds it makes no sense. Gossip and direct request are interchangeable and transparent here in that layer.

marcinwadon (changes requested):

modules/core/src/main/scala/org/tesselation/domain/cluster/programs/TrustPush.scala
tbekas: Yep, I think that `make` should return just `TrustPath[F]`.

modules/core/src/main/scala/org/tesselation/infrastructure/trust/handler.scala
```diff
@@ -0,0 +1,30 @@
...
+  def trustHandler[F[_]: Async: KryoSerializer](
+    trustStorage: TrustStorage[F]
+  ): RumorHandler[F] = {
+    val logger = Slf4jLogger.getLogger[F]
+
+    RumorHandler.fromReceivedRumorFn[F, PublicTrust](latestOnly = true) {
+      case ReceivedRumor(origin, trust) =>
+        logger.info(s"Received trust=${trust} from id=${origin.show}") >> {
+          // Placeholder for updating trust scores
+          Async[F].delay(())
```
tbekas, line 26: You can replace it with `Applicative[F].unit`

modules/shared/src/main/scala/org/tesselation/schema/trust.scala
tbekas: @ryle-goehausen I think after recent addition of counters by @marcinwadon it's not necessary. Spread messages are unique because counter is unique so you can spread twice the same content if you need to do so.

ryle-goehausen: As long as you guys can verify removing this is fine, that's fine with me. I remember in earlier comments it was mentioned that you're filtering repeats by recently seen hashes -- but if you do the counter comparison before the filter stage I can remove this -- or if it doesn't actually impact it that works. It seems in code like the hash is being computed and used without reference to the counter, but if that doesn't matter then I can just remove it.

marcinwadon: Give it a try and check on cluster. IMO it will produce different rumor hashes, and there is a counter as well.

modules/core/src/main/scala/org/tesselation/infrastructure/trust/handler.scala
ryle-goehausen: Fixed

modules/shared/src/main/scala/org/tesselation/schema/trust.scala
ryle-goehausen: Removed this but had to leave it as kryo class due to rumor type signature:

>/Users/ryle/projects/tessellation/modules/core/src/main/scala/org/tesselation/domain/cluster/programs/TrustPush.scala:29:19
inferred type arguments [org.tesselation.schema.trust.PublicTrust] do not conform to method spread's type parameter bounds [A <: AnyRef]
      _ <- gossip.spread(trust)

Doesn't with work @newtype

marcinwadon (approved):

tbekas (approved):
LGTM

PR #63, opened 2021-11-09
DAG L1 with validation

TheMMaciek:


modules/core/src/main/scala/org/tesselation/domain/cluster/storage/AddressStorage.scala
```diff
@@ -1,12 +1,21 @@
...
 import org.tesselation.schema.balance.Balance
 
+sealed trait DBUpsertAction[A] { val value: A }
+case class Insert[A](value: A) extends DBUpsertAction[A]
+case class Update[A](value: A) extends DBUpsertAction[A]
+
 trait AddressStorage[F[_]] {
   def getBalance(address: Address): F[Balance]
 
+  def getMaybeBalance(address: Address): F[Option[Balance]]
```
marcinwadon, line 14: can we combine it into a single method?
can be there `getOrElse(0L)`? or just throw an error if the get query failed 🤔

modules/core/src/main/scala/org/tesselation/domain/cluster/storage/AddressStorage.scala
```diff
@@ -1,12 +1,21 @@
 package org.tesselation.domain.cluster.storage
 
+import org.tesselation.infrastructure.db.schema.StoredAddress
```
marcinwadon, line 3: you cannot import `infrastructure` into `domain`. if so, `StoredAddress` should be moved and/or renamed

modules/core/src/main/scala/org/tesselation/domain/cluster/storage/AddressStorage.scala
```diff
@@ -1,12 +1,21 @@
...
 import org.tesselation.schema.address.Address
 import org.tesselation.schema.balance.Balance
 
+sealed trait DBUpsertAction[A] { val value: A }
+case class Insert[A](value: A) extends DBUpsertAction[A]
+case class Update[A](value: A) extends DBUpsertAction[A]
```
marcinwadon, line 9: implementation leak. it can be stored in memory as well. you can remove `db` and keep it just as `UpsertAction` probably.

modules/dag/src/main/scala/org/tesselation/dag/l1/DAGBlock.scala
```diff
@@ -0,0 +1,21 @@
...
+
+import org.tesselation.schema.L1Block
+import org.tesselation.schema.height.Height
+import org.tesselation.schema.transaction.Transaction
+import org.tesselation.security.hash.Hash
+import org.tesselation.security.signature.Signed
+
+import io.estatico.newtype.ops.toCoercibleIdOps
+
+case class BlockReference(hash: Hash, height: Height)
```
marcinwadon, line 13: isn't it a Fiber?

modules/dag/src/main/scala/org/tesselation/dag/l1/DAGBlockValidator.scala
```diff
@@ -0,0 +1,170 @@
...
+  blockStorage: BlockStorage[F],
+  addressStorage: AddressStorage[F],
+  transactionValidator: TransactionValidator[F]
+) extends BlockValidator[F, DAGBlock] {
+
+  override def accept(signedBlock: Signed[DAGBlock]): F[Unit] =
+    for {
+      _ <- validateBlockSignatures(signedBlock).ifM(
+        Async[F].unit,
+        Async[F].raiseError[Unit](new Throwable("Invalid block signatures!"))
```
marcinwadon, line 39: Applicative, ApplicativeError, MonadThrow

modules/dag/src/main/scala/org/tesselation/dag/l1/DAGBlockValidator.scala
```diff
@@ -0,0 +1,170 @@
...
+      )
+      _ <- validateBlock(signedBlock.value).ifM(
+        Async[F].unit,
+        Async[F].raiseError[Unit](new Throwable("Invalid block data!"))
+      )
+      hashedBlock <- signedBlock.hashWithSignatureCheck.flatMap(_.liftTo[F])
```
marcinwadon, line 45: lift swallows errors. probably it should be attemptT, rethrowT, or sth like that

modules/dag/src/main/scala/org/tesselation/dag/l1/DAGBlockValidator.scala
```diff
@@ -0,0 +1,170 @@
...
+    for {
+      arePassingValidation <- areTransactionsPassingValidation(signedTransactions)
+      areChainedCorrectly <- areAllTransactionsFormingCorrectChain(signedTransactions.map(_.value))
+    } yield arePassingValidation && areChainedCorrectly
+
+  private def areTransactionsPassingValidation(signedTransactions: Seq[Signed[Transaction]]): F[Boolean] =
+    transactionValidator
+      .validateBatchOfTransactions(signedTransactions)
+      .map(_.isValid)
+      .handleError(_ => false)
```
marcinwadon, line 95: error logging

modules/dag/src/main/scala/org/tesselation/dag/l1/DAGBlockValidator.scala
```diff
@@ -0,0 +1,170 @@
...
+
+  private def areAllTransactionsFormingCorrectChain(transactions: Seq[Transaction]): F[Boolean] =
+    transactions
+      .groupBy(_.source)
+      .toList
+      .traverse {
+        case (address, txs) =>
+          for {
+            lastAccepted <- transactionStorage.getLastAcceptedTransactionRef(address)
+            sorted = txs.sortBy(_.parent.ordinal.coerce.value)
```
marcinwadon, line 105: you should implement Order instance for ordinal, and then just sort by ordinal

modules/dag/src/main/scala/org/tesselation/dag/l1/DAGBlockValidator.scala
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  private def prepareBalances(transactions: Seq[Transaction]): F[Seq[DBUpsertAction[StoredAddress]]] =
+    transactions.traverse { tx =>
+      for {
+        sourceBalance <- addressStorage.getBalance(tx.source)
+        newSourceBalance <- refineV[NonNegative]
+          .apply(sourceBalance.coerce.value - tx.amount.coerce.value)
+          .bimap(new Throwable(_), Balance(_))
+          .liftTo[F]
```
marcinwadon, line 140: aa

modules/dag/src/main/scala/org/tesselation/dag/l1/L1Routes.scala
```diff
@@ -0,0 +1,45 @@
...
+        for {
+          signedTransaction <- req.as[Signed[Transaction]]
+          validation <- transactionValidator.validateSingleTransaction(signedTransaction)
+          response <- validation match {
+            case Valid(signedTx) =>
+              transactionQueue
+              // TODO: we could move the queue out of the L1Routes and instead have it in TransactionManager along with validator
+                .offer(signedTx)
+                // TODO: I don't like it regenareting hash here, maybe validator should returned a Hashed[Transaction]
+                .flatMap(_ => signedTx.value.hashF)
```
marcinwadon, line 37:
```
transactionService.process(signedTx): F[Hashed[Transaction]]
```
???

modules/dag/src/main/scala/org/tesselation/dag/l1/TransactionValidator.scala
```diff
@@ -0,0 +1,238 @@
...
+    }.map(_.forall(identity))
+      .ifM(
+        Async[F].pure(signedTransaction.value.validNel),
+        Async[F].pure(SourceAddressAndSignerIdsDontMatch(signedTransaction.value).invalidNel)
+      )
+
+  private def validateSourceSignature(signedTransaction: Signed[Transaction]): F[ValidationResult[Transaction]] =
+    signedTransaction.hasValidSignature.ifM(
+      Async[F].pure(signedTransaction.value.validNel),
+      Async[F].pure(InvalidSourceSignature.invalidNel)
```
marcinwadon, line 119: `.pure[F]`

modules/dag/src/main/scala/org/tesselation/dag/l1/storage/BlockStorage.scala
```diff
@@ -0,0 +1,28 @@
+package org.tesselation.dag.l1.storage
+
+import cats.effect.Sync
```
marcinwadon, line 3: Async is a base type now

marcinwadon (changes requested):

modules/dag/src/main/scala/org/tesselation/dag/l1/DAGBlock.scala
```diff
@@ -0,0 +1,21 @@
...
+
+case class DAGBlock(transactions: Seq[Signed[Transaction]], parent: NonEmptyList[BlockReference]) extends L1Block {
+
+  // TODO: looks ugly
+  // TODO: confirm if we should increment the lower or higher parent height value
```
ryle-goehausen, line 18: Probably need someone to double check that this is still using same definition as earlier -- but height should be a max. It's supposed to represent how many references are required to trace back to some origin. With 2 parents of height 10 and 20, you'd need to trace back a maximum of (20+1)=21 references for the new child

modules/dag/src/main/scala/org/tesselation/dag/l1/DAGBlockValidator.scala
```diff
@@ -0,0 +1,170 @@
...
+      )
+      _ <- validateBlock(signedBlock.value).ifM(
+        Async[F].unit,
+        Async[F].raiseError[Unit](new Throwable("Invalid block data!"))
+      )
+      hashedBlock <- signedBlock.hashWithSignatureCheck.flatMap(_.liftTo[F])
+
+      hashedTransactions <- signedBlock.value.transactions.traverse(_.hashWithSignatureCheck.flatMap(_.liftTo[F]))
+
+      _ <- acceptTransactions(hashedTransactions)
```
ryle-goehausen, line 49: The transactions and block are being accepted separately here, what happens if there's an error in accepting the block? How do we roll the transactions back as a single operation? All the operations of validation before this are fine, but we need to explicitly capture errors associated with a single transact operation here, since this will eventually reflect database updates and should be treated that way. 

If there's a failure in updating anything, it should all be easy to roll back atomically. Not sure if that's within the scope of this PR.

modules/dag/src/main/scala/org/tesselation/dag/l1/DAGBlock.scala
buckysballs: Currently we're using the maxheight for doing diffs before redownload. We're using minheight for pretty much everything else. We're don't need to worry about greatest common ancestors except for optimizing tip conflict handling in the L0.

modules/dag/src/main/scala/org/tessellation/dag/l1/L1DAGBlockCell.scala
```diff
@@ -0,0 +1,516 @@
...
+            }
+          } yield result
+        }
+
+        scheme.hyloM(
+          AlgebraM[F, StackF, Either[CellError, Ω]] {
+            case More(a) => a.pure[F]
+            case Done(Right(cmd: L1DAGBlockAlgebraCommand)) =>
+              cmd match {
+                //TODO: should we check if at this moment all the proposals are present (for the case of single facilitator consensus e.g.)
```
marcinwadon, line 144: Please create a ticket

modules/dag/src/main/scala/org/tessellation/dag/l1/L1DAGBlockCell.scala
```diff
@@ -0,0 +1,516 @@
...
+                    case None               => (roundData.some, roundData.some)
+                  }
+                  def handleUpdateResult(errorMessage: String): Option[RoundData] => F[Either[CellError, Ω]] = {
+                    case Some(roundData) =>
+                      for {
+                        signedProposal <- Signed
+                          .forAsyncKryo[F, L1PeerDAGBlockData](roundData.ownProposal, context.keyPair)
+                        // TODO: when broadcast failed the own consensus round got stuck and wasn't cleaned.
+                        //  Should we automatically clean up after the failed broadcast (which may not work - I mean sending cancellation)
+                        //  or should it be handled by consensus health check mechanism(that's not here yet)?
```
marcinwadon, line 157: Please create a ticket

modules/shared/src/main/scala/org/tessellation/security/signature/Signed.scala
```diff
@@ -39,12 +44,30 @@ object Signed {
...
           .traverse(signature.verifySignatureProof(hash, _))
           .map(_.forall(identity))
       } yield isValid
+
+    def hashWithSignatureCheck[F[_]: Async: SecurityProvider: KryoSerializer]
+      : F[Either[InvalidSignatureForHash[A], Hashed[A]]] =
+      for {
+        hash <- signed.value.hashF
+        // TODO: could become ProofHash/SignatureHash instead of Hash - we won't mix them up
+        // TODO: maybe we can avoid sorting by using a NES insted of NEL
```
marcinwadon, line 63: Please create tickets for that

modules/shared/src/main/scala/org/tessellation/security/Hashed.scala
```diff
@@ -0,0 +1,11 @@
+package org.tessellation.security
+
+import org.tessellation.security.hash.Hash
+import org.tessellation.security.signature.Signed
+
+import derevo.cats.{eqv, show}
+import derevo.derive
+
+// TODO: maybe a proofsHash should have it's own type?
```
marcinwadon, line 9: If you think so, do that :)

modules/shared/src/main/scala/org/tessellation/schema/transaction.scala
```diff
@@ -59,6 +63,9 @@ object transaction {
             salt.coerce.toHexString
           )
         )
+
+    // TODO: do we want this, or would we rather use parent.ordinal.next everywhere we need this tx's ordinal
```
marcinwadon, line 67: can't it be a value instead of def?

modules/shared/src/main/scala/org/tessellation/schema/BlockValidator.scala
```diff
@@ -0,0 +1,7 @@
+package org.tessellation.schema
+
+import org.tessellation.security.signature.Signed
+
+trait BlockValidator[F[_], A <: L1Block] {
+  def accept(signedBlock: Signed[A]): F[Unit]
```
marcinwadon, line 6: validator should validate, not accept :P

modules/dag/src/main/scala/org/tessellation/dag/l1/storage/TransactionStorage.scala
```diff
@@ -0,0 +1,108 @@
...
+  def pullTransactions(): F[Option[NonEmptyList[Signed[Transaction]]]]
+}
+
+object TransactionStorage {
+  import org.tessellation.dag.l1.DAGStateChannel.{ordinalNumberOrder, ordinalNumberOrdering}
+
+  def make[F[_]: Sync]: TransactionStorage[F] =
+    make(
+      Ref.unsafe(Map.empty),
+      Ref.unsafe(Map.empty)
```
marcinwadon, line 38: we shouldn't use `Ref.unsafe` at all. when creating, you can use Ref.of

modules/dag/src/main/scala/org/tessellation/dag/l1/storage/SetRefUtils.scala
```diff
@@ -0,0 +1,18 @@
+package org.tessellation.dag.l1.storage
```
marcinwadon, line 1: can you move it to shared?

modules/dag/src/main/scala/org/tessellation/dag/l1/storage/MapRefUtils.scala
```diff
@@ -0,0 +1,32 @@
+package org.tessellation.dag.l1.storage
```
marcinwadon, line 1: can you move it to shared?

modules/dag/src/main/scala/org/tessellation/dag/l1/storage/ConsensusStorage.scala
```diff
@@ -0,0 +1,40 @@
...
+  ownConsensus: Ref[F, Option[RoundData]],
+  peerConsensuses: MapRef[F, RoundId, Option[RoundData]]
+)
+
+object ConsensusStorage {
+
+  def make[F[_]: Sync]: F[ConsensusStorage[F]] =
+    for {
+      peerConsensuses <- MapRef.ofConcurrentHashMap[F, RoundId, RoundData]()
+      ownConsensus = Ref.unsafe(none[RoundData])
```
marcinwadon, line 25: please don't use Ref.unsafe. check how other services/storages are created (in core)

modules/dag/src/main/scala/org/tessellation/dag/l1/storage/ConsensusStorage.scala
```diff
@@ -0,0 +1,40 @@
...
+  peers: Set[Peer],
+  owner: PeerId,
+  ownProposal: Proposal,
+  ownBlock: Option[Signed[DAGBlock]],
+  ownCancellation: Option[CancellationReason],
+  peerProposals: Map[PeerId, Proposal],
+  peerBlocks: Map[PeerId, Signed[DAGBlock]],
+  peerCancellations: Map[PeerId, CancellationReason],
+  tips: PulledTips
+)
```
marcinwadon, line 40: can you move it into separate file?

modules/dag/src/main/scala/org/tessellation/dag/l1/ConsensusConfig.scala
```diff
@@ -0,0 +1,4 @@
+package org.tessellation.dag.l1
+
+// TODO: consider using posInt
+case class ConsensusConfig(peersCount: Int = 2, tipsCount: Int = 2)
```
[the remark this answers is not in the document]
TheMMaciek: It's a part of `DAGStateChannelContext` which is created during startup, so this config taken from context is not intended to change during runtime. Though it could change during state channel startup.

modules/dag/src/main/scala/org/tessellation/dag/l1/DAGBlockRumorHandler.scala
```diff
@@ -0,0 +1,35 @@
...
+object DAGBlockRumorHandler {
+
+  def getHandler[F[_]: Async: SecurityProvider: KryoSerializer](blockStorage: BlockStorage[F]): RumorHandler[F] = {
+    val logger = Slf4jLogger.getLogger[F]
+
+    RumorHandler.fromReceivedRumorFn[F, Signed[DAGBlock]]() {
+      case ReceivedRumor(_, signedBlock) =>
+        for {
+          maybeHashed <- signedBlock.hashWithSignatureCheck
+          _ <- maybeHashed match {
```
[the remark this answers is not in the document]
TheMMaciek: It's `Either[InvalidSignature, Hashed[Block]]` and I do a match on it to determine which it is, but I can call it hashed here if you prefer it.

modules/core/src/main/scala/org/tesselation/domain/cluster/storage/AddressStorage.scala
TheMMaciek: The `getBalance` is used to return a value - it defaults to 0 if there is no balance in the table for that address. I added `getMaybeBalance` to determine whether I need to insert the value or update it during the batch upsert based on transactions in a block. I could implement `getBalance` in terms of `getMaybeBalance` if that helps?

modules/core/src/main/scala/org/tesselation/domain/cluster/storage/AddressStorage.scala
TheMMaciek: Refactored.

modules/core/src/main/scala/org/tesselation/domain/cluster/storage/AddressStorage.scala
TheMMaciek: Fixed. I'm now passing `(Address, Balance)` tuple.

modules/dag/src/main/scala/org/tesselation/dag/l1/DAGBlockValidator.scala
TheMMaciek: Do you mean:
```
Applicative[F].unit,
new Throwable("Invalid block signatures!").raiseError[F, Unit]
```

TheMMaciek: I guess I will also create a separate types for this errors.

modules/dag/src/main/scala/org/tesselation/dag/l1/DAGBlockValidator.scala
TheMMaciek: Ok I will check that and refactor.

modules/dag/src/main/scala/org/tessellation/dag/l1/DAGBlockValidator.scala
```diff
@@ -0,0 +1,178 @@
...
+      _ <- acceptTransactions(hashedTransactions)
+      _ <- blockStorage.acceptBlock(hashedBlock)
+      // TODO: could happen outside of acceptance flow???
+      _ <- blockStorage.handleTipsUpdate(hashedBlock)
+    } yield ()
+
+  def validateBlockSignatures(signedBlock: Signed[DAGBlock]): F[Boolean] =
+    for {
+      areSignaturesValid <- signedBlock.hasValidSignature
+      areAtLeastThreeSignatures = signedBlock.proofs.map(_.id).toList.toSet.size >= 3 //At least 3 distinct signers
```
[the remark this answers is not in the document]
TheMMaciek: Yep it was supposed to come from config but I started with "magic" numbers and forgot to reactor that part apparently.

modules/dag/src/main/scala/org/tessellation/dag/l1/DAGBlockRumorHandler.scala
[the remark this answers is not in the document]
marcinwadon: Maybe = Option. If its na Either it shouldn't be named "maybe"

modules/core/src/main/scala/org/tesselation/domain/cluster/storage/AddressStorage.scala
marcinwadon: Only one is needed. Either what's already done and second method can be "exists", or just the implementation with option

modules/dag/src/main/scala/org/tessellation/dag/l1/DAGStateChannel.scala
```diff
@@ -0,0 +1,166 @@
...
+
+}
+
+object DAGStateChannel {
+  implicit val ordinalNumberOrder: Order[Signed[Transaction]] = (x: Signed[Transaction], y: Signed[Transaction]) =>
+    implicitly[Order[BigInt]].compare(x.value.ordinal.coerce, y.value.ordinal.coerce)
+
+  implicit val ordinalNumberOrdering: Ordering[Signed[Transaction]] =
+    (x: Signed[Transaction], y: Signed[Transaction]) =>
+      implicitly[Ordering[BigInt]].compare(x.value.ordinal.coerce, y.value.ordinal.coerce)
```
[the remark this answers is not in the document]
TheMMaciek: I think that's a good idea.

modules/dag/src/main/scala/org/tesselation/dag/l1/DAGBlockValidator.scala
TheMMaciek: Right, I actually have it defined, just need to use it.

modules/dag/src/main/scala/org/tesselation/dag/l1/DAGBlockValidator.scala
TheMMaciek: Checked it and it seems the error is not being swallowed. `liftTo` for `Either` leads to this piece of code and it looks like it rethrows by using `ApplicativeError` instance
```
  def fromEither[A](x: E Either A): F[A] =
    x match {
      case Right(a) => pure(a)
      case Left(e)  => raiseError(e)
    }
```

modules/core/src/main/scala/org/tessellation/ConstellationApp.scala
```diff
@@ -24,16 +25,41 @@ import org.tessellation.schema.node.NodeState
 import org.tessellation.schema.peer.PeerId
 import org.tessellation.security.SecurityProvider
 
+import org.http4s.HttpRoutes
 import org.typelevel.log4cats.slf4j.Slf4jLogger
 
-object Main extends IOApp {
```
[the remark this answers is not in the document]
TheMMaciek: Yeah I think you're right.

modules/core/src/main/scala/org/tessellation/modules/HttpApi.scala
```diff
@@ -20,17 +20,19 @@ object HttpApi {
     queues: Queues[F],
     services: Services[F],
     programs: Programs[F],
-    environment: AppEnvironment
+    environment: AppEnvironment,
+    stateChannelp2pRoutes: HttpRoutes[F]
```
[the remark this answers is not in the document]
TheMMaciek: We will have to add `stateChannelPublicRoutes` as well actually.

modules/dag/src/main/scala/org/tessellation/dag/l1/DAGBlockValidator.scala
```diff
@@ -0,0 +1,178 @@
...
+    for {
+      _ <- validateBlockSignatures(signedBlock).ifM(
+        Async[F].unit,
+        Async[F].raiseError[Unit](new Throwable("Invalid block signatures!"))
+      )
+      _ <- validateBlock(signedBlock.value).ifM(
+        Async[F].unit,
+        Async[F].raiseError[Unit](new Throwable("Invalid block data!"))
+      )
+      hashedBlock <- signedBlock.hashWithSignatureCheck.flatMap(_.liftTo[F])
```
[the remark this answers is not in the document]
TheMMaciek: `hashWIthSignatureCheck` returns `F[Either[_, _]]`, `liftTo` lifts `Either` to `F`, so we need to lift inside flatMap

modules/dag/src/main/scala/org/tessellation/dag/l1/DAGBlockValidator.scala
```diff
@@ -0,0 +1,178 @@
...
+//        case Some(tr) =>
+//        case None =>
+//      }
+//    }.sequence[F, (String, Boolean)]
+//      .map(l => l.collect { case (h, true) => h })
+//
+//    containsAccepted
+//  }
+
+  private def areTransactionsValid(signedTransactions: Seq[Signed[Transaction]]): F[Boolean] =
```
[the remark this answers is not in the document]
TheMMaciek: We are using a `Set` of transactions in a block, but I don't think we need to restrict validators to `Set` as most of them would actually convert it to a `Seq` at some point to perform the validations. And if we would actually get a `Seq` of transactions from somewhere we can validate it and fail on duplicates whereas if someone goes from `Seq` to `Set` it will deduplicate during the conversion and it could pass validation when it shouldn't based on the original data. WDYT?

modules/dag/src/main/scala/org/tessellation/dag/l1/DAGBlockValidator.scala
```diff
@@ -0,0 +1,178 @@
...
+    for {
+      arePassingValidation <- areTransactionsPassingValidation(signedTransactions)
+      areChainedCorrectly <- areAllTransactionsFormingCorrectChain(signedTransactions)
+    } yield arePassingValidation && areChainedCorrectly
```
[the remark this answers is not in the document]
TheMMaciek: There is a separate validator for transactions. These are only used to bring together validations from `TransactionValidator` and chain validation from `BlockValidator` into one function for block validation purpose.

modules/dag/src/main/scala/org/tessellation/dag/l1/DAGBlockValidator.scala
```diff
@@ -0,0 +1,178 @@
...
+      .toList
+      .traverse {
+        case (address, txs) =>
+          for {
+            lastAccepted <- transactionStorage.getLastAcceptedTransactionRef(address)
+            sorted = txs.sortBy(_.value.parent.ordinal.coerce.value)
+            // TODO: do we also need to validate if the first lastRef isn't referencing itself?
+            //  I guess it makes no sense given that ordinal and hash of the current tx isnt specified so it's impossible to self-reference
+            chainableTransactions = takeConsecutiveTransactions(lastAccepted, txs)
+            //  was txs.toSet == chainableTransactions.toSet but that deduplicates, went with sorted lists as these should match
```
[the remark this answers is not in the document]
TheMMaciek: I purposefully removed validations that are checking the same conditions. Here if not all transactions are chainable it means simply that either there are gaps, or duplicates, or invalid transactions. I would avoid overcomplicating validations by creating redundant ones.

modules/dag/src/main/scala/org/tessellation/dag/l1/storage/SetRefUtils.scala
TheMMaciek: Yep, along with `mapRefUtils`.

modules/dag/src/main/scala/org/tesselation/dag/l1/DAGBlockValidator.scala
TheMMaciek: Yep I also thought about this problem during the implementing, so it's nice that you are also pointing it out. Didn't come up with a good solution though at that time. To be honest if the transactions acceptance passes (database is updated) and the block acceptance fails (which is a simple blocking Ref modification, so I think the chances are very low) I can't guarantee that any rollback operation on the database will succeed (or can I somehow?), so we may still end up in a broken state. I don't think we can atomically update database and blockStorage at the same time - or maybe we can but I don't know how. The only solution I can think of is balances hash as a part of snapshot and majority being picked. Now that we plan to switch to transactions for rewards this actually seems possible to achieve. Do you have some idea how that could be tackled?

modules/dag/src/main/scala/org/tessellation/dag/l1/DAGBlockValidator.scala
```diff
@@ -0,0 +1,178 @@
...
+      }
+      .map(_.forall(identity))
+
+  private def prepareBalances(transactions: Seq[Transaction]): F[Seq[UpsertAction[StoredAddress]]] =
+    transactions.traverse { tx =>
+      for {
+        sourceBalance <- addressStorage.getBalance(tx.source)
+        newSourceBalance <- refineV[NonNegative]
+          .apply(sourceBalance.coerce.value - tx.amount.coerce.value)
+          .bimap(new Throwable(_), Balance(_))
```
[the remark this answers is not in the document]
TheMMaciek: Yep I want it to failfast.

modules/dag/src/main/scala/org/tessellation/dag/l1/DAGBlockValidator.scala
```diff
@@ -0,0 +1,178 @@
...
+          .liftTo[F]
+        sourceAction = Insert(StoredAddress(tx.source, newSourceBalance))
+        maybeDestinationBalance <- addressStorage.getMaybeBalance(tx.destination)
+        destinationAction = maybeDestinationBalance.map(db => Balance(db.coerce |+| tx.amount.coerce)) match {
+          case Some(newBalance) =>
+            Update(StoredAddress(tx.destination, newBalance))
+          case None =>
+            Insert(StoredAddress(tx.destination, Balance(tx.amount.coerce)))
+        }
```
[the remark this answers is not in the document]
TheMMaciek: That would be great, but I couldn't pull that of (thought maybe it can be done) to fetch balance, adjust and generate needed action and do it transactionally for a batch of transaction. So I came up with this idea of `UpsertAction` because I could pre-generate the actions based on current balances (this also requires that only block acceptance updates the balances single block at a time). You are right that for the multiple transactions it won't work - good catch - but I think we could still leverage it just by aggregating the balances from all transactions. We can create a ticket for it and tackle that issue again, because currently I don't know what is the perfect solution.

modules/dag/src/main/scala/org/tessellation/dag/l1/DAGBlockValidator.scala
```diff
@@ -0,0 +1,178 @@
...
+      txs: Seq[Signed[Transaction]]
+    ): Seq[Signed[Transaction]] =
+      txs.find(_.value.parent == prevTxRef) match {
+        case Some(tx) =>
+          loop(tx +: acc, tx.value.parent, txs.diff(Seq(tx)))
+        case None => acc.reverse //to preserve order of the chain
+      }
+
+    loop(Seq.empty, lastAcceptedTxRef, txs)
+  }
```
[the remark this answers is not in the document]
TheMMaciek: It's pure and used in places which don't need the whole blockValidator. I can extract it to some other object if that would work better?

modules/dag/src/main/scala/org/tessellation/dag/l1/DAGStateChannel.scala
```diff
@@ -0,0 +1,166 @@
...
+
+class DAGStateChannel[F[_]: Async: KryoSerializer: SecurityProvider: Random](
+  context: DAGStateChannelContext[F]
+) {
+
+  private val logger = Slf4jLogger.getLogger[F]
+
+  private def enoughPeersForConsensus: F[Boolean] =
+    context.clusterStorage.getPeers
+      .map(_.filter(_.state == Ready))
```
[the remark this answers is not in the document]
TheMMaciek: Out of the states I see, this is the only one valid for consensus. At least at current moment. Apart from that that's state channel specific, so I don't think the core will decide wether state channel should or shouldn't perform consensus.

modules/dag/src/main/scala/org/tessellation/dag/l1/DAGStateChannel.scala
```diff
@@ -0,0 +1,166 @@
...
+      enoughPeersForConsensus
+        .flatMap(enoughPeers => context.consensusStorage.ownConsensus.get.map(rd => (rd.isEmpty, enoughPeers)))
+    )
+    .delayBy(120.seconds)
+    .metered(5.seconds)
+    .evalTap {
+      case (noRoundInProgress, areEnoughPeers) =>
+        logger.debug(s"noRoundInProgress = $noRoundInProgress areEnoughPeers=$areEnoughPeers")
+    }
+    .map { case (bool, bool1) => if (bool && bool1) OwnRoundTrigger.some else none[OwnRoundTrigger.type] }
```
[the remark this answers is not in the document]
TheMMaciek: Yeah I should :D

modules/dag/src/main/scala/org/tessellation/dag/l1/DAGStateChannel.scala
```diff
@@ -0,0 +1,166 @@
...
+    .unNone
```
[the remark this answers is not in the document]
TheMMaciek: Will check it out.

modules/dag/src/main/scala/org/tessellation/dag/l1/DAGStateChannel.scala
```diff
@@ -0,0 +1,166 @@
...
+      .map(_.size >= context.consensusConfig.peersCount)
+
+  // TODO: I guess we could have this evaluation happen inside of the cell as coalgebra step and we would only sent Tick from here?
+  private val ownerL1Input: Stream[F, L1OwnerDAGBlockData] = Stream
+    .repeatEval(
+      enoughPeersForConsensus
+        .flatMap(enoughPeers => context.consensusStorage.ownConsensus.get.map(rd => (rd.isEmpty, enoughPeers)))
+    )
+    .delayBy(120.seconds)
+    .metered(5.seconds)
```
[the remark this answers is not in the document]
TheMMaciek: Possibly, essentaily I want to delay it once and then repeat every 5 seconds.

modules/dag/src/main/scala/org/tessellation/dag/l1/DAGStateChannel.scala
```diff
@@ -0,0 +1,166 @@
...
+  private val peerL1Input: Stream[F, L1PeerDAGBlockData] = Stream.fromQueueUnterminated(context.l1PeerDAGBlockDataQueue)
+
+  private val l1Input: Stream[F, L1DAGBlockData] = ownerL1Input.merge(peerL1Input)
+
+  private val l1Pipeline: Pipe[F, L1DAGBlockData, FinalBlock] =
+    _.evalTap(l1Data => logger.debug(s"Received l1Data to process: $l1Data"))
+      .evalMap(new L1DAGBlockCell(_, context).run().handleErrorWith(e => CellError(e.getMessage).asLeft[Ω].pure[F]))
+      .flatMap {
+        case Left(ce) =>
+          Stream.eval(logger.warn(ce)(s"Error occurred during some step of L1 consensus.")) >>
```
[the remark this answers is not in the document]
TheMMaciek: Not necessarily 🤔

PR #66, opened 2021-11-11
Update data generator to new cats, add basic test.

ryle-goehausen:
Preparation for sybil attack example next, verify data generator
is still working properly and generating proper edges.

### DIFF
--- a/modules/core/src/test/scala/org/tessellation/trust/DataGenerationSuite.scala
+++ b/modules/core/src/test/scala/org/tessellation/trust/DataGenerationSuite.scala
@@ -1,0 +1,21 @@
+package org.tessellation.trust
+
+import cats.effect.IO
+import cats.effect.std.Random
+
+import weaver.SimpleIOSuite
+import weaver.scalacheck.Checkers
+
+object DataGenerationSuite extends SimpleIOSuite with Checkers {
+  test("Data generation has values populated") {
+    val numNodes = 30
+
+    val generator = Random.scalaUtilRandom[IO].map { implicit rnd =>
+      new DataGenerator[IO]
+    }
+    generator
+      .flatMap(_.generateData(numNodes = numNodes))
+      .map(_.head.edges)
+      .map(edges => expect.all(edges.exists(_.trust != 0), edges.nonEmpty, edges.length < numNodes))
+  }
+}
--- a/modules/core/src/test/scala/org/tessellation/trust/DataGenerator.scala
+++ b/modules/core/src/test/scala/org/tessellation/trust/DataGenerator.scala
@@ -1,0 +1,65 @@
+package org.tessellation.trust
+
+import cats.effect.std.Random
+import cats.syntax.all._
+import cats.{Applicative, Monad}
+
+import org.tessellation.infrastructure.trust.{TrustEdge, TrustNode}
+
+class DataGenerator[F[_]: Monad: Random] {
+
+  def randomEdgeLogic(distance: Double): F[Boolean] =
+    for {
+      value1 <- Random[F].nextDouble
+      value2 <- Random[F].nextDouble
+    } yield {
+      value1 > distance && value2 < 0.5
+    }
+
+  def randomEdge(logic: Double => F[Boolean] = randomEdgeLogic)(n: TrustNode, n2: TrustNode) =
+    for {
+      result <- logic(n.id.toDouble)
+      trustZeroToOne <- Random[F].nextDouble
+    } yield {
+      if (result) {
+        Some(TrustEdge(n.id, n2.id, 2 * (trustZeroToOne - 0.5), isLabel = true))
+      } else None
+    }
+
+  def seedCliqueLogic(maxSeedNodeIdx: Int = 1)(id: Double): F[Boolean] =
+    Applicative[F].pure(id <= maxSeedNodeIdx)
+
+  def cliqueEdge(logic: Double => F[Boolean] = seedCliqueLogic())(n: TrustNode, n2: TrustNode): F[Option[TrustEdge]] =
+    for {
+      result <- logic(n.distance(n2))
+      trustZeroToOne <- Random[F].nextDouble
+    } yield
+      if (result) Some(TrustEdge(n.id, n2.id, 1.0, isLabel = true))
+      else {
+        Some(TrustEdge(n.id, n2.id, 2 * (trustZeroToOne - 0.5)))
+      }
+
+  def generateData(
+    numNodes: Int = 30,
+    edgeLogic: (TrustNode, TrustNode) => F[Option[TrustEdge]] = randomEdge()
+  ): F[List[TrustNode]] =
+    for {
+      x <- Random[F].nextDouble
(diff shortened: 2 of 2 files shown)


modules/core/src/test/scala/org/tesselation/trust/DataGenerationSuite.scala
```diff
@@ -0,0 +1,20 @@
...
+import cats.effect.{Async, IO}
+
+import weaver.SimpleIOSuite
+import weaver.scalacheck.Checkers
+
+object DataGenerationSuite extends SimpleIOSuite with Checkers {
+  test("Data generation has values populated") {
+    val numNodes = 30
+    for {
+      _ <- Async[IO].pure(IO.pure(()))
```
marcinwadon, line 12: Why is that needed? If you have only one operation, you don't need to write a for comp and do such tricks at the beginning. Please consider following:

```scala
val numNodes = 30
val gen = new DataGenerator()

gen.generateData(numNodes)
  .map(_.head.edges)
  .flatMap { edges =>
    expect.all(edges.exists(_.trust != 0), edges.nonEmpty, edges.length < numNodes)
  }
```

modules/core/src/test/scala/org/tesselation/trust/DataGenerator.scala
```diff
@@ -0,0 +1,68 @@
...
+
+import cats.effect.{Async, Ref}
+import cats.syntax.all._
+
+import scala.util.Random
+
+import org.tesselation.infrastructure.trust.{TrustEdge, TrustNode}
+
+class DataGenerator[F[_]: Async] {
+  private final val randomEffect: Ref[F, Random] = Ref.unsafe(new Random())
```
marcinwadon, line 11: Why is `Ref` needed here? I don't see you're updating it (so just val/def with pure/delayed random would be enough). Anyway, there is a `SecureRandom` instance from cats if you want an effect.

There is also a second option, to use a State monad for Random, but I think that's not the case here.

modules/core/src/test/scala/org/tesselation/trust/DataGenerator.scala
```diff
@@ -0,0 +1,68 @@
+package org.tesselation.trust
+
+import cats.effect.{Async, Ref}
+import cats.syntax.all._
+
+import scala.util.Random
+
+import org.tesselation.infrastructure.trust.{TrustEdge, TrustNode}
+
+class DataGenerator[F[_]: Async] {
```
marcinwadon, line 10: I know it's just a test but looks like you don't need an Async here. Just `Ref.Make: Monad` would be enough :)

marcinwadon (changes requested):

kpudlik (changes requested):

modules/core/src/test/scala/org/tesselation/trust/DataGenerator.scala
```diff
@@ -0,0 +1,68 @@
...
+      }
+      res <- nodesWithEdges
+    } yield res
+
+  def bipartiteEdge(logic: Double => F[Boolean] = seedCliqueLogic())(n: TrustNode, n2: TrustNode) =
+    for {
+      result <- logic(n.id.toDouble)
+    } yield
+      if (result) Some(TrustEdge(n.id, n2.id, 1.0, isLabel = true))
+      else Some(TrustEdge(n.id, n2.id, -1.0))
```
kpudlik, line 67: for-comprehension is not needed if you are not `flatMapping`

```scala
logic(n.id.toDouble).map { result =>
  if (result) Some(TrustEdge(n.id, n2.id, 1.0, isLabel = true),
  else Some(TrustEdge(n.id, n2.id, -1.0)
}
```

kpudlik (changes requested):

modules/core/src/test/scala/org/tesselation/trust/DataGenerationSuite.scala
ryle-goehausen: So just to clarify -- the Async came from the Concurrent from previous code base. All I did was move this over and get it working with new cats version. It required an implicit here. I actually don't think it matters if any of this code has async or anything else, since it's purely for generating test data, but it's what was there before and I need to base changes off it.

The code you posted in snippet above won't work -- because it's missing the implicit required by Ref for the random -- I tried moving the code over as is and it's missing an implicit for Sync provided by Async. (this is due to changing cats version.)

modules/core/src/test/scala/org/tesselation/trust/DataGenerator.scala
ryle-goehausen: I didn't write this Ref originally -- this was here before. I honestly don't think this code needs to be async either, so if you're okay with removing this that's fine with me. It's only supposed to be used in the tests for generating data before passing it somewhere else, so it really doesn't need to be captured.

modules/core/src/test/scala/org/tesselation/trust/DataGenerator.scala
ryle-goehausen: Cats new version requires sync implicit on the ref signature. If I copy paste the code as is it doesn't work without it.

modules/core/src/test/scala/org/tesselation/trust/DataGenerator.scala
```diff
@@ -0,0 +1,68 @@
...
+        val trustZeroToOne = random.nextDouble()
+        Some(TrustEdge(n.id, n2.id, 2 * (trustZeroToOne - 0.5), isLabel = true))
+      } else None
+
+  def seedCliqueLogic(maxSeedNodeIdx: Int = 1)(id: Double): F[Boolean] =
+    for {
+      _ <- randomEffect.get
+    } yield
+      if (id <= maxSeedNodeIdx) true
+      else false
```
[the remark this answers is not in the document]
ryle-goehausen: I'll update this since it's simple but again none of this is new changes.

modules/core/src/test/scala/org/tesselation/trust/DataGenerator.scala
ryle-goehausen: Same, will update this but didn't modify any of this code.

ryle-goehausen:
@marcinwadon updated this to use the cats random as discussed. It still seems to require an async -- so if there's a better way to deal with the pure initializer in the test let me know.

modules/core/src/test/scala/org/tesselation/trust/DataGenerationSuite.scala
marcinwadon: image

image

marcinwadon (changes requested):
see screenshots

modules/core/src/test/scala/org/tesselation/trust/DataGenerationSuite.scala
ryle-goehausen: @marcinwadon fixed

marcinwadon (approved):

PR #77, opened 2021-12-02
Use decline for core CLI layer

marcinwadon:


modules/core/src/main/scala/org/tessellation/Main.scala
```diff
@@ -24,78 +22,90 @@ import org.tessellation.schema.node.NodeState
...
 
-  private def loadKeyPair[F[_]: Async: SecurityProvider](cfg: KeyConfig): F[KeyPair] =
+  private def loadKeyPair[F[_]: Async: SecurityProvider](cfg: Run): F[KeyPair] =
     KeyStoreUtils
       .readKeyPairFromStore[F](
-        cfg.keystore,
-        cfg.keyalias.value,
-        cfg.storepass.value.toCharArray,
-        cfg.keypass.value.toCharArray
+        cfg.keyStore.value.toString,
```
tbekas, line 106: Is the `.toString` necessary?

modules/core/src/main/scala/org/tessellation/Main.scala
```diff
@@ -24,78 +22,90 @@ import org.tessellation.schema.node.NodeState
...
+import com.monovore.decline.Opts
+import com.monovore.decline.effect.CommandIOApp
 import org.typelevel.log4cats.slf4j.Slf4jLogger
 
-object Main extends IOApp {
+object Main
+    extends CommandIOApp(
+      name = "",
+      header = "Constellation Node",
+      version = "0.0.x"
```
tbekas, line 33: It would be a cumbersome to keep version both in `build.sbt` as well as in `Main.scala`. I think we could use [sbt-buildinfo](https://github.com/sbt/sbt-buildinfo) plugin to keep them in sync. Could you create a separate task for using this plugin?

tbekas (approved):

modules/core/src/main/scala/org/tessellation/Main.scala
marcinwadon: Yes, it's a Path from fs2 (https://javadoc.io/doc/co.fs2/fs2-io_2.13/latest/fs2/io/file/Path.html).

modules/core/src/main/scala/org/tessellation/Main.scala
marcinwadon: I already did that yesterday and posted on slack @tbekas :)

tbekas: Awesome :)

modules/core/src/main/scala/org/tessellation/Main.scala
```diff
@@ -24,78 +22,90 @@ import org.tessellation.schema.node.NodeState
...
     KeyStoreUtils
       .readKeyPairFromStore[F](
-        cfg.keystore,
-        cfg.keyalias.value,
-        cfg.storepass.value.toCharArray,
-        cfg.keypass.value.toCharArray
+        cfg.keyStore.value.toString,
+        cfg.alias.value.value,
+        cfg.password.value.value.toCharArray,
+        cfg.password.value.value.toCharArray
```
TheMMaciek, line 109: Shouldn't we use `.coerce` for these values?

PR #78, opened 2021-12-06
Snapshot aggregation

tbekas:


modules/shared/src/main/scala/org/tessellation/schema/aci/StateChannelOutput.scala
```diff
@@ -0,0 +1,12 @@
...
+
+import org.tessellation.schema.address.Address
+
+sealed trait StateChannelOutput {
+  val address: Address
+}
+
+object StateChannelOutput {
+  case class Snapshot(address: Address, bytes: Array[Byte]) extends StateChannelOutput
+  case class Other(address: Address) extends StateChannelOutput
```
TheMMaciek, line 11: Does it make sens to even have this class if it doesn't bring almost any information with it about the actual output? 🤔

modules/core/src/main/scala/org/tessellation/infrastructure/aci/StateChannelRouter.scala
```diff
@@ -37,16 +42,25 @@ object StateChannelRouter {
...
+        cellOutputOrError <- cell.run()
+        cellOutput <- cellOutputOrError.liftTo[F]
+        interpretedOutput <- context.interpretOutput(cellOutput)
+        _ <- publishOutput(interpretedOutput)
+      } yield cellOutput
+
+    private def publishOutput(output: StateChannelOutput) =
+      output match {
+        case o @ Snapshot(_, _) => stateChannelOutputQueue.offer(o)
+        case _                  => Applicative[F].unit
```
TheMMaciek, line 61: I guess it suggests that it can be any other output, whereas for now `StateChannelOutput` is a `sealed trait` and we know that it will be the second case `Other`. Do we intend to unseal that trait?

TheMMaciek: Maybe we should have the special case called `StateChannelSnapshotOutput` as I guess it reflects better the restriction that we make here?

modules/kernel/src/main/scala/org/tessellation/kernel/StateChannelContext.scala
```diff
@@ -11,6 +12,8 @@ trait StateChannelContext[F[_]] {
     hypergraphContext: HypergraphContext[F]
   ): F[Cell[F, StackF, Ω, Either[CellError, Ω], Ω]]
 
+  def interpretOutput(output: Ω): F[StateChannelOutput]
```
TheMMaciek, line 15: That's very restricting and I guess only valid in case of L0 Cells. Was that the intention? I mean if we use `interpretOutput` on L1 it needs to have a different definition as not every output will converge into a `Snapshot`.

modules/shared/src/main/scala/org/tessellation/schema/aci/StateChannelOutput.scala
```diff
@@ -0,0 +1,12 @@
+package org.tessellation.schema.aci
+
+import org.tessellation.schema.address.Address
+
+sealed trait StateChannelOutput {
+  val address: Address
+}
+
+object StateChannelOutput {
+  case class Snapshot(address: Address, bytes: Array[Byte]) extends StateChannelOutput
```
TheMMaciek, line 10: I guess we will be trying to deserialize the bytes based on the Address, so we don't need custom types introduced by the state channel owner?

modules/shared/src/main/scala/org/tessellation/schema/aci/StateChannelOutput.scala
```diff
@@ -0,0 +1,12 @@
+package org.tessellation.schema.aci
+
+import org.tessellation.schema.address.Address
+
+sealed trait StateChannelOutput {
```
TheMMaciek, line 5: Maybe we should have this special case called `StateChannelSnapshotOutput` as I guess it reflects better the restriction that we make here?

marcinwadon (approved):

modules/shared/src/main/scala/org/tessellation/schema/aci/StateChannelOutput.scala
tbekas: You're right. I think I will use `Option[Output]`

modules/shared/src/main/scala/org/tessellation/schema/aci/StateChannelOutput.scala
tbekas: That is the case.

modules/core/src/main/scala/org/tessellation/infrastructure/aci/StateChannelContextLoader.scala
```diff
@@ -97,6 +100,12 @@ class StateChannelContextLoader[F[_]: Async] {
                 cellCtor.newInstance(input, hypergraphContext).asInstanceOf[StdCell[F]]
               }
 
+            override def interpretOutput(output: Ω): F[StateChannelOutput] =
+              output.getClass.some match {
+                case `snapshotClass` => output.toBinaryF.map(StateChannelOutput.Snapshot(address, _))
+                case _               => Applicative[F].pure(StateChannelOutput.Other(address))
+              }
+
```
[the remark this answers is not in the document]
tbekas: I agree. I will do it this way.

modules/core/src/main/scala/org/tessellation/infrastructure/aci/StateChannelRouter.scala
tbekas: I refactored it and there's no longer `Other` type.

modules/kernel/src/main/scala/org/tessellation/kernel/StateChannelContext.scala
tbekas: True. I removed it and I just pass output as it is from Cell.

modules/core/src/main/scala/org/tessellation/domain/aci/StateChannelInstance.scala
```diff
@@ -0,0 +1,15 @@
+package org.tessellation.domain.aci
+
+import org.tessellation.kernel._
+import org.tessellation.kryo.KryoSerializer
+import org.tessellation.schema.address.Address
+
+trait StateChannelInstance[F[_]] {
```
marcinwadon, line 7: maybe ```StateChannelDef``` or ```StateChannelDefinition``` ?

modules/core/src/main/scala/org/tessellation/infrastructure/aci/StateChannelRouter.scala
```diff
@@ -2,51 +2,71 @@ package org.tessellation.infrastructure.aci
...
 
 object StateChannelRouter {
 
-  def make[F[_]: Async]: F[StateChannelRouter[F]] =
-    new StateChannelContextLoader[F].loadFromClasspath.map(make(_))
+  def make[F[_]: Async](stateChannelOutputQueue: Queue[F, StateChannelOutput]): F[StateChannelRouter[F]] =
+    for {
+      instances <- new StateChannelInstanceLoader[F].loadFromClasspath
+      contexts <- instances.toList.traverse {
+        case (_address, _) =>
```
marcinwadon, line 27: maybe `stateChannelAddress` ?

examples/empty/src/main/resources/state-channel-impl
```diff
@@ -0,0 +1 @@
+org.example.EmptyDef
```
marcinwadon, line 1: I'd name the file `-def` to be consistent

build.sbt
```diff
@@ -1,7 +1,7 @@
 import Dependencies.{Libraries, _}
 
 ThisBuild / scalaVersion := "2.13.5"
-ThisBuild / version := "0.0.2"
+ThisBuild / version := "0.0.4"
```
marcinwadon, line 4: agreed on slack to try snapshots

marcinwadon (changes requested):

build.sbt
```diff
@@ -1,7 +1,7 @@
 import Dependencies.{Libraries, _}
 
 ThisBuild / scalaVersion := "2.13.5"
-ThisBuild / version := "0.0.2"
+ThisBuild / version := "0.0.3-SNAPSHOT"
```
marcinwadon, line 4: should be ```0.0.4-SNAPSHOT``` since `0.0.3` is already released

modules/core/src/main/scala/org/tessellation/infrastructure/aci/StateChannelRouter.scala
marcinwadon: 🙃

marcinwadon (changes requested):

marcinwadon (approved):
scalafix

examples/empty/src/main/scala/org/example/EmptyDef.scala
```diff
@@ -0,0 +1,29 @@
...
+
+  val address: Address = Address("DAG3k3VihUWMjse9LE93jRqZLEuwGd6a5Ypk4zYS")
+  val kryoRegistrar: Map[Class[_], KryoRegistrationId] = Map(classOf[EmptyInput] -> 1001)
+
+  def makeCell[F[_]: Monad](input: EmptyInput, hgContext: HypergraphContext[F]) =
+    new Cell[F, StackF, EmptyInput, Either[CellError, Ω], EmptyInput](
+      data = input,
+      hylo = scheme.hyloM(
+        AlgebraM[F, StackF, Either[CellError, Ω]] { _ =>
+          Monad[F].pure(NullTerminal.asInstanceOf[Ω].asRight[CellError])
```
TheMMaciek, line 21: I guess you could use widen on an either instead of asInstanceOf - not sure but I guess it gives more compile time safety.

TheMMaciek (approved):

marcinwadon (approved):

TheMMaciek (approved):

PR #82, opened 2021-12-10
Revert "Directed Acylic Transitive Trust implementation"

ryle-goehausen:
Reverts Constellation-Labs/tessellation#80

This was merged to master by mistake. My bad, we should probably set the default PR branch to 'develop' so less likely in the future. I'll just add this to my next PR I'm working on now to develop.


marcinwadon:
Reverted manually from master

PR #84, opened 2021-12-10
DAG state channel basic types and Hashed class for verified Signed

TheMMaciek:


marcinwadon (dismissed):

build.sbt
```diff
@@ -277,4 +277,18 @@ lazy val core = (project in file("modules/core"))
     )
   )
 
+lazy val dag = (project in file("modules/dag"))
+  .dependsOn(core)
```
marcinwadon, line 281: If dag needs to depend on core it should be separated (see examples, not as a module). IMO the dependency should be in the opposite direction (core depends on dag)

marcinwadon (changes requested):

build.sbt
tbekas: I think that @TheMMaciek is using `core` as an `sdk`. So the dependency is in the right direction (`dag` on `sdk`).

On the second issue of keeping it separate (like `examples`). I agree in principle, however `dag` is a special case of a state-channel, so maybe it is okay if we treat it specially and keep it as a module? What do you guys think?

modules/shared/src/main/scala/org/tessellation/schema/transaction.scala
```diff
@@ -23,7 +27,9 @@ object transaction {
 
   @derive(decoder, encoder, eqv, show)
   @newtype
-  case class TransactionOrdinal(value: NonNegBigInt)
+  case class TransactionOrdinal(value: NonNegBigInt) {
```
tbekas, line 30: I think it's safe to use `NonNegLong`. I don't think it's possible for ordinal to exceed `Long` max value in any foreseeable future.

modules/shared/src/main/scala/org/tessellation/schema/transaction.scala
```diff
@@ -59,8 +65,17 @@ object transaction {
...
+
+    val ordinal: TransactionOrdinal = parent.ordinal.next
   }
 
+  implicit val signedTransactionOrder: Order[Signed[Transaction]] = (x: Signed[Transaction], y: Signed[Transaction]) =>
+    implicitly[Order[BigInt]].compare(x.value.ordinal.coerce, y.value.ordinal.coerce)
+
+  implicit val signedTransactionOrdering: Ordering[Signed[Transaction]] =
+    (x: Signed[Transaction], y: Signed[Transaction]) =>
+      implicitly[Ordering[BigInt]].compare(x.value.ordinal.coerce, y.value.ordinal.coerce)
```
tbekas, line 77: You can use `Order[Signed[Transaction]].toOrdering` instead.

modules/shared/src/main/scala/org/tessellation/schema/transaction.scala
TheMMaciek: Yep it's most probably safe, though I don't know if there is a big value in changing to `NonNegLong`. Anyway transaction schema is outside of the scope of this PR, so I can create a ticket to think about and change if needed.

build.sbt
TheMMaciek: Maybe I could create something like `dag-core` module (probably dependent on schema only) and keep base types and validators there (stuff that's needed for snapshot creation in L0 also). And I would keep state channel specific implementation in `dag` module dependent on `core` or separate project? Regarding where to keep DAG state channel implementation dependent on `core` I think keeping it in a module would be nice, given that it is ours, it is closely related and kind of a backbone of the whole system.  And yes I am using the core as an sdk for the needed cluster functionality - I guess most of the state channels will probably do that. BTW that's why I think we shouldn't necessarily keep the stuff related to L0 in `core`. I guess `core` should be a base for both L1's and L0. WDYT?

marcinwadon: For me, `core` is the app that will be running on every L0 node. That's where all the base things should be, as well as DAG (maybe as a dependency). The other way would be to keep `core` as an sdk, and create module `l0` and separate project `l1`.

modules/shared/src/main/scala/org/tessellation/schema/transaction.scala
tbekas: I would use the smallest types that will do the job.

build.sbt
```diff
@@ -277,4 +277,21 @@ lazy val core = (project in file("modules/core"))
     )
   )
 
+lazy val dag = (project in file("modules/dag-shared"))
```
marcinwadon, line 280: please change the val name also

modules/dag-shared/src/main/scala/org/tessellation/dag/types/DAGBlock.scala
```diff
@@ -0,0 +1,16 @@
...
+import org.tessellation.schema.transaction.Transaction
+import org.tessellation.security.signature.Signed
+
+import derevo.circe.magnolia.{decoder, encoder}
+import derevo.derive
+import io.estatico.newtype.ops._
+
+@derive(encoder, decoder)
+case class DAGBlock(transactions: Set[Signed[Transaction]], parent: NonEmptyList[BlockReference]) {
+  val height: Height = Height(parent.sortBy(_.height.value).head.height.coerce + 1L)
```
kpudlik, line 15: `parent.maxBy(_.height.value)`

TheMMaciek: If there is a `minBy` I will use it!

tbekas (dismissed):
LGTM

modules/dag-shared/src/main/scala/org/tessellation/dag/types/DAGBlock.scala
kpudlik: There is :)

Also `coerce` is safer because you are not coupled with the name of constant encapsulated by newtype

marcinwadon (approved):

PR #85, opened 2021-12-13
SDK and DAG-L1 modules

marcinwadon:


build.sbt
```diff
@@ -217,6 +216,38 @@ lazy val testShared = (project in file("modules/test-shared"))
...
+    commonTestSettings,
+    libraryDependencies ++= Seq(
+      CompilerPlugin.kindProjector,
+      CompilerPlugin.betterMonadicFor,
+      CompilerPlugin.semanticDB
+    )
+  )
+
+lazy val dagL1 = (project in file("modules/dag-l1"))
+  .dependsOn(shared)
```
TheMMaciek, line 236: this will depend on sdk right?

TheMMaciek (changes requested):
Apart from review comments, there are some empty files pushed. We either need to remove them or add entry to .gitignore

build.sbt
marcinwadon: correct. done

PR #86, opened 2021-12-13
DAG state channel transaction validator

TheMMaciek:


modules/dag-shared/src/main/scala/org/tessellation/dag/validators/TransactionValidator.scala
```diff
@@ -0,0 +1,208 @@
+package org.tessellation.dag.validators
```
marcinwadon, line 1: Shouldn't we agree to group by type not by feature?
IMO transaction validator should be in ```org.tessellation.dag.transaction```

modules/dag-shared/src/main/scala/org/tessellation/dag/validators/TransactionValidator.scala
```diff
@@ -0,0 +1,208 @@
...
+      balanceValidation <- validateSourceBalances(Seq(signedTransaction.value))
+    } yield propertiesValidation.product(balanceValidation).map(_ => signedTransaction)
+
+  def validateBatchOfTransactions(
+    signedTransactions: Seq[Signed[Transaction]]
+  ): F[ValidationResult[Seq[Signed[Transaction]]]] =
+    for {
+      propertiesValidation <- signedTransactions.traverse(validateTransactionProperties).map(_.sequence)
+      balancesValidation <- validateSourceBalances(signedTransactions.map(_.value))
+    } yield propertiesValidation.product(balancesValidation).map(_ => signedTransactions)
```
marcinwadon, line 52: for public API maybe following would be better?

```scala
def validate(transaction: Signed[Transaction]) = validate(Seq(transaction))
def validate(transactions: Seq[Signed[Transaction]]) = implementation
```

Then, you just use it like:
```
TransactionValidator.validate(tx)
TransactionValidator.validate(Seq(tx1, tx2))
```

Imo there is no point to use single/batch and "transaction" in the method name in this case

modules/dag-shared/src/main/scala/org/tessellation/dag/validators/TransactionValidator.scala
```diff
@@ -0,0 +1,208 @@
...
+  private def validateSourceAddress(
+    signedTransaction: Signed[Transaction]
+  ): F[ValidationResult[Transaction]] =
+    signedTransaction.proofs.traverse {
+      case SignatureProof(id, _) => {
+        for {
+          publicKey <- id.hex.toPublicKey
+          address = publicKey.toAddress
+        } yield address == signedTransaction.value.source
+      }.handleError(_ => false)
```
marcinwadon, line 97: This will swallow the error. Is that supposed?

modules/dag-shared/src/main/scala/org/tessellation/dag/validators/TransactionValidator.scala
```diff
@@ -0,0 +1,208 @@
...
+    val errorMessage: String = s"Transactions source address=$source has insufficient balance"
+  }
+
+  case object InvalidSourceSignature extends TransactionValidationError {
+    val errorMessage: String = s"Transaction has invalid source signature"
+  }
+
+  object SourceAddressAndSignerIdsDontMatch {
+    def apply(tx: Transaction) = new SourceAddressAndSignerIdsDontMatch(tx.source)
+  }
```
marcinwadon, line 185: I think you should order it like:
```
class AA
object AA

class BB
object BB
```
instead of 
```
class AA
class BB
object AA
object BB
```

marcinwadon (changes requested):

modules/dag-shared/src/main/scala/org/tessellation/dag/validators/TransactionValidator.scala
TheMMaciek: Ok!

modules/dag-shared/src/main/scala/org/tessellation/dag/validators/TransactionValidator.scala
TheMMaciek: Sure, makes sense.

modules/dag-shared/src/main/scala/org/tessellation/dag/validators/TransactionValidator.scala
TheMMaciek: Yep, the error returned in the following lines is descriptive enough.

TheMMaciek: I mean we could log it, but then we would need a logger in the validator, maybe the failing validation case is good enough?

marcinwadon: but here you have an error that is different than inequality (something throws under the hood) and this error is gone (i.e. kryo issue)

modules/dag-shared/src/main/scala/org/tessellation/dag/validators/TransactionValidator.scala
TheMMaciek: I will try, not sure if that wasn't re-ordered like that byt the linter (thought that could have been some internal intellij linter, so I will check that again).

modules/dag-shared/src/main/scala/org/tessellation/dag/validators/TransactionValidator.scala
TheMMaciek: Yeah I know. I'm just wondering though if we should log everything everywhere 🤔.

modules/dag-shared/src/main/scala/org/tessellation/dag/transaction/TransactionValidator.scala
```diff
@@ -0,0 +1,207 @@
...
+
+  implicit val F: Async[F]
+  implicit val kryoSerializer: KryoSerializer[F]
+  implicit val securityProvider: SecurityProvider[F]
+
+  protected def getBalance(address: Address): F[Balance]
+
+  protected def getLastAcceptedTransactionRef(address: Address): F[TransactionReference]
+
+  def validate(
```
marcinwadon, line 38: you should be able to do either:
```
def validate(tx) = implementation
def validate(txs) => txs.traverse(validate)
```
or
```
def validate(tx) = validate(Seq(tx))
def validate(txs) = implementation
```

I'd opt for the second. Definitely you don't need the implementation twice

modules/dag-shared/src/main/scala/org/tessellation/dag/transaction/TransactionValidator.scala
```diff
@@ -0,0 +1,207 @@
...
+        for {
+          publicKey <- id.hex.toPublicKey
+          address = publicKey.toAddress
+        } yield address == signedTransaction.value.source
+      }.handleError(_ => false)
+    }.map(_.forall(identity))
+      .ifM(
+        signedTransaction.value.validNel.pure[F],
+        Async[F].pure(SourceAddressAndSignerIdsDontMatch(signedTransaction.value).invalidNel)
+      )
```
kpudlik, line 102: Eventually:

```scala
signedTransaction.proofs.existsM {
  case SignatureProof(...) => {
     for { ... } yield address != signedTransaction.value.source 
  }.handleError(_ => true)
}.ifM(...)
```

By negating the statement it won't iterate over all the proofs but will stop at the first one which is wrong. Also you get rid of `forall`

modules/dag-shared/src/main/scala/org/tessellation/dag/transaction/TransactionValidator.scala
TheMMaciek: The first one won't work as we need to validate balances of all the transactions together. The second one should work.

marcinwadon (approved):

PR #88, opened 2021-12-14
DAG state channel block validator

TheMMaciek:


modules/dag-shared/src/main/scala/org/tessellation/dag/block/BlockValidator.scala
```diff
@@ -0,0 +1,153 @@
...
+import org.tessellation.dag.transaction.TransactionValidator.TransactionValidationError
+import org.tessellation.dag.transaction.filter.Consecutive
+import org.tessellation.dag.types.{BlockReference, DAGBlock}
+import org.tessellation.kryo.KryoSerializer
+import org.tessellation.schema.address.Address
+import org.tessellation.schema.transaction.{Transaction, TransactionReference}
+import org.tessellation.security.SecurityProvider
+import org.tessellation.security.signature.Signed
+
+abstract class BlockValidator[F[_]](transactionValidator: TransactionValidator[F], consensusPeersCount: Int) {
```
marcinwadon, line 28: why it's an abstract class instead of trait?

marcinwadon, line 28: maybe `facilitators` instead of `consensus peers`?

modules/dag-shared/src/main/scala/org/tessellation/dag/block/BlockValidator.scala
```diff
@@ -0,0 +1,153 @@
...
+
+  implicit val F: Async[F]
+  implicit val kryoSerializer: KryoSerializer[F]
+  implicit val securityProvider: SecurityProvider[F]
```
marcinwadon, line 32: if you have an abstract class, you can put it into the class definition

modules/dag-shared/src/main/scala/org/tessellation/dag/block/BlockValidator.scala
```diff
@@ -0,0 +1,153 @@
...
+    signedBlock.hasValidSignature
+      .ifM(
+        signedBlock.validNel.pure[F],
+        Applicative[F].pure(InvalidBlockSignatures.invalidNel)
+      )
+
+  private def validateSignaturesCount(signedBlock: Signed[DAGBlock]): ValidationResult[Signed[DAGBlock]] = {
+    val uniqueSignersCount = signedBlock.proofs.map(_.id).toList.toSet.size
+
+    // TODO: is more signatures than expected ok?
```
marcinwadon, line 61: more than expected is fine, although such a case should not happen for now.

modules/dag-shared/src/main/scala/org/tessellation/dag/block/BlockValidator.scala
```diff
@@ -0,0 +1,153 @@
...
+    override def getMessage: String = errorMessage
+  }
+
+  case object InvalidBlockSignatures extends BlockValidationError {
+    override val errorMessage: String = "Block has invalid signatures!"
+  }
+
+  case class InvalidBlockSignaturesCount(is: Int, shouldBe: Int) extends BlockValidationError {
+    override val errorMessage: String = s"Block has invalid signatures number: $is not equal $shouldBe!"
+  }
```
marcinwadon, line 136: maybe ```$is less or equal than $shouldBe```?

modules/dag-shared/src/main/scala/org/tessellation/dag/transaction/filter/Consecutive.scala
```diff
@@ -0,0 +1,28 @@
+package org.tessellation.dag.transaction.filter
+
+import scala.annotation.tailrec
+
+import org.tessellation.schema.transaction.{Transaction, TransactionReference}
+import org.tessellation.security.signature.Signed
+
+object Consecutive {
```
marcinwadon, line 8: can you create a jira ticket to write a unit test for that?

marcinwadon (changes requested):

modules/dag-shared/src/main/scala/org/tessellation/dag/block/BlockValidator.scala
TheMMaciek: We also have the notion of facilitators (participating) and owner. Then it could be misleading. But I can actually make it more precise and go with `requiredUniqueBlockSigners` or something similar.

modules/dag-shared/src/main/scala/org/tessellation/dag/block/BlockValidator.scala
TheMMaciek: Right. After decision about abstract class vs trait I will refactor that.

modules/dag-shared/src/main/scala/org/tessellation/dag/block/BlockValidator.scala
TheMMaciek: Yeah I wonder if we should just go with equal then? Especially that it's configurable here.

modules/dag-shared/src/main/scala/org/tessellation/dag/transaction/filter/Consecutive.scala
TheMMaciek: Yep!

modules/dag-shared/src/main/scala/org/tessellation/dag/block/BlockValidator.scala
TheMMaciek: Right it makes sense if we leave the greater or equal condition in the validator.

modules/dag-shared/src/main/scala/org/tessellation/dag/block/BlockValidator.scala
marcinwadon: okay, I'm fine with that ;)

marcinwadon: but I think it's strictly related to the number of facilitators during the consensus round 🤔 I mean, that should be the same value. anyway, `requiredUniqueBlockSigners` would be fine :+1: name is descriptive enough and one can pass the same configuration value here and to the consensus :+1:

modules/dag-shared/src/main/scala/org/tessellation/dag/block/BlockValidator.scala
marcinwadon: I don't think so. IMO greater or equal is 100% correct

marcinwadon (approved):

modules/dag-shared/src/main/scala/org/tessellation/dag/block/BlockValidator.scala
```diff
@@ -0,0 +1,151 @@
...
+import org.tessellation.dag.types.{BlockReference, DAGBlock}
+import org.tessellation.kryo.KryoSerializer
+import org.tessellation.schema.address.Address
+import org.tessellation.schema.transaction.{Transaction, TransactionReference}
+import org.tessellation.security.SecurityProvider
+import org.tessellation.security.signature.Signed
+
+abstract class BlockValidator[F[_]: Async: KryoSerializer: SecurityProvider](
+  transactionValidator: TransactionValidator[F],
+  requiredUniqueBlockSigners: Int
```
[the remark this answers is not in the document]
TheMMaciek: Good idea, I will introduce it with the next PR probably.

PR #89, opened 2021-12-14
DAG state channel refined types and access modifiers in block validator

TheMMaciek:


modules/dag-shared/src/main/scala/org/tessellation/dag/block/BlockValidator.scala
```diff
@@ -25,14 +25,17 @@ import org.tessellation.schema.transaction.{Transaction, TransactionReference}
 import org.tessellation.security.SecurityProvider
 import org.tessellation.security.signature.Signed
 
+import eu.timepit.refined.auto._
```
marcinwadon, line 28: is it required here?

TheMMaciek: We use it throughout the repo so I imported it instead of extracting int with `.value`

marcinwadon: that's what I asked about :) if it's really used or just imported because we import it everywhere.

marcinwadon (dismissed):

modules/dag-shared/src/main/scala/org/tessellation/dag/block/BlockValidator.scala
TheMMaciek: Yep, used!

kpudlik: Also you can put `PosInt` in place where `Int` is required

marcinwadon (approved):

PR #91, opened 2021-12-15
Base types and services for DAG Snapshot

kpudlik:


modules/core/src/main/scala/org/tessellation/l0/SnapshotService.scala
```diff
@@ -0,0 +1,59 @@
...
+  ): SnapshotService[F] =
+    new SnapshotService[F] {
+
+      def createSnapshot(minTipHeight: PosLong): F[DAGSnapshot] =
+        for {
+          previousSnapshotHeight <- snapshotStorage.getLastSnapshotHeight
+          nextSnapshotHeight = Height(previousSnapshotHeight.coerce + config.heightInterval)
+
+          // In 1.0: validateSnapshotHeightIntervalCondition
+          //  if (minTipHeight > nextHeight + config.heightDelayInterval)
```
kpudlik, line 36: I am wondering if we should check it here

modules/core/src/main/scala/org/tessellation/l0/SnapshotService.scala
```diff
@@ -0,0 +1,59 @@
...
+          blocksForNextSnapshot = acceptedBlocks.filter { block =>
+            val blockHeight = block.signed.height.coerce
+            blockHeight > previousSnapshotHeight.coerce && blockHeight <= nextSnapshotHeight.coerce
+          }
+
+          // In 1.0: validateAcceptedCBsSinceSnapshot
+          //  if (blocksForNextSnapshot.nonEmpty) OK
+
+          // In 1.0: validateBlocksWithinHeightInterval
+          //  blocksForNextSnapshot.toSeq.map(_.signed).traverse(blockValidator.validate)
```
kpudlik, line 49: Can we assume that accepted blocks are already validated? Why we needed additional validation in 1.0?

modules/core/src/main/scala/org/tessellation/l0/SnapshotStorage.scala
```diff
@@ -0,0 +1,25 @@
...
+  def getLastSnapshotHeight: F[Height]
+}
+
+object SnapshotStorage {
+
+  def make[F[_]: Async]: SnapshotStorage[F] =
+    new SnapshotStorage[F] {
+      private val lastSnapshot: Ref[F, Option[DAGSnapshot]] = Ref.unsafe(None)
+
+      override def storeSnapshot(snapshot: DAGSnapshot): F[Unit] = lastSnapshot.set(Some(snapshot))
```
kpudlik, line 19: Should storage on disk be separated from storage in memory?

modules/core/src/main/scala/org/tessellation/l0/DAGSnapshot.scala
```diff
@@ -0,0 +1,7 @@
+package org.tessellation.l0
+
+import org.tessellation.dag.types.DAGBlock
+import org.tessellation.schema.height.Height
+import org.tessellation.security.Hashed
+
+case class DAGSnapshot(height: Height, blocks: Set[Hashed[DAGBlock]])
```
kpudlik, line 7: In 1.0 height was dynamically calculated based on `blocks`. We should probably do the same here. WDYT?

modules/core/src/main/scala/org/tessellation/l0/SnapshotService.scala
```diff
@@ -0,0 +1,59 @@
...
+
+          acceptedBlocks <- blockStorage.getAccepted
+
+          blocksForNextSnapshot = acceptedBlocks.filter { block =>
+            val blockHeight = block.signed.height.coerce
+            blockHeight > previousSnapshotHeight.coerce && blockHeight <= nextSnapshotHeight.coerce
+          }
+
+          // In 1.0: validateAcceptedCBsSinceSnapshot
+          //  if (blocksForNextSnapshot.nonEmpty) OK
```
kpudlik, line 46: We shouldn't create an empty snapshot but it shouldn't be checked here as well. Actually, I have a feeling that all the conditions should be checked outside of `createSnapshot` function. WDYT?

modules/shared/src/main/scala/org/tessellation/security/signature/Signed.scala
```diff
@@ -29,6 +29,8 @@ case class Signed[A](value: A, proofs: NonEmptyList[SignatureProof])
 object Signed {
   case class InvalidSignatureForHash[A <: AnyRef](signed: Signed[A]) extends NoStackTrace
 
+  implicit def autoUnwrap[T](t: Signed[T]): T = t.value
```
marcinwadon, line 32: @TheMMaciek you can use it in Transaction/Block validator then :)

modules/core/src/main/scala/org/tessellation/l0/SnapshotStorage.scala
marcinwadon: Can we store it only on disk? 🤔 
Maybe we can use a database here, instead of file storage

modules/core/src/main/scala/org/tessellation/l0/SnapshotStorage.scala
```diff
@@ -0,0 +1,25 @@
...
+trait SnapshotStorage[F[_]] {
+  def storeSnapshot(snapshot: DAGSnapshot): F[Unit]
+  def getLastSnapshotHeight: F[Height]
+}
+
+object SnapshotStorage {
+
+  def make[F[_]: Async]: SnapshotStorage[F] =
+    new SnapshotStorage[F] {
+      private val lastSnapshot: Ref[F, Option[DAGSnapshot]] = Ref.unsafe(None)
```
marcinwadon, line 17: Please don't use `Ref.unsafe`

modules/core/src/main/scala/org/tessellation/l0/SnapshotStorage.scala
```diff
@@ -0,0 +1,25 @@
...
+
+import cats.effect.kernel.{Async, Ref}
+import cats.syntax.functor._
+
+import org.tessellation.schema.height.Height
+
+trait SnapshotStorage[F[_]] {
+  def storeSnapshot(snapshot: DAGSnapshot): F[Unit]
+  def getLastSnapshotHeight: F[Height]
+}
```
marcinwadon, line 11: shouldn't it be in the domain?

modules/core/src/main/scala/org/tessellation/l0/DAGSnapshot.scala
marcinwadon: I think it can be dynamic as well. Also, facilitators are missing (next facilitators)

modules/core/src/main/scala/org/tessellation/cli/snapshot.scala
```diff
@@ -0,0 +1,29 @@
...
+
+  val heightIntervalOpts: Opts[PosLong] = Opts
+    .env[PosLong]("CL_SNAPSHOT_HEIGHT_INTERVAL", help = "Height interval between snapshots")
+    .withDefault(PosLong(2))
+
+  val heightDelayIntervalOpts: Opts[PosLong] = Opts
+    .env[PosLong]("CL_SNAPSHOT_HEIGHT_DELAY_INTERVAL", help = "Height delay of snapshot trigger")
+    .withDefault(PosLong(2))
+
+  val storedSnapshotPath: Opts[Path] = Opts
```
TheMMaciek, line 23: I wonder if we want these to be loaded from envs or just hardcoded in an object?

modules/core/src/main/scala/org/tessellation/cli/snapshot.scala
```diff
@@ -0,0 +1,29 @@
...
+
+object snapshot {
+
+  val heightIntervalOpts: Opts[PosLong] = Opts
+    .env[PosLong]("CL_SNAPSHOT_HEIGHT_INTERVAL", help = "Height interval between snapshots")
+    .withDefault(PosLong(2))
+
+  val heightDelayIntervalOpts: Opts[PosLong] = Opts
+    .env[PosLong]("CL_SNAPSHOT_HEIGHT_DELAY_INTERVAL", help = "Height delay of snapshot trigger")
+    .withDefault(PosLong(2))
```
TheMMaciek, line 21: I wonder if we want these to be loaded from envs or just hardcoded in an object? Stored snapshot path, is per node operator so it can be loaded, but height intervals are global. WDYT?

TheMMaciek: In the future maybe we will take configuration from the majority snapshot, like we discussed some time ago.

modules/core/src/main/scala/org/tessellation/l0/DAGSnapshot.scala
TheMMaciek: Yeah we need to figure out if there will be a rotation happening on L0 for DAG. Not sure if we need to add it right now or wait. What about lastSnapshot?

TheMMaciek: We don't need to add reputation yet I think.

TheMMaciek: Also I intended `Hashed` to be used internally, not sent over network as every node should regenerate hashes on it's own. So just like `DAGBlock` stores  `Set[Signed[Transaction]]` maybe the same case should be for the blocks? So not `Set[Hashed[DAGBlock]]` but `Set[Signed[DAGBlock]]`

modules/core/src/main/scala/org/tessellation/l0/SnapshotStorage.scala
kpudlik: Do you mean in memory DB or persisted?

modules/core/src/main/scala/org/tessellation/l0/SnapshotService.scala
TheMMaciek: Especially that we are not sure how the trigger will work. Old implementation was based on tips used up by the node itself. Not sure how the trigger will work in L0 (which not necessarily will create it's own blocks).

modules/core/src/main/scala/org/tessellation/l0/SnapshotStorage.scala
kpudlik: Sure. I will make it `F[SnapshotStorage[F]]`

modules/core/src/main/scala/org/tessellation/l0/SnapshotService.scala
TheMMaciek: I think it makes sense. At least for now, until we figure out the triggering mechanism.

TheMMaciek: The problem here is that you can't pull from the storage and expect that pulled blocks will not be and empty `List`/`Set`. It would need to be more like a pure function getting `NEL` or `NES` of blocks as a parameter.

modules/core/src/main/scala/org/tessellation/l0/SnapshotService.scala
TheMMaciek: I would go with single point of validation and strict control over storages at least for now. If possible all blocks should go through the same path to the block storage to have a control over the logic - hence be sure that accepted is known to be valid.

modules/core/src/main/scala/org/tessellation/l0/DAGSnapshot.scala
TheMMaciek: Other thing. Maybe we should keep the same principal as we do for transactions and blocks, so we have a complete reference to the previous snapshot? Given that it's like transactions `Height` and `Hash`, wdyt?

modules/core/src/main/scala/org/tessellation/l0/SnapshotService.scala
marcinwadon: keep in mind that some types of blocks (state channels) will be only validated during snapshot creation process

modules/core/src/main/scala/org/tessellation/l0/SnapshotStorage.scala
marcinwadon: persisted 🤔

modules/core/src/main/scala/org/tessellation/l0/SnapshotService.scala
buckysballs: We should be able to reuse essentially the same logic from the Constellation repo i.e. using tip height. However in order to promote separation of the L0 from L1 or rather the snapshot trigger from block creation, what if we updated tip height as blocks are received rather than from the checkpoint storage cache?

modules/core/src/main/scala/org/tessellation/l0/SnapshotService.scala
buckysballs: While it may seem redundant, we will want every cell to validate input before broadcasting the output. It's possible that invalid/forged blocks can be received from L1 processes. Filtering these out before proposing in a snapshot helps prevent redownloads and will help with convergence and cluster stability in general

modules/core/src/main/scala/org/tessellation/l0/DAGSnapshot.scala
buckysballs: We will need the same rotation in facilitators for L0 as for L1, at least in the final implementation.

kpudlik: @TheMMaciek but `BlockStorage` doesn't have `Signed[DAGBlock]`. You would need to store there `Hashed[Signed[DAGBlock]]` or sth 🤔

TheMMaciek: `Hashed` has the `Signed` inside of it, you can just extract it.

modules/core/src/main/scala/org/tessellation/l0/SnapshotService.scala
TheMMaciek: That's exactly what single point of validation would do I think. The block will not be saved as valid until it is actually valid. Here we are implementing the logic for our DAGBlocks, not some abstract general block, so we can handle them the way we want. For the blocks from other state channels the logic will be for sure different and they can be stored in a different place and they will also have a single point of validation (while being pulled to the snapshot). That's at least how I see it.

marcinwadon:
rebase needed

modules/core/src/main/scala/org/tessellation/l0/SnapshotService.scala
kpudlik: > While it may seem redundant, we will want every cell to validate input before broadcasting the output. It's possible that invalid/forged blocks can be received from L1 processes. Filtering these out before proposing in a snapshot helps prevent redownloads and will help with convergence and cluster stability in general

I agree with Maciek. Such issue occurs in Global Snapshot and we should filter these blocks out but in my case it is just a DAGSnapshot so if we implement single point of validation in block acceptance (DAG blocks) then we do not need to validate blocks again,

modules/core/src/main/scala/org/tessellation/l0/SnapshotStorage.scala
kpudlik: Ok DB works for me. Much better than storing a file. Also, I am thinking if we still need some `Ref` or in-memory information about the last snapshot if we decide to store *whole* snapshot in DB. Probably not.

modules/core/src/main/scala/org/tessellation/l0/SnapshotService.scala
kpudlik: But we can't guarantee that it will be NEL/NES. Eventually `Option[NonEmptyList]` 🤔

marcinwadon (approved):

modules/core/src/main/scala/org/tessellation/l0/domain/snapshot/GlobalSnapshot.scala
```diff
@@ -0,0 +1,19 @@
...
+import org.tessellation.dag.domain.block.DAGBlock
+import org.tessellation.kernel.StateChannelSnapshot
+import org.tessellation.schema.height.Height
+import org.tessellation.schema.peer.PeerId
+import org.tessellation.security.signature.Signed
+
+import io.estatico.newtype.ops._
+
+case class GlobalSnapshot(
+  blocks: Set[Signed[DAGBlock]],
```
TheMMaciek, line 14: I wonder if we will be actually creating snapshots with no DAGBlocks 🤔

kpudlik: I guess we will. That's why it is not `NonEmptySet`. I guess the types will change when we implement the trigger so I wouldn't worry about it now

modules/core/src/main/scala/org/tessellation/l0/domain/snapshot/GlobalSnapshot.scala
```diff
@@ -0,0 +1,19 @@
...
+  snapshots: Set[StateChannelSnapshot],
+  nextFacilitators: NonEmptySet[PeerId]
+) {
+  def height: Height = blocks.map(_.value.height).maxBy(_.coerce)
```
TheMMaciek, line 18: Isn't it that it may throw given that, `blocks` is a `Set` not a `NonEmptySet`?

modules/core/src/main/scala/org/tessellation/l0/infrastructure/snapshot/SnapshotService.scala
```diff
@@ -0,0 +1,52 @@
...
+      def createSnapshot(
+        blocks: NonEmptySet[Hashed[DAGBlock]],
+        snapshots: Set[StateChannelSnapshot],
+        nextFacilitators: NonEmptySet[PeerId]
+      ): F[GlobalSnapshot] =
+        for {
+          previousSnapshotHeight <- snapshotStorage.getLastSnapshotHeight
+          nextSnapshotHeight = Height(previousSnapshotHeight.coerce + heightInterval)
+
+          blocksForNextSnapshot = blocks.filter { block =>
```
TheMMaciek, line 38: If we are filtering the blocks here we need to also reconcile which blocks made it into the the GlobalSnapshot. Unless we will only be clearing the cache of blocks after picking majority and till this happens we will be passing all available blocks in the cache to this function.

modules/shared/src/main/scala/org/tessellation/ext/derevo/ordering.scala
```diff
@@ -0,0 +1,3 @@
+package org.tessellation.ext.derevo
+
+object ordering extends Derive[Ordering] {}
```
TheMMaciek, line 3: Looks like a leftover?

PR #92, opened 2021-12-15
Util functions for ref based collections

TheMMaciek:


build.sbt
```diff
@@ -175,6 +175,7 @@ lazy val shared = (project in file("modules/shared"))
       Libraries.guava,
       Libraries.log4cats,
       Libraries.logback % Runtime,
+      Libraries.mapref,
```
marcinwadon, line 178: so I think you can remove it from core

modules/shared/src/main/scala/org/tessellation/ext/collection/SetRefUtils.scala
```diff
@@ -0,0 +1,18 @@
+package org.tessellation.ext.collection
+
+import cats.effect.{Ref, Sync}
+import cats.syntax.functor._
+
+object SetRefUtils {
+  implicit class RefOps[F[_]: Sync, K, V](val ref: Ref[F, Set[K]]) {
```
marcinwadon, line 7: shouldn't it be Async?

marcinwadon (changes requested):

modules/shared/src/main/scala/org/tessellation/ext/collection/SetRefUtils.scala
```diff
@@ -0,0 +1,18 @@
...
+
+    def add(k: K): F[Unit] =
+      ref.modify(a => (a + k, ()))
```
tbekas, line 10: If you return `Unit` you can just use `ref.upddate(_ + k)`

tbekas (approved):

build.sbt
tbekas: But core also uses this lib.

modules/shared/src/main/scala/org/tessellation/ext/collection/SetRefUtils.scala
TheMMaciek: I can change it. It's a copy paste from what we had in constellation.

modules/shared/src/main/scala/org/tessellation/ext/collection/SetRefUtils.scala
TheMMaciek: If the functions in the `RefOps` require `Sync`, why require `Async` from the outside? I guess it will not be utilized anyway.

build.sbt
marcinwadon: @tbekas shouldn't it use it through this helpers?

modules/shared/src/main/scala/org/tessellation/ext/collection/SetRefUtils.scala
marcinwadon: to be consistent within the repository. at the very beginning we agreed that Async would be the base type for our effects (instead of Sync that we used with cats effect 2)

build.sbt
tbekas: I don't think so. We import `import io.chrisdavenport.mapref.MapRef` in core in couple of places, therefore I think that this dependency should stay.

modules/shared/src/main/scala/org/tessellation/ext/collection/SetRefUtils.scala
```diff
@@ -0,0 +1,18 @@
...
+import cats.syntax.functor._
+
+object SetRefUtils {
+  implicit class RefOps[F[_]: Sync, K, V](val ref: Ref[F, Set[K]]) {
+
+    def add(k: K): F[Unit] =
+      ref.update(a => a + k)
+
+    def remove(k: K): F[Unit] =
+      ref.update(a => a - k)
```
marcinwadon, line 13: (_ + k) (_ - k) ? :)

marcinwadon (approved):

modules/shared/src/main/scala/org/tessellation/ext/collection/SetRefUtils.scala
TheMMaciek: Right even better.

tbekas (approved):

marcinwadon (approved):

PR #93, opened 2021-12-15
DAG state channel transaction service

TheMMaciek:


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/storage/TransactionStorage.scala
```diff
@@ -0,0 +1,111 @@
...
+  def isParentAccepted(tx: Transaction): F[Boolean]
+
+  def getLastAcceptedTransactionRef(address: Address): F[TransactionReference]
+
+  def acceptTransaction(hashedTx: Hashed[Transaction]): F[Unit]
+
+  def putTransactions(transactions: Set[Signed[Transaction]]): F[Unit]
+
+  def pullTransactions(): F[Option[NonEmptyList[Signed[Transaction]]]]
+}
```
marcinwadon, line 31: imo it should be moved to the domain.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/storage/TransactionStorage.scala
```diff
@@ -0,0 +1,111 @@
...
+import org.tessellation.ext.collection.MapRefUtils._
+import org.tessellation.schema.address.Address
+import org.tessellation.schema.transaction.{Transaction, TransactionReference}
+import org.tessellation.security.Hashed
+import org.tessellation.security.signature.Signed
+
+import io.chrisdavenport.mapref.MapRef
+
+trait TransactionStorage[F[_]] {
+  def isParentAccepted(tx: Transaction): F[Boolean]
```
marcinwadon, line 22: is it required to pass a transaction here? maybe just a hash would be enough

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/storage/TransactionStorage.scala
```diff
@@ -0,0 +1,111 @@
...
+  ): TransactionStorage[F] =
+    new TransactionStorage[F] {
+
+      def isParentAccepted(transaction: Transaction): F[Boolean] =
+        (transaction.parent == TransactionReference.empty)
+          .pure[F]
+          .ifM(
+            true.pure[F],
+            lastAccepted(transaction.source).get.map(_.contains(transaction.parent))
+          )
```
marcinwadon, line 54:
```scala
(transaction.parent == TransactionReference.empty)
  .guard[Option]
  .fold(true.pure[F]) { _ =>
     lastAccepted(transaction.source).get.map(_.contains(transaction.parent)))
   }
```

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/storage/TransactionStorage.scala
```diff
@@ -0,0 +1,111 @@
...
+
+      // Either transactions should be accepted in the correct order (handled outside of this scope),
+      // or maybe we should check if we can update here? (current < new) then transactions accepted in wrong order
+      // will still end up in consistent state - though it feels kind of like a side effect that should be out of scope of storage
+      def acceptTransaction(hashedTx: Hashed[Transaction]): F[Unit] = {
+        val source = hashedTx.signed.value.source
+        val reference = TransactionReference(hashedTx.hash, hashedTx.signed.value.ordinal)
+
+        lastAccepted(source).set(reference.some)
+      }
```
marcinwadon, line 67: acceptance should be done in the TransactionService. it's not a role of TransactionStorage.
in this class (TransactionStorage) there should be only a move from pending/unknown to accepted state.
yet, there can be a check that reference can be set if and only if it's incremented by one.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/storage/TransactionStorage.scala
```diff
@@ -0,0 +1,111 @@
...
+
+      def putTransactions(transactions: Set[Signed[Transaction]]): F[Unit] =
```
marcinwadon, line 69: If we're in transaction storage, imo we can name it simpler:
```
TransactionStorage {
 isParentAccepted
 getLastAcceptedReference
 accept
 put
 pull
} 
```
instead of duplicating `putTransaction`, `pullTransaction`, etc. I wouldn't expect to put a block or a snapshot in transaction storage ;)

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/storage/TransactionStorage.scala
```diff
@@ -0,0 +1,111 @@
...
+                  .getOrElse(NonEmptySet.one(tx))
+
+                waiting.updated(tx.value.source, updatedAddressTxs)
+            }
+
+          (updatedWaiting, ())
+        }
+
+      // I would like to pull inside of a modify so I'm prefetching the whole lastAccepted (which could theoretically
+      // change during the pull which increases invalid block creation probability - though I'm not sure that's something worry about it)
```
marcinwadon, line 86: we had this issue before. pull should be an atomic change and we shouldn't split it into two parts - prefetch and then modify. if you need to do so, I'd use a semaphore to lock the access

marcinwadon (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/storage/TransactionStorage.scala
TheMMaciek: I think it's better to pass transaction, extract parent here and return info if the parent of this transaction is accepted. I think it's better to extract here in one place then to extract outside of it. If we would pass hash only, we would need to change the method to `isTransactionAccepted` and it's a different functionality.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/storage/TransactionStorage.scala
TheMMaciek: Needed to negate the first condition. `true` was passing as `Some`

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/storage/TransactionStorage.scala
TheMMaciek: Actually after thinking about it a bit more this scenario seems ok. I'm prefetching the `lastAcceptedTransaction` and then atomically pulling from `waitingTransactions` based on `lastAcceptedTransaction`. Even if in the meantime new transaction is being accepted which unblocks transactions from waiting during pull is happening - it's ok, simply the transactions will be pulled on the next go. We may have to use semaphores when we will have redownload where the state could actually be rolled back. Until we have that we don't need the semaphore I think.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionService.scala
```diff
@@ -0,0 +1,16 @@
...
+import cats.data.NonEmptyList
+
+import org.tessellation.schema.address.Address
+import org.tessellation.schema.transaction.{Transaction, TransactionReference}
+import org.tessellation.security.Hashed
+import org.tessellation.security.signature.Signed
+
+trait TransactionService[F[_]] {
+  def isParentAccepted(tx: Transaction): F[Boolean]
+  def getLastAcceptedTransactionRef(address: Address): F[TransactionReference]
```
marcinwadon, line 12: shouldn't we agree to name it `getLastAcceptedReference`? :thinking:

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/transaction/TransactionService.scala
```diff
@@ -0,0 +1,113 @@
...
+          .modify[UpdateResult] { lastAccepted =>
+            if (lastAccepted.isEmpty && reference.ordinal == TransactionOrdinal.first ||
+                lastAccepted.exists(_.ordinal.next == reference.ordinal))
+              (reference.some, Success)
+            else
+              (
+                lastAccepted,
+                UnexpectedStateShift(
+                  s"Transaction acceptance failed! Transaction not accepted in the correct order. source=${source.show} current=${lastAccepted.show} attempted=${reference.show}"
+                )
```
marcinwadon, line 66: IMO it should be a separate error. It's not a state shift but a domain error (missing parents, parents not accepted, or incorrect reference, etc)

marcinwadon (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/transaction/TransactionService.scala
```diff
@@ -0,0 +1,116 @@
...
+import org.tessellation.security.signature.Signed
+
+import io.chrisdavenport.mapref.MapRef
+
+object TransactionService {
+
+  def make[F[_]: Sync]: F[TransactionService[F]] =
+    for {
+      lastAccepted <- MapRef.ofConcurrentHashMap[F, Address, TransactionReference]()
+      waitingTransactions <- Ref.of[F, Map[Address, NonEmptySet[Signed[Transaction]]]](Map.empty)
```
marcinwadon, line 31: isn't it a MapRef ?

TheMMaciek: It isn't. I'm atomically pulling transactions from this cache and given that after a bit of thought decided that it's better to use `Ref[Map[]]` then `MapRef` here.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/transaction/TransactionService.scala
```diff
@@ -0,0 +1,118 @@
...
+                val pulledForAddress = Consecutive.take(lastTx, waiting.toNonEmptyList.toList)
+                val updatedStillWaiting =
+                  NonEmptySet.fromSet(SortedSet.from(pulledForAddress)) match {
+                    case Some(pulledTxs) => NonEmptySet.fromSet(waiting.diff(pulledTxs))
+                    case None            => waiting.some
+                  }
+                (updatedStillWaiting, pulledForAddress)
+              case None =>
+                (None, Seq.empty)
+            }
```
TheMMaciek, line 103: I guess it would be good to handle the error here on the modify, log the error and default to `Seq.empty`. Working with `MapRef` makes pull for single address atomic and not for all the addresses. I guess logging and defaulting to `Seq.empty` is the best way to not go down the rabbit hole of rollbacking the pull for some addresses.

TheMMaciek: Implemented the error handling as described above.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/transaction/TransactionService.scala
```diff
@@ -0,0 +1,125 @@
...
+                val updatedStillWaiting =
+                  NonEmptySet.fromSet(SortedSet.from(pulledForAddress)) match {
+                    case Some(pulledTxs) => NonEmptySet.fromSet(waiting.diff(pulledTxs))
+                    case None            => waiting.some
+                  }
+                (updatedStillWaiting, pulledForAddress)
+              case None =>
+                (None, Seq.empty)
+            }.handleErrorWith {
+              logger
```
kpudlik, line 107: `logger.warn("Error ...")` 🤔 

I am not sure if we should call warnings Error but I do not have any better alternative

kpudlik (approved):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/transaction/TransactionService.scala
marcinwadon: :+1: if it's an error, it should be `logger.error`

marcinwadon (approved):
Please see the comment about the error. Otherwise seems legit :)

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/transaction/TransactionService.scala
TheMMaciek: I think the log level `error` is not appropriate for this case so I went with `warn`.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionService.scala
```diff
@@ -0,0 +1,16 @@
...
+import org.tessellation.schema.transaction.{Transaction, TransactionReference}
+import org.tessellation.security.Hashed
+import org.tessellation.security.signature.Signed
+
+trait TransactionService[F[_]] {
+  def isParentAccepted(tx: Transaction): F[Boolean]
+  def getLastAcceptedReference(address: Address): F[TransactionReference]
+  def accept(hashedTx: Hashed[Transaction]): F[Unit]
+  def put(transactions: Set[Signed[Transaction]]): F[Unit]
+  def pull(): F[Option[NonEmptyList[Signed[Transaction]]]]
```
tbekas, line 15: Wouldn't it be better to just have just `F[List[Signed[Transaction]]]`?

tbekas (approved):
Please remove the empty file `modules/dag-l1/src/main/scala/org/tessellation/.gitkeep`

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionService.scala
TheMMaciek: Don't know. I kind of like the explicit indication that we either get `None`, or `Some` transactions. The client side can default to an empty list if there is a need to have an empty transaction list.

TheMMaciek: Regarding the `.gitkeep` file I think I did remove it 🤔 (at least GitHub shows I did)

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/transaction/TransactionService.scala
marcinwadon: So please change the error message @TheMMaciek

PR #95, opened 2021-12-16
DAG state channel block service

TheMMaciek:
Review the block service part only

### DIFF
--- a/modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
+++ b/modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
@@ -1,0 +1,221 @@
+package org.tessellation.dag.l1.domain.block
+
+import cats.Applicative
+import cats.data.NonEmptyList
+import cats.effect.Sync
+import cats.syntax.either._
+import cats.syntax.flatMap._
+import cats.syntax.functor._
+import cats.syntax.option._
+import cats.syntax.traverse._
+
+import scala.util.control.NoStackTrace
+
+import org.tessellation.dag.domain.block.{BlockReference, DAGBlock}
+import org.tessellation.dag.l1.config.TipsConfig
+import org.tessellation.dag.l1.domain.block
+import org.tessellation.dag.l1.domain.block.BlockService._
+import org.tessellation.ext.collection.MapRefUtils.MapRefOps
+import org.tessellation.security.Hashed
+import org.tessellation.security.hash.ProofsHash
+import org.tessellation.security.signature.Signed
+
+import eu.timepit.refined.auto._
+import eu.timepit.refined.types.numeric.PosInt
+import io.chrisdavenport.mapref.MapRef
+import io.estatico.newtype.ops._
+import monocle.macros.syntax.lens._
+
+class BlockService[F[_]: Sync](blocks: MapRef[F, ProofsHash, Option[StoredBlock]], tipsConfig: TipsConfig) {
+
+  def areParentsAccepted(block: DAGBlock): F[Map[BlockReference, Boolean]] =
+    block.parent.traverse { ref =>
+      isBlockAccepted(ref).map(ref -> _)
+    }.map(_.toNem.toSortedMap)
+
+  def accept(hashedBlock: Hashed[DAGBlock]): F[Unit] =
+    blocks(hashedBlock.proofsHash)
+      .modify[Either[BlockAcceptanceError, Unit]] {
+        case Some(storedBlock) =>
+          storedBlock match {
+            case unknown: UnknownBlock => (unknown.some, BlockIsUnknown(hashedBlock.proofsHash).asLeft)
+            case WaitingBlock(_, usages) =>
+              (AcceptedBlock(hashedBlock, Accepted, usages, isTip = false).some, ().asRight)
+            case accepted: AcceptedBlock => (accepted.some, BlockAlreadyAccepted(hashedBlock.proofsHash).asLeft)
+          }
+        case None =>
+          (None, BlockDoesNotExist(hashedBlock.proofsHash).asLeft)
+      }
+      .flatMap(_.liftTo[F])
+
+  def store(hashedBlock: Hashed[DAGBlock]): F[Unit] =
+    blocks(hashedBlock.proofsHash)
+      .modify[Either[BlockStoringError, Unit]] {
+        case Some(storedBlock) =>
+          storedBlock match {
+            case UnknownBlock(usages) => (WaitingBlock(hashedBlock.signed, usages).some, ().asRight)
+            case waiting: WaitingBlock =>
+              (waiting.some, BlockAlreadyStoredAndWaiting(hashedBlock.proofsHash).asLeft)
+            case accepted: AcceptedBlock =>
+              (accepted.some, BlockAlreadyStoredAndAccepted(hashedBlock.proofsHash).asLeft)
+          }
+        case None => (WaitingBlock(hashedBlock.signed, Set.empty[ProofsHash]).some, ().asRight)
+      }
+      .flatMap(_.liftTo[F]) >> updateParentUsages(hashedBlock)
+
+  def getWaiting: F[Map[ProofsHash, Signed[DAGBlock]]] =
+    blocks.toMap.map(_.collect { case (hash, WaitingBlock(block, _)) => hash -> block })
+
+  def handleTipsUpdate(hashedBlock: Hashed[DAGBlock]): F[Unit] =
+    for {
+      allBlocks <- blocks.toMap
+      allTips = allBlocks.collect { case (hash, ab @ AcceptedBlock(_, _, _, true)) => hash -> ab }
+      parentHashToTip = hashedBlock.signed.value.parent.map(ref => ref.hash -> allTips.get(ref.hash)).toList.toMap
+      tipParents = parentHashToTip.collect { case (hash, Some(accepted)) => hash -> accepted }
+      notTipParents = (parentHashToTip -- tipParents.keySet).keySet
+      tipParentsSize = tipParents.size
+      tipParentsToRemove = tipParents.filter {
+        case (_, block) =>
+          val enoughTips = allTips.size >= tipsConfig.minimumTipsCount + tipParentsSize
+          val tooManyTips = allTips.size >= tipsConfig.maximumTipsCount + tipParentsSize
+          val parentReachedUsageLimit = (block.usages + hashedBlock.proofsHash).size >= tipsConfig.maximumTipUsages
+          enoughTips && (tooManyTips || parentReachedUsageLimit)
+      }
+      tipParentsToRemain = tipParents -- tipParentsToRemove.keySet
+      remainingTips = allTips -- tipParentsToRemove.keySet
+      minTipHeight = remainingTips.values.map(_.block.signed.value.height.coerce).minOption.getOrElse(0L)
+      addAsTipCondition = remainingTips.size < tipsConfig.maximumTipsCount && minTipHeight < hashedBlock.signed.value.height.coerce
+      _ <- updateNoTipParents(notTipParents.toList, hashedBlock)
+      _ <- updateTipParentsToRemove(tipParentsToRemove.keySet.toList, hashedBlock)
+      _ <- updateTipParentsToRemain(tipParentsToRemain.keySet.toList, hashedBlock)
+      _ <- tryAddAsATip(addAsTipCondition, hashedBlock)
+    } yield ()
+
+  def pullTips(tipsCount: PosInt): F[Option[Tips]] =
+    for {
+      allBlocks <- blocks.toMap
+      allTips = allBlocks.collect { case (_, AcceptedBlock(block, _, _, true)) => block }.toList
+        .sortBy(_.signed.value.height.coerce)
+      pulledTips = allTips match {
+        case tips if tips.size >= tipsCount =>
+          NonEmptyList
+            .fromList(tips.take(tipsCount))
+            .map(parents => block.Tips(parents.map(p => BlockReference(p.proofsHash, p.signed.value.height))))
+        case _ => None
+      }
+    } yield pulledTips
+
+  private def isBlockAccepted(blockReference: BlockReference): F[Boolean] =
+    blocks(blockReference.hash).get.map(_.exists(_.isInstanceOf[AcceptedBlock]))
+
+  private def updateNoTipParents(parents: List[ProofsHash], hashedBlock: Hashed[DAGBlock]): F[Unit] =
+    parents.traverse { hash =>
+      blocks(hash)
+        .modify[Either[TipsUpdateError, Unit]] {
+          case Some(accepted @ AcceptedBlock(_, _, _, false)) =>
+            (accepted.addUsage(hashedBlock.proofsHash).some, ().asRight)
+          case unexpected =>
+            (unexpected, NoTipAcceptedBlockNotFound(hash, unexpected).asLeft)
+        }
+        .flatMap(_.liftTo[F])
+    }.void
+
+  private def updateTipParentsToRemove(parents: List[ProofsHash], hashedBlock: Hashed[DAGBlock]): F[Unit] =
+    parents.traverse { hash =>
+      blocks(hash)
+        .modify[Either[TipsUpdateError, Unit]] {
+          case Some(accepted @ AcceptedBlock(_, _, _, true)) =>
+            (accepted.addUsage(hashedBlock.proofsHash).unsetAsTip.some, ().asRight)
+          case unexpected =>
+            (unexpected, TipBlockToRemoveAsTipNotFound(hash, unexpected).asLeft)
+        }
+        .flatMap(_.liftTo[F])
+    }.void
+
+  private def updateTipParentsToRemain(parents: List[ProofsHash], hashedBlock: Hashed[DAGBlock]): F[Unit] =
+    parents.traverse { hash =>
+      blocks(hash)
+        .modify[Either[TipsUpdateError, Unit]] {
+          case Some(accepted @ AcceptedBlock(_, _, _, true)) =>
+            (accepted.addUsage(hashedBlock.proofsHash).some, ().asRight)
+          case unexpected =>
+            (unexpected, TipBlockToRemainAsTipNotFound(hash, unexpected).asLeft)
+        }
+        .flatMap(_.liftTo[F])
+    }.void
+
+  private def tryAddAsATip(condition: Boolean, hashedBlock: Hashed[DAGBlock]): F[Unit] =
+    if (condition)
+      blocks(hashedBlock.proofsHash).modify {
+        case Some(accepted @ AcceptedBlock(_, _, usages, false)) if usages.size < tipsConfig.maximumTipUsages =>
+          (accepted.setAsTip.some, ())
+        case other => (other, ())
+      } else Applicative[F].unit
+
+  private def updateParentUsages(hashedBlock: Hashed[DAGBlock]): F[Unit] = {
+    val hash = hashedBlock.proofsHash
+    hashedBlock.signed.value.parent.toList.traverse { blockReference =>
+      blocks(blockReference.hash).update {
+        case Some(storedBlock) =>
+          storedBlock match {
+            case unknown: UnknownBlock   => unknown.addUsage(hash).some
+            case waiting: WaitingBlock   => waiting.addUsage(hash).some
+            case accepted: AcceptedBlock => accepted.addUsage(hash).some
+          }
+        case None => UnknownBlock(Set(hash)).some
+      }
+    }
+  }.void
+}
+
+object BlockService {
+
+  def make[F[_]: Sync](tipsConfig: TipsConfig): F[BlockService[F]] =
+    MapRef.ofConcurrentHashMap[F, ProofsHash, StoredBlock]().map(new BlockService[F](_, tipsConfig))
+
+  sealed trait AcceptedBlockState
+  case object Accepted extends AcceptedBlockState
+  case object InMajority extends AcceptedBlockState
+  sealed trait StoredBlock
+  case class UnknownBlock(usages: Set[ProofsHash]) extends StoredBlock {
+    def addUsage(hash: ProofsHash): UnknownBlock = this.focus(_.usages).modify(_ + hash)
+  }
+  case class WaitingBlock(block: Signed[DAGBlock], usages: Set[ProofsHash]) extends StoredBlock {
+    def addUsage(hash: ProofsHash): WaitingBlock = this.focus(_.usages).modify(_ + hash)
+  }
+  case class AcceptedBlock(block: Hashed[DAGBlock], state: AcceptedBlockState, usages: Set[ProofsHash], isTip: Boolean)
+      extends StoredBlock {
+    def addUsage(hash: ProofsHash): AcceptedBlock = this.focus(_.usages).modify(_ + hash)
+    def setAsTip: AcceptedBlock = this.focus(_.isTip).replace(true)
+    def unsetAsTip: AcceptedBlock = this.focus(_.isTip).replace(false)
+  }
+
+  sealed trait BlockServiceError extends NoStackTrace
+  sealed trait TipsUpdateError extends BlockServiceError
+  case class TipBlockToRemoveAsTipNotFound(hash: ProofsHash, got: Option[StoredBlock]) extends TipsUpdateError {
+    override def getMessage: String = s"Tip block to remove as tip with hash=$hash not found! But got: $got"
+  }
+  case class TipBlockToRemainAsTipNotFound(hash: ProofsHash, got: Option[StoredBlock]) extends TipsUpdateError {
+    override def getMessage: String = s"Tip block to remain as tip with hash=$hash not found! But got: $got"
+  }
+  case class NoTipAcceptedBlockNotFound(hash: ProofsHash, got: Option[StoredBlock]) extends TipsUpdateError {
+    override def getMessage: String = s"Accepted block that's not a tip with hash=$hash not found! But got: $got"
+  }
+  sealed trait BlockAcceptanceError extends BlockServiceError
+  case class BlockAlreadyAccepted(hash: ProofsHash) extends BlockAcceptanceError {
+    override def getMessage: String = s"Block with hash=$hash is already accepted!"
+  }
+  case class BlockIsUnknown(hash: ProofsHash) extends BlockAcceptanceError {
+    override def getMessage: String = s"Block with hash=$hash is unknown so it can't be accepted!"
+  }
+  case class BlockDoesNotExist(hash: ProofsHash) extends BlockAcceptanceError {
+    override def getMessage: String = s"Block with hash=$hash doesn't exist so it can't be accepted!"
+  }
+  sealed trait BlockStoringError extends BlockServiceError
+  case class BlockAlreadyStoredAndWaiting(hash: ProofsHash) extends BlockStoringError {
+    override def getMessage: String = s"Block with hash=$hash is already stored and waiting."
+  }
+  case class BlockAlreadyStoredAndAccepted(hash: ProofsHash) extends BlockStoringError {
+    override def getMessage: String = s"Block with hash=$hash is already stored and accepted."
+  }
+}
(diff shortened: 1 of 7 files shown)


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/UpdateResult.scala
```diff
@@ -0,0 +1,27 @@
...
+import cats.effect.Sync
+import cats.syntax.flatMap._
+
+import scala.util.control.NoStackTrace
+
+import org.typelevel.log4cats.SelfAwareStructuredLogger
+
+sealed trait UpdateResult
+case object Success extends UpdateResult
+case class UnexpectedStateShift(message: String) extends NoStackTrace with UpdateResult {
```
TheMMaciek, line 12: I also thought about adding one more case of non-fatal-unexpected-shift which would only log and return unit.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
```diff
@@ -0,0 +1,144 @@
...
+  def make[F[_]: Sync](blocks: MapRef[F, ProofsHash, Option[StoredBlock]], tipsConfig: TipsConfig): BlockService[F] =
+    new BlockService[F] {
+      implicit val logger = Slf4jLogger.getLogger[F]
+
+      def areParentsAccepted(block: DAGBlock): F[Map[BlockReference, Boolean]] =
+        block.parent.traverse { br =>
+          blocks(br.hash).get.map(sb => br -> sb.exists(_.isInstanceOf[AcceptedBlock]))
+        }.map(_.toNem.toSortedMap)
+
+      def accept(hashedBlock: Hashed[DAGBlock]): F[Unit] =
```
TheMMaciek, line 39: We could assert here that the state is as expected (if we decide that the state can change from waiting to accepted only).

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
```diff
@@ -0,0 +1,144 @@
...
+                  )
+              }
+              .handle
+          )
+          _ <- if (tryAddAsATip)
+            blocks(hashedBlock.proofsHash).modify {
+              case Some(ap @ AcceptedBlock(_, _, usages, false)) if usages.size <= tipsConfig.maximumTipUsages - 1 =>
+                (ap.copy(isTip = true).some, ())
+              case other => (other, ())
+            } else Sync[F].unit
```
marcinwadon, line 119: in general (not related to this piece of code) - is it possible to split this really big function into smaller ones? especially the ones that returns nothing (like this one selected). it can be simplified as
```scala
_ <- removeParents(tipParentsToRemove, ...)
_ <- keepParents(tipParentsToLeave, ...)
_ <- tryAddAsATip(...)
```

marcinwadon (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionService.scala
```diff
@@ -0,0 +1,16 @@
...
+import cats.data.NonEmptyList
+
+import org.tessellation.schema.address.Address
+import org.tessellation.schema.transaction.{Transaction, TransactionReference}
+import org.tessellation.security.Hashed
+import org.tessellation.security.signature.Signed
+
+trait TransactionService[F[_]] {
+  def isParentAccepted(tx: Transaction): F[Boolean]
+  def getLastAcceptedTransactionRef(address: Address): F[TransactionReference]
```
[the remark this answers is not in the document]
kpudlik: +1 because we know that it is `Transaction` because `TransactionService`

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
```diff
@@ -0,0 +1,18 @@
+package org.tessellation.dag.l1.domain.block
+
+import org.tessellation.dag.l1.domain.block.types.PulledTips
+import org.tessellation.dag.types.{BlockReference, DAGBlock}
```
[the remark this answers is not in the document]
TheMMaciek: Not sure if under `l1` package because `l0` will also use it, but the rest is correct I think, I can move it to: `org.tessellation.dag.domain.block` in `dag-shared` module.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
TheMMaciek: That would be a leak of internal representation to the domain. And we need to base on `DAGBlock` and get a `Hashed` of `DAGBlock` to get the `ProofsHash` right. I meant in this comment that I could check during the update that the block that is being accepted is currently in Waiting (so expected state) and only then update. Not sure though if during redownload we won't simply insert blocks as accepted straight away (without the expected shift from waiting -> accepted).

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionService.scala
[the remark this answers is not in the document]
TheMMaciek: Right!

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/UpdateResult.scala
```diff
@@ -0,0 +1,27 @@
...
+}
+
+object UpdateResult {
+  implicit class updateResultOps[F[_]: Sync](result: F[UpdateResult])(implicit logger: SelfAwareStructuredLogger[F]) {
+
+    def handle: F[Unit] =
+      result.flatMap {
+        case Success => Sync[F].unit
+        case e @ UnexpectedStateShift(_) =>
+          logger.warn(e)("Unexpected state shift occurred in block service!") >>
```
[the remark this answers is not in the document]
TheMMaciek: Yeah it's intended to be used in many places. I started with use case in block service and forgot about changing the message.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
```diff
@@ -0,0 +1,144 @@
...
+      def getWaiting: F[Map[ProofsHash, Signed[DAGBlock]]] =
+        blocks.toMap.map(_.collect { case (hash, WaitingBlock(block, _)) => hash -> block })
+
+      def handleTipsUpdate(hashedBlock: Hashed[DAGBlock]): F[Unit] =
+        for {
+          allBlocks <- blocks.toMap
+          allTips = allBlocks.collect { case (hash, ab @ AcceptedBlock(_, _, _, true)) => hash -> ab }
+          parentHashToTip = hashedBlock.signed.value.parent.map(br => br.hash -> allTips.get(br.hash)).toList.toMap
+          notTipParents = parentHashToTip.collect { case (hash, None) => hash }.toList
+          tipParents = parentHashToTip.collect { case (hash, Some(value)) => hash -> value }
```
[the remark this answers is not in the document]
TheMMaciek: Yes I could substract and the outcome should be the same. I'm using collect because I also need to extract the accepted block from the option.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
TheMMaciek: Yeah, you are right! I will do that and it should become more readable.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
```diff
@@ -0,0 +1,144 @@
...
+          remainingTips = allTips -- tipParentsToRemove.keySet
+          minTipHeight = remainingTips.values.map(_.block.signed.value.height.coerce).minOption.getOrElse(0L)
+          tryAddAsATip = remainingTips.size < tipsConfig.maximumTipsCount && minTipHeight < hashedBlock.signed.value.height.coerce
+          _ <- notTipParents.traverse(
+            blocks(_)
+              .modify[UpdateResult] {
+                case Some(ap @ AcceptedBlock(_, _, usages, false)) =>
+                  (ap.copy(usages = usages + hashedBlock.proofsHash).some, Success)
+                case unexpected =>
+                  (unexpected, UnexpectedStateShift(s"Expected Accepted block that IS NOT a tip! Got $unexpected."))
```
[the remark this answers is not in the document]
TheMMaciek: Makes sense. I think the second option is a good idea (writing out all allowed states could be tedious). I will refactor.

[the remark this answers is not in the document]
TheMMaciek: Though I would keep it a string defined in the place of use to give a possibility of customization. We just need to follow some guidelines while writing the message.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
marcinwadon: hm, I think that the block state could be also a part of a domain.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
```diff
@@ -0,0 +1,17 @@
...
+import org.tessellation.security.signature.Signed
+
+import eu.timepit.refined.types.numeric.PosInt
+
+trait BlockService[F[_]] {
+  def areParentsAccepted(block: DAGBlock): F[Map[BlockReference, Boolean]]
+  def accept(hashedBlock: Hashed[DAGBlock]): F[Unit]
+  def store(hashedBlock: Hashed[DAGBlock]): F[Unit]
+  def getWaiting: F[Map[ProofsHash, Signed[DAGBlock]]]
+  def handleTipsUpdate(hashedBlock: Hashed[DAGBlock]): F[Unit]
```
marcinwadon, line 15: btw, shouldn't `handleTipsUpdate` be private and executed internally by `accept`?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
```diff
@@ -0,0 +1,17 @@
...
+  def pullTips(tipsCount: PosInt): F[Option[PulledTips]]
```
marcinwadon, line 16: maybe just `Tips` instead of `PulledTips`. does it need to be a different type?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
```diff
@@ -0,0 +1,173 @@
...
+        blocks(blockReference.hash).get.map(_.exists(_.isInstanceOf[AcceptedBlock]))
+
+      private def updateNoTipParents(parents: List[ProofsHash], hashedBlock: Hashed[DAGBlock]): F[Unit] =
+        parents.traverse { hash =>
+          blocks(hash)
+            .modify[Either[TipsUpdateError, Unit]] {
+              case Some(accepted @ AcceptedBlock(_, _, _, false)) =>
+                (accepted.addUsage(hashedBlock).some, ().asRight)
+              case unexpected =>
+                (unexpected, NotTipAcceptedBlockNotFound(hash, unexpected).asLeft)
```
marcinwadon, line 111: `not tip accepted block not found` - sorry, but I don't understand it :( 
tip is not accepted? or block is not found? or what? :)

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
```diff
@@ -0,0 +1,173 @@
...
+        }.void
+
+      private def updateTipParentsToRemove(parents: List[ProofsHash], hashedBlock: Hashed[DAGBlock]): F[Unit] =
+        parents.traverse { hash =>
+          blocks(hash)
+            .modify[Either[TipsUpdateError, Unit]] {
+              case Some(accepted @ AcceptedBlock(_, _, _, true)) =>
+                (accepted.addUsage(hashedBlock).unsetAsTip.some, ().asRight)
+              case unexpected =>
+                (unexpected, TipBlockToUnsetAsTipNotFound(hash, unexpected).asLeft)
```
marcinwadon, line 123: same here

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
```diff
@@ -0,0 +1,173 @@
...
+        }.void
+
+      private def updateTipParentsToRemain(parents: List[ProofsHash], hashedBlock: Hashed[DAGBlock]): F[Unit] =
+        parents.traverse { hash =>
+          blocks(hash)
+            .modify[Either[TipsUpdateError, Unit]] {
+              case Some(accepted @ AcceptedBlock(_, _, _, true)) =>
+                (accepted.addUsage(hashedBlock).some, ().asRight)
+              case unexpected =>
+                (unexpected, TipBlockToRemainAsTipNotFound(hash, unexpected).asLeft)
```
marcinwadon, line 135: and here as well

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
```diff
@@ -0,0 +1,173 @@
...
+      extends StoredBlock {
+    def addUsage(hashedBlock: Hashed[DAGBlock]): AcceptedBlock = this.focus(_.usages).modify(_ + hashedBlock.proofsHash)
+    def setAsTip: AcceptedBlock = this.focus(_.isTip).replace(true)
+    def unsetAsTip: AcceptedBlock = this.focus(_.isTip).replace(false)
+  }
+
+  sealed trait BlockServiceError extends NoStackTrace
+  sealed trait TipsUpdateError extends BlockServiceError
+  case class TipBlockToUnsetAsTipNotFound(hash: ProofsHash, got: Option[StoredBlock]) extends TipsUpdateError {
+    override def getMessage: String = s"Tip block to unset as tip with hash=$hash not found! But got: $got"
```
marcinwadon, line 165: why not just `Tip not found` ?

marcinwadon (changes requested):
comments before

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
TheMMaciek: Was thinking about it. I intended it to run synchronously as a next step in block acceptance flow (in the upcoming PRs), so semantically it will be the same (it could run in acceptance flow or here as a next step after accept). For now I'm not sure if we should hide it behind the interface, as I'm not sure how will the whole redownload flow work. The most appealing scenario is that we would only accept blocks one after the other and the tips will be calculated based on these dynamically no matter the case (normal acceptance, acceptance after redownload). But it could also be the case that during redownload we will only want a certain pool of blocks to go through the tip updating. I wouldn't hide it for now.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
TheMMaciek: It means: Accepted block that's not a tip wasn't found. The name starting with `Not` maybe misleading but if I change it to `NoTipAcceptedBlockNotFound` I think it works.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
marcinwadon: so I'd create a second function for redownload specifically that will override some checks and force tips update, wdyt?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
TheMMaciek: I think this and the one below are good enough. The one above may be misleading so I will try to figure out the better name. This says that: a block that is a tip and was gonna be removed from tips wasn't found (when it should be found). And I think the name reflects that.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
TheMMaciek: Because I would like to have a concrete error for every case. If I have more generic `TipNotFound` I won't be able to tell during which update functionality (remaining or removing as a tip) it happened. I think we are allowed to have such precise error messages in the implementation.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
TheMMaciek: I guess that depends on whether redownload will be able to reuse this one or not. If yes I would keep it in the interface to not have any functionality duplication, if no and redownload will need a separate function then we can hide it. Don't know at this point. I can create a ticket to reevaluate at some point.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
TheMMaciek: I guess there is no need right now. The `BlockService` uses it internally. Whether the block is kept in the `Waiting/Accepted` classes in the `BlockService` or in any other way isn't important I guess. If L0 needs it for something then we may extract it. Though I think no interface should rely on getting `Accepted` or `Waiting` block. We should just base on `DAGBlock` wrapped in a `Signed` (for external communication) and `Hashed` for local.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
marcinwadon: So I'd propose a general type like `TipNotFound` or `BlockNotFound` with specific details inside, like a part of round, etc.
```
case class BlockNotFound( shouldBeTip: yes | no, state: Any | Accepted | ..., hash: ...  ) ...
```

Or just more verbose names like `NoTipBlockNotFound(when: Accepted)` or `TipBlockNotFound(when: Unset)` or sth like that 🤔

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
kpudlik: We use `ToRemove` in 116 but `ToUnset` here. So either remove or unset. I would avoid mixing both. Name is fine for me as it fits `unsetAsTip`

kpudlik (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
marcinwadon: Ok, please do so

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
```diff
@@ -0,0 +1,173 @@
...
+          parentHashToTip = hashedBlock.signed.value.parent.map(ref => ref.hash -> allTips.get(ref.hash)).toList.toMap
+          tipParents = parentHashToTip.collect { case (hash, Some(accepted)) => hash -> accepted }
+          notTipParents = (parentHashToTip -- tipParents.keySet).keySet
+          tipParentsSize = tipParents.size
+          tipParentsToRemove = tipParents.filter {
+            case (_, block) =>
+              allTips.size >= tipsConfig.minimumTipsCount + tipParentsSize &&
+                (
+                  allTips.size > tipsConfig.maximumTipsCount + tipParentsSize ||
+                    block.usages.size >= tipsConfig.maximumTipUsages - 1
```
marcinwadon, line 74: why `-1`? can you assign these conditions to some expressive names, like:
```scala
val aboveOrEqualRequiredMinimum = ...
val aboveRequiredMinimum
val areParentsUsedEnoughTimes
```
or sth like that

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
```diff
@@ -0,0 +1,173 @@
...
+                (unexpected, TipBlockToRemainAsTipNotFound(hash, unexpected).asLeft)
+            }
+            .flatMap(_.liftTo[F])
+        }.void
+
+      private def tryAddAsATip(condition: Boolean, hashedBlock: Hashed[DAGBlock]): F[Unit] =
+        if (condition)
+          blocks(hashedBlock.proofsHash).modify {
+            case Some(accepted @ AcceptedBlock(_, _, usages, false))
+                if usages.size <= tipsConfig.maximumTipUsages - 1 =>
```
marcinwadon, line 144: I see the same condition twice

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
TheMMaciek: `-1` because after adding a usage the `maximumTipUsages` count will be reached. Yeah I will try to extract conditions to properly named vals.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
TheMMaciek: I will check if somehow extracting it for both cases makes sense.

TheMMaciek: This one ones simply needs to be `< maximumTipUsages` don't know why I did `<=` and subtracted `1`.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
TheMMaciek: If that's ok I would still like to go with the current approach. If we write the rest of the implementation correctly we shouldn't get many more errors like this added to the service. And I think it is easier to pinpoint where the error happened with a concrete error than with a more generic one. I like the one-to-one correspondence between an error and a place in the code.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
```diff
@@ -0,0 +1,222 @@
+package org.tessellation.dag.l1.infrastructure.block
```
marcinwadon, line 1: service should be in the domain

marcinwadon (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
```diff
@@ -0,0 +1,222 @@
...
+      private def updateNoTipParents(parents: List[ProofsHash], hashedBlock: Hashed[DAGBlock]): F[Unit] =
+        parents.traverse { hash =>
+          blocks(hash)
+            .modify[Either[TipsUpdateError, Unit]] {
+              case Some(accepted @ AcceptedBlock(_, _, _, false)) =>
+                (accepted.addUsage(hashedBlock.proofsHash).some, ().asRight)
+              case unexpected =>
+                (unexpected, NoTipAcceptedBlockNotFound(hash, unexpected).asLeft)
+            }
+            .flatMap(_.liftTo[F])
```
marcinwadon, line 125: what if it fails? one parent will have usage added, second not?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
```diff
@@ -0,0 +1,222 @@
...
+      private def updateTipParentsToRemove(parents: List[ProofsHash], hashedBlock: Hashed[DAGBlock]): F[Unit] =
+        parents.traverse { hash =>
+          blocks(hash)
+            .modify[Either[TipsUpdateError, Unit]] {
+              case Some(accepted @ AcceptedBlock(_, _, _, true)) =>
+                (accepted.addUsage(hashedBlock.proofsHash).unsetAsTip.some, ().asRight)
+              case unexpected =>
+                (unexpected, TipBlockToRemoveAsTipNotFound(hash, unexpected).asLeft)
+            }
+            .flatMap(_.liftTo[F])
```
marcinwadon, line 137: same here. `liftTo` will throw an error and stop traversing

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
```diff
@@ -0,0 +1,222 @@
...
+        }.void
+
+      private def tryAddAsATip(condition: Boolean, hashedBlock: Hashed[DAGBlock]): F[Unit] =
+        if (condition)
+          blocks(hashedBlock.proofsHash).modify {
+            case Some(accepted @ AcceptedBlock(_, _, usages, false)) if usages.size < tipsConfig.maximumTipUsages =>
+              (accepted.setAsTip.some, ())
+            case other => (other, ())
+          } else Sync[F].unit
```
marcinwadon, line 158: Applicative unit :P

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
TheMMaciek: The `BlockService` trait is in the domain. Should I also move this concrete implementation to the domain?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
TheMMaciek: If the first parent fails the update for the second one won't be attempted, that's right. Not 100% sure if that's how it should be.

buckysballs: If any of the parents are invalid, the block is invalid. So we should only accept both or none

TheMMaciek: And yes if the first one passes and the second fails we won't rollback the update for the first parent - so it's not atomic. Don't know if we could even make it atomic. At this point the block is validated and these cases shouldn't happen - if they happen it means we have some bug in our code and we need to fix it. I don't see a way to do it any better at this point. For the case of the tips usage it's not crucial, but in general it's the matter of atomic state shifts.

kpudlik (approved):

kpudlik (approved):

marcinwadon (approved):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/BlockService.scala
buckysballs: If any of the parents are invalid, the block is invalid. So we should only accept both or none

PR #97, opened 2021-12-20
Move node storage and gossip algorithm to the SDK

marcinwadon:


tbekas (approved):
LGTM

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/gossip/p2p/GossipClient.scala
```diff
@@ -0,0 +1,36 @@
...
+
+import org.tessellation.kryo.KryoSerializer
+import org.tessellation.schema.gossip._
+import org.tessellation.sdk.http.p2p.PeerResponse
+import org.tessellation.sdk.http.p2p.PeerResponse.PeerResponse
+
+import org.http4s.Method._
+import org.http4s.client.Client
+
+trait GossipClient[F[_]] {
```
TheMMaciek, line 13: Was not removing the gossip client from core purposeful?

PR #98, opened 2021-12-21
Proximity / neighbor / expanded sybil attacks & data serialization

ryle-goehausen:
Use pre-generated networks committed to resources with data
generation test to build 10 separate 100 node networks.

Proximity test to demonstrate use of ET. Variation with first
and second neighbors. Compound sybil test with single attack
and half compromise. More variations required but demonstrates
basics.


modules/core/src/test/resources/trust_100x10.json
```diff
@@ -0,0 +1,1000 @@
+{"id":0,"xCoordinate":0.5824825788177106,"yCoordinate":0.1554504534401322,"edges":[{"src":0,"dst":7,"trust":0.6916835694196743,"isLabel":true},{"src":0,"dst":21,"trust":0.12426747820353046,"isLabel":true},{"src":0,"dst":45,"trust":0.8724780028867717,"isLabel":true},{"src":0,"dst":58,"trust":0.9578025631164443,"isLabel":true},{"src":0,"dst":59,"trust":0.8680184625097882,"isLabel":true},{"src":0,"dst":66,"trust":0.4802369595509173,"isLabel":true},{"src":0,"dst":68,"trust":0.7511380674851966,"isLabel":true},{"src":0,"dst":70,"trust":0.8725482905272447,"isLabel":true},{"src":0,"dst":76,"trust":0.48550843711192526,"isLabel":true},{"src":0,"dst":82,"trust":0.8195464092924847,"isLabel":true},{"src":0,"dst":84,"trust":0.6253358451732758,"isLabel":true},{"src":0,"dst":86,"trust":0.41444862173569386,"isLabel":true},{"src":0,"dst":98,"trust":0.9367671106108932,"isLabel":true}]}
```
kpudlik: Does it mean that test works only for that particular case from `trust_100x10.json` but may fail for randomly generated ones? Are these tests deterministic?

modules/core/src/test/scala/org/tessellation/trust/SybilAttackSuite.scala
```diff
@@ -1,37 +1,134 @@
...
+          case (col, next) =>
+            next.flatMap { n =>
+              col.map(_ :+ n)
+            }
+        }
+    } yield {
+      val ser = networkData.flatMap { _.map { _.asJson.noSpaces } }
+      val res = decode[TrustNode](ser.head)
+      Files
+        .write(Paths.get("./src/test/resources/trust_100x10.json"), ser.mkString("\n").getBytes(StandardCharsets.UTF_8))
```
marcinwadon, line 103: it's impure, shouldn't be in `yield` like that

modules/core/src/test/scala/org/tessellation/trust/SybilAttackSuite.scala
```diff
@@ -1,37 +1,134 @@
...
+      expect(res.isRight)
+    }
+  }
+
+  def loadSerialized(): IO[List[List[TrustNode]]] = {
```
marcinwadon, line 108: Your test was non-deterministic. It means that for given preconditions there are some values that could lead to a failure. I'd advise not to lock the test on the optimistic version (what you did right now) but to fix the algorithm to pass all the edge cases. Otherwise, you're introducing a test that checks if you created a correct json file, but doesn't check if the algorithm works properly for the given arbitrary data.

ryle-goehausen: Most of the non-determinism is coming from the old SAW module which I'm incrementally refactoring. The 'optimistic' version here should absolutely remain as a purely deterministic version -- it verifies given known data that we get precise outputs. DATT for instance is purely deterministic and has no randomness associated with it -- so fixed inputs are totally fine and yield exact outputs. The current SAW outputs -- even on fixed inputs, are not deterministic at all due to random seed issue which also needs to be fixed. 

That's one level of tests, but doesn't evaluate model performance as properly as possible. I actually want a second level of unit tests that covers more dynamically generated configurations -- but there's several edge cases here that require much more refactoring to eliminate (primarily due to the old SAW code which cannot be easily fixed immediately.) For now though, I don't want to introduce accidentally failing tests -- and these at least demonstrate some level of security which can be improved upon.

Evaluating the actual model and fixing hyper-parameters for the model is completely outside the scope of a unit test, and belongs as part of a model training framework. It will take too long (on the order of hours of compute time) to properly evaluate a model on many many network configurations. That is what we need to demonstrate proper security anyways -- any unit test is not going to capture enough configurations to demonstrate anything beyond a very basic proof that shows the code makes sense -- that's the only thing being shown here are a few attack scenarios to demonstrate reasons for current model.

modules/core/src/test/resources/trust_100x10.json
ryle-goehausen: Answered in lower comment -- tests are avoiding some of the non-determinism from SAW, there's still a wild amount of variance on random configurations (which can be fixed in future tests.) Either way -- we need both types of tests, some that verify exact expected outputs, others that do dynamic configurations. It's fine to start with exact for now until I can eliminate some of the variance from SAW -- and actual model proofs will be done as part of an evaluator outside of unit tests.

modules/core/src/test/scala/org/tessellation/trust/SybilAttackSuite.scala
ryle-goehausen: Main purpose of these tests also is to prevent any regressions while model is being developed -- future development should not substantially break these without manual oversight of changes in scores. Fully proving security is not complete, but this is a good place to start.

modules/core/src/test/scala/org/tessellation/trust/SybilAttackSuite.scala
ryle-goehausen: Changed to IO.delay in the loop -- if there's more idiomatic way let me know.

marcinwadon (approved):

PR #105, opened 2021-12-23
Move TrustPush and related classes to new trust module

ryle-goehausen:
Not all trust related classes moved as part of this PR, only main interfaces
to the rest of the code.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -284,9 +284,32 @@ lazy val dagL1 = (project in file("modules/dag-l1"))
       Libraries.refinedCore
     )
   )
+
+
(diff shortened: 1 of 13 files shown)


marcinwadon:
Why a new project? Please keep it in the core project (because it's strictly related with L0 nodes) but in the `trust` module.
I told you that I see `TrustPush` program that is in the `cluster` module, but you created a separate `trust` module and didn't move `TrustPush` there.
image

We don't need a separate project just for trust 🤔

ryle-goehausen:
@marcinwadon my mistake, will open a new PR, misunderstood.

PR #106, opened 2021-12-28
DAG state channel block consensus mechanism

TheMMaciek:


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/http/p2p/clients/BlockConsensusClient.scala
```diff
@@ -0,0 +1,29 @@
...
+import org.http4s.circe.CirceEntityCodec.circeEntityEncoder
+import org.http4s.client.Client
+
+trait BlockConsensusClient[F[_]] {
+  def sendConsensusData(data: Signed[PeerBlockConsensusData]): PeerResponse[F, Unit]
+}
+
+object BlockConsensusClient {
+
+  def make[F[_]: Sync](client: Client[F]): BlockConsensusClient[F] =
```
marcinwadon, line 21: just fyi: most likely `Concurrent` will be required for http4s req/res signer

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/storage/ConsensusStorage.scala
```diff
@@ -0,0 +1,24 @@
...
+
+import org.tessellation.dag.l1.domain.consensus.RoundId
+import org.tessellation.dag.l1.domain.consensus.block.RoundData
+
+import io.chrisdavenport.mapref.MapRef
+
+case class ConsensusStorage[F[_]](
+  ownConsensus: Ref[F, Option[RoundData]],
+  peerConsensuses: MapRef[F, RoundId, Option[RoundData]]
+)
```
marcinwadon, line 15: If it's a kind of service/storage/repository, I'd not use `case class`.

marcinwadon (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/RoundId.scala
```diff
@@ -0,0 +1,9 @@
+package org.tessellation.dag.l1.domain.consensus
+
+import java.util.UUID
+
+import derevo.circe.magnolia.{decoder, encoder}
+import derevo.derive
+
+@derive(encoder, decoder)
+case class RoundId(uuid: UUID)
```
kpudlik, line 9: Shouldn't we use `newtype` here?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/AlgebraCommand.scala
```diff
@@ -0,0 +1,20 @@
+package org.tessellation.dag.l1.domain.consensus.block
+
+import org.tessellation.dag.l1.domain.consensus.block.BlockConsensusData.{
+  BlockProposal,
+  CancelledBlockCreationRound,
+  Proposal
+}
+import org.tessellation.kernel.Ω
+
+sealed trait AlgebraCommand extends Ω
```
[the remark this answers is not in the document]
kpudlik: @marcinwadon `RoundStarFailure` is not a step as well  :D

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/BlockConsensusContext.scala
```diff
@@ -0,0 +1,24 @@
...
+
+case class BlockConsensusContext[F[_]](
+  blockConsensusClient: BlockConsensusClient[F],
+  blockService: BlockService[F],
+  blockValidator: BlockValidator[F],
+  clusterStorage: ClusterStorage[F],
+  consensusConfig: ConsensusConfig,
+  consensusStorage: ConsensusStorage[F],
+  keyPair: KeyPair,
+  myId: PeerId,
```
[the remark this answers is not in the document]
kpudlik: Or `selfId`  as used in `Joining`. Even more consistent.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/storage/ConsensusStorage.scala
```diff
@@ -0,0 +1,24 @@
...
+
+object ConsensusStorage {
+
+  def make[F[_]: Sync]: F[ConsensusStorage[F]] =
+    for {
+      peerConsensuses <- MapRef.ofConcurrentHashMap[F, RoundId, RoundData]()
+      ownConsensus = Ref.unsafe(none[RoundData])
```
kpudlik, line 22:
```scala
ownConsensus <- Ref.of[F, Option[RoundData]](none)
```

kpudlik (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/AlgebraCommand.scala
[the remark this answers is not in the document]
marcinwadon: right @kpudlik, it's either an outcome/result or an error

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/RoundId.scala
TheMMaciek: We could I guess.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/storage/ConsensusStorage.scala
TheMMaciek: Right!

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/AlgebraCommand.scala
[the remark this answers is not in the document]
TheMMaciek: Right I can change the name to `InformAboutRoundStartFailure` - it would be consistent then.
Regarding the naming I think we should use the same convention across the project, though I must admit I like the `Command` convention a lot. `Step` is kind of passive/impersonal for me, whereas `Command` is more vibrant and suiting to what's actually happening - that's how I use it also `PersistBlockProposal`, `StartOwnRound` etc. For me that's more descriptive. Also a step suggests that after you make first step, you will then make a second one, whereas here it's not true - you can make the same step many times (process proposal for one peer than for the other etc.) and also the cell is stateless in a way (it reads and modifies the data living in a storage) so in a way it can even go backwards (it's not tied to the concrete state). That's my thoughts. WDYT?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/CoalgebraCommand.scala
```diff
@@ -0,0 +1,16 @@
+package org.tessellation.dag.l1.domain.consensus.block
+
+import org.tessellation.dag.l1.domain.consensus.block.BlockConsensusData.{
+  BlockProposal,
+  CancelledBlockCreationRound,
+  Proposal
+}
+
+sealed trait CoalgebraCommand
```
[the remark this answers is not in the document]
TheMMaciek: It's only used inside of the Cell, we have `ConsensusData` extending `Ohm` and coming from the outside to the cell, `CoalgebraCommand` which is only living inside of the cell and doesn't need to extend `Ohm` because of that, `AlgebraCommand` which extends `Ohm` only because using `StackF` requires that and `ConsensusOutput` which is an output of the cell and extends `Ohm`.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/RoundData.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  def addPeerProposal(proposal: Proposal): RoundData =
+    this.focus(_.peerProposals).modify(_ + (proposal.senderId -> proposal))
+
+  def addOwnBlock(block: Signed[DAGBlock]): RoundData = this.focus(_.ownBlock).replace(block.some)
+
+  def addPeerBlock(blockProposal: BlockProposal): RoundData =
+    this.focus(_.peerBlocks).modify(_ + (blockProposal.senderId -> blockProposal.signedBlock))
+
+  def addOwnCancellation(reason: CancellationReason): RoundData = this.focus(_.ownCancellation).replace(reason.some)
```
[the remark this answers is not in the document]
TheMMaciek: Sure, I will use `set` for singular items, it makes sense.
I guess it's not required to keep both peers and own cancellations separately, but I think it's more convenient to do so. It's also consistent with how I keep own proposal and own block, which are used in different places (especially own proposal). If I would keep them in the collections then extracting them from the collection every time I need them would be cumbersome - I wouldn't be able to match the `RoundData` simply. And there are cases where event is handled differently whether for example own cancellation is there or not. So I think it's better to have own data in separate fields, as it simplifies the code and makes it less boilerplate'y.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/RoundData.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  def addPeerCancellation(cancellation: CancelledBlockCreationRound): RoundData =
+    this.focus(_.peerCancellations).modify(_ + (cancellation.senderId -> cancellation.reason))
+
+  def formBlock(): DAGBlock = DAGBlock(
```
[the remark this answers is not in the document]
TheMMaciek: I guess it would kind of suggest that we are transforming `RoundData` to a `Block` which is not the case. We are using some of the information stored inside of `RoundData` to form a block. After we form it we are even storing it inside of the `RoundData` again.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/storage/ConsensusStorage.scala
TheMMaciek: I can use a `class` but both `ownConsensus` and `peerConsensuses` would still need to be `vals`, because I need to access them directly in `BlockConsensusCell`. Making proxy functions to access modify e.g. doesn't make sense I think.

kpudlik (approved):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/BlockConsensusCell.scala
```diff
@@ -0,0 +1,530 @@
...
+        case OwnRoundTrigger                           => StartOwnRound
+        case proposal: Proposal                        => ProcessProposal(proposal)
+        case blockProposal: BlockProposal              => ProcessBlockProposal(blockProposal)
+        case cancellation: CancelledBlockCreationRound => ProcessCancellation(cancellation)
+      }
+    )
+
+object BlockConsensusCell {
+
+  private def deriveConsensusPeerIds(proposal: Proposal, myId: PeerId): Set[PeerId] =
```
marcinwadon, line 97: `myId` -> `selfId`

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/BlockConsensusCell.scala
```diff
@@ -0,0 +1,530 @@
...
+    } yield ()
+
+  private def broadcast[F[_]: Async](
+    data: Signed[PeerBlockConsensusInput],
+    peers: Set[Peer],
+    blockConsensusClient: BlockConsensusClient[F]
+  ): F[Unit] =
+    peers.toList
+      .traverse(blockConsensusClient.sendConsensusData(data)(_))
+      .void
```
marcinwadon, line 133: what about gossip usage?

marcinwadon (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/BlockConsensusCell.scala
marcinwadon: already discussed

marcinwadon (approved):

PR #107, opened 2021-12-28
Base TessellationIOApp

marcinwadon:


modules/shared/src/main/scala/org/tessellation/ext/cats/effect/package.scala
```diff
@@ -0,0 +1,9 @@
+package org.tessellation.ext.cats
+
+import _root_.cats.effect.{IO, Resource}
```
tbekas, line 3: Where does this `_root_` package comes from? Are we repackaging cats?

tbekas (approved):
LGTM

modules/shared/src/main/scala/org/tessellation/ext/cats/effect/package.scala
kpudlik: @tbekas  You can't use `cats.effect.{IO, Resource}` because you are already in `cats` package `org.tesselation.ext.cats` so it will try to find `IO` and `Resource` in `org.tesselation.ext.cats.{IO, Resource}` (relatively).

`_root_` is a syntax to force absolute import instead of relative one

kpudlik (approved):

PR #109, opened 2022-01-04
Use SDK in L0 layer

marcinwadon:


modules/core/src/main/scala/org/tessellation/Main.scala
```diff
@@ -14,101 +13,78 @@ import org.tessellation.http.p2p.P2PClient
...
 import com.monovore.decline.Opts
-import com.monovore.decline.effect.CommandIOApp
-import org.typelevel.log4cats.slf4j.Slf4jLogger
 
 object Main
-    extends CommandIOApp(
+    extends TessellationIOApp[Run](
       name = "",
-      header = "Constellation Node",
+      header = "Tessellation Node",
```
TheMMaciek, line 31: Maybe we should cal it `Tessellation L0 Node` 🤔

modules/core/src/main/scala/org/tessellation/Main.scala
```diff
@@ -14,101 +13,78 @@ import org.tessellation.http.p2p.P2PClient
...
+    kryoPool: KryoSerializer[IO]
+  ): Resource[IO, Unit] = {
+    val cfg = method.appConfig
 
-                          _ <- services.stateChannelRunner.initializeKnownCells.asResource
+    Database.forAsync[IO](cfg.dbConfig).flatMap { implicit database =>
+      for {
+        res <- AppResources.make[IO](cfg)
+        nodeId = PeerId.fromPublic(keyPair.getPublic)
+        _ <- logger.info(s"This peerId ${nodeId.show}").asResource
```
TheMMaciek, line 51: Similar message is already logged in `TessellationIOApp`, not sure it needs to be logged here as well.

TheMMaciek (approved):

modules/core/src/main/scala/org/tessellation/Main.scala
marcinwadon: :+1: I'll remove it once unifying services

PR #110, opened 2022-01-04
DAG state channel address storage

TheMMaciek:


modules/dag-l1/src/main/resources/db.migration/V1__AddressTable.sql
```diff
@@ -0,0 +1,5 @@
+create table if not exists Address (
+    address varchar(40) primary key,
+    balance bigint default 0
```
marcinwadon, line 3: on L1 we would need also a majority balance

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/address/storage/AddressStorage.scala
```diff
@@ -0,0 +1,9 @@
+package org.tessellation.dag.l1.domain.address.storage
+
+import org.tessellation.schema.address.Address
+import org.tessellation.schema.balance.Balance
+
+trait AddressStorage[F[_]] {
+  def getBalance(address: Address): F[Balance]
+  def updateBalances(addressBalances: Map[Address, Balance]): F[Unit]
```
marcinwadon, line 8: I'd make two separate functions:
```
setBalance(address, balance)
```
and
```
transfer(source, destination)
```

The latter will allow using storage (or db) locks to make the transfer transactional

TheMMaciek: Currently I only need the transactional batch update. Hence `updateBalances` function. I could extract part of its implementation to  private `setBalance` that would return Quills abstraction `Free[ConnectionOp, Unit]` that could be reused once we need to set balance for a single address transactionally. Unfortunately I can't write transactional function for single transfer and then execute it again transactionally for multiple transfers. And I want either all to happen here or none here so I can't execute sequential transactional updates here.

modules/dag-l1/src/main/resources/db.migration/V1__AddressTable.sql
TheMMaciek: Right. But I guess we should add needed pieces the moment they are needed, so here that would be added with majority snapshot processing I think.

marcinwadon: I'm fine with it :+1:

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/address/storage/AddressStorage.scala
marcinwadon: Ok, got it. You want to update balances for the whole block or snapshot at once. I thought about accepting single txs. If so, it's not needed - you can ignore my comment ;)

marcinwadon (approved):

PR #112, opened 2022-01-04
DAG state channel new block service

TheMMaciek:


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockProcessor.scala
```diff
@@ -0,0 +1,104 @@
...
+            .liftTo[F]
+        } yield (address, newBalance)
+    }.map(_.toMap)
+  }
+
+  private def acceptTransactions(hashedTransactions: Set[Hashed[Transaction]]): F[Unit] =
+    for {
+      preparedBalances <- prepareBalances(hashedTransactions.map(_.signed.value))
+      _ <- addressStorage.updateBalances(preparedBalances)
+      _ <- hashedTransactions.toList.traverse(transactionService.accept)
```
marcinwadon, line 92: I think that it's a responsibility of transaction service to accept balances accordingly. If you want to accept multiple txs at once, I'd create a function that receives set of txs and accepts them all or nothing. Anyway, looks like you update balances before you accept the transaction and I think it's wrong

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockProcessor.scala
```diff
@@ -0,0 +1,104 @@
...
+    } yield ()
+}
+
+object BlockProcessor {
+  sealed trait BlockProcessingError extends NoStackTrace
+  case class BlockFailedValidation(reasons: NonEmptyList[BlockValidationError]) extends BlockProcessingError {
```
marcinwadon, line 98: Maybe just BlockInvalid? wdyt?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockProcessor.scala
```diff
@@ -0,0 +1,104 @@
...
+        address -> change
+    }
+
+    addressBalanceChanges.toList.traverse {
+      case (address, balanceChange) =>
+        for {
+          current <- addressStorage.getBalance(address)
+          newBalance <- refineV[NonNegative]
+            .apply(current.coerce + balanceChange)
+            .bimap(BalanceNegative, Balance(_))
```
marcinwadon, line 82: How balance can be negative if transactions are validated upfront (or should be) and applied in chain? A transaction that spends more is simply invalid and should be already rejected

marcinwadon (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockProcessor.scala
```diff
@@ -0,0 +1,104 @@
...
+import org.tessellation.schema.transaction.Transaction
+import org.tessellation.security.signature.Signed
+import org.tessellation.security.{Hashed, SecurityProvider}
+
+import eu.timepit.refined.auto._
+import eu.timepit.refined.numeric.NonNegative
+import eu.timepit.refined.refineV
+import io.estatico.newtype.ops.toCoercibleIdOps
+
+class BlockProcessor[F[_]: Async: KryoSerializer: SecurityProvider](
```
marcinwadon, line 34: btw - can it be simply a BlockService? I think we don't need to create additional services and split the logic between them

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockProcessor.scala
TheMMaciek: `refineV.apply` returns an `Either[String, Refined]` so I need to handle that. If there is no issue in validations and flow this error should never be thrown. If it throws the reason would be that the Balance got negative hence the name.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockProcessor.scala
TheMMaciek: Yeah I guess that makes sense.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockProcessor.scala
TheMMaciek: We would need to switch from `MapRef` to `Ref[Map[K, V]]` to be able to do an atomic update on transactions. Although it still won't give us complete atomicity unfortunately (on block acceptance level) because we store balances in a different cache (database actually) and blocks in yet different one. Theoretically any of these may fail so if that actually happens during the updates we have a problem either way. I think I intended the balances update to happen first because given that we chose to store these inside of a database (and the database seems like the most erratic of the storages we use here) there is the biggest probability that block acceptance will fail at that moment - if it does it's still ok, balances update failed and nothing else got accepted so we may fail the whole flow with no consequences. After accepting balances everything after that should succeed and the validations and synchronized acceptance of blocks should enforce that. But from the local reasoning point of view I can't guarantee that if the transactions got accepted the block will also get accepted these are side effects wrapped in `F` so they may throw. Thinking about a solution for that case (because I don't see a way to atomically change all three caches at the same time), after an unexpected error is thrown we can make the node switch to a `Misaligned` status and not perform any operations until new majority globalSnapshot is received and then if the node is Misaligned that would trigger a redownload to that new majority right of the bat - we could also eagerly redownload to the last known majority. I planned to create a ticket regarding atomic acceptance and I think it makes sense to do so and try to spike this case there.

Regarding the `transactionService` being responsible for balance update I'm not sure, after all `transactionService` doesn't store the balances so why would it be responsible for updating them. Also if we start to nest classes the solution may get less elegant. `BlockService` would need `TransactionService` which would need `AddressStorage`, so it's kind of getting a bit convoluted. I strived towards flatter structure and decoupling to avoid circular dependencies. Just my thoughts, not saying that's the way to go.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockProcessor.scala
TheMMaciek: I will give it some thought. If we move it to `BlockService` then it will require both `TransactionService` and `StorageService` inside, if we further move the balance acceptance to `TransactionService` we will need to have the `AddressStorage` there as well. So it's getting nested and in the end `BlockService` will be kind of doing everything. I would rather go with flatter structure with dependencies between services decreased to minimum. But that's just what I feel comfortable with. I will look into it and see - maybe I'm overdramatic.

tbekas: I am all in favor of having many small services rather than a few bigger ones.

tbekas (approved):
LGTM

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockProcessor.scala
marcinwadon: Me too, but this is what we had in v1.0 and because of many dependent services we decided to combine them to reduce complexity. And now, we’re doing something opposite 🤔

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockProcessor.scala
marcinwadon: But it means that when I take the transactionService and call “accept”, the transaction will not be fully accepted 🤔

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockProcessor.scala
kpudlik: You can always have many smaller ones and use fascade pattern by creating single `BlockService`

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockProcessor.scala
kpudlik: I am wondering if we can omit that validation if we set both `current` and `balanceChange` as `NonNegative`

TheMMaciek: Balance change is a `BigInt` and it may be negative. If you send dag to someone, the balance change will be negative.

marcinwadon (approved):

PR #114, opened 2022-01-05
ADR - Architecture decision record

marcinwadon:


TheMMaciek (approved):

tbekas (approved):
Well done. Let's wait for all approvals, so that it's communicated to all contributors that we're using ADRs.

PR #115, opened 2022-01-05
ADRs for snapshots

marcinwadon:


doc/adr/0004-global-snapshot-trigger.md
```diff
@@ -0,0 +1,23 @@
...
+
+Date: 2022-01-05
+
+## Status
+
+Proposed
+
+## Context
+
+Global snapshot must have a deterministic trigger so then snapshots created on
```
kpudlik, line 11: This is the main issue @buckysballs. We should discuss that on Friday and you can put your thoughts here anyway.

doc/adr/0002-global-and-state-channel-snapshots.md
```diff
@@ -0,0 +1,37 @@
...
+periodically creating StateChannel snapshots inside the L0 Cell. Overall, the
+StateChannel snapshot needs to be returned as an output of an L0 Cell so then it
+will be aggregated by Layer 0 and proposed to the Global snapshot.
+
+### Global snapshots
+
+Global snapshots are being created on Layer 0. They contains both DAG blocks and
+StateChannel snapshots. Trigger for the global snapshot will be described by its
+own ADR. Global snapshots that passed majority algorithm can be considered final
+and immutable aka. source of truth.
```
ryle-goehausen, line 31: I think there's potentially some ambiguity in this statement. Snapshots are the best way to organize large volumes of data for the long-term, but are not ideal as the only 'source of truth' as it introduces too many potential sources of delays. Any global value supporting machines on different configurations will hit the same problems as many other single-hash bound chains (with large networks and uncertain computing power.) The main value in a trust based partitioning structure / parallelism is attempting to offer finality in advance of a single hash merger (i.e. a Merkle root of the corresponding snapshot block.) This statement is unconstrained as well (does it require constraints on the snapshot node minimum sizes / latency?) Also -- does that mean that blocks formed in advance of the snapshot can never be considered 'finalized'? It implies that we're degenerate down to a single chain solution with the intermediary structure being essentially an ephemeral solution -- which is not ideal. 

Ideally, finality protections should be designed around preventing excessive forks in the event of legitimate confusion on the network (i.e. dropped packets, latency, other determinism issues.) The cut-off point for 'finality' should be based around nodes attempting to follow the rules as opposed to the data hierarchy. Most conflicts are immediately resolvable through some level of determinism based on trust or chain data or some trivial resolution criteria (with regards to two double spend transactions that are relatively close in time -- it's irrelevant which is actually accepted.) The main purpose of finality here is to prevent the situation of a double spend happening at distinctly different points in time (one much later than another.) Relying purely on the global snapshot phenomena to stop the second case imposes arbitrary constraints on the overall network architecture -- and so finality should be ensured well in advance of that.

Ideally, a snapshot is really only calculating a global view onto data that already exists across the network -- well after consensus has been achieved. The main benefits of it are for calculating diffs more efficiently to provide a foundational data structure.

buckysballs: Ideally each node should have the same contents in its 'view' or snapshot proposal, but that's really more of an optimization. We still need to be able to handle forks in a global hierarchy to perform the most important task of the protocol which is preventing double spends. At the end of the day, we want the network to fork and potentially stall even, if an invalid/double spend block is committed to a local chain; network partitions and ddos should be optimized using reputation and health checks to prevent massive O(n) redownloads.

To your point, the goal of snapshots are to form a linear data structure (out pf path connected blocks) concurrently, while L1 consensus commits data in parallel, forming the dag. Note also that redownloads only send a diff of missing data, which in general if properly optimized should be empty; so L0 is reduces to a concurrent subprocess.

The main optimizations that achieve the optimizations about finality are:

- What we were talking about yesterday using the rep model to identify conflicts/tips connected to double spend are the optimizations which should achieve finality locally without any holdups at the L0.
- A deterministic partitioning structure for L1 facilitator selection optimizes against diffs in the address/balance ledger and blocks/tips/double spends by federating addresses to facilitators.
- The linear address structure prevents against double spends at larger time gaps
- Reducing the dag dimensions (depth, width height) and height diff for snapshot creation reduces finallity overall

doc/adr/0002-global-and-state-channel-snapshots.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Consequences
+
+- Layer 1 needs a redownload algorithm when misaligned with the Global snapshot
```
buckysballs, line 35: Layer 1 will want to download relevant data (most of snapshot) from L0 always, redownloading if active tips are in conflict

doc/adr/0002-global-and-state-channel-snapshots.md
```diff
@@ -0,0 +1,37 @@
...
+- Majority Global snapshots are the source of truth
```
buckysballs, line 36: 👍

doc/adr/0002-global-and-state-channel-snapshots.md
```diff
@@ -0,0 +1,37 @@
...
+- Each StateChannel needs to register its own Layer 0 Cell.
```
buckysballs, line 37: Not necessarily. Running it's own L0 and forming/sending state channel snapshots is the optimal architecture. But state channels can pay to just commit their blocks to global L0, we have a miro board on this from a session a while back

doc/adr/0003-dag-l1-responsibility-and-l0-interconnection.md
```diff
@@ -0,0 +1,30 @@
...
+specified.
+
+## Decision
+
+DAG Layer 1 returns DAG blocks as an output and it doesn't create snapshots.
+DAG Layer 1 sends DAG blocks to the Layer 0 so then the Layer 0 can create a Global
+snapshot.
+
+Cross-chain swaps will take place on the Layer 0, so then Layer 0 can create
+additional DAG transactions and blocks.
```
buckysballs, line 21: 👍

buckysballs, line 21: The actual data types for the swap, called Emission currently, should be enough to resister transfer across addresses/liquidity pools. However, this will happen at the L0 level

doc/adr/0003-dag-l1-responsibility-and-l0-interconnection.md
```diff
@@ -0,0 +1,30 @@
...
+## Context
+
+DAG L1 responsibility and communication between DAG L1 and L0 needs to be
+specified.
+
+## Decision
+
+DAG Layer 1 returns DAG blocks as an output and it doesn't create snapshots.
+DAG Layer 1 sends DAG blocks to the Layer 0 so then the Layer 0 can create a Global
+snapshot.
```
buckysballs, line 18: 👍

doc/adr/0003-dag-l1-responsibility-and-l0-interconnection.md
```diff
@@ -0,0 +1,30 @@
...
+Date: 2022-01-05
+
+## Status
+
+Proposed
+
+## Context
+
+DAG L1 responsibility and communication between DAG L1 and L0 needs to be
+specified.
```
buckysballs, line 12: L0 sends relevent Snapshot data as a metamorphism/stream (using same algebra/coalgebra in reverse [anna cata)])
L1 sends blocks as a hylomorphism (using algebra/coalgebra as specified [cata/anna])

doc/adr/0003-dag-l1-responsibility-and-l0-interconnection.md
```diff
@@ -0,0 +1,30 @@
...
+DAG Layer 1 sends DAG blocks to the Layer 0 so then the Layer 0 can create a Global
+snapshot.
+
+Cross-chain swaps will take place on the Layer 0, so then Layer 0 can create
+additional DAG transactions and blocks.
+
+DAG Layer 1 will subscribe to Layer 0, to pull Global snapshots. If there is a
+difference between accepted DAG blocks and DAG blocks from Global snapshot, DAG
+Layer 1 will apply missing blocks. In the case of conflicting transactions,
+redownload algorithm will be triggered, to reapply data and converge with the Global snapshot.
```
buckysballs, line 26: 👍

doc/adr/0003-dag-l1-responsibility-and-l0-interconnection.md
marcinwadon: What we mean is that DAG transactions and blocks will not only be created on the L1 but also on L0.
And due to no direct communication (push) from L0 to L1, both these things imply that when L1 will download a new global snapshot, there can be blocks/transactions new (unknown) for L1 so the L1 needs to reapply them (or simply apply the whole global snapshot and skip already accepted data)

doc/adr/0002-global-and-state-channel-snapshots.md
marcinwadon: Ok, can we recall that on Monday's discussion mtg? I don't remember how it should work.

doc/adr/0004-global-snapshot-trigger.md
buckysballs: We should use the same two triggers we have in Constellation

- Height diff
- Fallback to timed based trigger

doc/adr/0004-global-snapshot-trigger.md
```diff
@@ -0,0 +1,23 @@
...
+different nodes should converge (contains the same data) and pass the majority algorithm.
```
buckysballs, line 12: See convo above, but tuning height diff and dag dimensions will help optimize redownload size if needed, in lieu of tuning the reputation model and optimizing facilitator/tip selection

doc/adr/0004-global-snapshot-trigger.md
```diff
@@ -0,0 +1,23 @@
...
+
+As described in [ADR#2](./0002-global-and-state-channel-snapshots.md), Global
+snapshot contains both DAG blocks and StateChannel snapshots.
```
buckysballs, line 15: In general this should be the case, but we can allow state channels to commit just blocks if they pay for the space in the L0 dag and L0 to validate their blocks

doc/adr/0002-global-and-state-channel-snapshots.md
buckysballs: But to be clear, we should provide examples on the correct architecture first: which is to create state channel snapshots and commit those to the L0. It's gonna take a lot more work on the rewards and on validation to make that scale

doc/adr/0002-global-and-state-channel-snapshots.md
```diff
@@ -0,0 +1,37 @@
...
+responsibilities.
+
+## Decision
+
+### StateChannel snapshots
+
+L0 Cell can be identified as a proxy point between Layer 1 and Layer 0. Each
+state channel will have its own L0 Cell registered in the Layer 0. One of the
+responsibilities of the L0 Cell is to create a StateChannel snapshot. It can be
+performed either by proxying the snapshot received from Layer 1 or by
```
buckysballs, line 21: If you mean that the state channel L1 is a linear blockchain then yes

doc/adr/0002-global-and-state-channel-snapshots.md
```diff
@@ -0,0 +1,37 @@
...
+periodically creating StateChannel snapshots inside the L0 Cell. Overall, the
+StateChannel snapshot needs to be returned as an output of an L0 Cell so then it
+will be aggregated by Layer 0 and proposed to the Global snapshot.
```
buckysballs, line 24: 👍 we need this to work before working on optimizing the "dumb" case where state channels just pay absorbent fees from their liquidity pool to commit blocks

doc/adr/0002-global-and-state-channel-snapshots.md
buckysballs: Absolutely and it's on the bottom left of the big miro board if I remember correctly

doc/adr/0003-dag-l1-responsibility-and-l0-interconnection.md
buckysballs: Sure, yes L0 needs to create a block for the base/timed trigger case. Also Transactions for Rosetta standard rewards etc.

doc/adr/0002-global-and-state-channel-snapshots.md
marcinwadon: Yes, some kind of "majority" from layer 1.
I see that as an algorithm working on the L0 Cell that is receiving all StateChannel snapshots proposals from L1 (from all the state channel L1 peers), checks whether all required peers made the snapshot (checking against "next facilitators" written in snapshot n-1, and checks the majority. Then, L0 Cell return it as an output from the Cell and L0 layer can handle that. 
image
This is what I mean by "proxying" snapshot from L1 to L0. Basically from the L0 point of view, L0 Cell needs to propose a snapshot, no matter how it's obtained.

doc/adr/0003-dag-l1-responsibility-and-l0-interconnection.md
marcinwadon: I'd argue with that. L1 knows the "connection point" to the Layer 0. It can be a load balancer or any single peer (data should be gossiped across the whole layer anyway). While L0 doesn't know the L1 at all. Doesn't know the L1 peers etc. So what L0 can do is to share the stream of data (let's simplify it to an endpoint) so then L1 can subscribe to this and watch for changes or new events/data (pull-based instead of push-based).

marcinwadon: I think that it's wrong to say that L0 sends relevant Snapshot data to L1, because L0 doesn't know the L1 peers.

marcinwadon: and actually it implies that the L0 should work without any L1. L1 is a deliverer of data but if there are no data, L0 should work anyway.

doc/adr/0003-dag-l1-responsibility-and-l0-interconnection.md
marcinwadon: That's fine. L0 nodes can make similar consensus rounds as DAG L1 nodes.
cc: @TheMMaciek

doc/adr/0004-global-snapshot-trigger.md
marcinwadon: The issue is that those data we have on L1 but snapshot must be made on L0. It means, that either snapshot is triggered by receiving data from L1 (see 1.) or by some deterministic algorithm (see 2.).

1) L1 sends blocks to L0 and alongside sends metadata (like tips). L0 watches for these data and trigger a new snapshot on the height diff (based on received tips). There are 2 drowbacks. Firstly, there will be an additional overhead to send metadata over network, secondly L0 will depend on L1 and will not be able to create a snapshot without it.

2) We can agree on some conditions on both L0 and L1, like there are only 10 parents allowed on each height.
On L1, if there is 11th parent selected, node ignores it (or increments its height). On L0, if node received blocks and there are 10 parents at the same height, it means there will be no more blocks on that height and node can trigger a snapshot.

doc/adr/0004-global-snapshot-trigger.md
marcinwadon: it works if all the data are known on each layer (especially tips)

doc/adr/0003-dag-l1-responsibility-and-l0-interconnection.md
buckysballs: Well for the DAG L0, in order to validate blocks, it will need to be able to validate the facilitators/edges for each block. So it will need the information about a subset of L1 that are selected for active consensus. We can reuse the same process cycling L0 nodes from pending to active as L1. But that should be it

doc/adr/0003-dag-l1-responsibility-and-l0-interconnection.md
buckysballs: Well, the L0 cells don't need to run a full L1 consensus to make an "empty" block for a snapshot, in the case there are no new blocks. We can consider a snapshot/block (which has its two tips being the edge of the previous snapshot) as a special case, just for snapshot proposals.

However, nodes runing L0 can run L1 cells as well, this is where the Arrow comes in

doc/adr/0004-global-snapshot-trigger.md
buckysballs: Given the need to update L1 nodes with metadata from snapshots, and considering that it's pretty much as we had it in Constellation, I would implement option 1 before optimizing based on determinism.

Also, option 1 would be able to create a snapshot without any new blocks created with a timed based trigger. in this case, call it an "empty anapshot", we just need to set edge from the previous snapshot as both the tips for the new one

marcinwadon: Can you elaborate more on this part?
> we just need to set edge from the previous snapshot as both the tips for the new one

For now there were no tips in snapshots

doc/adr/0003-dag-l1-responsibility-and-l0-interconnection.md
marcinwadon: I didn't mean an empty block but a normal consensus round, to process additional transactions on L0 (eg. from cross-chain swaps)

doc/adr/0004-global-snapshot-trigger.md
kpudlik: About time-based: How we can then figure out who should be the next facilitator if we do not have metadata from L1?

doc/adr/0003-dag-l1-responsibility-and-l0-interconnection.md
buckysballs: We shouldn't be running L1 consensus via L0. If we need to perform a swap, an emission from L1 should be enough to initiate the transfer

doc/adr/0004-global-snapshot-trigger.md
buckysballs: So a snapshot is basically just a collection of blocks that is path connected to the collection of blocks in the proceeding snapshot. So basically, it's a linear data structure that links clusters of blocks into a topological order. If there are no blocks received and a snapshot needs to be created via timed trigger, then the proposal can be of one single block, with tips coming from the highest order (sorted by height, width depth [use Ordering]); or in the case of an 'empty' snapshot proceeding an 'empty' snapshot, just use the previous block in the 'empty' snapshot as both tips for the edge

doc/adr/0004-global-snapshot-trigger.md
buckysballs: Exactly

doc/adr/0002-global-and-state-channel-snapshots.md
```diff
@@ -25,7 +25,7 @@ will be aggregated by Layer 0 and proposed to the Global snapshot.
 
 ### Global snapshots
 
-Global snapshots are being created on Layer 0. They contains both DAG blocks and
+Global snapshots are being created on Layer 0. They may contain both DAG blocks and
```
marcinwadon, line 28: @kpudlik if they "may contain" it means they can be completely empty? I'd specify that in the ADR

doc/adr/0004-global-snapshot-trigger.md
```diff
@@ -16,8 +16,22 @@ snapshot contains both DAG blocks and StateChannel snapshots.
...
+when tips reached some interval (will be hardcoded). That trigger creates "non-empty" Global Snapshot
+and should increment `height` and set `subHeight` to `0`.
+
+### Periodic trigger
+If there are no new blocks sent from L1 then L0 nodes - periodically try to create "empty" Global Snapshot.
+Each new "empty" Global Snapshot should increment `subHeight` until "non-empty" snapshot is created.
+
+### Ordinal
+Each snapshot has unique ordinal number. Ordinal number is assigned to new Global Snapshot
+by incrementing the value od previous Global Snapshot. It happens always no matter which trigger has been chosen.
```
marcinwadon, line 32: ~od~ of

doc/adr/0002-global-and-state-channel-snapshots.md
kpudlik: Yes, time-based trigger won't put DAG blocks there.

doc/adr/0006-selecting-facilitators.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Context
+
+Facilitators of L1 should be selected by L0 layer.
+
+## Decision
+
+L0 nodes can select next facilitators by looking at the active peer list from L1.
+These facilitators are put to Global Snapshot so L1 can fetch it and run the consensus.
+L0 nodes will validate signatures of previously selected facilitators when blocks are received from l1.
```
kpudlik, line 17: Guys there is one possible edge case @buckysballs, @marcinwadon, @TheMMaciek, @tbekas 

1. We have selected facilitators `a`, `b`,`c`.
2. L0 receives blocks with tips and at some point is ready to create a snapshot.
3. L0 creates a snapshot and selects the next facilitators:  `d`, `e`, `f`.
4. After selecting the next facilitators and creating a snapshot L0 receives a delayed block from facilitator `b` which is obviously invalid according to that logic because that guy is not a facilitator anymore.

buckysballs: Assuming these L1 facilitators? Regardless of whether an L1 facilitator is active or not, if a block doesn't make it into any snapshot proposal and misses the appropriate height window then the block effectively missed its chance and the data (transactions etc.) need to be re-enqueued into the mempool

kpudlik: Ok, I will post it inside ADR

doc/adr/0003-dag-l1-responsibility-and-l0-interconnection.md
```diff
@@ -0,0 +1,30 @@
...
+additional DAG transactions and blocks.
+
+DAG Layer 1 will subscribe to Layer 0, to pull Global snapshots and other data like selected facilitators. If there is a
+difference between accepted DAG blocks and DAG blocks from Global snapshot, DAG
+Layer 1 will apply missing blocks. In the case of conflicting transactions,
+redownload algorithm will be triggered, to reapply data and converge with the Global snapshot.
+
+## Consequences
+
+tbd
```
marcinwadon, line 30: @kpudlik missing

doc/adr/0004-global-snapshot-trigger.md
marcinwadon: @kpudlik still relevant

tbekas (approved):

PR #117, opened 2022-01-10
Transaction and Hashed orderings

TheMMaciek:


modules/shared/src/main/scala/org/tessellation/schema/transaction.scala
```diff
@@ -73,11 +73,21 @@ object transaction {
...
   }
 
+  implicit val transactionOrder: Order[Transaction] = (x: Transaction, y: Transaction) =>
+    implicitly[Order[BigInt]].compare(x.ordinal.coerce, y.ordinal.coerce)
+
+  implicit val transactionOrdering: Ordering[Transaction] = transactionOrder.toOrdering
+
   implicit val signedTransactionOrder: Order[Signed[Transaction]] = (x: Signed[Transaction], y: Signed[Transaction]) =>
-    implicitly[Order[BigInt]].compare(x.value.ordinal.coerce, y.value.ordinal.coerce)
+    implicitly[Order[Transaction]].compare(x.value, y.value)
```
marcinwadon, line 82: hmm, I see there is already a generic order for `Signed[A]`. maybe you could reuse it? 🤔

modules/shared/src/main/scala/org/tessellation/schema/transaction.scala
```diff
@@ -73,11 +73,21 @@ object transaction {
...
 
   implicit val signedTransactionOrdering: Ordering[Signed[Transaction]] = signedTransactionOrder.toOrdering
 
+  implicit val hashedTransactionOrder: Order[Hashed[Transaction]] = (x: Hashed[Transaction], y: Hashed[Transaction]) =>
```
marcinwadon, line 86: the same for `Hashed`. if there is no generic one, you can create one

modules/shared/src/main/scala/org/tessellation/security/Hashed.scala
```diff
@@ -8,3 +10,11 @@ import derevo.derive
 
 @derive(eqv, show)
 case class Hashed[A <: AnyRef](signed: Signed[A], hash: Hash, proofsHash: ProofsHash)
+
+object Hashed {
+  implicit def autoUnwrap[T <: AnyRef](t: Hashed[T]): T = t.signed
```
TheMMaciek, line 15: FYI it uses `autoUnwrap` from `Signed` also to give the `T`. Not sure if it should use `Signed.autoUnwrap` or unwrap `Signed` here on its own.

marcinwadon: imo it's fine

marcinwadon (approved):

PR #127, opened 2022-01-13
DAG state channel http routes and block rumor handler

TheMMaciek:


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/http/Routes.scala
```diff
@@ -0,0 +1,47 @@
...
+          case Right(hash) => Ok(hash.asJson)
+        }
+      } yield response
+  }
+
+  private val p2p: HttpRoutes[F] = HttpRoutes.of[F] {
+    case req @ POST -> Root / "consensus" / "data" =>
+      for {
+        peerBlockConsensusInput <- req.as[Signed[PeerBlockConsensusInput]]
+        _ <- Async[F].start(peerBlockConsensusInputQueue.offer(peerBlockConsensusInput))
```
marcinwadon, line 39: probably it should be a `Spawn[F]` instead of `Async[F]`

marcinwadon (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/block/rumor/handler.scala
```diff
@@ -0,0 +1,21 @@
...
+import org.tessellation.security.signature.Signed
+
+object handler {
+
+  def blockRumorHandler[F[_]: Async: KryoSerializer](
+    peerBlockQueue: Queue[F, Signed[DAGBlock]]
+  ): RumorHandler[F] =
+    RumorHandler.fromReceivedRumorFn[F, Signed[DAGBlock]]() {
+      case ReceivedRumor(_, signedBlock) =>
+        peerBlockQueue.offer(signedBlock)
```
marcinwadon, line 19: I still suggest to make it through a storage/service instead directly on the queue :D

TheMMaciek: If I make it through the storage I won't be able to stop the single stream with a semaphore and perform redownload e.g. That was my idea here.

TheMMaciek: At least not as easily as I intended it to be :)

marcinwadon: why? you shouldn't stop a queue itself but a service that pulls data

marcinwadon (approved):

PR #128, opened 2022-01-13
DAG state channel runtime stream

TheMMaciek:


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -0,0 +1,173 @@
...
+  gossip: Gossip[F],
+  keyPair: KeyPair,
+  myId: PeerId,
+  nodeStorage: NodeStorage[F],
+  peerBlockConsensusInputQueue: Queue[F, Signed[PeerBlockConsensusInput]],
+  peerBlockQueue: Queue[F, Signed[DAGBlock]],
+  transactionStorage: TransactionStorage[F]
+) {
+
+  private implicit val logger = Slf4jLogger.getLogger[F]
```
marcinwadon, line 62: why it needs to be implicit?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -0,0 +1,173 @@
...
+  private def enoughPeersForConsensus: F[Boolean] =
+    clusterStorage.getPeers
+      .map(_.filter(p => isReadyForBlockConsensus(p.state)))
+      .map(_.size >= consensusConfig.peersCount)
+
+  private val ownerBlockConsensusInputs: Stream[F, OwnerBlockConsensusInput] = Stream
+    .awakeEvery(5.seconds)
+    .evalFilter(_ => consensusStorage.ownConsensus.get.map(_.isEmpty))
+    .evalFilter(_ => nodeStorage.getNodeState.map(isReadyForBlockConsensus))
+    .evalFilter(_ => enoughPeersForConsensus)
```
marcinwadon, line 86: what about:

```scala
private def canStartOwnConsensus: F[Boolean] =
  for { 
    ownRoundInProgress <- ...
    stateReadyForConsensus <- ...
    enoughPeers <- ...
   } yield !ownRoundInProgress && stateReadyForConsensus && enoughPeers
   
Stream.awakeEvery(5.seconds)
   .evalFilter(_ => canStartOwnConsensus)
   .as(OwnRoundTrigger)
```
?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -0,0 +1,173 @@
...
+  private val peerBlockConsensusInputs: Stream[F, PeerBlockConsensusInput] = Stream
+    .fromQueueUnterminated(peerBlockConsensusInputQueue)
+    .evalFilter { signedPeerInput =>
+      for {
+        hasValidSignature <- signedPeerInput.hasValidSignature
+        isSignedBy = signedPeerInput.isSignedBy(
+          Set(PeerId._Id.get(signedPeerInput.value.senderId))
+        )
+      } yield hasValidSignature && isSignedBy
+    }
```
marcinwadon, line 98: I'd extract it as well `def isPeerInputValid` or something like that.
it will keep "validation logic" clean in one place and those pipelines will be easy to read as well

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -0,0 +1,173 @@
...
+    .map(_.value)
+
+  private val blockConsensusInputs: Stream[F, BlockConsensusInput] =
+    ownerBlockConsensusInputs.merge(peerBlockConsensusInputs)
+
+  private val blockConsensusPipeline: Pipe[F, BlockConsensusInput, FinalBlock] =
```
marcinwadon, line 104: I'd use `blockConsensusPipe` instead of `blockConsensusPipeline`

or even just what it does: `runConsensus`

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -0,0 +1,173 @@
...
+                .eval(logger.debug(s"Block created! Hash=${hashedBlock.hash} ProofsHash=${hashedBlock.proofsHash}"))
+                .as(fb)
+            case NullTerminal => Stream.empty
+            case other =>
+              Stream.eval(logger.warn(s"Unexpected ohm in block consensus occurred: $other")) >>
+                Stream.empty
+          }
+      }
+
+  private val blockGossipPipeline: Pipe[F, FinalBlock, FinalBlock] =
```
marcinwadon, line 128: as above: `gossipBlock` or `spreadBlock`

marcinwadon (changes requested):
additional discussion on slack

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
TheMMaciek: Currently `BlockConsensusCell` uses it, we can think about refactoring `BlockConsensusCell` in another PR.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
TheMMaciek: I guess the multiple `evalFilter` solution may be more performant. If the first condition fails the next ones are not even evaluated. Although that might be micro-optimizing.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -0,0 +1,161 @@
...
+        }
+        .void
+    )
+
+  private val blockConsensus: Stream[F, Unit] =
+    blockConsensusInputs
+      .through(runConsensus)
+      .through(gossipBlock)
+      .merge(peerBlocks)
+      .through(storeBlock)
```
marcinwadon, line 155: I have one thought 🤔 Is there a possibility that the block will be gossiped before being stored?
We've seen issues before with blocks that should be accepted but were missing (delay in algorithm) causing unnecessary resolving

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/Validator.scala
```diff
@@ -0,0 +1,55 @@
+package org.tessellation.dag.l1.domain.consensus.block
+
+import cats.Monad
+import cats.effect.kernel.Async
```
marcinwadon, line 4: without `kernel`

marcinwadon (dismissed):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
TheMMaciek: Discussed on slack. TLDR the block storing is idempotent and whether the block comes from local source or from gossip doesn't matter.

marcinwadon (approved):

PR #132, opened 2022-01-19
Unit tests for transaction validator

TheMMaciek:


modules/shared/src/main/scala/org/tessellation/schema/transaction.scala
```diff
@@ -37,6 +39,13 @@ object transaction {
   @derive(decoder, encoder, eqv, show)
   case class TransactionReference(hash: Hash, ordinal: TransactionOrdinal)
 
+  object TransactionReference {
+    val empty: TransactionReference = TransactionReference(Hash(""), TransactionOrdinal(BigInt(0L)))
+
+    val _Hash: Lens[TransactionReference, Hash] = GenLens[TransactionReference](_.hash)
```
kpudlik, line 45: Is it some kind of pattern to name lenses with `_`?

marcinwadon: It's not set on stone but I saw it somewhere 🤔
Lenses: https://github.com/gvolpe/fsm-streams/blob/main/src/main/scala/com/gvolpe/fsmstreams/analytics/Agg.scala
Prisms: https://github.com/gvolpe/fsm-streams/blob/main/src/main/scala/com/gvolpe/fsmstreams/game/Event.scala

kpudlik (approved):

PR #133, opened 2022-01-21
Fiber definition for DAG Transaction and Block

marcinwadon:


modules/shared/src/main/scala/org/tessellation/schema/transaction.scala
```diff
@@ -41,6 +41,14 @@ object transaction {
...
 
+  @derive(decoder, encoder, eqv, show)
+  case class TransactionData(
+    source: Address,
+    destination: Address,
+    amount: TransactionAmount,
+    fee: TransactionFee
+  )
+
   // TODO: figure out the Fiber usage for transaction
```
TheMMaciek, line 52: If it's figured out I guess we can remove the TODO

TheMMaciek: At the same time it's the definition, not usage here 🤔

marcinwadon: I think we can surely remove this TODO

modules/shared/src/main/scala/org/tessellation/schema/Fiber.scala
```diff
@@ -1,6 +1,6 @@
 package org.tessellation.schema
 
 trait Fiber[A, B] {
-  def parent: A
-  def reference: B
+  def reference: A
+  def data: B
```
TheMMaciek, line 5: Originally in my understanding the Fiber was supposed to be an abstraction allowing for the objects of the same type that implement `Fiber` to be interconnected. Hence we had `parent` referencing previous tx/blocks/snapshot/whatever, `reference` being something that this object can be referenced by (so `hash` for `block`, `hash + ordinal` for `transaction`) which is used as `parent` for child objects. Now the `data` itself isn't something that can be referenced directly. Is that okay and as intended? I guess if we had a function calculating `parent` from `data` implemented for every `Fiber` that would make it work, though because of the hashing it wouldn't be pure and would need to be evaluated for every object being determined as parent. I'm not sure though what is the intended use of `Fiber`, so just posting my thoughts and how I understood it some time ago. Maybe you could share your reflections on my thoughts @buckysballs?

modules/shared/src/main/scala/org/tessellation/schema/transaction.scala
```diff
@@ -41,7 +41,14 @@ object transaction {
...
   case class TransactionSalt(value: Long)
 
-  // TODO: figure out the Fiber usage for transaction
+  @derive(decoder, encoder, eqv, show)
+  case class TransactionData(
+    source: Address,
+    destination: Address,
+    amount: TransactionAmount,
+    fee: TransactionFee
+  )
```
TheMMaciek, line 50: Depending on what a `Fiber` should be, possibly `Salt` and `TransactionReference` should also be part of the data. They influence the transaction hash.

TheMMaciek: Though I don't have the full understanding of `Fiber` so just posting to get it discussed.

buckysballs: For example: Transaction would keep the TransactionReference as its first polymorphic type

modules/shared/src/main/scala/org/tessellation/schema/Fiber.scala
buckysballs: Yep it's indeed the case that Fiber is meant to provide covariance for linked structures (single like Transaction, multiple/graph like block). This will help with validation and ledger updates for custom data types for state channels. We should be able to access the polymorphic attributes as long as they are specified. But this will allow for the acceptance and update pipelines to be generic, which we will want for when we expand to new types of Cells

buckysballs (approved):

PR #141, opened 2022-01-28
Healthcheck - trait with trigger capability

marcinwadon:


modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/services/HealthCheckConsensus.scala
```diff
@@ -0,0 +1,41 @@
...
+abstract class HealthCheckConsensus[F[_]]()(implicit F: Monad[F]) extends HealthCheck[F] {
+  def allRounds: Ref[F, ConsensusRounds[F]]
+  def roundsInProgress = allRounds.get.map(_.inProgress)
+
+  final override def trigger(): F[Unit] =
+    triggerRound()
+
+  def triggerRound(): F[Unit] =
+    roundsInProgress.flatMap {
+      case inProgress if inProgress.nonEmpty => manageRounds(inProgress)
```
tbekas, line 21: Is this pattern match actually needed? I prefer using `if-else` for the simplest cases.
```
roundsInProgress.flatMap { inProgress =>
if (inProgress.nonEmpty)
  manageRounds(inProgress)
else
  Applicative[F].unit
}
```

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/services/HealthCheckConsensus.scala
```diff
@@ -0,0 +1,41 @@
...
+import cats.Monad
+import cats.effect.Ref
+import cats.syntax.bifunctor._
+import cats.syntax.flatMap._
+import cats.syntax.functor._
+import cats.syntax.traverse._
+
+import org.tessellation.sdk.domain.healthcheck.types.ConsensusRounds
+
+abstract class HealthCheckConsensus[F[_]]()(implicit F: Monad[F]) extends HealthCheck[F] {
```
tbekas, line 12: I know that it's sometimes useful to have direct access to the type class instance, however in the rest of the code we pass those instances through type constraints, so for the consistency of the codebase I would prefer using `HealthCheckConsensus[F[_]: Monad]`

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/types/ConsensusRounds.scala
```diff
@@ -0,0 +1,15 @@
...
+import org.tessellation.sdk.domain.healthcheck.types.ConsensusRounds.InProgress
+
+final case class ConsensusRounds[F[_]](
+  historical: List[HistoricalRound],
+  inProgress: InProgress[F]
+)
+
+object ConsensusRounds {
+  private type Rounds[F[_]] = Map[HealthCheckKey, HealthCheckRound[F]]
+  type InProgress[F[_]] = Rounds[F]
```
tbekas, line 13: Shouldn't that be a `sealed trait` with two classes for `InProgress` and `Finished` instead of the two type aliases?

tbekas (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/services/HealthCheckConsensus.scala
marcinwadon: Okay, I'll do as you want. IMO both syntaxes are okay and can be used interchangeably

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/services/HealthCheckConsensus.scala
marcinwadon: it's just a matter of clarity but else case or more cases will be added later

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/types/ConsensusRounds.scala
marcinwadon: no, I need both these types separately, e.g.:
```scala
(rounds: ConsensusRounds.InProgress[F])
...
F[(ConsensusRounds.Finished[F], ConsensusRounds.InProgress[F])]
```

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/HealthCheckRound.scala
```diff
@@ -0,0 +1,5 @@
+package org.tessellation.sdk.domain.healthcheck
+
+class HealthCheckRound[F[_]] {
```
TheMMaciek, line 3: Not sure it can be a `class` if there will be different kinds of healthchecks for different `healthcheckKeys` 🤔

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/services/HealthCheckConsensus.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  def triggerRound(): F[Unit] =
+    roundsInProgress.flatMap { inProgress =>
+      if (inProgress.nonEmpty) manageRounds(inProgress) else Applicative[F].unit
+    }
+
+  def manageRounds(rounds: ConsensusRounds.InProgress[F]): F[Unit] =
+    partition(rounds).flatMap {
+      case (finished, inProgress) => Applicative[F].unit
+    }
```
kpudlik, line 27: `partition(rounds).void` unless you will do something with `finished` and `inProgress` here

marcinwadon: I will :) I said it's a stub/dummy code for now

TheMMaciek (approved):

buckysballs (approved):

tbekas (approved):
LGTM

PR #142, opened 2022-01-31
Healthcheck - manage rounds logic

marcinwadon:


modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/types/HealthCheckConsensusDecision.scala
```diff
@@ -0,0 +1,7 @@
+package org.tessellation.sdk.domain.healthcheck.types
+
+sealed trait HealthCheckConsensusDecision
+
+final case class PositiveOutcome() extends HealthCheckConsensusDecision
```
tbekas, line 5: If there are no constructor arguments, let's use `case object` instead. If some arguments will occur in the future, then we will change it to `case class`.

tbekas (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/services/HealthCheckConsensus.scala
```diff
@@ -1,32 +1,65 @@
...
     triggerRound()
 
-  def triggerRound(): F[Unit] =
+  private def triggerRound(): F[Unit] =
     roundsInProgress.flatMap { inProgress =>
       if (inProgress.nonEmpty) manageRounds(inProgress) else Applicative[F].unit
     }
 
-  def manageRounds(rounds: ConsensusRounds.InProgress[F]): F[Unit] =
+  private def manageRounds(rounds: ConsensusRounds.InProgress[F]): F[Unit] = {
```
TheMMaciek, line 29: Don't you want to import `InProgress` and `Finished` directly?

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/services/HealthCheckConsensus.scala
```diff
@@ -1,32 +1,65 @@
...
+      case (finished, inProgress) => checkRounds(inProgress).map(r => (finished ++ r, inProgress -- r.keySet))
+    }.flatMap {
+      case (ready, toManage) =>
+        toManage.values.toList
+          .map(_.manage)
+          .map(Spawn[F].start)
+          .sequence
+          .as(ready)
+    }.flatMap { calculateOutcome }.map {
+      _.filter { case (_, (decision, _)) => decision.isInstanceOf[PositiveOutcome] }
```
TheMMaciek, line 44: I guess the logic for `NegativeOutcome` will come in the next PRs?

TheMMaciek: I wonder which one is neater:
`_.filter { case (_, (decision, _)) => decision.isInstanceOf[PositiveOutcome] }`
vs
`_.collect { case outcome @ (_, (PositiveOutcome(), _)) => outcome }`

TheMMaciek (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/services/HealthCheckConsensus.scala
marcinwadon: No, I think it looks cleaner that way

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/services/HealthCheckConsensus.scala
marcinwadon: PositiveOutcome will have many arguments, so IMO the first one anyway

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/types/HealthCheckConsensusDecision.scala
marcinwadon: There will be many arguments, to be added in next PRs

marcinwadon: I know about the `case object` but in that case I'd avoid switching it back and forth just for one PR. I want to model the healthcheck first starting from the most generic/abstract things and then dive deeper and deeper.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/services/HealthCheckConsensus.scala
TheMMaciek: It can be like this as well:
`_.collect { case outcome @ (_, (_: PositiveOutcome), _)) => outcome }`
but whatever you prefer is ok.

buckysballs (approved):

PR #144, opened 2022-01-31
Fix checking parent acceptance in DAG L1

TheMMaciek:
Currently going through NonEmptyMap resulted in unwanted deduplication because of the way Order is defined for BlockReference - to be revised and possibly fixed in another PR.

### DIFF
--- a/modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockStorage.scala
+++ b/modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockStorage.scala
@@ -31,7 +31,7 @@ class BlockStorage[F[_]: Sync](blocks: MapRef[F, ProofsHash, Option[StoredBlock]
   def areParentsAccepted(block: DAGBlock): F[Map[BlockReference, Boolean]] =
     block.parent.traverse { ref =>
       isBlockAccepted(ref).map(ref -> _)
-    }.map(_.toNem.toSortedMap)
+    }.map(_.toList.toMap)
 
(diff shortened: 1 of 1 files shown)


marcinwadon (approved):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockStorage.scala
```diff
@@ -31,7 +31,7 @@ class BlockStorage[F[_]: Sync](blocks: MapRef[F, ProofsHash, Option[StoredBlock]
   def areParentsAccepted(block: DAGBlock): F[Map[BlockReference, Boolean]] =
     block.parent.traverse { ref =>
       isBlockAccepted(ref).map(ref -> _)
-    }.map(_.toNem.toSortedMap)
+    }.map(_.toList.toMap)
```
marcinwadon, line 34: acceptance must be in order, so you need to sort the list at least

marcinwadon (changes requested):
but acceptance must be sorted

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockStorage.scala
marcinwadon: doesn't it? or it's for something else?

TheMMaciek: It's checking if parents are accepted already - it's used to determine if the block that uses these parents is valid and can be accepted as well.

marcinwadon (approved):

PR #145, opened 2022-01-31
Constructors for DAG L1 block and transaction services

TheMMaciek:


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
```diff
@@ -94,6 +94,15 @@ class BlockService[F[_]: Async: KryoSerializer: SecurityProvider](
 }
 
 object BlockService {
+
+  def make[F[_]: Async: KryoSerializer: SecurityProvider](
```
marcinwadon, line 98: if so, BlockService should be sealed abstract class probably, and at least private

marcinwadon (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
TheMMaciek: Actually I can't make them private because I won't be able to use them anywhere. I can just use the instantiation with `new` if we consider defining `make` for classes a bad practice - in which case I will just close this PR. I did it to have similar interface for all the components. But I don't think extracting interface for these services makes sense to just have the `make` function.

marcinwadon: Yes, you can. See my other comment:

> ```scala
> sealed abstract class TransactionValidator[F[_]: Async: KryoSerializer: SecurityProvider] private (...) {
>   ...
> }
> ```

marcinwadon (changes requested):

marcinwadon (approved):

PR #146, opened 2022-01-31
Make transaction validator an abstract class

TheMMaciek:
Transaction validator will not be mixed in anywhere I think and working with abstract class with type parameters instead of trait is easier.

### DIFF
--- a/modules/dag-shared/src/main/scala/org/tessellation/dag/transaction/TransactionValidator.scala
+++ b/modules/dag-shared/src/main/scala/org/tessellation/dag/transaction/TransactionValidator.scala
@@ -25,12 +25,8 @@ import org.tessellation.security.signature.signature.SignatureProof
 import eu.timepit.refined.auto._
 import io.estatico.newtype.ops._
 
-trait TransactionValidator[F[_]] {
+abstract class TransactionValidator[F[_]: Async: KryoSerializer: SecurityProvider] {
   import TransactionValidator._
-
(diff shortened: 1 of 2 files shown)


modules/dag-shared/src/main/scala/org/tessellation/dag/transaction/TransactionValidator.scala
```diff
@@ -25,13 +25,9 @@ import org.tessellation.security.signature.signature.SignatureProof
 import eu.timepit.refined.auto._
 import io.estatico.newtype.ops._
 
-trait TransactionValidator[F[_]] {
+abstract class TransactionValidator[F[_]: Async: KryoSerializer: SecurityProvider] {
```
marcinwadon, line 28: you can do:

```scala
sealed abstract class TransactionValidator[F[_]: Async: KryoSerializer: SecurityProvider] private (...) {
  ...
}
```

marcinwadon (changes requested):

modules/dag-shared/src/main/scala/org/tessellation/dag/transaction/TransactionValidator.scala
TheMMaciek: Not sure if it makes sense to make the constructor private. I need to be able to implement it in different places (DAG L1, L0) based on the storages created there.

marcinwadon (approved):

PR #148, opened 2022-01-31
Require single signer in block validation during DAG L1 consensus

TheMMaciek:


marcinwadon (approved):

marcinwadon:
Why one signer only? I think for DAG we require at least 3

PR #149, opened 2022-02-01
Cluster leave and peer removal capabilities

marcinwadon:


modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/storage/ClusterStorage.scala
```diff
@@ -13,4 +13,5 @@ trait ClusterStorage[F[_]] {
   def hasPeerId(id: PeerId): F[Boolean]
   def hasPeerHostPort(host: Host, p2pPort: Port): F[Boolean]
   def setPeerState(id: PeerId, state: NodeState): F[Unit]
+  def removePeer(id: PeerId): F[Unit]
```
TheMMaciek, line 16: I would move peer removal to a separate PR, as leaving cluster and peer removal are not related to each other. Or at least I would remark adding peer removal functionality in the commit message.

TheMMaciek (changes requested):
Apart from peer removal being merged along with cluster leaving it LGTM.

buckysballs (approved):

PR #151, opened 2022-02-02
Time-based snapshot trigger

kpudlik:


modules/core/src/main/scala/org/tessellation/l0/domain/snapshot/GlobalSnapshot.scala
```diff
@@ -8,12 +8,15 @@ import org.tessellation.schema.height.Height
 import org.tessellation.schema.peer.PeerId
 import org.tessellation.security.signature.Signed
 
-import io.estatico.newtype.ops._
+import eu.timepit.refined.types.numeric.NonNegLong
+
+case class GlobalSnapshotOrdinal(value: NonNegLong)
```
marcinwadon, line 13: could it be a newtype?

modules/core/src/main/scala/org/tessellation/l0/domain/snapshot/SnapshotService.scala
```diff
@@ -1,16 +1,10 @@
...
 trait SnapshotService[F[_]] {
 
-  def createSnapshot(
-    blocks: NonEmptySet[Hashed[DAGBlock]],
-    snapshots: Set[StateChannelSnapshot],
-    nextFacilitators: NonEmptySet[PeerId]
-  ): F[GlobalSnapshot]
+  def offer(block: Signed[DAGBlock]): F[Unit]
+  def createEmptySnapshot(): F[GlobalSnapshot]
+  def createNormalSnapshot(blocks: Set[Signed[DAGBlock]]): F[GlobalSnapshot]
```
marcinwadon, line 9: if that's possible I'd stop naming it `normal` snapshot. Just a snapshot:
```scala
def createEmptySnapshot(): ...
def createSnapshot(): ...
```
or maybe even better:
```scala
def createSnapshot(): F[GlobalSnapshot]
def createSnapshot(blocks: ...): F[GlobalSnapshot]
```

modules/core/src/main/scala/org/tessellation/l0/domain/snapshot/SnapshotStorage.scala
```diff
@@ -1,7 +1,11 @@
 package org.tessellation.l0.domain.snapshot
+import org.tessellation.kernel.StateChannelSnapshot
 import org.tessellation.schema.height.Height
 
 trait SnapshotStorage[F[_]] {
   def storeSnapshot(snapshot: GlobalSnapshot): F[Unit]
   def getLastSnapshotHeight: F[Height]
+  def getLastSnapshotSubHeight: F[Height]
+  def getLastSnapshotOrdinal: F[GlobalSnapshotOrdinal]
+  def getStateChannelSnapshots: F[Set[StateChannelSnapshot]]
```
marcinwadon, line 10: shouldn't it be a `Map[StateChannelAddress, Set[StateChannelSnapshot]]` or `def getStateChannelSnapshots(address: StateChannelAddress): F[Set[StateChannelSnapshot]]` or sth like that?

modules/core/src/main/scala/org/tessellation/l0/http/Routes.scala
```diff
@@ -0,0 +1,29 @@
...
+
+import org.tessellation.dag.domain.block.DAGBlock
+import org.tessellation.l0.domain.snapshot.SnapshotService
+import org.tessellation.security.signature.Signed
+
+import org.http4s.HttpRoutes
+import org.http4s.circe.CirceEntityCodec.circeEntityDecoder
+import org.http4s.dsl.Http4sDsl
+
+final case class Routes[F[_]: Async](
```
marcinwadon, line 15: please name it `SnapshotRoutes` or sth (same as the others that @TheMMaciek moved to sdk)

modules/core/src/main/scala/org/tessellation/l0/http/Routes.scala
```diff
@@ -0,0 +1,29 @@
...
+  snapshotService: SnapshotService[F]
+) extends Http4sDsl[F] {
+
+  private val p2p: HttpRoutes[F] = HttpRoutes.of[F] {
+    case req @ POST -> Root / "block" =>
+      for {
+        block <- req.as[Signed[DAGBlock]]
+        response <- snapshotService.offer(block)
+        res <- Ok()
+      } yield res
```
marcinwadon, line 26: IMO more clear is
```scala
req.as[Signed[DAGBlock]] >>
  snapshotService.offer >>
  Ok()
```
or 
```scala
req.as[Signed[DAGBlock]]
  .flatMap(snapshotService.offer)
  .flatMap(_ => Ok())
```

It's easy to do sth like that then:
```scala
req.as[Signed[DAGBlock]]
  .flatMap(snapshotService.offer)
  .flatMap(_ => Ok())
  .recoverWith {
    case MyError => Conflict(...)
    case MyOtherError => NotFound(...)
  }
```

modules/core/src/main/scala/org/tessellation/l0/infrastructure/snapshot/SnapshotService.scala
```diff
@@ -2,46 +2,82 @@ package org.tessellation.l0.infrastructure.snapshot
 
 import cats.data.NonEmptySet
 import cats.effect.kernel.Async
```
marcinwadon, line 4: if you see an import from `kernel` please remove that, to have:
```
import cats.effect.Async
```

modules/core/src/main/scala/org/tessellation/l0/infrastructure/snapshot/SnapshotService.scala
```diff
@@ -2,46 +2,82 @@ package org.tessellation.l0.infrastructure.snapshot
...
-        blocks: NonEmptySet[Hashed[DAGBlock]],
-        snapshots: Set[StateChannelSnapshot],
-        nextFacilitators: NonEmptySet[PeerId]
-      ): F[GlobalSnapshot] =
+      def offer(signedBlock: Signed[DAGBlock]): F[Unit] =
         for {
-          previousSnapshotHeight <- snapshotStorage.getLastSnapshotHeight
-          nextSnapshotHeight = Height(previousSnapshotHeight.coerce + heightInterval)
+          _ <- blocksQueue.offer(Set(signedBlock))
+        } yield ()
```
marcinwadon, line 36: what's for needed?
```scala
def offer(signedBlock: Signed[DAGBlock]): F[Unit] =
  blocksQueue.offer(Set(signedBlock)) // or with .void at the end
```

modules/core/src/main/scala/org/tessellation/l0/infrastructure/snapshot/SnapshotService.scala
```diff
@@ -2,46 +2,82 @@ package org.tessellation.l0.infrastructure.snapshot
...
+
+      private val emptySnapshot = Stream.eval(createEmptySnapshot())
+
+      private val normalSnapshot = Stream.fromQueueUnterminated(blocksQueue).evalMap(createNormalSnapshot)
+
+      val snapshotCreation: Stream[F, GlobalSnapshot] =
+        normalSnapshot.through(switchRepeat(10.seconds, emptySnapshot))
```
marcinwadon, line 43: please extract delays into one place (config preferable)

modules/core/src/main/scala/org/tessellation/l0/infrastructure/snapshot/SnapshotService.scala
```diff
@@ -2,46 +2,82 @@ package org.tessellation.l0.infrastructure.snapshot
...
+          snapshots <- snapshotStorage.getStateChannelSnapshots
+        } yield
+          GlobalSnapshot(
+            ordinal = GlobalSnapshotOrdinal(prevOrdinal.value |+| NonNegLong(1L)),
+            height = Height(prevHeight.coerce + 1L),
+            subHeight = Height(0L),
+            blocks = blocks,
+            snapshots = snapshots,
+            nextFacilitators = nextFacilitators
+          )
```
marcinwadon, line 78: what about unifying that into a common method?

```scala

def createEmptySnapshot(): F[GlobalSnapshot] =
  (snapshotStorage.getLastSnapshotHeight, snapshotStorage.getLastSnapshotSubHeight).mapN { (height, subheight) =>
    createSnapshot(Set.empty[Signed[DAGBlock]], height, Height(subheight.coerce[Long] + 1L)
    
def createNormalSnapshot(blocks: Set[Signed[DAGBlock]]): F[GlobalSnapshot] =
  snapshotStorage.getLastSnapshotHeight.flatMap { height =>
    createSnapshot(blocks, Height(height.coerce + 1L), Height(0L))

private def createSnapshot(blocks: Set[Signed[DAGBlock]], height: Height, subheight: Height): F[GlobalSnapshot] =
        for {
            prevOrdinal <- snapshotStorage.getLastSnapshotOrdinal
            nextFacilitators <- selectNextFacilitators
            snapshots <- snapshotStorage.getStateChannelSnapshots
          } yield
            GlobalSnapshot(
              ordinal = GlobalSnapshotOrdinal(prevOrdinal.value |+| NonNegLong(1L)),
              height = height,
              subHeight = subheight,
              blocks = blocks,
              snapshots = snapshots,
              nextFacilitators = nextFacilitators
            )


```

marcinwadon (changes requested):

modules/core/src/main/scala/org/tessellation/l0/infrastructure/snapshot/SnapshotService.scala
kpudlik: Yep, we will get rid of the naming "normal/empty"

modules/core/src/main/scala/org/tessellation/l0/infrastructure/snapshot/SnapshotService.scala
kpudlik: I can extract that but the question is if It can be configurable by the node operator

modules/core/src/main/scala/org/tessellation/l0/infrastructure/snapshot/SnapshotService.scala
kpudlik: We can use `blocksQueue` directly in Routes. I will get rid of it.

modules/core/src/main/scala/org/tessellation/l0/infrastructure/snapshot/SnapshotService.scala
kpudlik: Auto-import from IntelliJ. I will try to ban auto-imports from kernel in IDE

modules/core/src/main/scala/org/tessellation/l0/http/Routes.scala
kpudlik: That code will be modified by the implementation of tip-based trigger. But I agree :)

modules/core/src/main/scala/org/tessellation/l0/domain/snapshot/SnapshotStorage.scala
kpudlik: I am not sure in which form we should store state-channel snapshots. Maybe you are right.
In general I just need to store statechannel snapshots, then put them to Snapshot and clear the storage by removing these which are in Snapshot already (or maybe in Majority Snapshot)

modules/core/src/main/scala/org/tessellation/l0/domain/snapshot/SnapshotService.scala
kpudlik: Yeah, I agree

modules/core/src/main/scala/org/tessellation/l0/domain/snapshot/GlobalSnapshot.scala
kpudlik: Oh yes, I forgot to make it a newtype. I just need to move that to package object

modules/core/src/main/scala/org/tessellation/l0/http/Routes.scala
```diff
@@ -0,0 +1,29 @@
...
+import org.tessellation.dag.domain.block.DAGBlock
+import org.tessellation.l0.domain.snapshot.SnapshotService
+import org.tessellation.security.signature.Signed
+
+import org.http4s.HttpRoutes
+import org.http4s.circe.CirceEntityCodec.circeEntityDecoder
+import org.http4s.dsl.Http4sDsl
+
+final case class Routes[F[_]: Async](
+  l1BlocksQueue: Queue[F, Signed[DAGBlock]],
```
TheMMaciek, line 16: Leftover?

modules/core/src/main/scala/org/tessellation/l0/infrastructure/snapshot/SnapshotService.scala
```diff
@@ -2,46 +2,82 @@ package org.tessellation.l0.infrastructure.snapshot
...
-      def createSnapshot(
-        blocks: NonEmptySet[Hashed[DAGBlock]],
-        snapshots: Set[StateChannelSnapshot],
-        nextFacilitators: NonEmptySet[PeerId]
-      ): F[GlobalSnapshot] =
+      def offer(signedBlock: Signed[DAGBlock]): F[Unit] =
         for {
-          previousSnapshotHeight <- snapshotStorage.getLastSnapshotHeight
-          nextSnapshotHeight = Height(previousSnapshotHeight.coerce + heightInterval)
+          _ <- blocksQueue.offer(Set(signedBlock))
```
TheMMaciek, line 35: Why do we need to offer a `Set[Signed[DAGBlock]]` instead of just `Signed[DAGBlock]`?

modules/core/src/main/scala/org/tessellation/l0/infrastructure/snapshot/SnapshotService.scala
TheMMaciek: We can have a config class or even object with meaningful names that's not configurable. If it's not configurable then maybe let's call it differently, I don't know parameters?

modules/core/src/main/scala/org/tessellation/l0/infrastructure/snapshot/SnapshotService.scala
```diff
@@ -2,46 +2,82 @@ package org.tessellation.l0.infrastructure.snapshot
...
+        } yield ()
+
+      private val emptySnapshot = Stream.eval(createEmptySnapshot())
+
+      private val normalSnapshot = Stream.fromQueueUnterminated(blocksQueue).evalMap(createNormalSnapshot)
```
TheMMaciek, line 40: What about the naming convention: `emptySnapshots` and `normalSnapshots`?

modules/core/src/main/scala/org/tessellation/l0/infrastructure/snapshot/SnapshotService.scala
```diff
@@ -2,46 +2,82 @@ package org.tessellation.l0.infrastructure.snapshot
...
+
+      val snapshotCreation: Stream[F, GlobalSnapshot] =
```
TheMMaciek, line 42: `globalSnapshots`?

modules/core/src/main/scala/org/tessellation/l0/infrastructure/snapshot/SnapshotService.scala
marcinwadon: hmm, I'd use a service anyway for clarity, but both ways would be fine, up to you

modules/core/src/main/scala/org/tessellation/l0/infrastructure/snapshot/SnapshotService.scala
marcinwadon: @TheMMaciek we already have non-configurable config

modules/core/src/main/scala/org/tessellation/l0/infrastructure/snapshot/SnapshotService.scala
kpudlik: We will see how that code mutates during normal trigger implementation. For now I would just use method directly

marcinwadon (approved):
Overall looks good but I think we agreed on using `core` instead of `l0` name for base nodes. 🤔 I mean, we have `DAG` which is common for both layers, `L1` separately and `core` for L0

modules/core/src/main/scala/org/tessellation/l0/infrastructure/snapshot/SnapshotService.scala
kpudlik: It will be changed because previous concept was to send "package" of blocks for Snapshot. But now it will be just a `Block` because we changed the concept with Tomek

kpudlik:
> Overall looks good but I think we agreed on using `core` instead of `l0` name for base nodes. 🤔 I mean, we have `DAG` which is common for both layers, `L1` separately and `core` for L0

I will refactor it in the next PR then.

PR #155, opened 2022-02-03
Healthcheck round inspection and outcome action

marcinwadon:


modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/services/HealthCheckConsensus.scala
```diff
@@ -40,8 +54,15 @@ abstract class HealthCheckConsensus[F[_]: Spawn]() extends HealthCheck[F] {
           .map(Spawn[F].start)
           .sequence
           .as(ready)
-    }.flatMap { calculateOutcome }.map {
-      _.filter { case (_, (decision, _)) => decision.isInstanceOf[PositiveOutcome] }
+    }.flatMap {
+      calculateOutcome
+    }.flatTap { outcome =>
+      def positive = outcome.filter { case (_, (decision, _)) => decision.isInstanceOf[PositiveOutcome] }
+      def negative = outcome -- positive.keySet
```
TheMMaciek, line 61: What bugs me about subtracting like this is that if we add another case and forget to handle it here it will be treated as negative outcome.

TheMMaciek (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/services/HealthCheckConsensus.scala
marcinwadon: I don't think we will extend the base but if so, I can make it the opposite way:
```
def negative = ...filter
def nonNegative = outcome -- negative.keySet
```
what do you think @TheMMaciek?
probably we don't want to make any action if the outcome is other than negative 🤔

TheMMaciek: I guess doing it this way around is safer indeed!

marcinwadon: I'll do that in the next PR

PR #159, opened 2022-02-04
Healthcheck - round start

marcinwadon:


modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/services/HealthCheckConsensus.scala
```diff
@@ -91,37 +93,86 @@ abstract class HealthCheckConsensus[
...
 
   def startRound(
     key: HealthCheckKey,
     peers: Set[Peer],
-    roundId: RoundId,
+    roundId: HealthCheckRoundId,
     roundType: HealthCheckRoundType
-  ): F[Unit] = ???
+  ): F[Unit] =
+    Clock[F].monotonic
```
tbekas, line 112: 1. If you need an actual time of the start of the round then `Clock[F].realTime` should be used instead of `monotonic`
2. It looks like you're not using the `startedAt` paramter at all.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/services/HealthCheckConsensus.scala
```diff
@@ -91,37 +93,86 @@ abstract class HealthCheckConsensus[
...
+      .map(_.toSeconds)
+      .flatMap { startedAt =>
+        allRounds.get.map {
+          case ConsensusRounds(historical, inProgress) =>
+            def inProgressRound = inProgress.get(key)
+            def historicalRound = historical.find(_.roundId == roundId)
+
+            historicalRound.orElse(inProgressRound).map(_ => none).getOrElse {
+              HealthCheckRound.make[F, A, B]().some
```
tbekas, line 121: This will result in `Option[Option[A]]`. Is the nested option necessary? It makes reasoning about the code difficult to follow.

tbekas (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/services/HealthCheckConsensus.scala
marcinwadon: I'm going to pass it to the round and most likely I need monotonic instead of realTime. If so, I'll change that in next PR with actual usage in a particular round

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/services/HealthCheckConsensus.scala
marcinwadon: Yes, I think so.
tl;dr;
It checks if historical or inProgress round exists, and if no - creates a new round declaration.
It returns the Option (to indicate whether new round should be created or no) and the value is `F[Round]`
so it's not `Option[Option[A]]` but `Option[F[A]]`

tbekas: Yep, you're right, the final type of this expression is `F[Option[F[A]]]`.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/services/HealthCheckConsensus.scala
tbekas: Ok, it's difficult for reviewers to tell the difference between what's not used yet and will be used later on and what's not used because the author forgot to remove it.

tbekas (approved):

PR #161, opened 2022-02-07
Time based clean up mechanism for DAG L1 block consensus

TheMMaciek:


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/BlockConsensusCell.scala
```diff
@@ -94,6 +107,8 @@ class BlockConsensusCell[F[_]: Async: SecurityProvider: KryoSerializer: Random:
 
 object BlockConsensusCell {
 
+  private def getTime[F[_]: Clock](): F[FiniteDuration] = Clock[F].monotonic
```
marcinwadon, line 110: if you want to compare and calculate delta, you should do `.map(_.toNanos)`

marcinwadon (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/BlockConsensusCell.scala
marcinwadon: `FiniteDuration` as the name says, it's not a moment in time but some period (duration)

PR #162, opened 2022-02-08
Healthcheck - consensus round logic

marcinwadon:


modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensus.scala
```diff
@@ -92,67 +96,57 @@ abstract class HealthCheckConsensus[
...
-            def historicalRound = historical.find(_.roundId == roundId)
-
-            historicalRound.orElse(inProgressRound).map(_ => none).getOrElse {
-              HealthCheckRound.make[F, A, B]().some
+    allRounds.get.map {
+      case ConsensusRounds(historical, inProgress) =>
+        def inProgressRound = inProgress.get(key)
+        def historicalRound = historical.find(_.roundId == roundId)
+
+        historicalRound.orElse(inProgressRound).map(_ => none).getOrElse {
```
kpudlik, line 115: .map(_ => none)

No matter if you have historicalRound or inProgressRound 🤔 

Then getOrElse always triggers else part.

kpudlik (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensus.scala
marcinwadon: No, it's `map` not `flatMap`. Option is nested

kpudlik: Ah right

marcinwadon: tl;dr;
if there is any historical or in progress round -> return `None`
if there is no round, create one and return `Some(newRound)`

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensusRound.scala
```diff
@@ -0,0 +1,122 @@
...
+    } else Applicative[F].unit
+
+  def getRoundIds: F[Set[HealthCheckRoundId]] = roundIds.get
+
+  def addParallelRounds(key: HealthCheckKey)(roundIds: Set[HealthCheckRoundId]): F[Unit] = ???
+
+  def calculateOutcome: F[HealthCheckConsensusDecision] = {
+    def status: A = ???
+
+    proposals.get.map(_.values.toList).map { received =>
```
TheMMaciek, line 66: `allProposalsReceived` subtracts peers from `proposals` Map from the `peers` Set to say if the consensus is ready, but here it's passing all the received proposals. Doesn't seem correct if the `peers` Set is representing a collection of peers that should be taken into account as final consensus peers/facilitators.

TheMMaciek: I think we need to stride towards deterministic results. Currently it seems that arbitrary proposals may land in the `proposals` Map and they will influence the round outcome.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensusRound.scala
```diff
@@ -0,0 +1,122 @@
...
+      driver.calculateConsensusOutcome(key, status, selfId, received)
+    }
+  }
+
+  def manage: F[Unit] =
+    Clock[F].monotonic.map(_.toNanos).flatMap { currentTime =>
```
TheMMaciek, line 72: Not sure that conversion to nanos is necessary. Depends on how it's going to be used probably. But you can compare `FiniteDuration`s with each other with e.g. `< > ==` and also subtract and add with success I think, so not sure if you need to switch to `Long` here.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensusRound.scala
```diff
@@ -0,0 +1,122 @@
...
+  def ownConsensusHealthStatus: F[B] = {
+    def status: A = ???
+    Applicative[F].pure(driver.consensusHealthStatus(key, status, roundId, selfId))
+  }
+
+  private def sendProposal: F[Unit] =
+    ownConsensusHealthStatus.flatMap(gossip.spread)
+
+  private def allProposalsReceived: F[Boolean] =
+    (peers.get, proposals.get.map(_.keySet))
```
TheMMaciek, line 87: I guess we also need the functionality for adding peers to `peers` Set, based on the incoming proposals. Though I'm not sure what algorithm/flow you intended for the final implementation. Currently we are only subtracting from the initial locally known peers - this may influence deterministicity for the global healthcheck outcome on the whole cluster - we need to stride towards determinsticity on the cluster level.

TheMMaciek (changes requested):
Some questions and possible flaws that I feel we need to discuss.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensusRound.scala
```diff
@@ -0,0 +1,122 @@
...
+    if (proposal.key == key) {
+      proposals.modify { m =>
+        def updated =
+          m.get(proposal.owner)
+            .fold(m + (proposal.owner -> proposal))(_ => m)
+
+        (updated, ())
+      }.flatMap { _ =>
+        roundIds.modify { m =>
+          (m + proposal.roundId, ())
```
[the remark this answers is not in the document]
marcinwadon: there are two cases:
1. proposal already exists
2. proposal does not exist

In case 2 - `m + proposal.roundId` changes nothing, roundId should already be there and it's a `Set` (deduplication)
In case 1 - `m + proposal.roundId` should add the round id

It's the same as you did:
```scala
          roundIds.modify { roundIds => //just adding blindly seems better then checking and conditionally modifying
            (roundIds.add(healthProposal.roundId), ())
          }
```

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensus.scala
```diff
@@ -92,67 +96,57 @@ abstract class HealthCheckConsensus[
...
-        case Some(mkRound) =>
-          mkRound.flatMap { round =>
-            allRounds.modify {
-              case ConsensusRounds(historical, inProgress) =>
-                (ConsensusRounds(historical, inProgress + (key -> round)), round.some)
-            }
+    }.flatMap {
+      case Some(mkRound) =>
+        mkRound.flatMap { round =>
+          allRounds.modify {
```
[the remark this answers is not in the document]
marcinwadon: :+1: I'll try to combine `get` and `modify` together.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensusRound.scala
marcinwadon: @TheMMaciek it substracts `proposals` from `peers`. So if all peers made a proposal, we can finish the round.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensusRound.scala
marcinwadon: You'd want to compare a point in time to check if some delta time has elapsed. 
see: https://github.com/typelevel/fs2/blob/main/core/shared/src/main/scala/fs2/Stream.scala#L3166

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensusRound.scala
marcinwadon: It's not the initial list of peers, but the final peers that are required to make a proposal.
It starts with the initial list of peers but not necessary it is the same at the end.
We may want to add peers to `peers` later but I didn't see that in your code so I didn't want to change it for now

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensusRound.scala
```diff
@@ -0,0 +1,122 @@
...
+) {
+
+  def start: F[Unit] =
+    Spawn[F].start {
+      sendProposal
+    }.void
+
+  def getPeers: F[Set[PeerId]] = peers.get
+
+  def manageAbsent(absentPeers: Set[PeerId]): F[Unit] =
```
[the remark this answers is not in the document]
marcinwadon: I'll take a look, good catch :+1:

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensusRound.scala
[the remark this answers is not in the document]
TheMMaciek: I thinks it's not the same as I did in constellation - I'm only adding the `RoundId` there if the proposal was accepted. Any other case results in `Left` and the `RoundId` is not added. "Blindly adding" being better regarded the fact that if we add the `RoundId` which is there already in the set (because some other node sent its proposal with the same `RoundId` before us) that won't damage anything it will deduplicate. But for this implementation someone could blast us with proposals with different `RoundId`s, proposals which won't be processed/added again, but the round ids for these proposals will get added and it brakes data integrity and may lead to non determinism depending on how will we use the round ids.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensusRound.scala
TheMMaciek: Maybe my response was a bit tangled but that's what I meant. Note that it checks if all peers sent proposals, but then sent's all proposals (potentially also the ones from peers that are not on ours peers list) to the function calculating outcome. I think it can brake data integrity/influence the outcome, essentially be exploited purposefully or by some non determinism somewhere else (node rejoining and sending different proposal again e.g.).

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensusRound.scala
TheMMaciek: Adding is there in constellation - hence idea of proxy peers. And yeah that's exactly what I meant, we start with initial peers and after that we only have functionality for removing peers with the `manageAbsent` arriving at final peers set. I think it may not be valid as I mentioned in the first comment.

TheMMaciek (changes requested):
Some more potential flaws to discuss.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensusRound.scala
```diff
@@ -0,0 +1,134 @@
...
+        m.get(proposal.owner)
+          .fold {
+            (m + (proposal.owner -> proposal), proposal.some)
+          } { _ =>
+            (m, none)
+          }
+      }.flatMap {
+        _.fold { Applicative[F].unit } { proposal =>
+          roundIds.update(_ + proposal.roundId).flatMap { _ =>
+            peers.update(_ + proposal.owner)
```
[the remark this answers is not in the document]
TheMMaciek: We will figure this out in another ticket.

TheMMaciek (dismissed):

TheMMaciek (approved):

kpudlik (approved):

PR #163, opened 2022-02-08
Healthcheck - ping consensus implementation

marcinwadon:


modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensus.scala
```diff
@@ -20,17 +20,20 @@ import org.tessellation.sdk.domain.healthcheck.consensus.types.types.RoundId
...
 ](
   clusterStorage: ClusterStorage[F],
   selfId: PeerId,
-  driver: HealthCheckConsensusDriver[A, B],
+  driver: HealthCheckConsensusDriver[K, A, B],
   gossip: Gossip[F]
 ) extends HealthCheck[F] {
+  def allRounds: Ref[F, ConsensusRounds[F, K, A, B]]
+
+  def ownStatus(key: K): F[Fiber[F, Throwable, A]]
```
TheMMaciek, line 35: Not sure if instead of `Fiber` we shouldn't use `Deferred`? 🤔

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/healthcheck/ping/PingHealthCheckConsensus.scala
```diff
@@ -0,0 +1,49 @@
...
+      selfId,
+      driver,
+      gossip
+    ) {
+
+  def allRounds: Ref[F, ConsensusRounds[F, PingHealthCheckKey, PingHealthCheckStatus, PingConsensusHealthStatus]] =
+    rounds
+
+  def ownStatus(key: PingHealthCheckKey): F[Fiber[F, Throwable, PingHealthCheckStatus]] =
+    Spawn[F].start(Applicative[F].pure(PeerUnavailable(selfId)))
```
TheMMaciek, line 31: Should we merge something like this?

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensusRound.scala
```diff
@@ -68,29 +72,35 @@ class HealthCheckConsensusRound[F[_]: Spawn: Clock, A <: HealthCheckStatus, B <:
...
+  def ownConsensusHealthStatus: F[B] =
+    status.map {
+      driver.consensusHealthStatus(key, _, roundId, selfId)
+    }
+
+  private def status =
+    ownStatus.join.flatMap {
+      case Outcome.Succeeded(fa) => fa
+      case Outcome.Errored(e)    => MonadError[F, Throwable].raiseError[A](e)
+      case Outcome.Canceled()    => MonadError[F, Throwable].raiseError[A](new Throwable("Unexpected error"))
```
TheMMaciek, line 102: Not sure what should be the behavior here, but keep in mind that other nodes will expect to get a status from you, if you fail with an error they will keep on waiting. So maybe it should default to some value.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensus.scala
marcinwadon: `Deferred` will be computed once, but not in the background. `Fiber` is more or less like Deferred that can be computed in the background.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensusRound.scala
marcinwadon: Hm, I like that idea. Will do that :+1:

PR #167, opened 2022-02-11
Tip-based snapshot trigger

kpudlik:


modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/SnapshotService.scala
```diff
@@ -27,48 +29,57 @@ object SnapshotService {
...
     new SnapshotService[F] {
 
-      val snapshotCreation: Stream[F, GlobalSnapshot] =
+      val pipeline: Stream[F, Either[Signed[DAGBlock], SnapshotReference]] =
         Stream
-          .fromQueueUnterminated(blocksQueue)
-          .evalMap(createSnapshot)
-          .through(switchRepeat(config.fallbackTriggerTimeout, Stream.eval(createSnapshot())))
+          .fromQueueUnterminated(l1DataQueue)
+          .zip(Stream.eval(snapshotStorage.getLastSnapshotHeight))
```
marcinwadon, line 41: I think you can expose `lastSnapshotHeight` as a stream, directly from `snapshotStorage`

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/SnapshotService.scala
```diff
@@ -27,48 +29,57 @@ object SnapshotService {
...
+          .evalMap {
+            case (data, height) =>
+              snapshotPreconditionsValidator.validate(height, data) match {
+                case Valid(height) =>
+                  createSnapshotTrigger(height).map(_.asRight[Signed[DAGBlock]])
```
marcinwadon, line 46: shouldn't it send both: `data.block` and then `snapshotTrigger`?

marcinwadon (changes requested):

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/SnapshotService.scala
kpudlik: Fair point!

modules/core/src/main/scala/org/tessellation/domain/snapshot/SnapshotStorage.scala
```diff
@@ -1,13 +1,15 @@
...
-import org.tessellation.l0.schema.snapshot.{GlobalSnapshot, GlobalSnapshotOrdinal}
 import org.tessellation.schema.height.Height
 
+import fs2.Stream
+
 trait SnapshotStorage[F[_]] {
-  def storeSnapshot(snapshot: GlobalSnapshot): F[Unit]
+  def storeSnapshot(snapshot: SnapshotReference): F[Unit]
   def getLastSnapshotHeight: F[Height]
+  def lastSnapshotHeight: Stream[F, Height] = Stream.eval(getLastSnapshotSubHeight)
```
marcinwadon, line 11: it's not a stream of last snapshot heights. it emits once only

marcinwadon (changes requested):

marcinwadon (dismissed):

marcinwadon (dismissed):

modules/core/src/main/scala/org/tessellation/domain/snapshot/SnapshotReference.scala
```diff
@@ -0,0 +1,15 @@
...
+
+import org.tessellation.kernel.StateChannelSnapshot
+import org.tessellation.schema.height.Height
+import org.tessellation.schema.peer.PeerId
+
+case class SnapshotReference(
+  ordinal: SnapshotOrdinal,
+  height: Height,
+  subHeight: Height,
+  snapshots: Set[StateChannelSnapshot],
```
tbekas, line 13: I think that this doesn't have to be in the `SnapshotReference`. The `GlobalSnapshot` is going to contain state channel snapshot.

modules/core/src/main/scala/org/tessellation/domain/snapshot/SnapshotService.scala
```diff
@@ -0,0 +1,8 @@
+package org.tessellation.domain.snapshot
+
+import org.tessellation.schema.height.Height
+
+trait SnapshotService[F[_]] {
+  def createSnapshotTrigger(): F[SnapshotReference]
```
tbekas, line 6: The parenthesis are not necessary.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/SnapshotService.scala
```diff
@@ -27,48 +28,59 @@ object SnapshotService {
...
-          .fromQueueUnterminated(blocksQueue)
-          .evalMap(createSnapshot)
-          .through(switchRepeat(config.fallbackTriggerTimeout, Stream.eval(createSnapshot())))
+          .fromQueueUnterminated(l1DataQueue)
+          .zip(Stream.eval(snapshotStorage.getLastSnapshotHeight))
+          .flatMap {
+            case (data, lastSnapshotHeight) =>
+              snapshotPreconditionsValidator.validate(lastSnapshotHeight, data) match {
+                case Valid(lastSnapshotHeight) =>
+                  Stream.emit(data.block.asLeft[SnapshotReference]) ++ Stream
```
tbekas, line 45: I think it should be reversed, first the block and then the `SnapshotReference`.

tbekas (changes requested):

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/SnapshotService.scala
tbekas: I just realized that it is already reversed, I just misread that fragment 🤦

modules/core/src/main/scala/org/tessellation/config/SnapshotConfig.scala
```diff
@@ -0,0 +1,12 @@
...
+
+import scala.concurrent.duration.FiniteDuration
+
+import eu.timepit.refined.types.numeric.PosLong
+import fs2.io.file.Path
+
+case class SnapshotConfig(
+  storedSnapshotPath: Path,
+  fallbackTriggerTimeout: FiniteDuration,
+  heightInterval: PosLong
```
tbekas, line 11: I think we should hardcode this in the `Next` or `Previous` instances.

modules/core/src/main/scala/org/tessellation/domain/snapshot/SnapshotReference.scala
```diff
@@ -0,0 +1,15 @@
+package org.tessellation.domain.snapshot
+
+import cats.data.NonEmptySet
+
+import org.tessellation.kernel.StateChannelSnapshot
+import org.tessellation.schema.height.Height
+import org.tessellation.schema.peer.PeerId
+
+case class SnapshotReference(
```
tbekas, line 9: So after today's discussion, I think that `SnapshotReference` should contain only the `Height` as it's triggered by tips. The other trigger should produce some `case object` just to signal the occurrence of the trigger (even the `Unit` type would work, but for clarity, I would go with `case object`).

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/SnapshotService.scala
```diff
@@ -0,0 +1,59 @@
...
+import org.tessellation.config.SnapshotConfig
+import org.tessellation.dag.domain.block.DAGBlock
+import org.tessellation.domain.snapshot._
+import org.tessellation.ext.fs2.switchRepeat
+import org.tessellation.schema.height.Height
+import org.tessellation.security.signature.Signed
+
+import fs2.Stream
+
+object SnapshotService {
```
tbekas, line 20: Sorry for not mentioning this earlier in my review, but I think that this should be named something like `SnapshotTrigger`. It's not really a service I think.

tbekas (approved):
Looks good

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/SnapshotService.scala
marcinwadon: @tbekas don't you think it can be a SnapshotService responsibility as well? I mean, this class can be extended for other snapshot-related things. Just thinking out loudly 🤔

marcinwadon (approved):

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/SnapshotService.scala
tbekas: Yeah, but it may lead to creating `all-in-one` classes, for anything `Snapshot` related. It's nice to have small classes with a single responsibility.

kpudlik: Actually the only public API is the pipeline :) so technically it is just a `Stream`.
Maybe I can just do `make[F[_]]: Stream[F, Either[Signed[DAGBlock], SnapshotTrigger]]` and name it differently 🤔

marcinwadon (approved):

tbekas (approved):

PR #171, opened 2022-02-15
Ping Healthcheck - periodic check

marcinwadon:


modules/sdk/src/main/scala/org/tessellation/sdk/config/types.scala
```diff
@@ -46,4 +47,16 @@ object types {
...
     cliHttp: HttpServerConfig
   )
+
+  case class HealthCheckConfig(
+    ping: PingHealthCheckConfig,
+    removeUnresponsiveParallelPeersAfter: FiniteDuration
+  )
+
+  case class PingHealthCheckConfig(
+    concurrent: Int,
```
kpudlik, line 57: `PosInt`?  or `NonNegInt`

modules/sdk/src/main/scala/org/tessellation/sdk/config/types.scala
```diff
@@ -46,4 +47,16 @@ object types {
...
+    defaultCheckTimeout: FiniteDuration,
+    defaultCheckAttempts: Int,
```
kpudlik, line 59: same here

kpudlik (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/config/types.scala
TheMMaciek, line 57: maybe `concurrentChecks`? I guess that's what it means.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensus.scala
```diff
@@ -77,7 +87,7 @@ abstract class HealthCheckConsensus[
 
       def run = onNegativeOutcome(negative) >> onNonNegativeOutcome(nonNegative)
 
-      Spawn[F].start(run)
+      run.handleErrorWith(e => logger.warn(e)("Unexpected error on outcome action"))
```
tbekas, line 90: Shouldn't this be logged as an `error` instead of a `warn`?

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensusRound.scala
```diff
@@ -105,7 +134,7 @@ class HealthCheckConsensusRound[F[_]: Spawn: Clock, K <: HealthCheckKey, A <: He
     }
 
   private def sendProposal: F[Unit] =
-    ownConsensusHealthStatus.flatMap(gossip.spread)
+    ownConsensusHealthStatus.flatMap(status => gossip.spreadCommon(status))
```
tbekas, line 137: Remember that `spreadCommon` is identified by content, so the content should be unique. Maybe add `roundId` to the message?

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensus.scala
marcinwadon: We agreed that not every error should be logged as "error". This is only a fallback for the wrongly implemented `onNegative/onNonNegative` actions.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensusRound.scala
marcinwadon: there is a `roundId` and `owner` in the message, so it's unique for every round/peer

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/healthcheck/ping/PingHealthCheckConsensus.scala
```diff
@@ -1,49 +1,139 @@
...
+
+import eu.timepit.refined.auto.autoUnwrap
+import eu.timepit.refined.types.numeric.PosInt
 
-class PingHealthCheckConsensus[F[_]: Async: GenUUID: Ref.Make](
+class PingHealthCheckConsensus[F[_]: Async: GenUUID: Random](
   clusterStorage: ClusterStorage[F],
   selfId: PeerId,
   driver: HealthCheckConsensusDriver[PingHealthCheckKey, PingHealthCheckStatus, PingConsensusHealthStatus],
+  config: HealthCheckConfig,
```
TheMMaciek, line 33: Maybe passing `pingConfig` is enough here?

marcinwadon: No, because I need to pass full config to HealthCheckConsensus (see extend)

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensus.scala
tbekas: IMO errors in implementation always should be logged on `error` level.

tbekas (approved):

TheMMaciek (approved):

kpudlik (approved):

PR #172, opened 2022-02-15
Storage for GlobalSnapshot

tbekas:


modules/dag-shared/src/main/scala/org/tessellation/dag/dag.scala
```diff
@@ -0,0 +1,15 @@
...
+import org.tessellation.dag.snapshot.{GlobalSnapshot, SnapshotOrdinal, StateChannelSnapshotWrapper}
+
+package object dag {
+
+  val dagKryoRegistrar: Map[Class[_], Int] = Map(
+    classOf[GlobalSnapshot] -> 400,
+    classOf[StateChannelSnapshotWrapper] -> 401,
+    classOf[SnapshotOrdinal] -> 402,
+    classOf[DAGBlock] -> 403,
+    classOf[BlockReference] -> 404
```
TheMMaciek, line 13: These are under different numbers in `dag-L1` module which may possibly lead to errors.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/package.scala
```diff
@@ -0,0 +1,31 @@
...
+    .of(
+      PeerId(Hex("peer1")),
+      PeerId(Hex("peer2")),
+      PeerId(Hex("peer3"))
+    )
+
+  val genesis: GlobalSnapshot =
+    GlobalSnapshot(
+      SnapshotOrdinal.MinValue,
+      Height.MinValue,
```
kpudlik, line 23: I like that! :)

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/package.scala
```diff
@@ -0,0 +1,31 @@
...
+      SubHeight.MinValue,
+      Hash("0"),
+      Set.empty,
+      Map.empty,
+      genesisNextFacilitators
+    )
+
+}
```
TheMMaciek, line 31: Not sure if we should merge it to develop, given that it's a mock.

modules/core/src/main/scala/org/tessellation/domain/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -0,0 +1,19 @@
...
+import org.tessellation.dag.snapshot.{GlobalSnapshot, SnapshotOrdinal, StateChannelSnapshotWrapper}
+import org.tessellation.schema.address.Address
+
+trait GlobalSnapshotStorage[F[_]] {
+
+  def add(globalSnapshot: GlobalSnapshot): F[Unit]
+
+  def getLast: F[GlobalSnapshot]
+
+  def get(ordinal: SnapshotOrdinal): F[Option[GlobalSnapshot]]
```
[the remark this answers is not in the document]
tbekas: Until we have something like `GlobalSnapshotInfo`, we will need to store all snapshots, unfortunately.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -0,0 +1,71 @@
...
+      }
+
+    def getLast: F[GlobalSnapshot] = snapshotsRef.get.map(_.head)
+
+    def getStateChannelSnapshot(
+      address: Address,
+      ordinal: SnapshotOrdinal
+    ): F[Option[StateChannelSnapshotWrapper]] = snapshotsRef.get.map { snapshots =>
+      snapshots.find { snapshot =>
+        snapshot.ordinal <= ordinal && snapshot.stateChannelSnapshots.contains(address)
```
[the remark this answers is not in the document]
tbekas: I think that the method name may be misleading, I will rename it. The intention is to return the most recent state-channel snapshot that's not above passed ordinal. It may happen that in the GlobalSnapshot at ordinal 7 there will be no state-channel snapshot for a given address, so the most recent one will be earlier (if any).

modules/dag-shared/src/main/scala/org/tessellation/dag/dag.scala
tbekas: I didn't realize that! I think we should remove it then from `dag-l1` and keep it only in `dag-shared`.

modules/shared/src/main/scala/org/tessellation/schema/kryo/package.scala
```diff
@@ -22,6 +25,7 @@ package object kryo {
     classOf[SignRequest] -> 104,
     classOf[NonEmptyList[_]] -> 105,
     classOf[Signed[_]] -> 106,
+    classOf[Refined[_, _]] -> 107,
```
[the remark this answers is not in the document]
tbekas: Me neither but without it, serialization fails.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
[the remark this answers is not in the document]
marcinwadon: Ok, please change the name accordingly. I had the same thought as @TheMMaciek.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/package.scala
marcinwadon: I'm fine with merging mock/stub things into develop. It's out development branch and doesn't need to be production-ready. Yet, should be compilable. Hence, having such stubs helps to focus on a single thing and stubbing the others that can be implemented later.

modules/dag-shared/src/main/scala/org/tessellation/dag/dag.scala
marcinwadon: yup, things should be defined only in its own registrar. we should avoid cross-module registration

modules/shared/src/main/scala/org/tessellation/schema/kryo/package.scala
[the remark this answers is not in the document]
marcinwadon: are you able to track which class produces this error, @tbekas?
you should see the exact place (class and field) that is causing that. maybe that way we can identify whether we should register Refined or change the schema 🤔

marcinwadon (changes requested):

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -0,0 +1,71 @@
...
+import org.tessellation.domain.snapshot.GlobalSnapshotStorage
+import org.tessellation.ext.crypto._
+import org.tessellation.kryo.KryoSerializer
+import org.tessellation.schema.address._
+import org.tessellation.security.hash.Hash
+
+object GlobalSnapshotStorage {
+
+  def make[F[_]: Async: KryoSerializer: Ref.Make](genesis: GlobalSnapshot): F[GlobalSnapshotStorage[F]] =
+    Ref[F].of[NonEmptyList[GlobalSnapshot]](NonEmptyList.of(genesis)).map(ref => make(ref))
```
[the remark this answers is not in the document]
tbekas: Interestingly the `.map(make)` doesn't compile, I tried it already. The compiler error is very unambiguous though. The `map(make(_))` does work though so I will change it to that.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -0,0 +1,71 @@
...
+import org.tessellation.dag.snapshot.{GlobalSnapshot, SnapshotOrdinal, StateChannelSnapshotWrapper}
+import org.tessellation.domain.snapshot.GlobalSnapshotStorage
+import org.tessellation.ext.crypto._
+import org.tessellation.kryo.KryoSerializer
+import org.tessellation.schema.address._
+import org.tessellation.security.hash.Hash
+
+object GlobalSnapshotStorage {
+
+  def make[F[_]: Async: KryoSerializer: Ref.Make](genesis: GlobalSnapshot): F[GlobalSnapshotStorage[F]] =
```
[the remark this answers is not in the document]
tbekas: Fixing.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -0,0 +1,71 @@
...
+    Ref[F].of[NonEmptyList[GlobalSnapshot]](NonEmptyList.of(genesis)).map(ref => make(ref))
+
+  def make[F[_]: Async: KryoSerializer: Ref.Make](
```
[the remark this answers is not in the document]
tbekas: Fixing.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -0,0 +1,71 @@
...
+    snapshotsRef: Ref[F, NonEmptyList[GlobalSnapshot]]
```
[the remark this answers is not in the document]
tbekas: Then  how would you name this argument `snapshots.modify { argument => }`? If you would also use `snapshots`, then it would shadow the outer name and would be unambiguous when reading the code.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -0,0 +1,71 @@
...
+              val expectedLink = (lastSnapshotHash, SnapshotOrdinal.next.next(lastSnapshot.ordinal))
+              val actualLink = (globalSnapshot.lastSnapshotHash, globalSnapshot.ordinal)
+              if (expectedLink === actualLink)
+                (globalSnapshot :: snapshots, none[Throwable])
+              else
+                (snapshots, InvalidGlobalSnapshotChain(expectedLink, actualLink).some)
+          }
+        }
+        _ <- maybeError.traverse(_.raiseError[F, Unit])
+      } yield ()
```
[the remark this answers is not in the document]
tbekas: Ok, looks good.

modules/shared/src/main/scala/org/tessellation/schema/kryo/package.scala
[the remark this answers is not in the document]
tbekas: The field is `subHeight` in `GlobalSnapshot` which has a type defined like this:
```
@newtype
case class SubHeight(value: NonNegLong)
```
And the `NonNegLong` type is defined like this:
```
type NonNegLong = Long Refined NonNegative
```

The `Refined` is an infix notation for the wrapper class `Refined`.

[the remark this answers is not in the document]
marcinwadon: Ok @tbekas, looks correct so I'd leave that as you did.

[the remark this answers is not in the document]
tbekas: Interestingly, it doesn't need any registration when I use `PosLong` (we already use it in Gossip) 🤔

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -0,0 +1,68 @@
...
+
+  def make[F[_]: Async: KryoSerializer](genesis: GlobalSnapshot): F[GlobalSnapshotStorage[F]] =
+    Ref[F].of[NonEmptyList[GlobalSnapshot]](NonEmptyList.of(genesis)).map(make(_))
+
+  def make[F[_]: Async: KryoSerializer](
+    snapshots: Ref[F, NonEmptyList[GlobalSnapshot]]
+  ): GlobalSnapshotStorage[F] = new GlobalSnapshotStorage[F] {
+
+    def save(snapshot: GlobalSnapshot): F[Unit] =
+      for {
```
marcinwadon, line 30: why such a weird construction? :D remove this for comprehension, it does nothing

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/Main.scala
```diff
@@ -22,7 +23,7 @@ import com.monovore.decline.Opts
 object Main extends TessellationIOApp[Run]("", "DAG L1 node", version = BuildInfo.version) {
   val opts: Opts[Run] = cli.method.opts
 
-  val kryoRegistrar: Map[Class[_], Int] = stateChannelKryoRegistrar
+  val kryoRegistrar: Map[Class[_], Int] = stateChannelKryoRegistrar ++ dagSharedKryoRegistrar
```
marcinwadon, line 26: please check if `stateChannelKryoRegistrar` doesn't have `dagShared` already combined

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/kryo/package.scala
```diff
@@ -15,10 +12,7 @@ package object kryo {
     classOf[Proposal] -> 1002,
     classOf[BlockProposal] -> 1003,
     classOf[CancelledBlockCreationRound] -> 1004,
-    classOf[UUID] -> 1005,
```
marcinwadon, line 18: was UUID registered twice?

marcinwadon (changes requested):

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -0,0 +1,68 @@
...
+          }
+        }.flatten
+      } yield ()
+
+    def get(ordinal: SnapshotOrdinal): F[Option[GlobalSnapshot]] =
+      snapshots.get.map(_.find(_.ordinal === ordinal))
+
+    def getLast: F[GlobalSnapshot] = snapshots.get.map(_.head)
+
+    def getStateChannelSnapshotBelowOrdinal(
```
TheMMaciek, line 52: Implementation looks like it is more like 'UntilOrdinal' / `AtOrBelowOrdinal` or something. I mean it may also be the specified ordinal. It's not always gonna be below the ordinal.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/kryo/package.scala
tbekas: Yes, once in `shared` and secondly here. Also, I think we should have a rule, that we register all classes that aren't part of the current module only in the `shared` module. That would include all JDK and Scala and library classes.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
tbekas: You're right. `until` sounds good.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
tbekas: It is a leftover, I had more expressions there, but now they're gone and I didn't realize that this for is no longer necessary.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -0,0 +1,66 @@
...
+    snapshotsRef: Ref[F, NonEmptyList[GlobalSnapshot]]
+  ): GlobalSnapshotStorage[F] = new GlobalSnapshotStorage[F] {
+
+    def save(snapshot: GlobalSnapshot): F[Unit] =
+      snapshotsRef.modify { snapshots =>
+        val lastSnapshot = snapshots.head
+        lastSnapshot.hash match {
+          case Left(error) => (snapshots, error.raiseError[F, Unit])
+          case Right(lastSnapshotHash) =>
+            val expectedLink = (lastSnapshotHash, SnapshotOrdinal.next.next(lastSnapshot.ordinal))
```
marcinwadon, line 35: can you rebase and use `Next[SnapshotOrdinal]` syntax?

modules/core/src/main/scala/org/tessellation/modules/Storages.scala
```diff
@@ -5,31 +5,36 @@ import cats.syntax.flatMap._
...
 object Storages {
 
-  def make[F[_]: Async: Database](
+  def make[F[_]: Async: Database: KryoSerializer](
     sdkStorages: SdkStorages[F]
   ): F[Storages[F]] =
     for {
       addressStorage <- AddressStorage.make[F]
       trustStorage <- TrustStorage.make[F]
+      globalSnapshotStorage <- GlobalSnapshotStorage.make[F](org.tessellation.infrastructure.snapshot.genesis)
```
marcinwadon, line 28: can you import `org.tessellation.infrastructure.snapshot.genesis` and pass `genesis` here only?

modules/core/src/test/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorageSuite.scala
```diff
@@ -0,0 +1,65 @@
...
+    for {
+      storage <- GlobalSnapshotStorage.make(genesis)
+      nextSnapshot = genesis
+        .focus(_.ordinal)
+        .modify(SnapshotOrdinal.next.next)
+      maybeError <- storage
+        .save(nextSnapshot)
+        .map(_ => none[Throwable])
+        .handleError(_.some)
+    } yield verify(maybeError.exists(_.isInstanceOf[InvalidGlobalSnapshotChain]))
```
marcinwadon, line 48: I think that you can expect `failure` instead of mapping an error to option and checking the option itself 🤔

marcinwadon (changes requested):

modules/core/src/test/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorageSuite.scala
tbekas: Failure is not the same as expecting an error. Failure means that you unconditionally fail the test. I haven't found any syntax for expecting an error in Weaver.

marcinwadon: Ah, right. I misunderstood the documentation.

marcinwadon (approved):

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
tbekas: I used the `_.next` syntax instead.

marcinwadon (approved):

PR #174, opened 2022-02-16
L0 peers discovery functionality

TheMMaciek:
I will still look into whether L0 nodes can be discovered from L1 peers (except for the initial node starting L1 layer). But this is a working and tested implementation of L0 peers discovery based on declaring initial L0 peer during node startup. Doing it that way also seems correct to me.

### DIFF
--- a/modules/core/src/main/scala/org/tessellation/modules/HttpApi.scala
+++ b/modules/core/src/main/scala/org/tessellation/modules/HttpApi.scala
@@ -57,10 +57,13 @@ sealed abstract class HttpApi[F[_]: Async: SecurityProvider: KryoSerializer] pri
   private val metricRoutes = MetricRoutes[F](services).routes
 
   private val openRoutes: HttpRoutes[F] =
-    `X-Id-Middleware`.responseMiddleware(selfId) {
-      (if (environment == Testnet || environment == Dev) debugRoutes else HttpRoutes.empty) <+>
-        metricRoutes <+> stateChannelRoutes.publicRoutes <+> clusterRoutes.publicRoutes
-    }
+    PeerAuthMiddleware
+      .responseSignerMiddleware(privateKey, storages.session) {
+        `X-Id-Middleware`.responseMiddleware(selfId) {
+          (if (environment == Testnet || environment == Dev) debugRoutes else HttpRoutes.empty) <+>
+            metricRoutes <+> stateChannelRoutes.publicRoutes <+> clusterRoutes.publicRoutes
+        }
+      }
 
   private val getKnownPeersId: Host => F[Set[PeerId]] = { (host: Host) =>
     storages.cluster.getPeers(host).map(_.map(_.id))
--- a/modules/dag-l1/src/main/scala/org/tessellation/dag/l1/Main.scala
+++ b/modules/dag-l1/src/main/scala/org/tessellation/dag/l1/Main.scala
@@ -13,6 +13,7 @@ import org.tessellation.dag.l1.kryo.stateChannelKryoRegistrar
 import org.tessellation.dag.l1.modules._
 import org.tessellation.ext.cats.effect.ResourceIO
 import org.tessellation.schema.node.NodeState
+import org.tessellation.schema.node.NodeState.SessionStarted
 import org.tessellation.sdk.app.{SDK, TessellationIOApp}
 import org.tessellation.sdk.infrastructure.gossip.RumorHandlers
 import org.tessellation.sdk.resources.MkHttpServer
@@ -33,11 +34,12 @@ object Main extends TessellationIOApp[Run]("", "DAG L1 node", version = BuildInf
     Database.forAsync[IO](cfg.db).flatMap { implicit database =>
       for {
         queues <- Queues.make[IO](sdkQueues).asResource
-        storages <- Storages.make[IO](cfg.tips, sdkStorages).asResource
+        storages <- Storages.make[IO](cfg.tips, sdkStorages, method.l0Peer).asResource
         validators = Validators.make[IO](storages, cfg.blockValidator)
         services = Services.make[IO](storages, validators, sdkServices)
         p2pClient = P2PClient.make(sdkP2PClient, sdkResources.client)
         healthChecks <- HealthChecks.make[IO](storages, services, p2pClient, cfg.healthCheck, sdk.nodeId).asResource
+        programs = Programs.make(sdkPrograms, p2pClient, storages)
 
         rumorHandler = RumorHandlers.make[IO](storages.cluster, healthChecks.ping).handlers <+>
           blockRumorHandler(queues.peerBlock)
@@ -46,19 +48,22 @@ object Main extends TessellationIOApp[Run]("", "DAG L1 node", version = BuildInf
           .start(storages, services, queues, p2pClient, rumorHandler, nodeId, cfg)
           .asResource
 
-        api = HttpApi.make[IO](storages, queues, keyPair.getPrivate, services, sdkPrograms, sdk.nodeId)
+        api = HttpApi.make[IO](storages, queues, keyPair.getPrivate, services, programs, sdk.nodeId)
         _ <- MkHttpServer[IO].newEmber(ServerName("public"), cfg.http.publicHttp, api.publicApp)
         _ <- MkHttpServer[IO].newEmber(ServerName("p2p"), cfg.http.p2pHttp, api.p2pApp)
         _ <- MkHttpServer[IO].newEmber(ServerName("cli"), cfg.http.cliHttp, api.cliApp)
         stateChannel = new StateChannel(cfg, keyPair, p2pClient, queues, nodeId, services, storages, validators)
         _ <- {
           method match {
-            case _: RunInitialValidator =>
-              storages.node.tryModifyState(NodeState.Initial, NodeState.ReadyToJoin) >>
-                services.session.createSession
+            case cfg: RunInitialValidator =>
+              programs.l0PeerDiscovery.discoverFrom(cfg.l0Peer) >>
+                storages.node.tryModifyState(NodeState.Initial, NodeState.ReadyToJoin) >>
+                services.session.createSession >>
+                storages.node.tryModifyState(SessionStarted, NodeState.Ready)
 
-            case _: RunValidator =>
-              storages.node.tryModifyState(NodeState.Initial, NodeState.ReadyToJoin)
+            case cfg: RunValidator =>
+              programs.l0PeerDiscovery.discoverFrom(cfg.l0Peer) >>
+                storages.node.tryModifyState(NodeState.Initial, NodeState.ReadyToJoin)
           }
         }.asResource
         _ <- stateChannel.runtime.compile.drain.asResource
--- a/modules/dag-l1/src/main/scala/org/tessellation/dag/l1/cli/method.scala
+++ b/modules/dag-l1/src/main/scala/org/tessellation/dag/l1/cli/method.scala
@@ -9,7 +9,8 @@ import org.tessellation.dag.block.config.BlockValidatorConfig
 import org.tessellation.dag.l1.config.TipsConfig
 import org.tessellation.dag.l1.config.types.{AppConfig, DBConfig}
 import org.tessellation.dag.l1.domain.consensus.block.config.ConsensusConfig
-import org.tessellation.sdk.cli.CliMethod
+import org.tessellation.schema.peer.L0Peer
+import org.tessellation.sdk.cli.{CliMethod, L0PeerOpts}
 import org.tessellation.sdk.config.AppEnvironment
 import org.tessellation.sdk.config.types._
 
@@ -20,6 +21,7 @@ object method {
 
   sealed trait Run extends CliMethod {
     val dbConfig: DBConfig
+    val l0Peer: L0Peer
 
     val appConfig: AppConfig = AppConfig(
       environment = environment,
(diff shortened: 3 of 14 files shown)


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/cli/method.scala
```diff
@@ -52,20 +56,38 @@ object method {
...
+  val l0PeerHostOpts: Opts[Host] = Opts
+    .option[Host]("l0-peer-host", help = "L0 peer HTTP host")
+    .orElse(Opts.env[Host]("CL_L0_PEER_HTTP_HOST", help = "L0 peer HTTP host"))
+
+  val l0PeerPortOpts: Opts[Port] = Opts
+    .option[Port]("l0-peer-port", help = "L0 peer HTTP port")
+    .orElse(Opts.env[Port]("CL_L0_PEER_HTTP_PORT", help = "L0 peer HTTP port"))
+    .withDefault(port"9000")
+
+  val l0PeerOpts: Opts[L0Peer] =
```
TheMMaciek, line 72: Should we put it in `L0Peer` object in shared? 🤔

modules/core/src/main/scala/org/tessellation/domain/cluster/services/L0Cluster.scala
```diff
@@ -0,0 +1,7 @@
+package org.tessellation.domain.cluster.services
+
+import org.tessellation.schema.peer.L0Peer
+
+trait L0Cluster[F[_]] {
+  def getL0Peers: F[Set[L0Peer]]
+}
```
marcinwadon, line 7: L0 cluster is native in this layer/module, so `clusterStorage.getPeers` should be used

modules/core/src/main/scala/org/tessellation/http/routes/L0ClusterRoutes.scala
```diff
@@ -0,0 +1,24 @@
...
+import org.http4s.circe.CirceEntityCodec.circeEntityEncoder
+import org.http4s.dsl.Http4sDsl
+import org.http4s.server.Router
+
+final case class L0ClusterRoutes[F[_]: Monad](l0Cluster: L0Cluster[F]) extends Http4sDsl[F] {
+  private val prefixPath = "/l0-cluster"
+
+  private val httpRoutes: HttpRoutes[F] = HttpRoutes.of[F] {
+    case GET -> Root / "peers" =>
+      Ok(l0Cluster.getL0Peers)
```
marcinwadon, line 17: this should be obtained by using `cluster/info` endpoint. you can also add the response signer there if you want

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/Main.scala
```diff
@@ -45,19 +46,21 @@ object Main extends TessellationIOApp[Run]("", "DAG L1 node", version = BuildInf
...
+        api = HttpApi.make[IO](storages, queues, keyPair.getPrivate, services, programs, sdk.nodeId)
         _ <- MkHttpServer[IO].newEmber(ServerName("public"), cfg.http.publicHttp, api.publicApp)
         _ <- MkHttpServer[IO].newEmber(ServerName("p2p"), cfg.http.p2pHttp, api.p2pApp)
         _ <- MkHttpServer[IO].newEmber(ServerName("cli"), cfg.http.cliHttp, api.cliApp)
         stateChannel = new StateChannel(cfg, keyPair, p2pClient, queues, nodeId, services, storages, validators)
         _ <- {
           method match {
-            case _: RunInitialValidator =>
-              storages.node.tryModifyState(NodeState.Initial, NodeState.ReadyToJoin) >>
+            case RunInitialValidator(_, _, _, _, _, _, l0Peer) =>
```
marcinwadon, line 56: instead of ignoring all values, I think it would be cleaner using:
```scala
case cfg: RunInitialValidator =>
  ....discoverFrom(cfg.l0Peer)
```

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/cli/method.scala
marcinwadon: I don't think so. I'd put it into `sdk` rather than `shared`. Yet, I still stand for L0 peer hardcoding for genesis L1 node only and peer-discovery for the rest L1 nodes. Anyway, I can agree to have such a functionality as a middle-point to test L0/L1 interoperability.

marcinwadon (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/cli/method.scala
TheMMaciek: `opts` for classes like `Password` `KeyAlias`, live in companion objects in `shared` module, that's why I wonder. Would `org.tessellation.sdk.cli` and `peer` object make sense? 🤔

modules/core/src/main/scala/org/tessellation/domain/cluster/services/L0Cluster.scala
TheMMaciek: I could use `cluster.info` in the routes class and do the adjusting to `L0Peer` there if that's what you mean? That would indeed make more sense than creating additional `L0Cluster`.

TheMMaciek: I could use `cluster.info` in the routes class and do the adjusting to `L0Peer` there if that's what you mean? That would indeed make more sense than creating additional `L0Cluster`.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/Main.scala
TheMMaciek: Right!

modules/core/src/main/scala/org/tessellation/http/routes/L0ClusterRoutes.scala
TheMMaciek: I would rather not reuse the `cluster/info` endpoint because then the adjusting to `L0Peer` would happen on the client side. Hence every cluster can be treated as `L0`. I would rather have a specific endpoint for fetching L0 peers on L0 layer.

marcinwadon: I think that's proper way of doing that. L1 shouldn't recognize L0 as actually L0. It should use it as a "layer". Eventually, you can create an L2 cluster that will connect to L1 layer (so no L0 peers but L1 peers). and functionality stays the same

marcinwadon: So to sum up - L1 should know it's being connected to L0 layer but L0 layer shouldn't expose itself as L0 directly.

modules/core/src/main/scala/org/tessellation/modules/HttpApi.scala
```diff
@@ -59,8 +59,8 @@ sealed abstract class HttpApi[F[_]: Async: SecurityProvider: KryoSerializer] pri
   private val openRoutes: HttpRoutes[F] =
     `X-Id-Middleware`.responseMiddleware(selfId) {
       (if (environment == Testnet || environment == Dev) debugRoutes else HttpRoutes.empty) <+>
-        metricRoutes <+> stateChannelRoutes.publicRoutes <+> clusterRoutes.publicRoutes
-    }
+        metricRoutes <+> stateChannelRoutes.publicRoutes
+    } <+> PeerAuthMiddleware.responseSignerMiddleware(privateKey, storages.session)(clusterRoutes.publicRoutes)
```
marcinwadon, line 63: Can you check again if we can sign all public responses?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/Main.scala
```diff
@@ -46,19 +47,21 @@ object Main extends TessellationIOApp[Run]("", "DAG L1 node", version = BuildInf
...
+        api = HttpApi.make[IO](storages, queues, keyPair.getPrivate, services, programs, sdk.nodeId)
         _ <- MkHttpServer[IO].newEmber(ServerName("public"), cfg.http.publicHttp, api.publicApp)
         _ <- MkHttpServer[IO].newEmber(ServerName("p2p"), cfg.http.p2pHttp, api.p2pApp)
         _ <- MkHttpServer[IO].newEmber(ServerName("cli"), cfg.http.cliHttp, api.cliApp)
         stateChannel = new StateChannel(cfg, keyPair, p2pClient, queues, nodeId, services, storages, validators)
         _ <- {
           method match {
-            case _: RunInitialValidator =>
-              storages.node.tryModifyState(NodeState.Initial, NodeState.ReadyToJoin) >>
+            case cfg: RunInitialValidator =>
```
marcinwadon, line 57: That peer should has Ready state, not ReadyForJoining

marcinwadon: and I'd wrap peer discovery into own statuses (see genesis on L0) but it's up to you

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/L0PeerDiscovery.scala
```diff
@@ -0,0 +1,43 @@
...
+) {
+
+  def discoverFrom(peer: P2PContext): F[Unit] = {
+    for {
+      peers <- l0ClusterClient.getPeers.run(peer)
+      l0Peers = peers.map(p => L0Peer(p.id, p.ip, p.publicPort))
+      _ <- l0ClusterStorage.addPeers(l0Peers)
+    } yield ()
+  }.handleErrorWith {
+    case _ => Sync[F].raiseError(UnknownL0PeerDiscoveryError)
```
marcinwadon, line 41: If so, MonadThrow raiseError

modules/shared/src/main/scala/org/tessellation/ext/collection/NonEmptyMapUtils.scala
```diff
@@ -0,0 +1,14 @@
+package org.tessellation.ext.collection
+
+import cats.data.{NonEmptyMap, NonEmptySet}
+
+import scala.collection.immutable.SortedSet
+
+object NonEmptyMapUtils {
```
marcinwadon, line 7: Can we name it accordingly to cats naming convention for ops?

marcinwadon (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/Main.scala
TheMMaciek: Not sure if it needs to be Ready. It's ready to join (or be joined) then it creates the session and changes to Ready just like other nodes do at the first peer joining. Anyway that's not the scope of this PR so if we want to we can change it in a separate one. Regarding own statuses even if I created own list of them the current ones seem to be baked in into the cluster functionality in SDK pretty tightly. We could maybe have some more generic statuses added to SDK so that state channels could use them. I can wrap the L0 peer discovery in switch from `Initial` to `ReadyToJoin` but I guess that's not what you meant.

marcinwadon: I believe it is in the scope of this PR because you're adding peer discovery. If node starts as an Initial peer, it needs to start a session and change its status to Ready, and accept incoming joins. 
Nodes that are starting as a validator needs to set status as ReadyToJoin, so then they can be connected to the network

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/L0PeerDiscovery.scala
```diff
@@ -0,0 +1,44 @@
...
+    override def getMessage: String = "Error during L0 peer discovery!"
+  }
+}
+
+sealed abstract class L0PeerDiscovery[F[_]: Sync] private (
+  l0ClusterClient: L0ClusterClient[F],
+  l0ClusterStorage: L0ClusterStorage[F]
+) {
+
+  def discoverFrom(peer: P2PContext): F[Unit] = {
```
marcinwadon, line 35: I'd be against using for comprehension. Instead, you can think about data flow and use locally reasoning. (for comp introduces kinda global scope)

```scala
l0ClusterClient.getPeers.run(peer)
  .map(_.map(L0Peer.fromPeer))
  .flatMap(l0ClusterStorage.addPeers)
  .handleErrorWith(e => UnknownError...(e).raiseError[F, Unit])
```

TheMMaciek: I get what you mean. With this kind of linear flow of transformations it's unnecessary to name the arguments to then just pass them to the next step. I guess that could be the rule exactly, if you are going from one step to the other depending only on the last step's outcome then using for comprehension isn't needed. Though sometimes I would still use syntactic sugar like this if it improves readability. Not the case here though.

marcinwadon: but history shows that we have a problem with reasoning locally about arguments in for comp :)

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/Main.scala
```diff
@@ -46,19 +48,22 @@ object Main extends TessellationIOApp[Run]("", "DAG L1 node", version = BuildInf
...
         _ <- MkHttpServer[IO].newEmber(ServerName("cli"), cfg.http.cliHttp, api.cliApp)
         stateChannel = new StateChannel(cfg, keyPair, p2pClient, queues, nodeId, services, storages, validators)
         _ <- {
           method match {
-            case _: RunInitialValidator =>
-              storages.node.tryModifyState(NodeState.Initial, NodeState.ReadyToJoin) >>
-                services.session.createSession
+            case cfg: RunInitialValidator =>
+              programs.l0PeerDiscovery.discoverFrom(cfg.l0Peer) >>
+                storages.node.tryModifyState(NodeState.Initial, NodeState.ReadyToJoin) >>
```
marcinwadon, line 60: we may think about it as a `Genesis` for later usage, when the initial peer will make a download to pre-fetch latest global snapshot

marcinwadon (approved):

kpudlik (approved):

PR #176, opened 2022-02-17
ADR for Kryo class registration policy

tbekas:


marcinwadon (dismissed):
I like that idea :+1:

doc/adr/0009-kryo-class-registration-policy.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Context
+
+Having reliable serialization across different modules require using consistent and 
+non-conflicting registration ids.
+
+## Decision
+
+* Each module containing classes that are a subject of a kryo serialization declares 
+its kryo registrar - a public constant of type `Map[Class[_], Int]` representing kryo
```
kpudlik, line 17: As a bonus feature, we can check if `Class[_]` have duplicates and if `Int` have duplicates and throw if it happens as we talked.

doc/adr/0009-kryo-class-registration-policy.md
```diff
@@ -0,0 +1,40 @@
...
+| module     | min id | max id |
+| ---------- | ------ | ------ |
+| shared     | 100    | 299    |
+| kernel     | 300    | 399    |
+| sdk        | 400    | 499    |
+| dag-shared | 500    | 599    |
+| core       | 600    | 699    |
+| dag-l1     | 700    | 799    |
+
+* State-channels can use registration ids starting from 1000.
```
kpudlik, line 35: That one as well can be secured by refined type

doc/adr/0009-kryo-class-registration-policy.md
tbekas: Fully agree, that's an implementation detail however I think so I would like to not include it in ADR.

doc/adr/0009-kryo-class-registration-policy.md
tbekas: It actually already is. There's a type alias `KryoRegistrationId = Int Refined Greater[1000]` in `org.tesselation.kernel`.

doc/adr/0009-kryo-class-registration-policy.md
```diff
@@ -0,0 +1,40 @@
...
+
+| module     | min id | max id |
+| ---------- | ------ | ------ |
+| shared     | 100    | 299    |
+| kernel     | 300    | 399    |
+| sdk        | 400    | 499    |
+| dag-shared | 500    | 599    |
+| core       | 600    | 699    |
+| dag-l1     | 700    | 799    |
+
```
TheMMaciek, line 34: Is the range [800, 999] intended as a backup?

tbekas: Yes, it's for future use.

TheMMaciek (approved):
Sounds good!

kpudlik (approved):

PR #177, opened 2022-02-17
Fix double joining

marcinwadon:


modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/SignClient.scala
```diff
@@ -28,9 +27,9 @@ object SignClient {
       def getRegistrationRequest: PeerResponse[F, RegistrationRequest] =
         PeerResponse[F, RegistrationRequest]("registration/request")(client)
 
-      def joinRequest(jr: JoinRequest): PeerResponse[F, Unit] =
-        PeerResponse[F, Unit]("cluster/join", POST)(client) { (req, c) =>
-          c.successful(req.withEntity(jr)).void
+      def joinRequest(jr: JoinRequest): PeerResponse[F, Boolean] =
```
tbekas, line 30: Maybe instead of a `Boolean` continue using `Unit` and change the response code from `200` to `409`. It may be useful if someone would create additional tooling around the CLI API (`curl` will return error exit code on responses outside of the 2xx family).

marcinwadon: Boolean isn't received from the endpoint. Take a look how `client.successful` works. I just expect that the request succeeded and it returns me yes or no.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
```diff
@@ -153,6 +161,15 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
       _ <- verifySignRequest(signRequest, signedSignRequest, PeerId._Id.get(withPeer.id))
         .ifM(Applicative[F].unit, HandshakeSignatureNotValid.raiseError[F, Unit])
 
+      _ <- if (skipJoinRequest) {
+        Applicative[F].unit
+      } else {
+        cluster.getRegistrationRequest
+          .map(JoinRequest.apply)
+          .flatMap(signClient.joinRequest(_).run(withPeer))
+          .ifM(Applicative[F].unit, new Throwable("Unexpected error occured.").raiseError[F, Unit])
```
tbekas, line 170: You can use more descriptive error, like "Unexpected error occurred when joining with peer ${withPeer.id}"

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
```diff
@@ -129,11 +126,22 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
...
+
+  private def validateHandshakeConditions(id: PeerId, ip: Host, p2pPort: Port): F[Unit] =
+    for {
+      hasPeerId <- clusterStorage.hasPeerId(id)
+      _ <- if (hasPeerId) PeerIdInUse(id).raiseError[F, Unit] else Applicative[F].unit
 
-      hasPeerHostPort <- clusterStorage.hasPeerHostPort(toPeer.ip, toPeer.p2pPort)
-      _ <- if (hasPeerHostPort) PeerHostPortInUse(toPeer.ip, toPeer.p2pPort).raiseError[F, Unit]
+      hasPeerHostPort <- clusterStorage.hasPeerHostPort(ip, p2pPort)
+      _ <- if (hasPeerHostPort) PeerHostPortInUse(ip, p2pPort).raiseError[F, Unit]
```
TheMMaciek, line 144: Maybe you could use `.ifM` here given that neither `hasPeerId` nor `hasPeerHostPort` are used anywhere further.

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/SignClient.scala
tbekas: Ah, alright then, I misread that!

tbekas (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
marcinwadon: Done :+1:

kpudlik (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
```diff
@@ -129,14 +126,22 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
...
+    validateHandshakeConditions(toPeer.id, toPeer.ip, toPeer.p2pPort)
+
+  private def validateHandshakeConditions(req: RegistrationRequest): F[Unit] =
+    validateHandshakeConditions(req.id, req.ip, req.p2pPort)
+
+  private def validateHandshakeConditions(id: PeerId, ip: Host, p2pPort: Port): F[Unit] =
+    clusterStorage.hasPeerId(id).ifM(PeerIdInUse(id).raiseError[F, Unit], Applicative[F].unit).flatMap { _ =>
+      clusterStorage
+        .hasPeerHostPort(ip, p2pPort)
+        .ifM(PeerHostPortInUse(ip, p2pPort).raiseError[F, Unit], Applicative[F].unit)
```
TheMMaciek, line 142: I think using for comprehension still could make it more readable. But it's up to you.

TheMMaciek (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
marcinwadon: I'm against using for..comp if it's not needed. Composing pipelines without syntax sugar makes me think better and upfront of data flow in the pipeline. for...comp looks more like imperative programming and can be misleading sometimes

PR #180, opened 2022-02-18
State channel output gist

tbekas:


marcinwadon (approved):

modules/core/src/main/scala/org/tessellation/kryo/package.scala
```diff
@@ -1,9 +1,11 @@
 package org.tessellation
 
+import org.tessellation.domain.aci.StateChannelGistedOutput
+
 package object kryo {
 
   val coreKryoRegistrar: Map[Class[_], Int] = Map(
-    // classOf[T] -> 123
+    classOf[StateChannelGistedOutput[_]] -> 189
```
marcinwadon, line 8: Please adjust it according to #176

marcinwadon (changes requested):

modules/core/src/main/scala/org/tessellation/kryo/package.scala
tbekas: I will rebase the change and fix it after #182 gets merged.

marcinwadon (approved):

PR #182, opened 2022-02-20
Adjust kryo registration ids to ADR #9

tbekas:


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/package.scala
```diff
@@ -0,0 +1,22 @@
+package org.tessellation.dag
+
+import org.tessellation.dag.l1.domain.block.Tips
+import org.tessellation.dag.l1.domain.consensus.block.BlockConsensusInput._
+import org.tessellation.dag.l1.domain.consensus.block.CancellationReason
+import org.tessellation.dag.l1.domain.consensus.block.CancellationReason._
+
+package object l1 {
+
+  val dagL1KryoRegistrar: Map[Class[_], Int] = Map(
```
marcinwadon, line 10: @kpudlik @TheMMaciek @tbekas let's discuss it tomorrow, whether it should be registered in its own scope like now, or above 1000 as a state channel. L1 is implemented as a Cell, and Cell and a "state channel" names the same.

tbekas: >L1 is implemented as a Cell, and Cell and a "state channel" names the same.

I think we need to agree on a common dictionary. 

For me, the State Channel term refers to something more than a Cell. A Cell is just a transformation from some input to output (with an effect) so it is stateless by its nature just as any other transformation is. A State Channel on the other hand represents a Cell registered under some specific address along with all the data that was ever produced by that Cell. A single Cell definition could be used in different State Channels, kinda like you could use the same ERC-20 definition in different smart contracts.

Having that in mind I don't see a problem with registering the `dag-l1` classes above the `1000`, I used the `700-799` range because that's what is in the [ADR #9](https://github.com/Constellation-Labs/tessellation/blob/develop/doc/adr/0009-kryo-class-registration-policy.md).

marcinwadon: Since the beginning state channel was a name for Cell, and this is how Cell was invented. Later on we used to name the instance of a Cell as a state channel (so it's registered). but I believe we should keep the naming that was presented in hypergraph hours and Wyatt's talks.

tbekas: Would you be able to provide a reference to this definition?

marcinwadon: Not really without searching

marcinwadon: I think we can find it in Alkimis doc as well

marcinwadon (approved):

PR #188, opened 2022-02-23
Global snapshot consensus

tbekas:


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -0,0 +1,182 @@
...
+        .updateState(key)(logIfUpdated(internalTryFacilitateConsensus(key, lastKeyAndArtifact)))
+
+    def tryAdvanceConsensus(
+      key: Key,
+      resources: ConsensusResources[Artifact]
+    ): F[Option[ConsensusState[Key, Artifact]]] =
+      consensusStorage
+        .updateState(key)(logIfUpdated(internalTryAdvanceConsensus(key, resources)))
+
+    private def logIfUpdated(fn: consensusStorage.StateUpdateFn): consensusStorage.StateUpdateFn =
```
marcinwadon, line 75: can you import `StateUpdateFn` directly?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -0,0 +1,182 @@
...
+                resources.peerDeclarations
+                  .get(peerId)
+                  .flatMap(peerDeclaration => peerDeclaration.signature.map(signature => (peerId, signature)))
+              }.map(_.map { case (id, signature) => SignatureProof(PeerId._Id.get(id), signature) })
+
+            val maybeSignedArtifact = for {
+              allSignatures <- maybeAllSignatures
+              allSignaturesNel <- NonEmptyList.fromList(allSignatures)
+              majorityArtifact <- resources.artifacts.get(majorityHash)
+            } yield Signed(majorityArtifact, allSignaturesNel)
```
marcinwadon, line 161: you can consider such a syntax as well:

```scala
(maybeAllSignatures.flatMap(NonEmptyList.fromList), resources.artifacts.get(majorityHash))
  .mapN(Signed(_, _)
```

marcinwadon (changes requested):

modules/core/src/main/scala/org/tessellation/infrastructure/aci/StateChannelRunner.scala
```diff
@@ -84,21 +84,25 @@ object StateChannelRunner {
...
-            .evalTap { output =>
-              instance.kryoSerializer.serialize(output).liftTo[F].flatMap { outputBytes =>
-                stateChannelOutputQueue.offer(StateChannelOutput(instance.address, output, outputBytes))
-              }
-            }
-            .compile
-            .drain
+          Spawn[F]
+            .start(
+              context.inputs
```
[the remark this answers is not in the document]
tbekas: Looks nice 👍

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensus.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  def make[F[_]: Async: KryoSerializer: SecurityProvider](
+    gossip: Gossip[F],
+    selfId: PeerId,
+    keyPair: KeyPair,
+    clusterStorage: ClusterStorage[F],
+    globalSnapshotStorage: GlobalSnapshotStorage[F]
+  ): F[Consensus[F, GlobalSnapshotEvent]] =
+    Consensus.make[F, GlobalSnapshotEvent, GlobalSnapshotKey, GlobalSnapshotArtifact](
+      GlobalSnapshotConsensusFunctions.make[F](globalSnapshotStorage, NonNegLong(2L)),
```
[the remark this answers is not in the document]
tbekas: Will do. That is `snapshotInterval`.

modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/GlobalSnapshot.scala
```diff
@@ -3,22 +3,22 @@ package org.tessellation.dag.snapshot
...
+@derive(encoder, decoder, eqv, show)
 case class GlobalSnapshot(
   ordinal: SnapshotOrdinal,
   height: Height,
   subHeight: SubHeight,
   lastSnapshotHash: Hash,
-  blocks: Set[DAGBlock],
-  stateChannelSnapshots: Map[Address, NonEmptyList[StateChannelSnapshotBinary]],
+  blocks: Set[Signed[DAGBlock]],
+  stateChannelSnapshots: Map[String, NonEmptyList[StateChannelSnapshotBinary]], // TODO change back to `Address`
```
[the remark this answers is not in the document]
tbekas: Yes, I am going to.

modules/core/src/main/scala/org/tessellation/package.scala
```diff
@@ -13,7 +15,15 @@ import io.estatico.newtype.ops._
...
-    classOf[StateChannelGistedOutput[_]] -> 700
+    classOf[StateChannelGistedOutput[_]] -> 700,
+    classOf[TipSnapshotTrigger] -> 701,
+    classOf[TimeSnapshotTrigger] -> 702,
+    classOf[SnapshotOrdinal] -> 703,
+    classOf[GlobalSnapshot] -> 704,
+    classOf[SnapshotOrdinal] -> 705,
+    classOf[SnapshotOrdinal] -> 706,
+    classOf[SnapshotOrdinal] -> 707,
+    classOf[SnapshotOrdinal] -> 708
```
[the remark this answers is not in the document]
tbekas: I think I was allocating entries, which I never used 😬 Thanks for catching it!

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -0,0 +1,107 @@
+package org.tessellation.sdk.infrastructure.consensus
+
+import cats.effect.Async
+import cats.kernel.Next
+import cats.syntax.all._
```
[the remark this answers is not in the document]
tbekas: Fixing it!

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -0,0 +1,107 @@
...
+      private val logger = Slf4jLogger.getLoggerFromClass[F](ConsensusManager.getClass)
+
+      def checkForTrigger(event: Event): F[Unit] =
+        consensusStorage.getLastKeyAndArtifact.flatMap(_.traverse(internalForCheckTrigger(_, event)).void)
+
+      def checkForStateUpdate(key: Key)(resources: ConsensusResources[Artifact]): F[Unit] =
+        for {
+          maybeState <- consensusStorage.getState(key)
+          _ <- maybeState.traverse(state => internalCheckForStateUpdate(key, state, resources))
+        } yield ()
```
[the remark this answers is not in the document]
tbekas: Yeah, I didn't like the way I wrote it either, but at the moment of writing it, I couldn't think of anything better.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -0,0 +1,107 @@
...
+
+      private def internalForCheckTrigger(lastKeyAndArtifact: (Key, Artifact), event: Event): F[Unit] =
+        if (consensusFns.triggerPredicate(lastKeyAndArtifact, event)) {
+          val nextKey = lastKeyAndArtifact._1.next
+          consensusStorage
+            .getState(nextKey)
+            .map(_.isEmpty)
+            .ifM(
+              logger.info(s"Triggering consensus for key ${nextKey.show}") >>
```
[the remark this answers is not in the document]
tbekas: Idk, I think it's actually a useful log and I would leave it on info. What do you think?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
tbekas: I can, but I will have to do it inside of the `make` function.  The `consensusStorage` is a parameter to `make` function. This is why I left it this way, to highlight that it isn't a static import.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -0,0 +1,182 @@
...
+            val maybeMajority = state.facilitators
+              .traverse(resources.peerDeclarations.get)
+              .flatMap(_.traverse(_.proposal))
+              .flatMap(pickMajority)
+
+            maybeMajority.traverse { majorityHash =>
+              for {
+                _ <- Applicative[F].unit
+                newState = state.copy(status = MajoritySelected[Artifact](majorityHash))
+                result <- setter(newState.some)
```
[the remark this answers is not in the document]
tbekas: Fixing it!

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
tbekas: Looks shorter, but Idk about the readability. I think that `for` in that case actually is more readable.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -0,0 +1,182 @@
...
+
+            val maybeSignedArtifact = for {
+              allSignatures <- maybeAllSignatures
+              allSignaturesNel <- NonEmptyList.fromList(allSignatures)
+              majorityArtifact <- resources.artifacts.get(majorityHash)
+            } yield Signed(majorityArtifact, allSignaturesNel)
+
+            maybeSignedArtifact.traverse { signedArtifact =>
+              for {
+                _ <- Applicative[F].unit
```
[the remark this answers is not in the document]
tbekas: Fixin'

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
marcinwadon: Ah, okay. I didnt catch it's a parameter

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStorage.scala
```diff
@@ -0,0 +1,222 @@
...
+              case (true, newState) =>
+                logger.trace(s"Consensus state for key ${key.show} transitioned to ${newState.show}") >>
+                  Applicative[F].pure(newState.some)
+              case (false, newState) =>
+                /** This should never occur as the access to the `ConsensusState` is protected by the semaphore */
+                logger.error(s"Consensus state for key ${key.show} failed to transition to ${newState.show}") >>
+                  Applicative[F].pure(none[ConsensusState[Key, Artifact]])
+            }
+          } yield maybeNewState
+        }
```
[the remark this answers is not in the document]
tbekas: I know that it may look strange, but I couldn't find any better way of dealing with the issue that I have there and the issue itself is quite complex. I will try to show the issue here, but if it will remain unclear we can have a session where it might be easier to explain it.

* Semaphore and the `access` are needed because there is at least one effectful computation inside of the update function, that cannot be moved outside of it  - pulling events from the storage in order to create the proposal.
  * If I use just the `access` without the semaphore, and there is more than one fiber that will do some effectful computations, I won't be able to revert these effects if their fiber fail to `set` the value (because only a single fiber will be able to successfully `set` the value).
  * If I use just the `semaphore` without the `access` I won't be able to perform the effects inside of the `.modify` function.
* The optional result`maybeNewState` is not actually needed but I added anyway it to lower the amount of locking with `.acquire` (in my case with `.permit`). There are 2 checks that decide if a state update is going to happen. The first check is happening before any call to `.permit`, and it's the call to `isStateUpdateRequired(state, resources)` in the `ConsensusManager`. Only if that will pass we then acquire the lock with `.permit` and then do the 2nd final check (we check again because before the lock the state may have been already updated by some other fiber). The 2nd check are the calls to: `state.facilitators.traverse(resources.peerDeclarations.get)` happening for each `case` in the big pattern-mach in `internalTryAdvanceConsensus`. This 2nd check returns `None` if the condition of having all necessary `peerDeclarations` isn't met, effectively skipping doing any effecful computations and also any state updates.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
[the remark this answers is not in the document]
marcinwadon: Personally, I have a rule of thumb that I try to write something without using for and if it's too complex or unreadable, I change that to a for comprehension.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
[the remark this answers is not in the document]
marcinwadon: I think we can keep in `info` information about selecting global snapshot only, and IMO there is no need to log all rounds or sth (I mean, it will be useful but for debugging)

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -0,0 +1,182 @@
...
+
+    private def internalTryFacilitateConsensus(key: Key, lastKeyAndArtifact: (Key, Artifact))(
+      maybeState: Option[ConsensusState[Key, Artifact]],
+      setter: Option[ConsensusState[Key, Artifact]] => F[Boolean]
+    ): F[Option[(Boolean, ConsensusState[Key, Artifact])]] =
+      maybeState match {
+        case None =>
+          for {
+            peers <- clusterStorage.getPeers
+            facilitators = selfId :: peers.filter(_.state == NodeState.Ready).map(_.id).toList
```
marcinwadon, line 90: filtering should be made like `NodeState.ready(peers)` (see: `node.scala`).
We had issues with such explicit filtering peers' states on the current mainnet, when we were adding new states (deriving from Ready), filters were not working.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStorage.scala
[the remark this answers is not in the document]
marcinwadon: Can you simplify it and wrap a "thunk" function into a state update transformation? In a similar way as NodeState is being changed. It will both simplify complexity and readability, and you'll get rid of semaphore locking.


```scala
stateTransition(from: ProposalMade, through: SelectingMajority, to: MajoritySelected) {
  ...
}
```

If there will be a second thread trying to make the same state shift, it will fail because of not-matching initial state:
```
threadA: from: ProposalMade -> SelectingMajority
threadB: from: ProposalMade, actual: SelectingMajority -> ...
```

marcinwadon (changes requested):

examples/empty/project/Dependencies.scala
```diff
@@ -1,17 +1,18 @@
 import sbt._
 
 object Dependencies {
+  val scope = Provided
```
TheMMaciek, line 4: Not sure extracting it to a val changes anything.

tbekas: True, I'm going to revert it or change to the `Runtime` scope. I was experimenting with different scopes, as I had trouble with `Provided` when running the `InputGenerator`.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
tbekas: Ok, I wasn't sure about this filtering either. Fixing it.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStorage.scala
```diff
@@ -0,0 +1,222 @@
...
+
+      def tryUpdateLastKeyAndArtifactWithCleanup(oldValue: (Key, Artifact), newValue: (Key, Artifact)): F[Boolean] =
+        lastKeyAndArtifactR.modify { maybeValue =>
+          if (maybeValue === oldValue.some)
+            (newValue.some, true)
+          else
+            (maybeValue, false)
+        }.flatTap(_ => cleanupStateAndResource(oldValue._1))
+
+      private def cleanupStateAndResource(key: Key): F[Unit] =
```
TheMMaciek, line 128: Is it compatible with `access` update used in `updateState` function. I mean no race condition possible?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -0,0 +1,182 @@
...
+  ): ConsensusStateUpdater[F, Key, Artifact] = new ConsensusStateUpdater[F, Key, Artifact] {
+
+    private val logger = Slf4jLogger.getLoggerFromClass(ConsensusStateUpdater.getClass)
+
+    def tryFacilitateConsensus(
+      key: Key,
+      lastKeyAndArtifact: (Key, Artifact)
+    ): F[Option[ConsensusState[Key, Artifact]]] =
+      consensusStorage
+        .updateState(key)(logIfUpdated(internalTryFacilitateConsensus(key, lastKeyAndArtifact)))
```
TheMMaciek, line 66: It seems like we are duplicating logs here. We have a log here and inside of the storage for the same case - success. I guess we could remove one of them.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStorage.scala
```diff
@@ -0,0 +1,222 @@
...
+
+      def getLastKeyAndArtifact: F[Option[(Key, Artifact)]] = lastKeyAndArtifactR.get
+
+      def tryUpdateLastKeyAndArtifactWithCleanup(oldValue: (Key, Artifact), newValue: (Key, Artifact)): F[Boolean] =
+        lastKeyAndArtifactR.modify { maybeValue =>
+          if (maybeValue === oldValue.some)
+            (newValue.some, true)
+          else
+            (maybeValue, false)
+        }.flatTap(_ => cleanupStateAndResource(oldValue._1))
```
TheMMaciek, line 126: The cleanup happens no matter if the update succeeded or not. Is that as intended?

modules/core/src/main/scala/org/tessellation/Main.scala
```diff
@@ -59,9 +63,20 @@ object Main
           cfg.snapshot
         )
 
+        _ <- GlobalSnapshotEventsPublisher
```
marcinwadon, line 66: don't you mind moving it into AppResources or some kind of nested dependency? I'd like to avoid what we have on the current Mainnet - the mess in the main initialisation flow because of hundreds of dependencies that are created in a single flow

modules/core/src/main/scala/org/tessellation/Main.scala
```diff
@@ -59,9 +63,20 @@ object Main
...
+          .publishEvents(queues.stateChannelOutput, snapshotTriggerStream, services.gossip)
+          .asResource
+
+        globalSnapshotConsensus <- GlobalSnapshotConsensus
```
marcinwadon, line 70: same here, it could go to services or programs, because indeed a consensus is kinda service

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -43,25 +41,32 @@ object GlobalSnapshotStorage {
...
       snapshotsRef.get.map(_.find(_.ordinal === ordinal))
 
-    def getLast: F[GlobalSnapshot] = snapshotsRef.get.map(_.head)
+    def getLast: F[Signed[GlobalSnapshot]] = snapshotsRef.get.map(_.head)
 
-    def getStateChannelSnapshotUntilOrdinal(
-      ordinal: SnapshotOrdinal
-    )(address: Address): F[Option[StateChannelSnapshotBinary]] =
+    /**
+      * Temporary implementation - until we agree on a representation and generation of the GlobalSnapshotInfo objects
```
marcinwadon, line 50: if it's going to stay in the codebase, can you mark it with `TODO: ` or `Note: `? it will be easier to track later

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -0,0 +1,107 @@
...
+
+      private val logger = Slf4jLogger.getLoggerFromClass[F](ConsensusManager.getClass)
+
+      def checkForTrigger(event: Event): F[Unit] =
+        consensusStorage.getLastKeyAndArtifact.flatMap(_.traverse(internalForCheckTrigger(_, event)).void)
+
+      def checkForStateUpdate(key: Key)(resources: ConsensusResources[Artifact]): F[Unit] =
+        consensusStorage.getState(key).flatMap(_.traverse(internalCheckForStateUpdate(key, _, resources)).void)
+
+      private def internalForCheckTrigger(lastKeyAndArtifact: (Key, Artifact), event: Event): F[Unit] =
```
marcinwadon, line 42: shouldn't it be `internalCheckForTrigger`?

marcinwadon (commented):
Changes look good :+1:

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
tbekas: Yes, absolutely

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
tbekas: Actually, I think I will remove this comment.

modules/core/src/main/scala/org/tessellation/Main.scala
tbekas: I wouldn't qualify it as a service yet, as it doesn't contain any methods to call, but in the future, it may become a service, as we may add some methods that will get called when we trigger/complete redownload. All in all, I think it won't hurt if I will add it to services already.

modules/core/src/main/scala/org/tessellation/Main.scala
tbekas: So that's an interesting case. There's only a single method that needs to get called upon the initialization so it's very similar to the `Daemon` that has only the `start` method. Maybe I should represent it then as a `Daemon`?

modules/core/src/main/scala/org/tessellation/Main.scala
```diff
@@ -43,25 +42,23 @@ object Main
...
-        services <- Services.make[IO](sdkServices, queues).asResource
+        services <- Services.make[IO](sdkServices, queues, storages, sdk.nodeId, keyPair, cfg).asResource
         programs = Programs.make[IO](sdkPrograms, storages, services)
         validators = Validators.make[IO](cfg.snapshot)
         healthChecks <- HealthChecks
           .make[IO](storages, services, p2pClient, cfg.healthCheck, sdk.nodeId)
           .asResource
 
-        _ <- services.stateChannelRunner.initializeKnownCells.asResource
+        streams = Streams.make(storages, validators, queues, cfg)
```
marcinwadon, line 52: shouldn't it be created inside a Daemon.make? It's a daemon itself as you said today

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -0,0 +1,156 @@
...
+    private val tipSnapshotTriggerTypable = Typeable[TipSnapshotTrigger]
+
+    def consumeSignedMajorityArtifact(signedArtifact: Signed[GlobalSnapshotArtifact]): F[Unit] =
+      globalSnapshotStorage.save(signedArtifact)
+
+    def triggerPredicate(
+      last: (GlobalSnapshotKey, GlobalSnapshotArtifact),
+      event: GlobalSnapshotEvent
+    ): Boolean = event.toOption.flatMap(_.toOption).fold(false) {
+      case TipSnapshotTrigger(height) => last._2.height.nextN(heightInterval) === height
```
marcinwadon, line 48: to use `.eqv`I think you need to import `syntax.eq._` 🤔

modules/sdk/src/main/scala/org/tessellation/sdk/domain/Daemon.scala
```diff
@@ -1,5 +1,20 @@
...
 }
+
+object Daemon {
+
+  def spawn[F[_]: Spawn](thunk: F[Unit]): Daemon[F] = new Daemon[F] {
+    def start: F[Unit] = Spawn[F].start(thunk).void
+  }
+
+  def drainStream[F[_]: Concurrent](stream: Stream[F, Unit]): Daemon[F] =
+    spawn[F] { stream.compile.drain }
```
marcinwadon, line 19: Don't you think it's an overkill? We'll use .compile.drain in many places not related to Daemon, as well as many other operators that are repeated over the code.

modules/shared/src/main/scala/org/tessellation/shared/package.scala
```diff
@@ -44,7 +45,7 @@ package object shared {
     classOf[CommonRumorBinary] -> 323,
     classOf[Transaction] -> 324,
     classOf[TransactionReference] -> 325,
-    classOf[SubHeight] -> 326
```
marcinwadon, line 47: why it's removed? 🤔

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
tbekas: `syntax.order._` already covers stuff in `syntax.eq._`

modules/shared/src/main/scala/org/tessellation/shared/package.scala
tbekas: It's a `@newtype` now (`Refine` is registered).

modules/sdk/src/main/scala/org/tessellation/sdk/domain/Daemon.scala
tbekas: Yep, it could be unnecessary, I'll remove it.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStorage.scala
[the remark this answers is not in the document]
tbekas: I will follow it up in the next PR. We agreed that we can merge this PR with the `Semaphore` in place as it is and the non-blocking solution will come up soon.

modules/core/src/main/scala/org/tessellation/Main.scala
tbekas: Ok, I'm removing `Streams` altogether and moving stream creation to `Daemons`.

marcinwadon (approved):

PR #189, opened 2022-02-23
Push L1 data to L0

kpudlik:


modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/SnapshotPreconditionsValidator.scala
```diff
@@ -22,9 +22,9 @@ class SnapshotPreconditionsValidator[F[_]](config: SnapshotConfig) {
 
   def validateMinTipHeight(
     previousSnapshotHeight: Height,
-    tips: Tips
+    tips: Option[Tips]
   )(implicit N: Next[Height]): ValidationResult[Height] = {
-    val minTipHeight = tips.value.toList.minBy(_.height).height
+    val minTipHeight = tips.map(_.value.toList.minBy(_.height).height).getOrElse(Height(0))
```
marcinwadon, line 27: afaik there is `Height.Min`, so you can use it for `getOrElse`. If there is not, you can create one

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -125,6 +126,16 @@ class StateChannel[F[_]: Async: KryoSerializer: SecurityProvider: Random](
       storages.block.store(fb.hashedBlock).handleErrorWith(e => logger.debug(e)("Block storing failed."))
     }
 
+  private val sendBlockToL0: Pipe[F, FinalBlock, Unit] =
+    _.evalMap { fb =>
+      (for {
+        tips <- storages.block
+          .pullTips(appConfig.tips.maximumTipsCount)
+        // l0Peer <- storages.l0Cluster.getPeers.map(Random.shuffle).map(_.head)
+        // req <- p2PClient.l0DAGCluster.sendL1Output(L1Output(fb.hashedBlock.signed, tips))(l0Peer)
```
marcinwadon, line 135: are you going to continue on that in this PR?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/http/p2p/L0DAGClusterClient.scala
```diff
@@ -0,0 +1,29 @@
...
+object L0DAGClusterClient {
+
+  def make[F[_]: Async: SecurityProvider](client: Client[F]): L0DAGClusterClient[F] =
+    new L0DAGClusterClient[F] {
+
+      def sendL1Output(output: L1Output): PeerResponse[F, Unit] =
+        PeerResponse[F, Unit]("block", POST)(client) { (req, c) =>
+          c.successful(req.withEntity(output)).void
+        }
+    }
```
marcinwadon, line 28: I think you can return `PeerResponse[F, Boolean]` and handle `false` later if needed (let's say to resend block or sth)

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -125,6 +126,16 @@ class StateChannel[F[_]: Async: KryoSerializer: SecurityProvider: Random](
...
+      } yield ()).handleErrorWith(e => logger.warn(e)("Sending block to L0 failed."))
```
marcinwadon, line 136: if request fails, you made `successful` so it will not throw an error but return a boolean whether succeeded or no. check out my second comment, I'd recommend to handle boolean and `handleErrorWith` will catch the error from the flow, not from the API

marcinwadon (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -125,6 +126,16 @@ class StateChannel[F[_]: Async: KryoSerializer: SecurityProvider: Random](
       storages.block.store(fb.hashedBlock).handleErrorWith(e => logger.debug(e)("Block storing failed."))
     }
 
+  private val sendBlockToL0: Pipe[F, FinalBlock, Unit] =
+    _.evalMap { fb =>
+      (for {
+        tips <- storages.block
+          .pullTips(appConfig.tips.maximumTipsCount)
```
TheMMaciek, line 133: Maybe we should only send the lowest tip height? All depends on how this information will be used on the L0.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
TheMMaciek, line 135: I think if we are sending additional data (tips) we need to sign the `L1Output` and then check the signature on the L0 side. Maybe signed request would be enough (depending on whether L0 checks signatures or not), but we could also send the Signed object.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/http/p2p/L0DAGClusterClient.scala
TheMMaciek: Regarding resending. We could change the flow and mark the blocks as sent or not in the storage (similarly to how we store information about whether block is a tip). Then we could pull the blocks to send (find not sent blocks signed by our id), send them and only after success mark them as sent. Then whatever happens during sending we may just log it and pull again, the failed blocks should still be there.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/SnapshotPreconditionsValidator.scala
kpudlik: Fair point!

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
kpudlik: Yes

kpudlik: Sure, So we will have `Signed[L1Output]` with `Signed[DAGBlock]` inside

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/SnapshotPreconditionsValidator.scala
```diff
@@ -11,20 +11,23 @@ import org.tessellation.dag.domain.block.{L1Output, Tips}
...
   def validateMinTipHeight(
     previousSnapshotHeight: Height,
-    tips: Tips
+    tips: Option[Tips]
   )(implicit N: Next[Height]): ValidationResult[Height] = {
-    val minTipHeight = tips.value.toList.minBy(_.height).height
+    val minTipHeight = tips.map(_.value.toList.minBy(_.height).height).getOrElse(Height.MinValue)
     val nextHeight = previousSnapshotHeight.next
 
     Validated.condNel(
```
tbekas, line 33: In the line below `34` you can use `Order[Height]` instance to compare two Heights with `>` operator. You will need to import `cats.syntax.order._`.

tbekas (approved):
Looks good to me, but I would like the other guys to review these changes as well before you're going to merge them.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -125,6 +128,25 @@ class StateChannel[F[_]: Async: KryoSerializer: SecurityProvider: Random](
       storages.block.store(fb.hashedBlock).handleErrorWith(e => logger.debug(e)("Block storing failed."))
     }
 
+  private val sendBlockToL0: Pipe[F, FinalBlock, Unit] =
+    _.evalMap { fb =>
+      for {
+        tips <- storages.block
+          .pullTips(appConfig.tips.minimumTipsCount)
```
TheMMaciek, line 135: I think you could send just a single tip height to L0 instead of 2 block references.

TheMMaciek: I would even create a function in `BlockStorage` for getting `getMinTipHeight: [Option[Height]]`

TheMMaciek (approved):

marcinwadon (approved):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
kpudlik: I will talk about it with Tomek.

PR #190, opened 2022-02-24
GlobalSnapshot persistent storage

marcinwadon:


modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -65,3 +73,63 @@ object GlobalSnapshotStorage {
...
+  ): F[GlobalSnapshotStorageV2[F]] = {
+
+    def cacheProcess =
+      Stream
+        .fromQueueUnterminated(cacheQueue)
+        .evalTap { snapshot =>
+          headRef.set(snapshot) >>
+            cache(snapshot.ordinal).set(snapshot.some) >>
+            NonNegLong
+              .from(snapshot.ordinal.value.value - inMemoryCapacity.value)
```
tbekas, line 106: I think that if we calculate the next `SnapshotOrdinal` using `Next[A]`, we should also calculate the previous values with `Previous[A]` or `PartialPrevious[A]` (the latter one makes more sense for types like `NonNegLong`). Maybe also a syntax `previousN`/`partialPreviousN` could be added, analogous to [nextN](https://github.com/Constellation-Labs/tessellation/blob/c9d24567f4f7ac24e4b85b73586065384b2e4d4d/modules/shared/src/main/scala/org/tessellation/ext/cats/syntax/NextSyntax.scala#L20).

modules/core/src/main/scala/org/tessellation/domain/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -16,3 +16,13 @@ trait GlobalSnapshotStorage[F[_]] {
   )(address: Address): F[Option[StateChannelSnapshotBinary]]
 
 }
+
+trait GlobalSnapshotStorageV2[F[_]] {
+
+  def prepend(snapshot: GlobalSnapshot): F[Unit]
```
tbekas, line 22: I think that `prepend` should return `F[Boolean]` so that we can tell if the new snapshot passed a validation against the current head.

marcinwadon: If we're going to check against Next/Prev, then yes

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -65,3 +73,68 @@ object GlobalSnapshotStorage {
...
+    def mkHeadRef = Ref.of[F, GlobalSnapshot](genesis)
+    def mkCache = MapRef.ofSingleImmutableMap[F, SnapshotOrdinal, GlobalSnapshot](Map.empty)
+    def mkOffloadQueue = Queue.unbounded[F, SnapshotOrdinal]
+
+    (mkCacheQueue, mkHeadRef, mkCache, mkOffloadQueue).mapN { make(_, _, _, _, inMemoryCapacity) }.flatten
+  }
+
+  def make[F[_]: Async: KryoSerializer](
+    cacheQueue: Queue[F, GlobalSnapshot],
+    headRef: Ref[F, GlobalSnapshot],
```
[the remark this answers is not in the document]
marcinwadon: No, I've started with SignallingRef but the documentation clearly states that it can skip some values and if all values matters, use Queue or Channel. 

Discrete returns only the last element on SignallingRef, continuous mode returns last element on each Pull (until changed, aka 1, 1, 1, 2, 2, 2, 2, 3, ...)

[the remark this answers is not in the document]
tbekas: Ok, good to know that!

modules/core/src/main/scala/org/tessellation/domain/snapshot/GlobalSnapshotStorage.scala
tbekas: I think we want to have this kind of validation.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -65,3 +78,85 @@ object GlobalSnapshotStorage {
...
+            case None => MonadThrow[F].raiseError[Unit](new Throwable("Unexpected state!"))
+          }
+        }
+        .compile
+        .drain
+
+    Spawn[F].start { offloadProcess } >> Spawn[F].start { cacheProcess }.map { _ =>
+      new GlobalSnapshotStorageV2[F] {
+        def prepend(snapshot: GlobalSnapshot) =
+          head.map { _.ordinal.next === snapshot.ordinal }.flatTap {
```
TheMMaciek, line 144: I think we can use `ifM`

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -65,3 +78,85 @@ object GlobalSnapshotStorage {
...
+
+  def make[F[_]: Async](
+    globalSnapshotLocalFileStorage: GlobalSnapshotLocalFileStorage[F],
+    genesis: GlobalSnapshot,
+    inMemoryCapacity: PosInt
+  ): F[GlobalSnapshotStorageV2[F]] = {
+    def mkCacheQueue = Queue.unbounded[F, GlobalSnapshot]
+    def mkHeadRef = Ref.of[F, GlobalSnapshot](genesis)
+    def mkCache = MapRef.ofSingleImmutableMap[F, SnapshotOrdinal, GlobalSnapshot](Map.empty)
+    def mkOffloadQueue = Queue.unbounded[F, SnapshotOrdinal]
```
TheMMaciek, line 92: Maybe we should store `Signed` or even `Hashed` `GlobalSnapshot`, of course if someone asks for it we should return `Signed[GlobalSnapshot]` but we could internally store it as `Hashed` and then we can compare hashes during update (if lastSnapshotHash matches)

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
marcinwadon: Why? I don't want to flatMap but flatTap

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -65,3 +78,85 @@ object GlobalSnapshotStorage {
...
+
+    def offloadProcess =
+      Stream
+        .fromQueueUnterminated(offloadQueue)
+        .evalTap { ordinal =>
+          cache(ordinal).get.flatMap {
+            case Some(snapshot) =>
+              globalSnapshotLocalFileStorage.write(snapshot).rethrowT.handleErrorWith { e =>
+                Logger[F].error(e)(s"Failed writing global snapshot to disk! Snapshot ordinal=${snapshot.ordinal.show}")
+              } >> cache(ordinal).set(none)
```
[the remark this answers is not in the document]
marcinwadon: I'll make a ticket then :+1:

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
TheMMaciek: Oh, ok you are returning it to the caller.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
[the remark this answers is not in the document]
marcinwadon: CONST-426

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -1,67 +1,112 @@
...
+      headRef.get.flatMap { h =>
+        cache(h.ordinal).set(h.some)
       }
-  }
+    }.map { _ =>
+      new GlobalSnapshotStorage[F] {
+        def prepend(snapshot: GlobalSnapshot) =
+          headRef.modify { current =>
+            isNextSnapshot(current, snapshot) match {
+              case Left(e) =>
```
tbekas, line 81: I think this should be `e.raiseError[F]` instead of logging.

modules/core/src/main/scala/org/tessellation/config/types.scala
```diff
@@ -37,8 +38,9 @@ object types {
   )
 
   case class SnapshotConfig(
-    storedSnapshotPath: Path,
-    fallbackTriggerTimeout: FiniteDuration
+    globalSnapshotPath: Path,
+    fallbackTriggerTimeout: FiniteDuration,
+    inMemoryCapacity: PosInt
```
tbekas, line 43: I would use `NonNegLong` / `NonNegInt` here.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -1,67 +1,112 @@
...
+                (current, Logger[F].error(e)(s"Unexpected error while checking next snapshot chain!").map(_ => false))
+              case Right(isNext) if isNext =>
+                def run =
+                  cache(snapshot.ordinal).set(snapshot.some) >>
+                    snapshot.ordinal
+                      .partialPreviousN(NonNegLong.unsafeFrom(inMemoryCapacity.toLong))
```
tbekas, line 87: I think we shouldn't use `unsafe` on refined (maybe tests could be an exception to the rule).

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotLocalFileStorage.scala
```diff
@@ -0,0 +1,31 @@
...
+import cats.effect.Async
+import cats.syntax.flatMap._
+
+import org.tessellation.dag.snapshot.{GlobalSnapshot, SnapshotOrdinal}
+import org.tessellation.kryo.KryoSerializer
+import org.tessellation.storage.LocalFileStorage
+
+import fs2.io.file.Path
+
+final class GlobalSnapshotLocalFileStorage[F[_]: Async: KryoSerializer] private (path: Path)
```
tbekas, line 14: I would rename it to `GlobalSnapshotLocalFileSystemStorage` or `GlobalSnapshotLocalFSStorage` because currently, someone could think that the whole storage is in a single file, like DB file (while it's in the directory).

tbekas (changes requested):

modules/core/src/main/scala/org/tessellation/config/types.scala
marcinwadon: we cannot store `0` snapshot in memory, it must be at least 1

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
marcinwadon: `inMemoryCapacity` is a positive long, so it's safe to convert it into non-negative long. unfortunately, there is no embedded function that will convert from more strict to more general type

modules/core/src/main/scala/org/tessellation/config/types.scala
tbekas: Why not? I think it would effectively mean turning off the in-memory cache.

marcinwadon: Then, we will end-up with asynchronous storage again

tbekas: Not really if the removal from the cache happens after the successful storage write.

tbekas (approved):

PR #191, opened 2022-02-28
Genesis balances

marcinwadon:


tbekas (commented):
Currently, the nodes started with `run-validator` won't be able to download/receive this data. I think we should introduce it together with either: GlobalSnapshot download or awaiting for GlobalSnapshot broadcast.

modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/GlobalSnapshot.scala
```diff
@@ -19,6 +21,22 @@ case class GlobalSnapshot(
...
+  def mkGenesis(balances: Map[Address, Balance]) =
+    GlobalSnapshot(
+      SnapshotOrdinal.MinValue,
+      Height.MinValue,
+      SubHeight.MinValue,
+      Hash("0"),
+      Set.empty,
+      balances,
+      Map.empty,
+      NonEmptyList.of(PeerId(Hex("peer1")))
```
TheMMaciek, line 40: Maybe we can pass the genesis node id?

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -71,39 +72,43 @@ object GlobalSnapshotStorage {
...
         def get(ordinal: SnapshotOrdinal): F[Option[Signed[GlobalSnapshotArtifact]]] = cache(ordinal).get.flatMap {
           case Some(s) => s.some.pure[F]
-          case None    => globalSnapshotLocalFileSystemStorage.read(ordinal).rethrowT.map(_.some)
+          case None =>
+            globalSnapshotLocalFileSystemStorage
+              .read(ordinal)
+              .rethrowT
+              .map(_.some)
+              .recover {
+                case _: NoSuchFileException => none[Signed[GlobalSnapshot]]
```
tbekas, line 110: As already mentioned in the online conversation, this should be handled by an abstraction.

tbekas (changes requested):

PR #197, opened 2022-03-01
Fix genesis balances by registering BigInteger in Kryo

marcinwadon:


marcinwadon:
Closing - the issue is more complex and because of the inconsistency of hashing BigInt, I'm closing the pull request and we will introduce another solution.

PR #202, opened 2022-03-03
Use sbt dotenv plugin to fix IntelliJ SBT issues.

ryle-goehausen:
Caused by Github actions sbt plugin, couldn't find easy resolution to this apart from using an env file due to environment.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,6 +8,8 @@ ThisBuild / organizationName := "tessellation"
 ThisBuild / evictionErrorLevel := Level.Warn
 ThisBuild / scalafixDependencies += Libraries.organizeImports
 
+ThisBuild  / envFileName := ".sbtenv"
+
 resolvers += Resolver.sonatypeRepo("snapshots")
 
 val scalafixCommonSettings = inConfig(IntegrationTest)(scalafixConfigSettings(IntegrationTest))
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -6,4 +6,5 @@ addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.30")
 addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "1.1.0")
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.10.0")
 addSbtPlugin("com.codecommit" % "sbt-github-packages" % "0.5.2")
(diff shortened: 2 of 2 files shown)


build.sbt
```diff
@@ -8,6 +8,8 @@ ThisBuild / organizationName := "tessellation"
 ThisBuild / evictionErrorLevel := Level.Warn
 ThisBuild / scalafixDependencies += Libraries.organizeImports
 
+ThisBuild  / envFileName := ".sbtenv"
```
marcinwadon, line 11: We have used `sbt-dotenv` previously but it conflicted with nix that we introduced in #138.
If you want to, you can change your task configuration to run the node from IntelliJ and it works without `sbt-dotenv` plugin, we're already using it that way.

Anyway, if you would still want to add `sbt-dotenv`, we already have the `.env` file so I'd use this one by default.

marcinwadon (changes requested):

build.sbt
ryle-goehausen: So the main problem I've been having with IntelliJ is with the code auto-complete. I can't get sbt to work without the environment variable for my github related token. Tried several configuration options but was unable to get it working properly. I have a .sbtenv file that only contains a personal github api token that is specific to me and not checked in here. That was the easiest way to fix IJ and it doesn't appear to be breaking any unit tests. 

If you have any more details on how to fix SBT in IJ I spent a while trying to get this to work, several suggested solutions from google just didn't work and I gave up and used this plugin from a recommendation. Not sure what you mean by changing task configuration? I tried some configs for SBT but this was only way I could get it to pass. To be clear this isn't required when I do CLI stuff since SBT works fine there, it's only IntelliJ having the issue.

kpudlik: You can try adding token there:

image


Otherwise I suggest you to put it into your `.bashrc`/`.zshrc` (depending on which shell you use). Assuming that you imported project as sbt (not BSP) then it should work.

ryle-goehausen: Its already loaded in my bash, and works totally fine on the command line again. I've tried re-importing the project and specifying sbt explicitly, and adding this path variable, but it still doesn't work. This PR is the only thing I can use to get my build environment working, and doesn't really cost very much to add since it's an optional value.

image

marcinwadon:
Closing as it has been fixed on a meeting

PR #203, opened 2022-03-03
Simplify consensus state update

tbekas:


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -58,52 +58,67 @@ object ConsensusStateUpdater {
...
+      fn: Option[ConsensusState[Key, Artifact]] => F[Option[(ConsensusState[Key, Artifact], F[Unit])]]
+    ): ModifyStateFn[ModifyStateResult] =
+      maybeState =>
+        fn(maybeState).map(_.map {
+          case (state, effect) => (state.some, (state.some, effect))
+        })
+
+    private def evalEffect(result: Option[ModifyStateResult]): F[Option[ConsensusState[Key, Artifact]]] =
+      result.flatTraverse {
+        case (maybeState, effect) => effect >> maybeState.pure[F]
```
marcinwadon, line 94: `effect.map(_ => maybeState)`

marcinwadon: what about error in `effect`? shouldn't it be handled here?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -58,52 +58,67 @@ object ConsensusStateUpdater {
...
         case None =>
           for {
             peers <- clusterStorage.getPeers
             facilitators = selfId :: NodeState.ready(peers).map(_.id).toList
             state = ConsensusState(key, facilitators, lastKeyAndArtifact, Facilitated[Artifact]())
-            result <- setter(state.some)
             bound <- consensusStorage.getUpperBound
-            _ <- gossip.spread(ConsensusFacility(key, bound))
-          } yield (result, state).some
+            effect = gossip.spread(ConsensusFacility(key, bound))
```
marcinwadon, line 112: `gossip.spread` raise errors sometimes (see above comment about error handling)

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
tbekas: So there's very little we can do at this point if an error happens, except for logging the error, and we already do it as a part of gossip error handling. This effect is evaluated after the state transition is "committed" and some effects have already taken place. The only sensible error handling would be to voluntarily leave the consensus, which isn't implemented and possibly will not have to be implemented if the node gets kicked out by the health-check anyway.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
tbekas: As stated above, there's very little we can do when evaluating an effect after the state transition. Once the health-check is in place the node will get kicked out anyway.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
marcinwadon: Didn't we agree yesterday that any error in effect during state transition should switch state to its previous/initial state?

tbekas: As stated above
>This effect is evaluated after the state transition is "committed"

marcinwadon (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -58,52 +58,67 @@ object ConsensusStateUpdater {
 
     private val logger = Slf4jLogger.getLoggerFromClass(ConsensusStateUpdater.getClass)
 
+    /** Pair of (maybeUpdatedState, effect) */
```
kpudlik, line 61: eventually alias:

```scala
type MaybeUpdatedState = Option[ConsensusState[Key, Artifact]]
```

it is used in many places so putting comment here isn't helpful as we need to keep scrolling to that comment. I do not have any better idea ;<

tbekas: You're right. It would be better to use the alias you've suggested since it's used more (I would just rename it to `MaybeState` since it's also used also before the update, not just after the update). Then we could remove the `MaybeStateResult` and just keep it as `(MaybeState, F[Unit])` since it's not used as much.

tbekas: I will add this in another PR.

PR #205, opened 2022-03-04
Use F[Option instead of EitherT in local filesystem storage

marcinwadon:


modules/shared/src/main/scala/org/tessellation/storage/LocalFileSystemStorage.scala
```diff
@@ -36,38 +38,40 @@ abstract class LocalFileSystemStorage[F[_]: KryoSerializer, A <: AnyRef: ClassTa
     F.delay { (a / fileName).exists }
   }
 
-  def read(fileName: String): EitherT[F, Throwable, A] = readBytes(fileName).flatMap { bytes =>
-    EitherT.fromEither { KryoSerializer[F].deserialize[A](bytes) }
-  }
+  def read(fileName: String): F[Option[A]] =
+    readBytes(fileName)
+      .map(_.flatMap { bytes =>
+        KryoSerializer[F].deserialize[A](bytes).toOption
```
TheMMaciek, line 44: So we won't be able to tell whether the file was not found or bytes couldn't be deserialized? Maybe we could raise the `Either` to `F`, to get the error whenever some error occurred and only get `None` when file wasn't there and `Some` when it was there and got deserialized successfully.

modules/shared/src/main/scala/org/tessellation/storage/LocalFileSystemStorage.scala
```diff
@@ -36,38 +40,40 @@ abstract class LocalFileSystemStorage[F[_]: KryoSerializer, A <: AnyRef: ClassTa
...
   }
 
-  def read(fileName: String): EitherT[F, Throwable, A] = readBytes(fileName).flatMap { bytes =>
-    EitherT.fromEither { KryoSerializer[F].deserialize[A](bytes) }
-  }
+  def read(fileName: String): F[Option[A]] =
+    readBytes(fileName)
+      .flatMap(_.map { bytes =>
+        EitherT.fromEither { KryoSerializer[F].deserialize[A](bytes) }.rethrowT
+      }.sequence)
```
tbekas, line 47: .flatMap(`_.traverse { bytes => bytes.fromBinaryF }`)

kpudlik (approved):

tbekas (approved):

PR #210, opened 2022-03-09
Change BigInt into Long for balance

marcinwadon:


modules/shared/src/main/scala/org/tessellation/schema/transaction.scala
```diff
@@ -82,7 +92,7 @@ object transaction {
           Seq(
             source.coerce,
             destination.coerce,
-            amount.coerce.value.toString(16),
+            amount.coerce.value.toHexString,
```
marcinwadon, line 95: @TheMMaciek Please confirm it's correct

marcinwadon: anyway, I think it is:
```scala
scala> 2L.toHexString
val res0: String = 2

scala> BigInt(2L).toString(16)
val res1: String = 2

scala> Long.MaxValue.toHexString
val res2: String = 7fffffffffffffff

scala> BigInt(Long.MaxValue).toString(16)
val res3: String = 7fffffffffffffff
```

modules/shared/src/main/scala/org/tessellation/schema/balance.scala
```diff
@@ -1,34 +1,51 @@
...
 import io.getquill.MappedEncoding
 
 object balance {
 
   @derive(decoder, encoder, eqv, show)
   @newtype
-  case class Balance(value: NonNegBigInt)
+  case class Balance(value: NonNegLong) {
+
+    def plus(that: Balance): Option[Balance] = {
```
tbekas, line 26: Is that correct that we add Balances? Shouldn't we add/subtract `Amount` to/from `Balance`?

kpudlik: Then we need `Balance` -> `Amount` conversion. I guess it won't work the opposite way.

marcinwadon: I think so. It's like a semigroup on balance (but semigroup isn't partial so I cannot use semigroup itself).

marcinwadon: I can change it into `def plus(that: Amount)` but then I'd have to have conversion from `Balance` to `Amount` which seems a little bit weird 🤔 
IMO both ways are correct. wdyt?

marcinwadon: As agreed, moved to separate task

tbekas (dismissed):
We need to address the issue of adding balances in the future.

modules/shared/src/main/scala/org/tessellation/schema/package.scala
```diff
@@ -54,15 +54,33 @@ trait OrphanInstances {
   implicit val posBigIntShow: Show[PosBigInt] =
```
TheMMaciek, line 54: Do we need the show/decoder/encoder for Pos/NonNeg BigInts still?

modules/shared/src/main/scala/org/tessellation/schema/transaction.scala
TheMMaciek: I guess. If it gives the same results. Maybe the best would be to look at the implementation and if they overlap somewhere underneath.

modules/shared/src/main/scala/org/tessellation/schema/balance.scala
TheMMaciek: I would build it around adding `Long` simply. Converting `Fee` or `Amount` to `Balance` just to perform adding doesn't look great. Anyway you are still getting `Option[Balance]` so you can pass `Long` and handle everything inside. Before adding or subtracting unwrap the type. That's what I think.

TheMMaciek: We could also be fine with just one function then, like combine instead of two separate for adding and subtracting.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
```diff
@@ -67,7 +66,7 @@ object BlockService {
         val all = sources.combine(destinations)
         val addressBalanceChanges = all.map {
           case (address, txs) =>
-            val change = txs.foldLeft(BigInt(0L))(
+            val change = txs.foldLeft(0L)(
               (change, tx) =>
                 tx match {
                   case Transaction(`address`, _, amount, fee, _, _) => change - amount.coerce - fee.coerce
```
TheMMaciek, line 72: If we are worried about overflowing `Long` it could theoretically happen here during `change` calculation as well.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
```diff
@@ -81,13 +80,25 @@ object BlockService {
 
         addressBalanceChanges.toList.traverse {
           case (address, balanceChange) =>
-            for {
-              current <- addressStorage.getBalance(address)
-              newBalance <- refineV[NonNegative]
-                .apply(current.coerce + balanceChange)
-                .bimap(BalanceNegative, Balance(_))
+            val positive = balanceChange > 0
```
TheMMaciek, line 83: I'm not sure if this approach with figuring out if the change is positive or negative, then constructing the `NonNegLong` and then wrapping it in `Balance` just to add or subtract is the way to go. I would rather base everything on simple types and have one point of conversion -> if after all the adding it's still correct lets get Some(Balance) otherwise None.

TheMMaciek: We could even go with `BigInt` for adding subtracting and so one, and pass the value to a function `Balance.fromBigInt()` and have this single function decide whether that is still a correct balance or not.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
```diff
@@ -81,13 +80,25 @@ object BlockService {
...
-            } yield (address, newBalance)
+
+            addressStorage
+              .getBalance(address)
+              .flatMap { currentBalance =>
+                getDiff.map(Balance(_)).map { diff =>
+                  val updated =
+                    if (positive) currentBalance.plus(diff) else currentBalance.minus(diff)
+
+                  updated.getOrElse(Balance.empty)
```
TheMMaciek, line 97: Are you defaulting to 0 if balance gets negative? For me it should fail fast and throw an error, we can't have such a situation at this point. If we do we have a flaw in our code that needs to come up.

TheMMaciek (changes requested):
At least the defaulting to 0 should be fixed.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
marcinwadon: fixed

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
marcinwadon: fixed

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
marcinwadon: We agreed on Slack to keep a single type and avoid conversions

modules/shared/src/main/scala/org/tessellation/schema/package.scala
marcinwadon: I don't think so. Removed :+1:

kpudlik (approved):

modules/shared/src/main/scala/org/tessellation/schema/balance.scala
```diff
@@ -1,39 +1,63 @@
...
+    def minus(that: Amount): Either[BalanceOutOfRange, Balance] =
+      NonNegLong.from(value.value - that.value.value).bimap(BalanceOutOfRange, Balance(_))
+  }
 
   object Balance {
-    val empty: Balance = Balance(NonNegBigInt.unsafeFrom(BigInt(0)))
+
+    val empty: Balance = Balance(NonNegLong(0L))
+
+    implicit def toAmount(balance: Balance): Amount = Amount(balance.value)
```
TheMMaciek, line 45: It's not used and I guess not needed. We won't be adding/subtracting balances right?

modules/shared/src/main/scala/org/tessellation/schema/height.scala
```diff
@@ -29,16 +29,18 @@ object height {
...
 
-  @derive(encoder, decoder, show, eqv)
+  @derive(encoder, decoder, eqv, show)
   @newtype
   case class SubHeight(value: NonNegLong)
 
   object SubHeight {
     val MinValue: SubHeight = SubHeight(NonNegLong.MinValue)
 
+    implicit val order: Order[SubHeight] = Order[Long].contramap(_.value.value)
```
TheMMaciek, line 39: It's not related to this PR, I think we shouldn't mix in stuff from outside of the scope in general. If we made a bug by doing that finding it would be a bit harder.

modules/shared/src/test/scala/org/tessellation/schema/BalanceSuite.scala
```diff
@@ -0,0 +1,43 @@
...
+  test("plus returns Left when sum overflows the Long range") {
+    val balance = Balance(NonNegLong(Long.MaxValue))
+    val amount = Amount(NonNegLong(2L))
+
+    not(exists(balance.plus(amount))(_ => success))
+  }
+
+  test("minus should subtract given amount from balance") {
+    val balance = Balance(NonNegLong(30L))
+    val amount = Balance(NonNegLong(1L))
```
TheMMaciek, line 30: Shouldn't this be `Amount` class as well?

modules/shared/src/test/scala/org/tessellation/schema/BalanceSuite.scala
```diff
@@ -0,0 +1,43 @@
...
+
+    val expected = Balance(NonNegLong(29L))
+
+    exists(balance.minus(amount))(expect.same(_, expected))
+  }
+
+  test("minus returns Left when amount lowers balance below zero") {
+    val balance = Balance(NonNegLong(1L))
+    val amount = Balance(NonNegLong(30L))
```
TheMMaciek, line 39: Here as well?

modules/shared/src/test/scala/org/tessellation/schema/BalanceSuite.scala
```diff
@@ -0,0 +1,43 @@
...
+
+    not(exists(balance.minus(amount))(_ => success))
```
TheMMaciek, line 41: Looks a bit weird maybe you could just assert that a value is equal to `Left`?

modules/shared/src/main/scala/org/tessellation/schema/balance.scala
marcinwadon: One use case would be to calculate total supply (we need to add all balances together). I'd keep that anyway

modules/shared/src/main/scala/org/tessellation/schema/height.scala
marcinwadon: it was required when I was changing the code. after all changes it looks like unused indeed. removed

modules/shared/src/test/scala/org/tessellation/schema/BalanceSuite.scala
marcinwadon: I can do that. anyway, `exists` work on `foldable` type (so on Either as well), and `not(exists` is imo self explanatory.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
```diff
@@ -62,35 +58,29 @@ object BlockService {
...
-          case (address, balanceChange) =>
-            for {
-              current <- addressStorage.getBalance(address)
-              newBalance <- refineV[NonNegative]
-                .apply(current.coerce + balanceChange)
-                .bimap(BalanceNegative, Balance(_))
-                .liftTo[F]
-            } yield (address, newBalance)
+        def applyTransactions(balance: Balance, address: Address): Either[BalanceOutOfRange, Balance] =
+          transactions.foldLeft(balance.asRight[BalanceOutOfRange]) { (acc, tx) =>
```
TheMMaciek, line 67: You are going through all the transactions just to extract balances for single addresses. Maybe there is something like `foldF` where you could traverse `Map[Address, Transaction]` and fold transactions with initial balance for every address?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
```diff
@@ -62,35 +58,29 @@ object BlockService {
...
+            tx match {
+              case Transaction(`address`, _, amount, fee, _, _) => acc.flatMap(_.minus(amount)).flatMap(_.minus(fee))
+              case Transaction(_, `address`, amount, _, _, _)   => acc.flatMap(_.plus(amount))
```
TheMMaciek, line 70: The problem I see with this adding/subtracting is that it is sensitive to an order in which transactions are folded. Currently that shouldn't be a problem because the validation will not pass transactions if we don't have sufficient balance up front, but in the future if we allow the received transactions to influence our balance during block validation this will no longer be valid. And I guess we may wan't to do this at some point because it seems a valid scenario. So we need to remember that.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
marcinwadon: I can use `filter` upfront, but I don't see any optimization in that. anyway I'd need to iterate through all transactions

TheMMaciek (approved):

PR #212, opened 2022-03-16
Get snapshot from storage by either ordinal or hash

marcinwadon:


modules/shared/src/main/scala/org/tessellation/storage/LocalFileSystemStorage.scala
```diff
@@ -66,6 +67,11 @@ abstract class LocalFileSystemStorage[F[_]: KryoSerializer, A <: AnyRef: ClassTa
       }
       .void
 
+  def link(fileName: String, to: String) =
```
TheMMaciek, line 70: We have type annotation on other methods, so we can add it here as well.

modules/core/src/test/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorageSuite.scala
```diff
@@ -130,4 +130,20 @@ object GlobalSnapshotStorageSuite extends MutableIOSuite with Checkers {
       }
     }
   }
+
+  test("get - should return snapshot by hash") { res =>
```
TheMMaciek, line 134: I think you are only testing the refs here because of the `inMemoryCapacity` config. Maybe we should test writing to disk as well - fetch both by ordinal and hash and then compare?

marcinwadon: I can do such tests but in a separate pull request (related to disk storage) as it requires not only this but other tests as well

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -93,10 +108,19 @@ object GlobalSnapshotStorage {
 
         def head: F[Option[Signed[GlobalSnapshotArtifact]]] = headRef.get
 
-        def get(ordinal: SnapshotOrdinal): F[Option[Signed[GlobalSnapshotArtifact]]] = cache(ordinal).get.flatMap {
-          case Some(s) => s.some.pure[F]
-          case None    => globalSnapshotLocalFileSystemStorage.read(ordinal)
-        }
+        def get(ordinal: SnapshotOrdinal): F[Option[Signed[GlobalSnapshotArtifact]]] =
+          ordinalCache(ordinal).get.flatMap {
+            _.fold(none[Signed[GlobalSnapshotArtifact]].pure[F])(hash => hashCache(hash).get)
```
TheMMaciek, line 113: Maybe you can reuse `get(hash: Hash)` function?

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -93,10 +108,19 @@ object GlobalSnapshotStorage {
...
         def head: F[Option[Signed[GlobalSnapshotArtifact]]] = headRef.get
 
-        def get(ordinal: SnapshotOrdinal): F[Option[Signed[GlobalSnapshotArtifact]]] = cache(ordinal).get.flatMap {
-          case Some(s) => s.some.pure[F]
-          case None    => globalSnapshotLocalFileSystemStorage.read(ordinal)
-        }
+        def get(ordinal: SnapshotOrdinal): F[Option[Signed[GlobalSnapshotArtifact]]] =
+          ordinalCache(ordinal).get.flatMap {
+            _.fold(none[Signed[GlobalSnapshotArtifact]].pure[F])(get(_))
+          }.flatMap {
```
TheMMaciek, line 114: Isn't this part already done in `get(hash: Hash)`? 🤔 I think you don't need to flatmap it again, right?

marcinwadon: yes, you're right

TheMMaciek (approved):
Seems ok as far as I can tell looking at the changes :)

TheMMaciek (approved):

PR #213, opened 2022-03-17
Add AddressInfo and update LastTxRefs with Balances

kpudlik:


modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -61,18 +67,20 @@ object GlobalSnapshotConsensusFunctions {
...
+      val blocksInRange: Set[Signed[DAGBlock]] =
         dagEvents
-          .mapFilter(_.swap.toOption)
+          .mapFilter[Signed[DAGBlock]](_.swap.toOption)
           .filter { signedBlock =>
             signedBlock.value.height <= heightLimit && signedBlock.value.height > lastGS.height
           }
           .toSet
 
+      val transactionsInRange: Set[Signed[Transaction]] = blocksInRange.flatMap(_.transactions)
```
kpudlik, line 82: It should be probably NonEmptySet 🤔

modules/shared/src/main/scala/org/tessellation/schema/balance.scala
```diff
@@ -57,6 +66,34 @@ object balance {
...
 
+    def applyTransactions[F[_]: Applicative: MonadCancelThrow](
+      transactions: Set[Transaction],
+      getBalanceFn: Address => F[Balance]
+    ): F[Map[Address, Balance]] = {
+      val sources = transactions.groupBy(_.source).keySet
+      val destinations = transactions.groupBy(_.destination).keySet
+      val addresses = sources ++ destinations
+
+      def applyTransactions(
```
kpudlik, line 77: It iterates over the whole transaction list for each balance so complexity could be improved probably

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -91,13 +99,33 @@ object GlobalSnapshotConsensusFunctions {
         ordinal = lastGS.ordinal.next
         maybeTipTrigger = tipTriggersEvents.find(_.height === heightLimit)
 
+        lastTxRefs <- transactionsInRange
+          .groupBy(_.source)
+          .toList
+          .traverse {
+            case (address, txs) =>
+              TransactionReference.of(txs.maxBy(_.ordinal)).map((address, _))
```
marcinwadon, line 107: Please make sure that `maxBy` does not throw for empty list. Scala's fn was throwing errors while cats' fn worked on Option. just double check

modules/shared/src/main/scala/org/tessellation/schema/balance.scala
```diff
@@ -57,6 +66,36 @@ object balance {
...
+      val sources = transactions.groupBy(_.source)
+      val destinations = transactions.groupBy(_.destination)
+      val addresses = sources ++ destinations
+
+      def applyTransactions(
+        balance: Balance,
+        address: Address,
+        txs: Set[Transaction]
+      ) =
+        txs.foldLeft(balance.asRight[schema.balance.BalanceOutOfRange]) { (acc, tx) =>
```
marcinwadon, line 82: please fix import `schema.balance.BalanceOutOfRange` :D

modules/shared/src/main/scala/org/tessellation/schema/transaction.scala
```diff
@@ -1,14 +1,20 @@
 package org.tessellation.schema
 
 import cats.Order
+import cats.effect.kernel.Async
```
marcinwadon, line 4: please import Async not from kernel

modules/shared/src/main/scala/org/tessellation/schema/transaction.scala
```diff
@@ -37,14 +43,20 @@ object transaction {
...
+  @derive(decoder, encoder, show)
   @newtype
   case class TransactionOrdinal(value: NonNegLong) {
     def next: TransactionOrdinal = TransactionOrdinal(value |+| 1L)
   }
 
   object TransactionOrdinal {
     val first: TransactionOrdinal = TransactionOrdinal(1L)
+
+    implicit val transactionOrder: Order[TransactionOrdinal] = (x: TransactionOrdinal, y: TransactionOrdinal) =>
```
marcinwadon, line 55: I'd use just `transactionOrdinalOrder` or just `order` and `ordering` (as we know it's for TransactionOrdinal)

marcinwadon (changes requested):

marcinwadon:
@kpudlik Please add a proper `label` to the pull request

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
kpudlik: That's why I mentioned that `Set[Signed[Transaction]]` should be probably `NonEmptySet` because we can't have empty blocks. I will doublecheck that.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -91,13 +99,33 @@ object GlobalSnapshotConsensusFunctions {
...
+        }
+
         globalSnapshot = GlobalSnapshot(
           ordinal,
           maybeTipTrigger.fold(lastGS.height)(_.height),
           maybeTipTrigger.fold(lastGS.subHeight.next)(_ => SubHeight.MinValue),
           lastGSHash,
           blocksInRange,
-          lastGS.balances, // TODO
+          addressInfo,
```
tbekas, line 128: I would suggest keeping balances and lastTxRefs in the `GlobalSnapshotInfo` object. To have a separation between the accumulated data and the raw events (blocks and state channel snapshots).

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
kpudlik: Actually it never throws because the list comes from `groupBy`.

So you either have at least 1 transaction for each `address` or you have just empty `Map`

marcinwadon (approved):
looks fine to me

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
marcinwadon: @tbekas but we've added balances into `GlobalSnapshot`. does it mean that they should be there as well?

kpudlik: I have moved it to GlobalSnapshotInfo. It makes sense

marcinwadon (dismissed):

marcinwadon (approved):

PR #220, opened 2022-03-25
Logback configuration for DAG L1

marcinwadon:


tbekas (commented):
Do you think we should have a copy of logback configurations between on L0 & L1, rather than a shared configuration between both of them, maybe in SDK?

marcinwadon:
@tbekas I'll check if we can share that. I'd go for that

tbekas (approved):

PR #223, opened 2022-03-25
DAG routes on L0 for balance and total supply

marcinwadon:


modules/core/src/main/scala/org/tessellation/http/routes/DagRoutes.scala
```diff
@@ -0,0 +1,49 @@
...
+
+}
+
+object DTO {
+
+  @derive(encoder)
+  case class BalanceDTO(balance: Balance)
+
+  @derive(encoder)
+  case class TotalSupplyDTO(total: BigInt)
```
kpudlik, line 48: Shouldn't total supply by `Amount` or at least `NonNegBitInt`? I know that it is a public service so BigInt is fine there probably but just for consistency.

kpudlik (approved):

modules/core/src/main/scala/org/tessellation/http/routes/DagRoutes.scala
marcinwadon: @kpudlik It's just for visualization (API output). That's why I made it next to the Routes object and it won't be used anywhere else.

PR #227, opened 2022-04-05
Return snapshot ordinal with balance and total-supply

marcinwadon:


modules/core/src/main/scala/org/tessellation/infrastructure/dag/DAGService.scala
```diff
@@ -15,23 +19,44 @@ object DAGService {
...
+        globalSnapshotStorage.head.flatMap {
+          case Some(snapshot) => getBalance(snapshot.value.ordinal, address)
+          case _              => none.pure[F]
         }
 
-      def getTotalSupply: F[BigInt] = {
+      def getBalance(ordinal: SnapshotOrdinal, address: Address): F[Option[(Balance, SnapshotOrdinal)]] =
+        globalSnapshotStorage.get(ordinal).map {
+          case Some(snapshot) =>
+            val balance = snapshot.value.info.balances.get(address).getOrElse(Balance.empty)
```
kpudlik, line 31: Can be shortened:

```scala
val balance = snapshot.value.info.balances.getOrElse(address, Balance.empty)
```

modules/core/src/main/scala/org/tessellation/infrastructure/dag/DAGService.scala
```diff
@@ -15,23 +19,44 @@ object DAGService {
...
+            val ordinal = snapshot.value.ordinal
+
+            (balance, ordinal).some
+          case _ => none
```
kpudlik, line 35: `Option` makes the fallback to `None` implicitly by design so if you need to do it manually then you probably shouldn't pattern match there. It is like `if (none) then none`.

I would consider using `map` then but it looks ugly with `maybeSnapshot`:

```scala
globalSnapshotStorage.get(ordinal).map { maybeSnapshot =>
  snapshot.map(...)
}
```

or would consider temporal monad transformer:

```scala
OptionT(globalSnapshotStorage.get(ordinal))
  .map { snapshot =>
     val balance = ...
     val ordinal = ...
     (balance, ordinal)
  }.value
```

modules/core/src/main/scala/org/tessellation/infrastructure/dag/DAGService.scala
```diff
@@ -15,23 +19,44 @@ object DAGService {
...
+        }
+
+      def getTotalSupply: F[Option[(BigInt, SnapshotOrdinal)]] =
+        globalSnapshotStorage.head.flatMap {
+          case Some(snapshot) => getTotalSupply(snapshot.value.ordinal)
+          case _              => none.pure[F]
+        }
```
kpudlik, line 42: Same here. With monad transformer we avoid explicit fallback from `None` to `F[None]`

```scala
OptionT(globalSnapshotStorage.head)
   .flatMap(latest => OptionT(getTotalSupply(latest.value.ordinal)))
   .value
```

modules/core/src/main/scala/org/tessellation/infrastructure/dag/DAGService.scala
```diff
@@ -15,23 +19,44 @@ object DAGService {
...
-            snapshot.value.info.balances.values
+            val supply = snapshot.value.info.balances.values
               .foldLeft(empty) { (acc, b) =>
                 acc + BigInt(b.coerce.value)
               }
-          case None => empty
+            val ordinal = snapshot.value.ordinal
+
+            (supply, ordinal).some
+          case None => none
```
kpudlik, line 57: Same here. It is a code smell imo.

kpudlik (changes requested):

kpudlik (approved):

PR #228, opened 2022-04-13
Model for global snapshot tips

tbekas:


modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/ActiveTip.scala
```diff
@@ -0,0 +1,10 @@
+package org.tessellation.dag.snapshot
+
+import org.tessellation.dag.domain.block.BlockReference
+
+import eu.timepit.refined.types.numeric.NonNegLong
+
+case class ActiveTip(
+  blockReference: BlockReference,
+  usageCount: NonNegLong
```
TheMMaciek, line 9: I wonder if we should have the count as a number vs a `Set` of child blocks hashes. For sure we would need to have the additional field for artificial usages and the total usage would be `childBlocks.size() + artificialUsages`. The advantage of a `Set` is that adding the hash again would be idempotent, with counts you need to be sure that the given block didn't increase the count already.

modules/dag-shared/src/main/scala/org/tessellation/dag/package.scala
```diff
@@ -19,6 +19,7 @@ package object dag {
     classOf[SnapshotOrdinal] -> 602,
     classOf[DAGBlock] -> 603,
     classOf[BlockReference] -> 604,
-    classOf[GlobalSnapshotInfo] -> 605
+    classOf[GlobalSnapshotInfo] -> 605,
+    classOf[ActiveTip] -> 606
```
TheMMaciek, line 23: Don't you need the `DeprecatedTip` and `GlobalSnasphotTips' as well?

modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/ActiveTip.scala
marcinwadon: We had similar issue on the current mainnet - we were not able to distinguish why tip has been used (because of which parent), so we've started using `Set` with children hashes.

modules/dag-shared/src/main/scala/org/tessellation/dag/package.scala
marcinwadon: We'll see later I guess :)

modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/ActiveTip.scala
tbekas: Likely not needed. Two reasons:
* We will be able to tell if the child is attempted to be accepted more than once, based on the tips (each block is a tip, and parents are not removed before children)
* We need to have 'artificial' usages on top of regular usages.

modules/dag-shared/src/main/scala/org/tessellation/dag/package.scala
tbekas: It's needed of course. Fixing.

modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/GlobalSnapshotTips.scala
```diff
@@ -0,0 +1,10 @@
+package org.tessellation.dag.snapshot
+
+import derevo.cats.{eqv, show}
+import derevo.derive
+
+@derive(eqv, show)
+case class GlobalSnapshotTips(
```
marcinwadon, line 7: One thing to mention, it's not necessary to repeat `tips` again, so instead of:
```scala
case class GlobalSnapshotTips(
   deprecatedTips: Set[DeprecatedTip],
   activeTips: Set[ActiveTip]
 )
 ```
 I'd see it as:
 ```scala
 case class GlobalSnapshotTips(
   deprecated: Set[DeprecatedTip],
   active: Set[ActiveTip]
 )
 ```
 
 Second thing is that if `active` are the tips from the previous snapshots, maybe it would be easier to understand by changing the name `active` into `previous`, `indirect`, `extended` or `prolonged`?

Comment (tbekas) [reply modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/GlobalSnapshotTips.scala:7]: I agree with removing the `tips` suffix, it's unnecessary.

Regarding the naming. Myabe `remainedActive`?

File: modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala [hunk 4]
```diff
@@ -124,15 +124,16 @@ object GlobalSnapshotConsensusFunctions {
           maybeTipTrigger.fold(lastGS.height)(_.height),
           maybeTipTrigger.fold(lastGS.subHeight.next)(_ => SubHeight.MinValue),
           lastGSHash,
-          blocksInRange,
+          blocksInRange.map(BlockAsActiveTip(_, 0L)),
```
TheMMaciek, line 127: Is that always true? I guess not all new blocks will have 0 usages. If that is here just for now and will be changed in the next PR, that's fine.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -124,15 +124,16 @@ object GlobalSnapshotConsensusFunctions {
...
           scSnapshots,
           rewards,
           NonEmptyList.of(PeerId(Hex("peer1"))), // TODO
           GlobalSnapshotInfo(
             lastGS.info.lastStateChannelSnapshotHashes ++ sCSnapshotHashes,
             (lastGS.info.lastTxRefs ++ lastTxRefs).groupMapReduce(_._1)(_._2) { case (_, updated) => updated },
             (lastGS.info.balances ++ balances).groupMapReduce(_._1)(_._2) { case (_, updated)     => updated }
-          )
+          ),
+          lastGS.tips
```
TheMMaciek, line 136: I guess the logic filtering moving the tips to deprecated will come later, right?

TheMMaciek (approved):

modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/GlobalSnapshotTips.scala
TheMMaciek: Or `stillActive`

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
tbekas: Yes, that's here only temporarily so that the whole change compiles.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
tbekas: Yep, in the next PR.

PR #229, opened 2022-04-15
Adjust L1 block storage to new tips logic

TheMMaciek:


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockStorage.scala
```diff
@@ -22,199 +21,116 @@ import org.tessellation.security.signature.Signed
...
     hashedBlock.signed.value.parent.toList.traverse { blockReference =>
-      blocks(blockReference.hash).update {
+      blocks(blockReference.hash).modify {
         case Some(storedBlock) =>
           storedBlock match {
-            case unknown: UnknownBlock   => unknown.addUsage(hash).some
-            case waiting: WaitingBlock   => waiting.addUsage(hash).some
-            case accepted: AcceptedBlock => accepted.addUsage(hash).some
+            case waiting: WaitingBlock =>
+              (waiting.some, TipUsageUpdateError(hash, blockReference.hash, "Waiting").asLeft)
```
marcinwadon, line 91: I see "Waiting", "Accepted", "Majority", .. a lot here. Would you mind creating some enum or more specific errors?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockStorage.scala
```diff
@@ -22,199 +21,116 @@ import org.tessellation.security.signature.Signed
...
+  sealed trait BlockStorageError extends NoStackTrace {
+    val errorMessage: String
+    override def getMessage: String = errorMessage
   }
-  case class BlockDoesNotExist(hash: ProofsHash) extends BlockAcceptanceError {
-    override def getMessage: String = s"Block with hash=$hash doesn't exist so it can't be accepted!"
+  case class TipUsageUpdateError(child: ProofsHash, parent: ProofsHash, encountered: String) extends BlockStorageError {
+
+    val errorMessage: String =
+      s"Parent block with hash=$parent not found in majority when updating usage! Child hash=$child. Encountered state: $encountered"
```
marcinwadon, line 126: You can use `.show` like this:
```
s"Parent block ${parent.show} not found in majority when updating usage. Child ${child.show}. Encountered state: ${encountered.show}"
```

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockStorage.scala
TheMMaciek: I purposefully flattened multiple errors (that we had before) into one more general with this additional field indicating what unexpected state was encountered. We have one error instead of three. I can go to the previous concept, but I remember that some time ago we discussed that single error per case may be overkill, so I changed it here. I was thinking about packing the encountered state (`Option[StoredBlock]`) into the error message and then dynamically creating the message, but I guess we also don't want to carry the whole object with state inside of an error class object and not all encountered states are erroneous. I could create an enum for this states (Waiting, Accepted, Majority, NotFound) but then they reflect the `StoredBlock` abstraction mostly and at the same time they are decoupled from that abstraction. In the end I decided that maybe just using String is good enough. We don't match these errors anywhere currently, they are only for information purposes. If you have some idea on how to efficiently tackle it without creating too much boilerplate, please share.

marcinwadon: You can use a sealed trait for each error and then specify more specific case classes. Also, if you would want to have a block inside the error I'd go with a companion object that does the logic and keep an error class simple. Companion object "apply" will use the full object just for creating the error rather than storing the object all the time.

kpudlik: Having unique errors for each kind of block would be helpful if handling differs depending on the type. Otherwise parametrized error IMO is ok.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockStorage.scala
```diff
@@ -22,199 +22,116 @@ import org.tessellation.security.signature.Signed
...
-              (accepted.some, BlockAlreadyStoredAndAccepted(hashedBlock.proofsHash).asLeft)
-          }
-        case None => (WaitingBlock(hashedBlock.signed, Set.empty[ProofsHash]).some, ().asRight)
-      }
-      .flatMap(_.liftTo[F]) >> updateParentUsages(hashedBlock)
+    blocks(hashedBlock.proofsHash).modify {
+      case Some(storedBlock) =>
+        storedBlock match {
+          case waiting: WaitingBlock =>
+            (waiting.some, BlockAlreadyStoredError(hashedBlock.proofsHash, "Waiting").asLeft)
```
kpudlik, line 56: Parametrized errors are fine but I do not like the fact that we manually put strings there. Especially that there is a fixed set of block types `String` kinda says that there are more types which may fit that error.

How about passing the classname instead?

```scala
case block @ BlockAlreadyStoredError(hashedBlock.proofsHash, block.getClass.getSimpleName).asLeft
```

kpudlik (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockStorage.scala
TheMMaciek: Ok I discussed it with @tbekas and we have come to a conclusion that by using `implicit show: Show[StoredBlock]` we can achieve the precise messages with no surplus boilerplate. We don't think there is any penalty in putting the whole `Option[StoredBlock]` into the error message as it's anyway just a reference to an object already in memory.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockStorage.scala
TheMMaciek: Already solved with passing current state and implicit show method defined in the scope.

kpudlik (approved):

PR #230, opened 2022-04-15
Prevent cross-layer joining by introducing cluster's session and id

kpudlik:


modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
```diff
@@ -206,10 +217,31 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
...
+        if (entries.contains(peer.id)) Applicative[F].unit else PeerNotWhitelisted(peer.id).raiseError[F, Unit]
+    }
+
   private def validateHandshake(registrationRequest: RegistrationRequest, remoteAddress: Option[Host]): F[Unit] =
     for {
       ip <- registrationRequest.ip.pure[F]
 
+      ownClusterId = clusterStorage.getClusterId
+
+      _ <- if (registrationRequest.clusterId == ownClusterId)
```
marcinwadon, line 233: can you use Eq from cats? `===`

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/cluster/services/Cluster.scala
```diff
@@ -85,6 +93,10 @@ object Cluster {
           clusterStorage.getPeers.map(_ + self)
         }
 
+      def createSession: F[ClusterSessionToken] =
+        GenUUID[F].make.map(ClusterSessionToken(_)).flatMap { session =>
```
marcinwadon, line 97: `flatMap` -> `flatTap`

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
```diff
@@ -206,10 +217,31 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
...
+
+      ownClusterSession <- clusterStorage.getClusterSession
+
+      _ <- ownClusterSession match {
+        case Some(session) if session === registrationRequest.clusterSession => Applicative[F].unit
+        case None                                                            => Applicative[F].unit
+        case _                                                               => ClusterSessionDoesNotMatch.raiseError[F, Unit]
+      }
+
       _ <- if (environment == Dev || ip.toString != host"127.0.0.1".toString && ip.toString != host"localhost".toString)
```
kpudlik, line 245: Regular `==` used

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
kpudlik: Sure but look at the comments below

marcinwadon (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
```diff
@@ -206,10 +217,31 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
...
+
+      _ <- if (registrationRequest.clusterId == ownClusterId)
+        Applicative[F].unit
+      else ClusterIdDoesNotMatch.raiseError[F, Unit]
+
+      ownClusterSession <- clusterStorage.getClusterSession
+
+      _ <- ownClusterSession match {
+        case Some(session) if session === registrationRequest.clusterSession => Applicative[F].unit
+        case None                                                            => Applicative[F].unit
```
TheMMaciek, line 241: Shouldn't we only allow empty session during an initial join?

TheMMaciek: Maybe we should reflect that here somehow?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/Main.scala
```diff
@@ -23,7 +23,7 @@ import org.tessellation.sdk.resources.MkHttpServer.ServerName
 import com.monovore.decline.Opts
 import eu.timepit.refined.boolean.Or
 
-object Main extends TessellationIOApp[Run]("", "DAG L1 node", version = BuildInfo.version) {
+object Main extends TessellationIOApp[Run]("", "DAG L1 node", "L1", version = BuildInfo.version) {
```
TheMMaciek, line 26: I think that if we want it to give some more realistic level of safety the `clusterId` should be a bit more convoluted than just "L1". I can imagine state channel owners calling their implementations as L1 as well. Maybe this could actually be autogenerated during build to some value? But then we would probably prevent joining between versions, which may not be desirable. Food for thought.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
marcinwadon: I think that for the rejoin (or healthcheck auto rejoin) we would require the same cluster session

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/Main.scala
marcinwadon: I'd address that together with network deployment ticket

kpudlik: On the other hand, we could validate `BuildInfo` hash as well so even if someone names his cluster `L1` then the version won't fit. But I get your point.

marcinwadon (approved):

marcinwadon (approved):

PR #231, opened 2022-04-25
Ping healtheck - outcome actions

marcinwadon:


modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensus.scala
```diff
@@ -218,7 +220,8 @@ abstract class HealthCheckConsensus[
         .map(_.isDefined)
         .ifM(
           logger.info(s"Positive outcome for peer: ${key.id}. Peer is already known."),
-          logger.info(s"Positive outcome for peer: ${key.id}. Peer is already unknown.")
+          logger.info(s"Positive outcome for peer: ${key.id}. Peer is already unknown. Adding to joining queue.") >>
```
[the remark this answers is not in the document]
marcinwadon: Initially we wanted to either remove some peers or keep them in the cluster, as an outcome from the healthcheck.
If the node is unhealthy - it should be removed, kept otherwise. So for the non-negative outcome from the healthcheck, we should either keep the peer or try to rejoin it.

marcinwadon:
> As far as I can tell from the code if we participate in the round with such key
> 
> ```
> { peerId: "A", ipAddress: "111.111.111.111" }
> ```
> 
> and in the cluster storage w have
> 
> ```
> { peerId: "A", ipAddress: "222.222.222.222" }
> ```
> 
> we will never try to ping `111.111.111.111`, so we may report `false-positive` or `false-negative` because we're pinging a different address.

@tbekas I think it may be a correct behavior. We should ping the node from the cluster storage and the key should match the local storage. If we participate in a round for the peer with a different ip/port, we should still ping the node from the cluster and there are two cases:
1. peer from key is correct, from storage is incorrect => we share negative outcome for the round:
 -> if the healthcheck outcome is negative, we remove the peer
 -> if the healthcheck outcome is non-negative, we rejoin to the peer from the key
2. peer from key is incorrect, from storage is correct => we share positive outcome for the round:
 -> if the healthcheck outcome is negative, we remove the peer (we are in minority)
 -> if the healthcheck outcome is positive, we keep the peer

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensus.scala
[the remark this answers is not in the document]
tbekas: I don't think that the result of the peer declaration health check should be adding nodes to the cluster, it doesn't make sense. This health check is for unlocking the consensus and kicking-out peers who are not publishing their declarations in a timely manner.

I think that if anything, joining the node on a positive outcome should be an action taken only after the ping health check and could be implemented as a hook `onPositiveOutcome` that's implemented by a ping health check only. It would also make more sense to add the `ipAddress` and `port` only to the ping health check key as these values could be actually used for pinging (or for comparing with what we have in cluster storage).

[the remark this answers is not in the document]
marcinwadon: @tbekas I just wanted to mimic the behavior that we have in mainnet 1.0 and it seems logical to me, to keep the cluster aligned. Missing declaration is just a trigger with its logic to determine the outcome. It's not an entirely different healthcheck, hence, when the peer is missing but healthcheck outcome is non-negative, it means that this peer should be in the cluster, and it doesn't matter what triggered that healthcheck. If one node does not have that peer, it should be rejoined.

> This health check is for unlocking the consensus and kicking-out peers who are not publishing their declarations in a timely manner.
Every healthcheck in our codebase is to keep the cluster aligned. Whether by removing peers that should be removed or by adding peers that should be added, thus the decision is binary - peer must be in the cluster or not. As an effect of cluster alignment, consensus will be unlocked.

@TheMMaciek what's your opinion on that?

tbekas:
> * peer from key is correct, from storage is incorrect => we share negative outcome for the round:
>   -> if the healthcheck outcome is negative, we remove the peer
>   -> if the healthcheck outcome is non-negative, we rejoin to the peer from the key
> 
> * peer from key is incorrect, from storage is correct => we share positive outcome for the round:
>   -> if the healthcheck outcome is negative, we remove the peer (we are in minority)
>   -> if the healthcheck outcome is positive, we keep the peer

Two things:
1. It doesn't work the way you described. If we try to join to a peer who is already in cluster storage we will get `PeerIdInUse` when validating conditions for joining, so we won't join to peer from the key anyway.
2. Even if it would work the way you described it would be simply incorrect because nodes would check the statuses of different IP addresses - yielding false values. I think there should be another status in the `PingHealthCheck` that says that there's an IP/port mismatch with what we have in the local cluster storage.

marcinwadon:
@tbekas I don't understand why it's incorrect. By following the two cases I described, it looks correct to me.

tbekas:
I think the issue is still there since the votes in the consensus are not matching the key. Example scenario:
1. There's a healthy node in a cluster `{ id: "A", ipAddress: "111" }`
2. As a malicious node we create a new round with key `{ id: "A", ipAddress: "222" }`
3. The node with `id: "A"` is pinged by everyone under ip `"111"`
4. Outcome is positive
5. Everyone removes node `"A"` and then tries to rejoin to it by using ip: `"222"` and fail
6. Node `"A"` has been successfully removed from the cluster

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/healthcheck/declaration/PeerDeclarationHealthCheck.scala
```diff
@@ -66,6 +66,20 @@ object PeerDeclarationHealthCheck {
             _ <- roundKeys.traverse(startOwnRound)
           } yield ()
 
+        def onOutcome(peers: ConsensusRounds.Outcome[F, Key[K], Health, Status[K]]): F[Unit] =
+          peers.toList.traverse {
+            case (key, t) =>
+              t match {
+                case (decision, _) if decision.isInstanceOf[PositiveOutcome] =>
```
tbekas, line 73: I think you can shorten it to:
```
case (decision: PositiveOutcome, _) =>
```

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/healthcheck/declaration/PeerDeclarationHealthCheck.scala
```diff
@@ -66,6 +66,20 @@ object PeerDeclarationHealthCheck {
...
+                  logger.info(s"Outcome for peer ${key.id}: positive - no action required")
+                case (decision, _) if decision.isInstanceOf[NegativeOutcome] =>
+                  logger.info(s"Outcome for peer ${key.id}: negative - removing peer") >>
+                    clusterStorage.removePeer(key.id)
+                case (decision, _) =>
```
tbekas, line 78: It's not very important, but it would be more elegant with a sealed trait - you could have an exhaustive match with just 2 cases. It would require another type parameter to the `HealthCheckConsensus` class, but maybe it's worth it.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/healthcheck/ping/PingHealthCheckConsensus.scala
```diff
@@ -46,70 +51,106 @@ class PingHealthCheckConsensus[F[_]: Async: GenUUID: Random](
...
     checkRandomPeers
 
+  def onOutcome(
+    peers: ConsensusRounds.Outcome[F, PingHealthCheckKey, PingHealthCheckStatus, PingConsensusHealthStatus]
+  ): F[Unit] =
+    peers.toList.traverse {
+      case (key, t) =>
+        def rejoin = joining.join(PeerToJoin(key.id, key.ip, key.p2pPort))
+        t match {
+          case (decision, round) if decision.isInstanceOf[DecisionKeepPeer] =>
```
tbekas, line 68: Similarly:
```
case (decision: DecisionKeepPeer, round) =>
```

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/healthcheck/ping/PingHealthCheckConsensusDecision.scala
```diff
@@ -0,0 +1,7 @@
+package org.tessellation.sdk.infrastructure.healthcheck.ping
+
+import org.tessellation.sdk.domain.healthcheck.consensus.types.HealthCheckConsensusDecision
+
+final case class DecisionKeepPeer() extends HealthCheckConsensusDecision
```
tbekas, line 5: `case object` maybe instead of a `case class`? Then it would simplify the pattern-match even more
```
case (DecitionKeepPeer, round) =>
```

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/healthcheck/declaration/PeerDeclarationHealthCheckDecision.scala
```diff
@@ -0,0 +1,7 @@
+package org.tessellation.sdk.infrastructure.healthcheck.declaration
+
+import org.tessellation.schema.peer.PeerId
+import org.tessellation.sdk.domain.healthcheck.consensus.types.HealthCheckConsensusDecision
+
+final case class PositiveOutcome(peer: PeerId) extends HealthCheckConsensusDecision
```
tbekas, line 6: I think that the `peer` argument is never used. If that's the case, it could be a `case object` as well.

tbekas (dismissed):
Minor comments

tbekas (commented):
LGTM

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensus.scala
```diff
@@ -29,25 +29,28 @@ abstract class HealthCheckConsensus[
...
+  def allRounds: Ref[F, ConsensusRounds[F, K, A, B, C]]
 
   def ownStatus(key: K): F[Fiber[F, Throwable, A]]
 
   def statusOnError(key: K): A
 
   def periodic: F[Unit]
 
-  def roundsInProgress: F[InProgress[F, K, A, B]] = allRounds.get.map(_.inProgress)
+  def onOutcome(peers: ConsensusRounds.Outcome[F, K, A, B, C]): F[Unit]
```
TheMMaciek, line 51: Shouldn't it be called `outcomes` instead of `peers`?

marcinwadon: I just updated the type by additional type parameter

marcinwadon: :+1: fixed

TheMMaciek (approved):

PR #232, opened 2022-04-27
Block acceptance and tips logic

tbekas:
To be added in this PR:
- tests for `BlockAcceptanceLogic`
- tests for `BlockAcceptanceManager`
- some additional validation for transactions (i.e. src != dst addres)

### DIFF
--- a/modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
+++ b/modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
@@ -3,27 +3,35 @@ package org.tessellation.infrastructure.snapshot
 import cats.data.NonEmptyList
 import cats.effect.Async
 import cats.syntax.applicative._
+import cats.syntax.applicativeError._
+import cats.syntax.bifunctor._
+import cats.syntax.either._
 import cats.syntax.flatMap._
+import cats.syntax.foldable._
 import cats.syntax.functor._
 import cats.syntax.functorFilter._
 import cats.syntax.list._
 import cats.syntax.option._
 import cats.syntax.order._
+import cats.syntax.show._
 import cats.syntax.traverse._
 import cats.{Applicative, Eval}
 
-import org.tessellation.dag.domain.block.DAGBlock
+import org.tessellation.dag.domain.block.{BlockReference, DAGBlock}
 import org.tessellation.dag.snapshot._
+import org.tessellation.domain.snapshot._
 import org.tessellation.domain.snapshot.rewards.Rewards
-import org.tessellation.domain.snapshot.{GlobalSnapshotStorage, TimeSnapshotTrigger, TipSnapshotTrigger}
 import org.tessellation.ext.cats.syntax.next._
 import org.tessellation.ext.crypto._
+import org.tessellation.infrastructure.snapshot.processing.{
+  BlockAcceptanceManager,
+  BlockAcceptanceState,
+  deprecationThreshold
+}
 import org.tessellation.kryo.KryoSerializer
 import org.tessellation.schema.address.Address
-import org.tessellation.schema.balance.Balance
-import org.tessellation.schema.height.SubHeight
+import org.tessellation.schema.height.{Height, SubHeight}
 import org.tessellation.schema.peer.PeerId
-import org.tessellation.schema.transaction.{Transaction, TransactionReference}
 import org.tessellation.sdk.domain.consensus.ConsensusFunctions
 import org.tessellation.security.SecurityProvider
 import org.tessellation.security.hash.Hash
@@ -33,7 +41,6 @@ import org.tessellation.security.signature.Signed
 import eu.timepit.refined.auto._
 import eu.timepit.refined.types.numeric.NonNegLong
 import org.typelevel.log4cats.slf4j.Slf4jLogger
-import shapeless.Typeable
 
 trait GlobalSnapshotConsensusFunctions[F[_]]
     extends ConsensusFunctions[F, GlobalSnapshotEvent, GlobalSnapshotKey, GlobalSnapshotArtifact] {}
@@ -42,11 +49,11 @@ object GlobalSnapshotConsensusFunctions {
 
   def make[F[_]: Async: KryoSerializer: SecurityProvider](
     globalSnapshotStorage: GlobalSnapshotStorage[F],
-    heightInterval: NonNegLong
+    heightInterval: NonNegLong,
+    blockAcceptanceManager: BlockAcceptanceManager[F]
   ): GlobalSnapshotConsensusFunctions[F] = new GlobalSnapshotConsensusFunctions[F] {
 
     private val logger = Slf4jLogger.getLoggerFromClass(GlobalSnapshotConsensusFunctions.getClass)
-    private val tipSnapshotTriggerTypable = Typeable[TipSnapshotTrigger]
 
     def consumeSignedMajorityArtifact(signedArtifact: Signed[GlobalSnapshotArtifact]): F[Unit] =
       globalSnapshotStorage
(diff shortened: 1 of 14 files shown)


modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -70,75 +77,99 @@ object GlobalSnapshotConsensusFunctions {
...
+            deprecated = deprecated.toSet,
+            remainedActive = remainedActive.toSet
+          )
         )
-        returnedEvents = returnedSCEvents.union(returnedDAGEvents)
+        returnedEvents = returnedSCEvents.union(returnedDAGEvents.toSet)
       } yield (globalSnapshot, returnedEvents)
     }
 
+    case class InvalidHeight(lastHeight: Height, currentHeight: Height) extends Throwable
```
marcinwadon, line 141: `Throwable` or maybe `NoStackTrace`?

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/processing/BlockAcceptanceLogic.scala
```diff
@@ -0,0 +1,132 @@
+package org.tessellation.infrastructure.snapshot.processing
```
marcinwadon, line 1: I'd say it's a `program`

```
snapshot/
  programs/
    BlockAcceptance
    BlockAcceptanceDriver
```
and I bet both can/should be in the domain 🤔

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/processing/BlockAcceptanceLogic.scala
```diff
@@ -0,0 +1,132 @@
...
+                          dstBalance <- state.balances.getOrElse(tx.destination, Balance.empty).plus(tx.amount)
+
+                          balances = state.balances.updated(tx.source, srcBalance).updated(tx.destination, dstBalance)
+                          lastTxRefs = state.lastTxRefs.updated(tx.source, txRef)
+                        } yield state.copy(balances = balances, lastTxRefs = lastTxRefs)
+
+                        balanceErrorOrState
+                          .map(state => (state, maybeNotReady))
+                          .leftMap(err => invalidBlock(blockRef, InvalidTransaction(txRef, InvalidBalance(err))))
+                          .toEitherT[F]
```
marcinwadon, line 115: I'd say that this kind of validation will be hard to maintain. I'd use `Validation` monad, extract separate pieces into own methods (like checking the parent hash / reference, checking the ordinal, etc) and I'd highly opt to reuse the transaction validator. Then, you can use tx validator to obtain whether tx is valid/invalid and add a custom validation for tips if needed here.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/processing/BlockAcceptanceManager.scala
```diff
@@ -0,0 +1,92 @@
...
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+
+trait BlockAcceptanceManager[F[_]] {
+
+  def processBlocks(
+    initState: BlockAcceptanceState,
+    blocks: List[Signed[DAGBlock]]
+  ): F[BlockAcceptanceResult]
+
+}
```
marcinwadon, line 28: didn't we agree some time ago that such traits we will put into a `domain`? to be able to reference them from services that are in the domain as well 🤔 I remember we had a discussion and that was the output, @TheMMaciek was changing some of the services to split them into domain and infra.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/processing/BlockAcceptanceManager.scala
```diff
@@ -0,0 +1,92 @@
...
+  def make[F[_]: Async: KryoSerializer: SecurityProvider](logic: BlockAcceptanceLogic[F]): BlockAcceptanceManager[F] =
+    new BlockAcceptanceManager[F] {
+      private val logger = Slf4jLogger.getLoggerFromClass[F](BlockAcceptanceManager.getClass)
+
+      def processBlocks(
+        initState: BlockAcceptanceState,
+        blocks: List[Signed[DAGBlock]]
+      ): F[BlockAcceptanceResult] = {
+
+        def go(prevResult: BlockAcceptanceResult): F[BlockAcceptanceResult] = {
```
marcinwadon, line 41: can we make it tail recursive?

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/processing/NotAcceptedBlock.scala
```diff
@@ -0,0 +1,43 @@
+package org.tessellation.infrastructure.snapshot.processing
```
marcinwadon, line 1: imo such things should be in the domain

marcinwadon (changes requested):

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/processing/BlockAcceptanceManager.scala
tbekas: We can but there's no need to. You can have a recursive function of arbitrary depth with `IO[_]` returned from `flatMap`. Also the code without tail-recursion is easier to read.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/processing/BlockAcceptanceManager.scala
tbekas: Maybe we can have an ADR that would clearly specify what classifies what? It's so ambiguous to me at the moment.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/processing/NotAcceptedBlock.scala
tbekas: This file will be removed (conclusion after the discussion about re-using the validators).

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/processing/BlockAcceptanceManager.scala
marcinwadon: yeah, I agree

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/processing/BlockAcceptanceLogic.scala
tbekas: Ok, so under that definition:

>Program is a part of business logic that combines multiple services

This class is definitely not a `program`.

marcinwadon (approved):

marcinwadon (dismissed):

modules/shared/src/main/scala/org/tessellation/schema/transaction.scala
```diff
@@ -27,73 +25,65 @@ import monocle.macros.GenLens
...
+  @derive(decoder, encoder, order, show)
   case class TransactionData(
     source: Address,
     destination: Address,
     amount: TransactionAmount,
     fee: TransactionFee
   )
 
-  @derive(decoder, encoder, eqv, show)
+  @derive(decoder, encoder, order, show)
```
marcinwadon, line 80: I think it can be misleading that it depends on the order of properties and it can be really easy to miss :|

tbekas: True. My intention was to have a sensible default order. However, I agree that it may lead to errors in the future if someone starts depending on the default order, while someone else changes the default order breaking the other places.

marcinwadon (approved):
Looks good for me but let's wait for Maciek's review :)

PR #233, opened 2022-04-29
L1 download/redownload and tips handling

TheMMaciek:
To be added in this PR:
- adjust 2 remaining tests for the old tip logic
- handle removing deprecated tips and moving tips from active to deprecated
- add tests for the new tip logic specific cases
- verify the correctness of the whole flow once again


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -67,7 +68,8 @@ class StateChannel[F[_]: Async: KryoSerializer: SecurityProvider: Random](
 
   private val ownRoundTriggerInput: Stream[F, OwnerBlockConsensusInput] = Stream
     .awakeEvery(5.seconds)
-    .evalFilter { _ =>
+    .evalTap(_ => semaphores.blockCreation.acquire)
```
marcinwadon, line 71: this semaphore is used here only. why can't you create it just before creating the StateChannel class, as we do with Ref's and similars?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -77,12 +79,14 @@ class StateChannel[F[_]: Async: KryoSerializer: SecurityProvider: Random](
         appConfig.consensus.tipsCount
       )
     }
+    .evalTap(_ => semaphores.blockCreation.release)
```
marcinwadon, line 82: I'd wrap `canStartOwnConsensus` and what's inside `evalMap` in a resource-fashion (`MonadCancel.uncancelable` or `.guarantee`), to be 100% sure the semaphore will be released. I have a gut feeling that this `evalTap` may not be executed if evalMap will throw an error, causing infinite semaphore acquisition.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -134,9 +138,17 @@ class StateChannel[F[_]: Async: KryoSerializer: SecurityProvider: Random](
...
+    _.evalTap(_ => semaphores.blockStoring.acquire).evalMap { fb =>
+      for {
+        lastSnapshotHeight <- storages.lastGlobalSnapshotStorage.getHeight.map(_.getOrElse(Height.MinValue))
+        _ <- if (lastSnapshotHeight.value < fb.hashedBlock.height.value)
+          storages.block.store(fb.hashedBlock).handleErrorWith(e => logger.debug(e)("Block storing failed."))
+        else
+          logger.debug(
+            s"Block can't be stored! Block height not above last snapshot height! block:${fb.hashedBlock.height} <= snapshot: $lastSnapshotHeight"
+          )
+      } yield ()
```
marcinwadon, line 150: what about a following syntax?
```scala
storages.lastGlobalSnapshotStorage.getHeight.map(_.getOrElse(Height.MinValue))
  .map(_.value < fb.hashedBlock.height.value)
  .ifM(
    storages.block.store(fb.hashedBlock).handleErrorWith(...),
    logger.debug(s"Block can't be stored - height above last snapshot height.")
  )
```

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -159,6 +171,7 @@ class StateChannel[F[_]: Async: KryoSerializer: SecurityProvider: Random](
 
   private val blockAcceptance: Stream[F, Unit] = Stream
     .awakeEvery(1.seconds)
+    .evalTap(_ => semaphores.blockAcceptance.acquire)
```
marcinwadon, line 174: aa. right now I have an idea that you can create an fs2 pipe that will take an effect and run it inside the acquire/release flow 🤔 like `evalMap` but with semaphore:
```
evalMapLocked(s: Semaphore[F], fn: A => F[B]): ??? = 
  s.acquire >> fn.guarantee(s.release) // or sth like that
```
maybe with a better name :D

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -177,16 +190,21 @@ class StateChannel[F[_]: Async: KryoSerializer: SecurityProvider: Random](
...
-      }
-    }
+    .evalTap(snapshot => logger.info(s"Pulled following global snapshot: $snapshot"))
+    .evalTap(
+      _ => semaphores.blockAcceptance.acquire >> semaphores.blockCreation.acquire >> semaphores.blockStoring.acquire
+    )
+    .evalMap(programs.snapshotProcessor.process)
+    .evalTap(
+      _ => semaphores.blockAcceptance.release >> semaphores.blockCreation.release >> semaphores.blockStoring.release
+    )
```
marcinwadon, line 206: if the only reason to use semaphore is to pause acceptance/creation/storage in that place I'd recommend using a `SignallingRef` and pausing above streams instead of locking 🤔

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockStorage.scala
```diff
@@ -44,7 +46,64 @@ class BlockStorage[F[_]: Sync: Random](blocks: MapRef[F, ProofsHash, Option[Stor
...
+      .flatMap(_ => addParentUsages(hashedBlock))
+
+  def adjustToMajority(
+    toAdd: Set[(Hashed[DAGBlock], NonNegLong)] = Set.empty,
+    toMarkMajority: Set[(ProofsHash, NonNegLong)] = Set.empty,
+    acceptedToRemove: Set[ProofsHash] = Set.empty,
+    obsoleteToRemove: Set[ProofsHash] = Set.empty,
+    toReset: Set[ProofsHash] = Set.empty
+  ): F[Unit] = {
+    val addMajorityBlocks = toAdd.toList.traverse {
```
marcinwadon, line 58: a rule of thumb is to write it as `def` instead of `val`. it doesn't matter because from FP standpoint both are programs that are pure but I read some time ago that it's better to use `def` wherever possible to emphasise that

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/snapshot/programs/SnapshotProcessor.scala
```diff
@@ -0,0 +1,258 @@
...
+                RedownloadPerformed(
+                  globalSnapshot.height,
+                  globalSnapshot.ordinal,
+                  globalSnapshot.hash,
+                  globalSnapshot.proofsHash,
+                  toAdd.map(_._1.proofsHash),
+                  acceptedToRemove,
+                  obsoleteToRemove
+                )
+            )
```
marcinwadon, line 194: you can write it using `{ }` instead of `( )` - it will be better formatted:
```scala
setSnapshot.map { _ =>
  ...
}
```

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/snapshot/programs/SnapshotProcessor.scala
```diff
@@ -0,0 +1,258 @@
...
+            if last.ordinal.next == globalSnapshot.ordinal && last.height.value < globalSnapshot.height.value =>
+          blockStorage.getBlocksForMajorityReconciliation(last.height, globalSnapshot.height).flatMap {
+            case MajorityReconciliationData(_, _, waitingInRange, acceptedInRange, acceptedAbove) =>
+              val acceptedLocally = acceptedInRange ++ acceptedAbove
+              val onlyInMajority = acceptedInMajority -- acceptedLocally
+              val toMarkMajority = acceptedInMajority.view.filterKeys(acceptedLocally.contains).mapValues(_._2)
+              val acceptedToRemove = acceptedInRange -- acceptedInMajority.keySet
+              lazy val toAdd = onlyInMajority.values.toSet
+              lazy val toReset = acceptedLocally -- toMarkMajority.keySet -- acceptedToRemove
+              val obsoleteToRemove = waitingInRange -- onlyInMajority.keySet
```
marcinwadon, line 234: I see some parts are repeated. can you extract both cases as separate fns? the whole for..comp will be easier to read and follow

marcinwadon (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
TheMMaciek: Initially that's how I implemented it, but then when writing a debug message I wanted to supply the context of snapshot height and block height, so I switched to for comprehension. If we are ok with more general debug message I can go with this notation.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
TheMMaciek: Makes sense as the requirement I have is a single permit semaphore. SignalingRef makes sense for that use.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockStorage.scala
TheMMaciek: Ok, let's do it like that.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
TheMMaciek: I can and initially I did, but then for some reason I thought that probably we will want to extract it somewhere, as we do with storages. I was wrong I guess ;)

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
TheMMaciek: Makes sense but also if anything throws in here the stream will close, so we are doomed either way I guess. In general we need to always handle errors if we don't want the stream to close.

TheMMaciek: And handling errors in evalMap also gives us security regarding the semaphore acquisition.

marcinwadon: We'll, I think no. Stream errors can be handled in many ways. One if it is a retry. And we don't know how errors are handled here in that place, that's why it's crucial to make safe semaphore operations and be 100% sure that the semaphore will be released when acquired, no matter what.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
marcinwadon: Hmm, for me, using a flatMap here is still more clear, but up to you. It's a detail

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
TheMMaciek: Looked into pausing the streams but it actually doesn't give me the same guarantees as using semaphores. Doesn't matter where I would put `pauseWhen` (I guess that's what was meant to be used here in conjunction with `SignalingRef`) the stream will pause on the next iteration or rather after reaching the `pauseWhen` clause for the next time. Hence storing a block after setting the `signalingRef` to `false` may still happen - so we get a race condition. Whereas when using semaphores we acquire a permit before performing an action and nothing else can use that semaphore (true when using single permit semaphores like I do here) in the meantime, so I have a guarantee that the state of the storage will not be changed. Correct me if I'm wrong but it seems to me that semaphores are still the best way to go.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -67,22 +71,27 @@ class StateChannel[F[_]: Async: KryoSerializer: SecurityProvider: Random](
...
     }
+    .evalTap(_ => blockCreationS.release)
+    .filter(identity)
     .as(OwnRoundTrigger)
 
   private val ownerBlockConsensusInputs: Stream[F, OwnerBlockConsensusInput] =
     inspectionTriggerInput.merge(ownRoundTriggerInput)
 
-  private val peerBlockConsensusInputs: Stream[F, PeerBlockConsensusInput] = Stream
+  private val peerBlockConsensusInputs: Stream[F, PeerBlockConsensusInput] = Stream // TODO: should I block it?
```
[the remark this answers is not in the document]
TheMMaciek: Or think about rather, I will ;)

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -134,9 +143,16 @@ class StateChannel[F[_]: Async: KryoSerializer: SecurityProvider: Random](
...
+    _.evalMapLocked(NonEmptyList.one(blockStoringS)) { fb =>
+      storages.lastGlobalSnapshotStorage.getHeight.map(_.getOrElse(Height.MinValue)).flatMap { lastSnapshotHeight =>
+        if (lastSnapshotHeight.value < fb.hashedBlock.height.value)
+          storages.block.store(fb.hashedBlock).handleErrorWith(e => logger.debug(e)("Block storing failed."))
+        else
+          logger.debug(
+            s"Block can't be stored! Block height not above last snapshot height! block:${fb.hashedBlock.height} <= snapshot: $lastSnapshotHeight"
+          )
+      }
+    }.evalTap(_ => blockStoringS.release)
```
marcinwadon, line 155: this is not needed

modules/shared/src/main/scala/org/tessellation/ext/fs2.scala
```diff
@@ -54,4 +58,12 @@ object fs2 {
     }
   }
 
+  implicit class StreamOps[F[_]: Sync, A](s: Stream[F, A]) {
+
+    def evalMapLocked[B](semaphores: NonEmptyList[Semaphore[F]])(fn: A => F[B]): Stream[F, B] =
```
marcinwadon, line 63: I'd write there a second fn:
```scala
def evalMapLocked[B](semaphore: Semaphore[F])(fn: A => F[B]): Stream[F, B] =
  evalMapLocked(NonEmptyList.one(semaphore))(fn)
```
because most of the time you pass one semaphore only

marcinwadon (changes requested):

modules/shared/src/main/scala/org/tessellation/ext/fs2.scala
TheMMaciek: Why not!

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
TheMMaciek: Good catch.

marcinwadon (approved):
Looks fine for me, but as a next step I'd try to simplify it a little bit, according to @tbekas comment

PR #237, opened 2022-05-10
Fix hash calculation for global snapshot consensus

marcinwadon:


modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -91,7 +91,7 @@ object GlobalSnapshotConsensusFunctions {
           .toSet
 
       for {
-        lastGSHash <- lastGS.hashF
+        lastGSHash <- lastGS.value.hashF
```
TheMMaciek, line 94: You could also `.hashWithSignatureCheck` and then take whichever hash you need: data, proofs. Other thing is that I think we should use hash from data and signatures (proofsHash) for last snasphot reference, but that's outside the scope.

TheMMaciek (approved):

PR #238, opened 2022-05-11
Partially fix the terraform script with v2 keys and new paths.

ryle-goehausen:
Updated the genesis arguments to reflect new values for the l0.
No service start script to distinguish the second L1 jar, but
available for testing purposes with manual pssh commands.

All keys were migrated using the internal keytool migration
function and then uploaded to same path but keys_v2.

Verified they could be loaded, sample command for genesis
comes from alkimi example but had a bug on working branch.

Committing this to merge the fixes so far.


ryle-goehausen:
Note this has only been tested up to the point of loading keys, but the setup script should now at least not crash and load the correct keys. Checkpointing this here for a partial implementation as switching to Rosetta. Have not tested on latest branch with @marcinwadon fix for the state issue on genesis loader.

Eventually should probably add another service loader for L1 -- unless we just have manual testing commands in the ./dag script to start it which would be fine as well for now.

marcinwadon:
Hey @ryle-goehausen. I see you mixed the dag script with kinda old terraform. We would like to avoid merging unfinished things or partially-working things into the develop and master branch

ryle-goehausen:
Leaving this here for usage by others until replacement script introduced. Not necessary to merge, can be closed as soon as there is a replacement option, this came from another branch which is not documented anywhere and was requested as example.

marcinwadon:
I'm closing it since it's a stale PR.

PR #242, opened 2022-05-13
Prevent empty blocks

marcinwadon:


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/Validator.scala
```diff
@@ -52,15 +59,17 @@ object Validator {
       stateReadyForConsensus <- nodeStorage.getNodeState.map(isReadyForBlockConsensus)
       enoughPeers <- enoughPeersForConsensus(clusterStorage, peersCount)
       enoughTips <- enoughTipsForConsensus(blockStorage, tipsCount)
+      enoughTxs <- atLeastOneTransaction(transactionStorage)
 
-      res = noOwnRoundInProgress && stateReadyForConsensus && enoughPeers && enoughTips
+      res = noOwnRoundInProgress && stateReadyForConsensus && enoughPeers && enoughTips && enoughTxs
```
kpudlik, line 64: Could have been shortened using `Applicative`:

```scala
res <- (noOwnRoundInProgress, stateReadyForConsensus, enoughPeers, enoughTips, enoughTxs)
    .mapN(_ && _)
```

kpudlik (approved):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionStorage.scala
```diff
@@ -85,6 +85,21 @@ class TransactionStorage[F[_]: Sync](
       }
       .void
 
+  def countAllowedForConsensus: F[Int] =
```
TheMMaciek, line 88: We should additionally add a validation to L1 consensus. As we are not sure that after checking condition here the transactions will still be there when starting the consensus during L1 consensus. Starting a round as facilitator may pull them before new own consensus starts. Additionally it's probably good idea to change the `DAGBlock` model to not allow adding empty Set of transactions.

PR #250, opened 2022-05-19
Reducing the global snapshot size ADR

tbekas:


marcinwadon (approved):
Looks reasonable and I really like the idea :+1:

marcinwadon:
bump @buckysballs @kpudlik @TheMMaciek @kriscieslak

doc/adr/0010-reducing-snapshot-size.md
```diff
@@ -0,0 +1,343 @@
...
+          "destination": "D",
+          "amount": 10
+        }
+      ]
+    }
+  ],
+  "info": {
+    "partial": {
+      "balances": {
+        "B": 10,
```
TheMMaciek, line 192: B wasn't influenced by any transaction so I guess it shouldn't be in `partial`

doc/adr/0010-reducing-snapshot-size.md
```diff
@@ -0,0 +1,343 @@
...
+The problem is that every single `GlobalSnapshot` contains all this data for all addresses. This is suboptimal because
+majority of this data never changes between two consecutive snapshots, and it results in increasing-only size of
+snapshots. In the long run this can cause issues with storage, network and memory.
+
+## Decision
+
+Replacing existing `GlobalSnapshotInfo` type with a product type of:
+1. `PartialInfo` - Contains the same fields as the existing `GlobalSnapshotInfo` type and represents overrides to the 
+    `balances`, `lastTransactionRef` and `lastStateChannelSnapshotHashes` that are applied by this snapshot. 
+    Missing values for an address should be treated as lack of overrides (the value from previous snapshot still holds).
```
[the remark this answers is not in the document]
marcinwadon: What exactly do you mean @TheMMaciek? From the perspective of this ADR, there are no deletions as well afaik.

doc/adr/0010-reducing-snapshot-size.md
```diff
@@ -0,0 +1,343 @@
...
+
+### Aggregated view resolution
+
+Algorithm for resolving an aggregated view of all data - from the genesis to the latest snapshot:
+1. Variable initialization:
+    * `snapshot := <latest-snapshot>`
+    * `acc := snapshot.info.aggregated`
+2. Loop:
+    * if `snapshot.info.aggregated.since` is equal to `0` return `acc`
+    * otherwise, get the snapshot with ordinal `snapshot.info.aggregated.since` and assign it to `snapshot`
```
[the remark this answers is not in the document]
marcinwadon: I support that only if downloading the last full snapshot is required

doc/adr/0010-reducing-snapshot-size.md
```diff
@@ -0,0 +1,343 @@
...
+  ],
+  "info": {
+    "partial": {
+      "balances": {
+        "B": 10,
+        "C": 10
+      }
+    },
+    "aggregated": {
+      "since": 1,
```
TheMMaciek, line 163: I'm starting to think that `since` should be the non-inclusive part of range and should be of type `SnapshotReference` (containing ordinal and hash) of the last full snapshot. Then the next snapshot after full one will still have the `partial` and `aggregated` equal (which feels natural). But the whole range of snapshots will reference the last full snapshot so then when downloading a snapshot during download/redownload we know upfront which full snapshot we need to fetch and we can verify if we got what we wanted.

TheMMaciek: We could also reflect the full snapshot nature with a separate type.

doc/adr/0010-reducing-snapshot-size.md
```diff
@@ -0,0 +1,343 @@
...
+        {
+          "source": "B",
+          "destination": "C",
+          "amount": 10
+        }
+      ]
+    }
+  ],
+  "info": {
+    "partial": {
```
TheMMaciek, line 156: I'm starting to think that having `partial` as well as `aggregated` gives us nothing. It is isomorphic to the blocks added in this snapshot and gives nothing during download/redownload on L1.

doc/adr/0010-reducing-snapshot-size.md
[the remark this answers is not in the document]
TheMMaciek: I know. That's exactly what I mean. This architecture doesn't allow for deletion and maybe the deletion should be allowed. I remember we discussed some time ago this case based on ethereum, where some smart contracts can and some can't be deleted. Possibly that's not the same thing, but I wanted to bring it into the discussion to be sure.

doc/adr/0010-reducing-snapshot-size.md
marcinwadon: it would be more complex but we gain nothing I believe

doc/adr/0010-reducing-snapshot-size.md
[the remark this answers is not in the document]
TheMMaciek: Well if the snapshot you download is a full snapshot you don't need to download anything else. If it's let's call it `PartialSnapshot` then you need to download the full snapshot that this partial one references as well. I would also like to find some better name that reflects the fact that these partial snapshots only build on top of the full snapshot. Maybe they should be called `DiffSnapshots` or `UpdateSnapshots` or something. I think that referencing the last full snapshot also gives us some freedom in changing the logic of full/partial snapshot creation. From the point of view of someone asking for a snapshot the fact that every full snapshot is created every 1000 ordinals is a magic number. But if we just reference the last full snapshot we can create them every 10, 100, 1000 or whatever dynamically changing number we want (it could rely on the network load as well) and the client doesn't need to follow the logic.

doc/adr/0010-reducing-snapshot-size.md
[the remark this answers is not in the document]
tbekas: Missing value means a lack of override. If we decide that for example `Hash.empty` means that the state channel has been removed, then we can use `Hash.empty` value to reflect that.

doc/adr/0010-reducing-snapshot-size.md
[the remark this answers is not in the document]
tbekas: >when downloading any of those you can then check if you got back what you requested.

So I've made a mistake in this algorithm. It should say get the snapshot with ordinal `since - 1` instead of `since`. So we're not referencing last `full` snapshot here, but the next one, because we're starting to aggregate after the last `full` snapshot.

doc/adr/0010-reducing-snapshot-size.md
tbekas: Yep, thanks for catching this!

doc/adr/0010-reducing-snapshot-size.md
tbekas: @TheMMaciek if it's not inclusive it would make impossible having aggregation from genesis. Having `since = 0` means that you're not including genesis, therefore any aggregation scheme would need +1 more download, just for the genesis.

doc/adr/0010-reducing-snapshot-size.md
tbekas: Please explain how `aggregated` is isomorphic with blocks added in this snapshot? 

`partial` is also not isomorphic btw. You can have two transactions for the same address that are updating balances with values `+1`  and `+1` or a single transaction with `+2`, or `+3 and -1`, etc... Isomorphism is injective (1 to 1 function, never 1 to many).

That being said, `partial` is a simplified view of the same data that is already in transactions and state channel snapshots, however, I think it's worth having it since clients processing a stream of global snapshots will have a clear view of what values have been updated in every single snapshot without having to apply those blocks and transactions manually.

marcinwadon: Exactly, the idea of partial data is to be able to quickly determine what has changed in a single snapshot (hence block explorer can use that).

doc/adr/0010-reducing-snapshot-size.md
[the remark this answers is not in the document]
marcinwadon: Actually I'd avoid using "diff" name because it will be misleading. That being said, we don't keep the difference between balances, but the partial update of balances. Hence, "partial" name suits the best IMO.

doc/adr/0010-reducing-snapshot-size.md
TheMMaciek: That's what I meant by saying isomorphic (so I used it wrongly), that by folding transactions you will get exactly what partial is.

doc/adr/0010-reducing-snapshot-size.md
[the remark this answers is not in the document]
tbekas: I think that the name proposed offline by @marcinwadon `SnapshotInfo` suits the best. Both `SnapshotInfo` and `AggregatedInfo` could be then composed into a product type `GlobalSnapshotInfos` I think (similar to `GlobalSnapshotTips`).

doc/adr/0010-reducing-snapshot-size.md
tbekas: Yep, so then the `since` has to be an `Option[SnapshotReference]` and `none` has to implicitly mean "since forever".

doc/adr/0010-reducing-snapshot-size.md
TheMMaciek: Whereas aggregated gives some added value, because we don't have these transactions in this snapshot. So partial is kind of duplication of information. But if we think it's a worthy optimization then let's have it. What bothers me is the ever problematic decoupling between transactions and folded balances from these transactions, they should be the same but programmatically it's not guaranteed.

tbekas: @TheMMaciek yes, but you have to have previous balances stored locally to apply the transactions, so clients are getting more complicated.

marcinwadon: yes, but the opposite direction does not hold

kpudlik (approved):
Sounds great :)

marcinwadon (approved):

doc/adr/0010-reducing-snapshot-size.md
```diff
@@ -0,0 +1,409 @@
...
+        "F": 0,
+        "G": 10
+      }
+    }
+  }
+}
+```
+
+
+## Consequences
```
marcinwadon, line 404: there can be also positive consequences: I'd also say that a consequence of that decision is ability to scan (by a block explorer, or any other network scanner) partial data and update the state partially.

kriscieslak (approved):
LGTM

TheMMaciek (approved):

doc/adr/0010-reducing-snapshot-size.md
```diff
@@ -0,0 +1,409 @@
...
+The problem is that every single `GlobalSnapshot` contains all this data for all addresses ever used. This is suboptimal
+because majority of this data never changes between two consecutive snapshots, and it causes an increasing-only size of
+snapshots. In the long run this can cause issues with storage, network and memory.
+
+## Decision
+
+Replacing existing `GlobalSnapshotInfo` type with a type `GlobalSnapshotInfos` which is a product type of:
+1. `SnapshotInfo` - Contains the same fields as the `GlobalSnapshotInfo` type and represents overrides to the 
+    `balances`, `lastTransactionRef` and `lastStateChannelSnapshotHashes` that are applied by this snapshot. 
+    Missing values for an address should be treated as lack of overrides - the value from previous snapshot still holds.
```
kpudlik, line 33: The cool thing is that it will be transparent for OpenSearch. We used the same pattern for balances where we just sort by `ordinal` and we return last available balance when user asks for balance on specific `ordinal`. I like the idea.

kpudlik (approved):

PR #251, opened 2022-05-20
Fix healthcheck

marcinwadon:
Mainly:
- register missing kryo classes
- more verbose logging
- trigger ping healthcheck on gossip error

### DIFF
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusState.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusState.scala
@@ -1,5 +1,7 @@
 package org.tessellation.sdk.infrastructure.consensus
 
+import cats.Show
+
 import scala.concurrent.duration.FiniteDuration
 
 import org.tessellation.schema.peer.PeerId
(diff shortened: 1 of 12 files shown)


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusState.scala
```diff
@@ -18,9 +20,20 @@ case class ConsensusState[Key, Artifact](
   statusUpdatedAt: FiniteDuration
 )
 
-@derive(eqv, show)
+@derive(eqv)
 sealed trait ConsensusStatus[Artifact]
 
+object ConsensusStatus {
```
tbekas, line 26: Thanks for fixing this, we need to limit logging to essential only info and this wasn't essential  👍

modules/sdk/src/main/scala/org/tessellation/sdk/package.scala
```diff
@@ -45,6 +46,12 @@ package object sdk {
       TimedOut.getClass -> 519,
       classOf[PeerMismatch] -> 520,
       classOf[PeerCheckTimeouted] -> 521,
-      classOf[PeerCheckUnexpectedError] -> 522
+      classOf[PeerCheckUnexpectedError] -> 522,
+      classOf[Host] -> 523,
```
tbekas, line 50: I will create a ticket for figuring out if we can somehow force Kryo to use our logger. We spend so much time because of these errors.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/gossip/GossipDaemon.scala
```diff
@@ -125,15 +128,15 @@ object GossipDaemon {
         selectedPeers <- Random[F].shuffleList(peers.toList).map(_.take(cfg.fanout))
         _ <- selectedPeers.parTraverse { peer =>
           runGossipRound(activeHashes, seenHashes, peer).handleErrorWith { err =>
-            logger.error(err)(s"Error running gossip round with peer ${peer.show}")
+            logger.error(err)(s"Error running gossip round with peer ${peer.show}") >>
```
tbekas, line 131: 👍

tbekas (approved):
LGTM

PR #254, opened 2022-05-23
Spread ping-healthcheck proposal as PeerRumor instead of CommonRumor

marcinwadon:


modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensusRound.scala
```diff
@@ -146,10 +145,9 @@ class HealthCheckConsensusRound[F[_]: Async, K <: HealthCheckKey, A <: HealthChe
     }
 
   private def sendProposal: F[Unit] =
-    ownConsensusHealthStatus.flatMap(
-      status =>
-        gossip.spreadCommon(status).flatTap(_ => logger.debug(s"Own status for round ${roundId.show} is ${status}"))
-    )
+    ownConsensusHealthStatus.flatMap { status =>
+      gossip.spread(status)
```
TheMMaciek, line 149: If so maybe, `.flatMap(gossip.spread)` or `.flatMap(gossip.spread(_))`

marcinwadon: It will not work that way because `gossip.spread` is not able to infere the rumor type properly, causing:
```scala
Unhandled rumor PeerRumorBinary(origin=93010bcb76161676eb35cc70e7c7b18a48cbca18e313157abc3fdbfe8c920b3b45e947c3a0069aaedd0aba21f55cf33fc0a6aad3c237dddf264c34eaa824da3e, ordinal=Ordinal{generation=1653306912927,counter=13}, contentType=Object) with hash 041e9ab94420067aff288cc2748092e917236e0fe7ba678916f3335c9e2728b9.
```

TheMMaciek: Interesting.

TheMMaciek (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensusRound.scala
tbekas: Yep, I found it interesting as well. Even though the compiler could inject the specific type of the rumor to the `gossip.spread` function, it won't do it. The actual injected type will be just `Object` as @marcinwadon showed above.

PR #257, opened 2022-05-24
Fixes for healthcheck determinism

marcinwadon:


modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensus.scala
```diff
@@ -121,11 +122,12 @@ abstract class HealthCheckConsensus[
   private def createRoundId: F[RoundId] = GenUUID[F].make.map(RoundId.apply)
 
   def startOwnRound(key: K) =
-    createRoundId
-      .map(HealthCheckRoundId(_, selfId))
-      .flatMap {
-        startRound(key, _)
-      }
+    clusterStorage
+      .hasPeerId(key.id)
```
TheMMaciek, line 126: Maybe this check should be more refined, checking if the node isn't offline or leaving.

marcinwadon: sure, I'll add that in next PR if so :+1:

PR #262, opened 2022-06-01
Fix non-stable hash calculation of GlobalSnapshot

tbekas:


modules/dag-shared/src/main/scala/org/tessellation/dag/domain/block/DAGBlock.scala
```diff
@@ -1,41 +1,31 @@
...
-import derevo.cats.{eqv, show}
+import derevo.cats.{order, show}
 import derevo.circe.magnolia.{decoder, encoder}
 import derevo.derive
 
 case class DAGBlockReference(parents: NonEmptyList[BlockReference])
 case class DAGBlockData(transactions: Set[Signed[Transaction]])
 
-@derive(encoder, decoder, eqv, show)
+@derive(encoder, decoder, order, ordering, show)
```
tbekas, line 22: FYI: I checked the usages of the existing `Order` instance, that I removed below and it was used only in a single place that required any deterministic order (not necessarily ordered by height as the first field), so I could safely replace it with generated `Order`.

modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/BlockAsActiveTip.scala
```diff
@@ -1,15 +1,20 @@
...
 
-@derive(eqv, show)
+@derive(order, show)
 case class BlockAsActiveTip(
   block: Signed[DAGBlock],
   usageCount: NonNegLong
 )
+
+object BlockAsActiveTip {
+  implicit object Ordering extends OrderBasedOrdering[BlockAsActiveTip]
```
tbekas, line 19: It's supposed to be `OrderingInstance` here to avoid naming clash with `scala.math.Ordering`. I will change it with any other PR fixes.

modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/ActiveTip.scala
```diff
@@ -1,15 +1,20 @@
...
-@derive(eqv, show)
+@derive(order, show)
 case class ActiveTip(
   block: BlockReference,
   usageCount: NonNegLong,
   introducedAt: SnapshotOrdinal
 )
+
+object ActiveTip {
+  implicit object OrderingInstance extends OrderBasedOrdering[ActiveTip]
```
TheMMaciek, line 19: It's called `Ordering` in `BlockAsActiveTip` maybe we can use the same naming convention everywhere?

tbekas: True, I think that `OrderingInstance` is a better name, because it doesn't conflict with `scala.math.Ordering`. I've noticed the inconsistency and I've put a comment about it https://github.com/Constellation-Labs/tessellation/pull/262#discussion_r887828217

modules/shared/src/main/scala/org/tessellation/schema/address.scala
```diff
@@ -56,6 +59,9 @@ object address {
 
     implicit val orderDAGAddress: Order[DAGAddress] =
       orderOf[String, DAGAddressRefined]
+
+    implicit val orderingDAGAddress: Ordering[DAGAddress] =
```
TheMMaciek, line 63: Can't you just convert the order above to an ordering?

tbekas: Yep, I can. Fixing it.

modules/shared/src/main/scala/org/tessellation/security/signature/Signed.scala
```diff
@@ -39,9 +41,9 @@ object Signed {
 
   implicit def autoUnwrap[T](t: Signed[T]): T = t.value
 
-  implicit def order[A: Order]: Order[Signed[A]] = (x: Signed[A], y: Signed[A]) => Order[A].compare(x, y)
+  implicit def order[A: Order]: Order[Signed[A]] = Order.by(s => (s.value, s.proofs))
```
TheMMaciek, line 44: 👍

modules/shared/src/main/scala/org/tessellation/security/signature/Signed.scala
```diff
@@ -106,4 +108,15 @@ object Signed {
         .hashF
         .map(hash => ProofsHash(hash.coerce))
   }
+
+  final class SignedOrdering[A: Ordering] extends Ordering[Signed[A]] {
```
TheMMaciek, line 112: Isn't the redefined order for `Signed` enough to derive ordering from it? Won't it give the same result as this definition?

PR #263, opened 2022-06-01
Updated the scoping syntax

cngo-github:
Updated the trait scoping syntax.

https://www.scala-sbt.org/1.x/docs/Migrating-from-sbt-013x.html#slash


marcinwadon:
@cngo-github you don't need to open and close pull requests to make a change. you can just update your branch and push it again

PR #265, opened 2022-06-01
ADR for rewards

kpudlik:


doc/adr/0012-rewards.md
```diff
@@ -0,0 +1,44 @@
+# Rewards
```
marcinwadon, line 1: I'd say "Dynamic network speed calculation"

doc/adr/0012-rewards.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Status
+
+Proposed
+
+## Context
+
+Facilitators get rewarded in DAG currency for participating in consensus rounds.
+We need to figure out the algorithm for minting/sending rewards to facilitators. Algorithm
+should take dynamic network speed into account to determine the size of rewards per node.
```
marcinwadon, line 13: Then this part could be:
"
The rewards model requires network speed (snapshots per min) as an input. In previous implementation this value was hardcoded. Currently, we need to figure out the algorithm for determining the network speed dynamically and deterministically, so then the reward model can be calculated locally as a pure function, assuming all inputs will be equal on each node.
"

doc/adr/0012-rewards.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Decision
+
+1. Each node will track the time difference between the creation of local snapshots.
+We can either store the average or keep the specific number of meaningful diffs (e.g. 100 last diffs) and calculate the average on demand.
+The calculated average is **local network speed** and will be proposed to `GlobalSnapshot` consensus.
+
+Sample output: `3 min 15s`
+
+2. The snapshot speed should be rounded (e.g. to minutes) to increase the probability of selecting the value which occurs most often (dominant value). There are 2 possible results of finding dominant value: found or not.
```
marcinwadon, line 23: I'd say that depending on the standard deviation, we can round to different value (for example up to 10 minutes). Just throwing out an idea

marcinwadon (changes requested):

doc/adr/0012-rewards.md
```diff
@@ -0,0 +1,44 @@
...
+## Context
+
+Facilitators get rewarded in DAG currency for participating in consensus rounds.
+We need to figure out the algorithm for minting/sending rewards to facilitators. Algorithm
+should take dynamic network speed into account to determine the size of rewards per node.
+
+## Decision
+
+1. Each node will track the time difference between the creation of local snapshots.
+We can either store the average or keep the specific number of meaningful diffs (e.g. 100 last diffs) and calculate the average on demand.
```
ryle-goehausen, line 18: One mention here is there's also a dependency on the download process to be updated to incorporate this information -- or otherwise have nodes that are new wait to propose these values until they have the requisite information. Since there is a state dependency, either the download must pull that state from another active node -- or more simply, it could propose None's and have them filtered out until it catches up and can determine it's own local value. The second solution is probably better as it doesn't deal with issues of modifying the download or dealing with synchronization issues across the node it's downloading from.

doc/adr/0012-rewards.md
ryle-goehausen: Agree strongly with this -- the only goal here presently is to give a rough approximation of time -- as opposed to dealing with scaling issues. In the long term, there is some dependency on the actual compute cost (which incentivizes nodes to spin up more instances in response to demand -- which requires some difficulty setting.) In the short term the only goal is to provide a rough measure of time for divisions, which means the time scale chosen can be relatively long, and has no requirement for capturing short variations. It can actually be much longer, but as mentioned earlier it's probably best to keep it medium because if it gets too great then nodes will have to wait to re-calculate that state -- or rely on externally downloaded non-synchronized state from another node.

tbekas (changes requested):
I think we need to use an approach that will be more secure and will not allow for arbitrary voting like this. Facilitators have no reason to vote truthfully, because it's in their benefit to vote with as high value of time delay as possible, while also triggering snapshots as frequently as possible. 

I think that snapshots triggered by data should be rewarded with DAG taken from transaction fee that's triggering snapshot. Time triggered snapshots should be rewarded with predefined and constant amount of DAGs (with respect to a single epoch).

marcinwadon:
@tbekas @kpudlik Can you provide an updated version of the ADR?

tbekas:
I won't be able to create an ADR this week as I have other assignments, but I can help in creating a draft that could be filled with text, examples, and diagrams. @kpudlik if you want we can collaborate on this.

tbekas:
This reward schema is no longer up to date. It has been amended by #563

PR #274, opened 2022-06-08
Check data equality when adding proofs for Signed

TheMMaciek:


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/BlockConsensusCell.scala
```diff
@@ -340,15 +339,27 @@ object BlockConsensusCell {
...
-                    .map(_ => FinalBlock(hashedBlock).asRight[CellError].widen[Ω])
+              maybeFinalBlock = roundData.peerBlocks.values.foldLeft {
+                ownBlock.asRight[MergingProofsForDistinctValues[DAGBlock]]
+              } { case (agg, sb) => agg.flatMap(_.addProofs(sb)) }
+              result <- maybeFinalBlock match {
+                case Left(error) =>
+                  cancelRound(roundData.ownProposal, ctx).map(
+                    _ =>
+                      CellError(
+                        s"Round cancelled after merging blocks to a final block failed! Original message: ${error.getMessage}"
```
marcinwadon, line 350: So "Round cancelled" or "Round ... failed" :D

marcinwadon (approved):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/BlockConsensusCell.scala
TheMMaciek: But the round got cancelled as a result of a failure during blocks merging ;).

PR #278, opened 2022-06-08
Queue proposals for future healthcheck rounds

marcinwadon:


tbekas (dismissed):

kpudlik (approved):

kriscieslak (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensus.scala
```diff
@@ -204,17 +209,20 @@ abstract class HealthCheckConsensus[
...
-                 historicalRound.map(handleProposalForHistoricalRound(proposal))
-               }
-               .getOrElse {
-                 if (depth > 0)
-                   participate.flatMap(_ => handleProposal(proposal, depth - 1))
-                 else
-                   (new Throwable("Unexpected recursion!")).raiseError[F, Unit]
-               }
+             inProgressRound.map { r =>
+               r.hasProposal(proposal.owner)
```
TheMMaciek, line 213: If it's exactly the same proposal then we can discard it immediately I guess (so the same observation and the same round id). If it's a proposal that matches the round id's but it's not exactly the same then it sounds like an error (node must have rejoined the consensus). If it's a different proposal yet then it's probably a proposal for the next round and then it should be put to a waiting proposals cache.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensus.scala
```diff
@@ -204,17 +209,20 @@ abstract class HealthCheckConsensus[
...
+                 .ifM(
+                   waitingProposals.update(_ + proposal),
+                   r.processProposal(proposal)
+                 )
+             }.orElse {
+               historicalRound.map(handleProposalForHistoricalRound(proposal))
```
TheMMaciek, line 219: Shouldn't we first check the historical rounds and only if the proposal is not matching the historical rounds, then try process it for the in progress ones?

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensus.scala
marcinwadon: Discarding is done within the round, in its `processProposal` function.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensus.scala
marcinwadon: Good point :+1: Changed

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensusRound.scala
```diff
@@ -70,6 +70,9 @@ class HealthCheckConsensusRound[F[_]: Async, K <: HealthCheckKey, A <: HealthChe
     }
   }
 
+  def hasProposal(peer: PeerId): F[Boolean] =
```
TheMMaciek, line 73: I think it should be more refined and it should tell wether that is the same proposal, or it's a different proposal but for the same `RoundId`s or it's a different proposal without relation to `RoundId`s of this healthcheck. Based on this information the proposal should be handled differently.

marcinwadon: IMO info whether round has a proposal from peer or no is enough. Proposal can be handled within the round by itself and from "manager" point of view I'd just need to know whether pass proposal to the round itself or no.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensus.scala
TheMMaciek: I guess this one will be discarded the minute this round becomes historical. Makes sense as well, we don't need to remove it prematurely even if we know we could.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensus.scala
```diff
@@ -116,6 +117,10 @@ abstract class HealthCheckConsensus[
 
             ConsensusRounds(updatedHistorical, updatedInProgress)
         }
+      }.flatTap { _ =>
+        waitingProposals.modify { proposals =>
+          (Set.empty[B], proposals.toList.traverse(handleProposal(_)))
```
TheMMaciek, line 122: Every time it's not processed for inProgress or historical round it will come back to the pool right?

marcinwadon: Yes, that's the current implementation

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensusRound.scala
TheMMaciek: If you ask the round for more details you may find the abnormal situation, like sending two different proposals for the same round from the same peer - this could also be checked for the historical round instead (the moment it becomes historical), so I guess it's ok. But in general I would rather know and figure out if we have an issue then to just let such proposals fly by.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/healthcheck/consensus/HealthCheckConsensus.scala
TheMMaciek: Because if it's not passed to the round then it won't be discarded in round's `processProposal` function.

TheMMaciek (approved):

PR #279, opened 2022-06-08
Use sbt directly in github actions (pull requests only)

marcinwadon:


.github/workflows/pull-requests.yml
```diff
@@ -10,31 +10,32 @@ on:
...
   build:
-    name: Build
+    name: Test
     runs-on: ubuntu-18.04
     strategy:
       matrix:
         java:
-          - jdk8
-          - jdk11
+          - zulu@1.8.282
```
tbekas, line 18: Why not openjdk?

marcinwadon: I haven't seen an openjdk 8 available but more importantly, openjdk build is slower than zulu or graalvm

kriscieslak (commented):
Do we need `nix` scripts after these changes?

marcinwadon:
> Do we need `nix` scripts after these changes?

I think yes. `nix` wasn't there just for CI, but for local development as well

kriscieslak (approved):

PR #284, opened 2022-06-09
GET transaction endpoint on L1

marcinwadon:


kpudlik (approved):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionStorage.scala
```diff
@@ -142,6 +143,11 @@ class TransactionStorage[F[_]: Async: KryoSerializer](
       }.map(_.flatten)
     } yield NonEmptyList.fromList(allPulled)
 
+  def find(hash: Hash): F[Option[Hashed[Transaction]]] =
+    waitingTransactions.toMap.flatMap { _.view.values.toList.flatMap(_.toList).traverse(_.toHashed) }.map {
```
TheMMaciek, line 147: Looks like a very costly operation. If we need to allow users to fetch the waiting transaction by hash, then maybe we should store it differently (with hash next to it).

TheMMaciek: Maybe as `Hashed[Transaction]` even

TheMMaciek: We are generating the hash either way when uploading the transaction, so we can as well store it is `Hashed`, then the consecutive function and this functionality should get simpler.

TheMMaciek: I think it may be fine to use `Hashed` internally, sending already `Hashed` value to someone would be problematic (everyone should verify Signed on their own)

TheMMaciek (changes requested):
I would consider storing `Hashed[Transactions]` to not recalculate the hash many times by different processes.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/BlockConsensusCell.scala
```diff
@@ -112,8 +112,16 @@ object BlockConsensusCell {
   private def deriveConsensusPeerIds(proposal: Proposal, selfId: PeerId): Set[PeerId] =
     proposal.facilitators + proposal.senderId + proposal.owner - selfId
 
-  private def returnTransactions[F[_]](ownProposal: Proposal, transactionStorage: TransactionStorage[F]): F[Unit] =
-    transactionStorage.put(ownProposal.transactions)
```
TheMMaciek, line 116: Maybe we should keep them as hashed transactions, but unwrap them only when sending to facilitators?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionStorage.scala
```diff
@@ -115,7 +118,8 @@ class TransactionStorage[F[_]: Async: KryoSerializer](
           maybePulled <- maybeWaiting.traverse { waiting =>
             val lastTx = lastAccepted.getOrElse(address, TransactionReference.empty)
             Consecutive
-              .take[F](waiting.toList, lastTx)
+              .take[F](waiting.toList.map(_.signed), lastTx)
```
TheMMaciek, line 121: Not sure if modifying take to operate on `Hashed[Transaction]` is not the better idea.

modules/shared/src/main/scala/org/tessellation/security/Hashed.scala
```diff
@@ -1,13 +1,28 @@
...
+  implicit def ordering[A <: AnyRef: Ordering]: Ordering[Hashed[A]] = new HashedOrdering[A]()
+}
+
+final class HashedOrdering[A <: AnyRef: Ordering] extends Ordering[Hashed[A]] {
+
+  def compare(x: Hashed[A], y: Hashed[A]): Int =
+    Order
+      .whenEqual(
+        Order.fromOrdering(Ordering[A]).contramap[Hashed[A]](h => h.signed),
+        Order[Hash].contramap[Hashed[A]](h => h.hash)
```
TheMMaciek, line 25: what about `ProofsHash`? I guess it should also serve as the field for defining order.

marcinwadon: `Order.whenEqual(x, y)` tries to order by `x` and then (if two elements are equal) falls back to `y` as a tiebreak. Hash is always a unique value so there is no need to combine another order 🤔 
@tbekas can you take a look and confirm what I wrote, or am I wrong?

TheMMaciek (approved):

PR #290, opened 2022-06-11
Remove ip checking from p2p middleware

tbekas:


modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/middlewares/PeerAuthMiddleware.scala
```diff
@@ -178,10 +171,7 @@ object PeerAuthMiddleware {
           .attemptT
           .toOption
 
-        id <- getRemoteHost(req).map { host =>
-          knownPeer(host).map(_.find(id => idFromHeaders.exists(_ == id))).attemptT.toOption.flatMap(_.toOptionT)
```
marcinwadon, line 182: Are you sure it's covered by another case/verifier? I'd still get the ID from clusterStorage, either by checking if ID from headers exists, or get ID by sessionToken. Wdyt?

tbekas: We could verify that, but I think it would duplicate the session token verification.
We're calling this method by passing `PeerId` from `X-Id` header. I think it's sufficient, wdyt?
```
def verifyToken(peer: PeerId, headerToken: Option[SessionToken]): F[TokenVerificationResult] =
        verifyToken(headerToken, clusterStorage.getPeer(peer))
```

marcinwadon: So we check if session token is known and we check if given Id signs the req/res, but I think we don't check if session token matches Id, am I right? If no, then it's sufficient.

tbekas: We're calling `clusterStorage.getPeer(idFromHeaders)`  and then we check if the resulting `Option[Peer]` contains peer that satisfies `_.session == tokenFromHeaders`. Isn't tat sufficient?

marcinwadon: That's fine :+1:

marcinwadon (approved):

PR #291, opened 2022-06-14
Port over rewards

kpudlik:


modules/core/src/main/scala/org/tessellation/domain/snapshot/rewards/Rewards.scala
```diff
@@ -1,15 +1,102 @@
...
+  val epochDurationInYears: PosDouble = 2.5
+  val epochRewards: PosDouble = 853333333.20
+  val normalizationFactor: PosDouble = 1e8
+  val softStaking = SoftStakingConfig(1, SnapshotOrdinal(1L))
+
+  private[rewards] def getRewardsPerSnapshot(
+    snapshotOrdinal: SnapshotOrdinal,
+    snapshotsPerMinute: NonNegInt
+  ): NonNegDouble =
+    getEpochNumber(snapshotOrdinal, snapshotsPerMinute) match {
```
tbekas, line 30: I think we cannot base epoch on snaphshot ordinal. I think we should have some kind of count-down to the next epoch stored in snapshot, as we don't know how many time-triggered snapshot's there will be. On each time-triggered snapshot we would decrement the count-down.

modules/core/src/main/scala/org/tessellation/domain/snapshot/rewards/Rewards.scala
```diff
@@ -1,15 +1,102 @@
...
+import org.tessellation.schema.transaction.{RewardTransaction, TransactionAmount}
+import org.tessellation.security.SecurityProvider
+
+import eu.timepit.refined.auto._
+import eu.timepit.refined.types.numeric._
 
 object Rewards {
+  val epochs: PosInt = 4
+  val epochDurationInYears: PosDouble = 2.5
+  val epochRewards: PosDouble = 853333333.20
```
tbekas, line 22: I'm very much in favor of doing all arithmetics on integer numbers. So `Long` or `Amount` type. And always normalized. Doing arithmetic on floating-point's will lead to rounding errors which can accumulate to a significant amount over time.

tbekas (changes requested):

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -106,7 +107,7 @@ object GlobalSnapshotConsensusFunctions {
         updatedLastTxRefs = lastGS.info.lastTxRefs ++ acceptanceResult.contextUpdate.lastTxRefs
         updatedBalances = lastGS.info.balances ++ acceptanceResult.contextUpdate.balances
 
-        rewards = Rewards.calculateRewards(lastGS.proofs.map(_.id))
+        rewards <- Rewards.calculateRewards(currentOrdinal, 1, lastGS.proofs.map(_.id))
```
tbekas, line 110: Balances need to be updated as well.

tbekas (changes requested):

modules/core/src/main/scala/org/tessellation/domain/snapshot/rewards/Rewards.scala
TheMMaciek: Even if the time trigger would change at some point then we can still calculate new value to countdown from and reach next epoch at the same moment as we planned for before, so it seems like a good idea.

tbekas (changes requested):
It's a partial review. The code is long and contains lots of mathematical operations - which mostly I think are unnecessary. I will try to finish the review in the second part of the day.

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -0,0 +1,197 @@
...
+        timeTriggerInterval: FiniteDuration,
+        facilitators: NonEmptySet[Id]
+      ): EitherT[F, RewardsError, (Map[Address, Amount], SortedSet[RewardTransaction])] =
+        EitherT
+          .fromEither(getRewardsPerSnapshot(snapshotOrdinal, timeTriggerInterval))
+          .biflatMap(
+            {
+              case LastEpochExceeded =>
+                EitherT.liftF {
+                  logger.warn("Couldn't distribute the rewards. Last epoch exceeded.") >>
```
marcinwadon, line 67: if last epoch exceeded, there should be no rewards I think 🤔

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/DTM.scala
```diff
@@ -0,0 +1,46 @@
...
+    private final val perSnapshot = Amount(
+      NonNegLong.unsafeFrom(
+        config.monthly.value / ((30 * 24 * 60) / timeTriggerInterval.toMinutes)
+      )
+    )
+
+    def getAddress: Address = config.address
+
+    def weight(distribution: NonEmptyMap[Address, Amount]): Either[RewardsError, NonEmptyMap[Address, Amount]] = {
+      val total = distribution.map(_.value.toLong).reduceLeft(_ + _)
```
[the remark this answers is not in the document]
kpudlik: `plus` operator exists only for `Balance`, not for `Amount`.
But I will add it there as well

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
kpudlik: And there are no rewards. We just return empty `Set`:

```scala
              case LastEpochExceeded =>
                EitherT.liftF {
                  logger.warn("Couldn't distribute the rewards. Last epoch exceeded.") >>
                    (Map.empty[Address, Amount], SortedSet.empty[RewardTransaction]).pure[F]
                }
```

I can change the warning to be more descriptive e.g. "Last epoch exceeded. Won't mint rewards" or sth

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/DTM.scala
```diff
@@ -0,0 +1,46 @@
...
+      val reduced = total - perSnapshot.value
```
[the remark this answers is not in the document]
kpudlik: It is already handled few lines below:


```scala
      val reduced = total - perSnapshot.value

      NonNegLong
        .from(
          if (reduced >= 0) {
            reduced / distribution.length
          } else
            0L
        )
```

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -112,7 +120,25 @@ object GlobalSnapshotConsensusFunctions {
         updatedLastTxRefs = lastGS.info.lastTxRefs ++ acceptanceResult.contextUpdate.lastTxRefs
         updatedBalances = lastGS.info.balances ++ acceptanceResult.contextUpdate.balances
 
-        rewards = Rewards.calculateRewards(lastGS.proofs.map(_.id))
+        (updatedBalancesWithRewards, rewardTxs) <- trigger match {
+          case EventTrigger => (updatedBalances, SortedSet.empty[RewardTransaction]).pure[F]
+          case TimeTrigger =>
+            rewards
+              .calculateRewards(currentTimeTriggerOrdinal, lastGS.proofs.map(_.id))
```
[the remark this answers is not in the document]
kpudlik: Previously we were rewarding the previous one but rewarding data-triggered snapshot is definitely wrong. We would need to have the information about the previous trigger type here so we can decide to reward previous facilitators or not.

kpudlik:
> It's a partial review. The code is long and contains lots of mathematical operations - which mostly I think are unnecessary. I will try to finish the review in a few hours.

What do you mean by saying "unnecessary"? It covers the feature parity. I could have hardcoded the reward per snapshot but then we would need to calculate it somewhere else e.g. in some Google Sheet. It gives more overhead for us and changes nothing imo. Just moves the responsibility to another software if this is what you meant.

tbekas:
>What do you mean by saying "unnecessary"? It covers the feature parity. I could have hardcoded the reward per snapshot but then we would need to calculate it somewhere else e.g. in some Google Sheet.

They're unnecessary for two reasons:
* the outcome doesn't change for fixed parameters
* lengthy calculation like there is currently in the code is very error prone - having a literal value like `123456L` DAGs per snapshot is very resilient to errors, additionally we can validate this number in some unit test

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -112,7 +120,25 @@ object GlobalSnapshotConsensusFunctions {
...
+                case (distribution, rewardTxs) =>
+                  Async[F].fromEither {
+                    (updatedBalances.keySet ++ distribution.keySet).toList.traverse { address =>
+                      ((updatedBalances.get(address), distribution.get(address)) match {
+                        case (Some(balance), Some(reward)) => balance.plus(reward)
+                        case (Some(balance), None)         => balance.asRight[BalanceArithmeticError]
+                        case (_, Some(reward))             => Balance.fromAmount(reward).asRight[BalanceArithmeticError]
+                        case _                             => Balance.empty.asRight[BalanceArithmeticError]
+                      }).map((address, _))
+                    }.map(SortedMap.from(_)).map((_, rewardTxs))
```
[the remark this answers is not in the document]
kpudlik: > into another (non-effectful) function

It has to be effectful due to effectful nature of `plus` method. Unless you burn but I am not sure if we should make such decisions right now as it changes the feature parity.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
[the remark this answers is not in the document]
tbekas: Just the information about the trigger type of the previous snapshot isn't sufficient. We would have to have facilitators of the previous `TimeTrigger`ed snapshot, which is not in the previous snapshot (there can be multiple data snapshots in-between `TimeTrigger`ed snapshots).

modules/core/src/test/scala/org/tessellation/infrastructure/rewards/RewardsSuite.scala
```diff
@@ -0,0 +1,157 @@
...
+      KryoSerializer.forAsync[IO](dagSharedKryoRegistrar.union(sdkKryoRegistrar)),
+      SecurityProvider.forAsync[IO],
+      Resource.liftK(Random.scalaUtilRandom[IO])
+    ).mapN((_, _, _))
+
+  implicit val order: Ordering[Id] = (x: Id, y: Id) => 1
+
+  val config = RewardsConfig(
+    epochs = 4,
+    epochDurationInYears = 2.5,
```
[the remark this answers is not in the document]
kpudlik: Yes and no.

Duration in ordinals (aka counter-up) depends on `timeTriggerInterval`. According to my knowledge the value may change in the future so algorithm need to adjust the counter-up to match 2,5 year

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/DTM.scala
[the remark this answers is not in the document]
kpudlik: But I will use `minus` method anyway

modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/GlobalSnapshot.scala
```diff
@@ -33,6 +33,7 @@ case class GlobalSnapshot(
   blocks: SortedSet[BlockAsActiveTip],
   stateChannelSnapshots: SortedMap[Address, NonEmptyList[StateChannelSnapshotBinary]],
   rewards: SortedSet[RewardTransaction],
+  timeTriggerOrdinal: SnapshotOrdinal,
```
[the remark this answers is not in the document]
kpudlik: Fun fact. `@newtype` is missing in `SnapshotOrdinal` too

```
@derive(encoder, decoder, show)
case class SnapshotOrdinal(value: NonNegLong)
```

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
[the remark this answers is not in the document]
tbekas: It cannot stay this way because it's stopping the consensus. If you need to consult it with someone, please do, but not handling such a case is an error.

modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/GlobalSnapshot.scala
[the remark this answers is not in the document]
tbekas: I don't remember what was the exact cause, but as far as I can remember, we had some kind of issue with `@newtype` on this class. That was a long time ago however so maybe we didn't have sufficient experience to solve it.

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -0,0 +1,197 @@
...
+
+        }
+
+      private[rewards] def createSharedDistribution(
+        snapshotHash: Hash,
+        snapshotReward: Amount,
+        addresses: NonEmptySet[Address]
+      ): F[Either[RewardsError, NonEmptyMap[Address, Amount]]] = {
+        val totalSpace = NonNegInt.unsafeFrom(addresses.length)
+        val (quotient, remainder) = snapshotReward.coerce.toLong /% totalSpace.toLong
```
tbekas, line 159: If you inline the `totalSpace` you get:
```
NonNegInt.unsafeFrom(addresses.length).toLong
```
which is doing nothing and is equivalent to just `address.length`.

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -0,0 +1,197 @@
...
+
+        val quotientDistribution = addresses.toNonEmptyList.map(_ -> quotient).toList.toMap.pure[F]
```
tbekas, line 161: Currently `quotientDistribution` is regular `Map`. You can get `NonEmptyMap` if you do
```
addresses.toNonEmptyList.map(_ -> quotient).toNem
```

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -0,0 +1,197 @@
...
+
+        val remainderDistribution = if (remainder > 0) {
+          Random.scalaUtilRandomSeedInt(snapshotHash.value.toInt).flatMap { random =>
+            random.shuffleList(addresses.toNonEmptyList.toList).map { shuffled =>
+              shuffled.take(remainder.toInt).map(_ -> remainder / remainder).toMap
```
tbekas, line 166: `reminder / reminder` always has to be `1`, I think you can safely hardcode it 🙂

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -0,0 +1,197 @@
...
+            }
+          }
+        } else Map.empty[Address, Long].pure[F]
+
+        (quotientDistribution, remainderDistribution)
+          .mapN(_ |+| _)
+          .map(_.view.mapValues(NonNegLong.unsafeFrom).mapValues(Amount(_)).toList)
+          .map(map => NonEmptyMap.fromMapUnsafe(SortedMap.from(map)))
+          .map(_.asRight[RewardsError])
+          .handleError(error => RewardsException(error.getMessage).asLeft[NonEmptyMap[Address, Amount]])
```
tbekas, line 176: Lot's of unsafe and throwing operations, where there shouldn't be many. I think you can rewrite it to:

```
val quotientDistribution = addresses.toNonEmptyList.map(_ -> NonNegLong.unsafeFrom(quotient)).toNem

Random.scalaUtilRandomSeedInt(snapshotHash.value.hashCode).flatMap { random =>
  random
    .shuffleList(addresses.toList)
    .map { shuffled =>
      shuffled.take(remainder.toInt)
    }
    .map { top =>
      top.toNel
        .fold(quotientDistribution) { topNel =>
          topNel.map(_ -> NonNegLong(1L)).toNem |+| quotientDistribution
        }
        .mapBoth { case (addr, reward) => (addr, Amount(reward)) }
    }
}
```

tbekas (changes requested):
Another batch of comments. I'm still not finished yet with the whole review.

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
kpudlik: Ah yes. It looked differently previously and I just changed the divisor to `remainder` and didn't notice that I ended up with `x/x` :D 

With `if (remainder > 0)` I can safely change it to `1` (actually without that check it would also throw on 0/0 anyway).

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/DTM.scala
```diff
@@ -0,0 +1,51 @@
...
+          .toRight[RewardsError](IgnoredAllAddressesDuringWeighting)
+
+        ignored <- NonEmptyMap
+          .fromMap(withoutIgnored.keys.foldLeft(distribution.toSortedMap) {
+            case (acc, nonIgnoredAddress) => acc - nonIgnoredAddress
+          })
+          .toRight[RewardsError](IgnoredAllAddressesDuringWeighting)
+
+        total = withoutIgnored.map(_.value.toLong).reduceLeft(_ + _)
+        perSnapshot = Math.floorDiv(Math.floorDiv(config.monthly.value, 30 * 24 * 60), timeTriggerInterval.toMinutes)
```
tbekas, line 38: Is this a correct calculation at all? According to this code the longer the time trigger the lower the rewards (since time trigger is in the denominator). It should be the other way around I think. Probably it works currently only by accident because the time trigger is `1.minutes`, and dividing by `1` or multiplying by `1` doesn't change the outcome.

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/DTM.scala
```diff
@@ -0,0 +1,51 @@
...
+import eu.timepit.refined.auto._
+import eu.timepit.refined.types.numeric._
+
+object DTM {
+
+  def make(config: DTMConfig, timeTriggerInterval: FiniteDuration): DTM = new DTM {
+
+    def getAddress: Address = config.address
+
+    def weight(
```
tbekas, line 21: Please replace the `weight` method with this:
```
def applyDtm(amount: Amount): Either[ArithmeticException, (Amount, NonEmptyMap[Address, Amount])] = {
  val epochDurationSeconds = NonNegLong.unsafeFrom(timeTriggerInterval.toSeconds)
  val monthInSeconds: PosLong = 30 * 24 * 60 * 60 // approx
  for {
    dtmRewards <- (config.monthly.value * epochDurationSeconds).map(_ / thirtyDaysInSeconds)
    remainingRewards <- amount.value - dtmRewards
  } yield (Amount(remainingRewards), NonEmptyMap.one(config.address, Amount(dtmRewards))
}
```

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -0,0 +1,153 @@
...
+  ): Rewards[F] =
+    new Rewards[F] {
+
+      implicit val logger = Slf4jLogger.getLogger[F]
+
+      def calculateRewards(
+        epochProgress: EpochProgress,
+        facilitators: NonEmptySet[Id]
+      ): F[SortedSet[RewardTransaction]] =
+        EitherT
```
tbekas, line 48: No need for `EitherT` in this method. Regular `Either` lifted to `F` as the last step should be sufficient.

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -0,0 +1,153 @@
...
+          .fromEither(getRewardsPerSnapshot(epochProgress))
+          .flatMap { snapshotReward =>
+            EitherT.liftF {
+              facilitators.toNonEmptyList
+                .traverse(_.toAddress)
+                .map(_.toNes)
+            }.flatMap(createSharedDistribution(epochProgress, snapshotReward, _))
+          }
+          // Note: Weights need to be validated
```
tbekas, line 57: Please compose the `apply*` methods in the following way:
1. Stardust collective (-10%)
2. DTM
3. Soft Staking
4. Distribution to facilitators

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -0,0 +1,153 @@
...
+          }
+          .toRight(LastEpochExceeded)
+
+      private[rewards] def createSharedDistribution(
+        epochProgress: EpochProgress,
+        snapshotReward: Amount,
+        addresses: NonEmptySet[Address]
+      ): EitherT[F, RewardsError, NonEmptyMap[Address, Amount]] = {
+
+        val (quotient, remainder) = snapshotReward.coerce.toLong /% addresses.length.toLong
```
tbekas, line 114: You can remove the `.toLong` once you have the `NonNegLong` arithmetic.

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -0,0 +1,153 @@
...
+          Random.scalaUtilRandomSeedLong(epochProgress.coerce).flatMap { random =>
+            random
+              .shuffleList(addresses.toNonEmptyList.toList)
+              .map { shuffled =>
+                shuffled.take(remainder.toInt)
+              }
+              .map { top =>
+                top.toNel
+                  .fold(quotientDistribution) { topNel =>
+                    topNel.map(_ -> NonNegLong(1L)).toNem |+| quotientDistribution
```
tbekas, line 128: I think that combining rewards isn't correct. If there are different rewards from different programs to the same address I believe it should be separate transactions instead of a single transaction for a combined amount.

modules/core/src/test/scala/org/tessellation/infrastructure/rewards/RewardsSuite.scala
```diff
@@ -0,0 +1,135 @@
...
+import org.tessellation.sdk.sdkKryoRegistrar
+import org.tessellation.security.SecurityProvider
+import org.tessellation.security.hash.Hash
+import org.tessellation.security.hex.Hex
+
+import eu.timepit.refined.auto._
+import weaver.MutableIOSuite
+import weaver.scalacheck.Checkers
+
+object RewardsSuite extends MutableIOSuite with Checkers {
```
tbekas, line 27: I haven't review this test yet, so there might be additional comments regarding it.

tbekas (changes requested):

tbekas:
Additionally I think that combining rewards isn't correct. If there are different rewards from different programs to the same address I believe it should be separate transactions instead of a single transaction for a combined amount.

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
marcinwadon: Composition is from right to left

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/StardustCollective.scala
```diff
@@ -0,0 +1,43 @@
...
+
+import eu.timepit.refined.auto._
+import eu.timepit.refined.types.numeric.NonNegLong
+
+object StardustCollective {
+
+  def make(config: StardustConfig): StardustCollective = new StardustCollective {
+    def getAddress: Address = config.address
+
+    def weight(
```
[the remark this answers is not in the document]
marcinwadon: Why do you implicitly state 10L as a 100%? Imo it could be 0.1 with refinement 0-1. In case of 0-10 you'd have to set a refinement too imo. Also, why the remaining weight cannot be calculated? What if I'll set it 1 and 8?

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/SoftStaking.scala
```diff
@@ -0,0 +1,78 @@
...
+
+import eu.timepit.refined.auto._
+import eu.timepit.refined.types.numeric.{NonNegInt, NonNegLong}
+
+object SoftStaking {
+
+  def make(config: SoftStakingConfig): SoftStaking = new SoftStaking {
+    def getAddress: Address = config.address
+
+    def weight(softStakingNodes: NonNegInt)(
```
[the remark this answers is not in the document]
marcinwadon: You can take a look into CL repo how testnet is rewarded because this is what we need to mimic 1:1

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/StardustCollective.scala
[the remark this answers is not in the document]
tbekas: I think you're misreading this. These are weights, there's no assumption that `10L` is 100%. It could be `1L` and `2L` giving total of 3L and that would be a 100%. 

>Imo it could be 0.1 with refinement 0-1.

0.1 is not an integer number. We should do calculations only on integer numbers.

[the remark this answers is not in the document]
marcinwadon: Ok, got it :+1:

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
tbekas: True, I meant sequencing computations.

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/StardustCollective.scala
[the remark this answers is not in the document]
marcinwadon: So it's more like a fraction, not weight 🤔

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/SoftStaking.scala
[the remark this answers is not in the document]
tbekas: As far as I can tell the owner of the testnet node receives the same amount of rewards as the owner of the staked dags. If that's the case, than this code should handle it:
```scala
def applySoftStakingAndTestnet(
  amount: Amount,
  facilitatorCount: NonNegLong
): Either[ArithmeticException, (Amount, NonEmptyMap[Address, Amount])] =
  for {
    f <- facilitatorCount * config.facilitatorWeight
    s <- config.softStakeCount * config.softStakeWeight
    t <- config.testnetCount * config.testnetWeight

    numeratorS <- amount.coerce * s
    numeratorT <- amount.coerce * t
    denominator <- f + s + t

    softStakingRewards <- numeratorS / denominator
    testnetRewards <- numeratorT / denominator
    facilitatorRewards <- (amount.value - softStakingRewards).flatMap(_ - testnetRewards)
  } yield
    (
      Amount(facilitatorRewards),
      NonEmptyMap(config.softStakeAddress -> Amount(softStakingRewards), SortedMap(config.testnetAddress -> Amount(testnetRewards)))
    )
```

With this config (without the default params of course):
```scala
case class SoftStakingAndTestnetConfig(
  softStakeAddress: Address,
  testnetAddress: Address,
  startingOrdinal: SnapshotOrdinal,
  testnetCount: NonNegInt,
  testnetWeight: NonNegLong = 4L,
  softStakeCount: NonNegLong,
  softStakeWeight: NonNegLong = 4L,
  facilitatorWeight: NonNegLong = 6L
)
```

modules/core/src/test/scala/org/tessellation/infrastructure/rewards/RewardsSuite.scala
```diff
@@ -0,0 +1,135 @@
...
+      ),
+      Id(
+        Hex(
+          "c7f9a08bdea7ff5f51c8af16e223a1d751bac9c541125d9aef5658e9b7597aee8cba374119ebe83fb9edd8c0b4654af273f2d052e2d7dd5c6160b6d6c284a17c"
+        )
+      )
+    )
+  )
+
+  test("epoch 1 snapshot reward is valid") { res =>
```
tbekas, line 61: I like these tests very much, however instead of having a test that's copied and pasted 4 times with different parameters, please create a test that generates`EpochProgress` based on `Gen.chooseNum`. `chooseNum` which has the following signature:

```scala
def chooseNum[T](minT: T, maxT: T, specials: T*)
```

this allows you to specify `specials` which are the values on the boundary of the epoch (these are keys in the `RewardsConfig.rewardsPerEpoch`). Also please do not hardcode expected values in the test (take them from `rewardsPerEpoch`).

You can create an additional test to check if the properties of the `rewardsPerEpoch` hold. See the next comment.

modules/core/src/test/scala/org/tessellation/infrastructure/rewards/RewardsSuite.scala
```diff
@@ -0,0 +1,135 @@
...
+      rewards <- Rewards.make[IO](config, 1.minutes, softStaking, dtm, stardust).pure[F]
+      txs <- rewards
+        .calculateRewards(
+          EpochProgress(1296001L * 4),
+          facilitators
+        )
+      sum = txs.toList.map(_.amount.value.toLong).sum
+
+    } yield expect(sum == 0)
+  }
```
tbekas, line 134: I think that there should be a test that checks if:

The sum of rewards in `RewardsConfig.rewardsPerEpoch` sums up the expected value which is `160000000000000000L`.

It might be the case that it won't (I haven't checked it), but in that case we have 2 options:
* change the value from `160000000000000000L` to one that's possible to distribute having the current length of the epoch and a constraint that each next epoch has exactly half of the rewards of the previous epoch (this allows us also to write a test that verifies this constraint)
* introduce a short 'reminder epoch' at the end of each epoch (or at the end of all epochs), where we distribute missing dags, so that they all sum up to the `160000000000000000L`

I think that the 1st option is better since it simplifies stuff (and this is how probably it works currently in the mainnet), but it has to be probably communicated to the CL management.

modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/epoch.scala
```diff
@@ -0,0 +1,36 @@
...
+import derevo.cats.show
+import derevo.circe.magnolia.{decoder, encoder}
+import derevo.derive
+import eu.timepit.refined.cats._
+import eu.timepit.refined.types.numeric.NonNegLong
+import io.estatico.newtype.macros.newtype
+
+object epoch {
+
+  @derive(encoder, decoder, show)
```
tbekas, line 19: Please add `order` here, or both `order` and `ordering` if needed

modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/epoch.scala
```diff
@@ -0,0 +1,36 @@
...
+  @newtype
+  case class EpochProgress(value: NonNegLong)
+
+  object EpochProgress {
+    val MinValue: EpochProgress = EpochProgress(NonNegLong.MinValue)
+
+    implicit val eqv: Eq[EpochProgress] = Eq[NonNegLong].contramap(_.value)
+    implicit val order: Order[EpochProgress] = Order[NonNegLong].contramap(_.value)
+    implicit val ordering: Ordering[EpochProgress] = order.toOrdering
```
tbekas, line 28: These instances can be autogenerated. Please look at the `Height` or `SnapshotOrdinal`.

modules/core/src/main/scala/org/tessellation/config/types.scala
```diff
@@ -41,8 +53,72 @@ object types {
...
+    rewardsPerEpoch: NonEmptyMap[EpochProgress, Amount]
+  )
+
+  object RewardsConfig {
+
+    val default = RewardsConfig(
+      epochs = 4,
+      epochDurationInYears = 2.5,
+      totalPool = Amount(160000000000000000L),
+      baseEpochReward = Amount(85333333320000000L),
```
tbekas, line 96: Unused config values. The `totalPool` can be hardcoded in the `RewardsSuite` since it will be the only place using it.

modules/core/src/main/scala/org/tessellation/config/types.scala
```diff
@@ -41,8 +53,72 @@ object types {
...
+      ),
+      dtm = DTMConfig(
+        address = Address("DAG0Njmo6JZ3FhkLsipJSppepUHPuTXcSifARfvK"),
+        monthly = Amount(200000000000000L)
+      ),
+      stardust = StardustConfig(
+        address = Address("DAGSTARDUSTCOLLECTIVEHZOIPHXZUBFGNXWJETZVSPAPAHMLXS"),
+        percentage = 10
+      ),
+      rewardsPerEpoch = NonEmptyMap(
```
tbekas, line 113: `NonEmptyMap` isn't needed here. `SortedMap` is sufficient.

tbekas (changes requested):
I've finished the review now.

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -0,0 +1,153 @@
...
+import org.tessellation.schema.transaction.{RewardTransaction, TransactionAmount}
+import org.tessellation.security.SecurityProvider
+
+import eu.timepit.refined.auto._
+import eu.timepit.refined.cats._
+import eu.timepit.refined.types.numeric._
+import io.estatico.newtype.ops.toCoercibleIdOps
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+
+object Rewards {
```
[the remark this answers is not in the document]
kpudlik: It is part of the bigger story. There is no way of course to derive `Numeric[A]` because some operations on refined types are not pure. I was thinking about creating a trait which is impure and ops return `Either[ArithmeticException, A]` and then implement that for the numeric types we use so the same as you suggested but in more generic way to keep it DRY. I decided to drop that idea because then we would need to refactor our code to use that ops everwhere instead of for example `Semigroup` in some places for adding 2 refined numbers.

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -0,0 +1,153 @@
...
+        facilitators: NonEmptySet[Id]
+      ): F[SortedSet[RewardTransaction]] =
+        EitherT
+          .fromEither(getRewardsPerSnapshot(epochProgress))
+          .flatMap { snapshotReward =>
+            EitherT.liftF {
+              facilitators.toNonEmptyList
+                .traverse(_.toAddress)
+                .map(_.toNes)
+            }.flatMap(createSharedDistribution(epochProgress, snapshotReward, _))
```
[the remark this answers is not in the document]
kpudlik: Seems ok but then to compose the methods we would need to return something like this:

```scala
case class Distribution(distribution: Map[Address, Amount], rest: Amount)
```

and then each method would take from `rest` and add addresses to `distribution`. Actually we could potentially get rid of `ignore` if we have the current distribution separately but only when the methods always take from `rest` and never from `distribution+rest`. If each composed method always ignores all the special addresses from previous methods then it sounds good to me.

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
kpudlik: To sequence methods in such way I need to change the output type of weight functions but as I said in the comment above - it will have some benefits.

modules/core/src/test/scala/org/tessellation/infrastructure/rewards/RewardsSuite.scala
kpudlik: Yeah. We were discussing the need of keeping total pool in snapshot but imo we should just keep that value in a suite and write another test to not make the Snapshot bigger (I know, it is just a Long value but still, the less we keep in snapshot the better)

I would go for 1st option.

modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/epoch.scala
kpudlik: `SnapshotOrdinal`:

```scala
@derive(encoder, decoder, show)
case class SnapshotOrdinal(value: NonNegLong)

object SnapshotOrdinal {
  val MinValue: SnapshotOrdinal = SnapshotOrdinal(NonNegLong.MinValue)

  implicit val order: Order[SnapshotOrdinal] = Order[NonNegLong].contramap(_.value)
  implicit val ordering: Ordering[SnapshotOrdinal] = order.toOrdering
```

From what I remember there was a problem with deriving these values and as you can see `SnapshotOrdinal` also uses `contramap`. But I will investigate the problem. For `Height` it works because it is just `Long` (unless you changed that to `NonNegLong` finally when I was off. I will pull the latest changes).

modules/core/src/main/scala/org/tessellation/config/types.scala
kpudlik: I assumed that there must be at least 1 epoch. I can make it `SortedMap` and just mint no rewards if that's the case.
Having `NonEmptyMap` was just fitting better for expressing the business case.

kpudlik:
> Additionally I think that combining rewards isn't correct. If there are different rewards from different programs to the same address I believe it should be separate transactions instead of a single transaction for a combined amount.

It will be a breaking change because I will need to rewrite the weighting process so it creates transactions instead of `Amount`. I am fine with that and it seems reasonable but on the other hand we never discussed that so I just left it as combined as it was in feature parity. Imo it is not a must have for mainnet release and we can change it later.

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -0,0 +1,153 @@
...
+          }
+          .find {
+            case ((_, _), (curr, _)) => epochProgress.value <= curr.value
+          }
+          .map {
+            case ((prevOrdinal, prevR), (_, currR)) =>
+              if (epochProgress.value <= prevOrdinal.value) prevR
+              else currR
+          }
+          .toRight(LastEpochExceeded)
```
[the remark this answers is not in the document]
kpudlik: `minAfter` is the operator I was looking for 👍🏻

modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/epoch.scala
kpudlik: Update: I fixed both `SnapshotOrdinal` and `EpochProgress` so it will be derived automatically.

tbekas:
>The issue you mentioned will happen if any of special addresses (DTM, Stardust, SoftStaking) is used in more than 1 process and such case won't happen.

There's no such guarantee. Also, there's nothing stopping someone who has access to those keys (DTM or SoftStaking) to start a node with that key. It's not a difficult change, just replace the return type from `(Address, NonEmptyMap[Address, Amount])` with `(Address, List[RewardTransaction])` and concatenate the lists from different programs.

marcinwadon:
but why?

modules/core/src/main/scala/org/tessellation/config/types.scala
tbekas: I don't see any reason why the type should guarantee that there is at least one epoch.

modules/core/src/test/scala/org/tessellation/infrastructure/rewards/RewardsSuite.scala
tbekas: >We were discussing the need of keeping total pool in snapshot 

I think we were discussing keeping the total amount of rewards minted so far in the snapshot. Copying a value ~`160000000000000000L` from snapshot to snapshot wouldn't make much sense.

marcinwadon:
Rewards are minted and the only reason why RewardTransaction exists is the need to visualize that. We don't distinguish different types of rewards and having multiple transactions does not give anything meaningful- you're not even able to say which transaction is from which reward algorithm. It's fine to keep it accumulated per address per snapshot, because that's was was required to achieve.

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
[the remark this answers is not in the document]
tbekas: Use any kind of abstraction you want, just as long as it's computationally safe and doesn't make a simple arithmetic operation too verbose. Using the abstraction proposed above it took just 6 lines of code to implement startust 10% share.

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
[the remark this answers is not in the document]
tbekas: Yep, this class `Distribution` is pretty much equivalent to the return type from the `apply*` functions provided below. I think that `List[RewardTransaction]` is a better choice than `Map[Address, Amount]` though.

tbekas:
>but why?

1. In the edge case, someone may not be able to receive `10L` as a single transaction, but if we keep it as separate transactions, let's say `4L` and `6L` he may be able to receive one of them.
2. If we decide to add a `remark` field to transactions in the future (which exists on many competing chains), it will be easier to implement it.
3. It's virtually no additional cost on the programming side (the cost is actually lower because there's no need to combine different maps of type `Map[Address, Amount]`)

marcinwadon:
I'd opt for focusing on important things first and deliver the functionality instead of thinking of a thousands of smaller improvements, because we're already late with that. We can always implement that later and imo it's nice to have, if really, instead of must have.

modules/core/src/test/scala/org/tessellation/infrastructure/rewards/RewardsSuite.scala
kpudlik: Right. It was about already minted value.

tbekas:
True, that's the 4th reason I guess, since It's less work to keep transactions separate (no need to sum up the rewards with handling possible errors there).

PR #294, opened 2022-06-17
Add collateral verification in Joining.

kriscieslak:


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/snapshot/programs/SnapshotProcessor.scala
```diff
@@ -120,6 +130,7 @@ sealed abstract class SnapshotProcessor[F[_]: Async: KryoSerializer: SecurityPro
             tipsToRemove = tipsToRemove
           )
           .flatMap(_ => lastGlobalSnapshotStorage.set(globalSnapshot))
+          .flatMap(_ => collateralUpdateBalances(globalSnapshot))
```
kpudlik, line 133: If you ignore returned values anyway then `flatTap`

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/snapshot/programs/SnapshotProcessor.scala
```diff
@@ -136,6 +147,7 @@ sealed abstract class SnapshotProcessor[F[_]: Async: KryoSerializer: SecurityPro
...
```
kpudlik, line 150: If you ignore returned values anyway then `flatTap`

modules/core/src/main/scala/org/tessellation/modules/Services.scala
```diff
@@ -64,5 +67,6 @@ sealed abstract class Services[F[_]] private (
   val metrics: Metrics[F],
   val gossip: Gossip[F],
   val consensus: Consensus[F, GlobalSnapshotEvent, GlobalSnapshotKey, GlobalSnapshotArtifact],
-  val dag: DAGService[F]
+  val dag: DAGService[F],
+  val collateralService: CollateralService[F]
```
marcinwadon, line 71: Can you name it accordingly to the pattern, please? without the Service suffix. The same for trait/class name

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/domain/snapshot/programs/SnapshotProcessorSuite.scala
```diff
@@ -40,6 +40,7 @@ import eu.timepit.refined.auto._
 import eu.timepit.refined.types.numeric.NonNegLong
 import io.chrisdavenport.mapref.MapRef
 import weaver.SimpleIOSuite
+import org.tessellation.sdk.infrastructure.CollateralService
```
marcinwadon, line 43: it will throw on linting 🤔

modules/sdk/src/main/scala/org/tessellation/sdk/domain/collateral/CollateralService.scala
```diff
@@ -0,0 +1,9 @@
+package org.tessellation.sdk.domain.collateral
+
+import org.tessellation.schema.address.Address
+import org.tessellation.schema.balance.Balance
+
+trait CollateralService[F[_]] {
+  def updateBalances(no: Long, balances: Map[Address, Balance]): F[Unit]
+  def verifyCollateral(addres: Address): F[Boolean]
```
marcinwadon, line 8: typo `addres` -> `address`

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/collateral/CollateralService.scala
```diff
@@ -0,0 +1,64 @@
...
+        .fromQueueUnterminated(updateBalanceQueue)
+        .filterWithPrevious {
+          case ((previousNo, _), (currentNo, _)) => previousNo < currentNo
+        }
+        .map {
+          case (_, balances) => balances
+        }
+        .evalMap(incomingBalances => {
+          val (improperBalances, properBalances) =
+            incomingBalances.partition { case (_, balance) => balance.value < numeric.NonNegLong(250000000L) }
```
marcinwadon, line 41: please don't hardcode such values, it should be parametrized and taken from configuration.
Or - written in genesis block and consecutive snapshots like "network metadata". What you guys think about that? @tbekas @kpudlik @TheMMaciek

Having "network metadata" we could enforce using the same values across the network and having it in snapshot would allow to change it in the go. I'm concerned about repeating the same values over and over again 🤔 any better idea?

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
```diff
@@ -146,6 +153,11 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
       _ <- if (!canJoinCluster) NodeStateDoesNotAllowForJoining(nodeState).raiseError[F, Unit]
       else Applicative[F].unit
 
+      address <- toPeer.id.value.toPublicKey.map(_.toAddress)
```
kpudlik, line 156: I have the same line in my Rewards code so I suggest you to do this:

```scala
  @derive(decoder, encoder, eqv, show, order)
  @newtype
  case class Id(hex: Hex) {
    def toPublicKey[F[_]: Async: SecurityProvider]: F[PublicKey] = hex.toPublicKey

    def toAddress[F[_]: Async: SecurityProvider]: F[Address] = toPublicKey.map(_.toAddress)
  }
```

and just `toPeer.id.toAddress` as such conversion is used very often.

marcinwadon (changes requested):
Overall looks good to me, but I think you don't need to store balances on a side and update it for each global snapshot. Instead, you can query global snapshot directly for the balances. Latest global snapshot is always loaded in memory, so you don't need to copy those values into another storage 🤔

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/collateral/CollateralService.scala
kpudlik: Yeah, I was initially thinking about just lifting that value in the file so it is not in between the lines but keeping it in genesis is also fine.

kpudlik: On the other hand repeating it in every snapshot is redundant so maybe hardcoding in the code is better.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/collateral/LatestBalances.scala
```diff
@@ -0,0 +1,8 @@
+package org.tessellation.sdk.domain.collateral
+
+import org.tessellation.schema.address.Address
+import org.tessellation.schema.balance.Balance
+
+trait LatestBalances[F[_]] {
+  def getLatestBalances: F[Option[Map[Address, Balance]]]
```
marcinwadon, line 7: IMO it's pointless to expose that. It can be an internal function of Collateral service that would depend on global snapshot storage

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/collateral/Collateral.scala
```diff
@@ -0,0 +1,29 @@
...
+    new Collateral[F] {
+
+      override def hasCollateral(peerId: PeerId): F[Boolean] =
+        for {
+          address <- peerId.value.toPublicKey.map(_.toAddress)
+          balances <- latestBalances.getLatestBalances
+        } yield
+          balances.flatMap(_.get(address)) match {
+            case None          => false
+            case Some(balance) => balance.value >= 250000000
```
marcinwadon, line 26: As previously commented

marcinwadon (changes requested):

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/LatestBalancesStorage.scala
```diff
@@ -0,0 +1,19 @@
...
+import cats.syntax.functor._
+
+import org.tessellation.domain.snapshot.GlobalSnapshotStorage
+import org.tessellation.schema.address.Address
+import org.tessellation.schema.balance.Balance
+import org.tessellation.sdk.domain.collateral.LatestBalances
+
+object LatestBalancesStorage {
+
+  def make[F[_]: Async](globalSnapshotStorage: GlobalSnapshotStorage[F]) =
```
marcinwadon, line 13: if you want, you can mix it together with GlobalSnapshotStorage, but on implementation level (not in the trait)

kriscieslak: Do you mean doing sth like this:
 ````
object GlobalSnapshotStorage {

  def make[F[_]: Async: KryoSerializer](
    globalSnapshotLocalFileSystemStorage: GlobalSnapshotLocalFileSystemStorage[F],
    inMemoryCapacity: NonNegLong
  ): F[GlobalSnapshotStorage[F] with LatestBalances[F]] = {
````
and then 
```
`object Storages {

  def make[F[_]: Async: KryoSerializer](
    sdkStorages: SdkStorages[F],
    snapshotConfig: SnapshotConfig
  ): F[Storages[F]] =
    for {
     [...]
      globalSnapshotStorage <- GlobalSnapshotStorage
        .make[F](globalSnapshotLocalFileSystemStorage, snapshotConfig.inMemoryCapacity)
      latestBalancesStorage = globalSnapshotStorage`
```
Or sth like this:
```
`object Services {

  def make[F[_]: Async: Random: KryoSerializer: SecurityProvider](
   [...]
  ): F[Services[F]] =
    for {
      metrics <- Metrics.make[F]
      consensus <- GlobalSnapshotConsensus
        .make[F](
          [...]
        )
      latestBalancesStorage = new LatestBalances {
        def getLatestBalances = storages.globalSnapshot.head.map(_.map(_.info.balances))
      }
      collateralService = Collateral.make[F](cfg.collateral,latestBalancesStorage)
```

File: modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala [hunk 11]
```diff
@@ -152,14 +152,17 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
...
+  def joinRequest(hasCollateral: PeerId => F[Boolean])(joinRequest: JoinRequest, remoteAddress: Host): F[Unit] =
     for {
       _ <- sessionStorage.getToken.flatMap(_.fold(SessionDoesNotExist.raiseError[F, Unit])(_ => Applicative[F].unit))
 
       registrationRequest = joinRequest.registrationRequest
 
       _ <- validateHandshakeConditions(joinRequest.registrationRequest)
 
+      _ <- hasCollateral(registrationRequest.id)
+        .ifM(Applicative[F].unit, CollateralNotSatisfied.raiseError[F, Unit])
```
TheMMaciek, line 164: `.unlessA`?

modules/sdk/src/main/scala/org/tessellation/sdk/domain/collateral/LatestBalances.scala
```diff
@@ -0,0 +1,8 @@
+package org.tessellation.sdk.domain.collateral
+
+import org.tessellation.schema.address.Address
+import org.tessellation.schema.balance.Balance
+
+trait LatestBalances[F[_]] {
```
tbekas, line 6: Do we actually need to track balances? Wouldn't be enough to keep track of a set of adresses that have more than 250k? On every snapshot we would add/remove addresses from that set. It would require less memory because we wouldn't store unnecessary information.

tbekas (changes requested):

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/LatestBalancesStorage.scala
marcinwadon: the first approach but then:

```scala
object Storages {

  def make[F[_]: Async: KryoSerializer](
    sdkStorages: SdkStorages[F],
    snapshotConfig: SnapshotConfig
  ): F[Storages[F]] =
    for {
     [...]
      globalSnapshotStorage <- GlobalSnapshotStorage
        .make[F](globalSnapshotLocalFileSystemStorage, snapshotConfig.inMemoryCapacity)
```

without

```scala
      latestBalancesStorage = globalSnapshotStorage
```
      
as globalSnapshotStorage already satisfies the `LatestBalances` trait, so this will work:

```scala
class Collateral[F[_]](latestBalances: LatestBalances[F]) { ... }

[...]

globalSnapshotStorage <- GlobalSnapshotStorage.make[F](...)
collateral = Collateral.make[F](cfg.collateral, globalSnapshotStorage)
```

modules/sdk/src/main/scala/org/tessellation/sdk/cli/CliMethod.scala
```diff
@@ -52,6 +53,8 @@ trait CliMethod {
     )
   )
 
+  val collateralConfig = CollateralConfig(Amount(25000000000000L))
```
marcinwadon, line 56: Please change it accordingly to @tbekas suggestion using `_` or even to keep it in more human-readable format like: `250_000L * NormalizationFactor`

modules/sdk/src/main/scala/org/tessellation/sdk/config/types.scala
```diff
@@ -69,4 +70,8 @@ object types {
     defaultCheckAttempts: PosInt,
     ensureCheckInterval: FiniteDuration
   )
+
+  case class CollateralConfig(
+    threshold: Amount
```
marcinwadon, line 75: I'd rename it into `minStaking` or `minStakingAmount`

modules/sdk/src/main/scala/org/tessellation/sdk/domain/collateral/LatestBalances.scala
marcinwadon: @kriscieslak please take a look here. our discussion related to LatestBalances trait could be irrelevant if above is true

modules/sdk/src/main/scala/org/tessellation/sdk/domain/collateral/LatestBalances.scala
marcinwadon: @tbekas we store balances cache anyway (either directly in global snapshot or later by applying diffs from light snapshots), so we would just reuse that (read-only). we definitely shouldn't store it twice

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/collateral/Collateral.scala
```diff
@@ -0,0 +1,30 @@
...
+
+      override def hasCollateral(peerId: PeerId): F[Boolean] =
+        for {
+          address <- peerId.value.toPublicKey.map(_.toAddress)
+          balances <- latestBalances.getLatestBalances
+        } yield
+          balances.flatMap(_.get(address)) match {
+            case None          => false
+            case Some(balance) => balance.value >= config.threshold.value
+          }
```
marcinwadon, line 28: what about using `.map(cond).getOrElse(false)` or `fold(false)(_ => cond)`?

```scala
balances.flatMap {
  _.get(address)
    .map(_ >= config.threshold)
    .getOrElse(false)
}
```

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/LatestBalancesStorage.scala
kriscieslak: But in this case I would also have to update `Storages` like this:
```scala
sealed abstract class Storages[F[_]] private (
  [...]
  val globalSnapshot: GlobalSnapshotStorage[F] with LatestBalances[F]
)
```
Is it ok?

modules/sdk/src/main/scala/org/tessellation/sdk/domain/collateral/LatestBalances.scala
tbekas: Makes sense 👍

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
```diff
@@ -152,14 +152,16 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
...
-  def joinRequest(joinRequest: JoinRequest, remoteAddress: Host): F[Unit] =
+  def joinRequest(hasCollateral: PeerId => F[Boolean])(joinRequest: JoinRequest, remoteAddress: Host): F[Unit] =
     for {
       _ <- sessionStorage.getToken.flatMap(_.fold(SessionDoesNotExist.raiseError[F, Unit])(_ => Applicative[F].unit))
 
       registrationRequest = joinRequest.registrationRequest
 
       _ <- validateHandshakeConditions(joinRequest.registrationRequest)
 
+      _ <- hasCollateral(registrationRequest.id).map(CollateralNotSatisfied.raiseError[F, Unit].unlessA)
```
marcinwadon, line 163: nice one! :+1:

marcinwadon (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/config/types.scala
TheMMaciek: Maybe just `amount` given that it's inside of `CollateralConfig` or `collateral` simply.

marcinwadon (dismissed):

modules/sdk/src/main/scala/org/tessellation/sdk/domain/collateral/CollateralService.scala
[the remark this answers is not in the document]
kriscieslak: Fixed.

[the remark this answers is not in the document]
kriscieslak: Changed.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/collateral/CollateralService.scala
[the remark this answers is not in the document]
kriscieslak: Changed as requested.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/snapshot/programs/SnapshotProcessor.scala
kriscieslak: Changed implementation - this code doesn't exist anymore.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/snapshot/programs/SnapshotProcessor.scala
kriscieslak: Changed implementation - this code doesn't exist anymore.

modules/core/src/main/scala/org/tessellation/modules/Services.scala
kriscieslak: Changed.

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/domain/snapshot/programs/SnapshotProcessorSuite.scala
kriscieslak: Fixed.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/collateral/CollateralService.scala
```diff
@@ -0,0 +1,9 @@
+package org.tessellation.sdk.domain.collateral
+
+import org.tessellation.schema.address.Address
+import org.tessellation.schema.balance.Balance
+
+trait CollateralService[F[_]] {
+  def updateBalances(no: Long, balances: Map[Address, Balance]): F[Unit]
```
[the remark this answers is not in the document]
kriscieslak: As above - this code doesn't exist anymore.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/collateral/CollateralService.scala
kriscieslak: Fixed.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/collateral/CollateralService.scala
kriscieslak: Threshold value extracted to `CollateralConfig` as `minStakingAmount`

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
```diff
@@ -152,14 +152,17 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
   private def validateHandshakeConditions(toPeer: PeerToJoin): F[Unit] =
     validateHandshakeConditions(toPeer.id, toPeer.ip, toPeer.p2pPort)
 
-  def joinRequest(joinRequest: JoinRequest, remoteAddress: Host): F[Unit] =
+  def joinRequest(hasCollateral: PeerId => F[Boolean])(joinRequest: JoinRequest, remoteAddress: Host): F[Unit] =
```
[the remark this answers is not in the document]
kriscieslak: As we discussed, because `Joining` is initialized in `sdk` and  snapshot storages are initialized on `core` or `dag-l1` level this is the way to pass verification function to `Joining`.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/collateral/LatestBalances.scala
kriscieslak: As we discussed `LatestBalances` interface is left to be implemented in `core` and `dag-l1`.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
kriscieslak: Done.

modules/sdk/src/main/scala/org/tessellation/sdk/cli/CliMethod.scala
kriscieslak: Changed.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/collateral/Collateral.scala
kriscieslak: Changed to first proposal.

modules/sdk/src/main/scala/org/tessellation/sdk/cli/CliMethod.scala
```diff
@@ -52,6 +54,9 @@ trait CliMethod {
     )
   )
 
+  private val normalizationFactor = 100_000_000L
```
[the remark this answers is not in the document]
kriscieslak: `normalizationFactor` moved to `balance` object in `sdk`.

modules/sdk/src/main/scala/org/tessellation/sdk/config/types.scala
kriscieslak: I ahve chabged it to `minStakingAmount`. @TheMMaciek ,if you are really unhappy with this I can change as you suggested.

tbekas (dismissed):
LGTM

modules/sdk/src/main/scala/org/tessellation/sdk/config/types.scala
kriscieslak: It is amount now.

modules/sdk/src/main/scala/org/tessellation/sdk/cli/CollateralConfigOpts.scala
```diff
@@ -0,0 +1,21 @@
...
+import org.tessellation.sdk.config.types.CollateralConfig
+
+import com.monovore.decline.Opts
+import com.monovore.decline.refined.refTypeArgument
+import eu.timepit.refined.types.numeric.NonNegLong
+
+object CollateralConfigOpts {
+
+  val opts: Opts[CollateralConfig] = Opts
+    .option[NonNegLong]("collateral", help = "User provided collateral")
```
marcinwadon, line 15: Could you change the help to be more descriptive what the collateral is? for example:
```
help = "Minimum staking amount to run a node"
```

marcinwadon (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/config/types.scala
```diff
@@ -69,4 +70,9 @@ object types {
     defaultCheckAttempts: PosInt,
     ensureCheckInterval: FiniteDuration
   )
+
+  case class CollateralConfig(
+    userAmount: Option[Amount],
```
kpudlik, line 75: Who is `user`? I mean we do not treat nodes as users so we should unify the glossary.
Node operators are not using the node but are hosting/operating the node.

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/collateral/CollateralSuite.scala
```diff
@@ -0,0 +1,80 @@
...
+      .map(expect.same(true, _))
+  }
+
+  test("improper collateral") { implicit sc =>
+    mkCollateral(Some(Map((address1, Balance(NonNegLong(249999))))))
+      .hasCollateral(peer1)
+      .map(expect.same(false, _))
+  }
+
+  test("pass for user provided collateral when env is 'Dev'") { implicit sc =>
```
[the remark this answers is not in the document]
marcinwadon: so I'd expect a different naming of those tests, something like:

```
should return false when balances are not initialized
should return false when the balance for a given address is empty or not found
should return true when the balance for a given address is equal to the required collateral
should return true when the balance for a given address is greater than the required collateral
should allow to set an arbitrary collateral for environment different than Mainnet
 -> or: should enforce hardcoded collateral for Mainnet environment
```

modules/sdk/src/main/scala/org/tessellation/sdk/config/types.scala
kriscieslak: Changed to `mainnetAmount`.

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/collateral/CollateralSuite.scala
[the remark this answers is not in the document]
kriscieslak: Changed names of tests.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/collateral/Collateral.scala
```diff
@@ -0,0 +1,42 @@
...
+          balances <- latestBalances.getLatestBalances
+        } yield
+          balances
+            .flatMap(_.get(address))
+            .map(_.value >= collateralAmount.value)
+            .getOrElse(false)
+
+      private def getCollateralAmount(env: AppEnvironment, config: CollateralConfig) =
+        if (env == Mainnet) Applicative[F].pure(config.amount)
+        else Applicative[F].pure(config.userAmount.getOrElse(config.amount))
```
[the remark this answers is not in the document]
kriscieslak: Changed.

kpudlik (approved):

PR #296, opened 2022-06-20
Grafana and Prometheus

tbekas:


kubernetes/prometheus/prometheus.yaml
```diff
@@ -0,0 +1,23 @@
...
+    scrape_timeout: 10s
+    metrics_path: /metrics
+    scheme: http
+    static_configs:
+      - targets:
+          - localhost:9090
+  - job_name: dynamic-targets
+    http_sd_configs:
+      - url: http://l0-initial-validator:9000/targets
+      - url: http://l1-initial-validator:9000/target
```
marcinwadon, line 23: what's the difference between `targets` and `target`?

modules/core/src/main/scala/org/tessellation/modules/HttpApi.scala
```diff
@@ -70,14 +72,16 @@ sealed abstract class HttpApi[F[_]: Async: SecurityProvider: KryoSerializer] pri
 
   private val debugRoutes = DebugRoutes[F](storages, services).routes
 
-  private val metricRoutes = MetricRoutes[F](services).routes
+  private val metricRoutes = routes.MetricRoutes[F]().routes
```
marcinwadon, line 75: why `routes.` prefix is needed?

marcinwadon (dismissed):
Looks fine, however `target` endpoint looks like typo

kubernetes/prometheus/prometheus.yaml
tbekas: It should be `targets` on both, thanks for catching it, I changed it without checking if it's working.

kpudlik (dismissed):
I like the fact that we got rid of ugly CRON ;d

kubernetes/prometheus/prometheus.yaml
```diff
@@ -0,0 +1,23 @@
...
+scrape_configs:
+  - job_name: prometheus
+    scrape_interval: 15s
+    scrape_timeout: 10s
+    metrics_path: /metrics
+    scheme: http
+    static_configs:
+      - targets:
+          - localhost:9090
+  - job_name: dynamic-targets
```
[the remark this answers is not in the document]
tbekas: Thanks :)

marcinwadon (approved):

PR #313, opened 2022-06-23
Add collateral verification for blocks

kriscieslak:


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/Validator.scala
```diff
@@ -26,12 +29,18 @@ object Validator {
 
   def isReadyForBlockConsensus(state: NodeState): Boolean = state == Ready
 
+  def withCollateral[F[_]: Applicative, A](collateral: Collateral[F])(peersIds: List[A])(id: A => PeerId) =
+    peersIds.toList.filterA(p => collateral.hasCollateral(id(p)))
```
marcinwadon, line 33: I think you may achieve `id: A => PeerId` using `Contramap` 🤔

marcinwadon (approved):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/Validator.scala
```diff
@@ -26,12 +29,18 @@ object Validator {
 
   def isReadyForBlockConsensus(state: NodeState): Boolean = state == Ready
 
+  def withCollateral[F[_]: Applicative, A: WithPeerId](collateral: Collateral[F])(peersIds: List[A]): F[List[A]] =
+    peersIds.toList.filterA(p => collateral.hasCollateral(WithPeerId[A].apply(p)))
```
tbekas, line 33: `toList` is not necessary as `peersIds` is already a list.

modules/shared/src/main/scala/org/tessellation/schema/peer.scala
```diff
@@ -44,6 +44,12 @@ object peer {
       fromId(publicKey.toId)
   }
 
+  type WithPeerId[A] = Function1[A, PeerId]
```
tbekas, line 47: It doens't bring much of functionality imo. I would go with simple `_.map(_.id)` function each time when set of peers need to be verified. If you want to be fancy maybe a syntax would be enough, like:
```scala
val peers: List[Peers] = ...
peers.map(_.id).filterCollateralized(collateral)
```

tbekas (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -163,7 +166,9 @@ class StateChannel[F[_]: Async: KryoSerializer: SecurityProvider: Random](
           .getTips(appConfig.consensus.tipsCount)
 
         l0Peer <- storages.l0Cluster.getPeers
-          .flatMap(peers => Random[F].shuffleList(peers.toNonEmptyList.toList))
+          .map(_.toNonEmptyList.toList)
+          .flatMap(_.filterA(p => services.collateral.hasCollateral(p.id)))
+          .flatMap(peers => Random[F].shuffleList(peers))
           .map(peers => peers.head)
```
TheMMaciek, line 172: Isn't it that right now it may throw and we should handle that?

tbekas (approved):
LGTM

PR #323, opened 2022-07-01
Implement rate limiting

TheMMaciek:


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionStorage.scala
```diff
@@ -96,51 +114,100 @@ class TransactionStorage[F[_]: Async: KryoSerializer](
...
 
         pulledM.handleErrorWith {
           logger.warn(_)(s"Error while pulling transactions for address=${address.show} from waiting pool.") >>
-            List.empty.pure[F]
+            none[NonEmptyList[Hashed[Transaction]]].pure[F]
         }
       }.map(_.flatten)
-    } yield NonEmptyList.fromList(allPulled)
+
+      selected = takeFirstNHighestFeeTxs(allPulled, count)
```
marcinwadon, line 162: should we sort by fee? that way there is an option that some transactions will be never pulled 🤔

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionStorage.scala
```diff
@@ -96,51 +114,100 @@ class TransactionStorage[F[_]: Async: KryoSerializer](
...
+      toReturn = allPulled.flatMap(_.toList).toSet.diff(selected.toSet)
```
marcinwadon, line 163: instead of pulling, checking and returning, we can just implement the limit in consecutive and do not pull more than expected

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/TransactionGenerator.scala
```diff
@@ -0,0 +1,43 @@
+package org.tessellation.dag.l1
```
marcinwadon, line 1: can it be in test-shared?

marcinwadon (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionStorage.scala
TheMMaciek: that's the whole point I guess of having transactions with fee, that they have the precedence

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionStorage.scala
TheMMaciek: We don't how many of these consecutive transactions for a single address will be pulled, because we first form the chains and only then take transactions by highest fee from all chains for all addresses. We don't know if another addresses will or won't have more transactions with higher fee and hence our transactions will have to wait. I don't know how many should I pull at the moment I'm calling consecutive.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -133,7 +133,9 @@ class StateChannel[F[_]: Async: KryoSerializer: SecurityProvider: Random](
 
   private val gossipBlock: Pipe[F, FinalBlock, FinalBlock] =
     _.evalTap { fb =>
-      services.gossip.spread(fb.hashedBlock.signed).handleErrorWith(e => logger.warn(e)("Block gossip spread failed!"))
+      services.gossip
+        .spreadCommon(fb.hashedBlock.signed)
+        .handleErrorWith(e => logger.warn(e)("Block gossip spread failed!"))
```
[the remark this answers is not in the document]
TheMMaciek: I didn't touch the logging part in this PR.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionStorage.scala
marcinwadon: that's true but imagine having a lot of transactions in a queue (100k, 500k). you'd shuffle them all back and forth. maybe we should keep them in consecutive order so we could see how many txs there are without moving them?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionStorage.scala
TheMMaciek: We could of course add a functionality to always take pulled feeless transactions but a consequence will be either a possibility to block fee transactions (not making the rate limit per block) or we will loose the control over limit of transactions per block (as we will need to pull over the limit). This in general touches upon very high network traffic and I'm not sure that there is a solution that would solve all the problems.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionStorage.scala
TheMMaciek: The problem is the fees and the not amount of the transactions. We need to give the precedence to the transactions with highest fee.

TheMMaciek: I could also change the `MapRef[Address, List[Transaction]]` to `Ref[Map[Address, List[Transaction]]]` and do all these operations in a `.modify` but then uploading the transactions may be less performant as I will block the whole ref instead of just one address in a mapref.

marcinwadon (approved):

PR #324, opened 2022-07-01
Endpoint returning wallet-count

marcinwadon:


modules/core/src/main/scala/org/tessellation/domain/dag/DagService.scala
```diff
@@ -10,4 +10,7 @@ trait DAGService[F[_]] {
 
   def getTotalSupply: F[Option[(BigInt, SnapshotOrdinal)]]
   def getTotalSupply(ordinal: SnapshotOrdinal): F[Option[(BigInt, SnapshotOrdinal)]]
+
+  def getTotalWallets: F[Option[(Int, SnapshotOrdinal)]]
```
TheMMaciek, line 14: ...`Count`. I would call it `getTotalWalletsCount`.

modules/core/src/main/scala/org/tessellation/http/routes/DagRoutes.scala
```diff
@@ -38,6 +38,13 @@ final case class DagRoutes[F[_]: Async](dagService: DAGService[F], mkDagCell: L0
         case _ => NotFound()
       }
 
+    case GET -> Root / "total-wallets" =>
```
TheMMaciek, line 41: Maybe "total-wallets-count" or just "wallets-count" for me "total-wallets" sounds confusing.

modules/core/src/main/scala/org/tessellation/infrastructure/dag/DAGService.scala
```diff
@@ -55,5 +55,20 @@ object DAGService {
 
       }
 
+      def getTotalWallets: F[Option[(Int, SnapshotOrdinal)]] =
+        globalSnapshotStorage.head.flatMap {
+          _.flatTraverse(snapshot => getTotalWallets(snapshot.value.ordinal))
```
TheMMaciek, line 60: You are fetching the same snapshot twice here. Maybe you should just extract the part retrieving needed fields into a separate function and use it in both `getTotalWallets` methods.

modules/core/src/main/scala/org/tessellation/http/routes/DagRoutes.scala
tbekas: I think the more english version is
`user-count`
(singular noun)

modules/core/src/main/scala/org/tessellation/infrastructure/dag/DAGService.scala
```diff
@@ -55,5 +55,20 @@ object DAGService {
...
+        }
+
+      def getTotalWallets(ordinal: SnapshotOrdinal): F[Option[(Int, SnapshotOrdinal)]] =
```
tbekas, line 63: This function receives ordinal and then returns it, which is odd. I think you don't need it at all. Once you have `head` you can do just `head.info.balances.size`.

tbekas (changes requested):

modules/core/src/main/scala/org/tessellation/infrastructure/dag/DAGService.scala
tbekas: Ok, I get it. Let's keep it then.

tbekas (approved):
LGTM

PR #325, opened 2022-07-01
Added test for writing trust values to the database

cngo-github:
Added test for writing trust values to the database and
the supporting structures.

Commits
=======
- Added SQL schema for TrustInfo.
- Updated the schema based on feedback from Ryle.
- Cleaned code.
- Changed peerId to varbinary(512).
- Updated schema for the TrustInfo table.
- Started the Score object.
- Debugging.
- Added TrustRating and supporting objects/method.
- Added objects for trust scores, optional values, and observation
amount, cleaned code.
- Debugging.
- Debugging.
- Added missing file.
- Debugging.
- Debugging.
- Debugging.
- Added working test for adding trust values to the database.
- Changed the database to in-memory.
- Cleaned code.
- Ran `sbt runLinter`.
- Ran `sbt scalafmt` and `sbt scalafix`.


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/trust/storage/TrustDbStorage.scala
```diff
@@ -0,0 +1,12 @@
+package org.tessellation.dag.l1.domain.trust.storage
+
+import org.tessellation.schema.peer.PeerId
+import org.tessellation.schema.trust.TrustDbValues
+
+trait TrustDbStorage[F[_]] {
```
marcinwadon, line 6: trait is an interface and it's part of a domain. domain by definition should be "framework-agnostic" as much as possible. that's the idea of abstraction and using interfaces, to make it exchangeable between different implementations. like an algebra and multiple interpreters. that being said, domain should be unaware of "database"

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/trust/storage/TrustDbStorage.scala
```diff
@@ -0,0 +1,12 @@
...
+  def getTrustScore(peerId: PeerId): F[Option[Double]]
```
marcinwadon, line 7: we prefer using nominal types https://github.com/estatico/scala-newtype for such values, to make it less error-prone (see transaction amount and fee)

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustDbStorageSuite.scala
```diff
@@ -0,0 +1,37 @@
...
+
+object TrustDbStorageSuite extends SimpleIOSuite {
+  test("Test writing trust values to the database.") {
+    val dbConfig = new DBConfig("org.sqlite.JDBC", "jdbc:sqlite::memory:", "sa", Secret(""))
+    Database.forAsync[IO](dbConfig).use { implicit db =>
+      val trustStorage = TrustDbStorage.make[IO]
+      val peerId = PeerId(Hex("123456789ABCDEF"))
+      val testScore = Option[Double](0.5)
+      val testRating = Option[Double](0.5)
+      val testObservationAdjustment = Option[Double](0.5)
```
marcinwadon, line 24: you can use a generator (arbitrary) for that purpose

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustDbStorageSuite.scala
```diff
@@ -0,0 +1,37 @@
...
+      val testTrustDbValues = TrustDbValues(peerId, testScore, testRating, testObservationAdjustment)
+
+      for {
+        _ <- trustStorage.updateTrustValues(Map(peerId -> testTrustDbValues)) // Map(A -> B) creates a tuple.
```
marcinwadon, line 28: such comment is not needed here and also `Map(A -> B)` does not create a tuple, it creates a single-element Map which is isomorphic to tuple

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustDbStorageSuite.scala
```diff
@@ -0,0 +1,37 @@
...
+        score <- trustStorage.getTrustScore(peerId)
+        rating <- trustStorage.getTrustRating(peerId)
+        observationAdjustment <- trustStorage.getTrustObservationAdjustment(peerId)
+      } yield
+        (expect.same(score, testScore) && expect.same(rating, testRating) && expect
```
marcinwadon, line 33: if possible, there should be one assertion per test (sometimes it's not possible)
here I think you can split it and describe the test more verbose

modules/shared/src/main/scala/org/tessellation/schema/trust.scala
```diff
@@ -1,13 +1,61 @@
...
+
+    def addToObservationAdjustment(that: Option[Double]): Either[ValueArithmeticError, TrustDbValues] = {
+      val newScore = add(observationAdjustment, that)
+
+      newScore match {
+        case Left(_)      => ValueOverflow.asLeft[TrustDbValues]
+        case Right(value) => TrustDbValues(peerId, score, rating, value).asRight[ValueArithmeticError]
+      }
+    }
+  }
```
marcinwadon, line 57: please take a look at monocle, it can simplify that code

marcinwadon (changes requested):
> PR Category: Test
Added test for writing trust values to the database and
the supporting structures.

I think it's wrong description, because you added the TrustDbStorage itself, not just test.

and can you remove that list, please? or just clean it up into list of changes that will be added in that pull request. imo it adds nothing to the pull request, especially: "-debugging" "-debugging" "-debugging", etc
image

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustDbStorageSuite.scala
cngo-github: I put that in there as I was learning. It will be removed.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/trust/storage/TrustStore.scala
```diff
@@ -0,0 +1,12 @@
+package org.tessellation.dag.l1.domain.trust.storage
+
+import org.tessellation.schema.peer.PeerId
+import org.tessellation.schema.trust.TrustDbValues
+
+trait TrustStore[F[_]] {
```
marcinwadon, line 6: `Store` -> `Storage` to keep the same naming within the codebase

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustDbStorage.scala
```diff
@@ -0,0 +1,114 @@
...
+import org.tessellation.schema.trust.TrustDbValues
+
+import doobie.implicits._
+import doobie.quill.DoobieContext
+import io.getquill.context.Context
+import io.getquill.{Literal, SqliteDialect}
+
+object TrustDbStorage {
+  case class StoredAddress(
+    peerId: String,
```
kpudlik, line 19: It should be `PeerId`. See comments about nominal types. We have such type already in schema.

modules/shared/src/main/scala/org/tessellation/schema/trust.scala
```diff
@@ -1,13 +1,113 @@
...
+
+    protected[trust] def addDoubles(
+      operand1: Option[Double],
+      operand2: Option[Double]
+    ): Either[ValueArithmeticError, Option[Double]] = {
+      val sum = operand1.flatMap(x => operand2.map(x + _)).get
+
+      if (sum >= -1.0 && sum <= 1.0) {
+        Option[Double](sum).asRight[ValueArithmeticError]
+      } else ValueOverflow.asLeft[Option[Double]]
```
kpudlik, line 26: I suggest you to use refined types for that:

https://github.com/fthomas/refined

You can then create a type like:

```scala
type TrustValue = Double Refined Interval.Closed[-1.0, 1.0]
```

and then use built-in `refineV` function which returns you `Either[String, TrustValue]`.

The nice thing if refined types is that it works at compile-time for refined types. For example such thing won't compile:


```scala
val x: TrustValue = 1.1 // Compilation error because it should be <= 1.0
```


Actually it applies to all the computations below :)

kpudlik (changes requested):

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustDbStorageSuite.scala
cngo-github: Removed.

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustDbStorageSuite.scala
cngo-github: The tests have been broken up.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/trust/storage/TrustDbStorage.scala
cngo-github: Converted `rating`, `score`, and `observationAdjustment` to `@newtype` value classes.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/trust/storage/TrustStore.scala
marcinwadon: or `TrustDbStorage` as indeed it's a database storage, not SQL storage :D

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustDbStorageSuite.scala
cngo-github: Added generators for the test.

modules/shared/src/main/scala/org/tessellation/implicits/trust.scala
```diff
@@ -0,0 +1,17 @@
+package org.tessellation.implicits
```
marcinwadon, line 1: you should not mix test with other code. test-specific code should be in test-related modules (see where other generators are)

marcinwadon:
Pull request label is missing

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/trust/storage/TrustStore.scala
cngo-github: Re-named `TrustSqlStorage` back to `TrustDbStorage`.

modules/shared/src/main/scala/org/tessellation/schema/trust.scala
cngo-github: Added refined types.

modules/shared/src/main/scala/org/tessellation/schema/trust.scala
cngo-github: Incorporated monocle.

cngo-github:
> Pull request label is missing

Added the labels `test` and `feature`

modules/shared/src/main/scala/org/tessellation/scalacheck/generators/trust.scala
```diff
@@ -0,0 +1,56 @@
...
+import org.scalacheck.Arbitrary.arbitrary
+import eu.timepit.refined.refineV
+import eu.timepit.refined.numeric.Interval
+import eu.timepit.refined.auto._
+
+object trust {
+
+  val genHex = for {
+    hexId <- arbitrary[String]
+  } yield Hex(hexId)
```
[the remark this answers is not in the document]
cngo-github: Simplified the generators.

modules/shared/src/main/scala/org/tessellation/implicits/trust.scala
cngo-github: Moved the implicits into the same object as the other generators.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustDbStorage.scala
cngo-github: Replaced the generic typed parameters with the values classes and added the appropriate decoders/encoders.

modules/shared/src/main/scala/org/tessellation/schema/trust.scala
cngo-github: `monocle` is no longer needed.

cngo-github:
Applied a workaround for the issue of `scalafmt` and the `-` in `-1.0`. Left a comment to detail why.

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustStorageSuite.scala
```diff
@@ -0,0 +1,52 @@
...
+import ciris.Secret
+import eu.timepit.refined.auto._
+import weaver._
+
+object TrustStorageSuite extends SimpleIOSuite {
+
+  val testTrustDbValues = genTrustDbValues.sample.get
+
+  test("Test writing trust values to the database - score.") {
+    val dbConfig = DBConfig("org.sqlite.JDBC", "jdbc:sqlite::memory:", "sa", Secret(""))
```
marcinwadon, line 18: `dbConfig` can be moved out of the test cases to keep it DRY ("don't repeat yourself")

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustStorageSuite.scala
```diff
@@ -0,0 +1,52 @@
...
+    Database.forAsync[IO](dbConfig).use { implicit db =>
```
marcinwadon, line 19: this creates a resource that can be provided at the suite level. please check the other tests with shared resources

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustStorageSuite.scala
```diff
@@ -0,0 +1,52 @@
...
+      val trustStorage = TrustDbStorage.make[IO]
+
+      for {
+        _ <- trustStorage.updateTrustValues(Map(testTrustDbValues.peerId -> testTrustDbValues))
+        score <- trustStorage.getTrustScore(testTrustDbValues.peerId)
+      } yield expect.same(score, testTrustDbValues.score)
```
marcinwadon, line 25:
```
val peer = testTrustDbValues.peerId
```
can be also moved above.
then, this can be simplified to something like that:
```scala
test("score should be written to the database") { implicit db =>
  trustStorage.updateTrustValues(Map(peer -> testTrustDbValues)) >>
    trustStorage.getTrustScore(peer).flatMap {
      expect.same(_, testTrustDbValues.score)
    }
}
```

modules/shared/src/main/scala/org/tessellation/scalacheck/generators/trust.scala
```diff
@@ -0,0 +1,64 @@
...
+    })
+
+  val genObservationAdjustment = arbitrary[Double]
+    .flatMap(refineV[TrustValueRefinement](_))
+    .flatMap(_ match {
+      case Left(_)      => Option[ObservationAdjustment](ObservationAdjustment(0.0))
+      case Right(value) => Option[ObservationAdjustment](ObservationAdjustment(value))
+    })
+
+  val genPeerId = genHex.flatMap(PeerId(_))
```
marcinwadon, line 50: the arbitrary for peer id already exists, please reuse it (abstract if needed)

modules/shared/src/main/scala/org/tessellation/scalacheck/generators/trust.scala
```diff
@@ -0,0 +1,64 @@
...
+    Gen.choose(0, 0xffffff).flatMap(_.toHexString)
+  }
+
+  implicit val arbitraryTrustDouble: Arbitrary[Double] = Arbitrary {
+    Gen.choose(-1.0, 1.0)
+  }
+
+  implicit val arbitraryOptionalTrustValue: Arbitrary[Option[Double]] = Arbitrary {
+    Gen.option(Gen.choose(-1.0, 1.0))
+  }
```
marcinwadon, line 25: you should reuse the existing arbitrary - `arbitraryTrustDouble` (simply map it or lift into option). please don't repeat the same logic

modules/shared/src/main/scala/org/tessellation/scalacheck/generators/trust.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  val genHex = arbitrary[String].flatMap(Hex(_))
+
+  val genScore = arbitrary[Double]
+    .flatMap(refineV[TrustValueRefinement](_))
+    .flatMap(_ match {
+      case Left(_)      => Option[Score](Score(0.0))
+      case Right(value) => Option[Score](Score(value))
+    })
```
marcinwadon, line 34: instead of pattern match, you can use
```scala
.flatMap(_.toOption.getOrElse(Score(0.0))
```
or even, add `empty` property to the Score (see: Hash or TransactionReference)
and then:
```scala
.flatMap(_.toOption.getOrElse(Score.empty))
```

modules/shared/src/main/scala/org/tessellation/scalacheck/generators/trust.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  val genRating = arbitrary[Double]
+    .flatMap(refineV[TrustValueRefinement](_))
+    .flatMap(_ match {
+      case Left(_)      => Option[Rating](Rating(0.0))
```
marcinwadon, line 39: as above

modules/shared/src/main/scala/org/tessellation/scalacheck/generators/trust.scala
```diff
@@ -0,0 +1,64 @@
...
+      case Right(value) => Option[Rating](Rating(value))
+    })
+
+  val genObservationAdjustment = arbitrary[Double]
+    .flatMap(refineV[TrustValueRefinement](_))
+    .flatMap(_ match {
+      case Left(_)      => Option[ObservationAdjustment](ObservationAdjustment(0.0))
```
marcinwadon, line 46: as above

modules/shared/src/main/scala/org/tessellation/scalacheck/generators/trust.scala
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  implicit val arbitraryOptionalTrustValue: Arbitrary[Option[Double]] = Arbitrary {
+    Gen.option(Gen.choose(-1.0, 1.0))
+  }
+
+  val genHex = arbitrary[String].flatMap(Hex(_))
+
+  val genScore = arbitrary[Double]
+    .flatMap(refineV[TrustValueRefinement](_))
```
marcinwadon, line 30: refined has scalacheck library that supports such generators. please check it out ("refined-scalacheck")

modules/shared/src/main/scala/org/tessellation/schema/TrustValueRefinement.scala
```diff
@@ -0,0 +1,11 @@
+package org.tessellation.schema
+
+import eu.timepit.refined.numeric.Interval
+
+object TrustValueRefinement {
+
+  // Moved this here so that I can exclude this file from scalafmt. This is necessary because of a bug in the current
+  // version of scalafmt that doesn't like the "-" in "-1.0", even though it is valid and the code compiles. scalafmt is
+  // configured to ignore this file.
```
marcinwadon, line 9: imagine you're a newcomer and you see: "moved this here". it doesn't make sense because it's out of context.
anyway, have you tried inline ignoring?

modules/shared/src/main/scala/org/tessellation/schema/trust.scala
```diff
@@ -1,13 +1,84 @@
...
+import io.estatico.newtype.macros.newtype
+import io.getquill.MappedEncoding
 
 object trust {
 
+  type TrustValue = Double Refined TrustValueRefinement
+
+  sealed trait ValueArithmeticError extends NoStackTrace
+  case object ValueOverflow extends ValueArithmeticError
+  case object ValueUnderflow extends ValueArithmeticError
```
marcinwadon, line 24: is it used?

marcinwadon (changes requested):

modules/shared/src/main/scala/org/tessellation/schema/TrustValueRefinement.scala
cngo-github: I have tried inline ignoring with `// scalafmt:ok` but it didn't work. Ignoring the file was the only thing that worked.

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustStorageSuite.scala
cngo-github: It is not. It will be removed.

modules/shared/src/main/scala/org/tessellation/schema/trust.scala
cngo-github: No. It will be removed.

modules/shared/src/main/scala/org/tessellation/scalacheck/generators/trust.scala
cngo-github: I intended to remove `arbitraryTrustDouble` but I forgot. It has been removed.

modules/shared/src/main/scala/org/tessellation/scalacheck/generators/trust.scala
cngo-github: The generators have been simplified.

modules/shared/src/main/scala/org/tessellation/scalacheck/generators/trust.scala
cngo-github: The generators have been simplified.

modules/shared/src/main/scala/org/tessellation/scalacheck/generators/trust.scala
cngo-github: The generators have been simplified.

modules/shared/src/main/scala/org/tessellation/scalacheck/generators/trust.scala
cngo-github: I tried `arbitrary[Double].flatmap(Score(_))` but that resulted in `compile-time refinement only works with literals`. The only way I know to work around that is `refineV`, but then the code is not meaningfully different.

modules/shared/src/main/scala/org/tessellation/scalacheck/generators/trust.scala
cngo-github: Changed.

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustStorageSuite.scala
cngo-github: Changed the test to use a shared `Resource`, but not `ResourceSuite`.

I was unable to figure out how to write a test based on:

https://github.com/Constellation-Labs/tessellation/blob/001ab0b555165272704e3ec6c9e1febce32e2eb8/modules/dag-shared/src/test/scala/org/tessellation/dag/transaction/ContextualTransactionValidatorSuite.scala

so, I based my test on:

https://github.com/cngo-github/tessellation/blob/20220623-ngo-dev-trust-test/modules/dag-l1/src/test/scala/org/tessellation/dag/l1/domain/transaction/TransactionStorageSuite.scala

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustStorageSuite.scala
cngo-github: The tests were simplified as part of the effort to use a shared `Resource`.

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustStorageSuite.scala
```diff
@@ -0,0 +1,61 @@
...
+
+object TrustStorageSuite extends SimpleIOSuite {
+
+  val dbConfig = DBConfig("org.sqlite.JDBC", "jdbc:sqlite::memory:", "sa", Secret(""))
+
+  type TestResources = (TrustStorage[IO], TrustDbValues)
+
+  def testResources: Resource[IO, TestResources] = Database.forAsync[IO](dbConfig).flatMap { implicit db =>
+    val testTrustDbValues = genTrustDbValues.sample.get
+    val trustStorage = TrustDbStorage.make[IO]
```
marcinwadon, line 24: `Database` is a resource only I think 🤔

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustStorageSuite.scala
cngo-github: I spent most of the day working and consulting with @ryle-goehausen about this. I’ve decided to keep using `SimpleIOSuite` because:

We cannot use a shared resource here because the database cannot be shared as the in-memory connection dropping leads to table dropping on all operations. Chaining any operations on top of the database resource leads to those operations not being re-run. There may be a way to get the chaining to work properly with writes, but all the things we tried didn’t work and it is not worth days of effort figuring it out.

The current implementation does reduce code repetition, and I think trying to jam the test into `ResourceSuite` will result in more complicated code.

marcinwadon: Got it. Let me try tomorrow

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/trust/storage/TrustStorage.scala
```diff
@@ -0,0 +1,12 @@
+package org.tessellation.dag.l1.domain.trust.storage
+
+import org.tessellation.schema.peer.PeerId
+import org.tessellation.schema.trust._
+
+trait TrustStorage[F[_]] {
+  def getTrustScore(peerId: PeerId): F[Option[Score]]
+  def getTrustRating(peerId: PeerId): F[Option[Rating]]
+  def getTrustObservationAdjustment(peerId: PeerId): F[Option[ObservationAdjustment]]
```
marcinwadon, line 9: when I think about this, I believe the return type can be with no Option type.
if I want to get, let's say` trustScore for peer A, it should fallback to the default value if nothing is stored. so the return type would be `F[Score]` anyway. what do you think @cngo-github?

marcinwadon (dismissed):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/trust/storage/TrustStorage.scala
cngo-github: `None != 0`. These `TrustValue`s are meant to be calculated locally, to compliment EigenTrust's global trust values. So, `None` indicates that there is no knowledge of an entity while `0` indicates there is some knowledge, but the value is `0` (i.e., no opinion). So, no knowledge != no opinion.

marcinwadon (dismissed):

kpudlik (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustDbStorage.scala
```diff
@@ -0,0 +1,124 @@
...
+  def make[F[_]: MonadCancelThrow: Database]: TrustStorage[F] =
+    make[F](new DoobieContext.SQLite[Literal](Literal))
+
+  def make[F[_]: MonadCancelThrow](ctx: Ctx)(implicit db: Database[F]): TrustStorage[F] =
+    new TrustStorage[F] {
+      val xa = db.xa
+
+      import ctx._
+
+      val getPeerIds = quote {
```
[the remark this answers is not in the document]
cngo-github: Renamed to `getStoredTrust` and updated code accordingly. Also renamed the other functions to replace `PeerId` with `StoredTrust`.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustDbStorage.scala
```diff
@@ -0,0 +1,124 @@
...
+          observationAmount: Option[ObservationAdjustment]
+        ) =>
+          getPeerIds
+            .filter(_.peerId == peerId)
+            .update(_.score -> score, _.rating -> rating, _.observationAdjustment -> observationAmount)
+      }
+
+      def getTrustScore(peerId: PeerId): F[Option[Score]] =
+        run(getPeerIdScore(lift(peerId)))
+          .map(_.headOption.flatMap(x => x))
```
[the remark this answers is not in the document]
cngo-github: Changed.

cngo-github:
`val genHex` no longer exists.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustDbStorage.scala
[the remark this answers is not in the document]
marcinwadon: @cngo-github I still see `.headOption.flatMap(x => x)` are you sure you pushed the code and/or updated the whole file?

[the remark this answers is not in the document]
cngo-github: I forgot to update the other occurrences. That has been rectified.

marcinwadon (approved):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustDbStorage.scala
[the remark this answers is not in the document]
kpudlik: There is no need to add `getStored` prefix to every single method because we already know that you get that from storage. The case class is named `TrustDbStorage`. That decision may lead to redundant naming like `storage.getStoredTrust` instead of just `storage.getTrust`. Also the prefix "Trust" is redundant (`TrustScore`, `TrustRating`, etc) because we already know that it is `TrustStorage`. It would be fine if case class was named `TrustScore` but you named it just `Score` so I would go for `getScore` which will be called like this: `trustStorage.getScore`

TL:DR:

Just use `getTrust`, `getScore`, `getRating` without `...StoredTrust` prefix

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/trust/storage/TrustStorage.scala
```diff
@@ -0,0 +1,12 @@
+package org.tessellation.dag.l1.domain.trust.storage
+
+import org.tessellation.schema.peer.PeerId
+import org.tessellation.schema.trust._
+
+trait TrustStorage[F[_]] {
+  def getTrustScore(peerId: PeerId): F[Option[Score]]
```
kpudlik, line 7: Just `getScore`.

For example in `AddressStorage` we have `getBalance` instead of `getAddressBalance` so you should name it `getScore`

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/trust/storage/TrustStorage.scala
kpudlik, line 9: just `getObservationAdjustment`

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustDbStorage.scala
```diff
@@ -0,0 +1,123 @@
...
+  def make[F[_]: MonadCancelThrow: Database]: TrustStorage[F] =
+    make[F](new DoobieContext.SQLite[Literal](Literal))
+
+  def make[F[_]: MonadCancelThrow](ctx: Ctx)(implicit db: Database[F]): TrustStorage[F] =
+    new TrustStorage[F] {
+      val xa = db.xa
+
+      import ctx._
+
+      val getStoredTrust = quote {
```
kpudlik, line 36: `getStoredTrusts` (should be plural because that query returns multiple elements)

kpudlik (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/trust/storage/TrustDbStorage.scala
cngo-github: Changed.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/trust/storage/TrustStorage.scala
```diff
@@ -0,0 +1,12 @@
...
+  def getTrustRating(peerId: PeerId): F[Option[Rating]]
```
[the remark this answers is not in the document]
cngo-github: Changed.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/trust/storage/TrustStorage.scala
cngo-github: Changed.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/trust/storage/TrustStorage.scala
cngo-github: Changed.

PR #326, opened 2022-07-04
Handling TimeTrigger & EventTrigger

tbekas:


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -1,73 +1,103 @@
...
-  def checkForTrigger(event: Event): F[Unit]
-  def checkForStateUpdate(key: Key)(resources: ConsensusResources[Artifact]): F[Unit]
-  def checkForStateUpdateSync(key: Key)(resources: ConsensusResources[Artifact]): F[Unit]
+  /** Schedules facilitating consensus with `TimeTrigger` after configured interval */
+  def scheduleTimeTrigger: F[Unit]
+
+  /** Tries to facilitate consensus immediately without any `TimeTrigger` or EventTrigger` */
+  def triggerWithNothing: F[Unit]
+
+  private[consensus] def triggerWithEvent: F[Unit]
```
kpudlik, line 30: Maybe we should use different names to avoid comments. Like:

`trigger`, `triggerWithEvent`, `triggerOnSchedule`

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStorage.scala
```diff
@@ -150,41 +171,63 @@ object ConsensusStorage {
...
+              if (updateTrigger) {
+                val maybeNewTrigger = events.map(_._1).maximumOption
+
+                maybeCurrentTrigger match {
+                  case Some(currentTrigger) =>
+                    maybeNewTrigger.fold(currentTrigger)(o => Order[Ordinal].max(currentTrigger, o)).some
+                  case None => maybeNewTrigger
+                }
+              } else
+                maybeCurrentTrigger
```
kpudlik, line 211: So you basically want to take `max` of 2 `Option[Ordinal]` or fallback to `maybeNewTrigger` or in worst case to `maybeCurrentTrigger`.

I guess then it can be simplified to:

```scala
(maybeCurrentTrigger, maybeNewTrigger)
   .mapN(Order[Ordinal].max(_, _))
   .orElse(maybeNewTrigger)
   .orElse(maybeCurrentTrigger)
```

eventually you can try with implicit conversion from Option to List and `maxOption` which uses `Ordering` implicitly

```scala
(maybeCurrentTrigger ++ maybeNewTrigger)
  .maxOption
  .orElse(maybeNewTrigger)
  .orElse(maybeCurrentTrigger)
```

kpudlik (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStorage.scala
tbekas: The first option you've suggested looks neat, I didn't think about that, fixing it.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
tbekas: Ok, I can remove the comments, but I would keep `triggerWithNothing` to be explicit about it.

tbekas: In the end, I replaced `triggerWithNothing` with `triggerOnStart` as it's used only initially.

kpudlik (approved):

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotEventsPublisherDaemon.scala
```diff
@@ -4,25 +4,29 @@ import cats.effect.Async
...
+    l1OutputQueue: Queue[F, Signed[L1Output]],
     gossip: Gossip[F]
   ): Daemon[F] = Daemon.spawn {
     Stream
       .fromQueueUnterminated(stateChannelOutputs)
       .map(_.asLeft[DAGEvent])
       .merge(
-        dagEvents
+        Stream
+          .fromQueueUnterminated(l1OutputQueue)
```
marcinwadon, line 28: is that queue used here, and only here? otherwise, events may be consumed by another subscriber

tbekas: It's used only here

PR #333, opened 2022-07-09
Add collateral verification during block acceptance

kriscieslak:


marcinwadon (changes requested):
IMO we shouldn't validate when pulling from the queue, but when pushing to the queue. Similarly how transactions work - first validation is before enqueue

marcinwadon:
Label is missing

kriscieslak:
IMO it is exactly how it works right now. Validation is added to method `processDAGL1` in `Coalgebra` which prepares data to be enqueued later in method `enqueueDAGL1Data` from `Algebra`. After the modification when validation fails `Algebra.enqueueDAGL1Data` will not be called, instead `CellError` will be propagated.

modules/core/src/main/scala/org/tessellation/http/routes/DagRoutes.scala
```diff
@@ -58,7 +59,10 @@ final case class DagRoutes[F[_]: Async](dagService: DAGService[F], mkDagCell: L0
         .map(L0CellInput.HandleDAGL1(_))
         .map(mkDagCell)
         .flatMap(_.run())
-        .flatMap(_ => Ok())
+        .flatMap {
+          case Right(_)             => Ok()
+          case Left(CellError(msg)) => BadRequest(msg)
```
marcinwadon, line 64: This should be somehow handled on caller-site

marcinwadon (dismissed):

modules/core/src/main/scala/org/tessellation/domain/cell/L0Cell.scala
```diff
@@ -85,10 +94,18 @@ object L0Cell {
...
   object Coalgebra {
 
-    def processDAGL1[F[_]: Async](data: Signed[L1Output]): CoalgebraR[F] = {
+    def processDAGL1[F[_]: Async: Logger](data: Signed[L1Output], collateral: Collateral[F]): CoalgebraR[F] = {
       def res: StackF[CoalgebraCommand] = Done(AlgebraCommand.EnqueueDAGL1Data(data).asRight[CellError])
-
-      res.pure[F]
+      def errorRes: F[StackF[CoalgebraCommand]] =
+        Logger[F]
+          .debug(s"Peers which signed block sent by ${data.proofs.head.id} have not sufficient collateral")
```
marcinwadon, line 101: each CellError is being logged (see: StateChannel.scala lines 105:130), so looks like this is a duplicated log

modules/core/src/test/scala/org/tessellation/domain/cell/L0CellSuite.scala
```diff
@@ -0,0 +1,93 @@
...
+      Transaction(
+        Address("DAG0y4eLqhhXUafeE3mgBstezPTnr8L3tZjAtMWB"),
+        Address("DAG07tqNLYW8jHU9emXcRTT3CfgCUoumwcLghopd"),
+        TransactionAmount(1L),
+        TransactionFee(1L),
+        TransactionReference(TransactionOrdinal(1L), Hash("hash")),
+        TransactionSalt(0L)
+      ),
+      NonEmptySet.one(SignatureProof(Id(Hex("")), Signature(Hex(""))))
+    )
```
marcinwadon, line 51: I think you can use a generator for those data

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -160,7 +160,7 @@ class StateChannel[F[_]: Async: KryoSerializer: SecurityProvider: Random](
     }
 
   private val sendBlockToL0: Pipe[F, FinalBlock, FinalBlock] =
-    _.evalTap { fb =>
+    _.evalFilter { fb =>
```
TheMMaciek, line 163: If sending to L0 failed, it's not a reason to discard such a block from further processing. Two other L1 facilitators may still send the block successfully to L0. Additionally we should rather optimize process of sending to L0 (retries?) then discard the block if sending failed.

modules/core/src/test/scala/org/tessellation/domain/cell/L0CellSuite.scala
TheMMaciek: Generating Addresses may be hard.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
kriscieslak: Code reverted.

modules/core/src/main/scala/org/tessellation/http/routes/DagRoutes.scala
kriscieslak: It is just logged on caller-site. There is used `successful` method from `http4s.Client` which should return false for `BadRequest` status and then warning message should be logged.

modules/core/src/main/scala/org/tessellation/domain/cell/L0Cell.scala
kriscieslak: `StateChannel` logs messages from `BlockConsensusCell`. I can't find similar code for `L0Cell`.

modules/dag-shared/src/test/scala/org/tessellation/dag/domain/block/generators.scala
```diff
@@ -0,0 +1,26 @@
...
+import org.tessellation.dag.domain.block.{BlockReference, DAGBlock}
+import org.tessellation.schema.generators._
+import org.tessellation.schema.height.Height
+import org.tessellation.security.hash.ProofsHash
+
+import org.scalacheck.{Arbitrary, Gen}
+
+object generators {
+
+  val blockReferenceGen: Gen[BlockReference] =
```
tbekas, line 14: Block Reference already has an `Arbitrary` instance (which can be mapped to `Gen` if necessary by calling `arbitrary` method).

modules/core/src/main/scala/org/tessellation/domain/cell/L0Cell.scala
```diff
@@ -85,10 +94,18 @@ object L0Cell {
...
+          .as(Done(CellError("Peers which signed block have not sufficient collateral").asLeft[Ω]))
```
tbekas, line 102: I'm not sure if rejecting blocks in Cell is necessary. Maybe it would be enough to reject them only during L0 consensus? 🤔

tbekas (changes requested):

modules/dag-shared/src/main/scala/org/tessellation/dag/block/processing/BlockAcceptanceLogic.scala
```diff
@@ -189,4 +194,30 @@ object BlockAcceptanceLogic {
...
+      signedBlock.proofs
+        .map(_.id.toPeerId)
+        .toList
+        .traverse(_.toAddress)
+        .flatMap(
+          _.filterA(
+            address =>
+              context
+                .getBalance(address)
+                .map(_.map(balance => !balance.satisfiesColalteral(context.getCollateral)).getOrElse(true))
```
tbekas, line 212: Please replace it with:
```
context.getBalance(address).map {
    _.getOrElse(Balance.empty).satisfiesCollateral(context.getCollateral))
}
```

modules/shared/src/main/scala/org/tessellation/schema/balance.scala
```diff
@@ -39,6 +39,8 @@ object balance {
       NonNegLong
         .from(value.value - that.value.value)
         .bimap(_ => BalanceUnderflow, Balance(_))
+
+    def satisfiesColalteral(collateral: Amount): Boolean = value >= collateral.value
```
tbekas, line 43: Typo: `colalteral` -> `collateral`

tbekas (changes requested):

PR #334, opened 2022-07-10
Tool prints progress every 1s instead of every 100 tx

tbekas:


modules/tools/src/main/scala/org/tessellation/tools/Main.scala
```diff
@@ -143,24 +143,33 @@ object Main
...
-        )
-        .zipWithIndex
-        .evalTap(printProgress[F](basicOpts.verbose, startTime))
-        .through(applyLimit(basicOpts.take))
-        .through(applyDelay(basicOpts.delay))
-        .handleErrorWith(e => Stream.eval(console.red(e.toString)))
-        .compile
-        .drain
+      Ref.of(0L).flatMap { counterR =>
+        val printProgressApplied = counterR.get.flatMap(printProgress(startTime, _))
```
marcinwadon, line 147: It's totally minor and irrelevant here, but just want to mention. It's good to write such a thing like "def" instead of "val", to emphasize the use of functions. We write most of the code using functional programming, and in FP there are only functions (value is a no-arg fn), so it's convenient to write everything using "def" until explicitly emphasizing something is indeed a value (like caching something for optimization).

marcinwadon (approved):

modules/tools/src/main/scala/org/tessellation/tools/Main.scala
tbekas: >and in FP there are only functions (value is a no-arg fn)

Not sure if I agree with that. In lambda calculus functions have exactly one argument (not zero) and there are values (arguments).

marcinwadon: Right, unit is an argument, that's what I meant

marcinwadon: Some time ago this influenced my mindset

image

tbekas: >Right, unit is an argument, that's what I meant

A unary function is not equivalent to a value, even if such function has a unit argument. If they were the same you could replace the occurrence of one with another, and that's not the case (not in lambda calculus and not in Scala). 

Lambda calculus defines [free variables](https://en.wikipedia.org/wiki/Lambda_calculus#Free_variables), and in my understanding `val printProgressApplied` is such a free variable in lambdas using it.

marcinwadon: Nevermind, do what you want, just sharing my standpoint.

tbekas: I appreciate that, thanks 👍

PR #335, opened 2022-07-12
Rumor validator

tbekas:


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/gossip/RumorValidator.scala
```diff
@@ -0,0 +1,93 @@
...
+        Validated.condNec(calculatedHash === hash, (hash, signedRumor), InvalidHash(calculatedHash, hash))
+      }
+
+    def validateOrigin(signedRumor: Signed[RumorBinary]): RumorValidationErrorOr[Signed[RumorBinary]] =
+      signedRumor.value match {
+        case _: CommonRumorBinary => signedRumor.validNec[RumorValidationError]
+        case rumor: PeerRumorBinary =>
+          val signers = signedRumor.proofs.map(_.id)
+          Validated.condNec(
+            signers.contains(rumor.origin.toId),
```
TheMMaciek, line 63: Contains or equals exactly to that given id? 🤔

tbekas: I don't think it matters much. Even if there are more signatures in a rumor, we're going to care only about the `origin` signature.

TheMMaciek: For sure it doesn't break anything but I wonder what should we allow.

TheMMaciek (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/gossip/RumorValidator.scala
marcinwadon: There is a method on Signed "isSignedBy" or "isSignedExclusivelyBy". Maybe we can reuse that 🤔

marcinwadon: @TheMMaciek I could intercept rumors and add my signature. It shouldn't invalidate the original rumor 🤔

PR #337, opened 2022-07-12
Send blocks to L0 without tips and remove L1Output

TheMMaciek:


modules/core/src/main/scala/org/tessellation/http/routes/DagRoutes.scala
```diff
@@ -54,7 +54,7 @@ final case class DagRoutes[F[_]: Async](dagService: DAGService[F], mkDagCell: L0
 
     case req @ POST -> Root / "l1-output" =>
```
TheMMaciek, line 55: Do we want to change the endpoint as well? All in all signed block is still a l1-output.

PR #341, opened 2022-07-14
Fix iterative acceptance of blocks

kriscieslak:


modules/dag-shared/src/main/scala/org/tessellation/dag/block/processing/BlockAcceptanceManager.scala
```diff
@@ -58,7 +58,7 @@ object BlockAcceptanceManager {
           toProcess: List[(Signed[DAGBlock], TxChains)]
         ): F[BlockAcceptanceState] =
           for {
-            currState <- toProcess.foldLeftM(initState) { (acc, blockAndTxChains) =>
+            currState <- toProcess.foldLeftM(initState.copy(awaiting = List.empty)) { (acc, blockAndTxChains) =>
```
tbekas, line 61: Great that you've found it. It was like that initially, but I mistakenly changed it. Thanks for fixing it.

tbekas (changes requested):
Great change, but please use valid hash values.

tbekas (changes requested):

tbekas (approved):

PR #343, opened 2022-07-14
Don't start genesis node without collateral

kriscieslak:


modules/core/src/main/scala/org/tessellation/Main.scala
```diff
@@ -106,6 +108,9 @@ object Main
 
               Signed.forAsyncKryo[IO, GlobalSnapshot](genesis, keyPair).flatMap { signedGenesis =>
                 storages.globalSnapshot.prepend(signedGenesis) >>
+                  services.collateral
+                    .hasCollateral(sdk.nodeId)
+                    .flatMap(OwnCollateralNotSatisfied.raiseError[IO, Unit].unlessA) >>
```
marcinwadon, line 113: does it stop the process and node stops immediately?

kriscieslak: Yes, I have checked it locally.

marcinwadon (approved):

PR #344, opened 2022-07-14
Don't check collateral for own rumors

kriscieslak:


marcinwadon:
Why collateral shouldn't be checked for own rumors? You should have the collateral in order to participate in the cluster. If you don't have one, you should be ignored anyway

tbekas:
I agree with @marcinwadon . Maybe the way to approach the problem of discarded rumors at the beginning is to **not reject** rumors, when there's no information about balances?

kriscieslak:
Ok, I have changed approach a bit, let me know what do you think.

PR #345, opened 2022-07-19
Improve error response for post transaction route

TheMMaciek:


marcinwadon (approved):
I'd love to have an "ApplicationError" returned by all the services and then conversion from ApplicationError to HttpError or ErrorResponse, so then it can be reused across the whole codebase. Let's talk about that later :)

PR #346, opened 2022-07-21
Rollback mechanism

TheMMaciek:


modules/core/src/main/scala/org/tessellation/Main.scala
```diff
@@ -51,7 +52,8 @@ object Main
       _ <- IO.unit.asResource
       p2pClient = P2PClient.make[IO](sdkP2PClient, sdkResources.client, sdkServices.session)
       queues <- Queues.make[IO](sdkQueues).asResource
-      storages <- Storages.make[IO](sdkStorages, cfg.snapshot).asResource
+      maybeRollbackHash = Option(method).collect { case rr: RunRollback => rr.rollbackHash }
+      storages <- Storages.make[IO](sdkStorages, cfg.snapshot, maybeRollbackHash).asResource
```
marcinwadon, line 56: How does it differ from the following code that is already used for running genesis?
```
storages.globalSnapshot.prepend(...)
```

modules/core/src/main/scala/org/tessellation/cli/method.scala
```diff
@@ -89,6 +90,41 @@ object method {
...
+    whitelistingPath: Option[Path],
+    collateralAmount: Option[Amount],
+    rollbackHash: Hash
+  ) extends Run
+
+  object RunRollback extends WithOpts[RunRollback] {
+
+    val whitelistingPathOpts: Opts[Option[Path]] = Opts.option[Path]("whitelisting", "").orNone
+
+    val rollbackHashOpts: Opts[Hash] = Opts.argument[Hash]("rollbackHash")
```
marcinwadon, line 110: I'd also allow for rollback from ordinal or the latest ordinal if nothing is specified. What do you think?

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorage.scala
```diff
@@ -36,28 +36,78 @@ import org.typelevel.log4cats.slf4j.Slf4jLogger
...
+      .read(rollbackHash)
+      .flatMap(ApplicativeError.liftFromOption(_, new Throwable("Rollback global snapshot not found!")))
+      .flatMap { rollbackSnapshot =>
+        makeResources().flatMap {
+          case (headRef, ordinalCache, hashCache, notPersistedCache, offloadQueue, logger) =>
+            implicit val l = logger
+
+            headRef.set(rollbackSnapshot.some) >>
+              ordinalCache(rollbackSnapshot.ordinal).set(rollbackHash.some) >>
+              hashCache(rollbackHash).set(rollbackSnapshot.some) >>
```
marcinwadon, line 82: Instead of that, you can just `prepend` the snapshot in the same way we prepend the genesis 🤔

marcinwadon (changes requested):

modules/core/src/main/scala/org/tessellation/Main.scala
TheMMaciek: Prepend will try to save it to disk again hence an error will be thrown constantly because of retry mechanism (such file already exists). By setting caches manually we don't have that error.

modules/core/src/main/scala/org/tessellation/cli/method.scala
TheMMaciek: Yeah we could add it as an additional feature, but it requires a bit more work and possibly some protection, so I didn't try to include it here. I can't create a ticket for it though.

marcinwadon (approved):

PR #347, opened 2022-07-22
Print app version on startup

tbekas:


marcinwadon (approved):

marcinwadon:
@tbekas next to the app version, it would be great to show also the build number (see BuildInfo object)

tbekas:
@marcinwadon This is the info that's available in `BuildInfo`. I don't see any build number.

```
case object BuildInfo {
  /** The value is "tessellation-shared". */
  val name: String = "tessellation-shared"
  /** The value is "0.11.0-SNAPSHOT". */
  val version: String = "0.11.0-SNAPSHOT"
  /** The value is "2.13.8". */
  val scalaVersion: String = "2.13.8"
  /** The value is "1.5.5". */
  val sbtVersion: String = "1.5.5"
  override val toString: String = {
    "name: %s, version: %s, scalaVersion: %s, sbtVersion: %s".format(
      name, version, scalaVersion, sbtVersion
    )
  }
}
```

marcinwadon:
@tbekas There is a simple option to return additional things like a timestamp and version + timestamp can most likely unique identify the build: https://github.com/sbt/sbt-buildinfo

With this two options:
```
buildInfoOptions ++= Seq(BuildInfoOption.BuildTime, BuildInfoOption.ToMap),
``` 
this is what we have in mainnet 1.0:
```
      BuildInfo.name,
      BuildInfo.version,
      BuildInfo.scalaVersion,
      BuildInfo.sbtVersion,
      BuildInfo.gitBranch,
      BuildInfo.gitCommit,
      BuildInfo.builtAtString,
      BuildInfo.builtAtMillis
```

tbekas:
How about `BuildInfo.gitCommit` instead of the timestamp? It might be useful since we have `-SNAPSHOT` versions deployed into some envs sometimes.

marcinwadon:
@tbekas sure :+1:
What I said is that just a `version` may be not distinguishable between runs so it would be helpful to have something more precise

PR #354, opened 2022-07-28
Periodic peer discovery on L1

marcinwadon:


modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/L0PeerDiscovery.scala
```diff
@@ -28,17 +31,44 @@ object L0PeerDiscovery {
...
+          .fromSet(SortedSet.from(s))
+          .fold {
+            (new Throwable("Unexpected state - no peers found but at least one should be available")).raiseError[F, NonEmptySet[L0Peer]]
+          }(Applicative[F].pure)
+      }
+
+  def discoverFrom(peer: P2PContext): F[Unit] =
+    getPeersFrom(peer)
+      .map(_.toSortedSet)
+      .map(_.toSet)
```
TheMMaciek, line 71: `addPeers` will not accept the sortedSet?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -90,6 +90,16 @@ class StateChannel[F[_]: Async: KryoSerializer: SecurityProvider: Random](
     .filter(identity)
     .as(OwnRoundTrigger)
 
+  private val peerDiscovery: Stream[F, Unit] = Stream
+    .awakeEvery(10.seconds)
```
TheMMaciek, line 94: I would go with less frequent check - not that much should change in 10s I think. We could also tie it to snapshot pulling (push snapshots to two separate pipelines) and check facilitators back then - maybe in case there are new facilitators that we don't know we should run the discovery and not pull them if no additional facilitators could be detected.

marcinwadon: There can be many joins or drops in a single 10s, especially when nodes are joining at the beginning after restart. It should not be strictly coupled with polling the global snapshot because changes can be in between as well, at least with the current algorithm. We agreed that the periodic check would be sufficient for now

kpudlik (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/L0PeerDiscovery.scala
marcinwadon: image

TheMMaciek: I mean isn't sortedSet a set?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
TheMMaciek: 10s though is too frequent in my opinion anyway

PR #358, opened 2022-08-02
Validate and pick consecutive txs in L1 consensus

TheMMaciek:


modules/dag-shared/src/main/scala/org/tessellation/dag/transaction/filter/Consecutive.scala
```diff
@@ -2,33 +2,56 @@ package org.tessellation.dag.transaction.filter
...
 import scala.annotation.tailrec
 
 import org.tessellation.kryo.KryoSerializer
 import org.tessellation.schema.transaction.{Transaction, TransactionReference}
 import org.tessellation.security.Hashed
+import org.tessellation.security.signature.Signed
 
 object Consecutive {
+  val hashedTxOrder: Order[Hashed[Transaction]] =
+    Order.whenEqual(Order.by(-_.fee.value.value), Order[Hashed[Transaction]])
```
kpudlik, line 20: You can achieve more DRY solution here.
`Hashed[Transaction]` has already `Signed[Transaction]` inside so you can create `Order[Hashed[Transaction]]` by using contravariant functor:

```scala
import cats.syntax.contravariant._

val hashedTxOrder: Order[Hashed[Transaction]] = signedTxOrder.contramap(_.signed)
```

kpudlik (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/RoundData.scala
```diff
@@ -44,9 +54,22 @@ case class RoundData(
...
-  def formBlock(): Option[DAGBlock] =
-    NonEmptySet
-      .fromSet((ownProposal.transactions ++ peerProposals.values.flatMap(_.transactions).toSet).toSortedSet)
-      .map(DAGBlock(tips.value, _))
+  def formBlock[F[_]: Async: KryoSerializer: Logger](validator: TransactionValidator[F]): F[Option[DAGBlock]] =
+    (ownProposal.transactions ++ peerProposals.values.flatMap(_.transactions)).toList
+      .traverse(validator.validate)
+      .flatMap { validatedTxs =>
+        val invalid = validatedTxs.collect { case Invalid(e) => e }
+        val valid = validatedTxs.collect { case Valid(tx) => tx }
```
kriscieslak, line 63: What about this: `val (invalid, valid) = validatedTxs.partitionMap(_.toEither)` ?

modules/dag-shared/src/main/scala/org/tessellation/dag/transaction/filter/Consecutive.scala
```diff
@@ -2,33 +2,56 @@ package org.tessellation.dag.transaction.filter
...
+    takeGeneric[Id, TransactionReference, Hashed[Transaction]](
+      hashedTx => Id(TransactionReference.of(hashedTx)),
       _.parent,
       txs,
       lastAcceptedTxRef
-    )(Applicative[F], Order[TransactionReference], hashedTxOrder)
-  }
+    )(Applicative[Id], Order[TransactionReference], hashedTxOrder)
 
   private def takeGeneric[F[_]: Applicative, Ref: Order, Item: Order](
```
kriscieslak, line 56: It is possible here to substitute `Ref` with `TransactionReference`:

```
private def takeGeneric[F[_]: Applicative, Item: Order](
    refOf: Item => F[TransactionReference],
    prevRefOf: Item => TransactionReference,
    items: List[Item],
    initRef: TransactionReference
  )
```

modules/dag-shared/src/main/scala/org/tessellation/dag/transaction/filter/Consecutive.scala
tbekas: Is it necessary though? `Hashed` has auto unwrap. This should work

```scala
val hashedTxOrder: Order[Hashed[Transaction]] = Order.reverse(Order.by(_.fee))
```

modules/dag-shared/src/main/scala/org/tessellation/dag/transaction/filter/Consecutive.scala
tbekas: `Ref` is a parameter and `TransactionReference` is an argument to that parameter, so it already is the way you describe.

tbekas (dismissed):
LGTM, but please use the `partitionMap` that @kriscieslak mentioned.

kpudlik:
Please add the label :)

modules/dag-shared/src/main/scala/org/tessellation/dag/transaction/filter/Consecutive.scala
TheMMaciek: We not only need it to sort by fee but also to be deterministic when different nodes run the same selection process. So I need to order by fee and then by everything else...

modules/dag-shared/src/main/scala/org/tessellation/dag/transaction/filter/Consecutive.scala
kriscieslak: I just thought we can avoid explicitly passing `Ref: Order` this way, but I agree there is not much profit from this.

kriscieslak (approved):

kpudlik (approved):

PR #359, opened 2022-08-03
Use autogenerated properties for testing rewards

kpudlik:


modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -59,8 +63,7 @@ object Rewards {
           .flatTap {
             case (remaining, _) =>
               Applicative[F].whenA(remaining =!= Amount(0L))(
-                Applicative[F].unit
-//                logger.error(s"Some rewards were not distributed {amount=${amount.show}, remainingAmount=${remaining.show}}")
+                logger.error(s"Some rewards were not distributed {amount=${amount.show}, remainingAmount=${remaining.show}}")
```
kpudlik, line 66: If such case happens it means that there is an implementation error in distribution programs. The question is if we should return reward transactions then anyway.

modules/core/src/test/scala/org/tessellation/infrastructure/rewards/RewardsSuite.scala
```diff
@@ -2,141 +2,109 @@ package org.tessellation.infrastructure.rewards
...
       KryoSerializer.forAsync[IO](dagSharedKryoRegistrar.union(sdkKryoRegistrar)),
       SecurityProvider.forAsync[IO],
       Resource.liftK(Random.scalaUtilRandom[IO])
-    ).mapN((_, _, _))
-
-  implicit val order: Ordering[Id] = (x: Id, y: Id) => 1
+    ).mapN {
+      case (ks, sp, r) =>
+        implicit val _sp: SecurityProvider[IO] = sp
+        (ks, sp, r, () => KeyPairGenerator.makeKeyPair.map(_.getPublic.toId).unsafeRunSync())
```
kpudlik, line 41: I couldn't reuse `idGen` because... it generates incorrect `Id`s in terms of `Id => Address` conversion.

modules/core/src/test/scala/org/tessellation/infrastructure/rewards/RewardsSuite.scala
```diff
@@ -2,141 +2,109 @@ package org.tessellation.infrastructure.rewards
...
-        .calculateRewards(
-          EpochProgress(1L),
-          facilitators
-        )
-      sum = txs.toList.map(_.amount.value.toLong).sum
+  val lowerBound: NonNegLong = EpochProgress.MinValue.value
+  val upperBound: NonNegLong = config.rewardsPerEpoch.keySet.max.value
+  val lowerBoundNoMinting: NonNegLong = NonNegLong.unsafeFrom(upperBound.value + 1)
+  val special: Seq[NonNegLong] =
+    config.rewardsPerEpoch.keys.flatMap(epochEnd => Seq(epochEnd.value, NonNegLong.unsafeFrom(epochEnd.value + 1L))).toSeq
```
kpudlik, line 50: I check here the epoch's upper bound and also the next number after the upper bound to see if halving works correctly.

modules/core/src/main/scala/org/tessellation/config/types.scala
```diff
@@ -90,12 +87,9 @@ object types {
   object RewardsConfig {
 
     val default = RewardsConfig(
-      epochs = 4,
-      epochDurationInYears = 2.5,
-      baseEpochReward = Amount(85333333320000000L),
       softStaking = SoftStakingAndTestnetConfig(
         softStakeAddress = Address("DAG77VVVRvdZiYxZ2hCtkHz68h85ApT5b2xzdTkn"),
-        softStakeCount = 0L,
+        softStakeCount = 20L,
```
kpudlik, line 92: I set it to some value to be sure that it returns `> 0 ` for soft staking numerator. Maybe we should test against arbitrary values of different node counts instead of hardcoding it.

modules/core/src/test/scala/org/tessellation/infrastructure/rewards/RewardsSuite.scala
```diff
@@ -2,141 +2,109 @@ package org.tessellation.infrastructure.rewards
...
+    val gen = for {
+      epochProgress <- meaningfulEpochProgressGen
+      facilitators <- facilitatorsGen
+    } yield (epochProgress, facilitators)
+
+    forall(gen) {
+      case (epochProgress, facilitators) =>
+        for {
+          rewards <- Rewards.make[IO](config.rewardsPerEpoch, softStaking, dtm, stardust, regular).pure[F]
+          txs <- rewards.calculateRewards(epochProgress, facilitators)
```
kpudlik, line 81: I was thinking about extracting an interface like `rewards.distributeAmount(amount: Amount, facilitators: Facilitators)` but it seemed to me like making some internal interfaces public just to make it available for the test suite.

Of course Scala has this fancy mechanism: `private[rewards]` but we cannot use that properly since we use the `make` factory pattern (whatever it's called).

modules/core/src/main/scala/org/tessellation/config/types.scala
tbekas: Actually, it's not a bad idea. If you decide to do it, please check for the `@derive(arbitrary)` to derive an Arbitrary instance for config generator.

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
tbekas: Hm, yeah I think you're right. Maybe we should throw here.

modules/core/src/test/scala/org/tessellation/infrastructure/rewards/RewardsSuite.scala
tbekas: I think (I'm not sure though) that the generator will put higher weights also on neighbors of special so +1/-1. But I guess it won't hurt to keep it.

modules/core/src/test/scala/org/tessellation/infrastructure/rewards/RewardsSuite.scala
```diff
@@ -2,141 +2,109 @@ package org.tessellation.infrastructure.rewards
...
-      sum = txs.toList.map(_.amount.value.toLong).sum
-
-    } yield expect(sum == 16460905347L)
+    val gen = for {
+      epochProgress <- meaningfulEpochProgressGen
+      facilitators <- facilitatorsGen
+    } yield (epochProgress, facilitators)
+
+    forall(gen) {
+      case (epochProgress, facilitators) =>
```
tbekas, line 78: Nice! I like the idea of randomizing facilitator set.

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -21,30 +24,31 @@ import eu.timepit.refined.auto._
...
       ): F[SortedSet[RewardTransaction]] = {
-        val amount = getAmountByEpoch(epochProgress)
+        val amount = getAmountByEpoch(epochProgress, rewardsPerEpoch)
 
         val programRewardsState = for {
-          dtmRewards <- dtm.distribute(epochProgress, facilitators)
-          softStakingRewards <- softStaking.distribute(epochProgress, facilitators)
           stardustCollective <- stardust.distribute(epochProgress, facilitators)
+          softStakingRewards <- softStaking.distribute(epochProgress, facilitators)
+          dtmRewards <- dtm.distribute(epochProgress, facilitators)
```
tbekas, line 51: Nice that you've caught it. I was prototyping, and at some point, I reversed the order to use composition instead of sequencing, and then I forgot to revert it back.

tbekas (approved):

modules/core/src/test/scala/org/tessellation/infrastructure/rewards/RewardsSuite.scala
marcinwadon, line 41: @kpudlik why do you need `unsafeRunSync` here?

PR #360, opened 2022-08-03
Update dependencies versions to newest compatible

TheMMaciek:
Tested on the local k8s cluster. The whole flow with transactions being sent up to snapshots being created and passed from L0 to L1 works.


project/Dependencies.scala
```diff
@@ -5,32 +5,32 @@ object Dependencies {
...
     val doobie = "1.0.0-RC1"
     val droste = "0.9.0"
     val enumeratum = "1.7.0"
-    val h2 = "2.1.212"
+    val h2 = "2.1.214"
     val javaxCrypto = "1.0.1"
-    val flyway = "8.5.12"
-    val fs2 = "3.2.7"
-    val fs2Data = "1.3.1"
+    val flyway = "9.1.0"
```
kpudlik, line 21: It is a major update. Have you tried backward compatibility of migrations?

TheMMaciek: I guess if the cluster gets up from the scratch and works correctly then the migrations must work as well right?

kpudlik (approved):

tbekas (approved):

kriscieslak (approved):

PR #362, opened 2022-08-04
Use EpochProgress from previous snapshot for calculating rewards

kpudlik:


modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -87,10 +86,7 @@ object GlobalSnapshotConsensusFunctions {
       for {
         lastGSHash <- lastGS.value.hashF
         currentOrdinal = lastGS.ordinal.next
-        epochProgress = trigger match {
-          case EventTrigger => lastGS.epochProgress
-          case TimeTrigger  => lastGS.epochProgress.next
-        }
+        currentEpochProgress = lastGS.epochProgress.next
```
tbekas, line 89: I think we should advance epoch only on TimeTrigger

tbekas (changes requested):

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
kpudlik: Ah right. I reverted that change by accident

tbekas (approved):

PR #364, opened 2022-08-04
Throw in case of incorrect calculation of reward distribution

kpudlik:


modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -60,11 +62,14 @@ object Rewards {
...
-              Applicative[F].whenA(remaining =!= Amount(0L))(
-                logger.error(s"Some rewards were not distributed {amount=${amount.show}, remainingAmount=${remaining.show}}")
-              )
+          .flatMap {
+            case result @ (remaining, _) =>
+              if (remaining =!= Amount(0L)) {
+                val error =
+                  new RuntimeException(s"Some rewards were not distributed {amount=${amount.show}, remainingAmount=${remaining.show}}")
+                logger.error(error)(s"Distributing rewards failed: ${error.getMessage}") >> error
+                  .raiseError[F, (Amount, List[(Address, Amount)])]
```
tbekas, line 71: As a rule of thumb, errors should be either raised or logged, but not both. Otherwise, errors will be logged twice. Once where they're raised and once where it's handled.

tbekas (changes requested):

tbekas (approved):

PR #365, opened 2022-08-05
Set DTM ratio to reflect latest changes

kpudlik:


tbekas (approved):
LGTM however we need to decide on default args in config case class vs default value of config. Most of the config currently uses a default value.

PR #366, opened 2022-08-05
Proportion based algorithm for minting rewards

kpudlik:


modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -156,7 +155,7 @@ object GlobalSnapshotConsensusFunctions {
       } yield (globalSnapshot, returnedEvents)
     }
 
-    private def updateBalancesByTransactions(balances: SortedMap[Address, Balance], txs: SortedSet[RewardTransaction]) =
+    private def updateBalancesByRewards(balances: SortedMap[Address, Balance], txs: SortedSet[RewardTransaction]) =
```
kpudlik, line 158: Please take a look at that logic.
If someone overflows `NonNegLong` then his balance will be replaces with `tx.amount`. It is not the correct behavior but such case should never happen because the total supply is lower than `NonNegLong.MaxValue`. I can throw there if you guys want.

tbekas: Wait, I think you pointed out an error here. I believe it should be:

```scala
maybeBalance
  .map(balance => balance.plus(tx.amount).getOrElse(balance))
  .getOrElse(Balance.fromAmount(tx.amount))
```

kpudlik: Fallback to previous balance means that reward transition exists but it didn't affect the balance :< so also incorrect but the lesser of 2 evils

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -1,143 +1,88 @@
...
+    rewardsPerEpoch: SortedMap[EpochProgress, Amount],
+    softStaking: SimpleRewardsDistributor,
+    dtm: SimpleRewardsDistributor,
+    stardust: SimpleRewardsDistributor,
+    regular: RewardsDistributor[F]
   ): Rewards[F] =
     new Rewards[F] {
 
-      implicit val logger = Slf4jLogger.getLogger[F]
+      private val logger = Slf4jLogger.getLogger[F]
```
kriscieslak, line 42: `logger` is not used, can be removed.

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -1,143 +1,88 @@
...
-import eu.timepit.refined.cats._
-import eu.timepit.refined.types.numeric._
+import eu.timepit.refined.numeric.Positive
+import eu.timepit.refined.refineV
 import io.estatico.newtype.ops.toCoercibleIdOps
 import org.typelevel.log4cats.slf4j.Slf4jLogger
 
 object Rewards {
 
   def make[F[_]: Async: SecurityProvider](
```
kriscieslak, line 33: `SecurityProvider` is not needed here.

modules/core/src/test/scala/org/tessellation/infrastructure/rewards/RewardsSuite.scala
```diff
@@ -2,137 +2,126 @@ package org.tessellation.infrastructure.rewards
...
-    for {
-      rewards <- Rewards.make[IO](config, 1.minutes, softStaking, dtm, stardust).pure[F]
-      txs <- rewards
-        .calculateRewards(
-          EpochProgress(1296001L * 4),
-          facilitators
-        )
-      sum = txs.toList.map(_.amount.value.toLong).sum
+  test("generated reward transactions sum up to the total snapshot reward") { res =>
+    implicit val (_, sp, rand, makeIdFn) = res
```
kriscieslak, line 78: `rand` is not used.

modules/core/src/test/scala/org/tessellation/infrastructure/rewards/RewardsSuite.scala
```diff
@@ -2,137 +2,126 @@ package org.tessellation.infrastructure.rewards
...
+          sum = txs.toList.map(_.amount.value.toLong).sum
+          expected = rewards.getAmountByEpoch(epochProgress, config.rewardsPerEpoch).value.toLong
+
+        } yield expect(sum == expected)
+    }
+  }
 
-    } yield expect(sum == 0)
+  test("reward transactions won't be generated after the last epoch") { res =>
+    implicit val (_, sp, rand, makeIdFn) = res
```
kriscieslak, line 105: `rand` is not used.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
tbekas: Currently, the logic is the following: if someone has a very high balance, i.e. `Long.MaxValue` and we want to reward him with `1L`, then his balance will be set to `1L`. This is incorrect as the balance should not decrease when adding rewards.

Regarding burning vs redistributing, I think we've discussed it already and the outcome was that burning is a safer option than redistributing.

kpudlik: I mean the current logic is incorrect (that's why I started the discussion) but the solution od burning is also wrong because the total pool of snapshot will be  smaller than expected. But in the end such case won't happen so it is better to burn anyway. I just need to filter out the tx.

tbekas: I wouldn't consider burning someone's reward as an error. Burning someone's total balance is an error for sure.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -155,16 +153,25 @@ object GlobalSnapshotConsensusFunctions {
...
+      val (toBurn, toUpdate) = txs.toList.map { tx =>
+        balances
+          .getOrElse(tx.destination, Balance.empty)
+          .plus(tx.amount)
+          .map((tx.destination, _))
+          .leftMap(_ => tx)
+      }.partitionBifold(identity)
+
+      val updatedBalances = toUpdate.foldLeft(balances) {
+        case (acc, (address, balance)) => acc.updated(address, balance)
```
tbekas, line 169: This will work incorrectly if there are 2 reward txs for some address.  Please replace it with a version that accumulates balances.

```scala
private def acceptRewardTxs(
  balances: SortedMap[Address, Balance],
  txs: SortedSet[RewardTransaction]
): (SortedMap[Address, Balance], SortedSet[RewardTransaction]) =
  txs.foldLeft((balances, SortedSet.empty[RewardTransaction])) { (acc, tx) =>
    val (updatedBalances, acceptedTxs) = acc

    updatedBalances
      .getOrElse(tx.destination, Balance.empty)
      .plus(tx.amount)
      .map(balance => (updatedBalances.updated(tx.destination, balance), acceptedTxs + tx))
      .getOrElse(acc)
  }
```

tbekas (changes requested):

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
kpudlik: Ooops, how could I not have seen it 🙈

kpudlik: I see that you are collecting accepted transactions instead of failed ones but it is fine because I can just subtract accepted from all to get incorrect ones.

tbekas (approved):
LGTM

PR #367, opened 2022-08-08
Fee distribution rewards

kpudlik:


modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -1,143 +1,121 @@
...
-              logError.as(SortedSet.empty[RewardTransaction])
-            }
+        val totalFee = transactions.toList
+          .map(_.fee)
+          .foldM(NonNegLong.MinValue) { case (acc, fee) => acc + fee.value }
+          .map(Amount(_))
+          .liftTo[F]
+
+        totalFee.flatMap { amount =>
+          regular.distribute(epochProgress, facilitators).run(amount).flatMap {
```
tbekas, line 60: I think it's better to base this distribution on `SnapshotOrdinal` instead of the `EpochProgress`. This would require passing a different param to `Random` so I think it's maybe better to pass the `Random` instance implicitly to the `Regular.distribute` method.

This would mean that we need to scrap the `RewardsDistributor` interface altogether since it constrains us to use two specific parameters `epochProgress: EpochProgress, facilitators: NonEmptySet[Id]`.

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -1,143 +1,121 @@
...
+                )
+              } else result.pure[F]
+          }
+        }.map {
+          case (_, rewards) =>
+            rewards.flatMap {
+              case (address, amount) =>
+                refineV[Positive](amount.coerce.value).toList.map(a => RewardTransaction(address, TransactionAmount(a)))
+            }.toSortedSet
+        }
```
tbekas, line 74: Please extract this duplicated code to some `private def`

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -120,16 +118,18 @@ object GlobalSnapshotConsensusFunctions {
...
+        rewardTxs <- trigger match {
+          case EventTrigger =>
+            rewards.feeDistribution(lastGS.epochProgress, transactions, facilitators)
           case TimeTrigger =>
-            rewards
-              .calculateRewards(epochProgress, lastGS.proofs.map(_.id))
-              .map { txs =>
-                (updateBalancesByTransactions(updatedBalances, txs), txs)
-              }
+            rewards.mintedDistribution(lastGS.epochProgress, facilitators)
```
tbekas, line 128: We should do `feeDistribution` on `TimeTrigger` as well.

tbekas (changes requested):

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
kpudlik: Even if we use `SnapshotOrdinal` as a `seed` for both fee and time then making it implicit won't change much because `Random` is not a singleton, so I need to instantiate a new one for each snapshot anyway. Or maybe I didn't get it.

Scraping the `RewardDistributor` is fine because not all the programs use both `epochProgress` and `facilitators` so we end up with `case(_, _)` :+1:

tbekas: That's true, maybe it's better to pass it explicitly in this case.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -120,16 +118,18 @@ object GlobalSnapshotConsensusFunctions {
         updatedLastTxRefs = lastGS.info.lastTxRefs ++ acceptanceResult.contextUpdate.lastTxRefs
         updatedBalances = lastGS.info.balances ++ acceptanceResult.contextUpdate.balances
 
-        (updatedBalancesWithRewards, rewardTxs) <- trigger match {
-          case EventTrigger => (updatedBalances, SortedSet.empty[RewardTransaction]).pure[F]
+        facilitators = lastGS.proofs.map(_.id)
+        transactions = lastGS.value.blocks.flatMap(_.block.transactions.toSortedSet).map(_.value)
```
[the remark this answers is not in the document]
tbekas: After considering it again, I think that we have to reward the previous snapshot, so it's correct I think.

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -1,80 +1,113 @@
...
-              new RuntimeException(s"Some rewards were not distributed {amount=${amount.show}, remainingAmount=${remaining.show}}")
-                .raiseError[F, Unit]
-                .whenA(remaining =!= Amount.empty)
+          .flatMap {
+            case (remaining, rewards) =>
+              if (remaining =!= Amount.empty) {
+                MonadThrow[F].raiseError[List[(Address, Amount)]](
+                  new RuntimeException(s"Some rewards were not distributed {amount=${amount.show}, remainingAmount=${remaining.show}}")
+                )
+              } else rewards.pure[F]
```
tbekas, line 100: Please extract lines 95-100 and lines 61-67 into a common function. Also I think that the previous notation was more concise:

```scala
.flatTap {
          case (remaining, _) =>
            new RuntimeException(s"Some rewards were not distributed {amount=${amount.show}, remainingAmount=${remaining.show}}")
              .raiseError[F, Unit]
              .whenA(remaining =!= Amount.empty)
```

tbekas (changes requested):
Looks good, but please fix the duplicated code.

tbekas (approved):
LGTM

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
kpudlik: Cool, I didn't think about using `flatTap` for throwing. Looks nice

tbekas (approved):
LGTM

marcinwadon (approved):

PR #373, opened 2022-08-10
Log reward distribution summary

tbekas:


modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -72,6 +72,7 @@ object Rewards {
                   refineV[Positive](amount.coerce.value).toList.map(a => RewardTransaction(address, TransactionAmount(a)))
               }.toSortedSet
           }
+          .flatTap(rewardTxs => logger.info(s"Minted amount of ${amount.show} and distributed in ${rewardTxs.size} transaction(s)"))
```
TheMMaciek, line 75: Maybe "and distributed among n addresses"? Unless that's not exactly the same thing.

tbekas: I wanted to do that, but there's no guarantee that these are distinct addresses.

TheMMaciek (approved):

PR #382, opened 2022-08-17
Additional logs for transactions

marcinwadon:


modules/shared/src/main/resources/logback-included.xml
```diff
@@ -82,5 +96,12 @@
         <appender-ref ref="GOSSIP" />
         <appender-ref ref="JSON" />
     </logger>
+    <logger name="org.tessellation.dag.l1.http.Routes" level="DEBUG" additivity="false">
+        <appender-ref ref="TRANSACTIONS" />
```
tbekas, line 100: Maybe a `JSON` appender as well? If we have centralized logs, we're likely going to use json format.

tbekas (approved):

TheMMaciek (approved):

modules/shared/src/main/resources/logback-included.xml
marcinwadon: We have a plan for better logging, and I'd like to avoid spamming with these logs right now. Tomorrow we'll call that part

PR #385, opened 2022-08-18
Separate logger for transactions

marcinwadon:


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionStorage.scala
```diff
@@ -136,13 +137,15 @@ class TransactionStorage[F[_]: Async: KryoSerializer](
...
-                .toNes
-                .flatMap(_.filter(_.ordinal > lastTxState.ref.ordinal).toNes)
+            val maybeStillWaiting = SortedSet.from(waiting.toList.diff(pulled)).toNes
+            val maybeStillWaitingAboveOrdinal = maybeStillWaiting.flatMap(_.filter(_.ordinal > lastTxState.ref.ordinal).toNes)
+            val expiredBelowOrdinal = maybeStillWaiting.flatMap(_.filter(_.ordinal <= lastTxState.ref.ordinal).toNes)
 
-            setter(maybeStillWaiting)
+            setter(maybeStillWaitingAboveOrdinal).flatTap { _ =>
+              transactionLogger.info(
+                s"Ignored transaction with ordinal lower or equal ${lastTxState.ref.ordinal}: ${expiredBelowOrdinal.map(_.map(_.hash)).show}"
```
TheMMaciek, line 146: `Expired transaction...`

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/package.scala
```diff
@@ -0,0 +1,5 @@
+package org.tessellation.dag.l1.domain
+
+package object transaction {
+  val transactionLoggerName = "org.tessellation.dag.l1.domain.TransactionLogger"
```
TheMMaciek, line 4: Should we name the logger as a class that doesn't exist? Maybe we could call it `TransactionLogger` simply? Or you could actually create an Object `TransactionLogger` at the given package and have a function for instantiating that logger to be used wherever you need it. You will not have to import the name and get the logger by name everywhere and the class will actually exist.

marcinwadon: I made it the same way as the RumorLogger. I think it should be any arbitrary name, it can be just `TransactionLogger` as well I think. @tbekas wdyt?

tbekas: I agree, let's use an arbitrary name, rather than non-existing class name. @marcinwadon could you change the logger in gossip to `RumorLogger` as well?

marcinwadon: done :+1:

TheMMaciek (approved):

tbekas (approved):
LGTM

PR #386, opened 2022-08-20
Added a test for to handle Rosetta's Construction test

cngo-github:
__PR Category__: Test

Added code for Rosetta's [construction test](https://www.rosetta-api.org/docs/rosetta_cli.html#writing-checkconstruction-tests). This is to meet the requirements of integrating Rosetta. This is being merged into the feature branch for Rosetta.

### DIFF
--- a/modules/rosetta/src/test/scala/org/tessellation/rosetta/server/RosettaConstructionSuite.scala
+++ b/modules/rosetta/src/test/scala/org/tessellation/rosetta/server/RosettaConstructionSuite.scala
@@ -1,0 +1,278 @@
+package org.tessellation.rosetta.server
+
+import java.util.concurrent.TimeUnit
+
+import cats.effect.unsafe.implicits.global
+import cats.effect.{Async, IO, Resource}
+
+import scala.concurrent.duration.Duration
+import scala.io.Source
+
+import org.tessellation.ext.cats.effect.ResourceIO
+import org.tessellation.ext.crypto._
+import org.tessellation.kryo.KryoSerializer
+import org.tessellation.rosetta.server.runner.runnerKryoRegistrar
+import org.tessellation.sdk.config.types.{HttpClientConfig, HttpServerConfig}
+import org.tessellation.sdk.resources.MkHttpServer.ServerName
+import org.tessellation.sdk.resources.{MkHttpClient, MkHttpServer}
+import org.tessellation.security.SecurityProvider
+import org.tessellation.shared.sharedKryoRegistrar
+
+import com.comcast.ip4s.{Host, Port}
+import io.circe.parser.parse
+import io.circe.{Json, JsonNumber}
+import org.http4s._
+import org.http4s.circe._
+import org.http4s.client.Client
+import org.http4s.server.Server
+import org.typelevel.log4cats.SelfAwareStructuredLogger
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+import suite.ResourceSuite
+import weaver.scalacheck.Checkers
+
+object RosettaConstructionSuite extends ResourceSuite with Checkers {
+  override type Res = (Resource[IO, Server], Resource[IO, Client[IO]])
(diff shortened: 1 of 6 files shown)


modules/rosetta/src/test/scala/org/tessellation/rosetta/server/RosettaConstructionSuite.scala
```diff
@@ -0,0 +1,151 @@
...
+      .newEmber(HttpClientConfig(Duration(5, TimeUnit.SECONDS), Duration(10, TimeUnit.SECONDS)))
+
+    IO.pure((server: Resource[IO, Server], client: Resource[IO, Client[IO]])).asResource
+  }
+
+  test("Construction test") { implicit resources: Res =>
+    val serverObj = resources._1
+    val httpClientObj = resources._2
+
+    serverObj.use { server =>
```
cngo-github, line 51: _For anyone that can answer this_: Why do I have to have `server =>` here for the tests to run even though I don't call it ever?

modules/rosetta/src/test/scala/org/tessellation/rosetta/server/JsonObjects.scala
```diff
@@ -0,0 +1,26 @@
+package org.tessellation.rosetta.server
+
+import io.circe.generic.auto._
+import io.circe.generic.semiauto.{deriveDecoder, deriveEncoder}
+import io.circe.{Decoder, Encoder}
+
+object JsonObjects {
```
cngo-github, line 7: I only created the case class for the payload because that is the only object that I need in its entirety that is common throughout all the tests.

modules/rosetta/src/test/scala/org/tessellation/rosetta/server/RosettaConstructionSuite.scala
ryle-goehausen: This can just be `_ =>` -- it has to be invoked here to start the server in background but not actually referenced.

modules/rosetta/src/test/scala/org/tessellation/rosetta/server/RosettaConstructionSuite.scala
```diff
@@ -0,0 +1,151 @@
...
+        val jsonString = Source.fromResource("rosetta.test_transactions.json").getLines().mkString
+        val cursor = parse(jsonString).getOrElse(Json.Null).hcursor
+        val params = cursor.downField("item").as[List[Json]].getOrElse(List.empty[Json])
+
+        val testResults = params.map { testEntry =>
+          val entryCursor = testEntry.hcursor
+          val payloadString =
+            entryCursor.downField("request").downField("body").downField("raw").as[String].getOrElse("{}")
+          val payload = decode[RequestJson](payloadString) match {
+            case Left(_)      => Json.Null
```
ryle-goehausen, line 62: Is this ignoring all the requests that don't conform to this schema? For instance `{{HOST}}:8080/construction/combine` has fields not present in this RequestJson schema. The response string appears to correspond to all the expected responses here but are we submitting an empty payload here?

ryle-goehausen (commented):
We can review here, but I'll merge to new branch separately -- made this branch while dealing with conflicts from dev merge so it's out of date and doesn't conform to current interface / L1 module, but will pull in to other updated branch after resolving other issues.

modules/rosetta/src/test/scala/org/tessellation/rosetta/server/RosettaConstructionSuite.scala
cngo-github: I am no longer using case classes to avoid this problem.

cngo-github:
Fixed logic errors, fixed tests, and cleaned code. The construction tests work now.

cngo-github:
This will be merged into the branch `feature/rosetta_ref2`.

PR #387, opened 2022-08-22
Consensus peer registration

tbekas:


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -38,134 +43,150 @@ object ConsensusManager {
...
+        consensusStateUpdater.tryUpdateConsensus(key, resources).flatMap {
+          case Some(state) =>
+            state.status match {
+              case Finished(signedArtifact, majorityTrigger) =>
+                consensusStorage
+                  .tryUpdateLastKeyAndArtifactWithCleanup(state.lastKey, key, signedArtifact)
+                  .ifM(
+                    afterConsensusFinish(majorityTrigger),
+                    logger.info("Skip triggering another consensus")
+                  ) >> nodeStorage.tryModifyStateB(Observing, Ready).void
```
marcinwadon, line 145: `nodeStorage.tryModifyStateB(Observing, Ready).void`
isn't it like `nodeStorage.tryModifyState(Observing, Ready)`?

marcinwadon (approved):
Looks legit :+1: I try to understand whether there is no recursion loop in scheduling (spawns/threads) but looks fine.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
tbekas: Also, I don't like the name of this function `tryModifyStateB`, but I couldn't come up with a better one.

marcinwadon: got it :+1: thanks

modules/core/src/main/scala/org/tessellation/modules/HttpApi.scala
```diff
@@ -104,7 +104,8 @@ sealed abstract class HttpApi[F[_]: Async: SecurityProvider: KryoSerializer: Met
                 gossipRoutes.p2pRoutes <+>
                 trustRoutes.p2pRoutes <+>
                 globalSnapshotRoutes.p2pRoutes <+>
-                healthcheckP2PRoutes
+                healthcheckP2PRoutes <+>
+                services.consensus.p2pRoutes
```
TheMMaciek, line 108: We don't want to follow the convention of creating routes outside of the service?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
TheMMaciek: maybe `tryModifyIfState(expected: State, to: State)`?

tbekas: I'm not convinced by this name, since the other function also modifies state if `actual == expected` and this name would suggest that only this function is doing such a check. 

Essentially it behaves exactly the same as `tryModifyState` but returns `Boolean` instead of raising an error.

PR #392, opened 2022-08-23
Use in-memory storage for AddressStorage

marcinwadon:


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/address/storage/AddressStorage.scala
```diff
@@ -1,67 +1,32 @@
...
-              .map(_.headOption)
-              .flatMap {
-                case Some(_) => run(updateAddressBalance(lift(address), lift(balance)))
-                case None    => run(insertAddressBalance(lift(address), lift(balance)))
-              }
-        }.transact(xa).as(())
-
-      def clean: F[Unit] = run(getAddresses.delete).transact(xa).as(())
+        balances.update { m =>
+          addressBalances.foldLeft(m) {
```
tbekas, line 24: Isn't this the same as just `++` operator?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/address/storage/AddressStorage.scala
```diff
@@ -1,67 +1,32 @@
...
+  def make[F[_]: Async]: F[AddressStorage[F]] =
+    Ref.of[F, Map[Address, Balance]](Map.empty).map(make(_))
 
-  type Ctx = DoobieContext.SQLite[Literal] with Context[SqliteDialect, Literal]
-
-  def make[F[_]: MonadCancelThrow: Database]: AddressStorage[F] =
-    make[F](new DoobieContext.SQLite[Literal](Literal))
-
-  def make[F[_]: MonadCancelThrow](ctx: Ctx)(implicit db: Database[F]): AddressStorage[F] =
+  def make[F[_]: Async](balances: Ref[F, Map[Address, Balance]]): AddressStorage[F] =
```
TheMMaciek, line 17: Maybe we can clean some dependencies from build.sbt at this point as well?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/address/storage/AddressStorage.scala
marcinwadon: tbh I don't know but currently it looks clear imo. `++` may be ambiguous whether we combine balances or update them

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/address/storage/AddressStorage.scala
TheMMaciek: Also there were decoder/encoder pairs created for `Balance` for sure, probably also for `Address` and other data. We can clean it as well I think.

marcinwadon: I think we cannot because Database currently is used i.e. by trust, but I'll double check that. Anyway, I'd keep that in a separate pull request

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/address/storage/AddressStorage.scala
```diff
@@ -1,67 +1,32 @@
...
-            run(getAddressBalance(lift(address)))
-              .map(_.headOption)
-              .flatMap {
-                case Some(_) => run(updateAddressBalance(lift(address), lift(balance)))
-                case None    => run(insertAddressBalance(lift(address), lift(balance)))
-              }
-        }.transact(xa).as(())
-
-      def clean: F[Unit] = run(getAddresses.delete).transact(xa).as(())
+        balances.update { m =>
```
TheMMaciek, line 23: I guess `m` is `currentBalances` maybe we could call it that or something else but more descriptive than just `m`

marcinwadon: I follow the same pattern that we used before, but I can name it differently. on it

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/infrastructure/address/storage/AddressStorage.scala
marcinwadon: @tbekas you're right. I'm changing it rn

TheMMaciek (approved):

TheMMaciek (approved):

PR #394, opened 2022-08-24
Fix stopping processing join requests after one failure

kriscieslak:


modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
```diff
@@ -100,7 +100,9 @@ object Joining {
     def join: Pipe[F, P2PContext, Unit] =
       in =>
         in.evalMap { peer =>
-          joining.twoWayHandshake(peer, none) >>
+          joining.twoWayHandshake(peer, none).void.handleErrorWith { err =>
+            logger.error(err)(s"Joining to peer ${peer.show} failed")
+          } >>
```
marcinwadon, line 105: If we cannot join the peer, why we should make a peer discovery from that peer?

marcinwadon (changes requested):
as commented above

marcinwadon (approved):

PR #398, opened 2022-08-25
Version check on joining

marcinwadon:


TheMMaciek (dismissed):

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
```diff
@@ -242,6 +243,13 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
 
   private def validateHandshake(registrationRequest: RegistrationRequest, remoteAddress: Option[Host]): F[Unit] =
     for {
+
+      versionHash <- BuildInfo.version.hashF
```
tbekas, line 247: I think that the version should be passed from `version` property of `TessellationIOApp`. Otherwise we're just comparing the version of the `shared` library.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
```diff
@@ -242,6 +243,13 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
...
+      _ <-
+        if (registrationRequest.version === versionHash)
```
tbekas, line 249: Maybe a tiny byt shorter:

```scala
VersionMismatch.raiseError[F, Unit].whenA(registrationRequest.version =!= versionHash)
```

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
TheMMaciek, line 247: We could possibly calculate it once and then reuse? 🤔

marcinwadon: where/when?

marcinwadon: `version` property is exactly the same one (taken from BuildInfo from shared module)

tbekas: I'm not in favor of tiny optimizations

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
marcinwadon: Should I change all the conditions in Joining to match that?

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
TheMMaciek: Maybe pass it to make as a Fiber? 🤔

marcinwadon: isn't it overcomplicated then?

TheMMaciek: Or as Tomek suggests we could calculate it in the TessellationIOApp? 🤔

tbekas: Yes, but that's just a coincidence. If someone would build a different app based on `TessellationIOApp`, then it would compare the version of the `shared` library rather than an app.

TheMMaciek: Maybe it's not worth it after all :)

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
tbekas: Not necessarily. Generally, I think when we figure out some better way of doing something we can introduce it gradually (when we change or refactor older code).

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
tbekas: My suggestion was to not cache it. It's a rare event that someone is joining.

TheMMaciek: Yeah, so not worth the additional code/obfuscation.

tbekas (dismissed):

TheMMaciek (approved):

TheMMaciek (approved):

PR #399, opened 2022-08-25
Stricter snapshot processing conditions

marcinwadon:


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -229,6 +229,9 @@ class StateChannel[F[_]: Async: KryoSerializer: SecurityProvider: Random](
     .evalMap {
       _.traverse(result => logger.info(s"Snapshot processing result: ${result.show}")).void
     }
+    .handleErrorWith { error =>
+      Stream.eval[F, Unit](logger.error(error)(s"Unexpected error during snapshot processing"))
+    }
```
TheMMaciek, line 234: You may not throw in the non critical case inside process function (before any state update begin). But in this place you don't know wether the error was ciritical or not. The balances update could fail e.g. and we will still continue processing whereas that seems like a fatal error.

tbekas (approved):
LGTM

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/domain/snapshot/programs/SnapshotProcessorSuite.scala
```diff
@@ -821,14 +823,9 @@ object SnapshotProcessorSuite extends SimpleIOSuite with TransactionGenerator {
...
             (
-              Left(
-                UnexpectedCaseCheckingAlignment(
-                  snapshotHeight6,
-                  snapshotSubHeight0,
-                  snapshotOrdinal10,
-                  snapshotHeight6,
-                  snapshotSubHeight0,
-                  snapshotOrdinal12
+              Right(
```
TheMMaciek, line 826: In the next PR we can remove wrapping in Either, as we don't expect an error anymore.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/snapshot/programs/SnapshotProcessor.scala
```diff
@@ -236,6 +239,13 @@ sealed abstract class SnapshotProcessor[F[_]: Async: KryoSerializer: SecurityPro
               obsoleteToRemove
             )
           }
+
+      case Ignore(lastHeight, lastSubHeight, lastOrdinal, processingHeight, processingSubHeight, processingOrdinal) =>
+        logger
+          .warn(
+            s"Unexpected case during global snapshot processing - ignoring snapshot! Last: (height: $lastHeight, subHeight: $lastSubHeight, ordinal: $lastOrdinal) processing: (height: $processingHeight, subHeight:$processingSubHeight, ordinal: $processingOrdinal)."
```
TheMMaciek, line 246: We may also log the expected and actual lastSnapshotHash as these may also differ. But let's add it in another PR.

TheMMaciek (approved):

PR #408, opened 2022-08-31
Peer discovery returns only not seen at all peers

kriscieslak:


modules/sdk/src/test/scala/org/tessellation/sdk/domain/cluster/programs/PeerDiscoverySuite.scala
```diff
@@ -0,0 +1,126 @@
...
+    }
+  }
+
+  test("getPeers - only peers not saved in cluster should be cached") {
+    val gen = for {
+      commonPeers <- Gen.containerOf[Set, Peer](peerGen)
+      newPeers <- Gen.containerOf[Set, Peer](peerGen)
+      discoverFromPeer <- peerGen
+      selfPeer <- peerGen.map(_.copy(id = nodeId))
+      knownPeers <- Gen.containerOf[Set, Peer](peerGen)
```
marcinwadon, line 112: I see this gen is duplicated so it can be extracted (IMO second test case can omit selfPeer and discoverFromPeer, so you can reuse it for all cases)

marcinwadon (approved):

PR #421, opened 2022-09-08
Gossip deregistration request before leaving the cluster

marcinwadon:


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -176,17 +178,29 @@ object ConsensusManager {
...
-        case _ => Applicative[F].unit
-      }.compile.drain)
-      .as(manager)
+    Spawn[F].start(
+      nodeStorage.nodeStates
+        .filter(NodeState.leaving.contains)
+        .evalTap { _ =>
+          consensusStorage.getLastKeyAndArtifact.map {
+            _.map { case (key, _) => key }
+              .map(Deregistration(_))
```
tbekas, line 187: I think it should be a `Order[Key].max(lastKey.next, ownRegistration)`. In most cases the `lastKey.next` would be sufficient, however when we're still observing, the `lastKey.next` is 1 ordinal less than `ownRegistration`.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -176,17 +178,29 @@ object ConsensusManager {
...
+          }.flatMap(_.fold(Applicative[F].unit)(d => gossip.spread(d)))
```
tbekas, line 188: Maybe a `traverse(gossip.spread).void` instead?

tbekas (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
marcinwadon: oh, I meant it to be `lastKey.next`. anyway, I'll make max of these two

kpudlik (dismissed):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -176,17 +179,36 @@ object ConsensusManager {
...
+            case (lastKey, ownRegistration) => lastKey.map(_._1).flatMap(k => ownRegistration.map((k, _)))
+          }.map {
+            _.map {
+              case (lastKey, ownRegistration) => Order[Key].max(lastKey.next, ownRegistration)
+            }
+          }
+            .map(_.map(Deregistration(_)))
+            .flatMap {
+              _.traverse(event => gossip.spread(event))
+            }
```
tbekas, line 196: Sorry for being annoying, but I think we can shorten this expression to:
```scala
(consensusStorage.getLastKey, consensusStorage.getOwnRegistration)
  .tupled
  .flatTap(_.mapN(Order[Key].max).map(Deregistration(_)).traverse(gossip.spread))
```

With the introduction of this function to `ConsensusStorage`:
```scala
def getLastKey: F[Option[Key]] = lastKeyAndArtifactR.get.map(_._1F)
```

tbekas (dismissed):

tbekas (approved):
LGTM

PR #428, opened 2022-09-14
Generation data type

marcinwadon:


tbekas (approved):
LGTM

kpudlik (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/gossip/Gossip.scala
```diff
@@ -38,7 +39,7 @@ object Gossip {
     for {
       counter <- Ref.of[F, PosLong](PosLong(1L))
       time <- Clock[F].realTime
-      generation <- PosLong.from(time.toMillis).leftMap(new RuntimeException(_)).liftTo[F]
+      generation <- PosLong.from(time.toMillis).map(Generation(_)).leftMap(new RuntimeException(_)).liftTo[F]
```
tbekas, line 42: I think that we can move the creation of a generation out of gossip unless you want to do it in the following PRs.

tbekas (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/gossip/Gossip.scala
marcinwadon: I'm working on that right now in separate PR, when reusing it for the session token.

PR #429, opened 2022-09-14
Use Generation for peer's SessionToken

marcinwadon:


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/cluster/storage/SessionStorage.scala
```diff
@@ -34,6 +33,6 @@ object SessionStorage {
       def clearToken: F[Unit] = sessionToken.set(none)
 
       private def generateToken: F[SessionToken] =
-        uid.make[F, SessionToken]
+        Generation.make[F].map(SessionToken(_))
```
tbekas, line 36: Ok, I think I understand why we're drawing `Generation` each time. It's because we want more or less unique `X-Session-Token` for each consecutive invocation?

If that's the case then maybe something like this:

```scala
case class SessionToken(generation: Generation, salt: SessionSalt)
```

And encoded as header with some separator like this:
```
X-Session-Token: 12345:xyzabc
```

This way we could have a single generation for the whole app and a random & unique header.

marcinwadon: What do you mean? We generate `SessionToken` once only 🤔

marcinwadon: Ah, so I didn't want to reuse a single generation for peer session, cluster session, and gossip generation 🤔 do you think they should be equal? IMO there is nothing wrong with creating it separately per use case

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/headers/X-Session-Token.scala
```diff
@@ -1,27 +1,28 @@
...
+    Try(s.toLong).toEither.flatMap { a =>
+      PosLong.from(a).leftMap(err => new Throwable(err))
+    }.map(Generation(_))
       .leftMap(e => ParseFailure("Invalid X-Session-Token header", e.getMessage()))
       .map(t => `X-Session-Token`(SessionToken(t)))
 
   implicit val headerInstance: Header[`X-Session-Token`, Header.Single] = Header.createRendered(
     ci"X-Session-Token",
-    _.token.coerce[UUID].toString,
+    _.token.value.value.value.toString,
```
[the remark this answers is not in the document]
marcinwadon: I can do `_.coerce[Generation].coerce[PosLong].toString`. fixed

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/cluster/storage/SessionStorage.scala
kpudlik: For me it is fine to keep these things separated to avoid the parsing overhead of combined one.
Also the endpoint should return just the cluster session so nodectl does not need to trim the Peer session part for comparing. I would go for KISS

kpudlik: Unless someone wants to compare both sessions in one shot then keeping it combined everywhere makes sense

tbekas: @marcinwadon @kpudlik Ok, but the thing is that it may not be unique. Consecutive invocations of `Generation.make[F]` may return the same object and therefore reply attack may be possible.

tbekas: Ok, it's not an issue. We call it just once when joining the cluster, therefore it doesn't have to be unique.

tbekas (approved):

PR #430, opened 2022-09-14
Format FacilitatorCalculatorSuite

TheMMaciek:


kpudlik (approved):
This is what I needed because it keeps formatting that file every time I compile :D

PR #440, opened 2022-09-20
Local healthcheck

marcinwadon:


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/healthcheck/LocalHealthcheck.scala
```diff
@@ -0,0 +1,94 @@
...
+              case Some(f) => (Some(f), false)
+              case _       => (Some(d.get), true)
+            }.ifM(
+              spawn(peer).flatMap(d.complete).void,
+              Applicative[F].unit
+            )
+          }
+      }
+
+    def cancel(peer: Peer): F[Unit] = cancel(peer.id)
```
kpudlik, line 61: As you can see, you use here `cancel(peer.id)` anyway instead of `cancel(peer)`

kpudlik (dismissed):

kpudlik (approved):

PR #447, opened 2022-09-23
Log gossip errors with error classname

tbekas:


tbekas:
Invalid rebase and invalid branch name

modules/shared/src/main/scala/org/tessellation/schema/package.scala
```diff
@@ -75,4 +75,6 @@ trait OrphanInstances {
 
   implicit val nonNegLongOrder: Order[NonNegLong] =
     orderOf[Long, NonNegative]
+
+  implicit val errorShow: Show[Throwable] = e => s"${e.getClass.getName}{message=${e.getMessage}}"
```
kpudlik, line 79: :+1:

kpudlik (approved):

PR #448, opened 2022-09-23
Log gossip errors with error classname

tbekas:


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/gossip/GossipRoundRunner.scala
```diff
@@ -40,7 +42,7 @@ object GossipRoundRunner {
             }
             .flatMap(_ => metrics.incrementGossipRoundSucceeded)
             .handleErrorWith { err =>
-              logger.error(s"Error running gossip round {peer=${peer.show}, reason=${err.getMessage}") >>
+              logger.error(s"Error running gossip round {peer=${peer.show}, reason=${err.show}") >>
```
marcinwadon, line 45: do you think `logger.error(err)(s"Error running gossip round {peer=${peer.show}")` would be too verbose?

marcinwadon (dismissed):

PR #452, opened 2022-09-26
Use coinbase hash in genesis snapshot

TheMMaciek:


modules/core/src/main/scala/org/tessellation/Main.scala
```diff
@@ -152,17 +152,19 @@ object Main
             NodeState.GenesisReady
           ) {
             GenesisLoader.make[IO].load(m.genesisPath).flatMap { accounts =>
-              val genesis = GlobalSnapshot.mkGenesis(
+              val genesisF = GlobalSnapshot.mkGenesis[IO](
```
marcinwadon, line 155: maybe:
```scala
def mkGenesis = GlobalSnapshot.mkGenesis..

mkGenesis.flatMap { genesis =>
  ...
}
```

marcinwadon (approved):

modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/Coinbase.scala
```diff
@@ -0,0 +1,18 @@
...
+
+object Coinbase {
+
+  val value: String = "Nature loves courage. You make the commitment and nature will respond to that commitment " +
+    "by removing impossible obstacles. Dream the impossible dream and the world will not grind you under, it will " +
+    "lift you up. This is the trick. This is what all these teachers and philosophers who really counted, who really " +
+    "touched the alchemical gold, this is what they understood. This is the shamanic dance in the waterfall. This is " +
+    "how magick is done. By hurling yourself into the abyss and discovering it's a feather bed."
+
+  def getHash[F[_]: MonadThrow: KryoSerializer]: F[Hash] = value.hashF
```
tbekas, line 17: Are we sure that the calculated version is the best option? I wouldn't mind seeing a hardcoded hash with a test that verifies if the calculated hash of a string matches the hardcoded hash.

TheMMaciek: We can do it the other way round as you describe it. Non of these solutions is perfect I think.

tbekas: I would go for it. It will be easier to handle it in other than JVM languages.

marcinwadon: @tbekas both will be done once only, so IMO no difference

tbekas (approved):

marcinwadon (approved):

PR #465, opened 2022-10-01
Prevent drawing the same peer more than once for a gossip round

tbekas:


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/gossip/GossipRoundRunner.scala
```diff
@@ -28,39 +29,55 @@ object GossipRoundRunner {
     round: Peer => F[Unit],
     roundLabel: String,
     cfg: GossipRoundConfig
-  ): GossipRoundRunner[F] = new GossipRoundRunner[F] {
-    private val logger = Slf4jLogger.getLogger[F]
+  ): F[GossipRoundRunner[F]] =
+    for {
+      selectedPeersQueue <- Queue.bounded[F, Peer](cfg.maxConcurrentRounds.value * 2)
```
marcinwadon, line 34: is it correct? I remember you said that gossip round is now for one peer 🤔

tbekas: Yep, it is for one, because we offer at most 1 item to that queue. The size of that queue is not very important.

marcinwadon (approved):

kpudlik (approved):

PR #471, opened 2022-10-05
Fix memory leak in L0 consensus storage

TheMMaciek:
FYI: this also impacts the way the healthcheck will work from now on for the corner case of checking proposal for an obsolete snapshot consensus. Example: if some slower node starts consensus for snapshot 10 when our node has already finished consensus for snapshot 11 we will respond with not required even if we had a proposal at height 10 originally - we will have those proposals cleaned up already. After analyzing it it doesn't seem like something we should be concerned about in the algorithm logic.


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStorage.scala
```diff
@@ -184,7 +184,7 @@ object ConsensusStorage {
         private def cleanupStateAndResource(key: Key): F[Unit] =
           condModifyState[Unit](key) { _ =>
             (none[ConsensusState[Key, Artifact]], ()).some.pure[F]
-          }.void
+          }.void >> cleanResources(key)
```
tbekas, line 187: Let's just inline this (no need for a new def).

tbekas (approved):

PR #478, opened 2022-10-06
Unidirectional registration

tbekas:
Code and flow simplification


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -239,8 +222,10 @@ object ConsensusManager {
             none[Peer]
         }
           .filter(_.isResponsive)
-          .filter(selfId < _.id)
-          .enqueueUnterminated(peersForRegistrationExchange)
+          .parEvalMapUnbounded { peer =>
```
marcinwadon, line 225: :+1:

modules/sdk/src/main/scala/org/tessellation/sdk/package.scala
```diff
@@ -58,9 +57,6 @@ package object sdk {
       classOf[IDN] -> 528,
       classOf[ConsensusPeerDeclaration[_, _]] -> 529,
       EventTrigger.getClass -> 530,
-      TimeTrigger.getClass -> 531,
-      classOf[RegistrationExchangeRequest[_]] -> 532,
-      classOf[RegistrationExchangeResponse[_]] -> 533,
-      classOf[Deregistration[_]] -> 534
```
marcinwadon, line 64: it's not needed because of json-encoding in gossip, right?

marcinwadon (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/package.scala
tbekas: Yes, that's the reason

PR #483, opened 2022-10-13
Use refined HexString instead of any Hex

tbekas:


modules/shared/src/main/scala/org/tessellation/security/Hash.scala
```diff
@@ -1,40 +1,47 @@
...
+  case class Hash(value: HexString64)
 
   object Hash {
 
     def fromBytes(bytes: Array[Byte]): Hash =
-      Hash(Hashing.sha256().hashBytes(bytes).toString)
+      Hash(refineV[HexString64Spec].unsafeFrom(Hashing.sha256().hashBytes(bytes).asBytes().toHexString))
 
-    def empty: Hash = Hash(s"%064d".format(0))
+    def empty: Hash = Hash("0000000000000000000000000000000000000000000000000000000000000000")
```
kpudlik, line 34: How about `Hash(0.toString * 64)`?

modules/core/src/test/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotStorageSuite.scala
```diff
@@ -61,7 +60,11 @@ object GlobalSnapshotStorageSuite extends MutableIOSuite with Checkers {
             SortedMap.empty,
             SortedSet.empty,
             genesis.value.epochProgress,
-            NonEmptyList.of(PeerId(Hex("peer1"))),
+            NonEmptyList.of(
+              PeerId(
+                "00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
```
kpudlik, line 65: How about `0.toString * 128` or `"0" * 128`?

modules/shared/src/main/scala/org/tessellation/schema/hex.scala
```diff
@@ -0,0 +1,103 @@
...
+import eu.timepit.refined.numeric.Even
+import eu.timepit.refined.refineV
+import eu.timepit.refined.scalacheck.arbitraryRefType
+import eu.timepit.refined.string.MatchesRegex
+import org.scalacheck.{Arbitrary, Gen}
+import shapeless.Witness
+
+object hex {
+
+  type HexStringSpec = And[MatchesRegex["^[0-9a-f]*$"], Size[Even]]
```
kpudlik, line 27: Or alternative syntax to read ltr:

`MatchesRegex["^[0-9a-f]*$"] And Size[Even]`

modules/shared/src/main/scala/org/tessellation/schema/hex.scala
```diff
@@ -0,0 +1,103 @@
...
+  type HexString = Refined[String, HexStringSpec]
```
kpudlik, line 28: Sometimes you have `Refined[String, A]` and sometimes `String Refined A`.
I think we should unify it. My personal preference is ltr but up to you:
```
String Refined A
A And B
```

instead of

```
Refined[String, A]
And[A, B]
```

kpudlik (changes requested):

marcinwadon (dismissed):

modules/shared/src/main/scala/org/tessellation/schema/hex.scala
tbekas: True, I will switch to infix-only notation.

modules/shared/src/main/scala/org/tessellation/security/Hash.scala
tbekas: Yeah, it won't compile, it has to be a literal

marcinwadon (approved):

kpudlik (approved):

tbekas:
Unfortunately, this change can't be merged, because it causes an issue with the deserialization of a GlobalSnapshot.

PR #485, opened 2022-10-13
Move L0 service to sdk and add functionality

TheMMaciek:


build.sbt
```diff
@@ -237,7 +237,7 @@ lazy val testShared = (project in file("modules/test-shared"))
   )
 
 lazy val sdk = (project in file("modules/sdk"))
-  .dependsOn(shared % "compile->compile;test->test", testShared % Test, keytool, kernel)
+  .dependsOn(shared % "compile->compile;test->test", testShared % Test, keytool, kernel, dagShared)
```
tbekas, line 240: As a general thought, we probably need to rethink how to organize modules. W have to many modules I believe, without having a clear separation/responsibility.
`kernel`, `shared`, `test-shared`, `dag-shared`, `sdk` and also `keytool` (which is also used as a library).

tbekas (approved):

build.sbt
marcinwadon: I'd reorganize them as part of the Tessellation SDK work from the roadmap

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/services/L0Service.scala
```diff
@@ -1,38 +1,54 @@
...
     l0ClusterStorage: L0ClusterStorage[F],
-    lastSnapshotStorage: LastGlobalSnapshotStorage[F]
+    lastSnapshotStorage: LastGlobalSnapshotStorage[F],
+    singlePullLimit: Option[PosLong]
   ): L0Service[F] =
     new L0Service[F] {
 
       private val logger = Slf4jLogger.getLogger[F]
 
+      def pullGlobalSnapshotByOrdinal(ordinal: SnapshotOrdinal): F[Option[Hashed[GlobalSnapshot]]] =
```
marcinwadon, line 39: maybe just:
```
def pullGlobalSnapshot(ordinal: SnapshotOrdinal): ...
```

Argument is self-explanatory imo

build.sbt
TheMMaciek: I agree. It would be best to do it before or at least during sdk development.

marcinwadon:
@TheMMaciek The commit message says nothing about the changes you made. "add functionality" or "some changes" isn't describing what changes, or what functionality you've added

kpudlik (approved):

kriscieslak (approved):

PR #487, opened 2022-10-13
Create snapshot validator with initial implementation

TheMMaciek:
blocked by:
https://github.com/Constellation-Labs/tessellation/pull/485


modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/Validator.scala
```diff
@@ -0,0 +1,19 @@
...
+object Validator {
+
+  def isNextSnapshot(previous: Hashed[GlobalSnapshot], next: Hashed[GlobalSnapshot]): Boolean =
+    previous.hash === next.lastSnapshotHash &&
+      previous.ordinal.next === next.ordinal &&
+      (
+        previous.height < next.height && next.subHeight === SubHeight.MinValue ||
+          previous.height === next.height && previous.subHeight.next === next.subHeight
+        )
+}
```
marcinwadon, line 19: Do you think it can return `Validated` instead of `Boolean`? You could split the conditionals and name them, as currently, it's a pretty complex condition.

TheMMaciek: It can but I would rather improve it to return info that it is either invalid, or valid in which case it says if it's the next snapshot by height or subheight that way we can reuse it in L1. Anyway for now I only need the boolean so that's what I implemented. Naming conditionals makes sense.

marcinwadon (approved):

PR #493, opened 2022-10-19
Fix CL_COLLATERAL env normalization and rumor logging

tbekas:


marcinwadon (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/cli/CliMethod.scala
```diff
@@ -33,7 +32,7 @@ trait CliMethod {
     CollateralConfig(
       amount = amount
         .filter(_ => environment != Mainnet)
-        .getOrElse(Amount(NonNegLong.unsafeFrom(250_000L * normalizationFactor)))
+        .getOrElse(Amount(250_000_00000000L))
```
marcinwadon, line 35: imo it was more clear before

tbekas: We use this format already in `RewardsConfig`. Also, multiplication means we cannot use auto-refine (we have to use `unsafeFrom`).

PR #494, opened 2022-10-21
Add info endpoint with peers registered for consensus

kriscieslak:


TheMMaciek (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusRoutes.scala
```diff
@@ -2,23 +2,45 @@ package org.tessellation.sdk.infrastructure.consensus
...
-    case GET -> Root / "consensus" / "registration" =>
+  private val prefixPath = "/consensus"
+
+  private val p2p: HttpRoutes[F] = HttpRoutes.of[F] {
+    case GET -> Root / "registration" =>
       storage.getOwnRegistration.map(RegistrationResponse(_)).flatMap(Ok(_))
   }
 
+  private val public: HttpRoutes[F] = HttpRoutes.of[F] {
+    case GET -> Root / "info" =>
```
tbekas, line 29: I don't think that `info` is a good name for returning a list of peers. Maybe `peers` or `next-peers`?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -42,7 +42,7 @@ trait ConsensusManager[F[_], Key, Artifact] {
   private[consensus] def facilitateOnEvent: F[Unit]
   private[consensus] def checkForStateUpdate(key: Key)(resources: ConsensusResources[Artifact]): F[Unit]
   private[sdk] def checkForStateUpdateSync(key: Key)(resources: ConsensusResources[Artifact]): F[Unit]
-
+  private[consensus] def registeredPeers: F[List[PeerId]]
```
tbekas, line 45: This is not a method that in any way "manages" consensus, therefore  I don't think it should be in `ConsensusManager`

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusRoutes.scala
```diff
@@ -2,23 +2,45 @@ package org.tessellation.sdk.infrastructure.consensus
...
+      {
```
marcinwadon, line 30: IMO this endpoint returns registered peers in consensus, so it's not a consensus info.
and I'd definitely return the ordinal for which you return these peers, as it may change ordinal by ordinal. What do you think guys @kriscieslak @kpudlik @tbekas @TheMMaciek?

marcinwadon (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -197,6 +198,14 @@ object ConsensusManager {
       private def afterTimeTrigger: F[Unit] =
         scheduleFacility >> consensusStorage.containsTriggerEvent
           .ifM(internalFacilitateWith(EventTrigger.some), Applicative[F].unit)
+
+      private[consensus] def registeredPeers: F[List[PeerId]] =
+        (consensusStorage.getLastKey, consensusStorage.getOwnRegistration).flatMapN {
+          case (Some(lastKey), Some(_)) =>
+            consensusStorage.getRegisteredPeers(lastKey.nextN(2L)).map(selfId :: _)
```
tbekas, line 205: Why are we asking for registrations for a key +2 from the last one?

tbekas (changes requested):
What's the use case for this endpoint on L1?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
kriscieslak: When node is in `Observing` status it is registered to consensus with last key + 2. But as we spoke, it is safe to send blocks to it, so I want to return it as well.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
kriscieslak: Should I introduce sth like `ConsensusService`?

tbekas: A private method in `ConsensusRoutes` would be enough I think.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
tbekas: I don't see how asking for +2 changes anything. Maybe it's better to query for facilitators of the last finished consensus? If we're still in the Observing state we won't have any finished consensus yet.

marcinwadon: @kriscieslak you need to think about this endpoint not by its use-case but by the meaning what it provides (what API client can expect).

`/consensus/info` -> as an API client I'd expect a consensus info (registered peers may be one of its properties, but definitely not the consensus info
`/consensus/peers`  -> as an API client, I'd expect to get a list of peers taking part in a consensus
or as @tbekas suggested `/consensus/{ordinal}/registrations` -> as an API client, I'd expect to get a list of registrations

Having that in mind, there should not be peers in "Observing" as they are not part of consensus yet. correct me if I'm wrong @tbekas

tbekas: Ok, so maybe then
```
/consensus/{ordinal}/peers
```
and
```
/consensus/latest/peers

**Review by tbekas — APPROVED** [review]

File: modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusRoutes.scala [hunk 5]
```diff
@@ -1,24 +1,59 @@
...
-    case GET -> Root / "consensus" / "registration" =>
+  private val prefixPath = "/consensus"
+
+  private val p2p: HttpRoutes[F] = HttpRoutes.of[F] {
+    case GET -> Root / "registration" =>
       storage.getOwnRegistration.map(RegistrationResponse(_)).flatMap(Ok(_))
   }
 
+  private val public: HttpRoutes[F] = HttpRoutes.of[F] {
+    case GET -> Root / "latest/peers" =>
```
kpudlik, line 35: WDYT about `peers/latest`? Because `latest` is a sub resource of `peers`. Not the other way.

kpudlik (changes requested):

kpudlik (dismissed):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusRoutes.scala
marcinwadon, line 35: actually it should be `"latest" / "peers"`

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusRoutes.scala
```diff
@@ -1,24 +1,59 @@
...
+      {
+        for {
+          clusterPeers <- cluster.info
+          registeredPeers <- allRegisteredPeers
+        } yield clusterPeers.filter(p => registeredPeers.contains(p.id))
+      }.flatMap(Ok(_))
```
marcinwadon, line 41: imo you should not mix for comprehension and chaining syntax. For comp. already uses flatMap, so either write it using chaining (flatMap) or for comprehension.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusRoutes.scala
```diff
@@ -1,24 +1,59 @@
...
+  }
+
+  private def allRegisteredPeers: F[List[PeerId]] =
+    (storage.getLastKey, storage.getOwnRegistration).flatMapN {
+      case (Some(lastKey), Some(_)) =>
+        storage.getRegisteredPeers(lastKey).map(selfId :: _)
+      case _ => Applicative[F].pure(List[PeerId](selfId))
```
marcinwadon, line 48: shouldn't it be:
```scala
...
case (None, Some(_)) => selfId
case _ => List.empty
```
?

and I'm trying to think if there can be a consensus without selfId 🤔

kriscieslak: Good point! I've impcoved it, but I've also added dependency that selfId should be returned only when it is registered for last key or earlier.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusRoutes.scala
```diff
@@ -1,24 +1,61 @@
...
   }
 
+  private val public: HttpRoutes[F] = HttpRoutes.of[F] {
+    case GET -> Root / "latest/peers" =>
+      for {
+        clusterPeers <- cluster.info
+        registeredPeers <- allRegisteredPeers
+        peers = clusterPeers.filter(p => registeredPeers.contains(p.id))
+        result <- Ok(peers)
+      } yield result
```
marcinwadon, line 41: I see there is no information about consensus ordinal, but it should be
image

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/info.scala
```diff
@@ -0,0 +1,15 @@
...
+import org.tessellation.schema.peer.PeerInfo
+
+import derevo.circe.magnolia._
+import derevo.derive
+
+object info {
+
+  @derive(encoder)
+  case class ConsensusInfo[Key](
+    latest: Option[Key],
```
marcinwadon, line 12: It's not really reusable if you specify `latest` as a property. I'd name it:
```
{
 key: ..
 peers: [..]
}
```.

Also, why it's an `Option[Key]`. Is it possible to return peers without the consensus key they're participate in? 🤔

**Review by marcinwadon — CHANGES_REQUESTED** [review]

File: modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/info.scala (hunk 9 above)
Comment (kriscieslak) [reply modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/info.scala]: You are right, it shouldn't be possible. I've changed it to return `NotFound` status in such situation.

File: modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusRoutes.scala [hunk 10]
```diff
@@ -1,24 +1,69 @@
...
+      storage.getLastKey.flatMap {
+        case Some(key) => Ok(consensusInfo(key))
+        case _         => NotFound()
+      }
+  }
+
+  private def consensusInfo(latestKey: Key) = for {
+    registeredPeers <- allRegisteredPeers(latestKey)
+    peers <- filterClusterPeers(registeredPeers)
+  } yield ConsensusInfo(latestKey, peers)
```
marcinwadon, line 45: Since you wrote everything in chaining style, I'd encourage you to write it too:
```
allRegisteredPeers(latestKey)
  .flatMap(filterClusterPeers)
  .map(ClusterInfo(latestKey, _))
```

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/info.scala
```diff
@@ -0,0 +1,15 @@
...
+
+import derevo.circe.magnolia._
+import derevo.derive
+
+object info {
+
+  @derive(encoder)
+  case class ConsensusInfo[Key](
+    key: Key,
+    peersInfo: Set[PeerInfo]
```
marcinwadon, line 13: Why not "peers"? IMO it's more intuitive as it's just a DTO for http api. You query for /peers. PeerInfo is just our internal name, to distinguish it from Peer class.

marcinwadon (approved):

PR #496, opened 2022-10-27
Filter out zero-balance addresses when creating global snapshot

kriscieslak:


modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -127,6 +127,7 @@ object GlobalSnapshotConsensusFunctions {
 
         updatedLastTxRefs = lastArtifact.info.lastTxRefs ++ acceptanceResult.contextUpdate.lastTxRefs
         balances = lastArtifact.info.balances ++ acceptanceResult.contextUpdate.balances
+        balancesNonZero = balances.filter { case (_, balance) => balance != Balance.empty }
```
marcinwadon, line 130: `positiveBalances` maybe?

marcinwadon: also, please use Eq if possible

PR #499, opened 2022-11-04
Validate artifact when choosing majority in consensus

marcinwadon:


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -266,34 +267,41 @@ object ConsensusStateUpdater {
...
-                } >> Applicative[F].whenA(majorityHash === proposalInfo.proposalArtifactHash) {
-                  gossip.spreadCommon(ConsensusArtifact(key, proposalInfo.proposalArtifact))
-                } >> Metrics[F].recordDistribution(
-                  "dag_consensus_proposal_affinity",
-                  proposalAffinity(allProposals, proposalInfo.proposalArtifactHash)
-                )
-              }.void
-              (newState, effect).pure[F]
+            (maybeLastArtifact match {
+              case Some(lastSignedArtifact) => pickValidatedMajority(lastSignedArtifact, majorityTrigger, resources)(allProposals)
```
tbekas, line 287: Let's use either `.map` combined with `.getOrElse` instead of the lengthy pattern match.

```
maybeLastArtifact.map { lastSignedArtifact =>
 ...
}.getOrElse(...)
```

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -338,6 +346,27 @@ object ConsensusStateUpdater {
...
+    private def pickValidatedMajority(
+      lastSignedArtifact: Signed[Artifact],
+      trigger: ConsensusTrigger,
+      resources: ConsensusResources[Artifact]
+    )(proposals: List[Hash]): F[Option[Hash]] =
+      proposals
+        .foldMap(a => Map(a -> 1))
+        .toList
+        .map(_.swap)
+        .sortBy(_._1)
```
tbekas, line 358: 1. This sorts only by `count` (occurrences). While it should sort by both `(count, hash)`
2. Default sort order is ascending, so aren't you taking the value with the smallest count first?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -338,6 +346,27 @@ object ConsensusStateUpdater {
...
+        .foldLeftM(none[Hash]) {
+          case (None, (_, majorityHash)) =>
```
tbekas, line 360: Before `foldLeftM` Please map all items to `hash` instead of `(count, hash)` since `count` is unused

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -338,6 +346,27 @@ object ConsensusStateUpdater {
...
+            resources.artifacts
+              .get(majorityHash)
+              .traverse { artifact =>
+                consensusFns.isValidProposal(lastSignedArtifact, trigger)(artifact)
+              }
+              .map(_ => majorityHash.some)
```
tbekas, line 366: This will ignore the result of validation, right?

tbekas (changes requested):

tbekas (changes requested):
Each time we find an invalid majority artifact, we should log this event

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -78,6 +79,25 @@ object GlobalSnapshotConsensusFunctions {
...
+    ): F[Boolean] = {
+      val events = proposal.blocks.unsorted.map(_.block.asRight[StateChannelOutput])
+
+      def proposalArtifact: F[GlobalSnapshot] = createProposalArtifact(lastSignedArtifact.ordinal, lastSignedArtifact, trigger, events)
+        .map(_._1)
+        .map { artifact =>
+          artifact.copy(
+            stateChannelSnapshots = proposal.stateChannelSnapshots,
+            info = artifact.info.copy(lastStateChannelSnapshotHashes = proposal.info.lastStateChannelSnapshotHashes)
+          )
```
tbekas, line 93: I think it wouldn't be a big deal if we would fix it right away instead doing it in two steps

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
marcinwadon: good catch :+1:

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
marcinwadon: fixed

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
marcinwadon: I'd like to keep those changes separated, as we don't have migration plan for testnet and I do not want to delay the deployment of this fix.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -338,6 +348,35 @@ object ConsensusStateUpdater {
...
+        .map(_.swap)
+        .sortBy { case (occurrences, hash) => (-occurrences, hash) }
+        .foldLeftM(none[Hash]) {
+          case (None, (occurrences, majorityHash)) =>
+            resources.artifacts
+              .get(majorityHash)
+              .traverse { artifact =>
+                consensusFns.isValidProposal(lastSignedArtifact, trigger)(artifact).map(_.toOption)
+              }
+              .map(_.flatten)
```
tbekas, line 368: One more crucial thing. We should differentiate between not finding an artifact and finding an invalid artifact. In the current implementation, you reduce it to the same thing, while it's not the same.

If we didn't find an artifact by hash in `resources.artifacts` we should return `None` as the result of the function `pickValidatedMajority` (because receiving artifact is asynchronous with receiving declarations so we may get artifact after we got all the declarations)

If we found an artifact by hash, but it wasn't a valid artifact then we log it and move to the next hash (and repeat both conditions above).

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -338,6 +348,35 @@ object ConsensusStateUpdater {
...
+    private def pickValidatedMajority(
+      lastSignedArtifact: Signed[Artifact],
+      trigger: ConsensusTrigger,
+      resources: ConsensusResources[Artifact]
+    )(proposals: List[Hash]): F[Option[Hash]] =
+      proposals
+        .foldMap(a => Map(a -> 1))
+        .toList
+        .map(_.swap)
+        .sortBy { case (occurrences, hash) => (-occurrences, hash) }
```
tbekas, line 360: Still, invalid sorting on the 2nd param (not the same as in `pickMajority`). This should work

```
.toList
.map(_.swap)
.sorted
.reverse
```

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -338,6 +348,35 @@ object ConsensusStateUpdater {
...
+            resources.artifacts
+              .get(majorityHash)
+              .traverse { artifact =>
+                consensusFns.isValidProposal(lastSignedArtifact, trigger)(artifact).map(_.toOption)
+              }
+              .map(_.flatten)
+              .map(_.map(_ => majorityHash))
+              .flatTap {
+                _.fold {
+                  logger.warn(s"Found invalid majority hash=${majorityHash.show} with occurrences=${occurrences}, selecting next majority")
```
tbekas, line 372: I like that you used `warn` level 👍

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -338,6 +348,35 @@ object ConsensusStateUpdater {
...
+                } { _ =>
+                  logger.debug(s"Found valid majority hash=${majorityHash.show} with occurrences=${occurrences}")
```
tbekas, line 374: Let's not add this log. Across all the codebase we log only anomalies (invalid things or errors), and we implicitly assume that the other 99,99% of things are valid.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/consensus/ConsensusFunctions.scala
```diff
@@ -10,6 +13,10 @@ trait ConsensusFunctions[F[_], Event, Key, Artifact] {
 
   def facilitatorFilter(lastSignedArtifact: Signed[Artifact], peerId: PeerId): F[Boolean]
 
+  def isValidProposal(lastSignedArtifact: Signed[Artifact], trigger: ConsensusTrigger)(
+    proposal: Artifact
+  ): F[Either[InvalidProposal, Artifact]]
```
tbekas, line 18: 👍

tbekas (changes requested):

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
tbekas: Ok

modules/sdk/src/main/scala/org/tessellation/sdk/domain/consensus/ConsensusFunctions.scala
```diff
@@ -10,6 +13,10 @@ trait ConsensusFunctions[F[_], Event, Key, Artifact] {
 
   def facilitatorFilter(lastSignedArtifact: Signed[Artifact], peerId: PeerId): F[Boolean]
 
+  def isValidProposal(lastSignedArtifact: Signed[Artifact], trigger: ConsensusTrigger)(
```
TheMMaciek, line 16: Maybe `isValidArtifact` if we are working with the `Artifact` abstraction?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusState.scala
```diff
@@ -28,17 +28,20 @@ case class ConsensusState[Key, Artifact](
...
 final case class CollectingSignatures[A](majorityArtifactHash: Hash, majorityTrigger: ConsensusTrigger) extends ConsensusStatus[A]
 final case class Finished[A](signedMajorityArtifact: Signed[A], majorityTrigger: ConsensusTrigger) extends ConsensusStatus[A]
 
 object ConsensusStatus {
   implicit def showInstance[A]: Show[ConsensusStatus[A]] = {
     case CollectingFacilities(maybeFacilityInfo) =>
       s"CollectingFacilities{maybeFacilityInfo=${maybeFacilityInfo.show}}"
-    case CollectingProposals(majorityTrigger, maybeProposalInfo) =>
-      s"CollectingProposals{majorityTrigger=${majorityTrigger.show}, maybeProposalInfo=${maybeProposalInfo.show}}"
+    case CollectingProposals(majorityTrigger, maybeProposalInfo, maybeLastArtifact) =>
```
TheMMaciek, line 43: maybeLastArtifact is not used, you can use wildcard notation

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
marcinwadon: That's why I used `debug` flag that we use for all debugging things (not a warn or error). I can remove it but I wanted to be able to see occurrences when debugging for misalignments.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -303,6 +327,8 @@ object GlobalSnapshotConsensusFunctions {
       (result, Set.empty)
     }
 
+    case object ProposalMismatch extends InvalidProposal
```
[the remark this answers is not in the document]
marcinwadon: `InvalidProposal` is in the domain while `ProposalMismatch` is a particular implementation in `infrastructure`, next to `GlobalSnapshotConsensusFunctions` which also is an implementation of generic `ConsensusFunctions`

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
marcinwadon: How this two cases should be handled then?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -265,35 +267,43 @@ object ConsensusStateUpdater {
...
-                } >> Metrics[F].recordDistribution(
-                  "dag_consensus_proposal_affinity",
-                  proposalAffinity(allProposals, proposalInfo.proposalArtifactHash)
-                )
-              }.void
-              (newState, effect).pure[F]
+            maybeLastArtifact.map { lastSignedArtifact =>
+              pickValidatedMajority(lastSignedArtifact, majorityTrigger, resources)(allProposals)
             }
+              .getOrElse(pickMajority(allProposals).pure[F])
```
TheMMaciek, line 290: Is this the case for the first snapshot when in observing state?

marcinwadon: currently both cases (missing artifact or no artifact) are handled in the same way. @tbekas suggested to distinguish them but I don't know how they should be handled then

TheMMaciek: If yes, maybe we should have to cases. One when last artifact exists and second when it doesn't. Then use `pickValidatedMajority` when `Some`, and `pickMajority` when `None`?

marcinwadon: Ok, so one case is when we do not store the artifact (observing state), then we can use `pickMajority`. What to do with the second case then? When the artifact should be store but we don't have it (because of async)

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
tbekas: It doesn't matter which one we choose until we choose the same on all nodes.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
tbekas: Ok, but it doesn't add too much information. We already log the hash of the majority hash during the state transition.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
marcinwadon: Answered on slack, thanks :+1:

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
marcinwadon: fixed :+1:

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -338,6 +348,31 @@ object ConsensusStateUpdater {
...
+          .flatMap {
+            case None => none[Hash].pure[F]
+            case Some(Left(cause)) =>
+              logger.warn(cause)(s"Found invalid majority hash=${majorityHash._2.show} with occurrences=${majorityHash._1}") >>
+                tail.headOption.fold(none[Hash].pure[F])(go(_, tail.tail))
+            case Some(Right(_)) => majorityHash._2.some.pure[F]
+          }
+
+      sortedProposals.headOption
+        .fold(none[Hash].pure[F])(go(_, sortedProposals.tail))
```
tbekas, line 373: This line and 368 look exactly the same. You could get rid of this small duplication (along with `._1` and `._2`)  If you would redefine `go` to:

```scala
def go(proposals: List[(Int, Hash)]): F[Option[Hash]] =
  proposals match {
    case (occurrences, majorityHash) :: tail =>
      resources.artifacts
        .get(majorityHash)
        .traverse { artifact =>
          consensusFns.validateArtifact(lastSignedArtifact, trigger)(artifact)
        }
        .flatMap { maybeValidatedArtifact =>
          maybeValidatedArtifact.flatTraverse { artifactOrError =>
            artifactOrError.fold(
              cause =>
                logger.warn(cause)(s"Found invalid majority hash=${majorityHash.show} with occurrences=$occurrences") >> go(tail),
              _ => majorityHash.some.pure[F]
            )
          }
        }
    case Nil => none[Hash].pure[F]
  }
```

tbekas (commented):
Overall looks good. Small comment regarding the `go` function.

tbekas (approved):
LGTM

modules/sdk/src/main/scala/org/tessellation/sdk/domain/consensus/ConsensusFunctions.scala
```diff
@@ -10,6 +13,10 @@ trait ConsensusFunctions[F[_], Event, Key, Artifact] {
 
   def facilitatorFilter(lastSignedArtifact: Signed[Artifact], peerId: PeerId): F[Boolean]
 
+  def validateArtifact(lastSignedArtifact: Signed[Artifact], trigger: ConsensusTrigger)(
+    proposal: Artifact
```
TheMMaciek, line 17: I guess `artifact` not `proposal`

marcinwadon: @TheMMaciek right, fixed :+1:

PR #500, opened 2022-11-04
Custom encoder/decoder for snapshot ordinal

kriscieslak:


modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/SnapshotOrdinal.scala
```diff
@@ -32,4 +30,10 @@ object SnapshotOrdinal {
     def partialPrevious(a: SnapshotOrdinal): Option[SnapshotOrdinal] =
       refineV[NonNegative].apply[Long](a.value.value |+| -1).toOption.map(r => SnapshotOrdinal(r))
   }
+
+  implicit val snapshotOrdinalDecoder: Decoder[SnapshotOrdinal] =
+    Decoder[Long].emap(v => NonNegLong.from(v).map(SnapshotOrdinal(_)))
+
+  implicit val snapshotOrdinalEncoder: Encoder[SnapshotOrdinal] =
```
tbekas, line 37:
```
Encoder[NonNegLong].contramap(_.value)
```

modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/SnapshotOrdinal.scala
```diff
@@ -4,18 +4,17 @@ import cats.Order
...
 import eu.timepit.refined.cats._
 import eu.timepit.refined.numeric.NonNegative
 import eu.timepit.refined.refineV
 import eu.timepit.refined.types.numeric.NonNegLong
+import io.circe.refined._
+import io.circe.{Decoder, Encoder}
 
-@derive(encoder, decoder, order, show)
+@derive(order, show)
 case class SnapshotOrdinal(value: NonNegLong)
```
marcinwadon, line 18: btw, why we don't have `newtype` here? 🤔 
@kpudlik @TheMMaciek @tbekas maybe you remember the reason?

marcinwadon: tl;dr; - it was required for Consensus (ClassTag / AnyRef). Probably it can be a newtype now, but it requires a schema migration

marcinwadon (approved):

PR #501, opened 2022-11-08
Properly decode snapshot ordinal in L1

kriscieslak:


marcinwadon (approved):
Looks good, but why decoder is imported each time? Can't the default one from SnapshotOrdinal be used?

kriscieslak:
I have defined new encoder/decoder in previous ticket and now SnapshotOrdinal is encoded as value not object. 
But in `global-snapshots/latest/ordinal` endpoint we wanted to return SnapshotOrdinal in old way (`{ value: ..}`) and I use "old" magnolia decoders to decode it.

PR #502, opened 2022-11-09
Verify facilitators hash between steps of consensus

kriscieslak:


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -245,95 +248,120 @@ object ConsensusStateUpdater {
...
-                  .warn(
-                    s"Removed ${(allSignatures.size - validSignatures.size).show} invalid signatures during consensus for key ${key.show}, " +
-                      s"${validSignatures.size.show} valid signatures left"
-                  )
-                  .whenA(allSignatures.size =!= validSignatures.size)
+                .flatMap(peerDeclaration => peerDeclaration.signature.map(signature => (peerId, signature)))
+            }
+
+          maybeAllSignatures
+            .traverseTap(checkFacilitatorsHashes(_._2.facilitatorsHash, ownFacilitatorsHash.some))
```
tbekas, line 337: Please use pattern match instead of `_._2`

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -345,6 +373,20 @@ object ConsensusStateUpdater {
       }
     }
 
+    private def checkFacilitatorsHashes[A](facilitatorsHash: A => Hash, ownFacilitatorsHash: Option[Hash])(
```
tbekas, line 376: 1. Please rename this function to something like `warnIfForking`
1. Argument name `facilitatorHash` doesn't fit its type, it's not a `Hash` but a mapping function. Also it would be better if you would simply add `def facilitatorHash: Hash` to `PeerDeclaration` while a adding here a constraint on `A` to be `A <: PeerDeclaration`, then you wouldn't need this argument at all.
1. `ownFacilitatorHash` shouldn't be `Option`
1. The result of this function should be only `F[Unit]` as it logs stuff at most

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -345,6 +373,20 @@ object ConsensusStateUpdater {
...
+    ): F[Option[List[A]]] =
+      pickMajority(declarations.map(facilitatorsHash)).filter { majorityFacilitatorsHash =>
+        ownFacilitatorsHash.map(_ === majorityFacilitatorsHash).getOrElse(true)
+      }
+        .map(_ => declarations)
+        .pure[F]
+        .flatTap(declarations =>
+          logger
+            .warn(s"Different facilitators hashes. This node is in fork")
+            .whenA(declarations.isEmpty)
```
tbekas, line 387: 99% of people reading this code will think that `declarations.isEmpty` checks if the list of declarations is empty

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -345,6 +373,20 @@ object ConsensusStateUpdater {
       }
     }
 
+    private def checkFacilitatorsHashes[A](facilitatorsHash: A => Hash, ownFacilitatorsHash: Option[Hash])(
+      declarations: List[A]
+    ): F[Option[List[A]]] =
+      pickMajority(declarations.map(facilitatorsHash)).filter { majorityFacilitatorsHash =>
```
tbekas, line 379: Please refactor this function to:

```scala
pickMajority(declarations.map(facilitatorsHash)).traverse { majorityHash =>
  logger
    .warn(s"Different facilitators hashes. This node is in fork")
    .whenA(majorityHash =!= ownHash)
}.void
```

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/consensus/FacilitatorCalculatorSuite.scala
```diff
@@ -62,4 +47,11 @@ object FacilitatorCalculatorSuite extends FunSuite with Checkers {
 
     expect.same(expected, result)
   }
+
+  private def facilityDeclaration(facilitators: Set[PeerId]) =
+    PeerDeclarations(
+      facility = Facility(Map.empty, facilitators, None, Hash("")).some,
```
tbekas, line 53: Please use valid hashes only, so i.e. `Hash.empty`

tbekas (changes requested):

tbekas (approved):
LGTM

PR #503, opened 2022-11-16
Consensus stall resolution

tbekas:
* Removal of PeerDeclarationHealthCheck
* Embedding stall detection and stall fix within the Consensus


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusState.scala
```diff
@@ -61,3 +88,10 @@ object ProposalInfo {
   implicit def showInstance[A]: Show[ProposalInfo[A]] = pi =>
     s"ProposalInfo{proposalArtifact=***, proposalArtifactHash=${pi.proposalArtifactHash.show}}"
 }
+
+@derive(eqv, show)
+sealed trait LockStatus
+
+case object Open extends LockStatus
+case object Locked extends LockStatus
+case object Unlocked extends LockStatus
```
kpudlik, line 97: As we discussed there is a fixed path of state transitions so for example we can't go from `Locked` to `Open`. To make it more bulletproof we could prevent wrong state transitions by using `IndexedStateT ` and secure that on type-level:

```scala
def lock: IndexedStateT[Eval, Open.type, Locked.type, Unit] = IndexedStateT.set(Locked)
def unlock: IndexedStateT[Eval, Locked.type, Unlocked.type, Unit] = IndexedStateT.set(Closed)
```

then if you do 

```
val check = for {
  _ <- lock
  _ <- unlock
  _ <- lock
} yield ()
```

then it gives you a type error.

Moreover if you do `valid.run(Open)` then it is fine because `_ <- lock` can be executed on `Open` but `valid.run(Unlocked)` also gives you a type error.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusResources.scala
```diff
@@ -1,14 +1,18 @@
...
 import org.tessellation.security.hash.Hash
 
+/** Represents various data received from other peers
+  */
 case class ConsensusResources[A](
   peerDeclarationsMap: Map[PeerId, PeerDeclarations],
-  artifacts: Map[Hash, A],
-  removedFacilitators: Set[PeerId]
+  acksMap: Map[(PeerId, PeerDeclarationKind), Set[PeerId]],
+  ackKinds: Set[PeerDeclarationKind],
```
kpudlik, line 12: isn't `ackKinds` just a `Set` of all `PeerDeclarationKind` from `acksMap`? If so then we are duplicating data here so maybe some flattening function will be better 🤔

modules/core/src/main/scala/org/tessellation/cli/snapshot.scala
```diff
@@ -15,5 +16,5 @@ object snapshot {
     .env[Path]("CL_SNAPSHOT_STORED_PATH", help = "Path to store created snapshot")
     .withDefault(Path("data/snapshot"))
 
-  val opts = globalSnapshotPath.map(SnapshotConfig(2L, _, 48.seconds, 10L))
+  val opts = globalSnapshotPath.map(SnapshotConfig(ConsensusConfig(48.seconds, 50.seconds, 10.seconds), _, 10L))
```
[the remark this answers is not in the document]
tbekas: Nope, these values are independent. `lockDelay` is the amount of time after which we lock the consensus if the state didn't transitioned

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusState.scala
```diff
@@ -2,28 +2,55 @@ package org.tessellation.sdk.infrastructure.consensus
...
+      case _: CollectingFacilities[A] => (Facility.some, Set.empty)
+      case _: CollectingProposals[A]  => (Proposal.some, Set(Facility))
+      case _: CollectingSignatures[A] => (MajoritySignature.some, Set(Facility, Proposal))
+      case _: Finished[A]             => (none, Set(Facility, Proposal, MajoritySignature))
+    }
+
+    def collectedKinds: Set[PeerDeclarationKind] = kindRelation._2
+    def maybeCollectingKind: Option[PeerDeclarationKind] = kindRelation._1
+    def locked: Boolean = value.lockStatus === Locked
+    def notLocked: Boolean = !locked
```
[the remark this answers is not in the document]
tbekas: Yep, exactly. It's very easy to miss the `!` when reading the code

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusResources.scala
tbekas: It is, but it's an optimization. Duplication (denormalization) of information is a common technique to optimize access to data.

There's almost no additional cost for writing to this Set, but there's a significant benefit on reading, and reading is as frequent as writing is (hundreds or even thousands of times during a consensus).

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -15,14 +15,15 @@ import cats.syntax.show._
 import cats.syntax.traverse._
 import cats.{Applicative, Order, Show}
 
-import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.duration.{FiniteDuration, _}
```
TheMMaciek, line 18: Either `_` or concrete imports.

tbekas: Fixing it

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusState.scala
tbekas: Ok, I tried to include this, but it would require a bigger refactoring, since we would have to include the type of the lock in the type parameters of the whole state `ConsensusState[Key, Artifact, LockStatus]`. This in turn would require changes to all methods that operate on a consensus state (not only method signature would change, but implementation would have to be adjusted in a couple of cases as well).

kpudlik (dismissed):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -159,19 +157,28 @@ object ConsensusManager {
...
+                  deregisterPeers(key, state.removedFacilitators) >>
                   consensusStorage
                     .tryUpdateLastKeyAndArtifactWithCleanup(state.lastKey, key, signedArtifact)
                     .ifM(
                       afterConsensusFinish(majorityTrigger),
                       logger.info("Skip triggering another consensus")
                     ) >> nodeStorage.tryModifyStateGetResult(Observing, Ready).void
               case _ =>
-                internalCheckForStateUpdate(key, resources)
+                stallDetection(key, state) >>
```
TheMMaciek, line 171: Improvement for the future already discussed offline - I propose running `stallDetection` only once for every stage as only the first run is useful the rest is bound to fail. So I guess we want the `stallDetection` background task to start in right after status switch (starting next stage).

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusState.scala
```diff
@@ -2,28 +2,55 @@ package org.tessellation.sdk.infrastructure.consensus
...
+      case _: CollectingFacilities[A] => (Facility.some, Set.empty)
+      case _: CollectingProposals[A]  => (Proposal.some, Set(Facility))
+      case _: CollectingSignatures[A] => (MajoritySignature.some, Set(Facility, Proposal))
+      case _: Finished[A]             => (none, Set(Facility, Proposal, MajoritySignature))
+    }
+
+    def collectedKinds: Set[PeerDeclarationKind] = kindRelation._2
+    def maybeCollectingKind: Option[PeerDeclarationKind] = kindRelation._1
+    def locked: Boolean = value.lockStatus === Closed
+    def notLocked: Boolean = !locked
```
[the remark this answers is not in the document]
TheMMaciek: Actually it does include that case automatically.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
tbekas: Agree, I think it's an important improvement because we want to keep the "flow" clean and under control. It will require returning `Ior` instead of `Maybe` from the `ConsensusStateUpdater`. I will create a ticket for this and fix it as a part of that ticket if you don't mind.

TheMMaciek (dismissed):
LGTM

TheMMaciek (approved):

kriscieslak (approved):

PR #505, opened 2022-11-24
Base project setup for Currency L0

kpudlik:


build.sbt
```diff
@@ -453,4 +453,70 @@ lazy val core = (project in file("modules/core"))
...
+      Libraries.cirisCore,
+      Libraries.cirisEnum,
+      Libraries.cirisRefined,
+      Libraries.derevoCore,
+      Libraries.derevoCats,
+      Libraries.derevoCirce,
+      Libraries.doobieCore,
+      Libraries.doobieHikari,
+      Libraries.doobieH2,
+      Libraries.doobieQuill,
```
marcinwadon, line 489: are all dependencies needed? I'd recommend starting with minimum and extend when required

modules/currency-l0/src/main/scala/org/tessellation/cli/snapshot.scala
```diff
@@ -0,0 +1,17 @@
...
+
+import com.monovore.decline._
+import eu.timepit.refined.auto._
+import fs2.io.file.Path
+
+object snapshot {
+
+  val snapshotPath: Opts[Path] = Opts
+    .env[Path]("CL_SNAPSHOT_STORED_PATH", help = "Path to store created snapshot")
+    .withDefault(Path("data/snapshot")) // TODO: Should path be prefixed to prevent overriding of another SC?
```
marcinwadon, line 14: imo each jar file should be executed in separate working directory. I think it should be responsibility of the runner to set working directory or correct paths 🤔

build.sbt
kpudlik: Not really, I just wanted to speed up the initial setup. I will clean it up.

modules/currency-l0/src/main/scala/org/tessellation/cli/http.scala
```diff
@@ -0,0 +1,51 @@
...
+
+import scala.concurrent.duration._
+
+import org.tessellation.ext.decline.decline._
+import org.tessellation.sdk.config.types.{HttpClientConfig, HttpConfig, HttpServerConfig}
+
+import com.comcast.ip4s.{Host, Port, _}
+import com.monovore.decline.Opts
+
+object http {
```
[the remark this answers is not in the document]
kpudlik: Sure. We can remove defaults from there and in DAG just do this:


`(externalIpOpts.withDefault(), publicHttpPortOpts.withDefault())` and for others just force passing via CLI

modules/currency-l0/src/main/scala/org/tessellation/cli/snapshot.scala
kpudlik: I totally agree with that

modules/currency-l0/src/main/scala/org/tessellation/CurrencyL0App.scala
```diff
@@ -0,0 +1,30 @@
...
+import org.tessellation.sdk.app.{SDK, TessellationIOApp}
+
+import com.monovore.decline.Opts
+
+final class CurrencyL0App(
+  header: String,
+  clusterId: ClusterId, // TODO: It should be token symbol probably but on the other hand it is hard to make it unique
+  helpFlag: Boolean = true,
+  version: String = "" // TODO: Should versioning be unique per currency?
+) extends TessellationIOApp[Run](clusterId.toString, header, clusterId, helpFlag, version) {
```
marcinwadon, line 17: imo we can create one more trait - `CurrencyApp`, and embed all the things there (inside). in the `CurrencyApp` we will specify just the details (like token name/symbol)

modules/sdk/src/main/scala/org/tessellation/sdk/cli/http.scala
```diff
@@ -0,0 +1,33 @@
...
+
+import org.tessellation.ext.decline.decline._
+import org.tessellation.sdk.config.types.HttpClientConfig
+
+import com.comcast.ip4s.{Host, Port}
+import com.monovore.decline.Opts
+
+object http {
+  val externalIpOpts: Opts[Host] = Opts
+    .option[Host]("ip", help = "External IP (a.b.c.d.)")
```
marcinwadon, line 13: typo `a.b.c.d.` -> `a.b.c.d`

marcinwadon (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/cli/http.scala
kpudlik: good catch, it has been there since the very beginning

marcinwadon (approved):

PR #506, opened 2022-11-30
Store signature of SCSnapshot in GlobalSnapshot

marcinwadon:
**IMPORTANT!**
This change is a breaking change for testnet and will require full restart (from genesis) on testnet.
**However, it is not a breaking change for mainnet.**

### DIFF
--- a/modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
+++ b/modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
@@ -274,7 +274,7 @@ object GlobalSnapshotConsensusFunctions {
     private def processStateChannelEvents(
       lastGlobalSnapshotInfo: GlobalSnapshotInfo,
       events: List[StateChannelEvent]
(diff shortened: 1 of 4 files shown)


modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -315,12 +315,7 @@ object GlobalSnapshotConsensusFunctions {
                 }
                 .getOrElse(Eval.now(List.empty))
 
-            unfold(initLsh).value.toNel.map(nel =>
-              address -> nel.map { event =>
-                StateChannelSnapshotBinary(event.snapshot.value.lastSnapshotHash, event.snapshot.content)
-
-              }.reverse
-            )
+            unfold(initLsh).value.toNel.map(address -> _.map(_.snapshot).reverse)
```
tbekas, line 318: 👍

tbekas (approved):
LGTM

tbekas:
Please also adjust the logic of snapshot validation, since now the state channel snapshots can be validated as well.

marcinwadon:
> Please also adjust the logic of snapshot validation, since now the state channel snapshots can be validated as well.

Afaik this is the ticket @kriscieslak is working on right now.

PR #515, opened 2022-12-09
Fetch majority global snapshot in L0Service

marcinwadon:


modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/services/L0Service.scala
```diff
@@ -36,56 +40,87 @@ object L0Service {
...
-              val nextOrdinal = lastStoredOrdinal.map(_.next).getOrElse(lastOrdinal)
-              val lastOrdinalCap = lastOrdinal.value.value
-                .min(singlePullLimit.map(nextOrdinal.value.value + _.value).getOrElse(lastOrdinal.value.value))
+      def fetchGlobalSnapshot(ordinal: SnapshotOrdinal)(peer: L0Peer): F[Option[Hashed[GlobalSnapshot]]] =
+        l0GlobalSnapshotClient
+          .get(ordinal)(peer)
+          .flatMap(_.toHashedWithSignatureCheck)
+          .flatMap(_.liftTo[F])
+          .map(_.some)
+          .handleError(_ => none[Hashed[GlobalSnapshot]])
```
TheMMaciek, line 69: Don't we want to have it logged in case it fails?

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/services/L0Service.scala
```diff
@@ -36,56 +40,87 @@ object L0Service {
...
+          }
+        }.handleErrorWith { e =>
+          logger.warn(e)(s"Failure pulling global snapshots!") >>
+            Applicative[F].pure(List.empty[Hashed[GlobalSnapshot]])
+        }
+
+      def pullMajorityGlobalSnapshots: F[List[Hashed[GlobalSnapshot]]] =
+        l0ClusterStorage.getPeers
+          .map(_.toNonEmptyList.toList)
+          .flatMap { peers =>
```
TheMMaciek, line 95: Why not keep the `Nel` here? 🤔

TheMMaciek (changes requested):
First of all I would rethink the race condition during majority pull.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/services/L0Service.scala
marcinwadon: actually I can add logger on some global level, but imo it's not worthing to add for every single fetch 🤔 wdyt @TheMMaciek ?

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/services/L0Service.scala
marcinwadon: I'll change that, it was from before refactoring

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/services/L0Service.scala
TheMMaciek: Well the way I was using that function is that first the node was asked what ordinals it has and after that these ordinals were fetched, so a failure here was unexpected hence logging. But If you have some idea on how to log it differently without loosing context then I would go for it.

marcinwadon: I've added logs :+1:

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/services/L0Service.scala
```diff
@@ -36,56 +40,87 @@ object L0Service {
...
+          .map(_.toList.map(o => SnapshotOrdinal(NonNegLong.unsafeFrom(o))))
 
-      def pullGlobalSnapshots: F[List[Hashed[GlobalSnapshot]]] = {
-        for {
-          lastStoredOrdinal <- lastSnapshotStorage.getOrdinal
-          l0Peer <- l0ClusterStorage.getRandomPeer
+      def fetchMajorityGlobalSnapshot(ordinal: SnapshotOrdinal, peers: List[L0Peer]): F[Option[Hashed[GlobalSnapshot]]] =
+        peers
+          .traverse(fetchGlobalSnapshot(ordinal))
+          .map(_.flatten)
```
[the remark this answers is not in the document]
marcinwadon: Let's talk about that on slack. I've added TODO in the code for now

marcinwadon:
We're not going to merge this pull request. Closing

PR #519, opened 2022-12-14
StateChannelSnapshotBinary Order implementation to fix serialization issue

marcinwadon:


kpudlik (approved):

modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/StateChannelSnapshotBinary.scala
```diff
@@ -14,10 +15,12 @@ case class StateChannelSnapshotBinary(
   lastSnapshotHash: Hash,
   content: Array[Byte]
 )
-
 object StateChannelSnapshotBinary {
   implicit val show: Show[StateChannelSnapshotBinary] = s => s"StateChannelSnapshotBinary(lastSnapshotHash=${s.lastSnapshotHash.show})"
 
+  implicit val arrayByteOrder: Order[Array[Byte]] = Order.by(_.toSeq)
```
tbekas, line 21: Please create a generic order for `Array[A]` and add it to the `OrhanInstances`

modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/StateChannelSnapshotBinary.scala
```diff
@@ -14,10 +15,12 @@ case class StateChannelSnapshotBinary(
...
+  implicit val order: Order[StateChannelSnapshotBinary] = Order.whenEqual(Order.by(_.lastSnapshotHash), Order.by(_.content))
```
tbekas, line 22: I think that this can be derived atuomatically by derevo once we have an order for `Array`

tbekas (changes requested):
Nice that you've fixed the issue! 👏

modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/StateChannelSnapshotBinary.scala
marcinwadon: :+1:

modules/dag-shared/src/main/scala/org/tessellation/dag/snapshot/StateChannelSnapshotBinary.scala
marcinwadon: yup, that's true

tbekas (approved):
LGTM

PR #524, opened 2022-12-23
Schema extracted to separate module

marcinwadon:


build.sbt
```diff
@@ -306,7 +327,7 @@ lazy val dagShared = (project in file("modules/dag-shared"))
 lazy val dagL1 = (project in file("modules/dag-l1"))
   .enablePlugins(AshScriptPlugin)
   .enablePlugins(JavaAppPackaging)
-  .dependsOn(kernel, shared % "compile->compile;test->test", dagShared, sdk, testShared % Test)
+  .dependsOn(dagShared % "compile->compile;test->test", sdk, kernel, testShared % Test)
```
TheMMaciek, line 330: What about transitive dependencies? For example I'm almost sure that dagL1 depends on schema as well.

modules/core/src/main/scala/org/tessellation/http/routes/DagRoutes.scala
```diff
@@ -7,8 +7,9 @@ import cats.syntax.functor._
 import org.tessellation.dag.domain.block.DAGBlock
 import org.tessellation.domain.cell.{L0Cell, L0CellInput}
 import org.tessellation.domain.dag.DAGService
-import org.tessellation.ext.http4s.{AddressVar, SnapshotOrdinalVar}
-import org.tessellation.security.signature.Signed
+import org.tessellation.ext.http4s.SnapshotOrdinalVar
```
TheMMaciek, line 10: Shouldn't `SnapshotOrdinalVar` also go to `schema` module?

modules/core/src/main/scala/org/tessellation/package.scala
```diff
@@ -6,18 +6,18 @@ import cats.effect.Async
...
 
 import eu.timepit.refined.auto._
-import eu.timepit.refined.numeric.Interval
+import eu.timepit.refined.numeric.Greater
 import io.estatico.newtype.ops._
 
 package object tessellation {
 
-  type CoreKryoRegistrationIdRange = Interval.Closed[700, 799]
+  type CoreKryoRegistrationIdRange = Greater[100]
```
TheMMaciek, line 20: We don't want any constraint for the max registration number?

marcinwadon: we already discussed that and this is what @kpudlik did in his pull request.
if we're going to rediscuss it, let's create a separate task and handle/change all kryo id ranges

modules/core/src/main/scala/org/tessellation/http/routes/DagRoutes.scala
marcinwadon: probably yes, I'll check

build.sbt
marcinwadon: in the current form, you were not able to handle all the dependencies and we used partially explicitly declared dependencies and partially implicitly (transitive). it's better to stay with one solution, so it would be way cleaner, to specify minimum set of libraries explicitly in the base package, and whatever is needed more on higher packages, should be declared separately. we cannot avoid transitive dependencies anyway, so let's use it instead of redeclaring them (and then noone is keeping an order there)

modules/kernel/src/main/scala/org/tessellation/kernel/package.scala
```diff
@@ -13,7 +13,6 @@ package object kernel {
   type KernelKryoRegistrationId = KryoRegistrationId[KernelKryoRegistrationIdRange]
 
   val kernelKryoRegistrar: Map[Class[_], KernelKryoRegistrationId] = Map(
-    classOf[ProcessSnapshot] -> 400,
-    classOf[StateChannelSnapshotGist] -> 401
```
TheMMaciek, line 17: Shouldn't we somehow mark that the 401 was already used, to avoid the situation where we reuse that registration id for another class? Nothing probably will happen in this particular case, but I wonder if it shouldn't be the approach in general to register new classes under new (never used) ids only.

modules/schema/src/main/scala/org/tessellation/schema/transaction.scala
```diff
@@ -67,7 +67,7 @@ object transaction {
     val _Ordinal: Lens[TransactionReference, TransactionOrdinal] = GenLens[TransactionReference](_.ordinal)
 
     def of[F[_]: Async: KryoSerializer](signedTransaction: Signed[Transaction]): F[TransactionReference] =
-      signedTransaction.value.hashF.map(TransactionReference(signedTransaction.ordinal, _))
+      signedTransaction.value.hashF.map(a => TransactionReference(signedTransaction.ordinal, a))
```
TheMMaciek, line 70: If `_` doesn't work anymore maybe call the value `h` or `hash` instead of `a`

modules/core/src/main/scala/org/tessellation/package.scala
tbekas: Why `Greater[100]`? It will likely conflict with twitter chill registrations.

modules/schema/src/main/scala/org/tessellation/schema/transaction.scala
marcinwadon: ah, good catch. it's a leftover

marcinwadon: fixed :+1:

modules/core/src/main/scala/org/tessellation/http/routes/DagRoutes.scala
marcinwadon: done :+1:

modules/shared/src/main/scala/org/tessellation/ext/kryo.scala
```diff
@@ -38,6 +39,14 @@ object kryo {
 
   type KryoRegistrationId[C] = Int Refined C
 
+  implicit class MapRegistrationGreaterId[A](mapa: Map[Class[_], KryoRegistrationId[Greater[A]]]) {
+
+    def union[B](mapb: Map[Class[_], KryoRegistrationId[Greater[A]]]) =
```
TheMMaciek, line 44: The `B` here is not used

marcinwadon: fixed :+1:

modules/core/src/main/scala/org/tessellation/package.scala
marcinwadon: after discussion, we've decided to keep it `Greater[300]`, because twitter/chill registers more than 100 classes

modules/kernel/src/main/scala/org/tessellation/kernel/package.scala
kpudlik: I think we should distinguish `never used` from `never serialized`. If it hasn't been serialized with these ids before then we should be fine. If you want to mark anyway then we can do some refined rule like `Not[400]` or sth.

build.sbt
TheMMaciek: I just remember that there was a discussion and we wanted to avoid transitive dependencies hence the comment. Given that external libraries pull in their versions of dependencies with them there is a chance that by removing some explicitly declared library from the base module, the version of this library in the depending module may change (because it's neither declared explicitly in the base module anymore but still can be available through base module dependencies and was never declared explicitly in the depending module). If we drop by a major version I guess it can mean trouble. Also in such a situation we are losing the compiler warning about possible incompatibilities afaik. Anyway I'm not sure what's the best approach here, just recalling the discussion we had some time ago and giving my 2 cents.

modules/kernel/src/main/scala/org/tessellation/kernel/package.scala
TheMMaciek: Just pointing to a possible problem. It may also not be that easy at some point to determine if a class was ever serialized or not.

build.sbt
kpudlik: We can run `sbt dependencyBrowseGraph` or `sbt dependencyTree` to find transitive dependencies eventually.

marcinwadon: @TheMMaciek I didn't remove dependencies from base modules (shared, schema, etc). I just removed redeclaration (probably not all, I'd keep it as a continuation process), so:
- if shared requires kryo, then schema don't need to redeclare kryo and can use it, because depends on shared.
- if kernel requires droste, but not shared, then kernel should declare droste dependency
- if shared depend on external library (let's say http4s) that is dependant on fs2, we should anyway declare fs2 explicitly.

so tl;dr - no transitive dependencies from external libraries. only between our internal modules

modules/shared/src/main/scala/org/tessellation/ext/kryo.scala
```diff
@@ -38,6 +39,14 @@ object kryo {
 
   type KryoRegistrationId[C] = Int Refined C
 
+  implicit class MapRegistrationGreaterEqualId[A](mapa: Map[Class[_], KryoRegistrationId[GreaterEqual[A]]]) {
```
tbekas, line 42: Why is it necessary? It seems like it's already covered by the existing `union`

build.sbt
```diff
@@ -89,155 +74,191 @@ lazy val wallet = (project in file("modules/wallet"))
...
+      Libraries.http4sCirce,
+      Libraries.http4sJwtAuth,
+      Libraries.weaverCats,
+      Libraries.weaverDiscipline,
+      Libraries.weaverScalaCheck
     )
   )
 
-lazy val shared = (project in file("modules/shared"))
+lazy val schema = (project in file("modules/schema"))
```
tbekas, line 150: We should have an ADR describing what's the purpose of each module so that we don't mix them up when adding stuff (like we do currently).

TheMMaciek (approved):

kpudlik (approved):

tbekas (changes requested):
What is the purpose/responsibility of the `schema` module? It's impossible to say if the correct classes were extracted without the context.

modules/shared/src/main/scala/org/tessellation/ext/kryo.scala
marcinwadon: otherwise, it throws an error that something cannot be `Greater[300] Or Greater[300]`

marcinwadon:
Closing the PR. I'll create new ones according to the ADR 12. (#526)

PR #525, opened 2022-12-27
Validate kryo registrations

tbekas:


marcinwadon (approved):

kpudlik (approved):

modules/shared/src/main/scala/org/tessellation/shared/package.scala
```diff
@@ -28,7 +27,6 @@ package object shared {
   type SharedKryoRegistrationId = KryoRegistrationId[SharedKryoRegistrationIdRange]
 
   val sharedKryoRegistrar: Map[Class[_], SharedKryoRegistrationId] = Map(
-    classOf[UUID] -> 300,
```
tbekas, line 31: I checked usages and we're not serializing it to disk, nor do we use it for hash calculation of stuff that we serialize to disk. But please double-check it if you can.

marcinwadon: we're using it with kryo indeed, but it's not a part of snapshot

TheMMaciek (approved):

PR #526, opened 2022-12-28
ADR - Modules' structure

marcinwadon:


doc/adr/0012-modules-structure.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+The existing structure of tessellation modules was mixed, hence hard to maintain.
+
+- `shared` contains both base classes and extensions, and tessellation schema,
```
TheMMaciek, line 13: "contains base classes, extensions and tessellation schema" or 
"contains both base classes with extensions and tessellation schema"
not sure which one you meant

doc/adr/0012-modules-structure.md
```diff
@@ -0,0 +1,105 @@
...
+- `keytool` contains key-pair generator and exposes key-related operations,
+- `wallet` contains operations related to transactions,
```
TheMMaciek, line 15: "related to transactions" or "related to creating transactions"? 🤔

doc/adr/0012-modules-structure.md
```diff
@@ -0,0 +1,105 @@
...
+- `kernel` contains part of state channels (hylomorphism's idea),
+- `sdk` contains all the things shared between Global L0 and DAG L1, but not the $DAG itself,
+- `dag-shared` contains all the things shared between Global L0 and DAG L1 related to $DAG,
+- `core` is the Global L0 module (but also contains state channels that must be extracted to be reused in other modules),
+- `dag-l1` is the DAG L1 module.
```
TheMMaciek, line 20: maybe: "contains implementation of DAG token L1 layer"

doc/adr/0012-modules-structure.md
```diff
@@ -0,0 +1,105 @@
...
+    E --> F
+    F --> C
+    F -.-> B
+    F -.-> A
+    G --> E
+    G -.-> C
+    G -.-> B
+    G -.-> A
+    G --> F
+    H --> F
```
TheMMaciek, line 49: If I understand correctly that dotted line is a transitive dependency. This can be marked as a dotted line as the SDK also requires DAG shared and DAG L1 depends on SDK. Right?

doc/adr/0012-modules-structure.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Decision
+
+To simplify the structure of modules, we will split `shared` into two separate modules - the first containing scala and libraries extensions, the latter containing the tessellation schema.
+Then, all the things related to node operations would be kept in `node-shared` module (including gossip protocol, consensus, etc.). This leads to emptying the SDK module, so we will be able to implement the SDK module as a proxy between the external world and the internal tessellation structure. SDK module should reexpose all necessary operations to be used by 3rd parties. For example - to create a keypair, global L0 can use the keytool module directly, but 3rd party will use the SDK module that will reexpose keypair creation function. That way, the internal structure of tessellation will be hidden behind the facade.
+
+Modules will be transient - for example: `wallet` does not need to depend on `schema` directly, because it depends on `keytool` that depends on `schema`.
```
tbekas: Wait, I thought we want to avoid transitive dependencies? So If some module is using classes from `shared` it's not enough to depend just on the `schema`, it has to depend on `shared`.

doc/adr/0012-modules-structure.md
```diff
@@ -0,0 +1,105 @@
...
+    E -.-> A
+    E --> F
+    F --> C
+    F -.-> B
+    F -.-> A
+    G --> E
+    G -.-> C
+    G -.-> B
+    G -.-> A
+    G --> F
```
TheMMaciek, line 48: Same here I think. It can be a transitive dependency.

doc/adr/0012-modules-structure.md
```diff
@@ -0,0 +1,105 @@
...
+Then, all the things related to node operations would be kept in `node-shared` module (including gossip protocol, consensus, etc.). This leads to emptying the SDK module, so we will be able to implement the SDK module as a proxy between the external world and the internal tessellation structure. SDK module should reexpose all necessary operations to be used by 3rd parties. For example - to create a keypair, global L0 can use the keytool module directly, but 3rd party will use the SDK module that will reexpose keypair creation function. That way, the internal structure of tessellation will be hidden behind the facade.
+
+Modules will be transient - for example: `wallet` does not need to depend on `schema` directly, because it depends on `keytool` that depends on `schema`.
+
+A separate ADR will organize dependencies (libraries) of modules.
+
+- `shared` will contain base classes and extensions (like kryo or refined extensions),
+- `schema` will contain tessellation schema (like Address, Balance, GlobalSnapshot, base Block or Transaction, base State Channels types),
```
tbekas: I think it's almost a circular definition "schema will contain schema".

My understanding is that a `schema` is all the stuff that's inside of the `GlobalSnapshot` (including the `GlobalSnapshot` itself) and nothing else. What's your understanding of schema?

doc/adr/0012-modules-structure.md
marcinwadon: fixed :+1:

doc/adr/0012-modules-structure.md
marcinwadon: actually I removed dotted lines. it was misleading (and not, it wasn't transient but could be)

doc/adr/0012-modules-structure.md
```diff
@@ -0,0 +1,104 @@
...
+Modules will be transient - for example: `wallet` does not need to depend on `shared` directly, because it depends on `schema` that depends on `shared`.
+
+A separate ADR will organize dependencies (libraries) of modules.
+
+- `shared` will contain base classes and extensions (like kryo or refined extensions),
+- `schema` will contain tessellation schema (like Address, Balance, GlobalSnapshot, base Block or Transaction, base State Channels types) and key-related operations,
+- `wallet` will contain operations required by the wallet-cli (like creating transactions), but also will expose key operations (like creating a new key)
+- `dag-shared` will contain shared types and classes between Global L0 and DAG L1, related to $DAG (like $DAG block or $DAG transaction),
+- `node-shared` will contain all the operations required to launch a node (like base TessellationIO app, gossip protocol, cluster management or joining algorithm),
+- `core` will remain the Global L0 module,
```
TheMMaciek, line 74: Shouldn't we rename it to `global-l0` or something?

doc/adr/0012-modules-structure.md
marcinwadon: fixed :+1:

doc/adr/0012-modules-structure.md
marcinwadon: imo it will make more burden but `core` also is a valid name, as the Global Layer 0 is the most underlying (base) layer, so a core for everything else.

doc/adr/0012-modules-structure.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Decision
+
+To simplify the structure of modules, we will split `shared` into two separate modules - the first containing scala and libraries extensions, the latter containing the tessellation schema. `keytool` module will be splitted into: operations/functions that we will put to `schema` and exposing an app that will be combined into `wallet` module.
+Then, all the things related to node operations would be kept in `node-shared` module (including gossip protocol, consensus, etc.). This leads to emptying the `sdk` module, so we will be able to implement the `sdk` module as a proxy between the external world and the internal tessellation structure. `sdk` module should reexpose all necessary operations to be used by 3rd parties. For example - to create a keypair, global L0 can use the `schema` module directly, but 3rd party will use the `sdk` module that will reexpose keypair creation function. That way, the internal structure of tessellation will be hidden behind the facade.
+
+Modules will be transient - for example: `wallet` does not need to depend on `shared` directly, because it depends on `schema` that depends on `shared`.
```
tbekas, line 65: Wait, I think we don't want to have transitive dependencies. If a module wants to use functionality from `shared` then it should depend on `shared`, not just rely on the fact that `schema` depends on `shared`

TheMMaciek: Not sure that we decided about it and that it's set in stone. I wouldn't state that here. We can say that in the new schema graph we don't explicitly mark the transitive dependencies, but it's not yet decided how the dependencies will be handled.

doc/adr/0012-modules-structure.md
kpudlik: Imo `core` is fine because `global-l0` sounds like there was `global-l1` as well

doc/adr/0012-modules-structure.md
marcinwadon: right, wrong words.
anyway, the current state of `schema` isn't just what's in `GlobalSnapshot`, but also gossip classes, node state, peer id, etc.

doc/adr/0012-modules-structure.md
TheMMaciek: For me it's a layer that brings all the layers (those which decide to use global L0 of course) together so rather the topmost layer hence global-l0. But I understand that in a way you can also think about it as a core layer, because lower layers check back with the topmost layer to get aligned. Anyway for me it's rather the topmost layer and naming it `core` sounds misleading for me. We don't have to change it if we don't agree, especially that it doesn't change any functionality and it's just a name.

doc/adr/0012-modules-structure.md
tbekas: Ok, but this part of the ADR is not about the current state but about the target state. I don't think that `NodeState` or `gossip` related stuff should be part of the schema (in my understanding of what the "schema" is).

doc/adr/0012-modules-structure.md
tbekas: I agree that `core` sounds misleading. I would go either with `dag-l0` or `global-l0` (preferably the first one)

marcinwadon: but it's not only a `dag-l0` @tbekas, because `dag-l0` is just a part of this module.

doc/adr/0012-modules-structure.md
```diff
@@ -0,0 +1,84 @@
...
+
+Modules' dependencies and libraries cannot be transient. We will use `% Optional` scope for all internal modules.
+
+- `shared` will contain base classes and extensions (like kryo or refined extensions), all the stuff used within the GlobalSnapshot (like Address, base Block or Transaction), base StateChannels types and key-related operations (like getting address from public key),
+- `wallet` will contain operations required by the wallet-cli (like creating transactions), but also will expose key operations (like creating a new key),
+- `dag-shared` will contain shared types and classes between Global L0 and DAG L1, related to $DAG (like $DAG block or $DAG transaction),
+- `node-shared` will contain all the operations required to launch a node (like base TessellationIO app, gossip protocol, cluster management or joining algorithm),
+- `global-l0` will be renamed from `core` and will remain the Global L0 module,
+- `dag-l1` will remain the DAG L1 module.
```
tbekas, line 58: Generally speaking I would rather see this list rephrased as definitions (in a present tense):

```
- `shared` - base library module containing `GlobalSnapshot` as well as extensions for 3rd party libraries
- `dag-shared` - an internal library module containing classes shared only between `dag-l1` and `global-l0`
- `node-shared` - an internal library module containing all the classes necessary to launch a node (like base TessellationIO app, gossip protocol, cluster management or joining algorithm),
- `sdk` - a public library module exposing a set of functionalities ... 
- `tools` an internal app used for generating transactions and state channel snapshots
- `wallet` - CLI app for creating transactions and managing keys
- `global-l0` - the Global L0 node app (renamed from `core`)
- `dag-l1` - the DAG L1 node app (remains unchanged)
```

tbekas (changes requested):
It looks pretty good already, but I would like to see a bit more precise definitions. Also we're missing a `tools` module.

marcinwadon:
@tbekas could you help with more precise ones then? I'll add tools module

doc/adr/0012-modules-structure.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Decision
+
+`keytool` module will be splitted into: operations/functions that we will put to `shared` and exposing an app that will be combined into `wallet` module.
+Then, all the things related to node operations would be kept in `node-shared` module (including gossip protocol, consensus, etc.). This leads to emptying the `sdk` module, so we will be able to implement the `sdk` module as a proxy between the external world and the internal tessellation structure. `sdk` module should reexpose all necessary operations to be used by 3rd parties. For example - to create a keypair, global L0 can use the `shared` module directly, but 3rd party will use the `sdk` module that will reexpose keypair creation function. That way, the internal structure of tessellation will be hidden behind the facade.
+
+Modules' dependencies and libraries cannot be transient. We will use `% Optional` scope for all internal modules.
```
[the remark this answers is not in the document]
marcinwadon: :+1:

doc/adr/0012-modules-structure.md
marcinwadon: :+1: rephrased

doc/adr/0012-modules-structure.md
```diff
@@ -0,0 +1,94 @@
...
+    H --> F
+```
+
+Furthermore, dependencies were partially explicitly declared and partially transient. After some time, it became unmaintainable.
+
+The second (and main) reason for restructuring modules is the fact that 3rd parties cannot use the SDK module in the current form. It does not expose all required operations, so internal modules were used directly, leaking the project's internal structure and locking further internal changes.
+
+## Decision
+
+`keytool` module will be splitted into: operations/functions that we will put to `shared` and exposing an app that will be combined into `wallet` module.
```
TheMMaciek, line 48: "will be split"

doc/adr/0012-modules-structure.md
```diff
@@ -0,0 +1,94 @@
...
+Then, all the things related to node operations would be kept in `node-shared` module (including gossip protocol, consensus, etc.). This leads to emptying the `sdk` module, so we will be able to implement the `sdk` module as a proxy between the external world and the internal tessellation structure. `sdk` module should reexpose all necessary operations to be used by 3rd parties. For example - to create a keypair, global L0 can use the `shared` module directly, but 3rd party will use the `sdk` module that will reexpose keypair creation function. That way, the internal structure of tessellation will be hidden behind the facade.
```
TheMMaciek, line 49: "will be kept"

doc/adr/0012-modules-structure.md
```diff
@@ -0,0 +1,94 @@
...
+- `sdk` 
+
+- `shared` - base internal library containing `GlobalSnapshot` schema as well as extensions for 3rd party libraries
+- `dag-shared` - an internal library containing $DAG-related classes (like $DAG block or $DAG transaction)
+- `node-shared` - an internal library containing all the classes necessary to launch a node (like base TessellationIO app, gossip protocol, cluster management or joining algorithm)
+- `wallet` - an app for creating transactions and managing keys
+- `global-l0` - the Global L0 node app (renamed from `core`)
+- `dag-l1` - the DAG L1 node app (remains unchanged)
+- `sdk` - a library exposing a set of functionalities as a Tessellation public interface (facade) for 3rd parties
+- `tools` - an internal app used for generating transactions, state channel snapshots and other things needed for development, maintenance or testing
```
tbekas, line 68: I like new definitions, but you kept both lists: the old one and the new one. I think we can keep only the new one.

tbekas (changes requested):

doc/adr/0012-modules-structure.md
marcinwadon: :+1:

tbekas (approved):
LGTM

tbekas (approved):

kriscieslak (approved):

kpudlik (approved):

TheMMaciek (approved):

PR #531, opened 2023-01-04
Extract state channels schema to shared module

marcinwadon:


modules/shared/src/main/scala/org/tessellation/statechannel/StateChannelSnapshot.scala
```diff
@@ -0,0 +1,7 @@
+package org.tessellation.statechannel
+
+import org.tessellation.security.hash.Hash
+
+trait StateChannelSnapshot {
```
TheMMaciek, line 5: Seems that we don't use it anywhere anymore. Do we want to keep it in schema for the future?

TheMMaciek (dismissed):

modules/shared/src/main/scala/org/tessellation/statechannel/StateChannelSnapshot.scala
marcinwadon: @TheMMaciek you're right. I've just removed it :+1:

TheMMaciek (approved):

PR #543, opened 2023-01-17
Added Rosetta classes, part 5.

cngo-github:
__PR Category__: Feature

Added the classes needed for Rosetta.

This lays the groundwork for the implementation of the systems
needed for the Rosetta endpoints, part 5.


modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/BlockSearchRequest.scala
```diff
@@ -0,0 +1,12 @@
+package org.tessellation.rosetta.server.model
+
+case class BlockSearchRequest(
```
marcinwadon, line 3: is it autogenerated as well?

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/BlockSearchRequest.scala
```diff
@@ -0,0 +1,12 @@
...
+  isOr: Boolean,
+  isAnd: Boolean,
```
marcinwadon, line 5: it's unsecure. what if `isOr` and `isAnd` will both be true or false?

marcinwadon: I see that rosetta uses enum `operator: or/and`, not booleans (https://www.rosetta-api.org/docs/models/Operator.html)

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/RosettaSearch.scala
```diff
@@ -0,0 +1,20 @@
...
+
+  def searchByDestinationAddress(transactionHash: String, snapshots: List[Transaction]) =
+    snapshots.filter(_.destination == transactionHash)
+
+  def searchByAddress(transactionHash: String, snapshots: List[Transaction]) = {
+    val sourceTransactions = searchBySourceAddress(transactionHash, snapshots)
+    val destinationTransactions = searchByDestinationAddress(transactionHash, snapshots)
+
+    sourceTransactions ++ destinationTransactions
+  }
```
marcinwadon, line 18: why it's in the model? also, argument `snapshot` is of type `List[Transaction]` which is wrong.

marcinwadon (changes requested):

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/RosettaSearch.scala
```diff
@@ -0,0 +1,20 @@
+package org.tessellation.rosetta.server.model
+
+import org.tessellation.schema.transaction.Transaction
+
+object RosettaSearch {
+
+  def searchBySourceAddress(transactionHash: String, snapshots: List[Transaction]) =
+    snapshots.filter(_.source == transactionHash)
```
tbekas, line 8: 1. This function is pretty much an alias to `filter`
2. It compares a property of type `Address` with a `String` parameter named `transactionHash`

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/RosettaSearch.scala
```diff
@@ -0,0 +1,20 @@
...
+
+  def searchByDestinationAddress(transactionHash: String, snapshots: List[Transaction]) =
+    snapshots.filter(_.destination == transactionHash)
```
tbekas, line 11: Same as above

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/GenerateCodecs.scala
```diff
@@ -0,0 +1,20 @@
...
+      .apply(Path("modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model"))
+      .toDirectory
+      .list
+      .filter(_.isFile)
+      .foreach { p =>
+        val name = p.name.replaceAll(".scala", "")
+        val gen = s"""implicit lazy val ${name}Encoder = deriveConfiguredMagnoliaEncoder[${name}]
+                      |implicit lazy val ${name}Decoder = deriveConfiguredMagnoliaDecoder[${name}]""".stripMargin
+        println(gen)
+      }
```
tbekas, line 18: What is this?

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/decoders.scala
```diff
@@ -0,0 +1,231 @@
+package org.tessellation.rosetta.server.model.dag
+
+import org.tessellation.rosetta.server.model._
+import org.tessellation.rosetta.server.model.dag.metadataSchema._
+import org.tessellation.rosetta.server.model.dag.schema._
+
+/** This whole thing has to be manually re-arranged after generating because of implicit dependencies requiring to be in order.
+  */
```
tbekas, line 8: I doubt we want to do this manually. I would rather see a complete yaml specification of OpenAPI as well as the generator configured either via the [SBT plugin](https://github.com/OpenAPITools/sbt-openapi-generator) or via the custom sbt task, that will use a custom Mustache [template](https://openapi-generator.tech/docs/templating/) that will add all the necessary derivations to the generated classes (so they don't need to be stored as .scala sources in the repo but generated during the build).

tbekas (changes requested):

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/RosettaSearch.scala
```diff
@@ -0,0 +1,20 @@
+package org.tessellation.rosetta.server.model
+
+import org.tessellation.schema.transaction.Transaction
+
+object RosettaSearch {
+
+  def searchBySourceAddress(transactionHash: String, snapshots: List[Transaction]) =
```
TheMMaciek, line 7: Why is the `snapshots` parameter of type `List[Transaction]`?

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/GenerateCodecs.scala
```diff
@@ -0,0 +1,20 @@
+package org.tessellation.rosetta.server.model.dag
+
+import scala.reflect.io.Path
+
+object GenerateCodecs {
+
+  def run(): Unit =
+    scala.tools.nsc.io.File
```
TheMMaciek, line 8: We are using `File` from fs2. I guess you could use it here as well.

kpudlik (changes requested):

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/GenerateCodecs.scala
cngo-github: This is a work around for `sbt autogen` being broken (openAPI). @ryle-goehausen tried to get an autogen working, and this was the one that he found to be the least broken. I have added a `TODO` noting the purpose of this code and to remove it when feasible.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/decoders.scala
cngo-github: See the above comment on `sbt autogen` being broken. An automated way to do this is necessary, but I think it will take considerable effort and should be considered out of scope for this PR.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/RosettaSearch.scala
cngo-github: This object is unneeded. Removed.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/BlockSearchRequest.scala
```diff
@@ -0,0 +1,12 @@
...
+  addressOpt: Option[String],
```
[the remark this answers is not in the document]
cngo-github: Changed.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/RosettaSearch.scala
cngo-github: This object is unneeded. Removed.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/RosettaSearch.scala
cngo-github: This object is unneeded. Removed.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/RosettaSearch.scala
cngo-github: This object is unneeded. Removed.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/BlockSearchRequest.scala
cngo-github: Changed to use an internal enum.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/BlockSearchRequest.scala
cngo-github: This is an intermediary class to hold the values from the Request that is relevant to our network.

marcinwadon: imo it should be similar to search request provided by rosetta api (example: operator as enum) instead of custom fields like here

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/BlockSearchRequest.scala
```diff
@@ -0,0 +1,12 @@
...
+  networkStatus: Option[String],
+  limit: Option[Long],
+  offset: Option[Long],
+  transactionHash: Option[String],
```
marcinwadon, line 10: what is the purpose of this field?

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/decoders.scala
marcinwadon: @cngo-github what's exactly broken with `sbt autogen`?
code like this (the file above) is risky and hard to maintain. even a custom sbt task would way better, but ideally it should be done via templates as @tbekas suggested.
if you want to do it out of scope for this PR, then please create separate PR for auto generated classes first and then separate one to regenerate all already created classes.

cngo-github: @ryle-goehausen , can you please elucidate what about `sbt autogen`?

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/BlockSearchRequest.scala
cngo-github: `isOr` and `isAnd` have been replaced by `operation: BlockSearchRequestOperation` where `BlockSearchRequestOperation` is an internal version of the operator enum.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/GenerateCodecs.scala
cngo-github: Can you please point me at a tutorial on using FS2 with a directory? So far, I've figured out

```scala
Files[F].walk(Path("modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model"))
```

However, I don't know how to check if a `Path` is a file and I can't call `foreach`.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/BlockSearchRequest.scala
cngo-github: This allows for the hash of a specific transaction in a block to be specified in the request call.

https://www.rosetta-api.org/docs/BlockApi.html#blocktransaction

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/decoders.scala
ryle-goehausen: Sure, the list of generators I tried was from here under the Scala section: https://github.com/OpenAPITools/openapi-generator/tree/master/docs/generators -- I believe I went through like 7 different generators I forget all the names, but it was from that repo. I also tried some of the code gen builtins in some IDEs, each one had a number of flaws -- either generating unnecessary or redundant code or throwing exceptions halfway through or failing to generate the correct classes. The last time I looked into this there didn't seem to be an easy solution, and a bunch of complaints for other OpenAPI failures related to Scala. I think it's specific to the yaml file that's being used, as it works in plenty of other cases but didn't for this. Possible I misconfigured it -- but I think it's appropriate to take a bit of a shortcut here rather than dealing with patching / custom SBT configurations since we don't anticipate needing to update for a while. 

I think it'd be reasonable to revisit this assuming there's a requirement to upgrade rosetta classes in the future, in which case what is present should be reasonably close to an sbt build step.

tbekas: @ryle-goehausen this is exactly what I addressed in my comment. Just use a custom Mustache [template](https://openapi-generator.tech/docs/templating/). If an sbt-plugin is "broken" as @cngo-github says we can write a custom sbt task as @marcinwadon suggested, but that's only when sbt-plugin can't be used.

Making everything manual makes `rosetta` module very difficult to maintain and prone to introducing errors (desyncing classes with their autogenerated counterparts).

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/GenerateCodecs.scala
tbekas: > Switching over to FS2 doesn't really provide real value.

It does provide a value. Using the same approach across the whole codebase - fs2 and cats-effect for all IO operations makes the code consistent and interoperable.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/BlockSearchRequest.scala
marcinwadon: but it's a search api (https://www.rosetta-api.org/docs/SearchApi.html) and a block search request, not transaction search request.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/GenerateCodecs.scala
tbekas: Even if `sbt autogen` would be broken it wouldn't justify this code. Please implement this `TODO` as a part of the current changes.

cngo-github:
I refactored the generated classes and their support classes to the `org.tessellation.rosetta.api.generated` package. Non-generated classes that are for the Rosetta endpoints will reside in the `org.tessellation.rosetta.server.model` package. I will add ore to that package in subsequent PRs as I merge in the implementations. The `rosetta` module now depends on the `dagShared` module because of [this](https://github.com/cngo-github/tessellation/blob/20230117-Rosetta_classes_pt5/modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/Registry.scala).

modules/rosetta/src/main/scala/org/tessellation/rosetta/api/generated/Operation.scala
```diff
@@ -5,9 +5,9 @@
   * NOTE: This class is auto generated by OpenAPI Generator (https://openapi-generator.tech). https://openapi-generator.tech
   */
 
-package org.tessellation.rosetta.server.model
+package org.tessellation.rosetta.api.generated
 
-import org.tessellation.rosetta.server.model.dag.metadataSchema.GenericMetadata
+import org.tessellation.rosetta.api.generated.dag.metadataSchema.GenericMetadata
```
marcinwadon, line 10: is `dag` also auto generated?

modules/rosetta/src/main/scala/org/tessellation/rosetta/api/generated/dag/GenerateCodecs.scala
```diff
@@ -0,0 +1,27 @@
...
+import scala.reflect.io.Path
+import scala.tools.nsc.io.File
+
+/** This is a work-around for `sbt autogen` being broken (openAPI). @ryle-goehausen tried to get an autogen working, and this was the one
+  * that he found to be the least broken.
+  *
+  * TODO: Remove this when feasible.
+  */
+
+object GenerateCodecs {
```
marcinwadon, line 12: this is not auto generated but it is in `api.generated` package

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/BlockSearchRequest.scala
marcinwadon: why `Operation`? it should be an `Operator`. and why there is a need for internal version of it, if Operator supports only `and` and `or` and there will be no different operators?

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/Registry.scala
```diff
@@ -0,0 +1,20 @@
...
+import org.tessellation.ext.kryo.KryoRegistrationId
+import org.tessellation.schema.transaction.RewardTransaction
+
+import eu.timepit.refined.numeric.Interval
+
+object Registry {
+
+  import eu.timepit.refined.auto._
+
+  type RunnerKryoRegistrationIdRange = Interval.Closed[300, 399]
```
marcinwadon, line 12: interval overlaps with existing range

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/Registry.scala
```diff
@@ -0,0 +1,20 @@
...
+
+  type RunnerKryoRegistrationId = KryoRegistrationId[RunnerKryoRegistrationIdRange]
+
+  val rosettaKryoRegistrar: Map[Class[_], RunnerKryoRegistrationId] = Map(
+    classOf[RewardTransaction] -> 389
+  )
```
marcinwadon, line 18: why it's needed?

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/network.scala
```diff
@@ -0,0 +1,52 @@
...
+  implicit def showNonEmptyString: Show[NonEmptyString] = s => s.value
+
+  @derive(show)
+  @newtype
+  case class NetworkStatus(value: NonEmptyString)
+
+  object NetworkStatus {
+    val accepted: NetworkStatus = NetworkStatus("Accepted")
+    val reverted: NetworkStatus = NetworkStatus("Reverted")
+    val canceled: NetworkStatus = NetworkStatus("Canceled")
```
marcinwadon, line 25: how network can be accepted, reverted or canceled?

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/network.scala
```diff
@@ -0,0 +1,52 @@
...
+
+    implicit val quillEncode: MappedEncoding[NetworkStatus, String] =
+      MappedEncoding[NetworkStatus, String](_.value.value)
+
+    implicit val quillDecode: MappedEncoding[String, NetworkStatus] = MappedEncoding[String, NetworkStatus](
+      refineV[NonEmpty].apply[String](_).leftMap(new Throwable(_)) match {
+        case Left(err)    => throw err
+        case Right(value) => NetworkStatus(value)
+      }
+    )
```
marcinwadon, line 35: why you need quill if there is no database yet?

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/network.scala
```diff
@@ -0,0 +1,52 @@
...
+        return Some(accepted)
+
+      if (normalizedString == reverted.value.value.toLowerCase())
+        return Some(reverted)
+
+      if (normalizedString == canceled.value.value.toLowerCase())
+        return Some(canceled)
+
+      None
+    }
```
marcinwadon, line 50: it can be handled by enum

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/network.scala
```diff
@@ -0,0 +1,52 @@
+package org.tessellation.rosetta.server.model
+
+import cats.Show
+import cats.implicits.toBifunctorOps
```
marcinwadon, line 4: please don't use `cats.implicits`. you should use `cats.syntax` as it's done in the whole project

modules/rosetta/src/main/scala/org/tessellation/rosetta/api/generated/Operation.scala
cngo-github: The stuff in `org.tessellation.rosetta.api.generated.dag` is not autogenerated, but it is used with the auto-generated classes in `org.tessellation.rosetta.api.generated`. For example, it has enums.

modules/rosetta/src/main/scala/org/tessellation/rosetta/api/generated/dag/GenerateCodecs.scala
cngo-github: This is a utility to generate the stuff. I'll move it into a package adjacent to `org.tessellation.rosetta.api.generated`.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/Registry.scala
cngo-github: This is needed for the Rosetta endpoint implementations. `KryoSerializer` is used a lot there.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/network.scala
cngo-github: This denotes the status of a transaction on a network. So a transaction can be `Accepted`, `Reverted`, or `Canceled`.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/network.scala
cngo-github: This is for when a transaction is written to a database in the implementations of the Rosetta server. Those implementations will be in the `dagL1` module and `quill` will be used there.

modules/rosetta/src/main/scala/org/tessellation/rosetta/api/generated/Operation.scala
cngo-github: I moved the `dag` package to `org.tessellation.rosetta.api` and out of `org.tessellation.rosetta.api.generated`.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/network.scala
cngo-github: I forgot to add `quill` to the dependencies for the `rosetta` module. That has been rectified.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/Registry.scala
cngo-github: Please elucidate this comment for me.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/BlockSearchRequest.scala
cngo-github: It is used in the [BlockApi](https://www.rosetta-api.org/docs/BlockApi.html).

modules/rosetta/src/main/scala/org/tessellation/rosetta/api/generated/Operation.scala
marcinwadon: Cool. You said that `api.generated` contains auto-generated classes so let's keep that exclusively for generated classes.

modules/rosetta/src/main/scala/org/tessellation/rosetta/api/generated/dag/GenerateCodecs.scala
marcinwadon: You don't need to move it. Please see https://github.com/Constellation-Labs/tessellation/pull/543#discussion_r1073309026 and https://github.com/Constellation-Labs/tessellation/pull/543#discussion_r1073286301

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/GenerateCodecs.scala
cngo-github: I will need assistance with this. We can go over it when we meet @tbekas next week. Alternatively, please point me to a good tutorial.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/decoders.scala
cngo-github: I will need assistance with this. We can go over it when we meet @tbekas next week. Additionally, please point me to a good tutorial.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/Registry.scala
marcinwadon: I don't see any need for this for now. Also, RewardTransaction (if it's the one from global snapshot, it's already registered in kryo). You will be mostly mapping rosetta classes to our domain classes and these classes are already registered in kryo.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/GenerateCodecs.scala
cngo-github: I will need assistance with this. We can go over it when we meet @tbekas next week. Additionally, please point me to a good tutorial.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/network.scala
marcinwadon: the name of this doesn't say that it's a status of a transaction on a network, but a status of the network itself.
also, is it 1:1 correspondence from rosetta? because transaction cannot be canceled or reverted in our network. eventually it can be not accepted

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/network.scala
marcinwadon: Can you separate this changes and add database support and quill in later pull request, when it would be needed?

marcinwadon: because this pull request was meant to add last auto generated classes (as the name says) and right now we have plenty of other stuff in here

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/BlockSearchRequest.scala
cngo-github: I've changed the enum to `BlockSearchRequestOperator`. The one I made is for internal use so that we don't have to tie. our code base to an enum specified by a external entity.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/network.scala
cngo-github: Changed the import to `cats.syntax.bifunctor._`.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/network.scala
cngo-github: Reworked this to an enum, renamed the file `NetworkStatus.scala`, and relocated it into the `org.tessellation.rosetta.server.model.enums` package.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/network.scala
cngo-github: Examples of usage are at [here](https://www.rosetta-api.org/docs/1.3.1/BlockApi.html#200---ok-1) and [here](https://www.rosetta-api.org/docs/1.3.1/MempoolApi.html#200---ok-1) where it is part of the transaction information returned. I will confer with @ryle-goehausen on the history of this class.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/network.scala
cngo-github: I'll make the requested changes.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/BlockSearchRequest.scala
marcinwadon: Yes, so it's used in a different request than you model here (it's used in block api, not in the search api)

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/network.scala
marcinwadon: Yes, it's used as a status of a transaction, not the network.

cngo-github: So `NetworkStatus` is the `status` property [here](https://www.rosetta-api.org/docs/models/Operation.html). I will change the name to `TransactionOperationStatus`.

modules/rosetta/src/main/scala/org/tessellation/rosetta/api/generated/dag/GenerateCodecs.scala
cngo-github: I will need assistance automating this task. Once that is done, this can be removed.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/BlockSearchRequest.scala
marcinwadon: but you will use it also for other requests, so it's not exclusively for BlockSearchRequest.
You can name it `Operator` as you will map it one way or another to rosetta operator, so it should be corresponding (isomorphic)

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/GenerateCodecs.scala
marcinwadon: @cngo-github both [cats-effect ](https://typelevel.org/cats-effect/)and [fs2](https://fs2.io/#/) have great documentation with plenty of examples and use-cases.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/BlockSearchRequest.scala
marcinwadon: additionally, in the API you provided above, there is transaction hash (identifier) required to get given transaction from a specific block (you specify the block identifier and transaction identifier you want to get), so following REST API specification, it's a request to the `/block/transaction` endpoint.
Here, you have `BlockSearchRequest`, so you search not a transaction but a block. The example provided by rosetta is to search transaction, that's why you need to provide a transaction identifier. When you change transaction->block, you should provide a block identifier to search for.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/Registry.scala
marcinwadon: Kryo can be used with or without pre-registration. We require it for security. You register a given class under numeric identifier which is used then for deserialization - hence, identifiers must be unique. You cannot serialize two different classes with the same identifier as you will not be able to deserialize it.
It's simple to keep everything clean when you have one module and one registration map, but in the project we have many modules and all of the registrations are combined into single one and single kryo pool.
That being said, we have to somehow manage these identifiers not to duplicate anything. 
See https://github.com/Constellation-Labs/tessellation/blob/develop/doc/adr/0009-kryo-class-registration-policy.md for the kryo class registration policy we use currently. Interval 300-399 is reserved exclusively for `shared` module, so you cannot reuse the same identifiers here, as they will be duplicated which will cause (de)serialization issues.

Nevertheless, for now you don't need kryo for the rosetta as there is nothing that currently uses it. You need kryo for two cases:
- when you need a hash of an object,
- when you need a binary format of an object.

I believe that rosetta classes will neither be required in binary format, nor hashed, so you don't need to register classes in kryo at all. You can (and for sure you will need) kryo pool, to get the hash of a transaction or a block, but these classes are already registered in their respective modules.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/network.scala
marcinwadon: :+1:

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/network.scala
marcinwadon: I have one more suggestion - in the whole codebase, we don't group files and classes by type but by feature.
So packages shouldn't be:
- utils
- traits
- enums
and so on, but rather
- transactions
- blocks
- snapshots
- consensus
- gossip
- ...

That being said, you can group these enums by feature (for example `transaction.TransactionOperationStatus` or `search.Operator`

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/network.scala
marcinwadon: :+1:

build.sbt
```diff
@@ -1,4 +1,4 @@
-import Dependencies._
+import Dependencies.{Libraries, _}
```
marcinwadon, line 1: why?

build.sbt
```diff
@@ -320,7 +320,9 @@ lazy val rosetta = (project in file("modules/rosetta"))
     libraryDependencies := Seq(
       CompilerPlugin.kindProjector,
       CompilerPlugin.betterMonadicFor,
-      CompilerPlugin.semanticDB
+      CompilerPlugin.semanticDB,
+        Libraries.doobieCore,
+    Libraries.doobieQuill,
```
marcinwadon, line 325: it's not required now, for auto generated classes

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/BlockSearchRequest.scala
cngo-github: I misspoke. `BlockSearchRequest` is used by the [`/search/transaction` endpoint](https://www.rosetta-api.org/docs/SearchApi.html#searchtransactions). The `/block` endpoints do not use intermediary request objects. One endpoint does use the `BlockResponse` intermediary response object. I have renamed `BlockSearchRequest` to `SearchTransactionRequest`.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/network.scala
cngo-github: The `quill` dependencies and decoders have been removed.

build.sbt
cngo-github: I'm not sure how that happened. It has been rectified.

build.sbt
cngo-github: Removed.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/Registry.scala
cngo-github: This file is not currently needed and may not be needed anymore. It has been removed.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/Registry.scala
cngo-github: This file is not currently needed and may not be needed anymore. It has been removed.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/network.scala
cngo-github: The `/search` endpoints stuff in `org.tessellation.rosetta.server.model` have all been relocated to `org.tessellation.rosetta.server.model.search`. In `org.tessellation.rosetta.api.dag`, `BlockEventType` has been relocated under `org.tessellation.rosetta.api.dag.block`, `SignatureType` has been relocated under `org.tessellation.rosetta.api.dag.crypto`, and `ChainObjectStatus` and `NetworkChainObjectStatus` has been relocated under `org.tessellation.rosetta.api.dag.network`. I am not really happy about the `network` package name and welcome suggestions for a better name.

build.sbt
```diff
@@ -320,7 +320,7 @@ lazy val rosetta = (project in file("modules/rosetta"))
     libraryDependencies := Seq(
       CompilerPlugin.kindProjector,
       CompilerPlugin.betterMonadicFor,
-      CompilerPlugin.semanticDB
+      CompilerPlugin.semanticDB,
```
marcinwadon, line 323: ?

modules/rosetta/src/main/scala/org/tessellation/rosetta/api/dag/schema.scala
```diff
@@ -1,17 +1,11 @@
-package org.tessellation.rosetta.server.model.dag
-
-import org.tessellation.rosetta.server.model.dag.enums.ChainObjectStatus
+package org.tessellation.rosetta.api.generated.dag
```
marcinwadon, line 1: why these classes are in generated package, if the comment clearly says it's not auto generated?

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/search/SearchTransactionRequest.scala
```diff
@@ -0,0 +1,14 @@
+package org.tessellation.rosetta.server.model.search
+
+/** This is an intermediary object used to hold the values from the Rosetta request that is relevant to our system.
+  */
+
+case class SearchTransactionRequest(
+  operation: SearchTransactionTransactionRequestOperator,
```
marcinwadon, line 7: so it's an operation or an operator? field name specify it differently that the type

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/search/SearchTransactionRequest.scala
```diff
@@ -0,0 +1,14 @@
+package org.tessellation.rosetta.server.model.search
+
+/** This is an intermediary object used to hold the values from the Rosetta request that is relevant to our system.
+  */
```
marcinwadon, line 4: if this is an intermediary object, shouldn't it hold our codebase types? what is the purpose of this case class and how it will be used?

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/search/SearchTransactionTransactionRequestOperator.scala
```diff
@@ -0,0 +1,32 @@
...
+import enumeratum.{Enum, EnumEntry}
+import io.circe.{Decoder, Encoder}
+
+/** This enumeration is for internal use only and meant to mirror:
+  *
+  * https://www.rosetta-api.org/docs/models/Operator.html
+  */
+
+@derive(eqv, show)
+sealed trait SearchTransactionTransactionRequestOperator extends EnumEntry
```
marcinwadon, line 16: imo it should be just `Operator` to mimic the rosetta specification. you already did that for other case classes that were auto-generated as empty and you created an enum for that. then you can reuse it in each search request. otherwise, you'd have to duplicate this code (because by its name it can be used for search transaction request only)

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/search/TransactionOperationStatus.scala
```diff
@@ -0,0 +1,26 @@
...
+
+@derive(eqv, show)
+sealed trait TransactionOperationStatus extends EnumEntry
+
+object TransactionOperationStatus extends Enum[TransactionOperationStatus] {
+  val values = findValues
+
+  case object Accepted extends TransactionOperationStatus
+  case object Reverted extends TransactionOperationStatus
+  case object Canceled extends TransactionOperationStatus
```
marcinwadon, line 18: have you confirmed what about this status?

modules/rosetta/src/main/scala/org/tessellation/rosetta/api/dag/network/NetworkChainObjectStatus.scala
```diff
@@ -0,0 +1,5 @@
+package org.tessellation.rosetta.api.dag.network
+
+import org.tessellation.rosetta.api.generated.dag.network.ChainObjectStatus
+
+case class NetworkChainObjectStatus(value: ChainObjectStatus)
```
marcinwadon, line 5: what is the purpose of this case class and how it will be used?

build.sbt
cngo-github: Typo. Corrected.

modules/rosetta/src/main/scala/org/tessellation/rosetta/api/dag/metadataSchema.scala
```diff
@@ -1,4 +1,4 @@
-package org.tessellation.rosetta.server.model.dag
+package org.tessellation.rosetta.api.generated.dag
```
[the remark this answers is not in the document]
cngo-github: I forgot to update the package name. They have been updated to `package org.tessellation.rosetta.api.dag`.

modules/rosetta/src/main/scala/org/tessellation/rosetta/api/dag/schema.scala
cngo-github: I forgot to update the package name. They have been updated to `package org.tessellation.rosetta.api.dag`.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/search/SearchTransactionRequest.scala
cngo-github: The purpose of this case class is to hold data extracted from the request and to be passed around to different methods that need that data. Basically, it holds the parameters for various methods. Where is the code base types?

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/search/SearchTransactionTransactionRequestOperator.scala
cngo-github: Changed to `Operator`.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/search/TransactionOperationStatus.scala
cngo-github: This denotes the status of a transaction operation on the network. Per the [Rosetta docs](https://www.rosetta-api.org/docs/models/Operation.html), a smart contracts may have non-atomic transactions that can be partially applied. This is the status for the whole of the transaction/smart contract, not the individual operations contained within.

modules/rosetta/src/main/scala/org/tessellation/rosetta/api/dag/network/NetworkChainObjectStatus.scala
cngo-github: I will confer with @ryle-goehausen and follow-up with you.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/search/SearchTransactionRequest.scala
marcinwadon: ok, sure. Is there no specification in the openapi so that the class can be auto-generated?
By codebase types I mean for example `Address` or `DAGAddress` instead of `String`, or `Hash` instead of `String`.

cngo-github: I'll confer with @tbekas and @ryle-goehausen on this.

modules/rosetta/src/main/scala/org/tessellation/rosetta/api/dag/network/NetworkChainObjectStatus.scala
cngo-github: This serves the same purpose as `TransactionOperationStatus` and is redundant. It has been removed.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/search/TransactionOperationStatus.scala
marcinwadon: to keep it in line with its name on Rosetta, it should be `OperationStatus`

cngo-github: Renamed.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/BlockSearchRequest.scala
cngo-github: It has been renamed.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/search/TransactionOperationStatus.scala
cngo-github: Renamed to `OperationStatus`.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/search/SearchTransactionRequest.scala
cngo-github: After discussing with @ryle-goehausen , I'll go ahead and add codebase types to this class.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/GenerateCodecs.scala
cngo-github: This is what I was able to figure out:

```
def run[F[_]: Files: Applicative](): Unit = {
    val generatedCodeDirectory = Path("modules/rosetta/src/main/scala/org/tessellation/rosetta/api/generated")
    Files[F].list(generatedCodeDirectory).filter(path => path.fileName.toString.endsWith(".scala")).foreach { entry =>
      val fileName = entry.fileName.toString.replaceFirst(".scala", "")
      val genString = s"""implicit lazy val ${fileName}Encoder = deriveConfiguredMagnoliaEncoder[${fileName}]
                         |implicit lazy val ${fileName}Decoder = deriveConfiguredMagnoliaDecoder[${fileName}]""".stripMargin
      println(genString)
      Applicative[F].pure(())
    }
  }
```

marcinwadon: @cngo-github using functional programming, we use effects to encapsulate the side-effects and delay their execution. for example: `println(genString)` is a side effect that will be executed immediately, but wrapping it into `IO(println(genString))` will create a declaration of run, so you can freely pass it further and execute on demand.

the whole codebase is running as a single "run" command, by chaining all the small sub-programs altogether.

that being said, you should chain (`flatMap`) all the operations.

`Applicative[F].pure(())` makes no sense because it does nothing.
`foreach` is imperative and it expects a unit type, and it expects a side-effect. instead of foreach loop, you should use fold or reduce.

anyway, it's not worth trying (other than learning purpose) because this code should be changed anyway to generate classes using template

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/search/SearchTransactionRequest.scala
cngo-github: Refined types and code base types have been added to the class as appropriate.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/search/SearchTransactionRequest.scala
```diff
@@ -0,0 +1,19 @@
...
+
+import eu.timepit.refined.types.numeric.{NonNegLong, PosLong}
+
+/** This is an intermediary object used to hold the values from the Rosetta request that is relevant to our system.
+  */
+
+case class SearchTransactionRequest(
+  operation: Operator,
+  address: Option[Address],
+  networkStatus: Option[OperationStatus],
```
marcinwadon, line 14: isn't `network status` and unchanged name?

modules/rosetta/src/main/scala/org/tessellation/rosetta/api/generated/dag/GenerateCodecs.scala
cngo-github: After discussing this with @ryle-goehausen , I'll be removing this file. I will work with @tbekas on Friday to automate this if we can do so relatively quickly. If not, we'll just leave this out of the code base.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/GenerateCodecs.scala
cngo-github: After discussing this with @ryle-goehausen , I'll be removing this file. I will work with @tbekas on Friday to automate this if we can do so relatively quickly. If not, we'll just leave this out of the code base.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/GenerateCodecs.scala
cngo-github: After discussing this with @ryle-goehausen , I'll be removing this file. I will work with @tbekas on Friday to automate this if we can do so relatively quickly. If not, we'll just leave this out of the code base.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/search/SearchTransactionRequest.scala
cngo-github: `networkStatus` is the same as `OperationStatus`. The `OperationStatus` class has supplanted the now deleted `NetworkStatus` class

marcinwadon: exactly, and you left the network status name as the class' field name. IMO it should be either just status or operation status, not network status.

cngo-github: I'll change it to `status`.

cngo-github:
I have added a `metabuild` declaration in the meta-project's `build.sbt` and incorporated the dependencies for the latest releases of `org.openapitools.openapi-generator-core` and `org.openapitools.openapi-generator-cli`. I also created the task `generateRosettaClasses` to generate Rosetta's model classes. They are being generated to the package `org.tessellation.rosetta.api`. I have also added the template needed for generation to `modules/rosetta/src/main/resources`. I added `modules/rosetta/src/main/scala/org/tessellation/rosetta/api` to the project's `.gitignore`. The package `org.tessellation.rosetta.api.dag` has been moved to `org.tessellation.rosetta.dag`.

Generate the Rosetta classes by running `sbt generateRosettaClasses` in the terminal.

Adding the custom `sbt` task and generator template allowed me to remove all the auto-generated Rosetta classes, `decoders.scala`, and `GenerateCodecs.scala`.

modules/rosetta/src/main/scala/org/tessellation/rosetta/api/generated/dag/GenerateCodecs.scala
cngo-github: The generation of the `Rosetta` classes has ben made a custom `sbt` task. This is no longer needed.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/decoders.scala
cngo-github: The generation of the `Rosetta` classes has ben made a custom `sbt` task. This is no longer needed.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/dag/GenerateCodecs.scala
cngo-github: The generation of the `Rosetta` classes has ben made a custom `sbt` task. This is no longer needed.

PR #544, opened 2023-01-20
Merkle tree implementation

marcinwadon:


kpudlik (approved):

modules/shared/src/main/scala/org/tessellation/merkletree/MerkleTree.scala
```diff
@@ -0,0 +1,125 @@
...
+          val lsib = pe.siblings.left.getOrElse(curr)
+          val rsib = pe.siblings.getOrElse(curr)
+          val hash = MerkleTree.hashIntermediate(lsib, rsib)
+          Option.when(hash == pe.target)(hash)
+        case (None, _) => None
+      }
+      .isDefined
+}
+
+case class MerkleTree(leafCount: Int, nodes: NonEmptyList[Hash]) {
```
kpudlik, line 26: How about making `leafCount` a `NonNegInt`?

kpudlik (changes requested):

modules/shared/src/main/scala/org/tessellation/merkletree/MerkleTree.scala
marcinwadon: let me check :+1:

modules/shared/src/main/scala/org/tessellation/merkletree/MerkleTree.scala
```diff
@@ -0,0 +1,125 @@
...
+
+      NonEmptyList.fromList(go()).map(Proof(_))
+    }
+}
+
+object MerkleTree {
+
+  implicit val eq: Eq[MerkleTree] = (x, y) => x.leafCount === y.leafCount && x.getRoot === y.getRoot
+
+  val leafPrefix: Byte = 0x00
```
kpudlik, line 79: can be `private`

modules/shared/src/main/scala/org/tessellation/merkletree/MerkleTree.scala
```diff
@@ -0,0 +1,125 @@
...
+  val intermediatePrefix: Byte = 0x01
```
kpudlik, line 80: can be `private`

modules/shared/src/main/scala/org/tessellation/merkletree/MerkleTree.scala
```diff
@@ -0,0 +1,125 @@
...
+
+  def hashLeaf(node: Hash): Hash =
```
kpudlik, line 82: can be package-private (probably)

modules/shared/src/main/scala/org/tessellation/merkletree/MerkleTree.scala
```diff
@@ -0,0 +1,125 @@
...
+    Hash.fromBytes(node.value.getBytes.prepended(leafPrefix))
+
+  def hashIntermediate(left: Hash, right: Hash): Hash =
```
kpudlik, line 85: can be package-private (probably)

kpudlik (changes requested):

kpudlik (dismissed):

kpudlik (dismissed):

modules/shared/src/main/scala/org/tessellation/merkletree/MerkleTree.scala
```diff
@@ -0,0 +1,128 @@
...
+        case (None, _) => None
+      }
+      .isDefined
+}
+
+case class MerkleTree(leafCount: NonNegInt, nodes: NonEmptyList[Hash]) {
+  def getRoot: Hash =
+    nodes.last
+
+  def findPath(node: Hash): Option[Proof] =
```
TheMMaciek, line 33: Aren't we looking for a path to a leaf rather than a node?

modules/shared/src/main/scala/org/tessellation/merkletree/MerkleTree.scala
```diff
@@ -0,0 +1,128 @@
...
+import cats.kernel.Eq
+import cats.syntax.eq._
+import cats.syntax.option._
+
+import org.tessellation.security.hash.Hash
+
+import eu.timepit.refined.cats._
+import eu.timepit.refined.types.numeric.NonNegInt
+
+case class ProofEntry(target: Hash, siblings: Either[Hash, Hash])
```
TheMMaciek, line 13: I think it would be beneficial to add a scala documentation for these classes/functions explaining the context and intended functionality - given that it is implemented with the base constructs (like e.g. fetching index from a list, left/right) which have some implicit meaning here.

marcinwadon: actually it would mean we have to explain how merkle trees should look like. does it make sense to just leave some url in the comment?

TheMMaciek, line 13: Regarding `ProofEntry` class maybe we should have the second parameter called `sibling` instead of `siblings` here given that there is only one value there (either left or right sibling).

modules/shared/src/main/scala/org/tessellation/merkletree/MerkleTree.scala
marcinwadon: when we create merkle tree, we provide `nodes`, so here is a `node` as well. I can change it to `element` (?)

modules/shared/src/main/scala/org/tessellation/merkletree/MerkleTree.scala
marcinwadon: :+1:

modules/shared/src/main/scala/org/tessellation/merkletree/MerkleTree.scala
marcinwadon: nvm. changed to `leaf`

modules/shared/src/main/scala/org/tessellation/merkletree/MerkleTree.scala
```diff
@@ -0,0 +1,128 @@
...
+
+case class Proof(entries: NonEmptyList[ProofEntry]) {
+  def verify(candidate: Hash): Boolean =
+    entries
+      .foldLeft(candidate.some.map(MerkleTree.hashLeaf)) {
+        case (Some(curr), pe) =>
+          val lsib = pe.siblings.left.getOrElse(curr)
+          val rsib = pe.siblings.getOrElse(curr)
```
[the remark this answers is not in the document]
marcinwadon: :+1:

modules/shared/src/main/scala/org/tessellation/merkletree/MerkleTree.scala
```diff
@@ -0,0 +1,130 @@
...
+      mt: MerkleTree = MerkleTree(leafCount, nodes),
+      levelLen: Int = nextLevelLen(leafCount.value),
+      levelStart: Int = leafCount.value,
+      prevLevelLen: Int = leafCount.value,
+      prevLevelStart: Int = 0
+    ): MerkleTree =
+      if (levelLen <= 0)
+        mt
+      else {
+        val newMt = (0 to levelLen - 1).foldLeft(mt) {
```
TheMMaciek, line 110: I think you can use `until` instead of `to levelLen - 1`

TheMMaciek (approved):
Looks good. Possibly I will try to make an improvement with less imperative implementation in another PR if I come up with one.

marcinwadon:
@TheMMaciek actually I was thinking to switch to more imperative implementation due to performance. functional programming is superb but take a look how most of libraries look under the hood. imperative programming, because in most cases it's just faster

PR #545, opened 2023-01-31
Added the models for the search endpoint

cngo-github:
__PR Category__: Feature

Added the intermediary models needed to hold the parameters for the Rosetta `search` endpoints ([link](https://www.rosetta-api.org/docs/SearchApi.html)).

### DIFF
--- a/modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/search/SearchTransactionRequest.scala
+++ b/modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/search/SearchTransactionRequest.scala
@@ -1,0 +1,20 @@
+package org.tessellation.rosetta.server.model.search
+
+import org.tessellation.schema.address.Address
+import org.tessellation.security.hash.Hash
+
+import eu.timepit.refined.types.numeric.{NonNegLong, PosLong}
+
+/** This is an intermediary object used to hold the values from the Rosetta request that is relevant to our system. It is used by the
+  * https://www.rosetta-api.org/docs/SearchApi.html endpoints.
+  */
+
+case class SearchTransactionRequest(
+  operation: Operator,
+  address: Option[Address],
+  status: Option[OperationStatus],
+  limit: Option[PosLong],
+  offset: Option[PosLong],
+  transactionHash: Option[Hash],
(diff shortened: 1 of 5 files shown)


modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/search/SearchTransactionRequest.scala
```diff
@@ -0,0 +1,20 @@
...
+  */
+
+case class SearchTransactionRequest(
+  operation: Operator,
+  address: Option[Address],
+  status: Option[OperationStatus],
+  limit: Option[PosLong],
+  offset: Option[PosLong],
+  transactionHash: Option[Hash],
+  maxBlock: Option[NonNegLong]
```
marcinwadon, line 19: What is the maxBlock? Which property of the block it represents?

marcinwadon (commented):
Overall looks good. One comment though

marcinwadon:
btw do we need a custom class for search api? Isn't it specified in the openapi thus can be autogenerated?

cngo-github:
> btw do we need a custom class for search api? Isn't it specified in the openapi thus can be autogenerated?

`SearchTransactionRequest` is not the same as the other auto-generated classes. It's purpose is to hold the parameters passed into the [`/search/transactions` endpoint](https://www.rosetta-api.org/docs/SearchApi.html#searchtransactions) and allow those parameters to be passed around in a compact way. It is not a part of the Rosetta specification, so I kept out of the `org.tessellation.rosetta.api` package.

modules/rosetta/src/main/scala/org/tessellation/rosetta/server/model/search/SearchTransactionRequest.scala
cngo-github: `maxBlock` is the `max_block` field in the request schema [here](https://www.rosetta-api.org/docs/SearchApi.html#searchtransactions). The endpoint searches for transactions inside of blocks (i.e., snapshots). `maxBlock` specifies the maximum height of the blocks to search.

marcinwadon (approved):

PR #547, opened 2023-01-31
Updated logic for joining consensus

tbekas:
Changes include:
- New states: WaitingForObserving, WaitingForReady
- Observation no longer means that we follow consensus logic locally
- Facilitators are copied from the previous consensus to the next
- Withdrawing from consensus can happen at any stage
- All consensus states (including not synchronized states yet) can be locked
- `ConsensusInfoRoutes` extracted from `ConsensusRoutes`


modules/core/src/main/scala/org/tessellation/domain/snapshot/programs/Download.scala
```diff
@@ -1,61 +1,30 @@
...
-              .flatMap { ordinal =>
-                consensus.manager.startObservingAfter(ordinal.next, peer)
-              }
-        }
-        .handleErrorWith { err =>
-          logger.error(err)(s"Error during download process") >>
-            err.raiseError[F, Unit]
-        }
-    }
+    nodeStorage.tryModifyState(NodeState.WaitingForDownload, NodeState.WaitingForObserving) >>
```
tbekas, line 28: This state transition is odd but I don't have a better idea for now on how to remove it without more extensive refactoring.

Currently, there's no download on L0 so download-related statuses don't make much sense. They will start to make sense once we have incremental snapshots and we will have to download either partial or full history.

Should I leave it as it is, refactor it now, or create a task for refactoring it in the future?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateRemover.scala
```diff
@@ -0,0 +1,71 @@
+package org.tessellation.sdk.infrastructure.consensus
+
+import cats.Show
+import cats.effect.Async
+import cats.kernel.Next
+import cats.syntax.all._
```
marcinwadon, line 6: in the whole project we try to import only required syntax, instead of all

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateRemover.scala
```diff
@@ -0,0 +1,71 @@
...
+  /** Withdraws from consensus.
+    */
+  def withdrawFromConsensus(key: Key): F[Unit]
+
+}
+
+object ConsensusStateRemover {
+  def make[F[
+    _
+  ]: Async, Event, Key: Show: Next: TypeTag: Encoder, Artifact](
```
marcinwadon, line 29: is it formatted? 🤔

tbekas: Yes, unfortunately, it is.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateRemover.scala
```diff
@@ -0,0 +1,71 @@
...
+import org.tessellation.sdk.domain.gossip.Gossip
+import org.tessellation.sdk.infrastructure.consensus.declaration.kind.{PeerDeclarationKind, _}
+import org.tessellation.sdk.infrastructure.consensus.message.ConsensusWithdrawPeerDeclaration
+
+import io.circe.Encoder
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+
+trait ConsensusStateRemover[F[_], Key, Artifact] {
+
+  /** Withdraws from consensus.
```
kpudlik, line 20: Imo this comment is redundant

kpudlik (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateRemover.scala
tbekas: I can change it in this file since it doesn't use much syntax, but I think we should be importing `cats.syntax.all._` when there are more than a few specific `cats.syntax` imports. Even cats do it internally. It simplifies development and solves potential conflicts in imports.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateRemover.scala
tbekas: True 👍

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateRemover.scala
marcinwadon: @tbekas what conflicts do you mean?

marcinwadon: > The reason for hand-optimizing your imports is that you limit the implicit scopes in your Scala files, reducing compilation time.

this argument convinces me to make this little effort and import only needed syntax and instances, instead of importing all everything, especially in such a large codebase we have

tbekas: I'm completely fine with trading the compiler's time for my own time. 

With hand optimizations, we need to know exactly what import has to be added for the syntax we want to use - which is problematic when something is rarely used, like `leftWiden` that is used in this file. If you don't know it you need to spend additional time searching for the right import. Also if someone isn't very proficient with cats API yet, it's nice to have a menu of methods available on hand.

>@tbekas what conflicts do you mean?

In the source of your quote, there's a next paragraph that explains it very well
https://blog.softwaremill.com/9-tips-about-using-cats-in-scala-you-might-want-to-know-e1bafd365f88

```
//we don't remember where `pure` is and
//we're trying to be smart
import cats.implicits._
import cats.instances.option._

"a".pure[Option]
//could not find implicit value for parameter F: cats.Applicative[Option]
```
>This happens because cats.implicits and cats.instances.option both extend cats.instances.OptionInstances. We’re essentially importing the implicit scope for that twice, and this confuses the compiler.

kpudlik (approved):

marcinwadon (approved):

tbekas:
@TheMMaciek feel free to add comments after the merge.

PR #551, opened 2023-02-03
Rosetta - Derive public key endpoint

marcinwadon:


build.sbt
```diff
@@ -309,18 +309,28 @@ lazy val dagShared = (project in file("modules/dag-shared"))
   )
 
 lazy val rosetta = (project in file("modules/rosetta"))
-  .dependsOn(shared % "compile->compile;test->test")
+  .enablePlugins(AshScriptPlugin)
+  .enablePlugins(JavaAppPackaging)
```
tbekas, line 313: Does this module actually use these plugins?

modules/shared/src/main/scala/org/tessellation/ext/derevo/snakeEncoder.scala
```diff
@@ -0,0 +1,18 @@
...
+object snakeEncoder extends Derivation[Encoder] with NewTypeDerivation[Encoder] {
+  type Typeclass[T] = Encoder[T]
+
+  def combine[T](caseClass: CaseClass[Typeclass, T]): Typeclass[T] =
+    semiauto.combine[T](caseClass)(Configuration.default.withSnakeCaseMemberNames)
+  def dispatch[T](sealedTrait: SealedTrait[Typeclass, T]): Typeclass[T] =
+    semiauto.dispatch[T](sealedTrait)(Configuration.default.withSnakeCaseMemberNames)
+
+  def instance[T]: Typeclass[T] = macro Magnolia.gen[T]
+}
```
tbekas, line 18: Wouldn't it be better to create only a configuration and then import it? I'm not sure if this works, but it looks simpler.

```scala
object jsonConfig {
 implicit val configuration:Configuration = Configuration.default.withSnakeCaseMemberNames
}
```

```scala
import derevo.derive
import derevo.circe.magnolia.{customizableDecoder, customizableEncoder}

import org.tessellation.rosetta.jsonConfig

@derive(customizableEncoder, customizableDecoder)
sealed trait SealedTrait
case class Foo(value: String)
```

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/error/RosettaError.scala
```diff
@@ -0,0 +1,54 @@
+package org.tessellation.rosetta.domain.error
+
+import cats.syntax.option._
+
+import derevo.circe.magnolia.encoder
+import derevo.derive
+import enumeratum.values._
+import io.circe.Encoder
+
+@derive(encoder)
```
tbekas, line 10: Regular `encoder` will be fine? I would suggest using in the whole scope of rosetta the same `encoder`s and `decoders`, even if the json look the same.

modules/shared/src/main/scala/org/tessellation/security/Hex.scala
```diff
@@ -52,6 +54,30 @@ object hex {
         }
       } yield pk
 
+    def toPublicKeyByEC[F[_]: Async: SecurityProvider]: F[PublicKey] =
```
tbekas, line 57: Is it going to be a different PublicKey than the one returned by `toPublicKey`?

build.sbt
marcinwadon: I tested something and left it. It's actually unused now

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/error/RosettaError.scala
marcinwadon: Good catch. The whole scope should use the same one

modules/shared/src/main/scala/org/tessellation/ext/derevo/snakeEncoder.scala
marcinwadon: It works but then you have to import the configuration everywhere and wrap case classes into objects. I did the same what customizableEncoder/Decoder does - just predefined the configuration and delegate encoder/decoder to magnolia.

marcinwadon: Oh, I got your point with importing the configuration. Indeed it's simpler. I'll change that!

modules/shared/src/main/scala/org/tessellation/security/Hex.scala
marcinwadon: It's encoded differently to hex, our method for decoding doesn't work (uses different Spec - X509 vs EC)

modules/rosetta/src/main/scala/org/tessellation/rosetta/http/routes/ConstructionRoutes.scala
```diff
@@ -0,0 +1,35 @@
...
+import org.tessellation.rosetta.domain.construction.ConstructionService
+import org.tessellation.rosetta.ext.http4s.refined._
+import org.tessellation.sdk.config.AppEnvironment
+
+import org.http4s.HttpRoutes
+import org.http4s.circe.CirceEntityCodec._
+import org.http4s.dsl.Http4sDsl
+import org.http4s.server.Router
+
+final case class ConstructionRoutes[F[_]: Async](
```
tbekas, line 15: I know that we already have routes as `case classes` but I think they shouldn't be `case classes` but regular classes or anonymous traits instead. `case class` implies some immutable data, but this is a service class with dependencies.

modules/shared/src/main/scala/org/tessellation/security/Hex.scala
tbekas: Ok, thanks for the explanation

tbekas (dismissed):
LGTM

tbekas (approved):

PR #555, opened 2023-02-06
Remove hylo and stackF usage from L1 consensus

TheMMaciek:
We need this change to avoid type erasure which is crucial for abstracting L1 functionality that will be reused in currency framework.

FYI the file `BlockConsensusCell` is not properly formatted right now - that's because the visibility of what has actually changed is much better with the additional indentation (caused by functions residing inside now removed nested objects). After I will get initial approval I will fix the formatting and re-request approval.


marcinwadon (approved):

tbekas (approved):
LGTM

PR #563, opened 2023-02-15
Rewards distribution update

tbekas:
* 5% for Stardust primary
* 5% for Stardust secondary
* 20% for soft staking
* 1% for testnet
* 65% for data pool
* 4% for facilitators


modules/core/src/main/scala/org/tessellation/config/types.scala
```diff
@@ -53,35 +50,22 @@ object types {
...
-    remainingWeight: NonNegLong = 18L
+  case class ProgramsDistributionConfig(
+    weights: Map[Address, Weight] = Map(
+      Address("DAGSTARDUSTCOLLECTIVEHZOIPHXZUBFGNXWJETZVSPAPAHMLXS") -> Weight(5L), // stardust tax primary
+      Address("DAG8VT7bxjs1XXBAzJGYJDaeyNxuThikHeUTp9XY") -> Weight(5L), // stardust tax secondary
+      Address("DAG77VVVRvdZiYxZ2hCtkHz68h85ApT5b2xzdTkn") -> Weight(20L), // soft staking
+      Address("DAG0qE5tkz6cMUD5M2dkqgfV4TQCzUUdAP5MFM9P") -> Weight(1L), // testnet
+      Address("DAG0Njmo6JZ3FhkLsipJSppepUHPuTXcSifARfvK") -> Weight(65L) // data pool
+    ),
+    remainingWeight: Weight = Weight(4L) // facilitators
```
kpudlik, line 64: What do you think about calculating it as `100L - sum(weights)`? We can also add a unit test which checks if all the weights sum up to `Weight(100L)`.

tbekas: Actually, that's a good idea. This test may have to be amended in the future because weights don't have to sum up to 100. It would be perfectly fine to have 3 weights of 1, so the sum would be 3 (not 100) and rewards would be distributed like this `{ 1/3, 1/3, 1/3 }`. But this test would help anyway if someone would unintentionally change the sum to different than 100, so I think it's worth adding it.

kpudlik: >This test may have to be amended in the future because weights don't have to sum up to 100

Makes sense. So maybe just checking if `<= 100`?

tbekas: I could also add the `exptectedWeightSum` to the config and validate it during the application start. What do you think about it? 🤔

kpudlik: Someone can still put wrong value in `expectedWeightSum` 🤔 . But assuming that expected one is set correctly then it is still better than nothing so fine for me. I just can't make my mind if `expectedWeightSum` should be in config or in unit test.

tbekas: I've added verification for `expextedWeightSum` only in the unit test for now, since rewards configuration isn't dynamically loaded

kpudlik (dismissed):

kpudlik (dismissed):

modules/core/src/main/scala/org/tessellation/config/types.scala
marcinwadon: Unit test is fine for me too

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -67,28 +64,22 @@ object Rewards {
...
-          .run(amount)
-          .flatTap(validateState(amount))
-          .map(toTransactions)
-      }
+      ): F[SortedSet[RewardTransaction]] =
+        Random.scalaUtilRandomSeedLong(epochProgress.coerce).flatMap { random =>
+          def eitherToF[A](either: Either[ArithmeticException, A]): F[A] = either.liftTo[F]
+
+          val allRewardsState = for {
+            programRewards <- programsDistributor.distribute().mapK[F](lift(eitherToF))
```
marcinwadon, line 72: is `eitherToF` used once only? won't `lift(_.liftTo[F])` work?

marcinwadon (dismissed):

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
tbekas: `lift(_.liftTo[F])` won't compile. However, I see that `liftFunction(_.liftTo[F])` does compile. Fixing it.

marcinwadon (approved):
:+1:

kpudlik (approved):

PR #566, opened 2023-02-20
Added Rosetta's construction parse endpoint.

cngo-github:
__PR Category__: Feature

Added Rosetta's [`construction/parse` endpoint](https://www.rosetta-api.org/docs/ConstructionApi.html#constructionparse).

### DIFF
--- a/modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/Currency.scala
+++ b/modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/Currency.scala
@@ -2,15 +2,18 @@ package org.tessellation.rosetta.domain
 
 import org.tessellation.ext.derevo.magnoliaCustomizable.snakeCaseConfiguration
 
+import derevo.cats.eqv
 import derevo.circe.magnolia._
 import derevo.derive
 import eu.timepit.refined.api.Refined
+import eu.timepit.refined.auto._
+import eu.timepit.refined.cats._
 import eu.timepit.refined.generic.Equal
 import io.circe.refined._
 import io.estatico.newtype.macros.newtype
 
 object currency {
-  @derive(customizableDecoder, customizableEncoder)
+  @derive(eqv, customizableDecoder, customizableEncoder)
   case class Currency(
     symbol: CurrencySymbol,
     decimals: CurrencyDecimal
@@ -18,13 +21,15 @@ object currency {
 
   type CurrencySymbolRefined = String Refined Equal["DAG"]
 
-  @derive(decoder, encoder)
+  @derive(eqv, decoder, encoder)
   @newtype
   case class CurrencySymbol(value: CurrencySymbolRefined)
 
   type CurrencyDecimalRefined = Long Refined Equal[8L]
 
-  @derive(decoder, encoder)
+  @derive(eqv, decoder, encoder)
   @newtype
   case class CurrencyDecimal(value: CurrencyDecimalRefined)
+
+  val DAG: Currency = Currency(CurrencySymbol("DAG"), CurrencyDecimal(8L))
 }
--- a/modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/api/construction/ConstructionParse.scala
+++ b/modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/api/construction/ConstructionParse.scala
@@ -1,0 +1,28 @@
+package org.tessellation.rosetta.domain.api.construction
+
+import cats.data.NonEmptyList
+
+import org.tessellation.ext.derevo.magnoliaCustomizable.snakeCaseConfiguration
+import org.tessellation.rosetta.domain.operation.Operation
+import org.tessellation.rosetta.domain.{AccountIdentifier, NetworkIdentifier}
+import org.tessellation.security.hex.Hex
+
+import derevo.circe.magnolia.{customizableDecoder, customizableEncoder}
+import derevo.derive
+
+case object ConstructionParse {
+  @derive(customizableDecoder)
+  case class Request(networkIdentifier: NetworkIdentifier, signed: Boolean, transaction: Hex)
+
+  @derive(customizableEncoder)
+  case class Response(operations: List[Operation], accountIdentifierSigners: Option[NonEmptyList[AccountIdentifier]])
+
+  object Response {
+    def fromParseResult(result: ParseResult) = Response(result.operations.toList, result.accountIdentifierSigners)
+  }
+
+  case class ParseResult(
+    operations: NonEmptyList[Operation],
+    accountIdentifierSigners: Option[NonEmptyList[AccountIdentifier]]
+  )
+}
--- a/modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
+++ b/modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
@@ -6,21 +6,31 @@ import cats.syntax.applicativeError._
 import cats.syntax.either._
 import cats.syntax.flatMap._
 import cats.syntax.functor._
+import cats.syntax.option._
 
 import org.tessellation.kryo.KryoSerializer
-import org.tessellation.rosetta.domain._
+import org.tessellation.rosetta.domain.amount.{Amount, AmountValue}
+import org.tessellation.rosetta.domain.api.construction.ConstructionParse
+import org.tessellation.rosetta.domain.currency.DAG
 import org.tessellation.rosetta.domain.error.{ConstructionError, InvalidPublicKey, MalformedTransaction}
-import org.tessellation.rosetta.domain.operation.Operation
+import org.tessellation.rosetta.domain.operation._
+import org.tessellation.rosetta.domain.{AccountIdentifier, RosettaPublicKey, TransactionIdentifier}
 import org.tessellation.schema.transaction.Transaction
 import org.tessellation.security.SecurityProvider
 import org.tessellation.security.hex.Hex
 import org.tessellation.security.key.ops._
 import org.tessellation.security.signature.Signed
 
+import eu.timepit.refined.auto._
+import eu.timepit.refined.boolean.Not
+import eu.timepit.refined.generic.Equal
+import eu.timepit.refined.refineV
+
 trait ConstructionService[F[_]] {
   def derive(publicKey: RosettaPublicKey): EitherT[F, ConstructionError, AccountIdentifier]
   def getAccountIdentifiers(operations: List[Operation]): Option[NonEmptyList[AccountIdentifier]]
   def getTransactionIdentifier(hex: Hex): EitherT[F, ConstructionError, TransactionIdentifier]
+  def parseTransaction(hex: Hex, isSigned: Boolean): EitherT[F, ConstructionError, ConstructionParse.ParseResult]
 }
 
 object ConstructionService {
@@ -48,5 +58,53 @@ object ConstructionService {
 
       NonEmptyList.fromList(accountIdentifiers)
     }
+
+    def transactionToOperations(transaction: Transaction): Either[ConstructionError, NonEmptyList[Operation]] = {
+      type Refinement = Not[Equal[0L]]
+
+      val negativeTransfer = refineV[Refinement](-transaction.amount.value.value).map(
+        (transaction.source, _, OperationIndex(0L))
+      )
+      val positiveTransfer = refineV[Refinement](+transaction.amount.value.value).map(
+        (transaction.destination, _, OperationIndex(1L))
+      )
+
+      NonEmptyList.of(negativeTransfer, positiveTransfer).traverse {
+        case (Right((address, amount, operationIndex))) =>
+          Operation(
+            OperationIdentifier(operationIndex),
+            none,
+            OperationType.Transfer,
+            none,
+            AccountIdentifier(address, none).some,
+            Amount(AmountValue(amount), DAG).some
+          ).asRight[ConstructionError]
+        case _ => Left(MalformedTransaction)
+      }
+    }
+
+    def parseSignedTransaction(hex: Hex): EitherT[F, ConstructionError, ConstructionParse.ParseResult] = {
+      val result = for {
+        signedTransaction <- KryoSerializer[F].deserialize[Signed[Transaction]](hex.toBytes).toEitherT
+        operations <- transactionToOperations(signedTransaction).toEitherT
+        proofs <- signedTransaction.proofs.toNonEmptyList.traverse(_.id.hex.toPublicKey).attemptT
+        accountIds = proofs.map(_.toAddress).map(AccountIdentifier(_, none))
+      } yield ConstructionParse.ParseResult(operations, accountIds.some)
+
+      result.leftMap(_ => MalformedTransaction)
+    }
+
+    def parseUnsignedTransaction(hex: Hex): EitherT[F, ConstructionError, ConstructionParse.ParseResult] =
+      for {
+        unsignedTransaction <- KryoSerializer[F].deserialize[Transaction](hex.toBytes).toEitherT.leftMap(_ => MalformedTransaction)
+        operations <- transactionToOperations(unsignedTransaction).toEitherT
+      } yield ConstructionParse.ParseResult(operations, none)
+
+    def parseTransaction(hex: Hex, isSigned: Boolean): EitherT[F, ConstructionError, ConstructionParse.ParseResult] =
+      if (isSigned) {
+        parseSignedTransaction(hex)
+      } else {
+        parseUnsignedTransaction(hex)
+      }
   }
 }
--- a/modules/rosetta/src/main/scala/org/tessellation/rosetta/http/routes/ConstructionRoutes.scala
+++ b/modules/rosetta/src/main/scala/org/tessellation/rosetta/http/routes/ConstructionRoutes.scala
@@ -2,7 +2,7 @@ package org.tessellation.rosetta.http.routes
 
 import cats.effect.Async
 
-import org.tessellation.rosetta.domain.api.construction.{ConstructionDerive, ConstructionHash, ConstructionPreprocess}
+import org.tessellation.rosetta.domain.api.construction._
 import org.tessellation.rosetta.domain.construction.ConstructionService
 import org.tessellation.rosetta.ext.http4s.refined._
 import org.tessellation.sdk.config.AppEnvironment
@@ -44,6 +44,19 @@ final class ConstructionRoutes[F[_]: Async](
 
         Ok(ConstructionPreprocess.Response(accountIdentifiers)).handleUnknownError
       }
+
+    case req @ POST -> Root / "parse" =>
+      req
+        .decodeRosettaWithNetworkValidation[ConstructionParse.Request](appEnvironment, _.networkIdentifier) { parseReq =>
+          constructionService
+            .parseTransaction(parseReq.transaction, parseReq.signed)
+            .bimap(
+              _.toRosettaError,
+              ConstructionParse.Response.fromParseResult
+            )
+            .asRosettaResponse
+        }
+        .handleUnknownError
   }
 
   val routes: HttpRoutes[F] = Router(
--- a/modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/construction/ConstructionServiceSuite.scala
+++ b/modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/construction/ConstructionServiceSuite.scala
@@ -1,5 +1,6 @@
 package org.tessellation.rosetta.domain.construction
 
+import cats.data.EitherT
 import cats.effect.{IO, Resource}
 import cats.syntax.contravariantSemigroupal._
 import cats.syntax.either._
@@ -11,18 +12,20 @@ import org.tessellation.keytool.KeyPairGenerator
 import org.tessellation.kryo.KryoSerializer
 import org.tessellation.rosetta.domain._
 import org.tessellation.rosetta.domain.amount.{Amount, AmountValue}
-import org.tessellation.rosetta.domain.currency.{Currency, CurrencyDecimal, CurrencySymbol}
-import org.tessellation.rosetta.domain.error.{InvalidPublicKey, MalformedTransaction}
+import org.tessellation.rosetta.domain.api.construction.ConstructionParse
+import org.tessellation.rosetta.domain.error.{ConstructionError, InvalidPublicKey, MalformedTransaction}
 import org.tessellation.rosetta.domain.operation.OperationType.Transfer
 import org.tessellation.rosetta.domain.operation.{Operation, OperationIdentifier, OperationIndex}
 import org.tessellation.schema.address.Address
 import org.tessellation.schema.generators.addressGen
-import org.tessellation.security.SecurityProvider
+import org.tessellation.schema.transaction.Transaction
 import org.tessellation.security.hex.Hex
 import org.tessellation.security.key.ops.PublicKeyOps
+import org.tessellation.security.{Hashed, SecurityProvider}
 import org.tessellation.shared.sharedKryoRegistrar
 
 import eu.timepit.refined.auto._
+import eu.timepit.refined.cats._
 import eu.timepit.refined.types.all.PosInt
 import weaver._
 import weaver.scalacheck.Checkers
@@ -35,6 +38,50 @@ object ConstructionServiceSuite extends MutableIOSuite with Checkers with Transa
     KryoSerializer
       .forAsync[IO](sharedKryoRegistrar)
       .flatMap(kryo => SecurityProvider.forAsync[IO].map(securityProvider => (securityProvider, kryo)))
+
+  def generateTestTransactions(wantSignedTransaction: Boolean)(implicit S: SecurityProvider[IO], K: KryoSerializer[IO]) = {
+    val getSignedTransaction = (transaction: Hashed[Transaction]) => transaction.signed
+    val getUnsignedTransaction = (transaction: Hashed[Transaction]) => transaction.signed.value
+
+    def txs(transactionConverter: Hashed[Transaction] => AnyRef) =
+      (KeyPairGenerator.makeKeyPair[IO], KeyPairGenerator.makeKeyPair[IO]).tupled.flatMap {
+        case (srcKey, dstKey) =>
+          val srcAddress = srcKey.getPublic.toAddress
+          val dstAddress = dstKey.getPublic.toAddress
+          val txCount = PosInt(100)
+
+          generateTransactions(srcAddress, srcKey, dstAddress, txCount)
+            .flatMap(_.traverse { hashedTransaction =>
+              KryoSerializer[F]
+                .serialize(transactionConverter(hashedTransaction))
+                .map(Hex.fromBytes(_))
+                .map((_, hashedTransaction.hash, hashedTransaction))
+                .liftTo[IO]
+            })
+      }
+
+    if (wantSignedTransaction) {
+      txs(getSignedTransaction)
+    } else {
+      txs(getUnsignedTransaction)
+    }
+  }
+
+  def testParseTransactions(
+    isSignedTransaction: Boolean,
+    testCaseCallbackHandler: (EitherT[F, ConstructionError, ConstructionParse.ParseResult], Hashed[Transaction]) => F[Expectations]
+  ) = sharedResource.use { res =>
+    implicit val (sp, k) = res
+
+    val cs = ConstructionService.make[IO]()
+
+    generateTestTransactions(wantSignedTransaction = isSignedTransaction)
+      .flatMap(_.traverse {
+        case (hex, _, transaction) =>
+          testCaseCallbackHandler(cs.parseTransaction(hex, isSignedTransaction), transaction)
+      })
+      .map(_.fold)
+  }
 
   test("derives public key") { res =>
     implicit val (sp, k) = res
@@ -72,28 +119,12 @@ object ConstructionServiceSuite extends MutableIOSuite with Checkers with Transa
 
     val cs = ConstructionService.make[IO]()
 
-    def txs = (KeyPairGenerator.makeKeyPair[IO], KeyPairGenerator.makeKeyPair[IO]).tupled.flatMap {
-      case (srcKey, dstKey) =>
-        val srcAddress = srcKey.getPublic.toAddress
-        val dstAddress = dstKey.getPublic.toAddress
-        val txCount = PosInt(100)
-
-        generateTransactions(srcAddress, srcKey, dstAddress, txCount)
-          .flatMap(_.traverse { hashedTransaction =>
-            KryoSerializer[F]
-              .serialize(hashedTransaction.signed)
-              .map(Hex.fromBytes(_))
-              .map((_, hashedTransaction.hash))
-              .liftTo[IO]
-          })
-    }
-
-    txs
+    generateTestTransactions(wantSignedTransaction = true)
       .flatMap(_.traverse {
-        case (hex, hash) =>
+        case (hex, hash, _) =>
           cs.getTransactionIdentifier(hex)
             .rethrowT
-            .map(expect.eql(_, TransactionIdentifier(hash)))
+            .map(expect.eql(TransactionIdentifier(hash), _))
       })
       .map(_.fold)
   }
@@ -126,7 +157,7 @@ object ConstructionServiceSuite extends MutableIOSuite with Checkers with Transa
         Transfer,
         none,
         AccountIdentifier(address, none).some,
-        Amount(AmountValue(-1L), Currency(CurrencySymbol("DAG"), CurrencyDecimal(8L))).some
+        Amount(AmountValue(-1L), currency.Currency(currency.CurrencySymbol("DAG"), currency.CurrencyDecimal(8L))).some
       )
 
       val result = cs
@@ -149,11 +180,121 @@ object ConstructionServiceSuite extends MutableIOSuite with Checkers with Transa
         Transfer,
         none,
         AccountIdentifier(address, none).some,
-        Amount(AmountValue(1L), Currency(CurrencySymbol("DAG"), CurrencyDecimal(8L))).some
+        Amount(AmountValue(1L), currency.Currency(currency.CurrencySymbol("DAG"), currency.CurrencyDecimal(8L))).some
       )
 
       val accountIdentifiers = cs.getAccountIdentifiers(List(operation))
       expect.eql(none, accountIdentifiers)
     }
+  }
+
+  test("returns operations for a valid signed transaction hex") { res =>
+    implicit val (sp, k) = res
+
+    val cs = ConstructionService.make[IO]()
+
+    generateTestTransactions(wantSignedTransaction = true)
+      .flatMap(_.traverse {
+        case (hex, _, _) =>
+          cs.parseTransaction(hex, true)
+            .rethrowT
+            .map(result => expect.eql(Amount(AmountValue(-1L), currency.DAG).some, result.operations.head.amount))
+      })
+      .map(_.fold)
+  }
+
+  test("returns operations for a valid unsigned transaction hex") { res =>
+    implicit val (sp, k) = res
+
+    val cs = ConstructionService.make[IO]()
+
+    generateTestTransactions(wantSignedTransaction = false)
+      .flatMap(_.traverse {
+        case (hex, _, _) =>
+          cs.parseTransaction(hex, false)
+            .rethrowT
+            .map(result => expect.eql(Amount(AmountValue(-1L), currency.DAG).some, result.operations.head.amount))
+      })
+      .map(_.fold)
+  }
+
+  test("parse signed transaction returns one positive and one negative operation only") {
+    val testCase = (parseResult: EitherT[F, ConstructionError, ConstructionParse.ParseResult], _: Hashed[Transaction]) =>
+      parseResult.rethrowT
+        .map(result => expect.eql(2, result.operations.length))
+
+    testParseTransactions(isSignedTransaction = true, testCaseCallbackHandler = testCase)
+  }
+
+  test("parse signed transaction returns a negative operation that takes from source address") {
+    val testCase = (parseResult: EitherT[F, ConstructionError, ConstructionParse.ParseResult], transaction: Hashed[Transaction]) =>
+      parseResult.rethrowT
+        .map(result => expect.eql(AccountIdentifier(transaction.source, none).some, result.operations.head.account))
+
+    testParseTransactions(isSignedTransaction = true, testCaseCallbackHandler = testCase)
+  }
+
+  test("parse signed transaction returns a negative operation with a negative amount") {
+    val testCase = (parseResult: EitherT[F, ConstructionError, ConstructionParse.ParseResult], _: Hashed[Transaction]) =>
+      parseResult.rethrowT
+        .map(result => expect.eql(Amount(AmountValue(-1L), currency.DAG).some, result.operations.head.amount))
+
+    testParseTransactions(isSignedTransaction = true, testCaseCallbackHandler = testCase)
+  }
+
+  test("parse signed transaction returns a positive operation that gives to the destination address") {
+    val testCase = (parseResult: EitherT[F, ConstructionError, ConstructionParse.ParseResult], transaction: Hashed[Transaction]) =>
+      parseResult.rethrowT
+        .map(result => expect.eql(AccountIdentifier(transaction.destination, none).some, result.operations.tail.head.account))
+
+    testParseTransactions(isSignedTransaction = true, testCaseCallbackHandler = testCase)
+  }
+
+  test("parse signed transaction returns a positive operation with a positive amount") {
+    val testCase = (parseResult: EitherT[F, ConstructionError, ConstructionParse.ParseResult], _: Hashed[Transaction]) =>
+      parseResult.rethrowT
+        .map(result => expect.eql(Amount(AmountValue(1L), currency.DAG).some, result.operations.tail.head.amount))
+
+    testParseTransactions(isSignedTransaction = true, testCaseCallbackHandler = testCase)
+  }
+
+  test("parse unsigned transaction returns one positive and one negative operation only") {
+    val testCase = (parseResult: EitherT[F, ConstructionError, ConstructionParse.ParseResult], _: Hashed[Transaction]) =>
+      parseResult.rethrowT
+        .map(result => expect.eql(2, result.operations.length))
+
+    testParseTransactions(isSignedTransaction = false, testCaseCallbackHandler = testCase)
+  }
+
+  test("parse unsigned transaction returns a negative operation that takes from source address") {
+    val testCase = (parseResult: EitherT[F, ConstructionError, ConstructionParse.ParseResult], transaction: Hashed[Transaction]) =>
+      parseResult.rethrowT
+        .map(result => expect.eql(AccountIdentifier(transaction.source, none).some, result.operations.head.account))
+
+    testParseTransactions(isSignedTransaction = false, testCaseCallbackHandler = testCase)
+  }
+
+  test("parse unsigned transaction returns a negative operation with a negative amount") {
+    val testCase = (parseResult: EitherT[F, ConstructionError, ConstructionParse.ParseResult], _: Hashed[Transaction]) =>
+      parseResult.rethrowT
+        .map(result => expect.eql(Amount(AmountValue(-1L), currency.DAG).some, result.operations.head.amount))
+
+    testParseTransactions(isSignedTransaction = false, testCaseCallbackHandler = testCase)
+  }
+
+  test("parse unsigned transaction returns a positive operation that gives to the destination address") {
+    val testCase = (parseResult: EitherT[F, ConstructionError, ConstructionParse.ParseResult], transaction: Hashed[Transaction]) =>
+      parseResult.rethrowT
+        .map(result => expect.eql(AccountIdentifier(transaction.destination, none).some, result.operations.tail.head.account))
+
+    testParseTransactions(isSignedTransaction = false, testCaseCallbackHandler = testCase)
+  }
+
+  test("parse unsigned transaction returns a positive operation with a positive amount") {
+    val testCase = (parseResult: EitherT[F, ConstructionError, ConstructionParse.ParseResult], _: Hashed[Transaction]) =>
+      parseResult.rethrowT
+        .map(result => expect.eql(Amount(AmountValue(1L), currency.DAG).some, result.operations.tail.head.amount))
+
+    testParseTransactions(isSignedTransaction = false, testCaseCallbackHandler = testCase)
   }
 }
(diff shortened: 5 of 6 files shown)


cngo-github:
I am implementing this [code](https://github.com/cngo-github/tessellation/blob/20221213-rosetta_test/modules/dag-l1/src/main/scala/org/tessellation/dag/l1/rosetta/RosettaRoutes.scala#L400-L464).

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
```diff
@@ -6,21 +6,27 @@ import cats.syntax.applicativeError._
...
 import org.tessellation.security.hex.Hex
 import org.tessellation.security.key.ops._
 import org.tessellation.security.signature.Signed
+import eu.timepit.refined.auto._
 
 trait ConstructionService[F[_]] {
   def derive(publicKey: RosettaPublicKey): EitherT[F, ConstructionError, AccountIdentifier]
   def getAccountIdentifiers(operations: List[Operation]): Option[NonEmptyList[AccountIdentifier]]
   def getTransactionIdentifier(hex: Hex): EitherT[F, ConstructionError, TransactionIdentifier]
+  def parseSignedTransaction(hex: Hex): EitherT[F, ConstructionError, (Option[NonEmptyList[Operation]], Option[NonEmptyList[Address]], Option[NonEmptyList[AccountIdentifier]])]
```
cngo-github, line 29: I don't like this return type: `EitherT[F, ConstructionError, (Option[NonEmptyList[Operation]], Option[NonEmptyList[Address]], Option[NonEmptyList[AccountIdentifier]])]` and recommendations are appreciated.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
```diff
@@ -48,5 +54,45 @@ object ConstructionService {
...
+                Amount(AmountValue(signedTransaction.value.amount.value.value), DAGCurrency)
+              )
+
+              val negativeOperation = Operation(
+                OperationIdentifier(OperationIndex(0L)),
+                none,
+                OperationType.Transfer,
+                none,
+                AccountIdentifier(signedTransaction.value.source, none).some,
+                Amount(AmountValue(-1L * signedTransaction.value.amount.value.value), DAGCurrency)
```
cngo-github, line 85: The only way I know to refine a value at runtime is with `refineV`, but that returns an `Either[...]`. What is the cleanest way to do this?

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
```diff
@@ -48,5 +54,45 @@ object ConstructionService {
...
+                first.appended(second)
+            })
+            .map(NonEmptyList.fromList(_)).map { maybeAddresses => 
+              val positiveOperation = Operation(
+                OperationIdentifier(OperationIndex(1L)),
+                none,
+                OperationType.Transfer,
+                none,
+                AccountIdentifier(signedTransaction.value.destination, none).some,
+                Amount(AmountValue(signedTransaction.value.amount.value.value), DAGCurrency)
```
cngo-github, line 76: The only way I know to refine a value at runtime is with `refineV`, but that returns an `Either[...]`. What is the cleanest way to do this?

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
```diff
@@ -48,5 +54,45 @@ object ConstructionService {
...
+              case (first, second) =>
+                first.appended(second)
+            })
+            .map(NonEmptyList.fromList(_)).map { maybeAddresses => 
+              val positiveOperation = Operation(
+                OperationIdentifier(OperationIndex(1L)),
+                none,
+                OperationType.Transfer,
+                none,
+                AccountIdentifier(signedTransaction.value.destination, none).some,
```
marcinwadon, line 75: it would be enough to write `signedTransaction.destination` as `Signed` has auto unwrap

marcinwadon (changes requested):

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/error/RosettaError.scala
```diff
@@ -42,6 +42,13 @@ object RosettaError extends IntEnum[RosettaError] with RosettaErrorEncoder {
         retriable = false,
         description = Some("Unable to decode the transaction.")
       )
+  case object DeserializationFailure
+      extends RosettaError(
+        value = 14,
```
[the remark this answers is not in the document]
cngo-github: Changed.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/error/RosettaError.scala
```diff
@@ -42,6 +42,13 @@ object RosettaError extends IntEnum[RosettaError] with RosettaErrorEncoder {
         retriable = false,
         description = Some("Unable to decode the transaction.")
       )
+  case object DeserializationFailure
```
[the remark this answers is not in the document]
cngo-github: Changed

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/Currency.scala
```diff
@@ -27,4 +28,6 @@ object currency {
   @derive(decoder, encoder)
   @newtype
   case class CurrencyDecimal(value: CurrencyDecimalRefined)
+
+  val DAGCurrency: Currency = Currency(CurrencySymbol("DAG"), CurrencyDecimal(8L))
```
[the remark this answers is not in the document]
cngo-github: Changed.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/api/construction/ConstructionParse.scala
```diff
@@ -0,0 +1,23 @@
...
+import cats.data.NonEmptyList
+
+case object ConstructionParse {
+  @derive(customizableDecoder)
+  case class Request(networkIdentifier: NetworkIdentifier, signed: Boolean, transaction: Hex)
+
+  @derive(customizableEncoder)
+  case class Response(
+                       operations: List[Operation],
+                       signers: Option[NonEmptyList[Address]],
```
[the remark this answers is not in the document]
cngo-github: I'll double check the specifications.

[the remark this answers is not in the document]
cngo-github: Since this field is not needed anymore, I will have to revisit the implementation.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
```diff
@@ -48,5 +54,45 @@ object ConstructionService {
...
+              val negativeOperation = Operation(
+                OperationIdentifier(OperationIndex(0L)),
+                none,
+                OperationType.Transfer,
+                none,
+                AccountIdentifier(signedTransaction.value.source, none).some,
+                Amount(AmountValue(-1L * signedTransaction.value.amount.value.value), DAGCurrency)
+              )
+
+              val operations = NonEmptyList.fromList(List(positiveOperation, negativeOperation))
```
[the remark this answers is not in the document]
cngo-github: Oversight.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
```diff
@@ -48,5 +54,45 @@ object ConstructionService {
...
+      KryoSerializer[F]
+        .deserialize[Signed[Transaction]](hex.toBytes)
+        .liftTo[F]
+        .flatMap { signedTransaction =>
+          signedTransaction.proofs.toNonEmptyList
+            .traverse(_.id.hex.toPublicKey.map(_.toAddress))
+            .map(_.foldLeft(List.empty: List[Address]) {
+              case (first, second) =>
+                first.appended(second)
+            })
```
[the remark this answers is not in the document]
cngo-github: I reworked the overall logic.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
```diff
@@ -48,5 +54,45 @@ object ConstructionService {
 
       NonEmptyList.fromList(accountIdentifiers)
     }
+
+    def parseSignedTransaction(hex: Hex): EitherT[F, ConstructionError, (Option[NonEmptyList[Operation]], Option[NonEmptyList[Address]], Option[NonEmptyList[AccountIdentifier]])] =
```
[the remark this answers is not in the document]
cngo-github: Added.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
cngo-github: Changed.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
```diff
@@ -48,5 +54,45 @@ object ConstructionService {
...
+      KryoSerializer[F]
+        .deserialize[Signed[Transaction]](hex.toBytes)
```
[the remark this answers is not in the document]
cngo-github: I reworked the overall logic.

cngo-github:
Here is where I am stuck.

I am trying to make

https://github.com/cngo-github/tessellation/blob/20230215-rosetta_parse/modules/rosetta/src/main/scala/org/tessellation/rosetta/http/routes/ConstructionRoutes.scala#L48-L58

work with these two:

https://github.com/cngo-github/tessellation/blob/20230215-rosetta_parse/modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala#L95-L114

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
```diff
@@ -48,5 +58,58 @@ object ConstructionService {
...
+      }
+    }.map(_.left.map(_ => MalformedTransaction))
+  }.map(_.left.map(_ => MalformedTransaction).flatten)
+
+  def parseUnsignedTransaction(hex: Hex): Either[MalformedTransaction.type, ConstructionParse.Response] = 
+    KryoSerializer[F].deserialize[Transaction](hex.toBytes).traverse { transaction =>
+    transactionToOperations(transaction).map { operationsList =>
+        ConstructionParse.Response(operationsList.toList, none)
+      }
+    }.map(_.left.map(_ => MalformedTransaction)).left.map(_ => MalformedTransaction).flatten
```
kpudlik, line 113:
```scala
      KryoSerializer[F]
        .deserialize[Transaction](hex.toBytes)
        .flatMap(transactionToOperations)
        .map(ops => ConstructionParse.Response(ops.toList, none))
        .leftMap(_ => MalformedTransaction)
```

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
```diff
@@ -48,5 +58,58 @@ object ConstructionService {
...
+        val proofAccountIdentifiers = proofs.map(_.toAddress).map { address =>
+          AccountIdentifier(address, none)
+        }
+        
+        ConstructionParse.Response(operationsList.toList, proofAccountIdentifiers.toList.some)
+      }
+    }.map(_.left.map(_ => MalformedTransaction))
+  }.map(_.left.map(_ => MalformedTransaction).flatten)
+
+  def parseUnsignedTransaction(hex: Hex): Either[MalformedTransaction.type, ConstructionParse.Response] =
```
kpudlik, line 108: Here you have `MalformedTransaction.type` but in trait you have `ParseFailure.type`. Which one is valid? Because it won't compile because it wants `ParseFailure` but you pass `MalformedTransaction`

kpudlik (changes requested):

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/error/RosettaError.scala
```diff
@@ -42,6 +42,13 @@ object RosettaError extends IntEnum[RosettaError] with RosettaErrorEncoder {
         retriable = false,
         description = Some("Unable to decode the transaction.")
       )
+  case object ParseFailure
+      extends RosettaError(
+        value = 7,
+        message = "Parse failure",
+        retriable = false,
+        description = Some("Unable to deserialize the class to its kryo type.")
+      )
```
marcinwadon, line 51: What's the difference between this and MalformedTransaction?

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
marcinwadon: yep, but the service should not return an API response

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
cngo-github: @marcinwadon requested reusing `MalformedTransaction`. I forgot to update the trait. That has been rectified.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/error/RosettaError.scala
cngo-github: I will use `MalformTransaction` instead. I forgot to remove this.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
cngo-github: Changed the code as pointed out by @kpudlik . Regarding @marcinwadon 's comment, I can change the return object to a case class made for that purpose, but I still have no idea how to convert `F[Either[MalformedTransaction.type, ConstructionParse.Response]]` and `Either[MalformedTransaction.type, ConstructionParse.Response]` into `F[Response[F]]`.

modules/rosetta/src/main/scala/org/tessellation/rosetta/http/routes/ConstructionRoutes.scala
```diff
@@ -44,6 +44,17 @@ final class ConstructionRoutes[F[_]: Async](
...
+
+    case req @ POST -> Root / "parse" =>
+      req.decodeRosettaWithNetworkValidation[ConstructionParse.Request](appEnvironment, _.networkIdentifier) { parseReq =>
+        if(parseReq.signed == true) {
+          val response = constructionService.parseSignedTransaction(parseReq.transaction)
+          val test = response.map(_.bimap(_.toRosettaError, _))
+        } else {
+          val response = constructionService.parseUnsignedTransaction(parseReq.transaction)
+          val test = response.left.map(_.toRosettaError)
+        }
```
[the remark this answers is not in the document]
cngo-github: I will need guidance on how to convert `F[Either[MalformedTransaction.type, ConstructionParse.Response]]` and `Either[MalformedTransaction.type, ConstructionParse.Response]` into `F[Response[F]]`. Once I understand that, I will address this comment.

cngo-github:
I am unable to make clean positive and negative tests because I cannot add `eqv` to the `@derive` of [`AmountValue`](https://github.com/cngo-github/tessellation/blob/20230215-rosetta_parse/modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/Amount.scala#L26-L28). The compiler cannot find the implicit for [`AmountValueRefined`](https://github.com/cngo-github/tessellation/blob/20230215-rosetta_parse/modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/Amount.scala#L22). I have already imported  `eu.timepit.refined.auto._`. Which import should I be using? How would I systematically find out in the future?

I can make the positive tests with [slightly ugly code](https://github.com/cngo-github/tessellation/blob/20230215-rosetta_parse/modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/construction/ConstructionServiceSuite.scala#L160-L220), but I cannot make the negative tests.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
```diff
@@ -48,5 +59,67 @@ object ConstructionService {
...
+          List(negativeOperation, positiveOperation)
+        }
+
+        operations.flatMap { operationList =>
+          NonEmptyList.fromList(operationList) match {
+            case Some(value) => value.asRight
+            case None        => "No valid operations found.".asLeft
+          }
+        }
+      }
```
kpudlik, line 93: You can use `for-comprehension` to get rid of nested code.

Also `NonEmptyList.fromList(List(negativeOperation, positiveOperation))` can be written as `NonEmptyList.of(negativeOperation, positiveOperation)` because you already know that it will pass.

kpudlik:
```scala
    def transactionToOperations(transaction: Transaction): Either[Throwable, NonEmptyList[Operation]] = {
      type Refinement = Not[Equal[0L]]

      val operations: Either[String, NonEmptyList[Operation]] = for {
        amount <- refineV[Refinement](transaction.amount.value.value)
        negAmount <- refineV[Refinement](-1L * transaction.amount.value.value)
        negativeOperation = Operation(
          OperationIdentifier(OperationIndex(0L)),
          none,
          OperationType.Transfer,
          none,
          AccountIdentifier(transaction.destination, none).some,
          Amount(AmountValue(negAmount), DAG).some
        )
        positiveOperation = Operation(
          OperationIdentifier(OperationIndex(1L)),
          none,
          OperationType.Transfer,
          none,
          AccountIdentifier(transaction.destination, none).some,
          Amount(AmountValue(amount), DAG).some
        )

        ops <- NonEmptyList.of(negativeOperation, positiveOperation).asRight
      } yield ops

      operations.leftMap(e => new Throwable(e))
    }
```

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
```diff
@@ -48,5 +59,67 @@ object ConstructionService {
...
+
+    def parseUnsignedTransaction(hex: Hex): Either[ConstructionError, ConstructionParse.ServiceReturn] =
+      KryoSerializer[F]
+        .deserialize[Transaction](hex.toBytes)
+        .flatMap(transactionToOperations)
+        .map(operations => ConstructionParse.ServiceReturn(operations, none))
+        .leftMap(_ => MalformedTransaction)
+
+    def parseTransaction(hex: Hex, isSigned: Boolean): EitherT[F, ConstructionError, ConstructionParse.ServiceReturn] =
+      EitherT(if (isSigned) {
```
kpudlik, line 119: You can just lift `Either[ConstructionError, ConstructionParse.ServiceReturn]` from `parseUnsignedTransaction` to `EitherT[F, ConstructionError, ConstructionParse.ServiceReturn]` https://typelevel.org/cats/datatypes/eithert.html#from-either-a-b-or-f-either-a-b-to-eithert-f-a-b

```scala
import cats.syntax.either._

if (isSigned) 
  parseSignedTransaction(hex)
else
  parseUnsignedTransaction.toEitherT

```

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
```diff
@@ -48,5 +59,67 @@ object ConstructionService {
...
+
+        operations.flatMap { operationList =>
+          NonEmptyList.fromList(operationList) match {
+            case Some(value) => value.asRight
+            case None        => "No valid operations found.".asLeft
+          }
+        }
+      }
+
+    def parseSignedTransaction(hex: Hex): F[Either[ConstructionError, ConstructionParse.ServiceReturn]] =
```
kpudlik, line 95: No need to use `traverse` there.

```scala
    def parseSignedTransaction(hex: Hex): EitherT[F, MalformedTransaction.type, ConstructionParse.Response]] = {
      val response = for {
        signedTransaction <- KryoSerializer[F].deserialize[Signed[Transaction]](hex.toBytes).toEitherT
        ops <- transactionToOperations(signedTransaction).toEitherT
        proofs <- signedTransaction.proofs.toNonEmptyList.traverse(_.id.hex.toPublicKey).attemptT
        ids = proofs.map(_.toAddress).map(AccountIdentifier(_, none))
      } yield ConstructionParse.Response(ops.toList, ids.toList.some)

      response.leftMap(_ => MalformedTransaction)
    }

    def parseUnsignedTransaction(hex: Hex): Either[MalformedTransaction.type, ConstructionParse.Response] =
      KryoSerializer[F]
        .deserialize[Transaction](hex.toBytes)
        .flatMap(transactionToOperations)
        .map(ops => ConstructionParse.Response(ops.toList, none))
        .leftMap(_ => MalformedTransaction)
```

Also for-comprehension is a syntax sugar to `flatMap` so you can avoid the callback hell: https://en.wiktionary.org/wiki/callback_hell by writing your nested code without nesting

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
```diff
@@ -48,5 +59,67 @@ object ConstructionService {
...
+            signedTransaction.proofs.toNonEmptyList.traverse(signature => signature.id.hex.toPublicKey).map { proofs =>
+              val proofAccountIdentifiers = proofs.map(_.toAddress).map { address =>
+                AccountIdentifier(address, none)
+              }
+
+              ConstructionParse.ServiceReturn(operationsList, proofAccountIdentifiers.some)
+            }
+          }.map(_.left.map(_ => MalformedTransaction))
+        }
+        .map(_.left.map(_ => MalformedTransaction).flatten)
```
kpudlik, line 109: `map` and then `flatten` is just a `flatMap`.

Also if you have `map` inside `map` then it means that you are missing `flatMap` somewhere because these methods look almost the same. See my other comments.

cngo-github:
I implemented `eqv` for the needed classes in order to make the tests cleaner.

I did not implement the negative tests because once I added this test:

```
  test("returns MalformedTransaction for an invalid signed transaction hex") { res =>
    implicit val (sp, k) = res

    val hex = Hex(
      "0483e4f38072fa59975fc796f220f4c07a7a6a3af1ad7fc091cbd6b8ebe78bac6a959da3587e6e761daf93693d4d2dc6b349fbc44dac5a9fcc5f809a59e93818ea"
    )

    val cs = ConstructionService.make[IO]()

    cs.parseTransaction(hex, true)
      .value
      .map(
        expect.eql(Left(MalformedTransaction), _)
      )
  }
```

I get the following error instead of `MalformedTransaction`:

```
ClassCastException: class java.lang.Float cannot be cast to class org.tessellation.security.signature.Signed (java.lang.Float is in module java.base of loader 'bootstrap'; org.tessellation.security.signature.Signed is in unnamed module of loader 'app')
```

kpudlik:
@cngo-github there is no need to implement the `eq`. The way how `@derive()` works is the following:

Assuming that you have product type (type which contains other types, for example `case class`):

```scala
case class C(a: A, b: B)
```

You can auto-generate (derive) `Eq[C]` for `C` by writing:

```scala
@derive(eqv)
case class C(a: A, b: B)
```
buuut there must be already `Eq[A]` and `Eq[B]` implemented.

In your case there is `Eq[ConstructionError]` but no `Eq[ServiceReturn]` so it can't derive `Eq[Either[ConstructionError, ServiceReturn]]`. So you need to:

```scala
  @derive(eqv)
  case class ServiceReturn(
    operations: NonEmptyList[Operation],
    accountIdentifierSigners: Option[NonEmptyList[AccountIdentifier]]
  )
```

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
cngo-github: Changed to using `NonEmptyList.of` and `for-comprehension`.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
cngo-github: Changed.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
cngo-github: I switched to `for-comprehension`

cngo-github:
> @cngo-github there is no need to implement the `eq`. The way how `@derive()` works is the following:
> 
> Assuming that you have product type (type which contains other types, for example `case class`):
> 
> ```scala
> case class C(a: A, b: B)
> ```
> 
> You can auto-generate (derive) `Eq[C]` for `C` by writing:
> 
> ```scala
> @derive(eqv)
> case class C(a: A, b: B)
> ```
> 
> buuut there must be already `Eq[A]` and `Eq[B]` implemented.
> 
> In your case there is `Eq[ConstructionError]` but no `Eq[ServiceReturn]` so it can't derive `Eq[Either[ConstructionError, ServiceReturn]]`. So you need to:
> 
> ```scala
>   @derive(eqv)
>   case class ServiceReturn(
>     operations: NonEmptyList[Operation],
>     accountIdentifierSigners: Option[NonEmptyList[AccountIdentifier]]
>   )
> ```
> 
> Ofc there must be `Eq[Operation]` and `Eq[AccountIdentifier]` etc, etc so it can derive the whole chain of types. You probably did it like that but just word `implement` is not a proper word for that because it is actually deriving instances for you.

I am not sure I understand. I changed the code quite a bit, so please review the new changes and clarify this comment.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/api/construction/ConstructionParse.scala
```diff
@@ -0,0 +1,24 @@
...
+  @derive(customizableDecoder)
+  case class Request(networkIdentifier: NetworkIdentifier, signed: Boolean, transaction: Hex)
+
+  @derive(customizableEncoder)
+  case class Response(operations: List[Operation], accountIdentifierSigners: Option[NonEmptyList[AccountIdentifier]])
+
+  case class ServiceReturn(
+    operations: NonEmptyList[Operation],
+    accountIdentifierSigners: Option[NonEmptyList[AccountIdentifier]]
+  )
```
marcinwadon, line 23: imo it's totally fine to return a tuple from the service. `ServiceReturn` is just a bad name and I cannot think of any better one 🤔

modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/construction/ConstructionServiceSuite.scala
```diff
@@ -126,7 +126,7 @@ object ConstructionServiceSuite extends MutableIOSuite with Checkers with Transa
         Transfer,
         none,
         AccountIdentifier(address, none).some,
-        Amount(AmountValue(-1L), Currency(CurrencySymbol("DAG"), CurrencyDecimal(8L))).some
+        Amount(AmountValue(-1L), currency.Currency(currency.CurrencySymbol("DAG"), currency.CurrencyDecimal(8L))).some
```
marcinwadon, line 129: bad auto import

modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/construction/ConstructionServiceSuite.scala
```diff
@@ -149,11 +149,73 @@ object ConstructionServiceSuite extends MutableIOSuite with Checkers with Transa
         Transfer,
         none,
         AccountIdentifier(address, none).some,
-        Amount(AmountValue(1L), Currency(CurrencySymbol("DAG"), CurrencyDecimal(8L))).some
+        Amount(AmountValue(1L), currency.Currency(currency.CurrencySymbol("DAG"), currency.CurrencyDecimal(8L))).some
```
marcinwadon, line 152: as above

modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/construction/ConstructionServiceSuite.scala
```diff
@@ -149,11 +149,73 @@ object ConstructionServiceSuite extends MutableIOSuite with Checkers with Transa
...
       )
 
       val accountIdentifiers = cs.getAccountIdentifiers(List(operation))
       expect.eql(none, accountIdentifiers)
     }
   }
+
+  test("returns operations for a valid signed transaction hex") { res =>
```
marcinwadon, line 160: there is no test checking whether the negative or positive operation is assigned to the correct address.
@kpudlik just found out that you use destination address twice

modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/construction/ConstructionServiceSuite.scala
```diff
@@ -149,11 +149,73 @@ object ConstructionServiceSuite extends MutableIOSuite with Checkers with Transa
...
+
+        generateTransactions(srcAddress, srcKey, dstAddress, txCount)
+          .flatMap(_.traverse { hashedTransaction =>
+            KryoSerializer[F]
+              .serialize(hashedTransaction.signed.value)
+              .map(Hex.fromBytes(_))
+              .map((_, hashedTransaction.hash))
+              .liftTo[IO]
+          })
+    }
```
marcinwadon, line 210: I see this is duplicated part (DRY - don't repeat yourself). Please take it outside the test case to reuse it here and there

marcinwadon (changes requested):

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
```diff
@@ -48,5 +58,54 @@ object ConstructionService {
...
+    def transactionToOperations(transaction: Transaction): EitherT[F, ConstructionError, NonEmptyList[Operation]] = {
+      val result = for {
+        transactionAmount <- refineV[Not[Equal[0L]]](transaction.amount.value.value)
+        negativeTransactionAmount <- refineV[Not[Equal[0L]]](-1L * transaction.amount.value.value)
+        negativeOperation = Operation(
+          OperationIdentifier(OperationIndex(0L)),
+          none,
+          OperationType.Transfer,
+          none,
+          AccountIdentifier(transaction.destination, none).some,
```
kpudlik, line 71: Shouldn't it be `transaction.source`? I mean if you transfer `100 DAG` from `A` to `B` then it probably should be:
`Operation(index = 0, address = 'A', value = -100), Operation(index = 1, address = 'B', value = +100)`.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
```diff
@@ -48,5 +58,54 @@ object ConstructionService {
 
       NonEmptyList.fromList(accountIdentifiers)
     }
+
+    def transactionToOperations(transaction: Transaction): EitherT[F, ConstructionError, NonEmptyList[Operation]] = {
+      val result = for {
+        transactionAmount <- refineV[Not[Equal[0L]]](transaction.amount.value.value)
+        negativeTransactionAmount <- refineV[Not[Equal[0L]]](-1L * transaction.amount.value.value)
+        negativeOperation = Operation(
+          OperationIdentifier(OperationIndex(0L)),
```
kpudlik, line 67: I see that you put `OperationIndex` manually here. I think we can improve it because you have a list of operations which are not related to each other (`relatedOperations` is None). For such case you can just use index of `NonEmptyList`:

```scala
    def transactionToOperations(transaction: Transaction): Either[Throwable, NonEmptyList[Operation]] = {
      type Refinement = Not[Equal[0L]]

      val negativeTransfer = refineV[Refinement](-transaction.amount.value.value)
        .map((transaction.source, _)) // Combine negative value with source address
      val positiveTransfer = refineV[Refinement](+transaction.amount.value.value)
        .map((transaction.destination, _)) // Combine positive value with destination address

      NonEmptyList
        .of(negativeTransfer, positiveTransfer) // Create list of operations in order
        .traverseWithLongIndexM { // Include index of element and convert it to `OperationIndex`
          case (Right((address, amount)), index) =>
            Operation(
              OperationIdentifier(OperationIndex(PosLong.unsafeFrom(index))), // We are sure that index is positive so we can use unsafeFrom here
              none,
              OperationType.Transfer,
              none,
              AccountIdentifier(address, none).some,
              Amount(AmountValue(amount), DAG).some
            ).asRight[Throwable]
          case _ => Left(MalformedTransaction)
        }
    }
```

Also the solution is a little bit more DRY

modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/construction/ConstructionServiceSuite.scala
kpudlik: Exactly. It returns operations but operations are incorrect due to a mistake in a code.

You should check following things:

1. `result.operations.lenght == 2`
2. `result.operations[0].address == srcAddress && result.operations[1].address == dstAddress`
3. `result.operations[0].amount = -result.operations[1].amount`

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/api/construction/ConstructionParse.scala
marcinwadon: I mean, this idea is brillant if you can find a better name (actually the name that describes what it is, instead of where it is used). Otherwise, just return a tuple

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
cngo-github: Changed.

marcinwadon: @cngo-github it changes nothing, so you can omit this, but @kpudlik's written that for better readability to put focus it's negative or positive operation

modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/construction/ConstructionServiceSuite.scala
cngo-github: Refactored.

modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/construction/ConstructionServiceSuite.scala
cngo-github: Added these tests.

modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/construction/ConstructionServiceSuite.scala
cngo-github: Changed.

modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/construction/ConstructionServiceSuite.scala
cngo-github: This change was because of the following linter error.

```
/Users/chuong/Documents/dev/constellationnetwork/tessellation/modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/construction/ConstructionServiceSuite.scala:164:43: reference to CurrencySymbol is ambiguous;
[error] it is imported twice in the same scope by
[error] import org.tessellation.rosetta.domain.currency._
[error] and import org.tessellation.rosetta.domain._
[error]         Amount(AmountValue(-1L), Currency(CurrencySymbol("DAG"), CurrencyDecimal(8L))).some
```

modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/construction/ConstructionServiceSuite.scala
```diff
@@ -36,6 +39,53 @@ object ConstructionServiceSuite extends MutableIOSuite with Checkers with Transa
       .forAsync[IO](sharedKryoRegistrar)
       .flatMap(kryo => SecurityProvider.forAsync[IO].map(securityProvider => (securityProvider, kryo)))
 
+  def generateTransaction(wantSignedTransaction: Boolean) =
```
marcinwadon, line 42: If I read this correctly, it creates more than one transaction

modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/construction/ConstructionServiceSuite.scala
marcinwadon: Yup, so don't import it twice and you should be good

modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/construction/ConstructionServiceSuite.scala
cngo-github: Updated.

modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/construction/ConstructionServiceSuite.scala
cngo-github: Without importing `org.tessellation.rosetta.domain.currency._`, `Currency`, `CurrencySymbol`, etc. don't resolve. If I take out `org.tessellation.rosetta.domain._`, I lose `RosettaError`, `TransactionIdentifier`, etc. Using `currency.` seems like the least bad way to solve the problem.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/api/construction/ConstructionParse.scala
cngo-github: Renamed to `ParseResult`.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
```diff
@@ -48,5 +58,53 @@ object ConstructionService {
...
+
+    def parseSignedTransaction(hex: Hex): EitherT[F, ConstructionError, ConstructionParse.ParseResult] = {
+      val response = for {
+        signedTransaction <- KryoSerializer[F].deserialize[Signed[Transaction]](hex.toBytes).toEitherT
+        operations <- transactionToOperations(signedTransaction).toEitherT
+        proofs <- signedTransaction.proofs.toNonEmptyList.traverse(_.id.hex.toPublicKey).attemptT
+        accountIds = proofs.map(_.toAddress).map(AccountIdentifier(_, none))
+      } yield ConstructionParse.ParseResult(operations, accountIds.some)
+
+      response.leftMap(_ => MalformedTransaction)
```
marcinwadon, line 94: why `response`? It's not a response in here

modules/rosetta/src/main/scala/org/tessellation/rosetta/http/routes/ConstructionRoutes.scala
```diff
@@ -44,6 +44,23 @@ final class ConstructionRoutes[F[_]: Async](
...
+        .decodeRosettaWithNetworkValidation[ConstructionParse.Request](appEnvironment, _.networkIdentifier) { parseReq =>
+          constructionService
+            .parseTransaction(parseReq.transaction, parseReq.signed)
+            .bimap(
+              _.toRosettaError,
+              serviceResponse =>
+                ConstructionParse.Response(
+                  serviceResponse.operations.toList,
+                  serviceResponse.accountIdentifierSigners
+                )
```
marcinwadon, line 59: in `ConstructionParse.Response` you can write a fn:
```scala
def fromParseResult(result: ParseResult): Response = ...
```
and you can use it here like:
```scala
.bimap(
  _.toRosettaError,
  ConstructionParse.Response.fromParseResult
)
```

modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/construction/ConstructionServiceSuite.scala
```diff
@@ -36,6 +39,53 @@ object ConstructionServiceSuite extends MutableIOSuite with Checkers with Transa
       .forAsync[IO](sharedKryoRegistrar)
       .flatMap(kryo => SecurityProvider.forAsync[IO].map(securityProvider => (securityProvider, kryo)))
 
+  def generateTestTransactions(wantSignedTransaction: Boolean) =
+    sharedResource.use { res =>
```
marcinwadon, line 43: instead of `sharedResource.use` you should specify required implicits:
```scala
def generateTestTransactions(wantSignedTransaction: Boolean)(implicit val S: SecurityProvider[IO], K: KryoSerializer[IO]) = ...
```

modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/construction/ConstructionServiceSuite.scala
marcinwadon: You can import the exact classes you need

marcinwadon (changes requested):

modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/construction/ConstructionServiceSuite.scala
cngo-github: If I import all the `currency` classes I need, I end up with

`import org.tessellation.rosetta.domain.currency.{Currency, CurrencySymbol, CurrencyDecimal, DAG}`

after the linter is run, that becomes

`import org.tessellation.rosetta.domain.currency._`

and then the `ambiguous reference` errors pop up.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
cngo-github: Renamed to `result`.

modules/rosetta/src/main/scala/org/tessellation/rosetta/http/routes/ConstructionRoutes.scala
cngo-github: Changed.

modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/construction/ConstructionServiceSuite.scala
cngo-github: Changed.

cngo-github:
> > @cngo-github there is no need to implement the `eq`. The way how `@derive()` works is the following:
> > Assuming that you have product type (type which contains other types, for example `case class`):
> > ```scala
> > case class C(a: A, b: B)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > You can auto-generate (derive) `Eq[C]` for `C` by writing:
> > ```scala
> > @derive(eqv)
> > case class C(a: A, b: B)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > buuut there must be already `Eq[A]` and `Eq[B]` implemented.
> > In your case there is `Eq[ConstructionError]` but no `Eq[ServiceReturn]` so it can't derive `Eq[Either[ConstructionError, ServiceReturn]]`. So you need to:
> > ```scala
> >   @derive(eqv)
> >   case class ServiceReturn(
> >     operations: NonEmptyList[Operation],
> >     accountIdentifierSigners: Option[NonEmptyList[AccountIdentifier]]
> >   )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Ofc there must be `Eq[Operation]` and `Eq[AccountIdentifier]` etc, etc so it can derive the whole chain of types. You probably did it like that but just word `implement` is not a proper word for that because it is actually deriving instances for you.
> 
> I am not sure I understand. I changed the code quite a bit, so please review the new changes and clarify this comment.

@kpudlik, Reading this again, I think I understand. I implemented `derive(eqv)` as appropriate, however, when running the negative tests, I get:

```
ClassCastException: class java.lang.Float cannot be cast to class org.tessellation.security.signature.Signed (java.lang.Float is in module java.base of loader 'bootstrap'; org.tessellation.security.signature.Signed is in unnamed module of loader 'app')
[error] 
[error]   EitherT.scala:446                     cats.data.EitherT#$anonfun$flatMap$1
[error]   ConstructionService.scala:88          flatMap @ org.tessellation.rosetta.domain.construction.ConstructionService$$anon$1#parseSignedTransaction
[error]   ConstructionService.scala:44          leftMap @ org.tessellation.rosetta.domain.construction.ConstructionService$$anon$1#derive
[error]   ConstructionServiceSuite.scala:313    map @ org.tessellation.rosetta.domain.construction.ConstructionServiceSuite$#$anonfun$new$57
[error]   Test.scala:21                         defer @ weaver.Test$#$anonfun$apply$2
[error]   Stream.scala:2201                     get @ fs2.Stream#$anonfun$parEvalMapActionImpl$1
```

I am not able to comprehend the error and code well enough to track it down right now. I am leaving the negative tests off for now.

marcinwadon (approved):
overall looks good to me

kpudlik (approved):
I added 2 comments but nothing critical so LGTM.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/construction/ConstructionService.scala
```diff
@@ -48,5 +58,53 @@ object ConstructionService {
...
+    }
+
+    def parseUnsignedTransaction(hex: Hex): EitherT[F, ConstructionError, ConstructionParse.ParseResult] =
+      for {
+        unsignedTransaction <- KryoSerializer[F].deserialize[Transaction](hex.toBytes).toEitherT.leftMap(_ => MalformedTransaction)
+        operations <- transactionToOperations(unsignedTransaction).toEitherT
+      } yield ConstructionParse.ParseResult(operations, none)
```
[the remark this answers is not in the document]
marcinwadon: I support that @cngo-github

PR #569, opened 2023-02-23
Create Currency L0 App

kpudlik:


modules/currency-l0/src/main/scala/org/tessellation/currency/CurrencyApp.scala
```diff
@@ -1,24 +1,157 @@
...
+import org.tessellation.sdk.resources.MkHttpServer
+import org.tessellation.sdk.resources.MkHttpServer.ServerName
+import org.tessellation.security.signature.Signed
+
+abstract class CurrencyApp[A <: CliMethod](
+  header: String,
+  clusterId: ClusterId,
+  helpFlag: Boolean,
+  version: String,
+  tokenSymbol: String
```
marcinwadon, line 33: newtype?

modules/currency-l0/src/main/scala/org/tessellation/currency/modules/HealthChecks.scala
```diff
@@ -0,0 +1,49 @@
...
+    storages: Storages[F],
+    services: Services[F],
+    programs: Programs[F],
+    p2pClient: P2PClient[F],
+    client: Client[F],
+    session: Session[F],
+    config: HealthCheckConfig,
+    selfId: PeerId
+  ): F[HealthChecks[F]] = {
+    def ping = PingHealthCheckConsensus.make(
```
marcinwadon, line 29: is it required? s it the local healthcheck we use currently?

modules/currency-l0/src/main/scala/org/tessellation/currency/http/routes/CurrencyRoutes.scala
```diff
@@ -0,0 +1,84 @@
...
+import org.http4s.circe.CirceEntityCodec.circeEntityDecoder
+import org.http4s.dsl.Http4sDsl
+import org.http4s.server.Router
+import org.tessellation.sdk.domain.snapshot.services.AddressService
+import shapeless._
+import shapeless.syntax.singleton._
+
+final case class CurrencyRoutes[F[_]: Async](addressService: AddressService[F, CurrencySnapshot], mkCell: L0Cell.Mk[F])
+    extends Http4sDsl[F] {
+  private[routes] val prefixPath = "/dag"
```
marcinwadon, line 22: probably not `/dag`. by the way. I think Routes can be abstracted (possibly as another PR), as it depends on an address service.
Also, I'd change `/dag` to `/currency` and make an alias `/dag` -> `/currency` (or redirection) in our core module, to make it an universal endpoint.

kpudlik: Alias redirection is safer because we won't break existing integration 👍🏻

modules/currency-l0/src/main/scala/org/tessellation/currency/modules/HealthChecks.scala
kpudlik: It was in `core` so I decided to just mimic it.

modules/currency-l0/src/main/scala/org/tessellation/currency/http/routes/CurrencyRoutes.scala
kpudlik: Also I will extract `CurrencyRoutes` because it is 1:1 with DAG routes. Let's make it in next PR.

modules/currency-l0/src/main/scala/org/tessellation/currency/http/routes/DebugRoutes.scala
```diff
@@ -0,0 +1,75 @@
...
+import org.tessellation.sdk.infrastructure.consensus.{ConsensusResources, PeerDeclarations}
+
+import derevo.circe.magnolia.{decoder, encoder}
+import derevo.derive
+import org.http4s.HttpRoutes
+import org.http4s.circe.CirceEntityCodec.circeEntityEncoder
+import org.http4s.dsl.Http4sDsl
+import org.http4s.server.Router
+
+final case class DebugRoutes[F[_]: Async](
```
kpudlik, line 21: I will extract these. I just need to pass all the services manually instead of `Services` fascade.

modules/currency-l0/src/main/scala/org/tessellation/currency/http/routes/SnapshotRoutes.scala
```diff
@@ -0,0 +1,67 @@
...
+import io.circe.Encoder
+import io.circe.shapes._
+import org.http4s.circe.CirceEntityEncoder
+import org.http4s.dsl.Http4sDsl
+import org.http4s.server.Router
+import org.http4s.{EntityEncoder, HttpRoutes}
+import shapeless.HNil
+import shapeless.syntax.singleton._
+
+final case class SnapshotRoutes[F[_]: Async: KryoSerializer](
```
kpudlik, line 24: It will also be extracted in next changes and reused (by passing prefix as param, for Global it will be `global-snapshot` and for Currency it will be `snapshot`).

modules/currency-l0/src/main/scala/org/tessellation/currency/infrastructure/healthcheck/HealthCheckDaemon.scala
```diff
@@ -0,0 +1,34 @@
...
+import org.tessellation.currency.modules.HealthChecks
+import org.tessellation.sdk.domain.Daemon
+
+import fs2._
+
+trait HealthCheckDaemon[F[_]] extends Daemon[F] {}
+
+object HealthCheckDaemon {
+
+  def make[F[_]: Async](healthChecks: HealthChecks[F])(implicit S: Supervisor[F]): HealthCheckDaemon[F] = new HealthCheckDaemon[F] {
```
kpudlik, line 18: It can be extracted. I will create a common trait with `def trigger` method and just `healthChecks.trigger()`. Thanks to that Daemon does not care what kind of triggers we have and can be reused.

modules/currency-l0/src/main/scala/org/tessellation/currency/infrastructure/snapshot/CurrencySnapshotEventsPublisherDaemon.scala
```diff
@@ -0,0 +1,28 @@
...
+import org.tessellation.sdk.infrastructure.consensus.message.ConsensusEvent
+import org.tessellation.security.signature.Signed
+
+import fs2.Stream
+
+object CurrencySnapshotEventsPublisherDaemon {
+
+  def make[F[_]: Async: Supervisor](
+    l1OutputQueue: Queue[F, Signed[CurrencyBlock]],
+    gossip: Gossip[F]
```
kpudlik, line 18: I will extract it and parametrize by `B <: Block`

modules/currency-l0/src/main/scala/org/tessellation/currency/cli/snapshot.scala
```diff
@@ -0,0 +1,35 @@
...
+import eu.timepit.refined.auto._
+import fs2.io.file.Path
+
+object snapshot {
+
+  val snapshotPath: Opts[Path] = Opts
+    .env[Path]("CL_SNAPSHOT_STORED_PATH", help = "Path to store created snapshot")
+    .withDefault(Path("data/snapshot"))
+
+  val opts = snapshotPath.map { snapshotPath =>
```
kpudlik, line 18: Default configs can be shared between currency and dag

modules/currency-l0/src/main/scala/org/tessellation/currency/modules/HttpApi.scala
```diff
@@ -0,0 +1,137 @@
...
+      healthchecks,
+      privateKey,
+      environment,
+      selfId,
+      nodeVersion,
+      httpCfg
+    ) {}
+}
+
+sealed abstract class HttpApi[F[_]: Async: SecurityProvider: KryoSerializer: Metrics] private (
```
kpudlik, line 56: It can also be extracted I think. We just need to parametryze by specific Cell.

modules/currency-l0/src/main/scala/org/tessellation/currency/modules/HealthChecks.scala
kpudlik: `PingHealthCheckConsensus` is running on both `core` and `dag-L1`. I will extract if ofc to be DRY for now but let's discuss the local healthcheck with the rest of teeam.

modules/core/src/main/scala/org/tessellation/http/routes/DagRoutes.scala
```diff
@@ -1,84 +1,44 @@
...
-
-    case req @ POST -> Root / "l1-output" =>
-      req
-        .as[Signed[DAGBlock]]
-        .map(L0CellInput.HandleDAGL1(_))
-        .map(mkDagCell)
-        .flatMap(_.run())
-        .flatMap(_ => Ok())
+    case req @ _ =>
+      println(req.uri.path)
```
marcinwadon: ups :)

modules/currency-l0/src/main/scala/org/tessellation/currency/modules/HealthChecks.scala
kpudlik: @TheMMaciek ^

marcinwadon (approved):

marcinwadon:
label is missing

PR #571, opened 2023-02-24
Global L0 peer discovery and state channel snapshot offload

kpudlik:
First: https://github.com/Constellation-Labs/tessellation/pull/569


marcinwadon:
label is missing

modules/currency-l0/src/main/scala/org/tessellation/currency/ExampleCurrencyL0.scala
```diff
@@ -0,0 +1,16 @@
...
+
+import java.util.UUID
+
+import org.tessellation.currency.schema.currency.TokenSymbol
+import org.tessellation.schema.address.Address
+import org.tessellation.schema.cluster.ClusterId
+
+import eu.timepit.refined.auto._
+
+object ExampleCurrencyL0
```
marcinwadon, line 11: do we need this in the currency module? shouldn't it be some kind of example project?

kpudlik: For now I decided to keep it there and wait for @TheMMaciek's changes. Then we will make proper modules and structure the project.

marcinwadon (dismissed):

modules/currency-l0/src/main/scala/org/tessellation/currency/ExampleCurrencyL0.scala
```diff
@@ -0,0 +1,16 @@
...
+    extends CurrencyL0App(
+      TokenSymbol("TST"),
+      ClusterId(UUID.fromString("517c3a05-9219-471b-a54c-21b7d72f4ae5")),
+      Address("DAG77VVVRvdZiYxZ2hCtkHz68h85ApT5b2xzdTkn")
```
tbekas, line 15: Couldn’t this be just a set of CLI parameters?
```
java -jar currency-l1.jar —symbol=TST —clusterId=517… —identifier=DAG7Q…
```

kpudlik: Yes, we plan to get rid of inheritance in favor of parametrized binary. For now I left it like that because @TheMMaciek also implemented it like that and the draft docs have it like that.

Also `clusterId` can be derived by calling something like `ClusterId(UUID.nameUUIDFromBytes(address.value.getBytes))`

kpudlik: + we can get rid of `TokenSymbol` because it isn't even used and is just for the debugging. According to latest agreements with Alex and Ryle - all the currencies will be isomorphic so there is no need to introduce symbol. At least for now.

TheMMaciek: At least that's what I intend to do in the currency L1 part.

modules/currency-l0/src/main/scala/org/tessellation/currency/modules/Programs.scala
```diff
@@ -1,30 +1,39 @@
...
+      storages.l0Cluster
+    )
+
+    new Programs[F](sdkPrograms.peerDiscovery, l0PeerDiscovery, sdkPrograms.joining, download) {}
   }
 }
 
 sealed abstract class Programs[F[_]] private (
   val peerDiscovery: PeerDiscovery[F],
+  val l0PeerDiscovery: L0PeerDiscovery[F],
```
TheMMaciek, line 36: I guess you could call it `globalL0PeerDiscovery` to help a little in identifying what it is actually (differentiating it from `peerDiscovery` which runs for other L0 currency peers).

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/StateChannelSnapshotClient.scala
```diff
@@ -0,0 +1,47 @@
...
+import org.tessellation.security.SecurityProvider
+import org.tessellation.security.signature.Signed
+import org.tessellation.statechannel.StateChannelSnapshotBinary
+
+import org.http4s.Method.POST
+import org.http4s.circe.CirceEntityCodec.circeEntityEncoder
+import org.http4s.client.Client
+
+trait StateChannelSnapshotClient[F[_]] {
+  def sendStateChannelSnapshot(data: Signed[Snapshot[_, _]]): PeerResponse[F, Boolean]
```
TheMMaciek, line 25: I think we will never send a `GlobalSnapshot` back to global L0, so maybe this should be parametrized by `CurrencySnapshot` for now, or simply `StateChannelSnapshotBinary` and let's have the logic preparing it outside of the client.

modules/currency-l0/src/main/scala/org/tessellation/currency/ExampleCurrencyL0.scala
kpudlik: For now we will use inheritance model because providing CLI parameters needs bigger refactoring.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/Main.scala
```diff
@@ -1,86 +1,72 @@
...
-  helpFlag: Boolean,
-  version: String,
-  symbol: TokenSymbol,
-  identifier: Address
-) extends TessellationIOApp[A](clusterId.toString, header, clusterId, helpFlag, version) {
+object Main
+    extends TessellationIOApp[Run](
+      "Currency-l0",
+      "Currency L0 node",
+      ClusterId(UUID.randomUUID()),
```
TheMMaciek, line 32: I guess this will prevent the nodes from connecting with each other, because for every node the clusterId will be different right? 🤔

TheMMaciek: That of course is the case if we are checking the clusterId during join, and I think we are.

marcinwadon: @TheMMaciek i think it works on compile time but let's confirm @kpudlik. If it runs on runtime then @TheMMaciek is right

kpudlik: You are right. I was debugging something so I forced different values for ClusterId and forgot to revert it to fixed one.

kpudlik: Of course it still works if you have just one L0 node

kpudlik: fixed

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelSnapshotService.scala
```diff
@@ -0,0 +1,67 @@
...
+    keyPair: KeyPair,
+    lastSignedBinaryHashStorage: LastSignedBinaryHashStorage[F],
+    stateChannelSnapshotClient: StateChannelSnapshotClient[F],
+    globalL0ClusterStorage: L0ClusterStorage[F],
+    snapshotStorage: SnapshotStorage[F, CurrencySnapshotArtifact]
+  ): StateChannelSnapshotService[F] =
+    new StateChannelSnapshotService[F] {
+      private val logger = Slf4jLogger.getLogger
+
+      override def createBinary(snapshot: Signed[CurrencySnapshotArtifact]): F[Signed[StateChannelSnapshotBinary]] = for {
```
TheMMaciek, line 42: I guess you can omit the `override`.

TheMMaciek (approved):

TheMMaciek (approved):

PR #574, opened 2023-02-27
Fix redirection for `/dag` routes

kpudlik:


marcinwadon (dismissed):

modules/core/src/main/scala/org/tessellation/modules/HttpApi.scala
```diff
@@ -81,8 +81,8 @@ sealed abstract class HttpApi[F[_]: Async: SecurityProvider: KryoSerializer: Met
   private val trustRoutes = TrustRoutes[F](storages.trust, programs.trustPush)
   private val stateChannelRoutes = StateChannelRoutes[F](services.stateChannel)
   private val snapshotRoutes = SnapshotRoutes[F, GlobalSnapshot](storages.globalSnapshot, "/global-snapshots")
-  private val dagRoutes = DagRoutes[F]()
-  private val currencyRoutes = CurrencyRoutes[F, DAGTransaction, DAGBlock, GlobalSnapshot](services.address, mkDagCell)
+  private val dagRoutes = CurrencyRoutes[F, DAGTransaction, DAGBlock, GlobalSnapshot]("/dag", services.address, mkDagCell)
+  private val currencyRoutes = CurrencyRoutes[F, DAGTransaction, DAGBlock, GlobalSnapshot]("/currency", services.address, mkDagCell)
```
TheMMaciek, line 85: We want it twice? Once after "dag" prefix and one more under "currency"?

kpudlik: @marcinwadon suggested to have both on DAG for consistency. But up to you.

marcinwadon: @TheMMaciek how would you keep both the backward compatibility and the new prefix `/currency` across all the implementations at the same time?

marcinwadon: yep, I want `/currency` to be consistent across all the implementations. but also `/dag` is required to keep backward compatibility

kpudlik: Imo if API of currency and dag are isomorphic then we can keep both for DAG. It will be easier for future integrations to use `currency`. Otherwise we would need to introduce symbol and make prefixes to be sure that you are sending tx to the right state channel.

TheMMaciek: Well what I wanted to do was just to parametrize clients with the matching prefix.

TheMMaciek: Up to you guys. FYI the `L0DAGClusterClient` should be adjusted if we are taking this approach.

TheMMaciek: It points to the `/dag` prefix ...

kpudlik: >Well what I wanted to do was just to parametrize clients with the matching prefix.

But we decided to get rid of tokensymbol so for now there will be no prefix.

kpudlik: - We keep `/dag` for `L0DAGClusterClient` and existing integrations
- We introduce `/currency` additionally for consistency and future integrations

marcinwadon: @TheMMaciek think not only about our tessellation, as this is a public API. Those endpoints are used by dag4js as well, so you would have to parametrize everything/everywhere. 

For future, it would be great to keep it just /currency. but due to backward compatibility, we wanted to make a redirection, but it didn't work out as expected

TheMMaciek: I'm talking about "currency" or "dag" prefixes not symbol.

Regarding `L0DAGClusterClient` what I understood is that we are keeping "/dag" prefix only for backwards compatibility, but we can change it where ever possible to use new common prefix. If we are keeping the "/dag" prefix in the client I will have to abstract it away anyway for the currency so to some extent what's the point? And why add "/currency" prefix if we don't intend to use it ourselves.

tbekas: Keep in mind that adding any new public endpoints creates a burden of maintaining them possibly indefinitely into the future.

TheMMaciek: I know what is the intention and I understand it. We can still change the mentioned client to use the new prefix.

kpudlik: Ok guys. For now I will not add `currency` to DAG unless we are on the same page. We need some prefix for all the currencies so it will be just `/currency` for any currency and `/dag` just for DAG.

kpudlik: But @TheMMaciek you need to then parametrize `L0DAGClusterClient` anyway if you want tio reuse that class for both DAG and Currency.

tbekas (approved):

PR #576, opened 2023-02-28
Abstract L1 functionality and implement DAG L1

TheMMaciek:


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/http/p2p/L0CurrencyClusterClient.scala
```diff
@@ -0,0 +1,27 @@
...
+import org.http4s.circe.CirceEntityCodec.circeEntityEncoder
+import org.http4s.client.Client
+
+trait L0CurrencyClusterClient[F[_], B <: Block[_]] {
+  def sendL1Output(output: Signed[B]): PeerResponse[F, Boolean]
+}
+
+object L0CurrencyClusterClient {
+
+  def make[F[_], B <: Block[_]: Encoder](pathPrefix: String, client: Client[F]): L0CurrencyClusterClient[F, B] =
```
kpudlik, line 19: Have you found any refined type for `pathPrefix`?

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/Validator.scala
```diff
@@ -0,0 +1,34 @@
...
+import cats.syntax.order._
+
+import org.tessellation.ext.cats.syntax.next.catsSyntaxNext
+import org.tessellation.schema.height.SubHeight
+import org.tessellation.schema.snapshot.Snapshot
+import org.tessellation.security.Hashed
+
+object Validator {
+
+  def isNextSnapshot[S <: Snapshot[_, _]](previous: Hashed[S], next: S): Boolean =
```
kpudlik, line 12: Verify if that method is even used. I remember that it wasn't used at all

modules/shared/src/main/scala/org/tessellation/shared/package.scala
```diff
@@ -72,7 +72,7 @@ package object shared {
     classOf[SnapshotTips] -> 607,
     classOf[ActiveTip] -> 608,
     ActiveTip.OrderingInstance.getClass -> 609,
-    classOf[BlockAsActiveTip[DAGBlock]] -> 610,
+    classOf[BlockAsActiveTip[_]] -> 610,
```
kpudlik, line 75: If it works then it means that `BlockAsActiveTip[CurrencyBlock]` is not needed. I remmeber that Kryo didn't care about the parameter of generic so good that it works.

kpudlik (changes requested):

marcinwadon (changes requested):
Use covariant for `NoData` and other case objects (discussed on slack), no other comments on my side :+1:

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/Validator.scala
TheMMaciek: It was used in snapshot-streaming (but it obviously can't be seen in the tessellation repo) and right now I'm also using it in `SnapshotProcessor`

modules/shared/src/main/scala/org/tessellation/shared/package.scala
TheMMaciek: Yeah afaik the type parameter here is ignored anyway and the id for `DAGBlock` (or for `CurrencyBlock`) is taken from the registry. I also tested it on the cluster and checked that all unit test work.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/http/p2p/L0CurrencyClusterClient.scala
TheMMaciek: I was thinking about it for a moment but then decided that we will probably figure it out if we decide to add prefixes to all currency endpoints in general. I didn't want to implement partial solution here with a type. We may need multiple things thought through e.g. ADT for prefixes possibly (if limited) or a single refined type or something else, how to inject it in non-boilerplate'y way, should it be in the code or configurable. For now I implemented it similarly to how it's implemented on L0 layer.

kpudlik (approved):

kpudlik (approved):

marcinwadon (approved):

PR #580, opened 2023-03-01
Extract common part for currency to shared module

TheMMaciek:


build.sbt
```diff
@@ -464,11 +464,30 @@ lazy val core = (project in file("modules/core"))
     )
   )
 
+lazy val currencyShared = (project in file("modules/currency-shared"))
+  .enablePlugins(AshScriptPlugin)
+  .enablePlugins(BuildInfoPlugin)
```
kpudlik, line 469: Do we need BuildInfo there?

kpudlik (changes requested):

build.sbt
kpudlik: Also add `currencyShared` to `root` so scalafix works

TheMMaciek: I guess we add buildInfo everywhere so I did the same here.

TheMMaciek: Not true actually. Removed the plugins that are not required at least right now.

kpudlik (approved):

marcinwadon (approved):

PR #581, opened 2023-03-02
Implement currency L1 layer

TheMMaciek:


modules/currency-l1/src/main/scala/org/tessellation/currency/l1/Main.scala
```diff
@@ -0,0 +1,177 @@
...
+              programs.globalL0PeerDiscovery.discoverFrom(cfg.globalL0Peer) >>
+              currencyStorages.node.tryModifyState(NodeState.Initial, NodeState.ReadyToJoin)
+        }
+      }.asResource
+      globalL0PeerDiscovery = Stream
+        .awakeEvery[IO](10.seconds)
+        .evalMap { _ =>
+          currencyStorages.lastGlobalSnapshot.get.flatMap {
+            _.fold(Applicative[IO].unit) { latestSnapshot =>
+              programs.globalL0PeerDiscovery.discover(latestSnapshot.signed.proofs.map(_.id).map(PeerId._Id.reverseGet))
```
kpudlik, line 170: I have a feeling that it should be Daemon/Program.

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/domain/snapshot/programs/CurrencySnapshotProcessor.scala
```diff
@@ -0,0 +1,169 @@
...
+          BlockStorage[F, CurrencyBlock],
+          LastSnapshotStorage[F, CurrencySnapshot],
+          TransactionStorage[F, CurrencyTransaction]
+        )
+      ] = {
+        val bs = blockStorage.getState().flatMap(BlockStorage.make[F, CurrencyBlock](_))
+        val lcss = lastCurrencySnapshotStorage.get.flatMap(LastSnapshotStorage.make(_))
+        val as = addressStorage.getState().flatMap(AddressStorage.make(_))
+        val ts =
+          transactionStorage.getState().flatMap { case (lastTxs, waitingTxs) => TransactionStorage.make(lastTxs, waitingTxs) }
```
kpudlik, line 149: Could you tell me more about these intermediate storages? I mean I am wondering why these are created inside the processor.

kpudlik (changes requested):

kpudlik (approved):

kpudlik (dismissed):

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/Main.scala
```diff
@@ -0,0 +1,177 @@
...
+      ClusterId(UUID.fromString("517c3a05-9219-471b-a54c-21b7d72f4ae5")),
+      version = BuildInfo.version
+    ) {
+
+  val opts: Opts[Run] = cli.method.opts
+
+  type KryoRegistrationIdRange = CurrencyKryoRegistrationIdRange Or SdkOrSharedOrKernelRegistrationIdRange Or DagL1KryoRegistrationIdRange
+
+  val kryoRegistrar: Map[Class[_], KryoRegistrationId[KryoRegistrationIdRange]] =
+    currencyKryoRegistrar.union(sdkKryoRegistrar).union(dagL1KryoRegistrar)
```
marcinwadon, line 51: why `dagL1` is required in currency L1?

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/Main.scala
```diff
@@ -0,0 +1,177 @@
...
+        )
+      snapshotProcessor = CurrencySnapshotProcessor.make(
+        method.identifier,
+        currencyStorages.address,
+        currencyStorages.block,
+        currencyStorages.lastGlobalSnapshot,
+        currencyStorages.lastSnapshot,
+        currencyStorages.transaction
+      )
+      programs = CurrencyPrograms.make(sdkPrograms, p2pClient, currencyStorages, snapshotProcessor)
```
marcinwadon, line 83: shouldn't it be just `Programs`, like we do with sdk and dag?

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/Main.scala
marcinwadon: @TheMMaciek shouldn't it be a part of state channel runtime?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/address/storage/AddressStorage.scala
```diff
@@ -4,6 +4,7 @@ import org.tessellation.schema.address.Address
 import org.tessellation.schema.balance.Balance
 
 trait AddressStorage[F[_]] {
+  def getState(): F[Map[Address, Balance]]
```
marcinwadon, line 7: imo it should be `def getState: F[Map[Address, Balance]]`

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/Main.scala
TheMMaciek: Currently all the abstractions for L1 functionality are in dag module. Hence currencyL1 depends on dagL1. We can extract it to a separate module as one of the next steps and have both currencyL1 and dagL1 depend on this common logic in a common module. But for the purpose of easier review I decided to keep the abstractions in dagL1 module for now (these were merged in the previous PR) so it's not like moving implementation and abstracting it away at the same time consequently obscuring what was actually changing in the logic. As to why we need `dagL1KryoRegistrar` there are class registered there that are needed for the abstraction e.g. p2p consensus input (which is being singed).

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/Main.scala
TheMMaciek: I guess it can be just `Programs`. Probably the reason I did it that way was that some of the modules from abstraction could be reused as they were, and possible to emphasize that these one's here actually come from currency-l1 module I might have added the prefix. Will remove it.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/address/storage/AddressStorage.scala
TheMMaciek: Without the parenthesis, right.

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/Main.scala
TheMMaciek: It could be a daemon, it could also be a part of the state channel class. But for now I would consider that this solution is at least ok. In a way it is a part of the state channel runtime because I merge both streams here. I just wanted to avoid premature abstracting of state channel class functionality (or injecting with a parameter) when this part is the only part I needed to add. I would revisit it the moment we will tackle the extendability of L0 token functionality as we may then decide on how the custom stuff should be added to the implementation and hence also this part could fall into place then.

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/Main.scala
TheMMaciek: Actually the reason might have been that the complier perceived it as a cyclic dependency, e.i. `Programs` extending `Programs` (from dagL1), but I will handle it with aliasing when imporitng.

marcinwadon (approved):

PR #588, opened 2023-03-17
Add merkle tree validation

kriscieslak:


modules/shared/src/main/scala/org/tessellation/merkletree/MerkleTreeValidator.scala
```diff
@@ -0,0 +1,39 @@
...
+import org.tessellation.schema.snapshot.{Snapshot, SnapshotInfo}
+import org.tessellation.schema.{GlobalIncrementalSnapshot, GlobalSnapshotInfo, SnapshotOrdinal}
+import org.tessellation.security.{Hashed, hash}
+
+import derevo.cats.{eqv, show}
+import derevo.derive
+
+object MerkleTreeValidator {
+
+  def validate[F[_]: std.Console: MonadThrow: KryoSerializer](snapshot: Hashed[GlobalIncrementalSnapshot], state: GlobalSnapshotInfo) =
```
marcinwadon, line 20: bad import

modules/shared/src/main/scala/org/tessellation/merkletree/MerkleTreeValidator.scala
```diff
@@ -0,0 +1,39 @@
...
+    validateGeneral(snapshot, state)(_.stateProof, GlobalSnapshotInfo.stateProof(_))
+
+  def validate[F[_]: MonadThrow: KryoSerializer](snapshot: Hashed[CurrencyIncrementalSnapshot], state: CurrencySnapshotInfo) =
+    validateGeneral(snapshot, state)(_.stateProof, CurrencySnapshotInfo.stateProof(_))
```
marcinwadon, line 24: I think you can abstract it over `IncrementalSnapshot` and `SnapshotInfo` 🤔

modules/shared/src/test/scala/org/tessellation/merkletree/MerkleTreeValidatorSuite.scala
```diff
@@ -0,0 +1,138 @@
...
+object MerkleTreeValidatorSuite extends MutableIOSuite {
+
+  type Res = KryoSerializer[IO]
+
+  override def sharedResource: Resource[IO, Res] =
+    KryoSerializer.forAsync[IO](sharedKryoRegistrar)
+
+  private val hash = Hash("abc")
+
+  test("should succeed when use the same snapshot info for incremental global snapshot") { implicit ks =>
```
marcinwadon, line 39: imo - the test name should specify what it tests:
`should succeed when state matches snapshot's state proof`
it checks whether the state is valid, so it matches the proof
also, it fails when it cannot proof it, or it doesn't match the proof

modules/shared/src/main/scala/org/tessellation/merkletree/MerkleTreeValidator.scala
kriscieslak: I can think of two ways to do this:
- I can add two interfaces and mixin them to IncrementalSnapshots and to SnapshotInfo
- I can add type classes to act similar to above.
Which do you prefer?

modules/shared/src/main/scala/org/tessellation/merkletree/MerkleTreeValidator.scala
kriscieslak: My bad. Forgot to delete it.

modules/shared/src/test/scala/org/tessellation/merkletree/MerkleTreeValidatorSuite.scala
kriscieslak: Changed it.

modules/shared/src/main/scala/org/tessellation/merkletree/MerkleTreeValidator.scala
marcinwadon: imo type class has a different purpose, so the first option seems to be ok.

marcinwadon: so, instead of mixing it, I'd just make a proper hierarchy:
```
Snapshot[T, B]
IncrementalSnapshot extends Snapshot[T, B] {
 val stateProof: MerkleTree
}
FullSnapshot extends Snapshot[T, B, SI <: SnapshotInfo] {
 val state: SI // fyi - I'd rename "info" to "state"
}

CurrencySnapshot extends FullSnapshot[..., CurrencySnapshotInfo]
CurrencyIncrementalSnapshot extends IncrementalSnapshot[Curr..]

```

modules/shared/src/main/scala/org/tessellation/currency/schema/currency.scala
```diff
@@ -79,12 +78,18 @@ object currency {
...
+      .map(MerkleTree.from)
+  }
 
   object CurrencySnapshotInfo {
     def stateProof[F[_]: MonadThrow: KryoSerializer](info: CurrencySnapshotInfo): F[MerkleTree] =
-      (info.lastTxRefs.hashF, info.balances.hashF).tupled
-        .map(_.toNonEmptyList)
+      NonEmptyList
+        .of(info.lastTxRefs.hashF, info.balances.hashF)
+        .sequence
```
TheMMaciek, line 92: Looks like only one of these is needed.

TheMMaciek: I mean only one `stateProof` function for `CurrencySnapshotInfo`

TheMMaciek (approved):

PR #591, opened 2023-03-20
Apply context fns for global and currency snapshot processors

marcinwadon:


marcinwadon:
@Constellation-Labs/protocol ready for review

PR #592, opened 2023-03-21
Serialize CurrencySnapshot using circe

kriscieslak:


modules/shared/src/main/scala/org/tessellation/json/JsonSerializer.scala
```diff
@@ -0,0 +1,13 @@
+package org.tessellation.json
+
+import io.circe.jawn.JawnParser
+import io.circe.syntax._
+import io.circe.{Decoder, Encoder}
+
+object JsonSerializer {
+  def serialize[A: Encoder](content: A) =
```
marcinwadon, line 8: Missing types

marcinwadon (approved):
LGTM, but imo there should be return types

modules/shared/src/main/scala/org/tessellation/json/JsonSerializer.scala
```diff
@@ -0,0 +1,13 @@
...
+    content.asJson.noSpaces.getBytes("UTF-8")
+
+  def deserialize[A: Decoder](content: Array[Byte]): Either[Throwable, A] =
+    JawnParser(false).decodeByteArray(content)
```
TheMMaciek, line 12: Specifying charset is not necessary?

TheMMaciek (changes requested):
We also need to use this functionality in currency L1 -> currency snapshot processor still uses kryo to deserialize binary.

modules/shared/src/main/scala/org/tessellation/json/JsonSerializer.scala
kriscieslak: There is no parameter to pass charset, and I've checked, that under the hood utf-8 is used.

cngo-github:
@sdavidp for reference

modules/shared/src/test/scala/org/tessellation/json/JsonSerializerSuite.scala
```diff
@@ -0,0 +1,71 @@
...
+    currencyIncrementalSnapshot[IO](Hash.empty, CurrencySnapshotInfo(SortedMap.empty, SortedMap.empty)).map { signedSnapshot =>
+      val serialized = JsonSerializer.serialize(signedSnapshot)
+      val deserialized = JsonSerializer.deserialize[CurrencySnapshot](serialized)
+      expect.same(true, deserialized.isLeft)
+    }
+  }
+
+  private def currencyIncrementalSnapshot[F[_]: MonadThrow: KryoSerializer](
+    hash: Hash,
+    currencySnapshotInfo: CurrencySnapshotInfo
```
TheMMaciek, line 49: You are using this function twice with the same "empty" value for this parameter. You could remove the parameter altogether or give a default value for it if you think injecting different value will be useful in the future.

modules/shared/src/main/scala/org/tessellation/statechannel/StateChannelSnapshotBinary.scala
```diff
@@ -7,8 +7,9 @@ import org.tessellation.schema.arrayOrder
 import org.tessellation.security.hash.Hash
 
 import derevo.cats.{eqv, order}
-import derevo.circe.magnolia.{decoder, encoder}
 import derevo.derive
+import derevo.circe.magnolia.decoder
+import derevo.circe.magnolia.encoder
```
TheMMaciek, line 12: No change needed in this file

TheMMaciek (approved):

TheMMaciek (approved):

PR #593, opened 2023-03-22
Add version field to snapshot

kpudlik:


modules/shared/src/main/scala/org/tessellation/schema/snapshot.scala
```diff
@@ -14,8 +14,30 @@ import org.tessellation.schema.transaction.{Transaction, TransactionReference}
...
+  @derive(encoder, decoder, eqv, show)
+  @newtype
+  case class SemVer(
+    version: String Refined MatchesRegex[
+      "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$"
+    ]
+  )
+
+  object SemVer {
+    val Empty: SemVer = SemVer("0.0.1")
```
TheMMaciek, line 38: I don't think we can call it empty.

marcinwadon: "Default" or "Initial" probably

modules/shared/src/main/scala/org/tessellation/schema/snapshot.scala
```diff
@@ -14,8 +14,30 @@ import org.tessellation.schema.transaction.{Transaction, TransactionReference}
...
+import eu.timepit.refined.cats._
+import eu.timepit.refined.string.MatchesRegex
+import io.circe.refined._
+import io.estatico.newtype.macros.newtype
+
 object snapshot {
 
+  @derive(encoder, decoder, eqv, show)
+  @newtype
+  case class SemVer(
```
TheMMaciek, line 31: Maybe it would be a good idea to make the `SemVer` an ADT? Every new version that's gonna be released will be a new case object added to the family? 🤔

modules/shared/src/main/scala/org/tessellation/currency/schema/currency.scala
```diff
@@ -80,7 +80,8 @@ object currency {
     lastSnapshotHash: Hash,
     blocks: SortedSet[BlockAsActiveTip[CurrencyBlock]],
     tips: SnapshotTips,
-    info: CurrencySnapshotInfo
+    info: CurrencySnapshotInfo,
+    version: SemVer = SemVer.Empty
```
TheMMaciek, line 84: I wonder. Shouldn't it be hardcoded/baked into the class upon definition instead of parametrizable?

modules/shared/src/main/scala/org/tessellation/schema/snapshot.scala
```diff
@@ -14,8 +14,30 @@ import org.tessellation.schema.transaction.{Transaction, TransactionReference}
...
+    version: String Refined MatchesRegex[
+      "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$"
```
TheMMaciek, line 33: Would be nice to have some test for this regex.

modules/shared/src/main/scala/org/tessellation/schema/snapshot.scala
TheMMaciek: something like:
```
  @derive(encoder, decoder, eqv, show)
  sealed trait SemVer {
    val version: String Refined MatchesRegex[...]
  }

  object SemVer {
    case object `0.0.1` extends SemVer { val version = "0.0.1" }
  }
```

TheMMaciek: In case we won't do anything in our codebase based on these versions we won't need to keep the history of them as ADT as well. Just a hardcoded version field for a snapshot should be enough. We need to think about the exact use case for this to really decide.

modules/shared/src/main/scala/org/tessellation/schema/snapshot.scala
kpudlik: or `MinValue` maybe?

marcinwadon (changes requested):
imo this also should be done against `feature/incremental-snapshots` branch

modules/shared/src/main/scala/org/tessellation/currency/schema/currency.scala
kpudlik: I baked more or less. Take a look

kpudlik: ofc assuming that we do not need history in form of ADT

marcinwadon (approved):

PR #595, opened 2023-03-27
Fix Signed unique signers validation

marcinwadon:


tbekas (approved):
LGTM

modules/shared/src/main/scala/org/tessellation/security/signature/SignedValidator.scala
```diff
@@ -37,7 +37,7 @@ trait SignedValidator[F[_]] {
 
   def isSignedExclusivelyBy[A <: AnyRef](
     signed: Signed[A],
-    minSignatureCount: Address
+    signerAddress: Address
```
TheMMaciek, line 40: I guess you also need to change the name for the implementation.

marcinwadon: it was already changed @TheMMaciek

PR #596, opened 2023-03-27
Added CSV loader for trust ratings

cngo-github:
Added CSV loader for trust ratings


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/trust/trustTypes.scala
```diff
@@ -0,0 +1,16 @@
...
+import fs2.data.csv.generic.semiauto.deriveRowDecoder
+
+object trustTypes {
+  case class TrustRating(value: Float)
+
+  object TrustRating {
+    implicit val rowDecoder: RowDecoder[TrustRating] = deriveRowDecoder
+  }
+
+  type TrustRatingRefined = Float Refined Interval.Closed[-1.0, 1.0]
```
cngo-github, line 15: My next step is to incorporate the refined type. I may have to make a custom `RowDecoder` in order to do that.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/trust/TrustRatingCsvLoader.scala
marcinwadon, line 1: Hey. Just a general question. Why do you need a CSV Loader to create a domain and the very first logic?
I think that loading data from CSV should be one of last steps.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/trust/trustTypes.scala
marcinwadon, line 1: can you use the same file name format as other files in the codebase? we've never used `fooTypes` name.
also, you declare types in the `infrastructure`, but that's the domain part actually. it's part of the schema, so it's not infrastructure related.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/trust/trustTypes.scala
marcinwadon: I think you can use map/contramap for that purpose. Anyway, again - it's not a correct order of doing things. Start from the domain part (model, types), then business logic, then talking with outside world (csv loader). That way, the CSV loader would work on the specific types that are final and refined.

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/trust/TrustRatingCsvLoaderSuite.scala
```diff
@@ -0,0 +1,19 @@
...
+
+object TrustRatingCsvLoaderSuite extends SimpleIOSuite with Checkers {
+  test("load trust ratings from csv") {
+    val testFileLocation = getClass.getResource("/ratings.sample.csv").getPath
+    val inFile = Path(testFileLocation)
+
+    for {
+      csv <- TrustRatingCsvLoader.make[IO].load(inFile)
+      _ = println(csv)
+    } yield success
```
marcinwadon, line 17: println shouldn't be a part of a test. you specified a sample rating, so the test assertion should expect that specific rating to be red

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/trust/TrustRatingCsvLoaderSuite.scala
```diff
@@ -0,0 +1,19 @@
...
+import weaver._
+import weaver.scalacheck._
+
+object TrustRatingCsvLoaderSuite extends SimpleIOSuite with Checkers {
+  test("load trust ratings from csv") {
+    val testFileLocation = getClass.getResource("/ratings.sample.csv").getPath
+    val inFile = Path(testFileLocation)
+
+    for {
+      csv <- TrustRatingCsvLoader.make[IO].load(inFile)
```
marcinwadon, line 15: for comprehension is not needed for a single element case.

it corresponds to that:
```scala
TrustRatingCsvLoader.make[IO].load(inFile).map { csv =>
  println(csv)
  success
}
```
or without println to that:
```scala
TrustRatingCsvLoader.make[IO].load(inFile).as(success)
```

marcinwadon (changes requested):

tbekas:
I think we need a high-level overview of the design of the trust functionality. A while ago we've employed a system of [ADRs](https://github.com/Constellation-Labs/tessellation/tree/develop/doc/adr) that are written in plain-text English. Any bigger functionality, like this one, should have its own ADR submitted first.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/trust/trustTypes.scala
cngo-github: I changed the overall object to `trust` and moved it to domain.

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/trust/TrustRatingCsvLoaderSuite.scala
cngo-github: Replaced with `expect.eql`.

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/trust/TrustRatingCsvLoaderSuite.scala
cngo-github: Changed.

marcinwadon:
hey @sdavidp. if possible, please ammend the original commit (as you fix it in the same PR) and make yourself a coauthor

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/trust/trustTypes.scala
cngo-github: Implemented the refined type.

cngo-github:
> hey @sdavidp. if possible, please ammend the original commit (as you fix it in the same PR) and make yourself a coauthor. as long as it's possible, we want to have a single commit per single PR. (exception for this is when a separate commit provides a meaningful, separate change and cannot be made into PR separately)

The commits have been squashed and further cleaned up.

sdavidp:
> hey @sdavidp. if possible, please ammend the original commit (as you fix it in the same PR) and make yourself a coauthor. as long as it's possible, we want to have a single commit per single PR. (exception for this is when a separate commit provides a meaningful, separate change and cannot be made into PR separately)

@marcinwadon hey... @cngo-github  did tell me about doing single commits -- Chuong and I were pairing to resolve some issues and this push was so he could get the code to his machine. I'm sure he'll fix it up according to our standards.

Re. making me a coauthor, Chuong is already looking into that.
Thanks Marcin, I appreciate the detail-oriented approach to the codebase.

cngo-github:
Added the code to integrate the CSV upload and parse with the CLI.

modules/core/src/main/scala/org/tessellation/domain/trust/storage/TrustStorage.scala
```diff
@@ -5,6 +5,7 @@ import org.tessellation.schema.trust.{InternalTrustUpdateBatch, PublicTrust, Tru
 
 trait TrustStorage[F[_]] {
   def updateTrust(trustUpdates: InternalTrustUpdateBatch): F[Unit]
+  def updateTrust(trustUpdates: Option[InternalTrustUpdateBatch]): F[Unit]
```
marcinwadon, line 8: There is no need for such function. You should fold the option and execute update only for some

modules/sdk/src/main/scala/org/tessellation/sdk/domain/trust.scala
```diff
@@ -0,0 +1,35 @@
...
+import eu.timepit.refined.cats._
+import eu.timepit.refined.numeric.Interval
+import eu.timepit.refined.refineV
+import fs2.data.csv.generic.semiauto.deriveRowDecoder
+import fs2.data.csv.{CellDecoder, DecoderError}
+
+object trust {
+  @derive(eqv)
+  case class TrustRating(peerId: PeerId, rating: TrustRatingRefined) {
+    def toInternalTrustUpdate = InternalTrustUpdate(peerId, rating.value)
```
marcinwadon, line 21: Imo internal trust update should use refined types

modules/sdk/src/test/resources/ratings.sample.csv
```diff
@@ -0,0 +1,4 @@
+698c4c20b6975ff2bf61b82b078d0253, -1
```
marcinwadon, line 1: Spaces are not required in the csv

modules/sdk/src/test/resources/ratings.sample.csv
```diff
@@ -0,0 +1,4 @@
...
+aadeb50882008f5796fa0c487a8c17da, 0
```
marcinwadon, line 2: These are not valid peer ids

marcinwadon (changes requested):

modules/core/src/main/scala/org/tessellation/domain/trust/storage/TrustStorage.scala
cngo-github: Changed.

modules/core/src/main/scala/org/tessellation/modules/Storages.scala
```diff
@@ -25,10 +27,12 @@ object Storages {
   def make[F[_]: Async: KryoSerializer: Supervisor](
     sdkStorages: SdkStorages[F],
     snapshotConfig: SnapshotConfig,
-    maybeRollbackHash: Option[Hash]
+    maybeRollbackHash: Option[Hash],
+    trustUpdates: Option[InternalTrustUpdateBatch]
   ): F[Storages[F]] =
     for {
       trustStorage <- TrustStorage.make[F]
+      _ <- trustUpdates.fold(Applicative[F].pure(()))(trustStorage.updateTrust(_))
```
marcinwadon, line 35: There is an Applicative unit, so you dont need to write pure(())

modules/sdk/src/test/resources/ratings.sample.csv
cngo-github: Replaced with ids from `peerGen`.

modules/core/src/main/scala/org/tessellation/modules/Storages.scala
cngo-github: Changed.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/cli/method.scala
```diff
@@ -82,6 +83,9 @@ object method {
 
     private val seedlistPathOpts: Opts[Option[Path]] = Opts.option[Path]("seedlist", "").orNone
 
+    val trustRatingsPathOpts: Opts[Option[Path]] =
+      Opts.option[Path]("ratings", "The path to the CSV of peer ID ratings").orNone
```
marcinwadon, line 87: I see this is duplicated everywhere. imo it would be great to define it once and reuse later on

modules/sdk/src/test/resources/ratings.sample.csv
marcinwadon: which are also not valid peer ids

modules/shared/src/main/scala/org/tessellation/cli/opts.scala
marcinwadon, line 1: why `shared` module? imo `shared` has no knowledge about the node, so it should be in the `sdk` imo

modules/sdk/src/main/scala/org/tessellation/sdk/domain/trust.scala
```diff
@@ -0,0 +1,35 @@
...
+  object TrustRating {
+    implicit val peerIdDecoder: CellDecoder[PeerId] = CellDecoder.stringDecoder.map(Hex(_)).map(PeerId(_))
+
+    implicit val trustRatingRefinedDecoder: CellDecoder[TrustRatingRefined] =
+      CellDecoder.doubleDecoder.emap(refineV[TrustRatingRefinement](_).leftMap(new DecoderError(_)))
+
+    implicit val rowDecoder = deriveRowDecoder[TrustRating]
+  }
+
+  type TrustRatingRefinement = Interval.Closed[-1.0, 1.0]
```
marcinwadon, line 33: I see it duplicated

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/cli/method.scala
cngo-github: `genesisPathOpts`, `seedlistPathOpts`, and `trustRatingsPathOpts` have been centralized in `org.tessellation.sdk.cli.opts`.

modules/sdk/src/test/resources/ratings.sample.csv
cngo-github: But they were taken from `peerGen` the arbitrary for `Peer`s. If those are not correct, then is that generator valid? Where should I be getting valid peer ids from?

modules/shared/src/main/scala/org/tessellation/cli/opts.scala
cngo-github: This file has been moved to the `sdk` module.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/trust.scala
cngo-github: `TrustRatingRefined` and `TrustRatingRefinement` have been replaced by `InternalTrustUpdateRefined` and `InternalTrustUpdateRefinement` respectively.

modules/core/src/main/scala/org/tessellation/infrastructure/trust/storage/TrustStorage.scala
```diff
@@ -12,42 +13,40 @@ import org.tessellation.schema.trust.{InternalTrustUpdateBatch, PublicTrust, Tru
...
+
+          def updatePeerPublicTrustInfo(peerId: PeerId, publicTrust: PublicTrust): F[Unit] = trust.update { t =>
+            t.updatedWith(peerId) { trustInfo =>
+              trustInfo
+                .orElse(Some(TrustInfo()))
+                .map(_.copy(peerLabels = publicTrust.labels))
+            }
+          }
+
+          override def updatePredictedTrust(trustUpdates: Map[PeerId, Double]): F[Unit] = trust.update { t =>
```
marcinwadon, line 44: what it overrides if the class extends nothing?

modules/sdk/src/test/resources/ratings.sample.csv
```diff
@@ -0,0 +1,4 @@
+eZNBzdERjxeHdbCRSOyVL,-1
+fKBudHJCIxgGTIZYFSOYwKKvPGIKjt,0
+QyNfuwYquLsVLsjaEixlp,1
+EGHOZIHIgMJsXVRmmVLxBSTEeKJoYwNPYFM,0.5
```
marcinwadon, line 4: still wrong peer ids

cngo-github: Can you tell me where to get valid `PeerId`s?

cngo-github: I took 4 IDs from https://constellationlabs-dag.s3.us-west-1.amazonaws.com/testnet-seedlist

modules/core/src/main/scala/org/tessellation/infrastructure/trust/storage/TrustStorage.scala
cngo-github: Removed.

modules/core/src/main/scala/org/tessellation/infrastructure/trust/storage/TrustStorage.scala
```diff
@@ -12,42 +13,40 @@ import org.tessellation.schema.trust.{InternalTrustUpdateBatch, PublicTrust, Tru
 object TrustStorage {
 
   def make[F[_]: Monad: Ref.Make]: F[TrustStorage[F]] =
-    for {
-      trust <- Ref[F].of[Map[PeerId, TrustInfo]](Map.empty)
-    } yield make(trust)
-
-  def make[F[_]: Monad](trust: Ref[F, Map[PeerId, TrustInfo]]): TrustStorage[F] =
```
TheMMaciek, line 19: I think it's not a bad idea to have two make methods the way it was. The one with ref as parameter can be useful in tests for example so that the non empty trust map can be injected.

modules/core/src/main/scala/org/tessellation/modules/Storages.scala
```diff
@@ -25,10 +27,12 @@ object Storages {
   def make[F[_]: Async: KryoSerializer: Supervisor](
     sdkStorages: SdkStorages[F],
     snapshotConfig: SnapshotConfig,
-    maybeRollbackHash: Option[Hash]
+    maybeRollbackHash: Option[Hash],
+    trustUpdates: Option[InternalTrustUpdateBatch]
   ): F[Storages[F]] =
     for {
       trustStorage <- TrustStorage.make[F]
+      _ <- trustUpdates.fold(Applicative[F].unit)(trustStorage.updateTrust(_))
```
TheMMaciek, line 35: Not sure if this should be done here as a part of general `Storages` initialization flow. Maybe it would be better to have an additional `make(trustUpdates: Option[InternalTrustUpdateBatch]): F[TrustStorage]` method in `TrustStorage`. And the trust updates would happen as a part of `TrustStorage` initialization. WDYT?

modules/core/src/main/scala/org/tessellation/infrastructure/trust/storage/TrustStorage.scala
```diff
@@ -12,42 +13,40 @@ import org.tessellation.schema.trust.{InternalTrustUpdateBatch, PublicTrust, Tru
...
-      def updateTrust(trustUpdates: InternalTrustUpdateBatch): F[Unit] =
-        trust.update { t =>
-          t ++ trustUpdates.updates.map { trustUpdate =>
-            val capped = Math.max(Math.min(trustUpdate.trust, 1), -1)
-            val updated = t.getOrElse(trustUpdate.id, TrustInfo()).copy(trustLabel = Some(capped))
-            trustUpdate.id -> updated
-          }.toMap
+    Ref[F]
+      .of[Map[PeerId, TrustInfo]](Map.empty)
+      .map(trust =>
```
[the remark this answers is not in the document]
cngo-github: Changed.

TheMMaciek (dismissed):
It looks good overall in my opinion. I would change the trust updating to happen during initialization of `TrustStorage` instead of inside `Storages` class though.

modules/core/src/main/scala/org/tessellation/modules/Storages.scala
marcinwadon: I agree with @TheMMaciek

modules/core/src/main/scala/org/tessellation/infrastructure/trust/storage/TrustStorage.scala
cngo-github: Reverted the change.

modules/core/src/main/scala/org/tessellation/modules/Storages.scala
cngo-github: Added the third `make`.

marcinwadon (dismissed):
The test suite can be refactored slightly not to repeat yourself (enter inputs and map to case classes once) (for future reference) but overall looks good for me :+1: feel free to merge

modules/sdk/src/main/scala/org/tessellation/sdk/domain/trust.scala
```diff
@@ -0,0 +1,30 @@
...
+  @derive(eqv)
+  case class TrustRating(peerId: PeerId, rating: InternalTrustUpdateRefined) {
+    def toInternalTrustUpdate = InternalTrustUpdate(peerId, rating)
+  }
+
+  object TrustRating {
+    implicit val peerIdDecoder: CellDecoder[PeerId] = CellDecoder.stringDecoder.map(Hex(_)).map(PeerId(_))
+
+    implicit val internalTrustUpdateRefinedDecoder: CellDecoder[InternalTrustUpdateRefined] =
+      CellDecoder.doubleDecoder.emap(refineV[InternalTrustUpdateRefinement](_).leftMap(new DecoderError(_)))
```
sdavidp, line 26: I'm curious: Is there a reason these two decoders are not in their respective companion objects? `InternalTrustUpdateRefined` is being used by at least two types; if the behavior is intended to be the same in both cases then it feels like it should be its own type. WDYT?

modules/sdk/src/main/scala/org/tessellation/sdk/domain/trust.scala
```diff
@@ -0,0 +1,30 @@
...
+import derevo.cats.eqv
+import derevo.derive
+import eu.timepit.refined.cats._
+import eu.timepit.refined.refineV
+import fs2.data.csv.generic.semiauto.deriveRowDecoder
+import fs2.data.csv.{CellDecoder, DecoderError}
+
+object trust {
+  @derive(eqv)
+  case class TrustRating(peerId: PeerId, rating: InternalTrustUpdateRefined) {
```
sdavidp, line 18: Do these two classes have semantically different uses despite looking identical?
```
case class InternalTrustUpdate(id: PeerId, trust: InternalTrustUpdateRefined)
```

marcinwadon: Good point

modules/sdk/src/main/scala/org/tessellation/sdk/domain/trust.scala
marcinwadon: @sdavidp I think because they're cell decoders (for csv) so it will be used in this (and only in this) place.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/trust.scala
cngo-github: @marcinwadon @sdavidp , yes. @ryle-goehausen wants to keep them separate because one is internal facing and the other is external facing.

marcinwadon: What is the use case for both of them then? For now you don't use one, it you just map one into another. I'm in favor of abstracting the external one only if there is a need for that.

cngo-github: The two classes have been deduplicated for now. When they diverge, they will be changed then.

marcinwadon (approved):

PR #598, opened 2023-03-28
Add fee field to currency snapshot

kpudlik:


modules/shared/src/main/scala/org/tessellation/currency/schema/currency.scala
```diff
@@ -80,7 +90,8 @@ object currency {
     lastSnapshotHash: Hash,
     blocks: SortedSet[BlockAsActiveTip[CurrencyBlock]],
     tips: SnapshotTips,
-    info: CurrencySnapshotInfo
+    info: CurrencySnapshotInfo,
+    fee: SnapshotFee = SnapshotFee.MinValue
```
TheMMaciek, line 94: You may need to register it in kryo.

kpudlik: `TransactionFee` is not registered 🤔 

It is a `newtype` so at the very end it is just a `Long`

modules/shared/src/main/scala/org/tessellation/currency/schema/currency.scala
```diff
@@ -72,6 +72,16 @@ object currency {
     balances: SortedMap[Address, Balance]
   ) extends SnapshotInfo {}
 
+  @derive(decoder, encoder, order, show)
+  @newtype
+  case class SnapshotFee(value: NonNegLong)
```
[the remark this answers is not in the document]
kpudlik: But then you can mix `TransactionFee` and `SnapshotFee` so `@newtype` won't help you with catching that 🤔

modules/shared/src/main/scala/org/tessellation/currency/schema/currency.scala
kpudlik: fair point, it means that we definitely need to register `SnapshotFee`

TheMMaciek: That depends on how are the newtypes treated during kryo serialization. If they are treated as base types then the registration may in fact be unnecessary.

modules/shared/src/main/scala/org/tessellation/currency/schema/currency.scala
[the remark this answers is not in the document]
TheMMaciek: It could theoretically happen though I think it's unlikely. Anyway I'm fine with the `SnapshotFee` newtype

marcinwadon (changes requested):
should be done against `feature/incremental-snapshots` branch

modules/shared/src/main/scala/org/tessellation/currency/schema/currency.scala
marcinwadon: @TheMMaciek every time you see a newtype, think of it as an underlying type. the underlying type must be registered in kryo, because in fact newtype is just a "transparent wrapper". that being said, `Int` or `Long` are already registered, so it's all fine.
having:
```
@newtype
case class Foo(value: Bar)
```
you would have to register `Bar` type

TheMMaciek (approved):

PR #600, opened 2023-03-28
Enhance global l0 service

kriscieslak:


modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/Validator.scala
```diff
@@ -3,16 +3,18 @@ package org.tessellation.sdk.domain.snapshot
...
 
 object Validator {
 
-  def isNextSnapshot[S <: Snapshot[_, _]](previous: Hashed[S], next: S): Boolean =
-    compare[S](previous, next).isInstanceOf[Next]
+  def isNextSnapshot[T <: Transaction, B <: Block[T]](previous: Hashed[Snapshot[T, B]], next: Snapshot[T, B]): Boolean =
+    compare(previous, next).isInstanceOf[Next]
 
-  def compare[S <: Snapshot[_, _]](previous: Hashed[S], next: S): ComparisonResult = {
+  def compare[T <: Transaction, B <: Block[T]](previous: Hashed[Snapshot[T, B]], next: Snapshot[T, B]): ComparisonResult = {
```
TheMMaciek, line 17: That implies that now the transaction and block types restrict which snapshots can be compared, e.i. any two snapshots operating on the same transaction and block types can be compared. Designed for incremental and full snapshot comparison. Do we think that's fine? @Constellation-Labs/protocol

marcinwadon: does it mean that we can compare two different snapshots now?

TheMMaciek: Yes and that's what is used for comparing FullGlobalSnapshot with the following IncrementalGlobalSnapshot (both based on `DAGTransaction` and `DAGBlock`). I guess that's exactly why this change was implemented here. Just thinking about possible drawbacks for such a solution.

marcinwadon: maybe it can be clearly specified that you can check FullSnapshot with IncrementalSnapshot 🤔 we have types for that

TheMMaciek (approved):

marcinwadon (approved):

PR #601, opened 2023-03-28
Add max size restriction for state channel snapshot

TheMMaciek:


modules/core/src/main/scala/org/tessellation/domain/statechannel/StateChannelValidator.scala
```diff
@@ -6,6 +6,8 @@ import cats.syntax.all._
 
 import org.tessellation.domain.statechannel.StateChannelValidator.StateChannelValidationErrorOr
 import org.tessellation.ext.cats.syntax.validated._
+import org.tessellation.ext.kryo.RefinedSerializerF
```
marcinwadon, line 9: we used to import just `org.tessellation.ext.kryo._`

marcinwadon (approved):

marcinwadon (changes requested):
anyway, it should be done against `feature/incremental-snapshots` branch

PR #602, opened 2023-03-28
Abstract run method of Currency Framework

kpudlik:


modules/currency-l0/src/main/scala/org/tessellation/currency/l0/Main.scala
marcinwadon, line 1: this should be done against `feature/incremental-snapshots` branch

marcinwadon (changes requested):

PR #603, opened 2023-03-29
Add max size restriction for state channel snapshot

TheMMaciek:


modules/sdk/src/main/scala/org/tessellation/sdk/domain/statechannel/StateChannelValidator.scala
```diff
@@ -40,12 +44,26 @@ object StateChannelValidator {
...
+        else
+          BinaryExceedsMaxAllowedSize(maxBinarySizeInBytes, actualSize).invalidNec
+      }
   }
 
   @derive(eqv, show)
   sealed trait StateChannelValidationError
   case class InvalidSigned(error: SignedValidationError) extends StateChannelValidationError
   case object NotSignedExclusivelyByStateChannelOwner extends StateChannelValidationError
+  case class BinaryExceedsMaxAllowedSize(maxSize: Long, was: Int) extends StateChannelValidationError
```
tbekas, line 66: Probably it's better to use the same types for both cases. I guess that would be `Int` or `NonNegInt`.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/statechannel/StateChannelValidator.scala
```diff
@@ -40,12 +44,26 @@ object StateChannelValidator {
       signedValidator
         .isSignedExclusivelyBy(signedSC, address)
         .map(_.errorMap[StateChannelValidationError](_ => NotSignedExclusivelyByStateChannelOwner))
+
+    private def validateSnapshotSize(
+      signedSC: Signed[StateChannelSnapshotBinary]
+    ): F[StateChannelValidationErrorOr[Signed[StateChannelSnapshotBinary]]] =
+      signedSC.toBinaryF.map { binary =>
```
tbekas, line 51: I find it odd that we're calculating the size of the binary with the signatures. Specifically that the size of the signatures may vary. But if that's the requirement than ok.

marcinwadon (approved):

tbekas (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/domain/statechannel/StateChannelValidator.scala
marcinwadon: well, actually it should take into account signatures 🤔 shouldn't it @TheMMaciek?

TheMMaciek: Yeah that is a requirement to take signatures into account, to not allow flooding with signatures in the future.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/statechannel/StateChannelValidator.scala
TheMMaciek: Size for an array returns `Int`, so I went with `Int` to not upcast it to long. And wanted to use long for the maxBytesSize to be safe.

PR #605, opened 2023-03-29
Do not require collateral for currency L0 and L1

marcinwadon:


kpudlik (approved):

marcinwadon:
After talk with @kriscieslak, looks like this PR is not sufficient.

PR #607, opened 2023-03-30
Make JsonBinarySerializer.serialize safe

sdavidp:
*PR Category:* fix

Change `JsonBinarySerializer.serialize` to return an `Either`


modules/shared/src/main/scala/org/tessellation/json/JsonBinarySerializer.scala
```diff
@@ -1,12 +1,13 @@
...
+
 import io.circe.jawn.JawnParser
 import io.circe.syntax._
 import io.circe.{Decoder, Encoder}
-
 object JsonBinarySerializer {
-  def serialize[A: Encoder](content: A): Array[Byte] =
-    content.asJson.noSpaces.getBytes("UTF-8")
+  def serialize[A: Encoder](content: A): Either[Throwable, Array[Byte]] =
+    Try(content.asJson.noSpaces.getBytes("UTF-8")).toEither
```
tbekas, line 10: What exception are we catching here?

sdavidp:
> Thanks @sdavidp :) Looks good to me. ps. You don't need to write pr category in description while setting up the proper label (which we use for changelog generator)

Thank you @marcinwadon. I updated our Notion PR Guidelines doc.

modules/shared/src/main/scala/org/tessellation/json/JsonBinarySerializer.scala
TheMMaciek: I guess it could be a problem if we were specifying charset dynamically. If it's hardcoded right here it shouldn't ever throw I think (given that it worked once, why wouldn't it work every other time). Still maybe it's safer to wrap it anyway.

marcinwadon:
Can you send me on priv the link to that Notion page? First see

modules/shared/src/main/scala/org/tessellation/json/JsonBinarySerializer.scala
tbekas: Exactly, we have hardcoded encoding so `UnsupportedEncodingException` isn't going to be thrown. The default semantics for any function in our codebase is that the argument is not null, so `NullPointerException` isn't an option as well (otherwise everywhere we would have to check for nulls). Therefore this change is not needed.

PR #609, opened 2023-03-31
Loki setup for skaffold & json logger configuration

tbekas:


marcinwadon (approved):
Overall looks good. Is there a way not to enable json logs by default? It will be used by our nodes, but it will cause all the other testnet and mainnet nodes be bloated by additional logs (actually duplicated and not used).

tbekas:
@marcinwadon Good question. I will look into the easiest way of doing this. Possibly we could add a flag like `CL_ENABLE_JSON_LOGGING` that defaults to `false` but we could set it to `true` on internal nodes.

kpudlik (approved):

marcinwadon (approved):

PR #614, opened 2023-04-04
Rollback program for Currency L0

kpudlik:


tbekas (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/services/GlobalL0Service.scala
```diff
@@ -36,7 +37,6 @@ object GlobalL0Service {
   ](
     l0GlobalSnapshotClient: L0GlobalSnapshotClient[F],
     globalL0ClusterStorage: L0ClusterStorage[F],
-    lastGlobalSnapshotStorage: LastSnapshotStorage[F, GlobalIncrementalSnapshot, GlobalSnapshotInfo],
```
TheMMaciek, line 39: Removing it is a step back in my opinion. I think it's better to have `GlobalL0Service` more atomic the way it was. Allowing injecting ordinal from outside may lead to errors. This service always downloads GlobalSnapshots hence `lastGlobalSnapshotStorage` is required.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -218,7 +218,7 @@ class StateChannel[
 
   private val globalSnapshotProcessing: Stream[F, Unit] = Stream
     .awakeEvery(10.seconds)
-    .evalMap(_ => services.globalL0.pullGlobalSnapshots)
+    .evalMap(_ => storages.lastSnapshot.getOrdinal >>= services.globalL0.pullGlobalSnapshots)
```
TheMMaciek, line 221: It will now fail for the currency. The `lastSnapshot` stores a snapshot native to the state channel, so for currency it's `CurrencyIncrementalSnapshot` and for DAG it's `GlobalIncrementalSnapshot`. After this change downloading consecutive global snapshots will fail for currency cause the last ordinal is taken from `CurrencyIncrementalSnapshot` and not `GlobalIncrementalSnapshot` like it should. Because of this error I think that changes in `GlobalL0Service` should be reverted.

TheMMaciek (changes requested):
Left comments with bits I think need to be fixed.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
kpudlik: I agree. Fun fact is that it is a leftover because I am not even using `pullGlobalSnapshots` method in rollback but just `pullLatestSnapshot` directly. I am reverting it.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/services/GlobalL0Service.scala
kpudlik: Yeah. The reason why I did it previously is because I need `pullLatestSnapshot` method from `GlobalL0Service` and that method doesn't need any `lastGlobalSnapshotStorage` to be instantiated. So I needed to either:
- extract storage from `GlobalL0Service`
or - instantiate `lastGlobalSnapshotStorage` on L0 for no reason
or - extract `pullLatestSnapshot` from `GlobalL0Service` to reuse it

Of course I broke DAG which is another story.
Anyway, I will revert that and use different approach.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/http/P2PClient.scala
```diff
@@ -37,5 +37,5 @@ sealed abstract class P2PClient[F[_]] private (
   val gossip: GossipClient[F],
   val node: NodeClient[F],
   val stateChannelSnapshotClient: StateChannelSnapshotClient[F],
-  val globalSnapshot: L0GlobalSnapshotClient[F]
+  val l0GlobalSnapshotClient: L0GlobalSnapshotClient[F]
```
TheMMaciek, line 40: I think refactoring to `l0GlobalSnapshot` is enough. I think we don't add `Client` suffix in the `Client` class.

TheMMaciek (approved):

marcinwadon (approved):

kpudlik:
@marcinwadon @TheMMaciek I fixed Maciek's comment. Review please :)

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/programs/Genesis.scala
```diff
@@ -20,15 +21,18 @@ import org.tessellation.sdk.domain.collateral.{Collateral, OwnCollateralNotSatis
...
 import org.typelevel.log4cats.slf4j.Slf4jLogger
 
-trait GenesisService[F[_]] {
+trait Genesis[F[_]] {
+  def accept(path: Path): F[Unit]
   def accept(genesis: CurrencySnapshot): F[Unit]
 }
 
-object GenesisService {
+object Genesis {
```
marcinwadon, line 35: both `Service` and `Program` are part of a domain therefore should have no dependency to infrastructure. changing `GenesisService` into `GenesisProgram` makes no difference. both previous and current code has a dependency do infrastructure. if there is an easy way to fix that, please check it out. otherwise we can think about bigger refactor out of scope of this change
image

kpudlik: I agree, but unfortunately, it is a part of more extensive refactoring. Changing Service to Program doesn't fix the dependency of course but at least points out that Genesis and Rollback should be programs. We need to plan a bigger refactoring because getting rid of infrastructure dependencies there is not that trivial. I have it fixed on a different change so you will see if you are fine with the consequences of making these programs infra-independent.

PR #616, opened 2023-04-11
Select a random majority view peer to observe

cngo-github:
Observation is done on a `Peer` that is in the `Ready` state and is on the majority fork.

### DIFF
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/Consensus.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/Consensus.scala
@@ -1,7 +1,6 @@
 package org.tessellation.sdk.infrastructure.consensus
 
 import java.security.KeyPair
-
 import cats.effect.Async
 import cats.effect.std.{Random, Supervisor}
 import cats.kernel.Next
@@ -10,7 +9,6 @@ import cats.syntax.functor._
 import cats.{Eq, Order, Show}
 
 import scala.reflect.runtime.universe.TypeTag
-
 import org.tessellation.kryo.KryoSerializer
 import org.tessellation.schema.peer.PeerId
 import org.tessellation.sdk.config.types.ConsensusConfig
@@ -19,10 +17,10 @@ import org.tessellation.sdk.domain.cluster.storage.ClusterStorage
 import org.tessellation.sdk.domain.consensus.ConsensusFunctions
 import org.tessellation.sdk.domain.gossip.Gossip
 import org.tessellation.sdk.domain.node.NodeStorage
+import org.tessellation.sdk.http.p2p.clients.{GlobalSnapshotClient, L0GlobalSnapshotClient}
 import org.tessellation.sdk.infrastructure.gossip.RumorHandler
 import org.tessellation.sdk.infrastructure.metrics.Metrics
 import org.tessellation.security.SecurityProvider
-
 import io.circe.{Decoder, Encoder}
 import org.http4s.client.Client
 
@@ -65,6 +63,7 @@ object Consensus {
         gossip
       )
       consClient = ConsensusClient.make[F, Key, Artifact](client, session)
+      l0GlobalSnapshotClient = L0GlobalSnapshotClient.make(client, session)
       manager <- ConsensusManager.make[F, Event, Key, Artifact](
         consensusConfig,
         storage,
@@ -74,7 +73,8 @@ object Consensus {
         nodeStorage,
         clusterStorage,
         consClient,
-        selfId
+        selfId,
+        l0GlobalSnapshotClient
       )
       handler = ConsensusHandler.make[F, Event, Key, Artifact](storage, manager, consensusFns)
       routes = new ConsensusRoutes[F, Key, Artifact](storage)
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
@@ -10,6 +10,8 @@ import cats.syntax.all._
 import scala.concurrent.duration._
 
 import org.tessellation.ext.cats.syntax.next._
+import org.tessellation.kryo.KryoSerializer
+import org.tessellation.schema.SnapshotOrdinal
 import org.tessellation.schema.node.NodeState
 import org.tessellation.schema.node.NodeState._
 import org.tessellation.schema.peer.Peer.toP2PContext
@@ -17,6 +19,8 @@ import org.tessellation.schema.peer.{Peer, PeerId}
 import org.tessellation.sdk.config.types.ConsensusConfig
 import org.tessellation.sdk.domain.cluster.storage.ClusterStorage
 import org.tessellation.sdk.domain.node.NodeStorage
+import org.tessellation.sdk.http.p2p.clients.GlobalSnapshotClient
+import org.tessellation.sdk.domain.http.p2p.SnapshotClient
 import org.tessellation.sdk.infrastructure.consensus.message.GetConsensusOutcomeRequest
 import org.tessellation.sdk.infrastructure.consensus.trigger.{ConsensusTrigger, EventTrigger, TimeTrigger}
 import org.tessellation.sdk.infrastructure.metrics.Metrics
@@ -27,6 +31,8 @@ import eu.timepit.refined.auto._
 import org.typelevel.log4cats.slf4j.Slf4jLogger
 import retry.RetryPolicies.{constantDelay, fullJitter, limitRetries}
 import retry.syntax.all._
+import org.http4s.client.Client
+import org.tessellation.schema.GlobalSnapshot
 
 trait ConsensusManager[F[_], Key, Artifact] {
 
@@ -41,7 +47,7 @@ trait ConsensusManager[F[_], Key, Artifact] {
 
 object ConsensusManager {
 
-  def make[F[_]: Async: Random: Metrics, Event, Key: Show: Order: Next, Artifact: Eq](
+  def make[F[_]: Async: Random: Metrics: KryoSerializer, Event, Key: Show: Order: Next, Artifact: Eq](
     config: ConsensusConfig,
     consensusStorage: ConsensusStorage[F, Event, Key, Artifact],
     consensusStateCreator: ConsensusStateCreator[F, Key, Artifact],
@@ -50,7 +56,8 @@ object ConsensusManager {
     nodeStorage: NodeStorage[F],
     clusterStorage: ClusterStorage[F],
     consensusClient: ConsensusClient[F, Key, Artifact],
-    selfId: PeerId
+    selfId: PeerId,
+    snapshotClient: SnapshotClient[F, GlobalSnapshot],
   )(implicit S: Supervisor[F]): F[ConsensusManager[F, Key, Artifact]] = {
     val logger = Slf4jLogger.getLoggerFromClass[F](ConsensusManager.getClass)
 
@@ -78,7 +85,9 @@ object ConsensusManager {
           def getObservedOutcome: F[ConsensusOutcome[Key, Artifact]] = for {
             readyPeers <- clusterStorage.getResponsivePeers
               .map(_.filter(_.state === Ready))
-            selectedPeer <- Random[F].elementOf(readyPeers)
+            peersToQuery <- getPeerSublist(readyPeers.toList)
+            peerCandidates <- filterPeerList(peersToQuery)
+            selectedPeer <- Random[F].elementOf(peerCandidates)
             observedOutcome <- observePeer(selectedPeer)
           } yield observedOutcome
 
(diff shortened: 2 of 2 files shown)


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -177,6 +190,28 @@ object ConsensusManager {
             .handleErrorWith(logger.error(_)(s"Error checking for consensus state update {key=${key.show}}"))
         }.void
 
+      private def getPeerCandidates(peers: List[Peer]): F[List[Peer]] = {
+        case class PeerSnapshot(ordinal: SnapshotOrdinal, hash: Hash, peer: Peer)
+        val groupFn: PeerSnapshot => (SnapshotOrdinal, Hash) = v => (v.ordinal, v.hash)
+
+        for {
```
marcinwadon, line 197: You have a single flatMap, and then all the other assignments are just pure values. You don't need a for comprehension for that. Also, there is no need to specify each and every type within the for comprehension.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -177,6 +190,28 @@ object ConsensusManager {
             .handleErrorWith(logger.error(_)(s"Error checking for consensus state update {key=${key.show}}"))
         }.void
 
+      private def getPeerCandidates(peers: List[Peer]): F[List[Peer]] = {
+        case class PeerSnapshot(ordinal: SnapshotOrdinal, hash: Hash, peer: Peer)
```
marcinwadon, line 194: It can be a named tuple. You don't use it outside of this fn anyway

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -77,7 +81,16 @@ object ConsensusManager {
           def getObservedOutcome: F[ConsensusOutcome[Key, Artifact]] = for {
             readyPeers <- clusterStorage.getResponsivePeers
               .map(_.filter(_.state === Ready))
-            selectedPeer <- Random[F].elementOf(readyPeers)
+            numberOfPeersToQuery <- Random[F].nextIntBounded(readyPeers.size / 4 + 1)
```
marcinwadon, line 84: What's that magic calculation? Every magic number or calculation should be named

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -77,7 +81,16 @@ object ConsensusManager {
...
+            peersToQuery <- Random[F]
```
marcinwadon, line 85: If you rely on the majority, why you don't query all the peers. If I understood this correctly, by querying a subset of peers, you'll build up wrong majority observation.

marcinwadon (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -77,7 +81,16 @@ object ConsensusManager {
...
+              .shuffleList(readyPeers.toList)
+              .map(
```
marcinwadon, line 87: See @TheMMaciek comment about multiline formatting. It's better to use { instead of ( in that case.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -77,7 +81,16 @@ object ConsensusManager {
           def getObservedOutcome: F[ConsensusOutcome[Key, Artifact]] = for {
             readyPeers <- clusterStorage.getResponsivePeers
               .map(_.filter(_.state === Ready))
-            selectedPeer <- Random[F].elementOf(readyPeers)
+            numberOfPeersToQuery <- Random[F].nextIntBounded(readyPeers.size / 4 + 1)
+            peersToQuery <- Random[F]
+              .shuffleList(readyPeers.toList)
```
tbekas, line 86: Do you realize, that 99% of the time (or more) `readyPeers` will be a collection of size 1?

All this logic isn't going to do anything, except for a very rare case when the single node that we joined is going to fail while we're observing it.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -177,6 +190,28 @@ object ConsensusManager {
...
         }.void
 
+      private def getPeerCandidates(peers: List[Peer]): F[List[Peer]] = {
+        case class PeerSnapshot(ordinal: SnapshotOrdinal, hash: Hash, peer: Peer)
+        val groupFn: PeerSnapshot => (SnapshotOrdinal, Hash) = v => (v.ordinal, v.hash)
+
+        for {
+          peerSnaps: List[PeerSnapshot] <- peers.toList.traverse(peer =>
+            for {
+              signedSnapshot <- gsClient.getLatest(peer)
```
tbekas, line 200: `getLatest` as the name suggests gets the outcome of the latest consensus, what's the point of comparing hashes of different snapshots?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -177,6 +190,28 @@ object ConsensusManager {
             .handleErrorWith(logger.error(_)(s"Error checking for consensus state update {key=${key.show}}"))
         }.void
 
+      private def getPeerCandidates(peers: List[Peer]): F[List[Peer]] = {
```
tbekas, line 193: `candidates` is already used in consensus in the context of peers that are going to be added as facilitators to the next consensus. Please use a different name to avoid confusion.

tbekas (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
cngo-github: There is no magic calculation. We want to sample a small subset of peers. So, I'm taking at most 25% of the ready `Peer`s and at least 1 ready `Peer`. There is no rhyme or reason for the 25% number. It just seemed like a good number.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
cngo-github: Changed.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
cngo-github: The idea is to determine the majority branch. If there are `n` `Peer`s where `n > 1`, not all the `Peer`s may agree on the latest snapshot (i.e. they are on different forks). We compare the hash and ordinals to determine what the majority branch is so that we can down against one of those `Peer`s.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
cngo-github: Changed.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
ryle-goehausen: If the short term, querying all the peers wouldn't cause much of an issue, but it's not ideal at much higher node counts. We've made a few other tickets for adding a random sampling by weighted trust to ensure nodes are sampled with respect to overall score -- and additionally to observe the gossip stream to collect data -- so planned improvements to this would be done separately (ideally after the initial single peer join process is improved). If every node restart incurs a bulk broadcast to all nodes, it's not really great. Furthermore, starting from now this is at least a minor improvement to dealing with a random peer.

Extending this to the full list would be okay in short term however, but not critical

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
ryle-goehausen: @cngo-github we need to ensure this process runs after peer discovery has populated peers, but before download has occurred, this is important change to make here so we should ensure the peer list has been populated by discovery process BEFORE attempting this.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
ryle-goehausen: There's no logic here to handle the case of timing differences with respect to ordinal, but I think Chuong has made another ticket to capture this. I.e. if ordinal 1000 and 1001 are returned by different queries (but are otherwise on the same fork), for now we're ignoring that and just capturing majority of both.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
cngo-github: Please elucidate. Aren't named tuples in Scala case classes?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
marcinwadon: First search from stackoverflow: https://stackoverflow.com/a/47890

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
marcinwadon:
```scala
type PeerSnapshot = (SnapshotOrdinal, Hash)
```

you create a case class to convert it to tuple anyway

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
marcinwadon: > The idea is to determine the majority branch. If there are n Peers where n > 1, not all the Peers may agree on the latest snapshot (i.e. they are on different forks).

Different case is when there are ordinals at the same ordinal number but with different hashes (nodes are misaligned and indeed in different forks) and different case is when you compare hash of snapshot at ordinal 10 and hash of snapshot at ordinal 11 - they will be different even if both nodes are in the same fork and are aligned.

@ryle-goehausen I see no point in introducing issues or partially working code and fixing it later. (exception for this is when we merge it to a feature branch or agree to somehow delay deployments on mainnet)

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
cngo-github: Changed.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
tbekas: I agree with @marcinwadon . Submitting a broken functionality only to create a technical debt, which needs to be addressed later on is a bad idea.  Please fix it, comparing hashes of snapshots with different ordinals is simply pointless.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
cngo-github: The process of sub-selecting the ready peers has been moved to and encapsulated in a function.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
marcinwadon: @cngo-github that's why we use pattern matching, for example:
image

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -178,6 +184,31 @@ object ConsensusManager {
...
+              signedSnapshot.toHashed.map(_.hash).map(PeerSnapshot(signedSnapshot.value.ordinal, _, peer))
+            }
+          )
+          .map(_.groupBy(groupFn))
+          .map(_.maxBy(_._2.size)._2.map(_.peer))
+      }
+
+      private def getPeerSublist(peers: List[Peer]) = {
+        val maxSublistPercent = 0.25
+        val maxSublistSize = (peers.size * maxSublistPercent).toInt + 1
```
kpudlik, line 203: So you want at most `0.25` of the `peers.size` but at least `1` as you wrote in the comment:
https://github.com/Constellation-Labs/tessellation/pull/616#discussion_r1164234646

According to that statement for `peer.size = 8` it should return `2` but your logic returns `3` which is not `0.25` but `0.375`.

To make the logic working just like you explained you should rather 
```scala
val maxSublistSize = Math.max((peers.size * maxSublistPercent).toInt, 1)
```

kpudlik (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
ryle-goehausen: We have a separate ticket for @cngo-github to implement the process of determining latest ordinal. Unless I am mistaken this PR is strictly an improvement upon the prior implementation (which might be referred to as technical debt that already exists, or a bug that is already in the code which this is not making worse.) This should be an incremental improvement upon the existing functionality by at least introducing some check of multiple peers (even if not accounting for ordinals.) Since we already have a ticket to fix the ordinal and you're not okay with splitting up these PRs into separate items, I'll discuss with Chuong the next steps to fixing the ordinal query as part of this PR.

cngo-github: I will implement the process of determining the latest ordinal as part of this PR.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
cngo-github: Changed.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/Consensus.scala
```diff
@@ -74,7 +75,8 @@ object Consensus {
         nodeStorage,
         clusterStorage,
         consClient,
-        selfId
+        selfId,
+        GlobalSnapshotClient.make(client, session)
```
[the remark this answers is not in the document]
cngo-github: Changed.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
tbekas: Ok, so @cngo-github what changes to the flow are you planning to introduce to make the implementation aligned with what @ryle-goehausen described?

PR #618, opened 2023-04-17
Parametrize the last full global snapshot ordinal

marcinwadon:


modules/core/src/main/scala/org/tessellation/cli/incremental.scala
```diff
@@ -0,0 +1,23 @@
...
+import eu.timepit.refined.types.numeric.NonNegLong
+
+object incremental {
+  val lastFullGlobalSnapshot: Map[AppEnvironment, SnapshotOrdinal] = Map(
+    Mainnet -> SnapshotOrdinal(0L), // TODO: set before mainnet release
+    Testnet -> SnapshotOrdinal(736766L)
+  )
+
+  val lastFullGlobalSnapshotOpts: Opts[Option[SnapshotOrdinal]] = Opts
+    .argument[NonNegLong]("lastFullGlobalSnapshot")
```
kpudlik, line 20: `lastFullGlobalSnapshotOrdinal` because someone can try to put `hash` there especially that we put hashes for rollback

modules/core/src/main/scala/org/tessellation/cli/incremental.scala
```diff
@@ -0,0 +1,23 @@
...
+    .map(SnapshotOrdinal(_))
+    .orNone
```
kpudlik, line 22: `orNone` + `getOrElse` is technically `.withDefault(SnapshotOrdinal(0L)` but I see that defaulting doesn't make sense for `AppEnvironment.Dev` so let's leave it as it is. Just wanted to point out `withDefault` method for future use-cases.

modules/core/src/main/scala/org/tessellation/cli/method.scala
```diff
@@ -140,8 +144,27 @@ object method {
...
+              http,
+              environment,
+              snapshot,
+              seedlistPath,
+              collateralAmount,
+              rollbackHash,
+              lastGlobalSnapshot
+            ) =>
+          val lastGS =
+            (if (environment === AppEnvironment.Dev) lastGlobalSnapshot else lastFullGlobalSnapshot.get(environment)).getOrElse(SnapshotOrdinal(0L))
```
kpudlik, line 164: Maybe `SnapshotOrdinal.MinValue`?

kpudlik (changes requested):

modules/core/src/main/scala/org/tessellation/cli/method.scala
marcinwadon: :+1:

modules/core/src/main/scala/org/tessellation/cli/incremental.scala
marcinwadon: right :+1: fixed

kpudlik (approved):

PR #619, opened 2023-04-18
Seedlist for state channels

TheMMaciek:


kubernetes/data/state-channel-seedlist.csv
```diff
@@ -0,0 +1,10 @@
+DAG7QrKeoHbYezJ6YHUCt7PJDixWgxM4vfKz8qVT
+DAG3FJFFsjzYeSNVeL3cykLqWbijkq25K6Eg23yp
+DAG5kq5J6gudAbyGsgqPReGz8ZDRM94UoPwCFSem
+DAG3ysjCYZCWoTw3SKoK13McfdA5LZWYBr7ue47d
+DAG7cCCNgP8D5rdEznhSNXpAfQ2KhuScxhg1YhVX
+DAG2T7jB5kXpQMEpDErB2NcoQfeJP7DwbueK3A5i
+DAG16Q8MxAvihBbFk8wuSaCVzieq75zZA3mh414r
+DAG6wD9h8NE6qDuVT8ZtJFaU3ZzD9BqBjWD8GDee
+DAG0R1CMwfLPNp4qkA2TTqkcKWp3mC7U4ZMD5n1B
+DAG7gcb396YoLVzNhCEF2nrB79uwV3wr4BvXSU9f
```
marcinwadon, line 10: are you sure it should be a DAGAddress, not a node id (so the public key)?

modules/shared/src/test/resources/logback-test.xml
marcinwadon, line 1: is it a leftover or is it required?

modules/core/src/main/scala/org/tessellation/cli/method.scala
```diff
@@ -122,6 +124,7 @@ object method {
     environment: AppEnvironment,
     snapshotConfig: SnapshotConfig,
     seedlistPath: Option[Path],
+    stateChannelSeedlistPath: Path,
```
kpudlik, line 127: Isn't it `seedListPath === stateChannelSeedlistPath` for Global L0?

kpudlik (changes requested):

kubernetes/data/state-channel-seedlist.csv
TheMMaciek: Yes it needs to be a DAGAddress. Currently it's derived from public key, but we will add functionality to derive the address from state channel genesis.

modules/core/src/main/scala/org/tessellation/cli/method.scala
TheMMaciek: No, state channel seedlist path is a list of DAGAddresses.

modules/shared/src/test/resources/logback-test.xml
TheMMaciek: It disables logs in tests hence makes test runtime more readable.

modules/sdk/src/main/scala/org/tessellation/sdk/cli/StateChannelSeedlist.scala
```diff
@@ -0,0 +1,12 @@
...
+
+import org.tessellation.ext.decline.decline._
+
+import com.monovore.decline.Opts
+import fs2.io.file.Path
+
+object StateChannelSeedlist {
+
+  val opts: Opts[Path] = Opts
+    .option[Path]("state-channel-seedlist", "List of allowed state channel addresses")
```
tbekas, line 11: Consider adding the env variable as well

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/genesis/Loader.scala
```diff
@@ -21,7 +21,7 @@ object Loader {
           decodeWithoutHeaders[GenesisCSVAccount]()
         )
         .map(_.toGenesisAccount)
-        .map(_.bimap(e => new RuntimeException(e), identity))
+        .map(_.leftMap(new RuntimeException(_)))
```
tbekas, line 24: 👍

modules/core/src/main/scala/org/tessellation/cli/method.scala
kpudlik: But isn't it still `seedList.map(_.toAddress) === stateChannelSeedlist` for Global L0?

tbekas (dismissed):

marcinwadon (approved):

kpudlik (dismissed):

modules/shared/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,11 @@
+<configuration>
```
marcinwadon, line 1: why is it needed?

marcinwadon (dismissed):
LGTM

kriscieslak (approved):

PR #621, opened 2023-04-19
Remove infrastructure dependency from Rollback and Genesis

kpudlik:
…atible with onion architecture

### DIFF
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/domain/consensus/ConsensusManager.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/domain/consensus/ConsensusManager.scala
@@ -1,0 +1,15 @@
+package org.tessellation.sdk.domain.consensus
+
+import org.tessellation.sdk.infrastructure.consensus.ConsensusResources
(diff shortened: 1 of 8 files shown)


modules/sdk/src/main/scala/org/tessellation/sdk/domain/consensus/ConsensusManager.scala
```diff
@@ -0,0 +1,15 @@
...
+import org.tessellation.security.signature.Signed
+
+trait ConsensusManager[F[_], Key, Artifact, Context] {
+
+  def startObserving: F[Unit]
+  def startFacilitatingAfter(lastKey: Key, lastArtifact: Signed[Artifact], lastContext: Context): F[Unit]
+
+  def withdrawFromConsensus: F[Unit]
+  def facilitateOnEvent: F[Unit]
+  def checkForStateUpdate(key: Key)(resources: ConsensusResources[Artifact]): F[Unit]
```
kpudlik, line 13: These methods were `private[consensus]` and I needed to make them public so as to be available in `infrastructure/consensus`. Let me know if you are comfortable with it.

PR #623, opened 2023-04-19
Minimize Ref accesses in PeerDiscovery.discoverFrom

sdavidp:


marcinwadon (approved):
I'm good with that change but why is it needed? Instead of adding peers one by one, it updates the cache in batch, right?

sdavidp:
> I'm good with that change but why is it needed? Instead of adding peers one by one, it updates the cache in batch, right?

Yes, that's one benefit. The intended improvement is to:

- reduce the number of times we enter the `Ref` lock
- simplify the `filterNot` predicate and eliminate temporary collection creation
- reduce the number of temporary objects created on update (Depends on implementation of `Set.concat`)
- use `Ref.update` functions (which are implemented in terms of `modify`)

One thing I wondered about was whether it was important to get the peers after the call to `clusterClient.getDiscoveryPeers`. I don't think it is but want to be sure.

For context, I was looking at this code when exploring a possible solution to the peer discovery task Chuong is working on.

sdavidp:
> One thing I wondered about was whether it was important to get the peers after the call to `clusterClient.getDiscoveryPeers`. I don't think it is but want to be sure.

In fact, now that I'm looking at it with fresh eyes,

- if it's important we can move it after the call to `clusterClient.getDiscoveryPeers`
- we're not using the `peersQueue` collection anyway, we're only looking up the `id`s in the predicate, so we can simplify it a bit

PR #629, opened 2023-04-25
Add delay to download to discover peers

cngo-github:
Added a delay to the download process to block the start of observation until peer discovery had been given time to run and discover some peers.

PR #630, opened 2023-04-25
Observe majority fork

cngo-github:
When choosing which peer to observe, select a peer on the majority fork. This PR implements the code needed to select a peer on the majority fork for observation. It supersedes PR https://github.com/Constellation-Labs/tessellation/pull/616 and is related to PR https://github.com/Constellation-Labs/tessellation/pull/629.

### DIFF
--- a/modules/core/src/main/scala/org/tessellation/http/p2p/P2PClient.scala
+++ b/modules/core/src/main/scala/org/tessellation/http/p2p/P2PClient.scala
@@ -13,7 +13,8 @@ object P2PClient {
       sdkP2PClient.sign,
       sdkP2PClient.cluster,
       sdkP2PClient.gossip,
-      sdkP2PClient.node
+      sdkP2PClient.node,
+      sdkP2PClient.l0GlobalSnapshotClient
     ) {}
 }
 
@@ -21,5 +22,6 @@ sealed abstract class P2PClient[F[_]] private (
   val sign: SignClient[F],
   val cluster: ClusterClient[F],
   val gossip: GossipClient[F],
-  val node: NodeClient[F]
+  val node: NodeClient[F],
+  val l0GlobalSnapshotClient: L0GlobalSnapshotClient[F]
 )
--- a/modules/core/src/main/scala/org/tessellation/modules/Programs.scala
+++ b/modules/core/src/main/scala/org/tessellation/modules/Programs.scala
@@ -3,30 +3,37 @@ package org.tessellation.modules
 import java.security.KeyPair
 
 import cats.effect.Async
+import cats.effect.std.Random
 
 import org.tessellation.config.types.AppConfig
 import org.tessellation.domain.cluster.programs.TrustPush
 import org.tessellation.infrastructure.snapshot.programs.RollbackLoader
 import org.tessellation.kryo.KryoSerializer
 import org.tessellation.sdk.domain.cluster.programs.{Joining, PeerDiscovery}
+import org.tessellation.sdk.domain.snapshot.PeerSelect
+import org.tessellation.sdk.http.p2p.clients.L0GlobalSnapshotClient
+import org.tessellation.sdk.infrastructure.snapshot.MajorityPeerSelect
 import org.tessellation.sdk.infrastructure.snapshot.programs.Download
 import org.tessellation.sdk.modules.SdkPrograms
 import org.tessellation.security.SecurityProvider
 
 object Programs {
 
-  def make[F[_]: Async: KryoSerializer: SecurityProvider](
+  def make[F[_]: Async: KryoSerializer: SecurityProvider: Random](
     sdkPrograms: SdkPrograms[F],
     storages: Storages[F],
     services: Services[F],
     keyPair: KeyPair,
-    config: AppConfig
+    config: AppConfig,
+    l0GlobalSnapshotClient: L0GlobalSnapshotClient[F]
   ): Programs[F] = {
     val trustPush = TrustPush.make(storages.trust, services.gossip)
+    val peerSelect: PeerSelect[F] = MajorityPeerSelect.make(storages.cluster, l0GlobalSnapshotClient)
     val download = Download
       .make(
         storages.node,
-        services.consensus
+        services.consensus,
+        peerSelect
       )
     val rollbackLoader = RollbackLoader.make(
       keyPair,
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/SdkP2PClient.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/SdkP2PClient.scala
@@ -18,7 +18,8 @@ object SdkP2PClient {
       ClusterClient.make[F](client, session),
       GossipClient.make[F](client, session),
       NodeClient.make[F](client, session),
-      TrustClient.make[F](client, session)
+      TrustClient.make[F](client, session),
+      L0GlobalSnapshotClient.make[F](client)
     ) {}
 
 }
@@ -28,5 +29,6 @@ sealed abstract class SdkP2PClient[F[_]] private (
   val cluster: ClusterClient[F],
   val gossip: GossipClient[F],
   val node: NodeClient[F],
-  val trust: TrustClient[F]
+  val trust: TrustClient[F],
+  val l0GlobalSnapshotClient: L0GlobalSnapshotClient[F]
 )
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
@@ -3,7 +3,7 @@ package org.tessellation.sdk.infrastructure.consensus
 import cats._
 import cats.data.Ior.{Both, Right}
 import cats.effect._
-import cats.effect.std.{Random, Supervisor}
+import cats.effect.std.Supervisor
 import cats.kernel.Next
 import cats.syntax.all._
 
@@ -31,7 +31,7 @@ import retry.syntax.all._
 
 object ConsensusManager {
 
-  def make[F[_]: Async: Random: Metrics, Event, Key: Show: Order: Next, Artifact: Eq, Context: Eq](
+  def make[F[_]: Async: Metrics, Event, Key: Show: Order: Next, Artifact: Eq, Context: Eq](
     config: ConsensusConfig,
     consensusStorage: ConsensusStorage[F, Event, Key, Artifact, Context],
     consensusStateCreator: ConsensusStateCreator[F, Key, Artifact, Context],
@@ -61,17 +61,9 @@ object ConsensusManager {
 
     val manager = new ConsensusManager[F, Key, Artifact, Context] {
 
-      def startObserving: F[Unit] =
+      def startObserving(peerToObserve: Peer): F[Unit] =
         S.supervise {
-
-          def getObservedOutcome: F[ConsensusOutcome[Key, Artifact, Context]] = for {
-            readyPeers <- clusterStorage.getResponsivePeers
-              .map(_.filter(_.state === Ready))
-            selectedPeer <- Random[F].elementOf(readyPeers)
-            observedOutcome <- observePeer(selectedPeer)
-          } yield observedOutcome
-
-          getObservedOutcome
+          observePeer(peerToObserve)
             .retryingOnAllErrors(
               observationRetryPolicy,
               (err, retryDetails) =>
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
@@ -1,0 +1,77 @@
+package org.tessellation.sdk.infrastructure.snapshot
+
+import cats.data.NonEmptyList
+import cats.effect.Async
+import cats.effect.std.Random
+import cats.syntax.applicative._
+import cats.syntax.applicativeError._
+import cats.syntax.eq._
+import cats.syntax.flatMap._
+import cats.syntax.functor._
+import cats.syntax.list._
+
+import scala.util.control.NoStackTrace
+
+import org.tessellation.kryo.KryoSerializer
+import org.tessellation.schema.node.NodeState.Ready
+import org.tessellation.schema.peer.Peer
+import org.tessellation.schema.peer.Peer.toP2PContext
+import org.tessellation.sdk.domain.cluster.storage.ClusterStorage
+import org.tessellation.sdk.domain.snapshot.PeerSelect
+import org.tessellation.sdk.http.p2p.clients.L0GlobalSnapshotClient
+
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+
+object MajorityPeerSelect {
+
+  case object NoPeersToSelect extends NoStackTrace
+
+  def make[F[_]: Async: KryoSerializer: Random](
+    storage: ClusterStorage[F],
+    snapshotClient: L0GlobalSnapshotClient[F]
+  ): PeerSelect[F] = new PeerSelect[F] {
+
+    val logger = Slf4jLogger.getLoggerFromClass[F](MajorityPeerSelect.getClass)
+
+    def select: F[Peer] = storage.getResponsivePeers
+      .map(_.filter(_.state === Ready))
+      .flatMap(getPeerSublist)
+      .flatMap { peers =>
+        peers.toNel match {
+          case Some(value) => value.pure[F]
+          case None =>
+            logger.error("No Ready peers were found to be selected.") >>
+              NoPeersToSelect.raiseError[F, NonEmptyList[Peer]]
+        }
+      }
+      .flatMap(filterPeerList)
+      .map(_.toList)
+      .flatMap(Random[F].elementOf)
+
+    def filterPeerList(peers: NonEmptyList[Peer]): F[NonEmptyList[Peer]] =
+      peers
+        .traverse(snapshotClient.getLatestOrdinal(_))
+        .map {
+          _.groupBy(identity).maxBy { case (_, ordinals) => ordinals.size }
+        }
+        .flatMap {
+          case (majorityOrdinal, _) =>
+            peers.traverse(snapshotClient.get(majorityOrdinal).run(_).flatMap(_.toHashed.map(_.hash)))
+        }
+        .map(_.zip(peers))
+        .map(_.groupMap { case (hash, _) => hash } { case (_, ps) => ps })
+        .map(_.maxBy { case (_, peers) => peers.size })
+        .map { case (_, peerCandidates) => peerCandidates }
+
+    def getPeerSublist(peers: Set[Peer]): F[List[Peer]] = {
+      val maxSublistPercent = 0.25
+      val maxSublistSize = Math.max((peers.size * maxSublistPercent).toInt, 1)
+
+      Random[F].nextIntBounded(maxSublistSize).flatMap { peerCount =>
+        Random[F]
+          .shuffleList(peers.toList)
+          .map(_.take(peerCount))
+      }
+    }
+  }
+}
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/programs/Download.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/programs/Download.scala
@@ -5,26 +5,30 @@ import cats.syntax.flatMap._
 
 import org.tessellation.schema.node.NodeState
 import org.tessellation.sdk.domain.node.NodeStorage
+import org.tessellation.sdk.domain.snapshot.PeerSelect
 import org.tessellation.sdk.infrastructure.snapshot.SnapshotConsensus
 
 object Download {
 
   def make[F[_]: Async](
     nodeStorage: NodeStorage[F],
-    consensus: SnapshotConsensus[F, _, _, _, _, _]
+    consensus: SnapshotConsensus[F, _, _, _, _, _],
+    peerSelecter: PeerSelect[F]
   ): Download[F] =
     new Download[F](
       nodeStorage,
-      consensus
+      consensus,
+      peerSelecter
     ) {}
 }
 
 sealed abstract class Download[F[_]: Async] private (
   nodeStorage: NodeStorage[F],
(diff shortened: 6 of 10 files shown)


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -102,6 +107,32 @@ object ConsensusManager {
             }
         }.void
 
+      private def filterPeerList(peers: List[Peer]): F[List[Peer]] = {
+        val majorityOrdinal = peers
+          .traverse(snapshotClient.getLatestOrdinal(_))
+          .map(_.groupBy(ordinal => ordinal))
+          .map(_.maxBy(_._2.size)._1)
```
marcinwadon, line 114: what is `_2` or `_1`? please use meaningful names, like:
```
{
 case (foo, bar) => foo.size
```

marcinwadon (commented):
In general, this should be done for download, not for observing 🤔 
The download process should already be pointed to a correct fork, observing is the very next step

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -102,6 +107,32 @@ object ConsensusManager {
             }
         }.void
 
+      private def filterPeerList(peers: List[Peer]): F[List[Peer]] = {
+        val majorityOrdinal = peers
+          .traverse(snapshotClient.getLatestOrdinal(_))
+          .map(_.groupBy(ordinal => ordinal))
```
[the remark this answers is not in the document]
cngo-github: Changed.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -102,6 +107,37 @@ object ConsensusManager {
...
+      private def filterPeerList(peers: List[Peer]): F[List[Peer]] = {
+        val majorityOrdinal = peers
+          .traverse(snapshotClient.getLatestOrdinal(_))
+          .map(_.groupBy(identity))
+          .map { ordinalMap =>
+            ordinalMap.maxBy {
+              case (_, ordinalList) =>
+                ordinalList.size
+            }._1
+          }
```
marcinwadon, line 119: you still use `_1`.

I'd suggest:
```scala
        val majorityOrdinal = peers
          .traverse(snapshotClient.getLatestOrdinal(_))
          .map(_.groupBy(identity))
          .map { 
            _.maxBy {
              case (_, ordinalList) =>
                ordinalList.size
            }
          }.map {
            case (ordinal, _) => ordinal
          }

```

cngo-github: Changed.

ryle-goehausen:
> image
> 
> `maxBy` is unsafe operation (cannot be executed on empty map/list). Please use safer version from cats that fallbacks to Option (None) for empty case.

Regarding this, I think we should also abort if we can't find any peer responses at any stage here and throw an error indicative that no peers were found to make it even clearer (in addition to option here)

ryle-goehausen:
> In general, this should be done for download, not for observing 🤔 The download process should already be pointed to a correct fork, observing is the very next step

The reason for this was mainly intercepting the code that does the random shuffle associated with selecting a peer during observation. I'm actually okay with moving these changes over to the download, so long as we're modifying what peers the consensus observer accepts (so we need to modify that list to guarantee all of them are correct candidates matching the appropriate hash.) @cngo-github  how much do you estimate making this change is?

marcinwadon:
> > In general, this should be done for download, not for observing 🤔 The download process should already be pointed to a correct fork, observing is the very next step
> 
> The reason for this was mainly intercepting the code that does the random shuffle associated with selecting a peer during observation. I'm actually okay with moving these changes over to the download, so long as we're modifying what peers the consensus observer accepts (so we need to modify that list to guarantee all of them are correct candidates matching the appropriate hash.) @cngo-github how much do you estimate making this change is?

Sure. Then probably it would have to be implemented in both download and consensus (I'm not sure yet) but download is the very first step when node decides which fork it should download (so here should be some majority calculation instead of random select). Then, when node filled all the gaps, it selects the node to observe (and here as implemented, it should select the node again with majority calculation)

cngo-github:
> > > In general, this should be done for download, not for observing 🤔 The download process should already be pointed to a correct fork, observing is the very next step
> > 
> > 
> > The reason for this was mainly intercepting the code that does the random shuffle associated with selecting a peer during observation. I'm actually okay with moving these changes over to the download, so long as we're modifying what peers the consensus observer accepts (so we need to modify that list to guarantee all of them are correct candidates matching the appropriate hash.) @cngo-github how much do you estimate making this change is?
> 
> Sure. Then probably it would have to be implemented in both download and consensus (I'm not sure yet) but download is the very first step when node decides which fork it should download (so here should be some majority calculation instead of random select). Then, when node filled all the gaps, it selects the node to observe (and here as implemented, it should select the node again with majority calculation)

I'll work on the this and let you know about a time estimate.

cngo-github:
@ryle-goehausen @marcinwadon , I have moved the peer selection code out of the observation code as requested. Please review.

cngo-github:
> image
> 
> `maxBy` is unsafe operation (cannot be executed on empty map/list). Please use safer version from cats that fallbacks to Option (None) for empty case.

The `List` has been changed to `NonEmptyList`.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelectionStrategy.scala
```diff
@@ -0,0 +1,53 @@
...
+) extends PeerSelectionStrategy[F] {
+  def select: F[Peer] = for {
+    readyPeers <- storage.getResponsivePeers
+      .map(_.filter(_.state === Ready))
+    peersToQuery <- getPeerSublist(readyPeers).flatMap { peers =>
+      Async[F].fromOption(peers.toNel, NoPeers)
+    }
+    peerCandidates <- filterPeerList(peersToQuery)
+    selectedPeer <- Random[F].elementOf(peerCandidates.toList)
+  } yield selectedPeer
```
marcinwadon, line 28: in general, you just create next variables based on the previous ones, so for comprehension is imo less readable. please consider this:

```scala
storage.getResponsivePeers
  .flatMap(getPeerSublist)
  .flatMap(peers => ApplicativeError[F].fromOption(peers.toNel, NoPeers))
  .flatMap(filterPeerList)
  .flatMap(Random[F].elementOf(_))
```

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelectionStrategy.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  private def filterPeerList(peers: NonEmptyList[Peer]): F[NonEmptyList[Peer]] =
+    for {
```
marcinwadon, line 31: as above

marcinwadon (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelectionStrategy.scala
```diff
@@ -0,0 +1,53 @@
...
+
+class MajorityPeerSelectionStrategy[F[_]: Async: KryoSerializer: Random](
+  val storage: ClusterStorage[F],
+  val snapshotClient: L0GlobalSnapshotClient[F]
+) extends PeerSelectionStrategy[F] {
+  def select: F[Peer] = for {
+    readyPeers <- storage.getResponsivePeers
+      .map(_.filter(_.state === Ready))
+    peersToQuery <- getPeerSublist(readyPeers).flatMap { peers =>
+      Async[F].fromOption(peers.toNel, NoPeers)
```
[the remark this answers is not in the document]
sdavidp: @marcinwadon Couldn't an argument be made that `Async` is the right typeclass to use because we'd then have to specify the error type parameter: doesn't that have to match the one defined in `Async[F] <- Sync[F] <- MonadCancel[F, Throwable] <- ...` ?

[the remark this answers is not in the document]
marcinwadon: but then you don't use the least power. You can use Async everywhere but then tagless algebra makes no sense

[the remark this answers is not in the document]
sdavidp: I agree, but that's not what I'm saying. I'm asking whether or not there is a dependency on us using the same error type as `Async` is using for the ApplicativeError. If there is, then we should just go through the class we're using as the constraint. If there is not, then we can change it to the defining class.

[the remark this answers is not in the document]
marcinwadon: Could you show an example? I don't fully understand what you mean

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -30,7 +31,7 @@ import retry.syntax.all._
 
 trait ConsensusManager[F[_], Key, Artifact, Context] {
 
-  def startObserving: F[Unit]
+  def startObserving(selectedPeer: Peer): F[Unit]
```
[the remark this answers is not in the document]
cngo-github: Changed.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelectionStrategy.scala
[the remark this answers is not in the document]
sdavidp: What I want to understand is whether we have a dependency on the defined constraint. In this code it's `Async`.

For example, given Sync is defined like this,
```
trait Sync[F[_]] extends MonadCancel[F, Throwable] with Clock[F] with Unique[F] with Defer[F]
```
here are two ways to call `fromOption`
```
Sync[F].fromOption(...)  // uses error type parameter defined in Sync
ApplicativeError[F, E].fromOption(...)
```
What should we set `E` to?  If `Throwable`, then why? Is it because our function requires evidence of `Sync[F]`? If so then we should just use `Sync` otherwise we're assuming we have knowledge of the implementation.

If we can set `E` to anything else and not change the correctness or get compiler error, then `ApplicativeError` is fine.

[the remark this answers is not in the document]
sdavidp: In other words, does the `E` we use in `ApplicativeError` have to match the type parameter used in the typeclass we're constrained against, which in this case is `Async`.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelectionStrategy.scala
cngo-github: Changed.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelectionStrategy.scala
cngo-github: Changed.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelectionStrategy.scala
```diff
@@ -0,0 +1,53 @@
...
+
+import org.tessellation.kryo.KryoSerializer
+import org.tessellation.schema.node.NodeState.Ready
+import org.tessellation.schema.peer.Peer
+import org.tessellation.schema.peer.Peer.toP2PContext
+import org.tessellation.sdk.domain.cluster.storage.ClusterStorage
+import org.tessellation.sdk.domain.snapshot.peerSelection.{NoPeers, PeerSelectionStrategy}
+import org.tessellation.sdk.http.p2p.clients.L0GlobalSnapshotClient
+
+class MajorityPeerSelectionStrategy[F[_]: Async: KryoSerializer: Random](
```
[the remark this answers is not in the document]
cngo-github: Changed.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/peerSelection.scala
```diff
@@ -0,0 +1,13 @@
+package org.tessellation.sdk.domain.snapshot
+
+import scala.util.control.NoStackTrace
+
+import org.tessellation.schema.peer.Peer
+
+object peerSelection {
+  trait PeerSelectionStrategy[F[_]] {
+    def select: F[Peer]
+  }
```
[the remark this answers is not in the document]
cngo-github: The trait is renamed `PeerSelection`, as well as the file, and the object has been removed. `NoPeers` was moved to the `MajorityPeerSelection` object.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelectionStrategy.scala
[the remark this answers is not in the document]
marcinwadon: we use `MonadThrow` in many places which is an alias that sets `E` to `Throwable`.
I'm good also to use `MonadThrow[F].fromOption(..)` but effectively using `MonadThrow`/`Sync`/`Async` ends up with `ApplicativeError[F, Throwable]`.
when answering to `why` - `Throwable` is the most generic type for errors. but if you want, I think you can pass more specified type there

modules/core/src/main/scala/org/tessellation/modules/Programs.scala
```diff
@@ -3,30 +3,39 @@ package org.tessellation.modules
...
     sdkPrograms: SdkPrograms[F],
     storages: Storages[F],
     services: Services[F],
     keyPair: KeyPair,
-    config: AppConfig
+    config: AppConfig,
+    client: Client[F]
   ): Programs[F] = {
     val trustPush = TrustPush.make(storages.trust, services.gossip)
+    val peerSelector: PeerSelection[F] = MajorityPeerSelection.make(storages.cluster, L0GlobalSnapshotClient.make(client))
```
marcinwadon, line 33: I think `L0GlobalSnapshotClient` is already instantiated in `p2pClient`. You can just pass it here instead of creating a new one

marcinwadon: Also, you have a name mismatch: `peerSelector` vs `PeerSelection`. btw - why not `PeerSelect`?

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/peerSelection.scala
marcinwadon, line 1: file name does not correspond to the trait name

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelection.scala
```diff
@@ -0,0 +1,60 @@
...
+    storage: ClusterStorage[F],
+    snapshotClient: L0GlobalSnapshotClient[F]
+  ): PeerSelection[F] = new PeerSelection[F] {
+
+    def select: F[Peer] = storage.getResponsivePeers
+      .map(_.filter(_.state === Ready))
+      .flatMap(getPeerSublist)
+      .flatMap(peers => ApplicativeError[F, Throwable].fromOption(peers.toNel, NoPeers))
+      .flatMap(filterPeerList)
+      .flatMap(filteredPeers => Random[F].elementOf(filteredPeers.toList))
```
marcinwadon, line 32: it can be simplified to `map(_.toList).flatMap(Random[F].elementOf)`

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/programs/Download.scala
```diff
@@ -5,26 +5,30 @@ import cats.syntax.flatMap._
...
+      consensus,
+      peerSelecter
     ) {}
 }
 
 sealed abstract class Download[F[_]: Async] private (
   nodeStorage: NodeStorage[F],
-  consensus: SnapshotConsensus[F, _, _, _, _, _]
+  consensus: SnapshotConsensus[F, _, _, _, _, _],
+  peerSelecter: PeerSelection[F]
```
marcinwadon, line 28: typo

cngo-github: Fixed.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelection.scala
cngo-github: Changed.

modules/core/src/main/scala/org/tessellation/modules/Programs.scala
cngo-github: Changed

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/peerSelection.scala
cngo-github: File has been renamed `PeerSelect.scala` to match the trait.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelectionStrategy.scala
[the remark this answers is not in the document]
cngo-github: Changed to using `MonadThrow`.

cngo-github:
> > image
> > `maxBy` is unsafe operation (cannot be executed on empty map/list). Please use safer version from cats that fallbacks to Option (None) for empty case.
> 
> Regarding this, I think we should also abort if we can't find any peer responses at any stage here and throw an error indicative that no peers were found to make it even clearer (in addition to option here)

Added logging to notify the user when no ready peers are found to be selected.

modules/core/src/main/scala/org/tessellation/modules/Programs.scala
marcinwadon: Ok, I was not aware of that. If so, please add `L0GlobalSnapshotClient` to the `P2PClients` itself and pass it here. That way we have a single place where we pass `client: Client[F]` and instantiate all p2p clients.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
```diff
@@ -0,0 +1,72 @@
+package org.tessellation.sdk.infrastructure.snapshot
+
+import cats.data.NonEmptyList
+import cats.effect.kernel.Async
```
marcinwadon, line 4: `cats.effect.Async`

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
```diff
@@ -0,0 +1,72 @@
...
+import cats.effect.std.Random
+import cats.syntax.all._
```
marcinwadon, line 6: if possible, please import exact syntax instead of all (as it impacts compilation time)

marcinwadon (approved):
LGTM but I'd instantiate the global snapshot client in P2PClient.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
cngo-github: Changed.

modules/core/src/main/scala/org/tessellation/modules/Programs.scala
cngo-github: Added.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
cngo-github: Changed.

sdavidp (approved):
Approving since Marcin's comments were addressed

modules/core/src/main/scala/org/tessellation/http/p2p/P2PClient.scala
```diff
@@ -13,13 +13,15 @@ object P2PClient {
...
     ) {}
 }
 
 sealed abstract class P2PClient[F[_]] private (
   val sign: SignClient[F],
   val cluster: ClusterClient[F],
   val gossip: GossipClient[F],
-  val node: NodeClient[F]
+  val node: NodeClient[F],
+  val l0GlobalSnapshotClient: L0GlobalSnapshotClient[F]
```
marcinwadon, line 26: Please name it accordingly to other variables in this place (no Client suffix)

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/SdkP2PClient.scala
```diff
@@ -28,5 +29,6 @@ sealed abstract class SdkP2PClient[F[_]] private (
   val cluster: ClusterClient[F],
   val gossip: GossipClient[F],
   val node: NodeClient[F],
-  val trust: TrustClient[F]
+  val trust: TrustClient[F],
+  val l0GlobalSnapshotClient: L0GlobalSnapshotClient[F]
```
marcinwadon, line 33: And here as well please

PR #633, opened 2023-05-01
Implement network api endpoints

sdavidp:


modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
```diff
@@ -0,0 +1,62 @@
...
+      EitherT.pure(NetworkApiOptions(version, Allow.default))
+
+    def status: EitherT[F, NetworkApiError, NetworkStatusResponse] =
+      EitherT
+        .fromOptionF[F, NetworkApiError, Hashed[GlobalSnapshot]](getLastSnapshot(), LatestSnapshotNotFound)
+        .map { lastSnapshot =>
+          val genesisSnapshot = GlobalSnapshot.mkGenesis(Map.empty, EpochProgress.MinValue) // TODO confirm this is correct
+
+          val genesisBlockId = BlockIdentifier(
+            genesisSnapshot.height, // TODO confirm: rosetta branch used .ordinal
```
ryle-goehausen, line 40: This should be ordinal

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
```diff
@@ -0,0 +1,62 @@
...
+      EitherT.pure(List(NetworkIdentifier.Mainnet, NetworkIdentifier.Testnet))
+
+    def options: EitherT[F, NetworkApiError, NetworkApiOptions] =
+      EitherT.pure(NetworkApiOptions(version, Allow.default))
+
+    def status: EitherT[F, NetworkApiError, NetworkStatusResponse] =
+      EitherT
+        .fromOptionF[F, NetworkApiError, Hashed[GlobalSnapshot]](getLastSnapshot(), LatestSnapshotNotFound)
+        .map { lastSnapshot =>
+          val genesisSnapshot = GlobalSnapshot.mkGenesis(Map.empty, EpochProgress.MinValue) // TODO confirm this is correct
```
ryle-goehausen, line 37: This needs to be injected as a dependency / parameter for initializing. We don't have a great way to get this right now other than the block explorer API -- so assume it's not supplied yet and must be supplied as an external parameter.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
```diff
@@ -0,0 +1,62 @@
...
+
+          val genesisBlockId = BlockIdentifier(
+            genesisSnapshot.height, // TODO confirm: rosetta branch used .ordinal
+            Hex(genesisSnapshot.lastSnapshotHash.value) // TODO confirm this is correct
+          )
+
+          NetworkStatusResponse(
+            currentBlockIdentifier = BlockIdentifier(
+              lastSnapshot.height, // TODO confirm: rosetta branch used .ordinal
```
ryle-goehausen, line 46: Should be ordinal

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
```diff
@@ -0,0 +1,62 @@
...
+              Hex(lastSnapshot.hash.value)
+            ),
+            currentBlockTimestamp = lastSnapshot.ordinal.value.value, // TODO is this correct?
```
ryle-goehausen, line 49: This needs to be injected as a dependency / parameter instead, i.e. getTimestampForSnapshot(ordinal) -- we're going to have to query block explorer API (for timestamps,) until schema modifications are done to add timestamps. This should be supplied as a function -- for tests it's fine to use ordinal as a placeholder but that should be done thru a trait

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
```diff
@@ -0,0 +1,62 @@
...
+            genesisBlockIdentifier = genesisBlockId,
+            oldestBlockIdentifier = genesisBlockId,
+            syncStatus = SyncStatus(
```
ryle-goehausen, line 52: This is fine, we don't need rosetta to cover synchronization progress yet -- we can potentially just ignore requests until it's in ready state.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/model/options.scala
```diff
@@ -0,0 +1,86 @@
...
+  )
+
+  object Allow {
+    val default: Allow = Allow(
+      operationStatuses = AllowedOperationStatus.allStatuses,
+      operationTypes = List(Transfer.toString),
+      errors = RosettaError.values.map(RosettaErrorJson(_)).toList,
+      historicalBalanceLookup = false,
+      timestampStartIndex = none, // TODO will we support this in future?
+      callMethods = Nil, // TODO is this correct response?
```
ryle-goehausen, line 76: Yes, we have no /call endpoints so far.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/model/options.scala
```diff
@@ -0,0 +1,86 @@
...
+    transactionHashCase: Option[Case]
+  )
+
+  object Allow {
+    val default: Allow = Allow(
+      operationStatuses = AllowedOperationStatus.allStatuses,
+      operationTypes = List(Transfer.toString),
+      errors = RosettaError.values.map(RosettaErrorJson(_)).toList,
+      historicalBalanceLookup = false,
+      timestampStartIndex = none, // TODO will we support this in future?
```
ryle-goehausen, line 75: We will need to add this in the future, for now it remains none, maybe make a note that this should be a hardcoded constant in the future.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/model/options.scala
```diff
@@ -0,0 +1,86 @@
...
+      callMethods = Nil, // TODO is this correct response?
+      balanceExemptions = Nil, // TODO is this correct response?
```
ryle-goehausen, line 77: Yes, we're implementing rewards as a set of Operations rather than exemptions.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/NetworkIdentifier.scala
```diff
@@ -21,6 +22,10 @@ case class NetworkIdentifier(
   network: NetworkEnvironment,
   subNetworkIdentifier: Option[SubNetworkIdentifier]
 )
+object NetworkIdentifier {
+  val Mainnet = NetworkIdentifier(BlockchainId.dag, NetworkEnvironment.Mainnet, none)
+  val Testnet = NetworkIdentifier(BlockchainId.dag, NetworkEnvironment.Testnet, none)
```
marcinwadon, line 27: we already have `NetworkEnvironment.fromAppEnvironment` so I think it's redundant to hardcode specific environments here. you may add `NetworkIdentifier.fromAppEnvironment` that would return you the correct identifier

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
```diff
@@ -0,0 +1,62 @@
...
+  private val rosettaVersion = "1.4.14" // TODO get from a config?
+  private val version = Version(rosettaVersion = rosettaVersion, nodeVersion = BuildInfo.version)
+
+  def make[F[_]: Async](
+    getLastSnapshot: () => F[Option[Hashed[GlobalSnapshot]]],
+    getGenesisSnapshot: () => F[GlobalSnapshot],
+    getTimestampForSnapshot: SnapshotOrdinal => F[Long]
+  ): NetworkApiService[F] = new NetworkApiService[F] {
+    def list: EitherT[F, NetworkApiError, List[NetworkIdentifier]] =
+      EitherT.pure(List(NetworkIdentifier.Mainnet, NetworkIdentifier.Testnet))
```
marcinwadon, line 31: IMO you should return only the active network (so the network you query against).
Rosetta client cannot request mainnet network with testnet network identifier to send testnet transaction (there is no connection between these networks in our protocol).

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
```diff
@@ -0,0 +1,62 @@
...
+
+    def options: EitherT[F, NetworkApiError, NetworkApiOptions] =
+      EitherT.pure(NetworkApiOptions(version, Allow.default))
+
+    def status: EitherT[F, NetworkApiError, NetworkStatusResponse] = for {
+      lastSnapshot <- EitherT.fromOptionF[F, NetworkApiError, Hashed[GlobalSnapshot]](
```
marcinwadon, line 37: in general, when you need to lift each and every operation to `EitherT` it may be beneficial to chain operations and lift to EitherT once at the very end

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
```diff
@@ -0,0 +1,62 @@
...
+
+      genesisBlockId = BlockIdentifier(genesisSnapshot.ordinal, Hex(genesisSnapshot.lastSnapshotHash.value))
+      currentBlockId = BlockIdentifier(lastSnapshot.ordinal, Hex(lastSnapshot.hash.value))
+    } yield
+      NetworkStatusResponse(
+        currentBlockIdentifier = currentBlockId,
+        currentBlockTimestamp = blockTimestamp,
+        genesisBlockIdentifier = genesisBlockId,
+        oldestBlockIdentifier = genesisBlockId,
+        syncStatus = SyncStatus(
```
marcinwadon, line 52: does it depend on the download process?

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
```diff
@@ -0,0 +1,62 @@
...
+}
+
+object NetworkApiService {
+  private val rosettaVersion = "1.4.14" // TODO get from a config?
+  private val version = Version(rosettaVersion = rosettaVersion, nodeVersion = BuildInfo.version)
+
+  def make[F[_]: Async](
+    getLastSnapshot: () => F[Option[Hashed[GlobalSnapshot]]],
+    getGenesisSnapshot: () => F[GlobalSnapshot],
+    getTimestampForSnapshot: SnapshotOrdinal => F[Long]
```
marcinwadon, line 28: do you need timestamp for each snapshot ordinal or just for the latest?

modules/shared/src/test/scala/org/tessellation/schema/generators.scala
```diff
@@ -127,4 +129,35 @@ object generators {
...
+  val snapshotOrdinalGen: Gen[SnapshotOrdinal] =
+    Arbitrary.arbitrary[NonNegLong].map(SnapshotOrdinal(_))
+
+  val heightGen: Gen[Height] =
+    Arbitrary.arbitrary[NonNegLong].map(Height(_))
+
+  val subHeightGen: Gen[SubHeight] =
+    Arbitrary.arbitrary[NonNegLong].map(SubHeight(_))
+
+  val globalSnapshotGen =
```
marcinwadon, line 141: this generator won't pass snapshot validation. I'd rather create a proper snapshot instead of using arbitrary gen

marcinwadon (changes requested):

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/NetworkIdentifier.scala
sdavidp: `NetworkIdentifier.fromAppEnvironment` returns an `Option` which means I need to add runtime logic to handle a case that will never happen because we know what we want to respond with at compile time. So a constant makes more sense to me.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
sdavidp: @marcinwadon
I don't understand. Only 2 of the 3 uses are lifted and that's so the first one can fail fast. As I understand it that's the point of using a monad transformer.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/NetworkIdentifier.scala
marcinwadon: yes, it returns `Option` because you can launch an environment that does not map into network identifier.
having it constant, you would just return invalid data

marcinwadon: so it's not a case that will never happen. you compile a single version and you specify the environment when running the node, not at compile time.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
```diff
@@ -0,0 +1,62 @@
...
+    def status: EitherT[F, NetworkApiError, NetworkStatusResponse] = for {
+      lastSnapshot <- EitherT.fromOptionF[F, NetworkApiError, Hashed[GlobalSnapshot]](
+        getLastSnapshot(),
+        LatestSnapshotNotFound
+      )
+      genesisSnapshot <- EitherT.liftF(getGenesisSnapshot())
+      blockTimestamp <- EitherT.liftF(getTimestampForSnapshot(lastSnapshot.ordinal))
+
+      genesisBlockId = BlockIdentifier(genesisSnapshot.ordinal, Hex(genesisSnapshot.lastSnapshotHash.value))
+      currentBlockId = BlockIdentifier(lastSnapshot.ordinal, Hex(lastSnapshot.hash.value))
```
marcinwadon, line 45: why `currentBlockId` uses `lastSnapshot.hash` but `genesisBlockId` uses `genesisSnapshot.lastSnapshotHash`?

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/NetworkIdentifier.scala
sdavidp: If I write `NetworkIdentifier.fromAppEnvironment`

1. What is the return type? If it returns an `Option` then my what about my comment above?
2. We have 3 `AppEnvironment`s defined but only 2 are mapped to a `NetworkEnvironment` -- `Dev` is excluded. Am I to do the same thing for `NetworkIdentifier.fromAppEnvironment` ?
3. If we don't map `Dev`, what should the endpoint return for that environment? Empty list?

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
marcinwadon: in this case you have one operation that is lifted to EitherT because it returns Option (and you want to fail fast).
but then, you have some other operations that you need to lift into Either (each one). if you would have more of them, it may be simpler to just chain them and use `EitherT.liftF { ...chainedOperations... }` once. something like that (or equivalent using for comp):
image

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/NetworkIdentifier.scala
marcinwadon: re.1: Which network identifier you want to return on development environment?
re.2: up to you (you can map every app environment if you want, depending where rosetta should work)
re.3: I have no idea but returning either mainnet or testnet, or both, seems to be invalid, so imo there is no point in hardcoding network identifiers

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
sdavidp: Seems to me "simpler" is subjective. If you insist I'll change it to that but I don't see value in it.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
sdavidp: `genesisSnapshot` is not a `Hashed[GlobalSnapshot]` so `.hash` is not defined.
Is `.lastSnapshotHash` not the valid hash to use?

sdavidp: Since we're now injecting the function that gives us genesis I could change the return type so that it's the caller's responsibility to give me a `Hashed[GlobalSnapshot]` then the block ids would be set similarly.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
marcinwadon: No problem. I've just written that if you'd have a more complex case with more operations. It's fine for now

modules/shared/src/test/scala/org/tessellation/schema/generators.scala
sdavidp: Is this function (taken from SnapshotProcessorSuite) creating a valid snapshot?
```
  def generateSnapshot(peerId: PeerId): GlobalSnapshot =
    GlobalSnapshot(
      snapshotOrdinal10,
      snapshotHeight6,
      snapshotSubHeight0,
      lastSnapshotHash,
      SortedSet.empty,
      SortedMap.empty,
      SortedSet.empty,
      EpochProgress.MinValue,
      NonEmptyList.one(peerId),
      GlobalSnapshotInfo(SortedMap.empty, SortedMap.empty, SortedMap.empty),
      SnapshotTips(SortedSet.empty, SortedSet.empty)
    )
```

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
sdavidp: The docs seem to specify the timestamp only for the current block.
https://www.rosetta-api.org/docs/models/NetworkStatusResponse.html

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
sdavidp: ok, thank you Marcin. I will continue to look for simpler implementations and only lift things into the EitherT if it feels like it makes the code easier to understand.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/NetworkIdentifier.scala
sdavidp: Refactored. The code now (once I push it) defines the network id in terms of the network env and will return an empty list for the Dev environment since that's not defined.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
sdavidp: 🤔 I'm not sure. Do you think this should be `NodeState` ?
If so, for which states would _synced_ be `true`?

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
marcinwadon: got it. so I think that you should specify a function for the current block only. the way it is right now suggests that we should calculate timestamp for any arbitrary snapshot

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
marcinwadon: I'm asking because I don't know what rosetta specifies under this property. anyway, imo it shouldn't be the node state (as it's custom to our network)

marcinwadon: but as I think about it, it says `stage = Synced` which is invalid if node has download in progress 🤔

modules/shared/src/test/scala/org/tessellation/schema/generators.scala
marcinwadon: yes, I think so. `SnapshotProcessorSuite` uses the full validation

marcinwadon:
@sdavidp can you change the base branch to `feature/incremental-snapshots`?

modules/shared/src/test/scala/org/tessellation/schema/generators.scala
sdavidp: I thought I based my generator on that function. I'll check again this morning.

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
sdavidp: @marcinwadon  `s/getTimestampForSnapshot/getLastSnapshotTimestamp`  ?

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
marcinwadon: `hash` and `lastSnapshotHash` are two different things. while `hash` references to the current data, `lastSnapshotHash` references to previous data. IMO `genesisBlockId` should present genesis hash. genesis `lastSnapshotHash` is an empty hash (no reference).

modules/shared/src/test/scala/org/tessellation/schema/generators.scala
marcinwadon: it has predefined/hardcoded variables I think. the only concern I have is about the `lastSnapshotHash` and I think that `Arbitrary[Hash]` produces wrong hashes

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
sdavidp: I haven't found anything yet that tells us what is expected. Since it's not mentioned in the schema I'm guessing it's our choice. Given that there's a `synced` flag I'm guessing `stage` is informative and that the flag is really telling whether the node is "ready".

Maybe we should have at least two values for staged, e.g. `Ready` when `synced = true` and `Not Ready`  otherwise.
Then we just need to decide which state (`NodeState` ?) will set the flag to true.

WDYT?
@ryle-goehausen any thoughts?

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
marcinwadon: something like that (and then without argument?) 🤔

marcinwadon: also, you can create a newtype for timestamp

sdavidp: What's the benefit of creating a newtype? are we going to constraint the values in some way?

marcinwadon: newtype is not for exactly for putting constraints. it's more about eliminating human error - if you see the type `Long` it can be anything (and you can use it anywhere). if you see the type `SnapshotOrdinal` (even if it's `Long` underlaying), you cannot misplace or misunderstand the type. hence, making it a `Timestamp` will differentiate it from (for example) snapshot ordinal, which is also of type `Long` but already wrapped into newtype.
also, newtypes are zero-cost runtime wrappers, so it's fine to use them in many places.

sdavidp: @marcinwadon Back to the `getLastSnapshotTimestamp` ...

If we're going to call it that and remove the argument then it doesn't make sense to me that it be a separate function from obtaining last snapshot itself since there is a dependency between the two.

So wouldn't merging the two make sense? E.g. something like
```
getLastSnapshotInfo: () => F[Option[(Hashed[GlobalSnapshot], SnapshotTimestamp)]]
```

marcinwadon: yup, it makes perfect sense. although, `GlobalSnapshot` should be `GlobalIncrementalSnapshot` (but you need to base the change to the `feature/incremental-snapshots` branch)

sdavidp: I noticed that. Thanks!

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
ryle-goehausen: We definitely don't need any more granularity other than Ready / not ready here, so synced = true when node state is Ready is fine.

This can also be optional and not required to implement as it's only for advanced uses.

modules/shared/src/test/scala/org/tessellation/schema/generators.scala
sdavidp: globalSnapshotGen removed

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
```diff
@@ -0,0 +1,68 @@
...
+import org.tessellation.security.hex.Hex
+
+trait NetworkApiService[F[_]] {
+  def list(appEnv: AppEnvironment): EitherT[F, NetworkApiError, List[NetworkIdentifier]]
+  def options: EitherT[F, NetworkApiError, NetworkApiOptions]
+  def status: EitherT[F, NetworkApiError, NetworkStatusResponse]
+}
+
+object NetworkApiService {
+  private val rosettaVersion = "1.4.14" // TODO get from a config?
```
marcinwadon, line 27: TODO left

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
```diff
@@ -0,0 +1,68 @@
...
+  def make[F[_]: Async](
+    lastSnapshot: F[Option[LastSnapshotInfo]],
+    genesisOrdinalAndHash: F[(SnapshotOrdinal, Hash)],
+    nodeState: F[NodeState]
+  ): NetworkApiService[F] = new NetworkApiService[F] {
+    def list(appEnv: AppEnvironment): EitherT[F, NetworkApiError, List[NetworkIdentifier]] =
+      EitherT.pure(NetworkIdentifier.fromAppEnvironment(appEnv).toList)
+
+    def options: EitherT[F, NetworkApiError, NetworkApiOptions] =
+      EitherT.pure(NetworkApiOptions(version, Allow.default))
```
marcinwadon, line 41: if these are pure fns, why wrapping into `EitherT`?
it can be simply:
```scala
def list(appEnv: AppEnvironment): List[NetworkIdentifier] = ...
def options: NetworkApiOptions = ...
```

modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiServiceSuite.scala
```diff
@@ -0,0 +1,149 @@
...
+import weaver.MutableIOSuite
+import weaver.scalacheck.Checkers
+
+object NetworkApiServiceSuite extends MutableIOSuite with Checkers {
+  type Res = (SecurityProvider[IO], KryoSerializer[IO])
+
+  override def sharedResource: Resource[IO, Res] =
+    KryoSerializer
+      .forAsync[IO](sharedKryoRegistrar)
+      .flatMap(kryo => SecurityProvider.forAsync[IO].map(securityProvider => (securityProvider, kryo)))
```
marcinwadon, line 39: it can be `map((_, kryo))`

modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiServiceSuite.scala
```diff
@@ -0,0 +1,149 @@
...
+  ): NetworkApiService[IO] =
+    NetworkApiService.make[IO](lastSnapshotInfo, genesisSnapshot, nodeState)
+
+  test("list returns mainnet identifier for that AppEnvironment ") { _ =>
+    val netId = NetworkIdentifier(BlockchainId.dag, NetworkEnvironment.Mainnet, none)
+    val expected = List(netId).asRight[NetworkApiError]
+    makeNetworkApiService()
+      .list(AppEnvironment.Mainnet)
+      .value
+      .map(actual => expect.eql(expected, actual))
```
marcinwadon, line 55: `map(expect.eql(expected, _))`

marcinwadon (changes requested):
There is one TODO left and also I left one comment about the EitherT which I think can be simplified. Other than that, LGTM

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiService.scala
sdavidp: removed

modules/rosetta/src/test/scala/org/tessellation/rosetta/domain/networkapi/NetworkApiServiceSuite.scala
sdavidp: code changed due to change in return type of `.list`

marcinwadon (approved):
I left you one comment on slack about `pureTest` and `test`, otherwise looks good :+1:

marcinwadon (approved):
looks good :+1:

PR #636, opened 2023-05-03
PeerSelect gracefully handles no snapshot errors

cngo-github:
The current `PeerSelect` doesn't gracefully handles errors that are the result of it querying a `Peer` for the snapshot for a `SnapshotOrdinal` value that doesn't exist. This fixes that. It also parallelizes the queries to other `Peer`s and addresses comments that were not addressed prior to merging PR https://github.com/Constellation-Labs/tessellation/pull/630.

### DIFF
--- a/modules/core/src/main/scala/org/tessellation/http/p2p/P2PClient.scala
+++ b/modules/core/src/main/scala/org/tessellation/http/p2p/P2PClient.scala
@@ -14,7 +14,7 @@ object P2PClient {
       sdkP2PClient.cluster,
       sdkP2PClient.gossip,
       sdkP2PClient.node,
-      sdkP2PClient.l0GlobalSnapshotClient
+      sdkP2PClient.l0GlobalSnapshot
     ) {}
 }
 
@@ -23,5 +23,5 @@ sealed abstract class P2PClient[F[_]] private (
   val cluster: ClusterClient[F],
   val gossip: GossipClient[F],
   val node: NodeClient[F],
-  val l0GlobalSnapshotClient: L0GlobalSnapshotClient[F]
+  val l0GlobalSnapshot: L0GlobalSnapshotClient[F]
 )
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/SdkP2PClient.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/SdkP2PClient.scala
@@ -30,5 +30,5 @@ sealed abstract class SdkP2PClient[F[_]] private (
   val gossip: GossipClient[F],
   val node: NodeClient[F],
   val trust: TrustClient[F],
-  val l0GlobalSnapshotClient: L0GlobalSnapshotClient[F]
+  val l0GlobalSnapshot: L0GlobalSnapshotClient[F]
 )
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
@@ -1,8 +1,10 @@
 package org.tessellation.sdk.infrastructure.snapshot
 
+import cats.MonadThrow
 import cats.data.NonEmptyList
 import cats.effect.Async
 import cats.effect.std.Random
+import cats.effect.syntax.concurrent._
 import cats.syntax.applicative._
 import cats.syntax.applicativeError._
 import cats.syntax.eq._
@@ -13,16 +15,22 @@ import cats.syntax.list._
 import scala.util.control.NoStackTrace
 
 import org.tessellation.kryo.KryoSerializer
+import org.tessellation.schema.SnapshotOrdinal
 import org.tessellation.schema.node.NodeState.Ready
 import org.tessellation.schema.peer.Peer
 import org.tessellation.schema.peer.Peer.toP2PContext
 import org.tessellation.sdk.domain.cluster.storage.ClusterStorage
 import org.tessellation.sdk.domain.snapshot.PeerSelect
 import org.tessellation.sdk.http.p2p.clients.L0GlobalSnapshotClient
+import org.tessellation.security.hash.Hash
 
 import org.typelevel.log4cats.slf4j.Slf4jLogger
 
 object MajorityPeerSelect {
+
+  val maxConcurrentPeerInquiries = 10
+  val peerSampleRatio = 0.25
+  val minSampleSize = 20
 
   case object NoPeersToSelect extends NoStackTrace
 
@@ -33,7 +41,9 @@ object MajorityPeerSelect {
 
     val logger = Slf4jLogger.getLoggerFromClass[F](MajorityPeerSelect.getClass)
 
-    def select: F[Peer] = storage.getResponsivePeers
+    def select: F[Peer] = selectPeer
+
+    def selectPeer: F[Peer] = storage.getResponsivePeers
       .map(_.filter(_.state === Ready))
       .flatMap(getPeerSublist)
       .flatMap { peers =>
@@ -50,28 +60,32 @@ object MajorityPeerSelect {
 
     def filterPeerList(peers: NonEmptyList[Peer]): F[NonEmptyList[Peer]] =
       peers
-        .traverse(snapshotClient.getLatestOrdinal(_))
-        .map {
-          _.groupBy(identity).maxBy { case (_, ordinals) => ordinals.size }
+        .parTraverseN(maxConcurrentPeerInquiries)(snapshotClient.getLatestOrdinal(_))
+        .map(_.groupBy(identity).maxBy { case (_, ordinals) => ordinals.size })
+        .map { case (majorityOrdinal, _) => majorityOrdinal }
+        .flatMap { majorityOrdinal =>
+          peers
+            .parTraverseN(maxConcurrentPeerInquiries)(peer => getSnapshotHashByPeer(peer, majorityOrdinal).attempt)
         }
-        .flatMap {
-          case (majorityOrdinal, _) =>
-            peers.traverse(snapshotClient.get(majorityOrdinal).run(_).flatMap(_.toHashed.map(_.hash)))
+        .flatMap { peerSnapshotHashes =>
+          MonadThrow[F].fromOption(
+            peerSnapshotHashes.collect { case Right(peerSnapshot) => peerSnapshot }.toNel,
+            NoPeersToSelect
+          )
         }
-        .map(_.zip(peers))
-        .map(_.groupMap { case (hash, _) => hash } { case (_, ps) => ps })
-        .map(_.maxBy { case (_, peers) => peers.size })
-        .map { case (_, peerCandidates) => peerCandidates }
+        .map(_.groupMap { case (_, hash) => hash } { case (peer, _) => peer })
+        .map(_.values.toList.sortBy(-_.size))
+        .map(_.head)
 
     def getPeerSublist(peers: Set[Peer]): F[List[Peer]] = {
-      val maxSublistPercent = 0.25
-      val maxSublistSize = Math.max((peers.size * maxSublistPercent).toInt, 1)
+      val sampleSize = (peers.size * peerSampleRatio).toInt
 
-      Random[F].nextIntBounded(maxSublistSize).flatMap { peerCount =>
-        Random[F]
-          .shuffleList(peers.toList)
-          .map(_.take(peerCount))
-      }
+      Random[F]
+        .shuffleList(peers.toList)
+        .map(_.take(Math.max(sampleSize, minSampleSize)))
     }
+
+    def getSnapshotHashByPeer(peer: Peer, ordinal: SnapshotOrdinal): F[(Peer, Hash)] =
+      snapshotClient.get(ordinal).run(peer).flatMap(_.toHashed).map(snapshot => (peer, snapshot.hash))
   }
 }
--- a/modules/core/src/main/scala/org/tessellation/Main.scala
+++ b/modules/core/src/main/scala/org/tessellation/Main.scala
@@ -64,7 +64,7 @@ object Main
           cfg
         )
         .asResource
-      programs = Programs.make[IO](sdkPrograms, storages, services, keyPair, cfg, p2pClient.l0GlobalSnapshotClient)
+      programs = Programs.make[IO](sdkPrograms, storages, services, keyPair, cfg, p2pClient.l0GlobalSnapshot)
       healthChecks <- HealthChecks
         .make[IO](
           storages,
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/programs/Download.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/programs/Download.scala
@@ -13,12 +13,12 @@ object Download {
   def make[F[_]: Async](
     nodeStorage: NodeStorage[F],
     consensus: SnapshotConsensus[F, _, _, _, _, _],
-    peerSelecter: PeerSelect[F]
+    peerSelect: PeerSelect[F]
   ): Download[F] =
     new Download[F](
       nodeStorage,
       consensus,
-      peerSelecter
+      peerSelect
     ) {}
 }
 


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
```diff
@@ -50,22 +56,28 @@ object MajorityPeerSelect {
...
     def filterPeerList(peers: NonEmptyList[Peer]): F[NonEmptyList[Peer]] =
       peers
-        .traverse(snapshotClient.getLatestOrdinal(_))
-        .map {
-          _.groupBy(identity).maxBy { case (_, ordinals) => ordinals.size }
+        .parTraverseN(maxConcurrentPeerInquiries)(snapshotClient.getLatestOrdinal(_))
+        .map { latestOrdinals =>
+          val (majorityOrdinal, _) = latestOrdinals.groupBy(identity).maxBy { case (_, ordinals) => ordinals.size }
+          majorityOrdinal
+        }
```
marcinwadon, line 63:
```scala
.map( _.groupBy(identity).maxBy { case (_, ordinals) => ordinals.size })
.map { case (majorityOrdinal, _) => majorityOrdinal }
```

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
```diff
@@ -50,22 +56,28 @@ object MajorityPeerSelect {
...
+        .flatMap { majorityOrdinal =>
+          peers.parTraverseN(maxConcurrentPeerInquiries)(snapshotClient.get(majorityOrdinal).run(_).attempt)
```
marcinwadon, line 65: if peer returned latest ordinal higher than the majority ordinal but does not have the majority ordinal, it's an unexpected error and should disqualify that peer from selecting.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
```diff
@@ -50,22 +56,28 @@ object MajorityPeerSelect {
...
+            maybeSnapshots.collect { case Right(snapshot) => snapshot }.toNel,
+            NoPeersToSelect
+          )
         }
-        .map(_.zip(peers))
-        .map(_.groupMap { case (hash, _) => hash } { case (_, ps) => ps })
-        .map(_.maxBy { case (_, peers) => peers.size })
-        .map { case (_, peerCandidates) => peerCandidates }
+        .flatMap(_.traverse(_.toHashed.map(_.hash)))
+        .map(_.zip(peers).groupMap { case (hash, _) => hash } { case (_, ps) => ps })
```
marcinwadon, line 74: didn't check it but from top of my head - is it right to zip it like that if you before filter out some snapshots?
IMO you should traverse by peer list, fetch and map it to tuple `(peer, snapshot)` or something that will index what you fetched. we use that in many places in the codebase, so we don't need to zip it with arbitrary collection and we're make sure we zip correct data together (it does not depend on the order of collection items)

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
```diff
@@ -50,22 +56,28 @@ object MajorityPeerSelect {
...
+        .map(_.values.toList.sortWith(_.size > _.size))
```
marcinwadon, line 75: `sortBy(_.size)`

marcinwadon (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
```diff
@@ -50,22 +56,28 @@ object MajorityPeerSelect {
...
+        .map(_.head)
 
     def getPeerSublist(peers: Set[Peer]): F[List[Peer]] = {
       val maxSublistPercent = 0.25
-      val maxSublistSize = Math.max((peers.size * maxSublistPercent).toInt, 1)
+      val maxSublistSize = Math.max((peers.size * maxSublistPercent).toInt, 20)
```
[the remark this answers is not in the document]
marcinwadon: Got it. Can you name it minSampleSize then?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
```diff
@@ -50,22 +56,28 @@ object MajorityPeerSelect {
...
 
       Random[F].nextIntBounded(maxSublistSize).flatMap { peerCount =>
```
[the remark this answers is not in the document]
cngo-github: Changed.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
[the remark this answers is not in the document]
cngo-github: Changed.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
cngo-github: Changed.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
cngo-github: If a `Peer` returns a `NotFound` for the snapshot given an ordinal, then the query for that `Peer` returns an error in the `F` context. The `attempt` makes that a `Left`, so the following `flatMap` ignores it.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
cngo-github: Changed.

sdavidp (approved):
LGTM. Will wait for @marcinwadon to review and we can address any concerns he has.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
marcinwadon: @sdavidp so it should be `sortBy(-_.size)` then

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
```diff
@@ -50,28 +59,34 @@ object MajorityPeerSelect {
...
-        .map(_.maxBy { case (_, peers) => peers.size })
-        .map { case (_, peerCandidates) => peerCandidates }
+        .map(_.groupMap { case (_, hash) => hash } { case (peer, _) => peer })
+        .map(_.values.toList.sortWith(_.size > _.size))
+        .map(_.head)
 
     def getPeerSublist(peers: Set[Peer]): F[List[Peer]] = {
-      val maxSublistPercent = 0.25
-      val maxSublistSize = Math.max((peers.size * maxSublistPercent).toInt, 1)
+      val minSampleSize = Math.max((peers.size * maxPeerSampleRatio).toInt, 20)
```
marcinwadon, line 80: is it min sample size or max sublist size? I thought that `20` is the min sample size.

imo it should look like this:
```scala
val sublistPercent = 0.25
val minSampleSize = 20
val sublistSize = Math.max((peers.size * maxSublistPercent).toInt, minSampleSize)

Random[F].nextIntBounded(sublistSize).map(Math.max(_, 1)).flatMap { peersToTake =>
  Random[F]
    .shuffleList(peers.toList)
    .map(_.take(peersToTake))
}
```

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
```diff
@@ -50,28 +59,34 @@ object MajorityPeerSelect {
...
 
-      Random[F].nextIntBounded(maxSublistSize).flatMap { peerCount =>
+      Random[F].nextIntBounded(1 + minSampleSize).flatMap { peerCount =>
```
marcinwadon, line 82: this does not fix the issue we discussed on slack. `nextIntBounded` may return `0`.
see above comment how I see that change

marcinwadon (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
marcinwadon: also, I'm thinking if `Random[F].nextIntBounded` is indeed correct 🤔 do you want to take an arbitrary peers count from 1 to 25% of all peers?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
sdavidp: @marcinwadon Why is that preferred over `sortWith` ?

marcinwadon: also, I see no `sortWith` in the codebase but I see 8 usages of `sortBy`

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
sdavidp: @marcinwadon You're correct, that's my bad. Sorry @cngo-github  :(
The +1 belongs in the take:
```
 Random[F].nextIntBounded(minSampleSize).flatMap { peerCount =>
        Random[F]
          .shuffleList(peers.toList)
          .map(_.take(1 + peerCount))
      }
```
`Random[F].nextIntBounded(minSampleSize)` should give us a value in interval `[0, minSampleSize)`
so adding 1 in the take will give us `[1, minSampleSize]`

sdavidp: .... which I just realized isn't really what we want.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
cngo-github: Changed the implementation to the following:

```
      val maxSampleSize = minPeerSelectionSize + (peers.size * peerSampleRatio).toInt

      Random[F].betweenInt(minPeerSelectionSize, maxSampleSize + 1).flatMap { peerCount =>
        Random[F]
          .shuffleList(peers.toList)
          .map(_.take(peerCount))
      }
```

where

```
  val peerSampleRatio = 0.25
  val minPeerSelectionSize = 20
```

So, at least 20 `Peer`s will be selected. If `peer.size < 20` then `_.take(peerCount)` will select all the `Peer`s. The upper bound for the number of `Peer`s selected is `maxSampleSize` or 20 `Peer`s more than 25% of the total number of discovered `Peer`s. At low number of discovered `Peer`s, this means that most or all of the discovered `Peer`s will be selected. As the number of discovered `Peer`s grows, the `20` becomes increasingly dwarfed by `peers.size * peerSampleRatio` and increasingly trivial. Since the point of selecting a subset of `Peer`s is avoid to many queries as the number of discovered `Peer`s grows, this is fine.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
cngo-github: See my above comment.

modules/core/src/main/scala/org/tessellation/http/p2p/P2PClient.scala
```diff
@@ -23,5 +23,5 @@ sealed abstract class P2PClient[F[_]] private (
   val cluster: ClusterClient[F],
   val gossip: GossipClient[F],
   val node: NodeClient[F],
-  val l0GlobalSnapshotClient: L0GlobalSnapshotClient[F]
+  val l0GlobalSnapshot: L0GlobalSnapshotClient[F]
```
sdavidp, line 26: Out of curiosity, why the rename?

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/SdkP2PClient.scala
```diff
@@ -30,5 +30,5 @@ sealed abstract class SdkP2PClient[F[_]] private (
   val gossip: GossipClient[F],
   val node: NodeClient[F],
   val trust: TrustClient[F],
-  val l0GlobalSnapshotClient: L0GlobalSnapshotClient[F]
+  val l0GlobalSnapshot: L0GlobalSnapshotClient[F]
```
sdavidp, line 33: Same question: why the rename?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
```diff
@@ -50,28 +60,34 @@ object MajorityPeerSelect {
...
-      val maxSublistPercent = 0.25
-      val maxSublistSize = Math.max((peers.size * maxSublistPercent).toInt, 1)
+      val maxSampleSize = minPeerSelectionSize + (peers.size * peerSampleRatio).toInt
 
-      Random[F].nextIntBounded(maxSublistSize).flatMap { peerCount =>
+      Random[F].betweenInt(minPeerSelectionSize, maxSampleSize + 1).flatMap { peerCount =>
         Random[F]
           .shuffleList(peers.toList)
           .map(_.take(peerCount))
       }
```
sdavidp, line 87: Here are a couple alternate implementations which IMO are cleaner:
```
      for {
        shuffled <- Random[F].shuffleList(peers.toList)
        takeAmount <- Random[F].betweenInt(minPeerSelectionSize, maxSampleSize + 1)
      } yield shuffled.take(takeAmount)
```
```
      Applicative[F].map2(
        Random[F].shuffleList(peers.toList),
        Random[F].betweenInt(minPeerSelectionSize, maxSampleSize + 1)
      )((ps, n) => ps.take(n))
```

sdavidp (dismissed):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
```diff
@@ -50,28 +60,34 @@ object MajorityPeerSelect {
...
-        .map(_.maxBy { case (_, peers) => peers.size })
-        .map { case (_, peerCandidates) => peerCandidates }
+        .map(_.groupMap { case (_, hash) => hash } { case (peer, _) => peer })
+        .map(_.values.toList.sortBy(-_.size))
+        .map(_.head)
 
     def getPeerSublist(peers: Set[Peer]): F[List[Peer]] = {
-      val maxSublistPercent = 0.25
-      val maxSublistSize = Math.max((peers.size * maxSublistPercent).toInt, 1)
+      val maxSampleSize = minPeerSelectionSize + (peers.size * peerSampleRatio).toInt
```
marcinwadon, line 81: shouldn't it be either `(peers.size * peerSampleRatio).toInt` or fallback to `minPeerSelectionSize`?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
```diff
@@ -50,28 +60,34 @@ object MajorityPeerSelect {
...
 
-      Random[F].nextIntBounded(maxSublistSize).flatMap { peerCount =>
+      Random[F].betweenInt(minPeerSelectionSize, maxSampleSize + 1).flatMap { peerCount =>
```
marcinwadon, line 83: why `+1` if using `betweenInt`?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
cngo-github: Changed.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
marcinwadon: I'd see it as:
```scala
val sampleSize = (peers.size * peerSampleRatio).toInt

Random[F].shuffleList(peers.toList)
  .map(_.take(Math.max(sampleSize, minSampleSize)))
```

so we either take 25% of the cluster or all the peers (if the sample size is lower than min sample size)

modules/core/src/main/scala/org/tessellation/http/p2p/P2PClient.scala
cngo-github: @marcinwadon requested it to keep it in line with the other clients.

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/SdkP2PClient.scala
cngo-github: @marcinwadon requested it to keep it in line with the other clients.

modules/core/src/main/scala/org/tessellation/http/p2p/P2PClient.scala
marcinwadon: imo, there should be no suffix `Client` (see other names).

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
cngo-github: Because if `(peers.size * peerSampleRatio).toInt` is 0, then `maxSampleSize` == `minPeerSelectionSize`. @sdavidp  and I were unsure what the behavior would be if the upper and lower bounds were the same value. `+1` ensures that the upper bound is always greater than the lower bound.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
cngo-github: Adding `minPeerSelectionSize` to `(peers.size * peerSampleRatio).toInt` ensures that `maxSampleSize` will always be greater than or equal to `minPeerSelectionSize`, which ensures that the upper bound for `Random[F].betweenInt` is greater than or equal to the lower bound.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
marcinwadon: I'd rather think about the algorithm and then implement it, instead of thinking how specific implementation works and craft the algorithm to match the implementation.

What's the requirement of this ticket? You want to select some subset of peers for majority selection, right?
IMO, `minPeerSelectionSize` says that it can be the lower bound. the upper bound is max of 25% peers or minPeerSelectionSize if I understood it correctly. so, if we have less peers than minPeerSelectionSize, we should select all the peers for majority selection. do we want to randomly choose number of peers instead of 25% of the cluster? if so, it can look like:

peers > minPeerSelectionSize => take random number between minPeerSelectionSize and 25% of peers
peers <= minPeerSelectionSize => take all peers

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
cngo-github: I think that implementation shortcoming is fine, @marcinwadon. I'll check with @ryle-goehausen to be sure.

cngo-github: Implemented @marcinwadon suggestion.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
cngo-github: Implemented the algorithm suggested by @marcinwadon. `PeerSelect` will either select the the discovered `Peer`s up to `minSampleSize` (20) or 25% of the discovered `Peer`s, whichever is higher.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
```diff
@@ -50,28 +60,32 @@ object MajorityPeerSelect {
...
-        Random[F]
-          .shuffleList(peers.toList)
-          .map(_.take(peerCount))
-      }
+      Random[F]
+        .shuffleList(peers.toList)
+        .map(_.take(Math.max(sampleSize, minSampleSize)))
     }
+
+    def getPeerHash(peer: Peer, ordinal: SnapshotOrdinal): F[(Peer, Hash)] =
```
marcinwadon, line 88: the function name suggests that it returns a peer hash but in fact it returns a snapshot hash from this peer. you cannot guess that from the type as well. IMO the name is confusing - could you rename it? `getSnapshotHash` or `getSnapshotHashByPeer` or `getSnapshotHashFromPeer`, or anything else

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
```diff
@@ -50,28 +60,32 @@ object MajorityPeerSelect {
...
+        .flatMap { majorityOrdinal =>
+          peers
+            .parTraverseN(maxConcurrentPeerInquiries)(peer => getPeerHash(peer, majorityOrdinal).attempt)
         }
-        .flatMap {
-          case (majorityOrdinal, _) =>
-            peers.traverse(snapshotClient.get(majorityOrdinal).run(_).flatMap(_.toHashed.map(_.hash)))
+        .flatMap { maybePeerSnapshots =>
+          MonadThrow[F].fromOption(
+            maybePeerSnapshots.collect { case Right(peerSnapshot) => peerSnapshot }.toNel,
```
marcinwadon, line 72: I think that `maybePeerSnapshots.sequence.toNel` would do the same (or flatSequence)

marcinwadon: also, `maybe` suggests it's an optional type, but collect uses `either`. so is it an Option or Either?

cngo-github: `maybePeerSnapshots` renamed to `peerSnapshotHashes` and `maybePeerSnapshots.collect { case Right(peerSnapshot) => peerSnapshot }.toNel` changed to `peerSnapshotHashes.sequence.toOption`.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
cngo-github: Renamed to `getSnapshotHashByPeer`.

marcinwadon (approved):
LGTM but make sure it works on cluster

sdavidp (approved):

PR #642, opened 2023-05-09
Ignore the incremental snapshots data directory

cngo-github:
Ignore the incremental snapshots data directory

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ scripts/
 /logs/
 .envrc
 .direnv/
+data/incremental_snapshot/


marcinwadon:
Why is it required? You can just select any directory for data storage. You can also run the node from different working directory. And anyway, this directory isn't the only one that is created (there are two or three).

cngo-github:
> Why is it required? You can just select any directory for data storage. You can also run the node from different working directory. And anyway, this directory isn't the only one that is created (there are two or three).

@marcinwadon , I saw this directory in my `git status` after running the validator nodes in IntelliJ. I'm adding that directory in order to not have that directory be accidentally added to `git`. If this is not the only directory created, what are the other ones? I would like to ignore all of them.

marcinwadon:
works for me too. otherwise, ignore whole `data/` directory

marcinwadon:
also, each node should have different path specified (so they should not be default ones)

sdavidp:
> 

Do we have to do that, or can each node create its own directory under a common base?

marcinwadon:
Assuming you're a node owner, you have two options: 
- either provide the path for storage (for example you want to keep data on different, bigger disk that can be unmounted and mounted into another instance),
- or fallback to the default storage path (created in your working directory, so in a place you run the jar file).

If you run everything from IntelliJ, you are simulating many nodes, hence you should set up storage path for each node, otherwise the data will mix up.

cngo-github:
Working with @sdavidp , the IntelliJ application configurations have been updated to use the `$PROJECT_DIR$/target` directory, which is already ignored by `git` and cleaned by `sbt clean`. This PR is not needed anymore.

PR #643, opened 2023-05-11
Added endpoint to get Snapshot hash

cngo-github:
Added an endpoint to get the hash of a `Snapshot` given a `SnapshotOrdinal`.

### DIFF
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/L0GlobalSnapshotClient.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/L0GlobalSnapshotClient.scala
@@ -5,6 +5,7 @@ import cats.effect.Async
 import org.tessellation.kryo.KryoSerializer
 import org.tessellation.schema._
 import org.tessellation.schema.snapshot.SnapshotMetadata
+import org.tessellation.sdk.domain.cluster.services.Session
 import org.tessellation.sdk.http.p2p.PeerResponse
 import org.tessellation.sdk.http.p2p.PeerResponse.PeerResponse
 import org.tessellation.security.SecurityProvider
@@ -14,6 +15,7 @@ import org.tessellation.security.signature.Signed
 import io.circe.Decoder
 import io.circe.magnolia.derivation.decoder.semiauto._
 import io.circe.refined._
+import org.http4s.Method.POST
 import org.http4s.Uri
 import org.http4s.client.Client
 
@@ -23,6 +25,7 @@ trait L0GlobalSnapshotClient[F[_]] {
   def getLatest: PeerResponse[F, (Signed[GlobalIncrementalSnapshot], GlobalSnapshotInfo)]
   def get(ordinal: SnapshotOrdinal): PeerResponse[F, Signed[GlobalIncrementalSnapshot]]
   def get(hash: Hash): PeerResponse[F, Signed[GlobalIncrementalSnapshot]]
+  def getHash(ordinal: SnapshotOrdinal): PeerResponse[F, Option[Hash]]
   def getFull(ordinal: SnapshotOrdinal): PeerResponse[F, Signed[GlobalSnapshot]]
 }
 
@@ -30,7 +33,7 @@ object L0GlobalSnapshotClient {
 
   def make[
     F[_]: Async: SecurityProvider: KryoSerializer
-  ](client: Client[F]): L0GlobalSnapshotClient[F] =
+  ](client: Client[F], session: Session[F]): L0GlobalSnapshotClient[F] =
     new L0GlobalSnapshotClient[F] {
 
       def getLatestOrdinal: PeerResponse[F, SnapshotOrdinal] = {
@@ -57,6 +60,14 @@ object L0GlobalSnapshotClient {
         import org.tessellation.ext.codecs.BinaryCodec.decoder
 
         PeerResponse[F, Signed[GlobalIncrementalSnapshot]](s"global-snapshots/${ordinal.value.value}")(client)
+      }
+
+      def getHash(ordinal: SnapshotOrdinal): PeerResponse[F, Option[Hash]] = {
+        import org.http4s.circe.CirceEntityCodec.circeEntityDecoder
+
+        PeerResponse(s"global-snapshots/${ordinal.value.value}/hash", POST)(client, session) { (req, client) =>
+          client.expectOption[Hash](req)
+        }
       }
 
       def get(hash: Hash): PeerResponse[F, Signed[GlobalIncrementalSnapshot]] = {
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/http/routes/SnapshotRoutes.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/http/routes/SnapshotRoutes.scala
@@ -91,6 +91,16 @@ final case class SnapshotRoutes[F[_]: Async: KryoSerializer, S <: Snapshot[_, _]
           }
         }.getOrElse(NotFound())
 
+    case GET -> Root / SnapshotOrdinalVar(ordinal) / "hash" =>
+      import org.http4s.circe.CirceEntityCodec.circeEntityEncoder
+
+      snapshotStorage
+        .getHash(ordinal)
+        .flatMap {
+          case None           => NotFound()
+          case Some(snapshot) => Ok(snapshot)
+        }
+
     case req @ GET -> Root / HashVar(hash) =>
       resolveEncoder[F, Signed[S]](req) { implicit enc =>
         snapshotStorage.get(hash).flatMap {
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
@@ -5,7 +5,6 @@ import cats.data.NonEmptyList
 import cats.effect.Async
 import cats.effect.std.Random
 import cats.effect.syntax.concurrent._
-import cats.syntax.applicativeError._
 import cats.syntax.eq._
 import cats.syntax.flatMap._
 import cats.syntax.functor._
@@ -74,10 +73,10 @@ object MajorityPeerSelect {
       ordinalDistribution = peerOrdinals.groupMap { case (_, ordinal) => ordinal } { case (peer, _) => peer }
       (majorityOrdinal, _) = latestOrdinals.groupBy(identity).maxBy { case (_, ordinals) => ordinals.size }
       peerDistribution <- peers
-        .parTraverseN(maxConcurrentPeerInquiries)(peer => getSnapshotHashByPeer(peer, majorityOrdinal).attempt)
-        .flatMap { peerSnapshotHashes =>
+        .parTraverseN(maxConcurrentPeerInquiries)(getSnapshotHashByPeer(_, majorityOrdinal))
+        .flatMap { maybePeerSnapshotHashes =>
           MonadThrow[F].fromOption(
-            peerSnapshotHashes.collect { case Right(peerSnapshot) => peerSnapshot }.toNel,
+            maybePeerSnapshotHashes.toList.flatten.toNel,
             NoPeersToSelect
           )
         }
@@ -103,7 +102,7 @@ object MajorityPeerSelect {
         .map(_.take(Math.max(sampleSize, minSampleSize)))
     }
 
-    def getSnapshotHashByPeer(peer: Peer, ordinal: SnapshotOrdinal): F[(Peer, Hash)] =
-      snapshotClient.get(ordinal).run(peer).flatMap(_.toHashed).map(snapshot => (peer, snapshot.hash))
+    def getSnapshotHashByPeer(peer: Peer, ordinal: SnapshotOrdinal): F[Option[(Peer, Hash)]] =
+      snapshotClient.getHash(ordinal).run(peer).map(_.map((peer, _)))
   }
 }
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/storage/SnapshotStorage.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/storage/SnapshotStorage.scala
@@ -181,6 +181,10 @@ object SnapshotStorage {
             case None    => snapshotLocalFileSystemStorage.read(hash)
           }
 
+        def getHash(ordinal: SnapshotOrdinal): F[Option[Hash]] = get(ordinal).flatMap {
+          _.traverse(_.toHashed.map(_.hash))
+        }
+
         def getLatestBalances: F[Option[Map[Address, Balance]]] =
           headRef.get.map(_.map(_._2.balances))
 
--- a/modules/core/src/test/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctionsSuite.scala
+++ b/modules/core/src/test/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctionsSuite.scala
@@ -63,6 +63,8 @@ object GlobalSnapshotConsensusFunctionsSuite extends MutableIOSuite with Checker
 
       override def get(hash: Hash): IO[Option[Signed[GlobalIncrementalSnapshot]]] = ???
 
+      override def getHash(ordinal: SnapshotOrdinal): F[Option[Hash]] = ???
+
     }
 
   val bam: BlockAcceptanceManager[IO, DAGTransaction, DAGBlock] = new BlockAcceptanceManager[IO, DAGTransaction, DAGBlock] {
--- a/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/Main.scala
+++ b/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/Main.scala
@@ -44,7 +44,7 @@ object Main
 
     for {
       _ <- Resource.unit
-      p2pClient = P2PClient.make[IO](sdkP2PClient, sdkResources.client, keyPair, method.identifier)
+      p2pClient = P2PClient.make[IO](sdkP2PClient, sdkResources.client, method.identifier)
       queues <- Queues.make[IO](sdkQueues).asResource
       storages <- Storages.make[IO](sdkStorages, cfg.snapshot, method.globalL0Peer).asResource
       validators = Validators.make[IO](seedlist)
--- a/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/http/P2PClient.scala
+++ b/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/http/P2PClient.scala
@@ -1,7 +1,5 @@
 package org.tessellation.currency.l0.http
 
-import java.security.KeyPair
-
 import cats.effect.Async
 
 import org.tessellation.currency.l0.snapshot.CurrencySnapshotClient
@@ -19,7 +17,6 @@ object P2PClient {
   def make[F[_]: Async: SecurityProvider: KryoSerializer](
     sdkP2PClient: SdkP2PClient[F],
     client: Client[F],
-    keyPair: KeyPair,
     identifier: Address
   ): P2PClient[F] =
     new P2PClient[F](
@@ -28,7 +25,7 @@ object P2PClient {
       sdkP2PClient.gossip,
       sdkP2PClient.node,
       StateChannelSnapshotClient.make(client, identifier),
-      L0GlobalSnapshotClient.make[F](client),
+      sdkP2PClient.l0GlobalSnapshot,
       CurrencySnapshotClient.make[F](client)
     ) {}
 }
--- a/modules/dag-l1/src/main/scala/org/tessellation/dag/l1/http/p2p/P2PClient.scala
+++ b/modules/dag-l1/src/main/scala/org/tessellation/dag/l1/http/p2p/P2PClient.scala
@@ -3,7 +3,6 @@ package org.tessellation.dag.l1.http.p2p
 import cats.effect.Async
 
 import org.tessellation.dag.l1.domain.consensus.block.http.p2p.clients.BlockConsensusClient
-import org.tessellation.kryo.KryoSerializer
 import org.tessellation.schema.Block
 import org.tessellation.schema.transaction.Transaction
 import org.tessellation.sdk.http.p2p.SdkP2PClient
@@ -17,7 +16,7 @@ import org.http4s.client._
 object P2PClient {
 
   def make[
-    F[_]: Async: SecurityProvider: KryoSerializer,
+    F[_]: Async: SecurityProvider,
     T <: Transaction: Encoder,
     B <: Block[T]: Encoder
   ](
@@ -33,7 +32,7 @@ object P2PClient {
       L0CurrencyClusterClient.make(currencyPathPrefix, client),
       sdkP2PClient.gossip,
       BlockConsensusClient.make(client),
-      L0GlobalSnapshotClient.make[F](client)
+      sdkP2PClient.l0GlobalSnapshot
     ) {}
 }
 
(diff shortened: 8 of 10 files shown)


modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/L0GlobalSnapshotClient.scala
```diff
@@ -56,5 +57,11 @@ object L0GlobalSnapshotClient {
 
         PeerResponse[F, Signed[GlobalIncrementalSnapshot]](s"global-snapshots/$hash")(client)
       }
+
+      def getHash(ordinal: SnapshotOrdinal): PeerResponse[F, Hash] = {
+        import org.http4s.circe.CirceEntityCodec.circeEntityDecoder
+
+        PeerResponse[F, Hash](s"global-snapshots/hash/${ordinal.value.value}")(client)
+      }
```
marcinwadon, line 65: if `NotFound` is an expected http status (I believe it is as someone can request for future ordinal which does not exist yet), then the client should handle that status accordingly. IMO the returned type should be `PeerResponse[F, Option[Hash]]` and `NotFound` should be mapped to `None` then

marcinwadon (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/http/routes/SnapshotRoutes.scala
```diff
@@ -64,6 +65,17 @@ final case class SnapshotRoutes[F[_]: Async: KryoSerializer, S <: Snapshot[_, _]
         }
       }
 
+    case GET -> Root / "hash" / SnapshotOrdinalVar(ordinal) =>
+      import org.http4s.circe.CirceEntityCodec.circeEntityEncoder
+
+      snapshotStorage
+        .get(ordinal)
+        .flatMap(_.traverse(_.toHashed.map(_.hash)))
```
TheMMaciek, line 73: I think it makes more sense to keep that logic in the `snapshotStorage`. I mean additional method there e.g. `getHash(ordinal: SnapshotOrdinal): F[Option[Hash]]` that fetches the snapshot and hashes it, there could also be an optimization there where in the case the hash is in the `ordinalCache` we don't need to rehash the snapshot.

In the future I would consider if having a cache of ordinal to hash mapping would make sense (to not hash a snapshot on every request which could be computationally costly), maybe we could figure out a way to check what file does the ordinal link to and get the hash that way (snapshots are saved on the disk under hash name and a link by ordinal is created to them). But let's not optimize prematurely and get it working first.

TheMMaciek, line 73: I think it makes more sense to keep that logic in the `snapshotStorage`. I mean additional method there e.g. `getHash(ordinal: SnapshotOrdinal): F[Option[Hash]]` that fetches the snapshot and hashes it, there could also be an optimization there where in the case the hash is in the `ordinalCache` we don't need to rehash the snapshot.

In the future I would consider if having a cache of ordinal to hash mapping would make sense (to not hash a snapshot on every request which could be computationally costly), maybe we could figure out a way to check what file does the ordinal link to and get the hash that way (snapshots are saved on the disk under hash name and a link by ordinal is created to them). But let's not optimize prematurely and get it working first.

TheMMaciek (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/http/routes/SnapshotRoutes.scala
marcinwadon: I support that part
> to not hash a snapshot on every request which could be computationally costly
it's a heavy operation and can be some vector of attack on the node (DoS), just keep that in mind

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/L0GlobalSnapshotClient.scala
```diff
@@ -59,6 +61,12 @@ object L0GlobalSnapshotClient {
         PeerResponse[F, Signed[GlobalIncrementalSnapshot]](s"global-snapshots/${ordinal.value.value}")(client)
       }
 
+      def getHash(ordinal: SnapshotOrdinal): PeerResponse[F, Option[Hash]] = {
+        import org.http4s.circe.CirceEntityCodec.circeEntityDecoder
+
+        PeerResponse[F, Option[Hash]](s"global-snapshots/hash/${ordinal.value.value}")(client)
```
marcinwadon, line 67: `hash` and `ordinal` are a property of snapshot. `ordinal` is not a property of `hash`, so IMO the URL should be:
`/global-snapshots` -> list of all snapshots
`/global-snapshots/{ordinal}` -> one snapshot by ordinal
`/global-snapshots/{ordinal}/hash` -> hash of that snapshot
cc: @AlexBrandes

modules/sdk/src/main/scala/org/tessellation/sdk/http/routes/SnapshotRoutes.scala
```diff
@@ -91,6 +91,23 @@ final case class SnapshotRoutes[F[_]: Async: KryoSerializer, S <: Snapshot[_, _]
           }
         }.getOrElse(NotFound())
 
+    case GET -> Root / "hash" / SnapshotOrdinalVar(ordinal) :? FullSnapshotQueryParam(fullSnapshot) =>
+      import org.http4s.circe.CirceEntityCodec.circeEntityEncoder
+
+      if (!fullSnapshot) {
+        snapshotStorage.getHash(ordinal).flatMap {
```
marcinwadon, line 98: you can use `.flatMap(Ok(_))` as you did few lines below

modules/sdk/src/main/scala/org/tessellation/sdk/http/routes/SnapshotRoutes.scala
```diff
@@ -91,6 +91,23 @@ final case class SnapshotRoutes[F[_]: Async: KryoSerializer, S <: Snapshot[_, _]
...
+          Ok(_)
+        }
+      } else {
+        fullGlobalSnapshotStorage
+          .map(
+            _.readHash(ordinal)
+          )
+          .traverse(identity)
```
marcinwadon, line 106: `traverse(identity)` is just a `sequence` (and with additional `flatten` it might be a `flatSequence`. I'm not sure as I just deduct how it works by github code)

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
```diff
@@ -103,7 +102,7 @@ object MajorityPeerSelect {
         .map(_.take(Math.max(sampleSize, minSampleSize)))
     }
 
-    def getSnapshotHashByPeer(peer: Peer, ordinal: SnapshotOrdinal): F[(Peer, Hash)] =
-      snapshotClient.get(ordinal).run(peer).flatMap(_.toHashed).map(snapshot => (peer, snapshot.hash))
+    def getSnapshotHashByPeer(peer: Peer, ordinal: SnapshotOrdinal): F[Option[(Peer, Hash)]] =
+      snapshotClient.getHash(ordinal).run(peer).map(_.map((peer, _)))
```
marcinwadon, line 106: I see you don't use `getHashFull` method. Why is it required then?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
```diff
@@ -50,6 +51,9 @@ final class SnapshotLocalFileSystemStorage[F[_]: Async: KryoSerializer, S <: Sna
   def read(hash: Hash): F[Option[Signed[S]]] =
     read(hash.coerce[String])
 
+  def readHash(ordinal: SnapshotOrdinal): F[Option[Hash]] =
+    read(ordinal).flatMap(_.map(_.toHashed.map(_.hash)).traverse(identity))
```
marcinwadon, line 55: as above, traverse with identity is just a sequence

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/storage/SnapshotStorage.scala
```diff
@@ -181,6 +181,10 @@ object SnapshotStorage {
             case None    => snapshotLocalFileSystemStorage.read(hash)
           }
 
+        def getHash(ordinal: SnapshotOrdinal): F[Option[Hash]] = get(ordinal).flatMap(
+          _.map(_.toHashed.map(_.hash)).traverse(identity)
```
marcinwadon, line 185: as above

marcinwadon: but also I think it might be just `_.traverse(_.toHashed.map(_.hash))` but I may be wrong

marcinwadon (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
cngo-github: I added `getHashFull` because there was `get` and `getFull` for `SnapshotOrdinal`. It has been removed.

modules/sdk/src/main/scala/org/tessellation/sdk/http/routes/SnapshotRoutes.scala
cngo-github: Changed

```scala
flatMap {
  Ok(_)
}
```

to `.flatMap(Ok(_))`.

modules/sdk/src/main/scala/org/tessellation/sdk/http/routes/SnapshotRoutes.scala
cngo-github: Changed

```scala
          .map(
            _.readHash(ordinal)
          )
          .traverse(identity)
```
to

```scala
          .map(
            _.readHash(ordinal)
          )
          .flatSequence
```

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/storage/SnapshotStorage.scala
cngo-github: Changed

```scala
_.map(_.toHashed.map(_.hash)).traverse(identity)
```

to

```scala
_.traverse(_.toHashed.map(_.hash))
```

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/L0GlobalSnapshotClient.scala
cngo-github: I will need guidance here.

```scala
    case GET -> Root / SnapshotOrdinalVar(ordinal) :? FullSnapshotQueryParam(fullSnapshot) / "hash" =>
```

gives me

```
pattern type is incompatible with expected type;
 found   : org.http4s.Uri.Path
 required: scala.collection.immutable.Map[String,scala.collection.Seq[String]]
```

marcinwadon: I think it's because the url you constructed `/global-snapshots/{ordinal}?full/hash` is invalid.
see https://en.wikipedia.org/wiki/Query_string and https://www.ibm.com/docs/en/rtw/9.0.0?topic=resources-creating-web-url-schema

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
```diff
@@ -50,6 +51,12 @@ final class SnapshotLocalFileSystemStorage[F[_]: Async: KryoSerializer, S <: Sna
   def read(hash: Hash): F[Option[Signed[S]]] =
     read(hash.coerce[String])
 
+  def readHash(ordinal: SnapshotOrdinal): F[Option[Hash]] =
+    read(ordinal).flatMap { test =>
+      val test2 = test.map(_.toHashed.map(_.hash)).sequence
+      test2
```
marcinwadon: what is `test2`? `map` + `sequence` = `traverse`

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
cngo-github: Changed

```scala
read(ordinal).flatMap(_.map(_.toHashed.map(_.hash)).traverse(identity))
```

to

```scala
read(ordinal).flatMap(_.map(_.toHashed.map(_.hash)).sequence)
```

modules/sdk/src/main/scala/org/tessellation/sdk/http/routes/SnapshotRoutes.scala
```diff
@@ -91,6 +91,22 @@ final case class SnapshotRoutes[F[_]: Async: KryoSerializer, S <: Snapshot[_, _]
...
+      if (!fullSnapshot) {
+        snapshotStorage
+          .getHash(ordinal)
+          .flatMap(Ok(_))
+      } else {
+        fullGlobalSnapshotStorage
+          .map(
+            _.readHash(ordinal)
+          )
+          .flatSequence
```
marcinwadon, line 106: if you use map then sequence, it can be a single operation like `traverse`, but if you use flatten after that, it can be a `flatTraverse`

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
```diff
@@ -50,6 +51,9 @@ final class SnapshotLocalFileSystemStorage[F[_]: Async: KryoSerializer, S <: Sna
   def read(hash: Hash): F[Option[Signed[S]]] =
     read(hash.coerce[String])
 
+  def readHash(ordinal: SnapshotOrdinal): F[Option[Hash]] =
+    read(ordinal).flatMap(_.map(_.toHashed.map(_.hash)).sequence)
```
marcinwadon, line 55: map + sequence = traverse

marcinwadon (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
cngo-github: `test2` was for debugging. It has been removed. This has been changed to

```scala
read(ordinal).flatMap(_.traverse(_.toHashed.map(_.hash)))
```

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
cngo-github: This has been changed to

```scala
read(ordinal).flatMap(_.traverse(_.toHashed.map(_.hash)))
```

modules/sdk/src/main/scala/org/tessellation/sdk/http/routes/SnapshotRoutes.scala
cngo-github: This has been changed to

```scala
        fullGlobalSnapshotStorage
          .flatTraverse(
            _.readHash(ordinal)
          )
          .flatMap(Ok(_))
```

modules/sdk/src/main/scala/org/tessellation/sdk/http/routes/SnapshotRoutes.scala
```diff
@@ -91,6 +91,21 @@ final case class SnapshotRoutes[F[_]: Async: KryoSerializer, S <: Snapshot[_, _]
           }
         }.getOrElse(NotFound())
 
+    case GET -> Root / "hash" / SnapshotOrdinalVar(ordinal) :? FullSnapshotQueryParam(fullSnapshot) =>
```
marcinwadon, line 94: What about the NotFound http status that was there before?

modules/sdk/src/main/scala/org/tessellation/sdk/http/routes/SnapshotRoutes.scala
cngo-github: The logics to get the hash have been moved to the storages. I made a ticket for the optimization referenced.

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/L0GlobalSnapshotClient.scala
cngo-github: The route is returning an `Option[Hash]` now.

modules/sdk/src/main/scala/org/tessellation/sdk/http/routes/SnapshotRoutes.scala
cngo-github: The route is returning an `Option[Hash]` now. So, if it doesn't find the hash for the snapshot at the specified `SnapshotOrdinal`, it returns an `Ok(None)`. There is no scenario where it will return a `NotFound`.

marcinwadon: Why does it return Some/None instead of Ok and NotFound status codes? IMO if there is no resource, the endpoint should return NotFound
@AlexBrandes what do you think?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
marcinwadon: and as written in many other comments - map + sequence = traverse

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/L0GlobalSnapshotClient.scala
marcinwadon: @cngo-github the URL looks to be in different order (`ordinal` is not a property of `hash`, so it shouldn't be `/hash/{ordinal}`). Take a look at a suggested URL above

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
cngo-github: This has been removed since it is not being used by an endpoint.

modules/sdk/src/main/scala/org/tessellation/sdk/http/routes/SnapshotRoutes.scala
cngo-github: The route is now returning `Ok(Hash)` or `NotFound`. `L0GlobalSnapshotClient` converts successful calls to `Some(Hash)` and failures to `none[Hash]`.

marcinwadon: Great. All failures or just the NotFound status code?

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/L0GlobalSnapshotClient.scala
cngo-github: The route has been changed to `/global-snapshots/${ordinal}/hash`. There is no route for full snapshots.

modules/sdk/src/main/scala/org/tessellation/sdk/http/routes/SnapshotRoutes.scala
cngo-github: All failures.

marcinwadon: Is it correct behavior? Shouldn't anything else throw to stop the algorithm or something? If no,and it's correct then we're good.

modules/sdk/src/main/scala/org/tessellation/sdk/http/routes/SnapshotRoutes.scala
```diff
@@ -91,6 +91,16 @@ final case class SnapshotRoutes[F[_]: Async: KryoSerializer, S <: Snapshot[_, _]
...
 
+    case GET -> Root / SnapshotOrdinalVar(ordinal) / "hash" =>
+      import org.http4s.circe.CirceEntityCodec.circeEntityEncoder
+
+      snapshotStorage
+        .getHash(ordinal)
+        .flatMap {
+          case None           => NotFound()
+          case Some(snapshot) => Ok(snapshot)
+        }
```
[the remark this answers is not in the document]
marcinwadon: Nice, didn't know about that 💪

modules/sdk/src/main/scala/org/tessellation/sdk/http/routes/SnapshotRoutes.scala
cngo-github: This behavior was implemented because as @sdavidp and I were tracing `PeerResponse`, we were unsure of how it handled errors. We thought that it might be converting all errors into `unauthenticated` throwables. Seeing as we didn't want to modify `PeerResponse` or the middleware it uses for the actual call, we simply converted all errors to `None`.
Part of the rationale for doing that was also the thought that if any error was encountered, a `none[Hash]` makes sense as a return value.

marcinwadon: I think the easiest would be to talk with protocol_eu and make quick pair programming. Anyway, for me personally, "none"/"not found" sounds like a domain error (cannot find a snapshot) but I know the request was correctly processed. If I have an error, I can retry the request if needed or do some other logic. Right now we're not able to distinguish whether the request was incorrect or the node has no snapshot. If you're good with that then fine but in my opinion it should reflect the real state (found, not found, issue with the request/response).

sdavidp: > Right now we're not able to distinguish whether the request was incorrect or the node has no snapshot. If you're good with that then fine but in my opinion it should reflect the real state (found, not found, issue with the request/response).

Yes, this is why we were asking about where the 404 was mapped to an error in `F` so we could distinguish between that and real errors. I still haven't found it in _http4s_ or our code.

modules/sdk/src/main/scala/org/tessellation/sdk/http/routes/SnapshotRoutes.scala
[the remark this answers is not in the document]
sdavidp: > Nice, didn't know about that 💪

I could be mistaken :(
I've used helper methods so much that I often forget whether it's something we wrote in-house or whether it was part of the library. I found `.orNotFound` for Kleisli's but haven't been able to find it for this case yet.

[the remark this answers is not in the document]
marcinwadon: there is `orNotFound` which is a handler for unmatched paths, but you can always write a helper fn if you want

modules/sdk/src/main/scala/org/tessellation/sdk/http/routes/SnapshotRoutes.scala
cngo-github: Only `Status.NotFound` and `Status.None` return `none[Hash]` now.

marcinwadon: You mean Status.Gone? :) 
Cool, I'll take a look tomorrow morning and approve/merge

marcinwadon (approved):

TheMMaciek (approved):

PR #648, opened 2023-05-15
Replace inheritance with composition

tbekas:
The work isn't finished yet however, the common part is already extracted under the `CurrencyData` model. And most of the functions are switched to using it.


marcinwadon:
Work continued in #703

PR #649, opened 2023-05-17
DataApplication implementation for Currency L0

marcinwadon:
Here is an up-to-date complete and working example:

```scala
object Acme {

  @derive(decoder, encoder)
  sealed trait Usage extends DataUpdate {
    val value: Long
  }

  @derive(decoder, encoder)
  case class EnergyUsage(value: Long) extends Usage

  @derive(decoder, encoder)
  case class WaterUsage(value: Long) extends Usage

  @derive(decoder, encoder)
  case class State(totalEnergyUsage: Long, totalWaterUsage: Long) extends DataState

  case object EnergyNotPositive extends DataApplicationValidationError {
    val message = "energy ussage must be positive"
  }
  case object WaterNotPositive extends DataApplicationValidationError {
    val message = "water ussage must be positive"
  }
  case object Noop extends DataApplicationValidationError {
    val message = "invalid update"
  }

  val energy: DataApplicationL0Service[IO, Usage, State] = new DataApplicationL0Service[IO, Usage, State] {
    def validateData(oldState: State, updates: NonEmptyList[Usage]): IO[DataApplicationValidationErrorOr[Unit]] =
      IO {
        updates.map {
          case update: EnergyUsage =>
            if (update.value > 0L) ().validNec else EnergyNotPositive.asInstanceOf[DataApplicationValidationError].invalidNec
          case update: WaterUsage =>
            if (update.value > 0L) ().validNec else WaterNotPositive.asInstanceOf[DataApplicationValidationError].invalidNec
        }.reduce
      }

    def combine(oldState: State, updates: NonEmptyList[Usage]): IO[State] =
      IO {
        updates.foldLeft(oldState) { (acc, update) =>
          update match {
            case EnergyUsage(usage) => State(acc.totalEnergyUsage + usage, acc.totalWaterUsage)
            case WaterUsage(usage)  => State(acc.totalEnergyUsage, acc.totalWaterUsage + usage)
          }
        }
      }

    def serializeState(state: State): IO[Array[Byte]] = IO {
      state.asJson.toString.getBytes(StandardCharsets.UTF_8)
    }

    def deserializeState(bytes: Array[Byte]): IO[Either[Throwable, State]] = IO {
      parse(new String(bytes, StandardCharsets.UTF_8)).flatMap { json =>
        json.as[State]
      }
    }

    def dataEncoder: Encoder[Usage] = deriveEncoder

    def dataDecoder: Decoder[Usage] = deriveDecoder

    def genesis: State = State(0L, 0L)
  }
}

```

### DIFF
--- a/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/Main.scala
+++ b/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/Main.scala
@@ -4,8 +4,10 @@ import java.util.UUID
 
 import cats.effect.{IO, Resource}
 import cats.syntax.semigroupk._
+import cats.syntax.traverse._
 
 import org.tessellation.BuildInfo
+import org.tessellation.currency._
 import org.tessellation.currency.l0.cli.method
 import org.tessellation.currency.l0.cli.method._
 import org.tessellation.currency.l0.http.P2PClient
@@ -37,6 +39,8 @@ object Main
   val kryoRegistrar: Map[Class[_], KryoRegistrationId[KryoRegistrationIdRange]] =
     sdkKryoRegistrar
 
+  val dataApplication: Option[BaseDataApplicationL0Service[IO]] = None
+
   def run(method: Run, sdk: SDK[IO]): Resource[IO, Unit] = {
     import sdk._
 
@@ -52,14 +56,13 @@ object Main
           p2pClient,
           sdkServices,
           storages,
-          validators,
           sdkResources.client,
           sdkServices.session,
           sdk.seedlist,
           sdk.nodeId,
           keyPair,
           cfg,
-          method.identifier
+          dataApplication
         )
         .asResource
       programs = Programs.make[IO](
@@ -88,7 +91,7 @@ object Main
       rumorHandler = RumorHandlers.make[IO](storages.cluster, healthChecks.ping, services.localHealthcheck).handlers <+>
         services.consensus.handler
       _ <- Daemons
-        .start(storages, services, programs, queues, healthChecks)
+        .start(storages, services, programs, queues, healthChecks, dataApplication)
         .asResource
       api = HttpApi
         .make[IO](
@@ -143,7 +146,11 @@ object Main
             NodeState.Initial,
             NodeState.LoadingGenesis,
             NodeState.GenesisReady
-          )(programs.genesis.accept(m.genesisPath)) >> gossipDaemon.startAsInitialValidator >>
+          ) {
+            dataApplication
+              .traverse(_.serializedGenesis)
+              .flatMap(programs.genesis.accept(m.genesisPath, _))
+          } >> gossipDaemon.startAsInitialValidator >>
             services.cluster.createSession >>
             services.session.createSession >>
             programs.globalL0PeerDiscovery.discoverFrom(cfg.globalL0Peer) >>
--- a/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensus.scala
+++ b/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensus.scala
@@ -7,6 +7,7 @@ import cats.effect.std.{Random, Supervisor}
 
 import org.tessellation.currency.l0.snapshot.services.StateChannelSnapshotService
 import org.tessellation.currency.schema.currency._
+import org.tessellation.currency.{BaseDataApplicationL0Service, DataUpdate}
 import org.tessellation.kryo.KryoSerializer
 import org.tessellation.schema.SnapshotOrdinal
 import org.tessellation.schema.balance.Amount
@@ -22,6 +23,8 @@ import org.tessellation.sdk.infrastructure.metrics.Metrics
 import org.tessellation.sdk.infrastructure.snapshot.{CurrencySnapshotAcceptanceManager, SnapshotConsensus}
 import org.tessellation.security.SecurityProvider
 
+import io.circe.Decoder
+import io.circe.disjunctionCodecs._
 import org.http4s.client.Client
 
 object CurrencySnapshotConsensus {
@@ -39,14 +42,22 @@ object CurrencySnapshotConsensus {
     client: Client[F],
     session: Session[F],
     stateChannelSnapshotService: StateChannelSnapshotService[F],
-    snapshotAcceptanceManager: CurrencySnapshotAcceptanceManager[F]
-  ): F[SnapshotConsensus[F, CurrencyTransaction, CurrencyBlock, CurrencySnapshotArtifact, CurrencySnapshotContext, CurrencySnapshotEvent]] =
+    snapshotAcceptanceManager: CurrencySnapshotAcceptanceManager[F],
+    maybeDataApplication: Option[BaseDataApplicationL0Service[F]]
+  ): F[
+    SnapshotConsensus[F, CurrencyTransaction, CurrencyBlock, CurrencySnapshotArtifact, CurrencySnapshotContext, CurrencySnapshotEvent]
+  ] = {
+    def noopDecoder: Decoder[DataUpdate] = Decoder.failedWithMessage[DataUpdate]("not implemented")
+
+    implicit def daDecoder: Decoder[DataUpdate] = maybeDataApplication.map(_.dataDecoder).getOrElse(noopDecoder)
+
     Consensus.make[F, CurrencySnapshotEvent, SnapshotOrdinal, CurrencySnapshotArtifact, CurrencySnapshotContext](
       CurrencySnapshotConsensusFunctions.make[F](
         stateChannelSnapshotService,
         snapshotAcceptanceManager,
         collateral,
-        rewards
+        rewards,
+        maybeDataApplication
       ),
       gossip,
       selfId,
@@ -58,5 +69,5 @@ object CurrencySnapshotConsensus {
       client,
       session
     )
-
+  }
 }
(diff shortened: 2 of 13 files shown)


modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensus.scala
```diff
@@ -39,14 +42,22 @@ object CurrencySnapshotConsensus {
     client: Client[F],
     session: Session[F],
     stateChannelSnapshotService: StateChannelSnapshotService[F],
-    snapshotAcceptanceManager: CurrencySnapshotAcceptanceManager[F]
-  ): F[SnapshotConsensus[F, CurrencyTransaction, CurrencyBlock, CurrencySnapshotArtifact, CurrencySnapshotContext, CurrencySnapshotEvent]] =
+    snapshotAcceptanceManager: CurrencySnapshotAcceptanceManager[F],
+    maybeDataApplication: Option[BaseDataApplicationL0Service[F]]
+  ): F[
+    SnapshotConsensus[F, CurrencyTransaction, CurrencyBlock, CurrencySnapshotArtifact, CurrencySnapshotContext, CurrencySnapshotEvent]
```
TheMMaciek, line 48: Can we revert this formatting?

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
```diff
@@ -61,7 +71,11 @@ object CurrencySnapshotConsensusFunctions {
       events: Set[CurrencySnapshotEvent]
     ): F[(CurrencySnapshotArtifact, CurrencySnapshotContext, Set[CurrencySnapshotEvent])] = {
 
-      val blocksForAcceptance = events
+      val (blocks: List[Signed[CurrencyBlock]], dataBlocks: List[Signed[DataApplicationBlock]]) = events.filter { event =>
+        if (maybeDataApplicationService.isEmpty) event.isLeft else true
```
TheMMaciek, line 75: I think it should be possible to write the filter without returning `true`

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
```diff
@@ -71,6 +85,46 @@ object CurrencySnapshotConsensusFunctions {
...
+        maybeLastDataState <- (maybeLastDataApplication, maybeDataApplicationService).mapN {
+          case ((lastDataApplication, service)) =>
+            service
+              .deserializeState(lastDataApplication)
+              .flatTap {
+                case Left(err) => logger.warn(err)(s"Cannot deserialize custom state")
+                case Right(a)  => logger.info(s"Deserialized state: ${a}") >> Applicative[F].unit
+              }
+              .map(_.toOption)
+              .handleErrorWith(err => logger.error(err)(s"Cannot deserialize custom state") >> none[DataState].pure[F])
```
TheMMaciek, line 99: A different log message (then the one at 95) would be good.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
marcinwadon: I've copied as we have it in the global l0 consensus. any idea how to change that?

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensus.scala
marcinwadon: we cannot, it's too large now

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
marcinwadon: changed to use `collect`

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/Main.scala
```diff
@@ -144,7 +147,11 @@ object Main
             NodeState.Initial,
             NodeState.LoadingGenesis,
             NodeState.GenesisReady
-          )(programs.genesis.accept(m.genesisPath)) >> gossipDaemon.startAsInitialValidator >>
+          ) {
+            dataApplication.traverse { da =>
+              da.serializeState(da.genesis)
+            }.flatMap(programs.genesis.accept(m.genesisPath, _))
```
kpudlik, line 153: `da.serializeState(da.genesis)` seems to operate on the wrong abstraction level. You could hide the details in something like: `da.serializedGenesis` so `Main` doesn't need to know the details.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
```diff
@@ -61,7 +71,12 @@ object CurrencySnapshotConsensusFunctions {
       events: Set[CurrencySnapshotEvent]
     ): F[(CurrencySnapshotArtifact, CurrencySnapshotContext, Set[CurrencySnapshotEvent])] = {
 
-      val blocksForAcceptance = events
+      val (blocks: List[Signed[CurrencyBlock]], dataBlocks: List[Signed[DataApplicationBlock]]) = events.collect {
+        case event @ Left(_)                                           => event
+        case event @ Right(_) if maybeDataApplicationService.isDefined => event
+      }.toList.partitionMap(identity)
```
kpudlik, line 77: `Left` case is just `a => a`. Also `Right` case is just a filter on `Right` side.

```scala
events
  .filter(_ => maybeDataApplicationService.isDefined)
  .toList
  .partitionMap(identity)
```

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
```diff
@@ -71,6 +86,52 @@ object CurrencySnapshotConsensusFunctions {
...
 
+        maybeLastDataApplication = lastArtifact.data
+
+        maybeLastDataState <- (maybeLastDataApplication, maybeDataApplicationService).mapN {
+          case ((lastDataApplication, service)) =>
+            service
+              .deserializeState(lastDataApplication)
+              .flatTap {
+                case Left(err) => logger.warn(err)(s"Cannot deserialize custom state")
+                case Right(a)  => logger.info(s"Deserialized state: ${a}") >> Applicative[F].unit
```
kpudlik, line 97: Isn't `Applicative[F].unit` redundant?

kpudlik (changes requested):

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/Main.scala
marcinwadon: :+1:

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
marcinwadon: yes, it is

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
TheMMaciek: I don't think filter is right biased here. I think it will filter out all events (lefts and rights) based on whether  dataApplicationService is none or some.

marcinwadon: @kpudlik: @TheMMaciek is right. `filter` on Either is right biased, but `events` are `Set[Either`. anyway, I've updated the code to use `filter` instead of `collect` which seems to be more readable

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
```diff
@@ -61,7 +70,15 @@ object CurrencySnapshotConsensusFunctions {
       events: Set[CurrencySnapshotEvent]
     ): F[(CurrencySnapshotArtifact, CurrencySnapshotContext, Set[CurrencySnapshotEvent])] = {
 
-      val blocksForAcceptance = events
+      val (blocks: List[Signed[CurrencyBlock]], dataBlocks: List[Signed[DataApplicationBlock]]) =
+        events.filter {
+          case Left(_)  => true
+          case Right(_) => maybeDataApplicationService.isDefined
+
+        }.toList
```
kpudlik, line 78: eventually then:

```scala
events.filter(_.isLeft || maybeDataApplicationService.isDefined)
```

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
kpudlik: `events.filter(_.isLeft || maybeDataApplicationService.isDefined)` then :D

kpudlik (approved):

PR #653, opened 2023-05-24
Additional validations for currency snapshot on Global L0

kpudlik:


marcinwadon (approved):
Approved. I see a duplication of currency snapshot definition but I'm going to remove it anyway in my next PR.

PR #655, opened 2023-05-25
Expand SnapshotClient, use in MajorityPeerSelect

sdavidp:


modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/PeerResponse.scala
```diff
@@ -41,14 +41,21 @@ object PeerResponse {
...
+      }
   }
 
   def apply[F[_], G[_], A](uri: Uri => Uri, method: Method) =
     new PeerResponsePartaillyApplied[F, G, A](uri, method)
   def apply[F[_], G[_], A](path: String, method: Method) =
     new PeerResponsePartaillyApplied[F, G, A]((uri: Uri) => uri.addPath(path), method)
 
-  private[p2p] final class GetPeerResponsePartaillyApplied[F[_], A](val uri: Uri => Uri) {
+  private[p2p] final class GetPeerResponsePartiallyApplied[F[_], A](val uri: Uri => Uri) {
```
marcinwadon, line 58: :+1:

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/L0SnapshotClient.scala
```diff
@@ -4,35 +4,68 @@ import cats.effect.Async
...
 import org.tessellation.security.signature.Signed
 
 import io.circe.Decoder
 import io.circe.magnolia.derivation.decoder.semiauto._
 import io.circe.refined._
+import org.http4s.Method.GET
 import org.http4s.client.Client
 
-object L0SnapshotClient {
+abstract class L0SnapshotClient[
```
marcinwadon, line 22: if this is an abstract class, why you specify a separate trait for definitions?
imo either you can use that abstract class as a type, or move the implementation to trait (which is also possible). I don't know which option would be better, I think it's up to you. what do you think?

Also, I think that there should be:
```
SnapshotClient - this is that class, the base one
GlobalSnapshotClient - (initialized on the Global L0 layer with session and on the other layers with no session)
CurrencySnapshotClient - (initialized on the Currency L0 layer with session and on the other layers with no session)
```
`L0SnapshotClient` name can be misleading as it's a generic implementation now

sdavidp: Yes, I agree we can remove the trait and rename the abstract class.

I'm confused about your second comment: The implementation for the `CurrencySnapshotClient` was refactored to use the session. So should it do that, or not? Or should there be another client for currency that does not use the session?

Unless I'm missing something there are only 2 clients in use in the codebase right now:
The global client with no session
The currency client with the session

sdavidp: Changes pushed

marcinwadon: as written on slack yesterday, there should be four (or three) clients:
- The global client with session (should be used in Global L0 download)
- The global client with no session (should be used in DAG L1 and Currency L0/L1)
- The currency client with session (should be used on Currency L0)
- The currency client with no session (should be used on Currency L1 if needed. probably not needed)

sdavidp: @marcinwadon Ok, I think the safest thing to do is remove the global snapshot client from `SdkP2PClient` otherwise it's confusing/error-prone how to create it and too easy to use the wrong client in the wrong place. Then anyone who needs to create the Global and Currency Snapshot Clients should explicitly create the one they need.

marcinwadon: Good idea :+1: as long as it's not used in the SDK module (but definitions imo should be in the sdk)

sdavidp: Ok, changes pushed. I removed the global client from the there and now the `P2PClient` objects in each module define the global (and currency, if needed) snapshot clients appropriately for that module.

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/L0GlobalSnapshotClient.scala
```diff
@@ -4,82 +4,40 @@ import cats.effect.Async
...
+    val urlPrefix = "global-snapshots"
+
+    def getFull(ordinal: SnapshotOrdinal): PeerResponse[F, Signed[GlobalSnapshot]] = {
+      import org.tessellation.ext.codecs.BinaryCodec.decoder
+
+      PeerResponse[F, Signed[GlobalSnapshot]]((uri: Uri) => uri.addPath(s"$urlPrefix/${ordinal.value.value}").withQueryParam("full"))(
+        client
+      )
     }
+  }
```
marcinwadon, line 42: one tip. you can use an anonymous class, instead of `Impl` suffix.

```
def make[F[_]: ...](...): L0GlobalSnapshotClient[F] = new SnapshotClient[F, GlobalIncrementalSnapshot, GlobalSnapshotInfo](client, maybeSession) extends L0GlobalSnapshotClient {
  // implement methods there
}
```
or some kind of that

marcinwadon (dismissed):
LGTM. See one comment about the anonymous class - imo it's more readable that way, but up to you :)

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/L0GlobalSnapshotClient.scala
sdavidp: Do you have a working example? You can't do it that way because `SnapshotClient` is abstract - compiler won't let us create it.

marcinwadon: I think you can create an instance of an abstract class. For sure you can create an instance of a trait. We use that in almost all storages/services. I cannot send you the link now but I'm sure you can quickly find that in repo

sdavidp: I couldn't get it to work that way but I removed the constructor args from the abstract class and did it that way.

sdavidp:
Confirmed the issue is fixed by using IntelliJ to launch 2 Genesis nodes (joined), 2 Currency L0 nodes (joined), and 2 Currency L1 nodes (joined).  On the develop branch I received this error joining the two Currency L0 nodes:
```
org.http4s.client.UnexpectedStatus: unexpected HTTP status: 401 Unauthorized for request GET http://127.0.0.1:9901/global-snapshots/latest/ordinal
```

ryle-goehausen (approved):
lgtm

PR #661, opened 2023-06-07
Adding send transactions action

IPadawans:


.github/config/containers/global-l0/.gitkeep
marcinwadon, line 1: those directories are being used by the github action only, so there is no need to keep them in the git repository.
can you create directories directly from the github action?

.github/action_scripts/send_transactions.js
```diff
@@ -0,0 +1,375 @@
...
+        if( secondWalletBalance === 9900 && thirdWalletBalance === 5000 ){
+            logMessage( `Amount sent to third wallet ${thirdWalletBalance} and not to second: ${secondWalletBalance}` );
+            logMessage( `FirstWalletBalance: ${firstWalletBalance}` );
+            logMessage( `SecondWalletBalance: ${secondWalletBalance}` );
+            logMessage( `ThirdWalletBalance: ${thirdWalletBalance}` );
+            return;
+        }
+
+        if( secondWalletBalance === 14900 && thirdWalletBalance === 0 ){
+            logMessage( `Amount sent to second wallet ${thirdWalletBalance} and not to third: ${secondWalletBalance}` );
```
marcinwadon, line 230: `second wallet ${thirdWalletBalance}`
shouldn't it be `secondWalletBalance`? the same with third at the end of log message

.github/action_scripts/send_transactions.js
```diff
@@ -0,0 +1,375 @@
...
+    const expectedDestinationBalance = isInitial ? 11000 : 9900;
+
+    if(
+        Number( originBalance ) !== expectedOriginBalance ||
+        Number( destinationBalance ) !== expectedDestinationBalance
+    ) {
+        throw Error( `
+        Error sending transactions. Wallet balances are different than expected:
+        expectedOriginBalance: ${expectedOriginBalance} ---- originBalance: ${originBalance}
+        expectedDestinationBalance: ${expectedDestinationBalance} ---- ${destinationBalance}
```
marcinwadon, line 260: I'd write `destinationBalance:` in a log message as you did with `originBalance:`

.github/workflows/send-transactions-test.yml
```diff
@@ -0,0 +1,586 @@
+name: Test commit
+
+on:
+  pull_request: {}
```
marcinwadon, line 4: do we want to run it automatically on every push to pull requests? It may be time consuming and may block the pull request for a long time (so I'd recommend to turn it off as a required check). IMO it may be worth considering running it manually (then there is a button to manually trigger the action) or using some cron action (once a day for example).

.github/workflows/send-transactions-test.yml
```diff
@@ -0,0 +1,586 @@
...
+  push:
+    branches:
+      - develop
+      - hotfix/*
+      - release/*
```
marcinwadon, line 9: do we want also to run it on `main` branch and releases (versioned tags)?

.github/workflows/send-transactions-test.yml
```diff
@@ -0,0 +1,586 @@
...
+          mv modules/core/target/scala-2.13/tessellation-core-assembly-* .github/config/containers/jars
+          mv modules/keytool/target/scala-2.13/tessellation-keytool-assembly-* .github/config/containers/jars
+          mv modules/wallet/target/scala-2.13/tessellation-wallet-assembly-* .github/config/containers/jars
+          mv modules/dag-l1/target/scala-2.13/tessellation-dag-l1-assembly-* .github/config/containers/jars
+          mv modules/currency-l0/target/scala-2.13/tessellation-currency-l0-assembly-* .github/config/containers/jars
+          mv modules/currency-l1/target/scala-2.13/tessellation-currency-l1-assembly-* .github/config/containers/jars
+          
+          cd .github/config/containers/jars
+          
+          mv tessellation-core-assembly-* global-l0.jar
```
marcinwadon, line 52: I think you can name the jar file in the command above, like this:
```
mv modules/core/target/scala-2.13/tessellation-core-assembly-* .github/config/containers/jars/global-l0.jar
```

.github/workflows/send-transactions-test.yml
```diff
@@ -0,0 +1,586 @@
...
+        run: |          
+          cd .github/config/containers/global-l0
+          java -jar ../jars/cl-keytool.jar generate
+          
+          export CL_PUBLIC_HTTP_PORT=9000
+          export CL_P2P_HTTP_PORT=9001
+          export CL_CLI_HTTP_PORT=9002
+          export CL_APP_ENV=dev
+          export CL_ENV=dev
+          export CL_COLLATERAL=0
```
marcinwadon, line 74: I see you pass some attributes as cmd arguments and some as exported env variables. I think you can unify it either to env variables or cmd line 🤔 just food for thought

.github/workflows/send-transactions-test.yml
IPadawans: It was informed to me that this should behave the same as our current action that checks the code. Could we sync this on our meeting to see if we need to change. It takes around 15min to run completely it's not absolutely high but takes some time.

.github/workflows/send-transactions-test.yml
IPadawans: I think we can add the main branch too, I'll do that

.github/workflows/send-transactions-test.yml
marcinwadon: Sure, please add that to our meeting notes for tomorrow

.github/config/containers/global-l0/.gitkeep
IPadawans: Sure, good catch

marcinwadon:
> > @IPadawans in this repository we agreed to keep a linear history as much as possible with a single commit being a single self-contained change. Commits like "CR fixes" or "changes", or other like that should be either ammended or squashed to the related commit (the commit the fix is related to) if the original commit has not been merged yet.
> 
> Oh sorry for that @marcinwadon , what I should do now to keep aligned with you guys? Should I create another PR with all the changes in another branch or something like that

I think there is no need to create another PR. You can always change the commits you already pushed to your branch and pull request will keep updated. IMO either ammend the fix commit to one of the other commits, or just squash it using interactive rebase `rebase -i`. up to you, whatever is easier for you :)

IPadawans:
> 

Hmm, I've never done this before so I'm kinda confused for the best approach to take, I've tried to squash the commits on Intellij, but I'm not sure if I did it right

marcinwadon (approved):

PR #665, opened 2023-06-13
Break Currency L0 validator joining to Currency L0 initial validator

sdavidp:
**This PR is not meant to be merged**

It's sole purpose is to see if the automated github actions will fail by reverting a fix done a few PRs ago:
We expect the github actions to fail.

When I ran it locally:

1. Launch Global initial (genesis) node
2. Launch Currency L0 initial validator (c0)
3. Launch Currency L0 validator (c1)
4. Join c1 to c0
5. 404 error reported in the node logs

The error does not happen in the current develop branch

cc: @ryle-goehausen @AlexBrandes

### DIFF
--- a/modules/core/src/main/scala/org/tessellation/http/p2p/P2PClient.scala
+++ b/modules/core/src/main/scala/org/tessellation/http/p2p/P2PClient.scala
@@ -1,7 +1,6 @@
 package org.tessellation.http.p2p
 
 import cats.effect.Async
-import cats.syntax.option._
 
 import org.tessellation.kryo.KryoSerializer
 import org.tessellation.sdk.domain.cluster.services.Session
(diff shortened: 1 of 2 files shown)


marcinwadon:
@sdavidp make sure that the action runs multiple L0 nodes. If the error does not occur, probably it runs a single node

TheMMaciek:
Is this PR still needed? 🤔

sdavidp:
> 

I kept it alive so we could re-check at a later date.

marcinwadon:
@sdavidp you can take a look at https://github.com/nektos/act and I believe we can close the pull request. whenever there would be a need for re-test you can create a new PR based on that branch. also, the branch is already outdated and noone is keeping it up to date

PR #666, opened 2023-06-13
Added ip address, alias, and bias to the seedlist

cngo-github:
Added ip address, alias, and bias to the seedlist


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/seedlist/types.scala
```diff
@@ -1,20 +1,40 @@
...
 
 object types {
 
-  case class SeedlistCSVEntry(id: String) {
-
-    def toPeerId: PeerId = PeerId(Hex(id))
-  }
+  case class SeedlistCSVEntry(
+    id: PeerId,
+    ipAddress: String,
```
ryle-goehausen, line 14: These types here are not backwards compatible, if an entry is missing an IpAddress (i.e. the current entries) you are enforcing an expected value here.

This entire PR appears not to be backwards compatible with existing seedlist. It should load the existing file throughout here independently, so we don't need to perform an update at the same time the code is changed.

These should be changed to optional fields, and the rest of the code should reflect that.

ryle-goehausen: NVM re: backwards compat, after more discussion offline, this appears to be correct, but the types still need to change to optional for clarity and test added for showing mixed line parsing.

cngo-github: `SeedlistCSVEntry` has been changed to:

```
  case class SeedlistCSVEntry(
    id: PeerId,
    ipAddress: Option[String],
    alias: Option[String],
    bias: Option[TrustValueRefined]
  )
```

The tests and decoder have been updated appropriately. I also added a test for mixed formats in the same seedlist file.

marcinwadon:
@cngo-github @ryle-goehausen 
when I analyze the pull request, I have two thoughts.

1. Is it possible to set peer id and node alias, but no ip and no trust for example? I mean - is it possible to set partial data or it's either peer id or all (peer id, alias, ip, trust)? If the latter, then instead of optional types, I'd specify two case classes:
```
case class Foo(peerId)
case class Bar(peerId, nodeAlias, nodeIp, trust)
```
and load either this or that. if all the fields can be optional and loaded as partial data, then it's fine to keep it as a single case class with optional fields.

2. What is the reason to store node ip? If you want to check it against the peer that is joining to the node it may be not possible. IP address can be easily spoofed using many headers. If you setup a floating ip on digital ocean (to have a static ip to your server), the DO by default route the public traffic through a different IP address than the floating ip. I'm trying to say that there can be many misalignments in the IP specified in the file and the IP node is joining from. What's more, having the seedlist centralized like it is right now will effectively block migrating a node to a new server (for example after credentials breach). We had an IP specified in the seedlist on mainnet 1.0 and due to those issues, we decided to remove it from the seedlist and not rely on it.

cngo-github:
> @cngo-github @ryle-goehausen when I analyze the pull request, I have two thoughts.
> 
> 1. Is it possible to set peer id and node alias, but no ip and no trust for example? I mean - is it possible to set partial data or it's either peer id or all (peer id, alias, ip, trust)? If the latter, then instead of optional types, I'd specify two case classes:
> 
> ```
> case class Foo(peerId)
> case class Bar(peerId, nodeAlias, nodeIp, trust)
> ```
> 
> and load either this or that. if all the fields can be optional and loaded as partial data, then it's fine to keep it as a single case class with optional fields.
> 
> 2. What is the reason to store node ip? If you want to check it against the peer that is joining to the node it may be not possible. IP address can be easily spoofed using many headers. If you setup a floating ip on digital ocean (to have a static ip to your server), the DO by default route the public traffic through a different IP address than the floating ip. I'm trying to say that there can be many misalignments in the IP specified in the file and the IP node is joining from. What's more, having the seedlist centralized like it is right now will effectively block migrating a node to a new server (for example after credentials breach). We had an IP specified in the seedlist on mainnet 1.0 and due to those issues, we decided to remove it from the seedlist and not rely on it.

The seedlist must specify the `PeerId`. All of the other values can be left blank, missing, or set. However, the seedlist must have either 1 field or 4 fields in a row. The number of fields can vary per line, but they must always be 1 or 4.

modules/sdk/src/main/scala/org/tessellation/sdk/app/TessellationIOApp.scala
```diff
@@ -104,7 +104,7 @@ abstract class TessellationIOApp[A <: CliMethod](
                           _generation <- Generation.make[IO].asResource
                           versionHash <- version.hash.liftTo[IO].asResource
                           _seedlist <- method.seedlistPath
-                            .fold(none[Set[PeerId]].pure[IO])(SeedlistLoader.make[IO].load(_).map(_.some))
+                            .fold(none[Set[types.SeedlistCSVEntry]].pure[IO])(SeedlistLoader.make[IO].load(_).map(_.some))
```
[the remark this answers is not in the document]
cngo-github: This change has been made.

marcinwadon:
@cngo-github okay, if so, then using two case classes makes sense to have either a single entry (peer id) or four entries (all values)

modules/sdk/src/main/scala/org/tessellation/sdk/app/TessellationIOApp.scala
```diff
@@ -132,7 +135,7 @@ abstract class TessellationIOApp[A <: CliMethod](
                               session,
                               p2pClient.node,
                               validators,
-                              _seedlist,
+                              _seedlist.map(_.map(_.id)),
```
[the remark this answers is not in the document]
marcinwadon: Can you change the class name to something that doesn't leak the implementation details? Right now you pass CSV entry into the domain, but the seedlist can be loaded from any place (for example json file). The rest of the code should not depend that it's a CSV.

cngo-github:
> @cngo-github okay, if so, then using two case classes makes sense to have either a single entry (peer id) or four entries (all values)

The decoder now decodes an `Either[PeerId, SeedlistEntry]`. If there is only 1 peer, a `PeerId` is returned as a `Left`. If there are 4 fields, a `SeedlistEntry` is created with those 4 fields. `ipAddress`, `alias`, and `bias` will be kept as `Option`-wrapped values at @ryle-goehausen 's request, and `None` will be used when the field has an empty value.

marcinwadon:
@ryle-goehausen can you elaborate on keeping `Option` types when all fields are present?
We talked on slack with @cngo-github and @sdavidp that using `Either[PeerId, SeedlistEntry]` we would benefit over maintaining all optional types in the code. If the `Option` type would still be there, there is no need to distinguish it between `PeerId` or `SeedlistEntry` so it can be as it was before (a single case class with optional types) as there is no benefit of `Either` at all.

modules/sdk/src/main/scala/org/tessellation/sdk/app/TessellationIOApp.scala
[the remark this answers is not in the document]
cngo-github: `SeedlistCSVEntry` has been renamed `SeedlistEntry`. All intermediate objects use `Option[Set[Either[PeerId, SeedlistEntry]]]` or `Set[Either[PeerId, SeedlistEntry]]`, which are converted to `Option[Set[PeerId]]` and `Set[PeerId]`, respectively, at the end.

cngo-github:
> @ryle-goehausen can you elaborate on keeping `Option` types when all fields are present? We talked on slack with @cngo-github and @sdavidp that using `Either[PeerId, SeedlistEntry]` we would benefit over maintaining all optional types in the code. If the `Option` type would still be there, there is no need to distinguish it between `PeerId` or `SeedlistEntry` so it can be as it was before (a single case class with optional types) as there is no benefit of `Either` at all.

I want to clarify. A field is present when there is a comma. For example, `AABBCC,127.0.0.1,,1` has 4 fields fields: `peerId`, `ipAddress`, `alias`, and `bias`. All fields except for `alias` have a non-empty value. Since there are 4 fields, this row would be decoded as a `SeedlistEntry`. If there were no `Option`s for the fields in `SeedlistEntry`, you would get:

```
SeedlistEntry(
peerId = PeerId(Hex("AABBCC")),
ipAddress = "127.0.0.1",
alias = "",
bias = TrustValueRefined(1.0)
)
```

currently, you would get

```
SeedlistEntry(
peerId = PeerId(Hex("AABBCC")),
ipAddress = Some("127.0.0.1"),
alias = None[String],
bias = Some(TrustValueRefined(1.0))
)
```

@ryle-goehausen  indicated he would rather see `None` when there are empty fields instead of a default value like `""`.

ryle-goehausen:
@marcinwadon 

>@ryle-goehausen can you elaborate on keeping Option types when all fields are present?
We talked on slack with @cngo-github and @sdavidp that using Either[PeerId, SeedlistEntry] we would benefit over maintaining all optional types in the code. If the Option type would still be there, there is no need to distinguish it between PeerId or SeedlistEntry so it can be as it was before (a single case class with optional types) as there is no benefit of Either at all.

In general, there is probably much more information we want to add to this in the future, so new fields should be treated as optional so it doesn't force us to update prior values / prior data.

It makes to support the case where some new fields are added, and all fields are not present. So we shouldn't expect the new format to necessarily supply all rows. Someone may want to update their existing peerId without supplying a static ip address / host, so optional. Additionally the score should have a default (requiring it not to be supplied,) and only if present should it be overridden. It's useful to distinguish these cases.

cngo-github:
> @marcinwadon
> 
> > @ryle-goehausen can you elaborate on keeping Option types when all fields are present?
> > We talked on slack with @cngo-github and @sdavidp that using Either[PeerId, SeedlistEntry] we would benefit over maintaining all optional types in the code. If the Option type would still be there, there is no need to distinguish it between PeerId or SeedlistEntry so it can be as it was before (a single case class with optional types) as there is no benefit of Either at all.
> 
> In general, there is probably much more information we want to add to this in the future, so new fields should be treated as optional so it doesn't force us to update prior values / prior data.
> 
> It makes to support the case where some new fields are added, and all fields are not present. So we shouldn't expect the new format to necessarily supply all rows. Someone may want to update their existing peerId without supplying a static ip address / host, so optional. Additionally the score should have a default (requiring it not to be supplied,) and only if present should it be overridden. It's useful to distinguish these cases.

@marcinwadon , do I proceed with reverting the decoder to decoding `SeedlistEntry` instead of `Either[PeerId, SeedlistEntry]`?

marcinwadon:
> @marcinwadon
> 
> > @ryle-goehausen can you elaborate on keeping Option types when all fields are present?
> > We talked on slack with @cngo-github and @sdavidp that using Either[PeerId, SeedlistEntry] we would benefit over maintaining all optional types in the code. If the Option type would still be there, there is no need to distinguish it between PeerId or SeedlistEntry so it can be as it was before (a single case class with optional types) as there is no benefit of Either at all.
> 
> In general, there is probably much more information we want to add to this in the future, so new fields should be treated as optional so it doesn't force us to update prior values / prior data.
> 
> It makes to support the case where some new fields are added, and all fields are not present. So we shouldn't expect the new format to necessarily supply all rows. Someone may want to update their existing peerId without supplying a static ip address / host, so optional. Additionally the score should have a default (requiring it not to be supplied,) and only if present should it be overridden. It's useful to distinguish these cases.

Got it. It means that @cngo-github there should be a single case class with optional fields instead of `Either` but it also means it should be a separate thing that is being loaded from disk or the check for seedlist alignment should be done against peer ids only. Otherwise nodes will not be able to join the network.

cngo-github:
> > @marcinwadon
> > > @ryle-goehausen can you elaborate on keeping Option types when all fields are present?
> > > We talked on slack with @cngo-github and @sdavidp that using Either[PeerId, SeedlistEntry] we would benefit over maintaining all optional types in the code. If the Option type would still be there, there is no need to distinguish it between PeerId or SeedlistEntry so it can be as it was before (a single case class with optional types) as there is no benefit of Either at all.
> > 
> > 
> > In general, there is probably much more information we want to add to this in the future, so new fields should be treated as optional so it doesn't force us to update prior values / prior data.
> > It makes to support the case where some new fields are added, and all fields are not present. So we shouldn't expect the new format to necessarily supply all rows. Someone may want to update their existing peerId without supplying a static ip address / host, so optional. Additionally the score should have a default (requiring it not to be supplied,) and only if present should it be overridden. It's useful to distinguish these cases.
> 
> Got it. It means that @cngo-github there should be a single case class with optional fields instead of `Either` but it also means it should be a separate thing that is being loaded from disk or the check for seedlist alignment should be done against peer ids only. Otherwise nodes will not be able to join the network.

Please elucidate this: `should be a separate thing that is being loaded from disk or the check for seedlist alignment should be done against peer ids only`. Just map the `SeedlistEntry` to `PeerId` at the end?

cngo-github:
> > @marcinwadon
> > > @ryle-goehausen can you elaborate on keeping Option types when all fields are present?
> > > We talked on slack with @cngo-github and @sdavidp that using Either[PeerId, SeedlistEntry] we would benefit over maintaining all optional types in the code. If the Option type would still be there, there is no need to distinguish it between PeerId or SeedlistEntry so it can be as it was before (a single case class with optional types) as there is no benefit of Either at all.
> > 
> > 
> > In general, there is probably much more information we want to add to this in the future, so new fields should be treated as optional so it doesn't force us to update prior values / prior data.
> > It makes to support the case where some new fields are added, and all fields are not present. So we shouldn't expect the new format to necessarily supply all rows. Someone may want to update their existing peerId without supplying a static ip address / host, so optional. Additionally the score should have a default (requiring it not to be supplied,) and only if present should it be overridden. It's useful to distinguish these cases.
> 
> Got it. It means that @cngo-github there should be a single case class with optional fields instead of `Either` but it also means it should be a separate thing that is being loaded from disk or the check for seedlist alignment should be done against peer ids only. Otherwise nodes will not be able to join the network.

`Either[PeerId, SeedlistEntry]` has been replaced `SeedlistEntry`, and transformed into `PeerId` at the end.

marcinwadon:
> Please elucidate this: `should be a separate thing that is being loaded from disk or the check for seedlist alignment should be done against peer ids only`. Just map the `SeedlistEntry` to `PeerId` at the end?

I think you need to load the seedlist as it was before (as it's centralized) and all metadata from a separate file. If no, then the alignment check should be done against `PeerId` only.

modules/sdk/src/main/scala/org/tessellation/sdk/app/TessellationIOApp.scala
```diff
@@ -158,26 +164,26 @@ abstract class TessellationIOApp[A <: CliMethod](
...
-                            val sdkStorages = storages
-                            val sdkServices = services
-                            val sdkPrograms = programs
-                            val sdkValidators = validators
-
-                            def restartSignal = _restartSignal
+                            val random: Random[IO] = _random
+                            val securityProvider: SecurityProvider[IO] = _securityProvider
+                            val kryoPool: KryoSerializer[IO] = _kryoPool
+                            val metrics: Metrics[IO] = _metrics
```
marcinwadon, line 170: why? if so, can you make such changes in separate PR to separate it from the current change (seedlist)?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/seedlist/types.scala
```diff
@@ -1,20 +1,49 @@
...
+          Left(new DecoderError(s"Rows must have 1 or 4 fields, but found ${row.values.size}"))
+        }
+    }
   }
 
+  implicit val order: Order[SeedlistEntry] = Order[PeerId].contramap(_.peerId)
+  implicit val ordering: Ordering[SeedlistEntry] = order.toOrdering
+
+  // implicit val orderDecodedEntry: Order[Either[PeerId, SeedlistEntry]] = Order[PeerId].contramap(_.fold(identity, _.peerId))
+  // implicit val orderingDecodedEntry: Ordering[Either[PeerId, SeedlistEntry]] = orderDecodedEntry.toOrdering
```
marcinwadon, line 47: leftover

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/seedlist/types.scala
```diff
@@ -1,20 +1,49 @@
...
-import fs2.data.csv.RowDecoder
+import fs2.data.csv._
 import fs2.data.csv.generic.semiauto.deriveRowDecoder
 
 object types {
 
-  case class SeedlistCSVEntry(id: String) {
+  case class SeedlistEntry(
+    peerId: PeerId,
+    ipAddress: Option[String],
```
marcinwadon, line 17: you can use a dedicated type for IP address - `Host` or `IpAddress` from `com.comcast.ip4s`

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/seedlist/types.scala
```diff
@@ -1,20 +1,49 @@
...
+    alias: Option[String],
```
marcinwadon, line 18: it would be best to create a newtype for alias

modules/sdk/src/test/resources/seedlist.1field.sample
marcinwadon, line 1: I think you can put all these seedlists in a single separate directory, what do you think?

modules/sdk/src/main/scala/org/tessellation/sdk/app/TessellationIOApp.scala
```diff
@@ -63,7 +65,7 @@ abstract class TessellationIOApp[A <: CliMethod](
     */
   val kryoRegistrar: Map[Class[_], KryoRegistrationId[KryoRegistrationIdRange]]
 
-  protected val logger = Slf4jLogger.getLogger[IO]
+  protected val logger: SelfAwareStructuredLogger[IO] = Slf4jLogger.getLogger[IO]
```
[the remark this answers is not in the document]
cngo-github: IntelliJ pointed this out as a warning.

modules/sdk/src/main/scala/org/tessellation/sdk/app/TessellationIOApp.scala
cngo-github: IntelliJ pointed these out as warnings.

modules/sdk/src/main/scala/org/tessellation/sdk/app/TessellationIOApp.scala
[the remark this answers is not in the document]
marcinwadon: can you show this warning? I see no such thing during sbt compilation 🤔

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
```diff
@@ -245,7 +246,8 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
     seedlist match {
       case None => Applicative[F].unit
       case Some(entries) =>
-        if (entries.contains(peer.id)) Applicative[F].unit else PeerNotInSeedlist(peer.id).raiseError[F, Unit]
+        if (entries.map(_.peerId).contains(peer.id)) { Applicative[F].unit }
+        else { PeerNotInSeedlist(peer.id).raiseError[F, Unit] }
```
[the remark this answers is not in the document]
cngo-github: The brackets have been removed.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/seedlist/types.scala
cngo-github: The IP Address has been changed to `IpAddress`.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/seedlist/types.scala
cngo-github: `alias` is now the `Alias` type:

```
type Alias = String
```

modules/sdk/src/test/resources/seedlist.1field.sample
cngo-github: The seedlists have been moved to the sub-directory `seedlists`.

modules/sdk/src/main/scala/org/tessellation/sdk/app/TessellationIOApp.scala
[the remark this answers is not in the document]
cngo-github: Per our offline discussion, this change has been reverted.

modules/sdk/src/main/scala/org/tessellation/sdk/app/TessellationIOApp.scala
cngo-github: Per our offline discussion, these changes have been reverted.

marcinwadon:
@cngo-github please test the PR on the cluster as I believe it will fail with joining error for different seedlists present on the nodes

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/seedlist/types.scala
marcinwadon: `type Alias = String` is not a newtype and does not prevent human mistakes (you can assign any string there and it won't throw a compile error). please take a look how `PeerId` is created, you already used that construction.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/seedlist/SeedlistEntry.scala
```diff
@@ -0,0 +1,55 @@
...
+import org.tessellation.schema.trust._
+
+import com.comcast.ip4s.IpAddress
+import fs2.data.csv._
+import fs2.data.csv.generic.semiauto.deriveRowDecoder
+
+case class SeedlistEntry(
+  peerId: PeerId,
+  ipAddress: Option[IpAddress],
+  alias: Option[String],
```
kpudlik, line 17: Shouldn't it be `Option[Alias]`?

modules/sdk/src/main/scala/org/tessellation/sdk/domain/seedlist/SeedlistEntry.scala
```diff
@@ -0,0 +1,55 @@
...
+  bias: Option[TrustValueRefined]
+)
+
+object SeedlistEntry {
+
+  type Alias = String
```
kpudlik, line 23: You should probably use a `@newtype` there.

Type aliases differ from new types. Type alias as you did here is just a reference to `String`. Consider such example which leads to errors:

```scala
type Age = Long
type Speed = Long

val age: Age = 18
val speed: Speed = age // No compile error!
```

where with `@newtype`:

```scala
@newtype
case class Age(value: Long)
@newtype
case class Speed(value: Long)

val age: Age = Age(18)
val speed: Speed = age // Compile error because Age is not assiblable to Speed
```

kpudlik (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/domain/seedlist/SeedlistEntry.scala
cngo-github: `Alias` has been changed to a case class.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/seedlist/SeedlistEntry.scala
cngo-github: You are correct. This has been fixed.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/cluster/services/Cluster.scala
```diff
@@ -44,15 +45,19 @@ object Cluster {
         for {
           session <- sessionStorage.getToken.flatMap {
             case Some(s) => Applicative[F].pure(s)
-            case None    => MonadThrow[F].raiseError[SessionToken](SessionDoesNotExist)
+            case None =>
```
marcinwadon, line 48: Looks like you changed the format unexpectedly

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/cluster/services/Cluster.scala
```diff
@@ -44,15 +45,19 @@ object Cluster {
...
+              MonadThrow[F].raiseError[SessionToken](SessionDoesNotExist)
           }
           clusterSession <- clusterStorage.getToken.flatMap {
             case Some(s) => Applicative[F].pure(s)
-            case None    => MonadThrow[F].raiseError[ClusterSessionToken](ClusterSessionDoesNotExist)
+            case None =>
+              MonadThrow[F].raiseError[ClusterSessionToken] {
```
marcinwadon, line 54: Looks like you changed the format unexpectedly

cngo-github: Reverted this change.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/cluster/services/Cluster.scala
cngo-github: Reverted this change.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/seedlist.scala
```diff
@@ -0,0 +1,68 @@
...
+    peerId: PeerId,
+    ipAddress: Option[IpAddress],
+    alias: Option[Alias],
+    bias: Option[TrustValueRefined]
+  )
+
+  object SeedlistEntry {
+
+    implicit val maybeIpAddressCellDecoder: CellDecoder[Option[IpAddress]] =
+      CellDecoder.stringDecoder.map(IpAddress.fromString)
```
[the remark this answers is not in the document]
cngo-github: `IpAddress.fromString` returns an `Option[IpAddress]`. It only returns a `Some(...)` if the string is a valid IP. So all invalid IP strings, including an empty string, returns `None`.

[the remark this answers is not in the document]
marcinwadon: Yes, exactly that's why I think it's incorrect. Empty string is not an invalid value but no value (empty field) and you use None to represent an empty field, not an invalid field.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/seedlist.scala
```diff
@@ -0,0 +1,77 @@
...
+  object SeedlistEntry {
+
+    implicit val maybeIpAddressCellDecoder: CellDecoder[Option[IpAddress]] =
+      CellDecoder.stringDecoder.emap { value =>
+        if ("".equals(value))
+          none.asRight[DecoderError]
+        else
+          IpAddress.fromString(value) match {
+            case v @ Some(_) => v.asRight[DecoderError]
+            case None        => Left(new DecoderError("Invalid IP address."))
```
marcinwadon, line 48: Please add a value there as well if that's possible. Finding an invalid entry within hundreds may be hard

modules/sdk/src/main/scala/org/tessellation/sdk/domain/seedlist.scala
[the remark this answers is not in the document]
cngo-github: The decoder no distinguishes between an empty string and a non-empty string that is an invalid IP Address.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/seedlist.scala
cngo-github: Added the invalid value to the error's message.

marcinwadon (approved):

kpudlik (approved):

PR #668, opened 2023-06-15
Updating action to new Metagraph ID generation

IPadawans:


.github/workflows/pull-requests.yml
```diff
@@ -89,8 +89,8 @@ jobs:
           export CL_PUBLIC_HTTP_PORT=9000
           export CL_P2P_HTTP_PORT=9001
           export CL_CLI_HTTP_PORT=9002
-          export CL_APP_ENV=dev
-          export CL_ENV=dev
+          export CL_APP_ENV=testnet
+          export CL_ENV=testnet
```
marcinwadon, line 93: Why do you want to run the `testnet` environment on development env?

IPadawans: To create the log folder and put the logs inside that folder, instead of only logging on the bash.

marcinwadon: ok. so that should be done via `logback.xml` but still having the environment `dev`. we can manage that on slack. `testnet` and `mainnet` envs should be used on testnet/mainnet only (maybe with small exceptions)

IPadawans: @marcinwadon I think we can let as `dev` this way, for now, we need to merge this to fix the action to the next PRs. Can I merge?

marcinwadon (approved):

PR #670, opened 2023-06-20
Adding logs of the action as artifacts of the execution

IPadawans:


.github/workflows/pull-requests.yml
```diff
@@ -609,4 +609,32 @@ jobs:
         run: |
           cd .github/action_scripts
           npm install @stardust-collective/dag4
-          node send_transactions.js
\ No newline at end of file
+          node send_transactions.js
+
+      - name: Save Global L0 logs
+        uses: actions/upload-artifact@v3
+        if: always()
```
marcinwadon, line 616: is the condition required? doesn't `always()` work like no condition?

.github/workflows/pull-requests.yml
```diff
@@ -590,23 +589,232 @@ jobs:
...
           done
 
-      - name: Request clusters info
+      - name: Waiting 120s before check nodes health
         run: |
-          sleep 5
-          echo "Dag L1"
-          curl http://127.0.0.1:9100/cluster/info
+          echo "Waiting for 120s before checking cluster individual nodes health..."
+          sleep 120
```
marcinwadon, line 595: I think you can query the health in the loop with 120s timeout instead of fixed delay. That way we won't wait 120s if nodes are ready earlier.

.github/workflows/pull-requests.yml
```diff
@@ -590,23 +589,232 @@ jobs:
...
+            echo "Ordinals are the same:"
+            echo "CL0 - 1: $CL0_1_ORDINAL"
+            echo "CL0 - 2: $CL0_2_ORDINAL"
+            echo "CL0 - 3: $CL0_3_ORDINAL"
+          else
+            echo "Ordinals are not the same:"
+            echo "CL0 - 1: $CL0_1_ORDINAL"
+            echo "CL0 - 2: $CL0_2_ORDINAL"
+            echo "CL0 - 3: $CL0_3_ORDINAL"
+            exit 1
```
marcinwadon, line 731: ordinals may be not the same quite often, so this won't be a deterministic test.

.github/workflows/pull-requests.yml
```diff
@@ -590,23 +589,232 @@ jobs:
...
+          fi
+
+      - name: Check Currency L0 nodes snapshots
+        run: |
+          export CL0_1_SNAPSHOT=$(curl -s http://127.0.0.1:9400/snapshots/latest -H "Accept: application/json" | grep -Eo -m 1 "\"lastSnapshotHash\":\".*?\""  )
+          export CL0_2_SNAPSHOT=$(curl -s http://127.0.0.1:9500/snapshots/latest -H "Accept: application/json" | grep -Eo -m 1 "\"lastSnapshotHash\":\".*?\""  )
+          export CL0_3_SNAPSHOT=$(curl -s http://127.0.0.1:9600/snapshots/latest -H "Accept: application/json" | grep -Eo -m 1 "\"lastSnapshotHash\":\".*?\""  )
+
+          if [[ $CL0_1_ORDINAL == $CL0_2_ORDINAL && $CL0_1_ORDINAL == $CL0_3_ORDINAL ]]; then
```
marcinwadon, line 740: as above

.github/workflows/pull-requests.yml
IPadawans: Hmm, how to you recommend testing this? Ryle as said to catch the latest one

.github/workflows/pull-requests.yml
```diff
@@ -590,23 +589,144 @@ jobs:
...
-          echo "Currency L0"
-          curl http://127.0.0.1:9400/cluster/info
-          echo ""
-          
-          echo "Currency L1"
-          curl http://127.0.0.1:9700/cluster/info
-          echo ""
+          for ((i = 1; i <= 12; i++)); do
+            if curl -s http://127.0.0.1:9000/node/info | grep -o "\"state\":\"Ready\"" >> /dev/null; then
+              echo "Global L0 health and Ready"
```
TheMMaciek, line 596: *healthy

.github/workflows/pull-requests.yml
```diff
@@ -590,23 +589,144 @@ jobs:
...
+              exit 0 
+            else
+              if [ $i -eq 12 ]; then
+                echo "Global L0 is not health, check the logs"
```
TheMMaciek, line 600: *healthy

.github/workflows/pull-requests.yml
TheMMaciek: I would just say that they are Ready actually. We are not checking healthiness as a different kind of check from status being Ready, so it always will be both or none.

.github/workflows/pull-requests.yml
```diff
@@ -590,23 +589,144 @@ jobs:
...
+            export SMALLEST_ORDINAL=$CL0_3_ORDINAL
+          fi
+          
+          export BIGGEST_ORDINAL=$CL0_1_ORDINAL
+          if [[ $CL0_2_ORDINAL > $BIGGEST_ORDINAL ]]; then
+            export BIGGEST_ORDINAL=$CL0_2_ORDINAL
+          fi
+          if [[ $CL0_3_ORDINAL > $BIGGEST_ORDINAL ]]; then
+            export BIGGEST_ORDINAL=$CL0_3_ORDINAL
+          fi
```
TheMMaciek, line 676: Maybe you could do something like creating an array out of all the ordinals, sorting it numerically and taking first and last as min and max, to not have multiple if statements?

.github/workflows/pull-requests.yml
```diff
@@ -590,23 +589,144 @@ jobs:
...
+          echo "The biggest ordinal is $BIGGEST_ORDINAL"
+          
+          if [[ $(expr $BIGGEST_ORDINAL - $SMALLEST_ORDINAL)  > 3 ]]; then
+            echo "Difference between BIGGEST_ORDINAL: $BIGGEST_ORDINAL and SMALLEST_ORDINAL: $SMALLEST_ORDINAL is greater than 3"
+            exit 1
+          fi
+          
+          export CL0_1_SNAPSHOT=$(curl -s http://127.0.0.1:9400/snapshots/$SMALLEST_ORDINAL -H "Accept: application/json" | grep -Eo -m 1 "\"lastSnapshotHash\":\".*?\""  )
+          export CL0_2_SNAPSHOT=$(curl -s http://127.0.0.1:9500/snapshots/$SMALLEST_ORDINAL -H "Accept: application/json" | grep -Eo -m 1 "\"lastSnapshotHash\":\".*?\""  )
+          export CL0_3_SNAPSHOT=$(curl -s http://127.0.0.1:9600/snapshots/$SMALLEST_ORDINAL -H "Accept: application/json" | grep -Eo -m 1 "\"lastSnapshotHash\":\".*?\""  )
```
TheMMaciek, line 688: I guess you could use `jq` to fetch the fields you want instead of grep.

IPadawans: Thanks for the tip, that's a great tool =]

.github/workflows/pull-requests.yml
```diff
@@ -48,6 +52,7 @@ jobs:
 
       - name: Create directories
         run: |
+          jq --version
```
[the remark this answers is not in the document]
IPadawans: Yeah, I know haha, just debugging it, I'll clean everything yet

.github/workflows/pull-requests.yml
```diff
@@ -590,23 +595,138 @@ jobs:
...
+              exit 0 
+            else
+              if [ $i -eq 12 ]; then
+                echo "Global L0 is not ready, check the logs"
+                exit 1
+              fi
+              echo "Global L0 not ready yet, trying again in 10s"
+              sleep 10
+            fi
+          done
```
marcinwadon, line 612: there is a lot of repeating code. I'd suggest (either now or in some future PR) to extract the common logic (like execute a code with loop check and exit timeout) and reuse it here and there.

.github/workflows/pull-requests.yml
```diff
@@ -590,23 +595,138 @@ jobs:
...
+
+      - name: Check if DAG L1 nodes are Ready
+        run: |
+          for ((i = 1; i <= 12; i++)); do
+            if [[ $(curl -s http://127.0.0.1:9100/node/info | jq '.state=="Ready"') == true && $(curl -s http://127.0.0.1:9200/node/info | jq '.state=="Ready"') == true && $(curl -s http://127.0.0.1:9300/node/info | jq '.state=="Ready"') == true ]]; then
+              echo "All DAG L1 nodes are ready"
```
marcinwadon, line 618: here the same. you can extract the ready check to some bash script and execute it like:
```
./node-is-ready --port 9100 && ./node-is-ready --port 9200 ...
```

marcinwadon: or... the best would be to write typescript scripts and execute them only in here. bash scripts are fine to some extent but at some point for more complex checks ts is just easier (or python, go, whatever)

IPadawans: Done

.github/action_scripts/check_clusters.js
```diff
@@ -0,0 +1,218 @@
...
+        }s, check the logs.`
+    );
+};
+
+const validateOrdinalsAndSnapshots = async ( urls ) => {
+    const ordinals = [];
+    for( const url of urls ) {
+        const snapshot = await fetchData( `${url}/latest` );
+        ordinals.push( snapshot.value.ordinal );
+    }
```
marcinwadon, line 55: I don't know how much you're into functional programming but you can use that array with URLs and using `Promise.all` get responses that you can map then to snapshot ordinals like:

```
.then(_ => _.value.ordinal)
```

marcinwadon (dismissed):
fyi: you can extract the joining curl command as well

.github/action_scripts/check_clusters.js
IPadawans: Hmm, good catch. I know this functionality and maybe I've missed this part, I'll update =].

marcinwadon: in general, FP approach is that you avoid mutations and side effects.
```
const ordinals = []
...
ordinals.push(...)
```
is a mutation. in FP, you read the code as small programs and knowing that all structures are immutable, it's just easier for local reasoning and understanding the flow of the code. I'm not sure if you want to follow FP in the code but this is something we rely on in the protocol, so just instilling the idea :)

PR #672, opened 2023-06-21
Return BadRequest in case of malformed state channel snapshot binary

kpudlik:


marcinwadon (approved):

modules/core/src/main/scala/org/tessellation/http/routes/StateChannelRoutes.scala
```diff
@@ -26,7 +26,10 @@ final case class StateChannelRoutes[F[_]: Async](
         .as[Signed[StateChannelSnapshotBinary]]
         .map(StateChannelOutput(address, _))
         .flatMap(stateChannelService.process)
-        .flatMap(_ => Ok())
+        .flatMap {
+          case Left(_) => BadRequest()
+          case Right(_) => Ok()
```
TheMMaciek, line 31: I think it's good initially but we may have to return concrete errors to the client so that the client can handle them. I'm aware that there is a tradeoff between security and being concrete, but I think that for this endpoint being the client endpoint vs peer endpoint it makes sense to return concrete error messages.

TheMMaciek (approved):

modules/core/src/main/scala/org/tessellation/http/routes/StateChannelRoutes.scala
marcinwadon: @TheMMaciek does it make sense to cover that while refactoring api endpoints for common error structure?

TheMMaciek: Yeah, I think so, that was my intuition as well.

PR #673, opened 2023-06-22
Simplify/refactor ahead of a couple of PRs

sdavidp:
The affected code is being refactored in a couple of upcoming PRs (#666 and for PROT-244).
Putting this out now will make it a little easier to review those PRs.


modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
```diff
@@ -242,11 +242,10 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
...
   private def validateSeedlist(peer: PeerToJoin): F[Unit] =
-    seedlist match {
-      case None => Applicative[F].unit
-      case Some(entries) =>
-        if (entries.contains(peer.id)) Applicative[F].unit else PeerNotInSeedlist(peer.id).raiseError[F, Unit]
-    }
+    if (seedlist.forall(_.contains(peer.id)))
+      Applicative[F].unit
+    else
+      PeerNotInSeedlist(peer.id).raiseError[F, Unit]
```
marcinwadon, line 248: if you actually touch this part, you can use `whenA`/`unlessA` method from Applicative.

image

marcinwadon:
@sdavidp would you like to keep this pull request to aggregate all changes or should we merge them one by one?
if the latter is true, please change the commit name to something more descriptive what has been changed so we can already merge it as a small simple change (and refactor the code in small steps)

TheMMaciek (approved):
Nice. I would only think about moving `loadSeedlist` somewhere else to not pollute the `TessellationIOApp` if not necessary.

TheMMaciek (changes requested):
Unapproved so that the commit message gets adjusted.

TheMMaciek (approved):

PR #674, opened 2023-06-27
Fix forked snapshots cleanup during download

TheMMaciek:


sdavidp (commented):
LGTM

modules/core/src/main/scala/org/tessellation/domain/snapshot/storages/SnapshotDownloadStorage.scala
```diff
@@ -19,4 +19,5 @@ trait SnapshotDownloadStorage[F[_]] {
   def moveTmpToPersisted(snapshot: Signed[GlobalIncrementalSnapshot]): F[Unit]
 
   def readGenesis(ordinal: SnapshotOrdinal): F[Option[Signed[GlobalSnapshot]]]
+  def movePersistedAboveToTmp(ordinal: SnapshotOrdinal): F[Unit]
```
marcinwadon, line 22: does it make sense to name it more like `move persisted to tmp above ... ordinal` - so `movePersistedToTmpAbove(ordinal)`?

TheMMaciek: How I intended it to sound was `move persisted above (ordinal) to tmp` hence the name. In your version `above` sounds like it refers to tmp somehow 🤔 . Maybe: `movePersistedToTmpIfAbove` or `movePersistedAboveOrdinalToTmp`?

marcinwadon (approved):

PR #675, opened 2023-06-27
Endpoint for accepting custom data blocks on Currency L0

kpudlik:


marcinwadon (approved):

marcinwadon (approved):

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/http/routes/DataBlockRoutes.scala
```diff
@@ -0,0 +1,38 @@
...
+    extends Http4sDsl[F] {
+  import org.http4s.circe.CirceEntityCodec.circeEntityDecoder
+
+  private val httpRoutes: HttpRoutes[F] = HttpRoutes.of[F] {
+    case req @ POST -> Root / "l1-data-output" =>
+      req
+        .as[Signed[DataApplicationBlock]]
+        .map(_.asRight[Signed[CurrencyBlock]])
+        .map(mkCell)
+        .flatMap(_.run())
```
TheMMaciek, line 31: Does `run` never fail? Maybe we should at least return BadRequest in case some error occurs.

ryle-goehausen: It would be good to have some kind of error returned here

kpudlik: Good point. For DAG cell, we also do not check for errors so either it never fails of we missed something.

kpudlik: I included basic error handling

marcinwadon (approved):

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/p2p/L0CurrencyBlockOutputClient.scala
```diff
@@ -0,0 +1,33 @@
...
+import org.http4s.circe.CirceEntityCodec.circeEntityEncoder
+import org.http4s.client.Client
+
+trait L0CurrencyBlockOutputClient[F[_]] {
+  def sendL1Output(output: Signed[CurrencyBlock]): PeerResponse[F, Boolean]
+  def sendL1DataOutput(output: Signed[DataApplicationBlock]): PeerResponse[F, Boolean]
+}
+
+object L0CurrencyBlockOutputClient {
+  def make[F[_]](l0BlockOutputClient: L0BlockOutputClient[F, CurrencyBlock], client: Client[F])(
```
TheMMaciek, line 21: Does it make sense to aggregate client like this and have proxy methods?

PR #677, opened 2023-06-28
New rewards minting logic starting July first

TheMMaciek:


modules/core/src/main/scala/org/tessellation/config/types.scala
```diff
@@ -54,23 +57,51 @@ object types {
...
+
+  object RewardsConfig {
+    val stardustPrimary: Address = Address("DAGSTARDUSTCOLLECTIVEHZOIPHXZUBFGNXWJETZVSPAPAHMLXS")
+    val stardustSecondary: Address = Address("DAG8VT7bxjs1XXBAzJGYJDaeyNxuThikHeUTp9XY")
+    val softStaking: Address = Address("DAG77VVVRvdZiYxZ2hCtkHz68h85ApT5b2xzdTkn")
+    val testnet: Address = Address("DAG0qE5tkz6cMUD5M2dkqgfV4TQCzUUdAP5MFM9P")
+    val dataPool: Address = Address("DAG3RXBWBJq1Bf38rawASakLHKYMbRhsDckaGvGu")
+
+    val mainnetProgramsDistributionConfig: EpochProgress => ProgramsDistributionConfig = {
+      case epoch if epoch < EpochProgress(1336392L) =>
```
TheMMaciek, line 77: 1336392 is more or less UTC 12:00 am on July 1 2023. Please verify.

ryle-goehausen (approved):
This looks good to me, it might be useful here to describe the steps for replicating the epoch progress calculation, even in just a PR comment, i.e. describe current, estimation function, estimation of time passed, to make this easily verifiable to others.

marcinwadon:
> This looks good to me, it might be useful here to describe the steps for replicating the epoch progress calculation, even in just a PR comment, i.e. describe current, estimation function, estimation of time passed, to make this easily verifiable to others.

Assuming rate 1 snapshot/min, calculate the time difference between now and the estimated date in minutes and add that to the current epoch progress (returned by the `GET /global-snapshots/latest` endpoint on mainnet nodes).

kpudlik (approved):

modules/core/src/main/scala/org/tessellation/config/types.scala
TheMMaciek: Now that I calculate it it's more like 10am UTC, so 2 hours earlier. I may have calculated the time for UTC+2 by mistake. I will monitor it and adjust if it will be the case that it wasn't my mistake during calculation and the consensus rushes a bit.

PR #679, opened 2023-06-29
Data application POST /data endpoint and ApplicationError

marcinwadon:


modules/currency-l1/src/main/scala/org/tessellation/currency/l1/Main.scala
```diff
@@ -120,5 +120,5 @@ object Main
       ClusterId(UUID.fromString("517c3a05-9219-471b-a54c-21b7d72f4ae5")),
       version = BuildInfo.version
     ) {
-  override def dataApplication: Option[BaseDataApplicationL1Service[IO]] = BaseDataApplicationL1Service(Acme.energy).some
+  override def dataApplication: Option[BaseDataApplicationL1Service[IO]] = None // BaseDataApplicationL1Service(Acme.energy).some
```
marcinwadon, line 123: this is an intended change to make PR pass the Currency L1 checks. It will be removed when merging `custom-data-l1` branch to the `develop`.

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/Routes.scala
```diff
@@ -1,29 +1,61 @@
...
 import org.typelevel.log4cats.slf4j.Slf4jLogger
+import shapeless._
+import shapeless.syntax.singleton._
 
-final case class Routes[F[_]: Async](
+final case class Routes[F[_]: Async: KryoSerializer: SecurityProvider](
   dataApplicationPeerConsensusInput: Queue[F, Signed[ConsensusInput.PeerConsensusInput]],
   l0ClusterStorage: L0ClusterStorage[F],
-  dataDecoder: Decoder[DataUpdate]
+  dataDecoder: Decoder[DataUpdate],
```
kpudlik, line 29: You can make it implicit and get rid of line `39`. We rarely pass decoders explicitly.

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/Routes.scala
```diff
@@ -1,29 +1,61 @@
...
 import org.http4s.HttpRoutes
-import org.http4s.circe.CirceEntityCodec.circeEntityDecoder
+import org.http4s.circe.CirceEntityCodec.{circeEntityDecoder, circeEntityEncoder}
 import org.http4s.dsl.Http4sDsl
 import org.typelevel.log4cats.slf4j.Slf4jLogger
+import shapeless._
+import shapeless.syntax.singleton._
 
-final case class Routes[F[_]: Async](
+final case class Routes[F[_]: Async: KryoSerializer: SecurityProvider](
```
kpudlik, line 26: Imo the name is too generic. How about `DataApplicationRoutes`?

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/modules/HttpApi.scala
```diff
@@ -110,7 +109,7 @@ sealed abstract class HttpApi[
           nodeRoutes.publicRoutes <+>
           metricRoutes <+>
           targetRoutes <+>
-          routes.as(HttpRoutes.empty[F]).getOrElse(currencyRoutes.publicRoutes)
+          routes.map(_.publicRoutes).getOrElse(currencyRoutes.publicRoutes)
```
kpudlik, line 112: So either data routes or currency routes. Is that correct?

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/modules/HttpApi.scala
```diff
@@ -89,7 +88,7 @@ sealed abstract class HttpApi[
   private val registrationRoutes = RegistrationRoutes[F](services.cluster)
   private val gossipRoutes = GossipRoutes[F](storages.rumor, services.gossip)
   private val routes = maybeDataApplication.map { da =>
```
kpudlik, line 90: `implicit da`

kpudlik (changes requested):

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/Routes.scala
marcinwadon: fine for me, I'll change that (but I used the same pattern we already have in DAG L1)

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/Routes.scala
marcinwadon: I don't want to pass it implicitly as anyway I need to pass it explicitly to the next decoder. I know we pass all decoders/encoders implicitly, but here the decoder is provided by the client (end-user), hence it's passed explicitly as an argument.

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/modules/HttpApi.scala
marcinwadon: yes, that's intended. I have confirmed that with Alex and Ryle yesterday.

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/modules/HttpApi.scala
marcinwadon: why?

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/Routes.scala
marcinwadon: anyway, I see that I pass also the data application service, so I can extract it from there.

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/Routes.scala
```diff
@@ -1,29 +1,61 @@
...
   def logger = Slf4jLogger.getLogger[F]
 
+  private val public: HttpRoutes[F] = HttpRoutes.of[F] {
+    case req @ POST -> Root / "data" =>
+      implicit val decoder = dataDecoder
+
+      req
+        .asR[Signed[DataUpdate]] {
+          _.toHashedWithSignatureCheck[F](dataApplication.serializeUpdate _).flatMap {
+            case Left(e) => InternalServerError(InvalidSignature.toApplicationError)
```
sdavidp, line 44:
```suggestion
            case Left(_) => InternalServerError(InvalidSignature.toApplicationError)
```

sdavidp (approved):
LGTM. Left small suggestion.

sdavidp (approved):

kpudlik (approved):

PR #680, opened 2023-06-30
Splitting the Github action on 2: DAG and Currency

IPadawans:


marcinwadon:
@IPadawans by splitting actions I meant one action to be compilation and scala tests (exactly as it was before) and the second action your tests on a running cluster. That way we can simply require the code to be compilable (with passing unit tests) separately from the long running tests on the cluster.

marcinwadon:
btw - missing label

marcinwadon:
hey @IPadawans, will you work on that or it's already deprecated?

IPadawans:
> hey @IPadawans, will you work on that or it's already deprecated?

I'll work on this, I didn't work yet because we have the high priority things with metagraphs

PR #681, opened 2023-06-30
Adjust ProgramsDistributionConfig for future epoch

sdavidp:
UPDATE 2: The only effective change for this PR will be to introduce a new rewards distribution. The majority of the other changes described below will be in a follow-up release.

We're also adding the _Stardust Primary_ address to the set of locked addresses.

UPDATE: As per Ryle's comment we recalculated the expected Epoch to be `1352274` rather than using the computation below.

Calculation for the added `EpochProgress` is based on the recently added epoch progress value of `1349352` estimated to happen on July 1st, 10am UTC. Using the _1 snapshopt / minute_ guideline,
```
scala> val July1st = Instant.parse("2023-07-01T10:00:00Z")
val July1st: java.time.Instant = 2023-07-01T10:00:00Z

scala> val July10th = Instant.parse("2023-07-10T10:00:00Z")
val July10th: java.time.Instant = 2023-07-10T10:00:00Z

scala> ChronoUnit.MINUTES.between(July1st, July10th)
val res0: Long = 12960

scala> 1336392L + 22000 // We increased to 22,000 because EpochProgress seems to be increasing faster than expected
val res1: Long = 1358392
```

### DIFF
--- a/modules/core/src/main/scala/org/tessellation/config/types.scala
+++ b/modules/core/src/main/scala/org/tessellation/config/types.scala
@@ -9,8 +9,10 @@ import org.tessellation.config.types.RewardsConfig._
 import org.tessellation.schema.address.Address
 import org.tessellation.schema.balance.Amount
 import org.tessellation.schema.epoch.EpochProgress
+import org.tessellation.schema.transaction.TransactionAmount
 import org.tessellation.sdk.config.AppEnvironment
 import org.tessellation.sdk.config.types._
+import org.tessellation.sdk.domain.transaction.TransactionValidator.stardustPrimary
 
 import ciris.Secret
 import eu.timepit.refined.auto._
@@ -61,17 +63,21 @@ object types {
     remainingWeight: Weight
   )
 
+  case class OneTimeReward(epoch: EpochProgress, address: Address, amount: TransactionAmount)
+
   case class RewardsConfig(
     programs: EpochProgress => ProgramsDistributionConfig = mainnetProgramsDistributionConfig,
-    rewardsPerEpoch: SortedMap[EpochProgress, Amount] = mainnetRewardsPerEpoch
+    rewardsPerEpoch: SortedMap[EpochProgress, Amount] = mainnetRewardsPerEpoch,
+    oneTimeRewards: List[OneTimeReward] = List.empty
   )
 
   object RewardsConfig {
-    val stardustPrimary: Address = Address("DAGSTARDUSTCOLLECTIVEHZOIPHXZUBFGNXWJETZVSPAPAHMLXS")
+    val stardustNewPrimary = Address("DAG8vD8BUhCpTnYXEadQVGhHjgxEZZiafbzwmKKh")
     val stardustSecondary: Address = Address("DAG8VT7bxjs1XXBAzJGYJDaeyNxuThikHeUTp9XY")
     val softStaking: Address = Address("DAG77VVVRvdZiYxZ2hCtkHz68h85ApT5b2xzdTkn")
     val testnet: Address = Address("DAG0qE5tkz6cMUD5M2dkqgfV4TQCzUUdAP5MFM9P")
     val dataPool: Address = Address("DAG3RXBWBJq1Bf38rawASakLHKYMbRhsDckaGvGu")
+    val integrationNet: Address = Address("DAG8jE4CHy9T2izWFEv8K6rp5hNJq11SyLEVYnt8")
 
     val mainnetProgramsDistributionConfig: EpochProgress => ProgramsDistributionConfig = {
       case epoch if epoch < EpochProgress(1336392L) =>
@@ -85,7 +91,7 @@ object types {
           ),
           remainingWeight = Weight(4L) // facilitators
         )
-      case _ =>
+      case epoch if epoch < EpochProgress(1352274L) =>
         ProgramsDistributionConfig(
           weights = Map(
             stardustPrimary -> Weight(5L),
@@ -94,6 +100,17 @@ object types {
             dataPool -> Weight(55L)
           ),
           remainingWeight = Weight(30L) // facilitators
+        )
+      case _ =>
+        ProgramsDistributionConfig(
+          weights = Map(
+            stardustNewPrimary -> Weight(5L),
+            stardustSecondary -> Weight(5L),
+            testnet -> Weight(3L),
+            dataPool -> Weight(55L),
+            integrationNet -> Weight(15L)
+          ),
+          remainingWeight = Weight(17L) // facilitators
         )
     }
 
--- a/modules/core/src/main/scala/org/tessellation/infrastructure/dag/DAGService.scala
+++ b/modules/core/src/main/scala/org/tessellation/infrastructure/dag/DAGService.scala
@@ -8,6 +8,7 @@ import org.tessellation.domain.snapshot.SnapshotStorage
 import org.tessellation.schema.address.Address
 import org.tessellation.schema.balance.Balance
 import org.tessellation.schema.{GlobalSnapshot, SnapshotOrdinal}
+import org.tessellation.sdk.domain.transaction.TransactionValidator.lockedAddresses
 import org.tessellation.security.signature.Signed
 
 import io.estatico.newtype.ops._
@@ -16,7 +17,6 @@ object DAGService {
 
   def make[F[_]: Async](globalSnapshotStorage: SnapshotStorage[F, GlobalSnapshot]): DAGService[F] =
     new DAGService[F] {
-
       private def getBalance(snapshot: Signed[GlobalSnapshot], address: Address): (Balance, SnapshotOrdinal) = {
         val balance = snapshot.value.info.balances.getOrElse(address, Balance.empty)
         val ordinal = snapshot.value.ordinal
@@ -34,25 +34,31 @@ object DAGService {
           _.map(getBalance(_, address))
         }
 
-      private def getTotalSupply(snapshot: Signed[GlobalSnapshot]): (BigInt, SnapshotOrdinal) = {
+      private def calculateTotalSupply(balances: Iterable[Balance], ordinal: SnapshotOrdinal): (BigInt, SnapshotOrdinal) = {
         val empty = BigInt(Balance.empty.coerce.value)
-        val supply = snapshot.value.info.balances.values
+        val supply = balances
           .foldLeft(empty) { (acc, b) =>
             acc + BigInt(b.coerce.value)
           }
-        val ordinal = snapshot.value.ordinal
 
         (supply, ordinal)
       }
 
       def getTotalSupply: F[Option[(BigInt, SnapshotOrdinal)]] =
         globalSnapshotStorage.head.map {
-          _.map(getTotalSupply)
+          _.map(s => calculateTotalSupply(s.info.balances.values, s.ordinal))
+        }
+
+      def getFilteredOutTotalSupply: F[Option[(BigInt, SnapshotOrdinal)]] =
+        globalSnapshotStorage.head.map { maybeSignedSnapshot =>
+          maybeSignedSnapshot.map { snapshot =>
+            calculateTotalSupply(snapshot.info.balances.filterNot { case (a, _) => lockedAddresses.contains(a) }.values, snapshot.ordinal)
+          }
         }
 
       def getTotalSupply(ordinal: SnapshotOrdinal): F[Option[(BigInt, SnapshotOrdinal)]] =
         globalSnapshotStorage.get(ordinal).map {
-          _.map(getTotalSupply)
+          _.map(s => calculateTotalSupply(s.info.balances.values, s.ordinal))
         }
 
       private def getWalletCount(snapshot: Signed[GlobalSnapshot]): (Int, SnapshotOrdinal) = {
--- a/modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
+++ b/modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
@@ -9,7 +9,6 @@ import cats.syntax.applicativeError._
 import cats.syntax.either._
 import cats.syntax.eq._
 import cats.syntax.flatMap._
-import cats.syntax.foldable._
 import cats.syntax.functor._
 import cats.syntax.show._
 
@@ -74,10 +73,16 @@ object Rewards {
 
           val mintedAmount = getAmountByEpoch(epochProgress, config.rewardsPerEpoch)
 
-          allRewardsState
+          val rewards: F[SortedSet[RewardTransaction]] = allRewardsState
             .run(mintedAmount)
             .flatTap(validateState(mintedAmount))
             .map(toTransactions)
+
+          val oneTimeRewards = config.oneTimeRewards
+            .filter(_.epoch === epochProgress)
+            .map(otr => RewardTransaction(otr.address, otr.amount))
+
+          rewards.map(_ ++ oneTimeRewards)
         }
 
       private def validateState(totalPool: Amount)(state: (Amount, List[(Address, Amount)])): F[Unit] = {
--- a/modules/core/src/test/scala/org/tessellation/trust/SybilAttackSuite.scala
+++ b/modules/core/src/test/scala/org/tessellation/trust/SybilAttackSuite.scala
@@ -37,7 +37,6 @@ object SybilAttackSuite extends SimpleIOSuite with Checkers {
 
   def benchmark(allData: List[TrustNode], filterGood: Int => Boolean): ModelBenchmark = {
 
-    val numNodes = allData.size / 2
     val scores = SelfAvoidingWalk.runWalkFeedbackUpdateSingleNode(
       0,
       allData
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/domain/transaction/TransactionValidator.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/domain/transaction/TransactionValidator.scala
@@ -22,17 +22,16 @@ trait TransactionValidator[F[_], T <: Transaction] {
 }
 
 object TransactionValidator {
+  val stardustPrimary: Address = Address("DAGSTARDUSTCOLLECTIVEHZOIPHXZUBFGNXWJETZVSPAPAHMLXS")
+  val lockedAddresses: Set[Address] = Set(
+    Address("DAG0qgcEbMk8vQL6VrnbhMreNeEFXk12v1BvERCb"),
+    Address("DAG2KQrN97LpA5gRerJAQ5mDuy6kjC2dDtMr58fe")
+  )
 
   def make[F[_]: Async, T <: Transaction](
     signedValidator: SignedValidator[F]
   ): TransactionValidator[F, T] =
     new TransactionValidator[F, T] {
-
-      val lockedAddresses: Set[Address] = Set(
-        Address("DAG0qgcEbMk8vQL6VrnbhMreNeEFXk12v1BvERCb"),
-        Address("DAG2KQrN97LpA5gRerJAQ5mDuy6kjC2dDtMr58fe")
-      )
-
       def validate(
         signedTransaction: Signed[T]
       ): F[TransactionValidationErrorOr[Signed[T]]] =
--- a/modules/core/src/main/scala/org/tessellation/domain/dag/DagService.scala
+++ b/modules/core/src/main/scala/org/tessellation/domain/dag/DagService.scala
@@ -9,6 +9,7 @@ trait DAGService[F[_]] {
   def getBalance(ordinal: SnapshotOrdinal, address: Address): F[Option[(Balance, SnapshotOrdinal)]]
 
   def getTotalSupply: F[Option[(BigInt, SnapshotOrdinal)]]
+  def getFilteredOutTotalSupply: F[Option[(BigInt, SnapshotOrdinal)]]
   def getTotalSupply(ordinal: SnapshotOrdinal): F[Option[(BigInt, SnapshotOrdinal)]]
 
   def getWalletCount: F[Option[(Int, SnapshotOrdinal)]]
--- a/modules/core/src/test/scala/org/tessellation/infrastructure/rewards/RewardsSuite.scala
+++ b/modules/core/src/test/scala/org/tessellation/infrastructure/rewards/RewardsSuite.scala
@@ -12,6 +12,7 @@ import scala.collection.immutable.{SortedMap, SortedSet}
 import org.tessellation.config.types.RewardsConfig._
 import org.tessellation.config.types._
 import org.tessellation.domain.rewards.Rewards
+import org.tessellation.ext.cats.syntax.next.catsSyntaxNext
 import org.tessellation.ext.kryo._
 import org.tessellation.keytool.KeyPairGenerator
 import org.tessellation.kryo.KryoSerializer
@@ -21,6 +22,7 @@ import org.tessellation.schema.balance.Amount
 import org.tessellation.schema.epoch.EpochProgress
 import org.tessellation.schema.generators.{chooseNumRefined, transactionGen}
 import org.tessellation.schema.transaction.{RewardTransaction, TransactionAmount, TransactionFee}
+import org.tessellation.sdk.domain.transaction.TransactionValidator.stardustPrimary
 import org.tessellation.sdk.sdkKryoRegistrar
 import org.tessellation.security.SecurityProvider
 import org.tessellation.security.key.ops.PublicKeyOps
@@ -181,7 +183,7 @@ object RewardsSuite extends MutableIOSuite with Checkers {
     } yield expect.eql(expected, txs)
   }
 
-  test("minted rewards for the logic at epoch progress 1336392 and later are as expected") { res =>
+  test("minted rewards for the logic at epoch progress 1336392 until 1352274 are as expected") { res =>
     implicit val (_, sp, makeIdFn) = res
 
     val epoch = EpochProgress(1336392L)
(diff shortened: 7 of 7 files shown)


modules/core/src/main/scala/org/tessellation/config/types.scala
```diff
@@ -67,8 +67,8 @@ object types {
   )
 
   object RewardsConfig {
-    val stardustPrimary: Address = Address("DAGSTARDUSTCOLLECTIVEHZOIPHXZUBFGNXWJETZVSPAPAHMLXS")
-    val stardustSecondary: Address = Address("DAG8VT7bxjs1XXBAzJGYJDaeyNxuThikHeUTp9XY")
```
ryle-goehausen, line 71: I'd leave this address in place since it's not being touched / changed and not used in the same way as the others.

marcinwadon:
What's the reasoning behind this change? The stardust primary address is already locked without this change.

modules/sdk/src/main/scala/org/tessellation/sdk/config/FixedAddresses.scala
```diff
@@ -0,0 +1,17 @@
+package org.tessellation.sdk.config
+
+import org.tessellation.schema.address.Address
+
+import eu.timepit.refined.auto._
+
+object FixedAddresses {
```
ryle-goehausen, line 7: This is fine to separate in a different file, but it should still carry the named LockedAddresses I think, and stardustSecondary should be left in the other configuration location since that is remaining untouched.

modules/core/src/main/scala/org/tessellation/infrastructure/dag/DAGService.scala
```diff
@@ -34,9 +35,10 @@ object DAGService {
           _.map(getBalance(_, address))
         }
 
+      // TODO This needs tests; I can break it with the wrong `filter` and no tests fail
       private def getTotalSupply(snapshot: Signed[GlobalSnapshot]): (BigInt, SnapshotOrdinal) = {
         val empty = BigInt(Balance.empty.coerce.value)
-        val supply = snapshot.value.info.balances.values
+        val supply = snapshot.value.info.balances.filterNot { case (a, _) => lockedAddresses.contains(a) }.values
```
marcinwadon, line 41: fyi - it just filters out the balance for total supply endpoint but it does not change the snapshot which in fact will have an increased total supply. IMO the change should be reflected in the snapshot itself to keep the total supply at the same level, not just on the visual layer

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -74,10 +73,19 @@ object Rewards {
...
             .run(mintedAmount)
             .flatTap(validateState(mintedAmount))
             .map(toTransactions)
+
+          val oneTimeRewards = config.oneTimeRewards
+            .filter(_.epoch === epochProgress)
+            .map(otr => RewardTransaction(otr.address, otr.txnAmount))
+
+          // TODO should we handle case where Address already in rewards map and sum TransactionAmounts?
+          rewards.map(_.concat(oneTimeRewards))
```
marcinwadon, line 86: we don't need to sum tx amounts. it's fine to keep it as a separate reward transaction

modules/sdk/src/main/scala/org/tessellation/sdk/config/LockedAddresses.scala
```diff
@@ -0,0 +1,16 @@
+package org.tessellation.sdk.config
+
+import org.tessellation.schema.address.Address
+
+import eu.timepit.refined.auto._
+
+object LockedAddresses {
+  val stardustPrimary = Address("DAG8vD8BUhCpTnYXEadQVGhHjgxEZZiafbzwmKKh")
```
marcinwadon, line 8: it shouldn't be there, right?

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -74,10 +73,19 @@ object Rewards {
...
+
+        // TODO Add test if this was correct place to add one-time rewards
```
marcinwadon, line 88: I'm not sure how it will affect unit tests. I believe we already check the total sum of rewards if it matches the expected value

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
sdavidp: @marcinwadon If we keep it separate then that requires a change in the return type doesn't it? And that will have a ripple effect at the caller.

modules/core/src/main/scala/org/tessellation/config/types.scala
```diff
@@ -95,6 +101,17 @@ object types {
...
+        ProgramsDistributionConfig(
+          weights = Map(
+            stardustNewPrimary -> Weight(5L),
+            stardustSecondary -> Weight(5L),
+            testnet -> Weight(3L),
+            dataPool -> Weight(55L),
+            integrationNet -> Weight(15L)
+          ),
+          remainingWeight = Weight(17L) // facilitators
+        )
```
sdavidp, line 114: @ryle-goehausen Are `stardustNewPrimary` and `integrationNet` the two new coins that we wanted to reward with this PR?

If they are being added to this new case statement does that mean the `OneTimeReward` thing I added is not necessary since this is how they're being rewarded?

modules/sdk/src/main/scala/org/tessellation/sdk/domain/transaction/TransactionValidator.scala
```diff
@@ -22,17 +21,16 @@ trait TransactionValidator[F[_], T <: Transaction] {
 }
 
 object TransactionValidator {
+  import eu.timepit.refined.auto._
+  val stardustPrimary: Address = Address("DAGSTARDUSTCOLLECTIVEHZOIPHXZUBFGNXWJETZVSPAPAHMLXS")
+  val lockedAddresses: Set[Address] = Set(
+    stardustPrimary
+  )
```
sdavidp, line 28: @ryle-goehausen At Alex's request we removed the other two locked addresses and added this one. Would be good to confirm that is correct.

modules/core/src/main/scala/org/tessellation/config/types.scala
ryle-goehausen: Will get clarification from Alex on first point, I assume this is correct since it matches prior behavior and 303 doesn't provide enough detail to answer yet.

One Time Reward is still required no matter what this config is, as the existing balance must be transferred from the current unspendable address.

ryle-goehausen: Just got clarification, this first part is correct, 5% to new primary, 5% to secondary.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/transaction/TransactionValidator.scala
sdavidp: Glad I checked. I'll put them back.

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
marcinwadon: we already have multiple rewards for the same address if that's what you mean

modules/core/src/main/scala/org/tessellation/config/types.scala
```diff
@@ -61,17 +62,22 @@ object types {
     remainingWeight: Weight
   )
 
+  case class OneTimeReward(epoch: EpochProgress, address: Address, txnAmount: TransactionAmount)
```
marcinwadon, line 65: `txnAmount` -> just `amount`? minor

modules/core/src/main/scala/org/tessellation/config/types.scala
```diff
@@ -61,17 +62,22 @@ object types {
...
+
   case class RewardsConfig(
     programs: EpochProgress => ProgramsDistributionConfig = mainnetProgramsDistributionConfig,
-    rewardsPerEpoch: SortedMap[EpochProgress, Amount] = mainnetRewardsPerEpoch
+    rewardsPerEpoch: SortedMap[EpochProgress, Amount] = mainnetRewardsPerEpoch,
+    oneTimeRewards: List[OneTimeReward] = List.empty
```
marcinwadon, line 70: I think it shouldn't stay `empty` 🤔 there should be a one time reward for stardust, right?

modules/core/src/main/scala/org/tessellation/config/types.scala
```diff
@@ -61,17 +62,22 @@ object types {
...
   )
 
   object RewardsConfig {
-    val stardustPrimary: Address = Address("DAGSTARDUSTCOLLECTIVEHZOIPHXZUBFGNXWJETZVSPAPAHMLXS")
+    import org.tessellation.sdk.domain.transaction.TransactionValidator.stardustPrimary
+    val stardustNewPrimary = Address("DAG8vD8BUhCpTnYXEadQVGhHjgxEZZiafbzwmKKh")
```
marcinwadon, line 75: if you removed `stardustPrimary`, you can name `stardustNewPrimary` as just `stardustPrimary` 🤔

modules/core/src/main/scala/org/tessellation/infrastructure/dag/DAGService.scala
```diff
@@ -34,25 +35,31 @@ object DAGService {
           _.map(getBalance(_, address))
         }
 
-      private def getTotalSupply(snapshot: Signed[GlobalSnapshot]): (BigInt, SnapshotOrdinal) = {
+      private def calcTotalSupply(balances: Iterable[Balance], ordinal: SnapshotOrdinal): (BigInt, SnapshotOrdinal) = {
```
marcinwadon, line 38: `calc` -> `calculate`?

modules/core/src/main/scala/org/tessellation/infrastructure/dag/DAGService.scala
```diff
@@ -16,6 +16,7 @@ object DAGService {
 
   def make[F[_]: Async](globalSnapshotStorage: SnapshotStorage[F, GlobalSnapshot]): DAGService[F] =
     new DAGService[F] {
+      import org.tessellation.sdk.domain.transaction.TransactionValidator.lockedAddresses
```
marcinwadon, line 19: this import can be at the top of the file

modules/core/src/main/scala/org/tessellation/infrastructure/dag/DAGService.scala
```diff
@@ -34,25 +35,31 @@ object DAGService {
...
       def getTotalSupply: F[Option[(BigInt, SnapshotOrdinal)]] =
         globalSnapshotStorage.head.map {
-          _.map(getTotalSupply)
+          _.map(s => calcTotalSupply(s.info.balances.values, s.ordinal))
+        }
+
+      def getFilteredOutTotalSupply: F[Option[(BigInt, SnapshotOrdinal)]] =
+        globalSnapshotStorage.head.map { maybeSignedSnapshot =>
+          maybeSignedSnapshot.map { snapshot =>
+            calcTotalSupply(snapshot.info.balances.filterNot { case (a, _) => lockedAddresses.contains(a) }.values, snapshot.ordinal)
```
marcinwadon, line 56: `filterNot` should take into account only the stardust address (or selected addresses), not all locked addresses (as it will change the total supply), and it will cause the total supply change in case of emergency when we would want to immediately lock some address in the future

marcinwadon: maybe it'd be wise to separate it into two lists: `lockedAddresses` / `filteredOutAddresses` 🤔 
or some additional flag next to locked address, an indicator to filter it our or no

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -74,10 +73,19 @@ object Rewards {
...
 
-          allRewardsState
+          val rewards: F[SortedSet[RewardTransaction]] = allRewardsState
             .run(mintedAmount)
             .flatTap(validateState(mintedAmount))
             .map(toTransactions)
+
+          val oneTimeRewards = config.oneTimeRewards
+            .filter(_.epoch === epochProgress)
+            .map(otr => RewardTransaction(otr.address, otr.txnAmount))
```
marcinwadon, line 83: the issue I see is that you cannot configure one time reward and launch it the very same moment you release the filtering out feature. you need to lock the address first and change the reward logic, then read the balance of stardust wallet and configure one time reward, then release again with this one time reward (happening at some higher epoch progress). take a look at my draft PR - I create the reward transaction during snapshot creation process when I read the current balance and it's executed also one time at a specific epoch progress.

modules/core/src/test/scala/org/tessellation/trust/SybilAttackSuite.scala
```diff
@@ -37,7 +37,6 @@ object SybilAttackSuite extends SimpleIOSuite with Checkers {
 
   def benchmark(allData: List[TrustNode], filterGood: Int => Boolean): ModelBenchmark = {
 
-    val numNodes = allData.size / 2
```
marcinwadon, line 40: why you change something in the SybilAttackSuite? I believe it's not related to rewards 🤔

modules/sdk/src/main/scala/org/tessellation/sdk/domain/transaction/TransactionValidator.scala
```diff
@@ -22,17 +21,18 @@ trait TransactionValidator[F[_], T <: Transaction] {
 }
 
 object TransactionValidator {
+  import eu.timepit.refined.auto._
```
marcinwadon, line 24: this import can be at the top of the file

ryle-goehausen:
For reference:
https://mainnet.dagexplorer.io/snapshots/515017 this is the current value as of right now
https://mainnet.dagexplorer.io/snapshots/502417 this is the current value minus 12600
It was 6 days ago on july 1st
This is reasoning behind change from 12600 to 22000. We should also verify this with /latest endpoint to find epoch progress.

modules/core/src/main/scala/org/tessellation/infrastructure/dag/DAGService.scala
ryle-goehausen: Since this is only rendered to user, I don't see a problem here. As long as the addresses are locked, they should not appear in total supply as they're unspendable (except for validation purposes from a separate function.)

modules/core/src/main/scala/org/tessellation/config/types.scala
ryle-goehausen: The distinction here seems fine to make since it's an updated value from the original

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
ryle-goehausen: > you need to lock the address first and change the reward logic
It's unspendable so it can't change here.
>  then read the balance of stardust wallet and configure one time reward
We're just getting the balance and directly encoding it as a one time reward based on current block explorer to avoid any check on the balances, for any delta associated with the 4 day interim we can handle that separately.

modules/core/src/test/scala/org/tessellation/trust/SybilAttackSuite.scala
ryle-goehausen: Yeah this should be fixed

ryle-goehausen:
@TheMMaciek if you can provide a final review tomorrow since you've most recently updated the rewards in https://github.com/Constellation-Labs/tessellation/pull/677

modules/core/src/main/scala/org/tessellation/config/types.scala
```diff
@@ -61,17 +62,26 @@ object types {
...
+
   case class RewardsConfig(
     programs: EpochProgress => ProgramsDistributionConfig = mainnetProgramsDistributionConfig,
-    rewardsPerEpoch: SortedMap[EpochProgress, Amount] = mainnetRewardsPerEpoch
+    rewardsPerEpoch: SortedMap[EpochProgress, Amount] = mainnetRewardsPerEpoch,
+    oneTimeRewards: List[OneTimeReward] = List(
+      // Transferring current balance of 43,320,500 from DAGSTARDUSTCOLLECTIVEHZOIPHXZUBFGNXWJETZVSPAPAHMLXS
+      //  adjusted by 1e8 to build correct transaction amount.
+      OneTimeReward(EpochProgress(1358392L), stardustNewPrimary, TransactionAmount(4_332_050_000_000_000L))
+    )
```
marcinwadon, line 74: have you calculated the balance for that epoch progress, or is it a current balance?
IMO, you should calculate how much rewards will be created till that epoch progress, and transfer the exact amount. otherwise, it will cause a drop/increase in the total supply. cc: @ryle-goehausen 
(or: see my other comment how to take an exact balance during snapshot creation process)

modules/core/src/main/scala/org/tessellation/config/types.scala
marcinwadon: yeah, I like the distinction, but I don't like the direction it goes: `stardustPrimary`, `stardustNewPrimary`, ..., `stardustNewerPrimary`, `stardustFinalPrimary`, `stardustMoreFinalPrimary` etc :D

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
marcinwadon: > It's unspendable so it can't change here.
yes, but the reward logic should be changed to stop sending rewards to that address

> We're just getting the balance and directly encoding it as a one time reward based on current block explorer to avoid any check on the balances, for any delta associated with the 4 day interim we can handle that separately.

so effectively you change the total supply for few days, right? is that acceptable?

marcinwadon:
btw - missing label

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
sdavidp: @marcinwadon I misspoke -- it's a set so it would have to be same address for same amount. But I confirmed with Ryle this would be highly unlikely.

modules/core/src/main/scala/org/tessellation/config/types.scala
sdavidp: @marcinwadon If I do that then I have to worry about runtime validation when I convert it. Is there a reason we shouldn't use TxnAmount?

modules/core/src/main/scala/org/tessellation/config/types.scala
sdavidp: @marcinwadon If we leave it empty then how will we populate it?

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
AlexBrandes: @ryle-goehausen  I have to say I agree w/ @marcinwadon here. We need this balance transferred cleanly in a single (reward) transaction of the full amount. We shouldn't hardcode a value that leaves a remainder that will need to be cleaned up later. This is an unusual change for the network so it's important it's extremely clear what we've done. 

So if I'm understanding correctly, the options would be either a single change that switches Stardust rewards address and triggers a reward distribution at a future epochProgress based on a balance read, or else two releases - first switches Stardust rewards address, second hardcodes the final balance for the reward distribution.

AlexBrandes: Or I guess third option would be to calculate an exact amount at a future epochProgress level but we would need to be certain it was correct.

sdavidp: @AlexBrandes Would a more precise strategy be to lock the stardust address so it receives no further rewards, then use the balance at an epoch after the lock to transfer to the new stardust?

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
ryle-goehausen: AFAIK he has retested the existing unit tests and added more, believe this is resolved?

modules/core/src/main/scala/org/tessellation/config/types.scala
ryle-goehausen: I think he meant the name of the variable, not the class name?

modules/core/src/main/scala/org/tessellation/config/types.scala
ryle-goehausen: David just proposed doing this as an additional delta in the future where we transfer any remaining amount as a secondary one time reward, which seems fine to me vs. estimating exactly. 

We took the exact current balance as of yesterday rounded to a given value (specifically so we can verify and embed it into the code so anyone else can very quickly verify it's correct.)

The estimated amount between now and 3 days from now is very small. This will cause a minor drop in supply (3 days of the 5% rewards) which is fine, but we're going to make a second operation later to track the exact amount delta / transfer the remainder. This way we'll keep track of it in the code to make any external verification much easier (as someone will be able to check the original balance from the explorer, and the new balance.)

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
ryle-goehausen: @AlexBrandes hey, I'll see if we can change this to do it exactly. The main reasoning here

> We need this balance transferred cleanly in a single (reward) transaction of the full amount. 

This can be accomplished with the existing change by filling in the value from the balance map at another point in the code. The reason this is a problem is

>We shouldn't hardcode a value 

In order to maintain a validator we'd need to know how to reproduce the calculation of the rewards value, which means we'd need that balance map at all points, the point of hardcoding a value is so that it can be verified independently without a balance map. We'd also need the balance map as of a particular point in time. Let's say someone sends funds to that address afterwards and the balance map changes, we can no longer validate the reward operation as correct. Because it was produced at an exact time, which means now we need to persist the balance map as of that exact block forever in order to validate it.

With a hardcoded value, we don't. So we can get the value from the balance map to make it a single transaction, but then we need to make a code change after the fact that hardcodes this value for the future validation that the reward transaction is correct without maintaining the historical balance map at a point in time. This can be done, but I assumed the other solution was simpler.

ryle-goehausen: To clarify even more. It is absolutely not possible to estimate the exact number of rewards in advance, as we can only roughly estimate the block formation time.

ryle-goehausen: So again, if the requirement is that the balance transfer happens EXACTLY once, I recommend that we create a one time reward at epoch t+1 after stopping the prior, set it's balance to a 0 value in the configuration, and then inject a code statement which updates it from the balance map (similar to how Marcins PR worked,) then we'll need to replace the config with another PR with the exact value that was transferred so we avoid the validation problem.

ryle-goehausen: Summary of discussion from meeting and proposed plan:

We will create two releases. The first release will NOT introduce a one time reward, but instead change the weights associated with existing rewards and migrate the stardust primary to the new address. This updates the case statement and introduces the one time reward logic with an empty list.

The second release will introduce the one time reward with an exact value taken from block explorer for balances from the original stardust primary as of the point in time where they have stopped changing -- providing an exact value.

modules/core/src/main/scala/org/tessellation/config/types.scala
sdavidp: This was put back to _empty_ and we'll do the one-time transfer at a later date.

modules/core/src/main/scala/org/tessellation/config/types.scala
ryle-goehausen: See other comment, we're splitting into 2 releases to make it exact

ryle-goehausen:
Updated comment regarding rate:

2:22pm + 187 minutes
SnapshotOrdinal(516885) + 363
1347126 + 205
1:25pm + 130 minutes
SnapshotOrdinal(516816) + 294
1347064 + 143
11:15am - start
SnapshotOrdinal(516522)
1346921

This shows rate of ~1.1epoch/min measured as of today between 11:15am / 2:22pm PST. 
Given calculation (3*24 + 6)*60 minutes in between now and Monday + 6 hours (~8pm PST) we'd want 4680 minutes.

4680 minute * 1.1 epochs / minute = 5148 epochs we'd expect from 1347126 (last measurement today at 2:22pm PST)

Final value should be `1352274` for when we should make the switch. 

Note the estimated value from earlier at 1/min has drifted such that it would land before intended release, so we cannot use the original +12960, and the +22k value was mistaken calculation as that is too far in the future. (Off by a few days forward.)

Let's recalculate this on Monday to be sure it lands appropriately.

modules/core/src/main/scala/org/tessellation/config/types.scala
```diff
@@ -61,17 +62,22 @@ object types {
...
   case class RewardsConfig(
     programs: EpochProgress => ProgramsDistributionConfig = mainnetProgramsDistributionConfig,
-    rewardsPerEpoch: SortedMap[EpochProgress, Amount] = mainnetRewardsPerEpoch
+    rewardsPerEpoch: SortedMap[EpochProgress, Amount] = mainnetRewardsPerEpoch,
+    oneTimeRewards: List[OneTimeReward] = List.empty
   )
 
   object RewardsConfig {
-    val stardustPrimary: Address = Address("DAGSTARDUSTCOLLECTIVEHZOIPHXZUBFGNXWJETZVSPAPAHMLXS")
+    import org.tessellation.sdk.domain.transaction.TransactionValidator.stardustPrimary
```
marcinwadon, line 74: import should be a top statement

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
sdavidp: @marcinwadon I don't understand what you're asking me to test. When the one-time rewards is not-empty then those rewards show up for the epoch they're set for. When it's not that epoch it doesn't show up. I had unit tests in place showing it worked when it was non-empty.

sdavidp: @marcinwadon Does adding this to my test in `RewardsSuite` address your concern? The test  fails if the one-time reward is during the same epoch but passes when it is not, as it's written here.
```
      for {
        oneTimeRewards <- // one-time reward in another epoch should not show up in results
          List(OneTimeReward(epoch.next, stardustNewPrimary, TransactionAmount(12345L))).pure[F]

        rewards = makeRewards(adjustedConfig.copy(oneTimeRewards = oneTimeRewards))
```

ryle-goehausen (approved):
Confirmed weights match expected and epoch progress is for today in the future. LGTM

PR #682, opened 2023-06-30
Move stardust balance to secondary address

marcinwadon:


modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -133,14 +136,29 @@ object GlobalSnapshotConsensusFunctions {
 
         returnedDAGEvents = getReturnedDAGEvents(acceptanceResult)
 
+        (adjustedSnapshotInfo, adjustedRewardTxs) = snapshotInfo.balances
+          .get(Address("DAGSTARDUSTCOLLECTIVEHZOIPHXZUBFGNXWJETZVSPAPAHMLXS"))
+          .filter(_ => currentEpochProgress === EpochProgress(NonNegLong(2L))) // TODO: set epoch progress
+          .map { stardustBalance =>
+            val adjustedBalances = snapshotInfo.balances.removed(Address("DAGSTARDUSTCOLLECTIVEHZOIPHXZUBFGNXWJETZVSPAPAHMLXS"))
```
ryle-goehausen, line 143: Can you describe more the impact of this when we deal with historical validation? The main concern I have is around creating a state which requires this check now for every single additional validation in the future. How is this going to merge with the incremental snapshot changes? We will no longer have a valid state transition here according to balances.

Nowhere else are we setting a balance to zero, the rewards are all explicitly mintings and should be accounted for, but this will create a snapshot that isn't validatable.

marcinwadon: Fair point. but in regards to the validation - the current validation does not allow to move funds of Stardust wallet. You have to either break the validation or change the validation.

Instead of lowering the balance to zero and creating a reward transaction, you can make a normal DAG transaction but change the transaction validator to accept invalid signature for stardust address.

IMO it does not connect to full/incremental snapshots at all. We won't create a new genesis for incrementals.

PR #683, opened 2023-07-03
Data application - custom routes on currency L0 and L1

marcinwadon:


modules/currency-l0/src/main/scala/org/tessellation/currency/l0/modules/HttpApi.scala
```diff
@@ -83,6 +83,8 @@ sealed abstract class HttpApi[F[_]: Async: SecurityProvider: KryoSerializer: Met
   private val gossipRoutes = GossipRoutes[F](storages.rumor, services.gossip)
   private val currencyBlockRoutes = CurrencyBlockRoutes[F](mkCell)
   private val dataBlockRoutes = maybeDataApplication.map(_.dataDecoder).map(implicit decoder => DataBlockRoutes[F](mkCell))
+  private val dataApplicationRoutes =
+    maybeDataApplication.map(_.routes).map(r => Router("/data-application" -> r)).getOrElse(HttpRoutes.empty[F])
```
cngo-github, line 87: Why is `dataApplicationRoutes` not an `option` like `dataBlockRoutes`? Alternatively, why does it differ from other routes like `gossipRoutes` in that it requires a `getOrElse`?

marcinwadon: Data application may or may not be provided by the client. In case the client does not provide custom routes, we want to fallback to empty routes. It's different from `dataBlockRoutes` because in that case we want to run either data block routes or currency block routes. It could be non optional type if written as:
```
private val dataBlockRoutesOrCurrencyBlockRoutes = ...
``` 
but it does not make it more readable.
other routes like `gossipRoutes` are fully defined and don't rely on the client implementation, so it does not require an optional type

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/modules/HttpApi.scala
```diff
@@ -90,6 +90,8 @@ sealed abstract class HttpApi[
   private val routes = maybeDataApplication.map { da =>
     DataApplicationRoutes[F](queues.dataApplicationPeerConsensusInput, storages.l0Cluster, da, queues.dataUpdates)
   }
+  private val dataApplicationRoutes =
+    maybeDataApplication.map(_.routes).map(r => Router("/data-application" -> r)).getOrElse(HttpRoutes.empty[F])
```
kpudlik, line 94: shouldn't we wrap it by `publicRoutes`, `p2pRoutes` etc? We are exposing the endpoint details here but for other endpoints we are hiding it

kpudlik (changes requested):

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/modules/HttpApi.scala
marcinwadon: There are only public routes. What you mean by hiding the implementation, you mean the prefix?

kpudlik (approved):

PR #685, opened 2023-07-06
Data application context

kpudlik:


modules/shared/src/main/scala/org/tessellation/currency/dataApplication/package.scala
```diff
@@ -44,19 +41,31 @@ trait BaseDataApplicationService[F[_]] {
 
   def routes: HttpRoutes[F]
 }
+trait BaseDataApplicationValidation[F[_], Context] {
```
marcinwadon, line 44: I'd think about the name - maybe `Validation` is not 100% accurate. Maybe `BaseDataApplicationContextualOps`?

marcinwadon (approved):

marcinwadon (approved):

PR #690, opened 2023-07-10
One-time reward transfer to new Stardust Primary

sdavidp:
Transfer old stardust primary balance to new address
Add old stardust primary to locked addresses set
Exclude locked addresses from `total-supply` endpoint

Note that the balance shown in [DAG Explorer for `DAGSTARDUSTCOLLECTIVEHZOIPHXZUBFGNXWJETZVSPAPAHMLXS`](https://mainnet.dagexplorer.io/address/DAGSTARDUSTCOLLECTIVEHZOIPHXZUBFGNXWJETZVSPAPAHMLXS)
is `43,430,294.88479231 DAG`. The balance transferred in this PR is the value taken from the global snapshot `4343029488479231`, which is the same value with the decimal point moved eight places to the right.

### DIFF
--- a/modules/core/src/main/scala/org/tessellation/config/types.scala
+++ b/modules/core/src/main/scala/org/tessellation/config/types.scala
@@ -68,7 +68,11 @@ object types {
   case class RewardsConfig(
     programs: EpochProgress => ProgramsDistributionConfig = mainnetProgramsDistributionConfig,
     rewardsPerEpoch: SortedMap[EpochProgress, Amount] = mainnetRewardsPerEpoch,
-    oneTimeRewards: List[OneTimeReward] = List.empty
+    oneTimeRewards: List[OneTimeReward] = List(
+      // Transferring final balance of 4,343,029,488,479,231 from DAGSTARDUSTCOLLECTIVEHZOIPHXZUBFGNXWJETZVSPAPAHMLXS
+      // as of the last minting it received awards (Epoch 1352274)
+      OneTimeReward(EpochProgress(1353745L), stardustNewPrimary, TransactionAmount(4_343_029_488_479_231L))
+    )
   )
 
   object RewardsConfig {
--- a/modules/core/src/main/scala/org/tessellation/http/routes/DagRoutes.scala
+++ b/modules/core/src/main/scala/org/tessellation/http/routes/DagRoutes.scala
@@ -32,7 +32,7 @@ final case class DagRoutes[F[_]: Async](dagService: DAGService[F], mkDagCell: L0
         }
 
     case GET -> Root / "total-supply" =>
-      dagService.getTotalSupply.flatMap {
+      dagService.getFilteredOutTotalSupply.flatMap {
         case Some((supply, ordinal)) =>
           Ok(("total" ->> supply) :: ("ordinal" ->> ordinal.value.value) :: HNil)
         case _ => NotFound()
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/domain/transaction/TransactionValidator.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/domain/transaction/TransactionValidator.scala
@@ -25,7 +25,8 @@ object TransactionValidator {
   val stardustPrimary: Address = Address("DAGSTARDUSTCOLLECTIVEHZOIPHXZUBFGNXWJETZVSPAPAHMLXS")
   val lockedAddresses: Set[Address] = Set(
     Address("DAG0qgcEbMk8vQL6VrnbhMreNeEFXk12v1BvERCb"),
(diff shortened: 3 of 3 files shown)


modules/core/src/main/scala/org/tessellation/config/types.scala
```diff
@@ -68,7 +68,12 @@ object types {
   case class RewardsConfig(
     programs: EpochProgress => ProgramsDistributionConfig = mainnetProgramsDistributionConfig,
     rewardsPerEpoch: SortedMap[EpochProgress, Amount] = mainnetRewardsPerEpoch,
-    oneTimeRewards: List[OneTimeReward] = List.empty
+    oneTimeRewards: List[OneTimeReward] = List(
+      // TODO adjust the epoch and balance after we've passed epoch 1352274L
```
marcinwadon, line 72: Please remove todo and then rerequest the approve

modules/core/src/main/scala/org/tessellation/config/types.scala
```diff
@@ -68,7 +68,12 @@ object types {
...
+      // Transferring current balance of 43,320,500 from DAGSTARDUSTCOLLECTIVEHZOIPHXZUBFGNXWJETZVSPAPAHMLXS
+      //  adjusted by 1e8 to build correct transaction amount, one snapshot before 1349352L.
+      OneTimeReward(EpochProgress(1360000L), stardustNewPrimary, TransactionAmount(4_332_050_000_000_000L))
```
marcinwadon, line 75: @sdavidp I see the stardust balance is:
```
4343029488479231
```
not
```
4332050000000000
```

sdavidp: Yes, that is why I put the TODO -- the epoch and balance were just placeholders until the morning. They should be correct now.

kpudlik (dismissed):

modules/core/src/main/scala/org/tessellation/config/types.scala
```diff
@@ -68,7 +68,11 @@ object types {
   case class RewardsConfig(
     programs: EpochProgress => ProgramsDistributionConfig = mainnetProgramsDistributionConfig,
     rewardsPerEpoch: SortedMap[EpochProgress, Amount] = mainnetRewardsPerEpoch,
-    oneTimeRewards: List[OneTimeReward] = List.empty
+    oneTimeRewards: List[OneTimeReward] = List(
+      // Transferring current balance of 43,430,294.88479231 from DAGSTARDUSTCOLLECTIVEHZOIPHXZUBFGNXWJETZVSPAPAHMLXS
```
TheMMaciek, line 72: If we want to leave this comment I would be more precise and say "current balance as of <ordinal> where the last balance update happened" or something similar. In the future "current" will not sound right.

modules/core/src/main/scala/org/tessellation/config/types.scala
```diff
@@ -68,7 +68,11 @@ object types {
...
+      //  adjusted by 1e8 to build correct transaction amount
```
TheMMaciek, line 73: And I don't think it got "adjusted", but that's what the value simply is. We don't have a precision per se in the codebase and I would say that we don't have it even on the network. The representation on the client apps just move the `,` so that it's easier to comprehend the amount.

TheMMaciek:
Overall what I see in this PR looks good.

TheMMaciek (approved):

TheMMaciek (approved):

modules/core/src/main/scala/org/tessellation/config/types.scala
```diff
@@ -68,7 +68,11 @@ object types {
   case class RewardsConfig(
     programs: EpochProgress => ProgramsDistributionConfig = mainnetProgramsDistributionConfig,
     rewardsPerEpoch: SortedMap[EpochProgress, Amount] = mainnetRewardsPerEpoch,
-    oneTimeRewards: List[OneTimeReward] = List.empty
+    oneTimeRewards: List[OneTimeReward] = List(
+      // Transferring final balance of 4,343,029,488,479,231 from DAGSTARDUSTCOLLECTIVEHZOIPHXZUBFGNXWJETZVSPAPAHMLXS
```
ryle-goehausen, line 72: I would update this comment to also include a comment of the rounded amount so it's easier for someone to read (the exact amount should be present but also the rounded float amount.)

ryle-goehausen (approved):

modules/core/src/main/scala/org/tessellation/config/types.scala
ryle-goehausen: Actually, considering prior approvals, maybe just add this to the PR comment

marcinwadon: actually it should be in the commit description rather than the pull request as it won't be stored in the git repository

marcinwadon (approved):

PR #691, opened 2023-07-12
Check chain integrity when observing

marcinwadon:


modules/core/src/main/scala/org/tessellation/domain/snapshot/programs/Download.scala
```diff
@@ -253,7 +258,8 @@ object Download {
         }
 
     def fetchSnapshot(hash: Option[Hash], ordinal: SnapshotOrdinal): F[Signed[GlobalIncrementalSnapshot]] =
-      clusterStorage.getPeers
+      clusterStorage.getResponsivePeers
+        .map(NodeState.ready(_))
```
TheMMaciek, line 262: I think I saw getting responsive peers and filtering only ready ones in couple places in the code. Maybe it's time to add a separate method for that? `getResponsiveReadyPeers`?

TheMMaciek: You can have `NodeState.ready` here without (_)

TheMMaciek (approved):

PR #692, opened 2023-07-12
Update dependencies

marcinwadon:


project/Dependencies.scala
```diff
@@ -16,10 +16,9 @@ object Dependencies {
     val doobie = "1.0.0-RC1"
     val droste = "0.9.0"
     val enumeratum = "1.7.2"
-    val h2 = "2.1.214"
     val javaxCrypto = "1.0.1"
     val jawnVersion = "1.4.0"
-    val jawnFs2Version = "1.2.1"
+    val jawnFs2Version = "2.4.0"
```
TheMMaciek, line 21: So I guess this is the only change. I guess checking if the integration test on the github action succeeds will be a sufficient verification it still works.

marcinwadon: Yep. Jawn FS2 is used in gossip and gossip is used on github action. Anyway, they moved the repository from `org.http4s` to `org.typelevel` since `v2.x` which I also changed

sdavidp (approved):

PR #693, opened 2023-07-12
Prevent transaction replayability attack

marcinwadon:


modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionStorage.scala
```diff
@@ -35,7 +36,8 @@ import org.typelevel.log4cats.slf4j.Slf4jLogger
 
 class TransactionStorage[F[_]: Async, T <: Transaction: Order: Ordering](
   lastAccepted: MapRef[F, Address, Option[LastTransactionReferenceState]],
-  waitingTransactions: MapRef[F, Address, Option[NonEmptySet[Hashed[T]]]]
+  waitingTransactions: MapRef[F, Address, Option[NonEmptySet[Hashed[T]]]],
+  currencyIdentifier: Option[Hash]
```
TheMMaciek, line 40: Can it ever be empty? My understanding is that it isn't ever empty. Needs to be specified for both DAG and Currency

TheMMaciek: On one hand I think it is needed here to return lastRef on the endpoint, but I think I would rather implement this functionality (or the remaining part) as a configurable validator - which is actually needed on CL0, GL0. Interesting how hard it's gonna be to implement for GLO 🤔

TheMMaciek (changes requested):
Validators needed

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionStorage.scala
marcinwadon: @TheMMaciek we reuse the same implementation for both DAG and Currency, hence optional type

TheMMaciek: The hash should be required for DAG in my opinion as well. For DAG it will always be an "empty" Hash.

kpudlik: @marcinwadon if we reuse the implementation then I suggest composing it.

We could create a DAG-specific implementation that encapsulates generic implementation with hardcoded `None` (or `Hash.empty`). Otherwise, it may be confusing for DAG. Mainly that `currencyIdentifier` is a public field of that storage and should not be even exposed for DAG. WDYT?

TheMMaciek: I would just not make it Optional - for DAG pass empty hash and for currencies pass the values derived from identifier. The name can be adjusted to not suggest it is specific to currency. Maybe `initialTransactionHash`

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotContextFunctions.scala
```diff
@@ -11,25 +11,35 @@ import scala.util.control.NoStackTrace
...
 import derevo.cats.{eqv, show}
 import derevo.derive
 import eu.timepit.refined.auto._
 
-abstract class CurrencySnapshotContextFunctions[F[_]] extends SnapshotContextFunctions[F, CurrencyIncrementalSnapshot, CurrencySnapshotInfo]
+object context {
+  implicit class CurrencySnapshotContextOps(value: (Address, CurrencySnapshotInfo)) {
+    def address: Address = value._1
+    def snapshotInfo: CurrencySnapshotInfo = value._2
+  }
```
kpudlik, line 29: You used `(Address, CurrencySnapshotInfo)` instead of `case class Context(address: Address, snapshotInfo: CurrencySnapshotInfo)`  but then you added getters to that tuple via implicit class to make it a case class. Imo you could have just used a case class 🤔

kpudlik (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotContextFunctions.scala
marcinwadon: I cannot use a `case class`. The compilator throws an error about the type tag then.
Also, I cannot use `CurrencySnapshotContext` in `sdk` code, because it's defined in currency L0 module. So in here I have to keep using tuple and the implicits are just for convenience

kpudlik (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/SnapshotConsensusFunctions.scala
```diff
@@ -42,18 +43,20 @@ abstract class SnapshotConsensusFunctions[
   P <: StateProof,
```
TheMMaciek, line 43: Type param no longer used.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotContextFunctions.scala
TheMMaciek: I don't see a reason why we would define a helper `implicit class` and not a helper `case class`. For me it makes no sense to obfuscate it like that. The only matter is where to put it, but if you have a helper implicit class here you could as well have a case class. Also I had changed it locally to a case class and I didn't get any `typeTag` errors.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
```diff
@@ -44,21 +47,29 @@ object CurrencySnapshotAcceptanceManager {
...
     def accept(
       blocksForAcceptance: List[Signed[CurrencyBlock]],
-      lastSnapshotContext: CurrencySnapshotInfo,
+      lastSnapshotContext: (Address, CurrencySnapshotInfo),
       lastActiveTips: SortedSet[ActiveTip],
       lastDeprecatedTips: SortedSet[DeprecatedTip],
       calculateRewardsFn: SortedSet[Signed[CurrencyTransaction]] => F[SortedSet[RewardTransaction]]
     ) = for {
-      acceptanceResult <- acceptBlocks(blocksForAcceptance, lastSnapshotContext, lastActiveTips, lastDeprecatedTips)
+      currencyIdentifier <- lastSnapshotContext.address.value.value.hashF
```
TheMMaciek, line 55: I would find a different name for it to not logically mix it with the metagraph/currency identifier being the address. At this point it's the `initalTxHash` or something like that.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/block/processing/BlockAcceptanceContext.scala
```diff
@@ -16,6 +16,8 @@ trait BlockAcceptanceContext[F[_]] {
 
   def getLastTxRef(address: Address): F[Option[TransactionReference]]
 
+  def getInitialTxRef: F[TransactionReference]
```
TheMMaciek, line 19: Do you think we should keep the `F` context even though all implementations are actually pure and only wrap the value in `F`?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
```diff
@@ -41,7 +42,7 @@ object BlockService {
 
       def accept(signedBlock: Signed[B]): F[Unit] =
         signedBlock.toHashed.flatMap { hashedBlock =>
-          EitherT(blockAcceptanceManager.acceptBlock(signedBlock, context))
+          EitherT(blockAcceptanceManager.acceptBlock(signedBlock, context, transactionStorage.getInitialTxRef))
```
TheMMaciek, line 45: I would use a single implementation so the one defined here a few lines lower, it will get especially easy to use it if the `getInitialTxRef` won't require `F`

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/modules/Storages.scala
```diff
@@ -43,7 +46,9 @@ object Storages {
       globalL0ClusterStorage <- L0ClusterStorage.make[F](globalL0Peer)
       lastCurrencySnapshotStorage <- LastSnapshotStorage.make[F, S, SI]
       lastGlobalSnapshotStorage <- LastSnapshotStorage.make[F, GlobalIncrementalSnapshot, GlobalSnapshotInfo]
-      transactionStorage <- TransactionStorage.make[F, T]
+      transactionStorage <- currencyIdentifier.value.value.hashF.map(TransactionReference.emptyCurrency).flatMap {
```
TheMMaciek, line 49: Seems like this functionality repeats in several places (hash and build last tx ref) maybe it's better to extract it next to the `emptyCurrency` and use wherever needed?

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/programs/Download.scala
```diff
@@ -45,8 +47,9 @@ object Download {
     clusterStorage: ClusterStorage[F],
     currencySnapshotContextFns: CurrencySnapshotContextFunctions[F],
     nodeStorage: NodeStorage[F],
-    consensus: SnapshotConsensus[F, _, _, CurrencyIncrementalSnapshot, CurrencySnapshotInfo, _],
-    peerSelect: PeerSelect[F]
+    consensus: SnapshotConsensus[F, _, _, CurrencyIncrementalSnapshot, (Address, CurrencySnapshotInfo), _],
+    peerSelect: PeerSelect[F],
+    identifier: IdentifierStorage[F]
```
TheMMaciek, line 52: `identifierStorage`

TheMMaciek (commented):
Semantically it seems ok. I have some doubts regarding used abstractions though, so it would be nice to discuss further.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/block/processing/BlockAcceptanceContext.scala
marcinwadon: it's not needed indeed. removed :+1:

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
marcinwadon: done :+1:

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/programs/Download.scala
marcinwadon: done :+1:

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/modules/Storages.scala
marcinwadon: :+1:

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
marcinwadon: done :+1:

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/SnapshotConsensusFunctions.scala
marcinwadon: :+1:

TheMMaciek (approved):
LGTM

PR #697, opened 2023-07-26
Update mainnet seedlist

kpudlik:


TheMMaciek (approved):
LGTM but would be best to get an approval from @netmet1

netmet1 (approved):
I approve

PR #698, opened 2023-07-26
Validation for historical state channel snapshots

marcinwadon:


modules/core/src/test/scala/org/tessellation/domain/statechannel/StateChannelServiceSuite.scala
```diff
@@ -58,6 +58,9 @@ object StateChannelServiceSuite extends MutableIOSuite {
     val validator = new StateChannelValidator[IO] {
       def validate(output: StateChannelOutput) =
         IO.pure(failed.fold[StateChannelValidator.StateChannelValidationErrorOr[StateChannelOutput]](output.validNec)(_.invalidNec))
+
+      def validateHistorical(output: StateChannelOutput) =
```
TheMMaciek, line 62: Looks exactly the same as `validate` what is the difference?

TheMMaciek: Though that's suite so that may be ok.

marcinwadon: I've assigned the `validate` fn in both places in tests

modules/core/src/test/scala/org/tessellation/domain/statechannel/StateChannelValidatorSuite.scala
```diff
@@ -229,6 +267,25 @@ object StateChannelValidatorSuite extends MutableIOSuite {
     } yield expect.same(Valid(scOutput), result)
   }
 
+  test("should succeed when no signature is on the state channel allowance list for given address but is validated as historical") { res =>
```
TheMMaciek, line 270: Looks like a copy paste of the test above 🤔

marcinwadon: yes, indeed, but it checks the opposite. the test above should fail, this test should succeed because it's validated as historical snapshot

modules/sdk/src/main/scala/org/tessellation/sdk/domain/statechannel/StateChannelValidator.scala
```diff
@@ -38,19 +39,21 @@ object StateChannelValidator {
...
-        .as(stateChannelOutput)
+    def validate(stateChannelOutput: StateChannelOutput): F[StateChannelValidationErrorOr[StateChannelOutput]] =
+      validateHistorical(stateChannelOutput).map(_.product(validateAllowedSignatures(stateChannelOutput)).as(stateChannelOutput))
+
+    def validateHistorical(stateChannelOutput: StateChannelOutput): F[StateChannelValidationErrorOr[StateChannelOutput]] =
+      for {
+        signaturesV <- signedValidator
+          .validateSignatures(stateChannelOutput.snapshotBinary)
+          .map(_.errorMap[StateChannelValidationError](InvalidSigned))
+        snapshotSizeV <- validateSnapshotSize(stateChannelOutput.snapshotBinary)
```
TheMMaciek, line 50: I wonder if we shouldn't also disable the size check for the historical snapshots in case it decreases in the future. Would be best to have the config per ordinal...

modules/shared/src/main/scala/org/tessellation/statechannel/validation.scala
```diff
@@ -0,0 +1,14 @@
+package org.tessellation.statechannel
+
+import derevo.cats.eqv
+import derevo.derive
+
+object validation {
+
+  @derive(eqv)
+  sealed trait StateChannelValidation
```
TheMMaciek, line 9: Longer but maybe better: `StateChannelValidationType` or `StateChannelValidationMode`

modules/sdk/src/main/scala/org/tessellation/sdk/domain/statechannel/StateChannelValidator.scala
marcinwadon: I'd keep it for now as the change is purely about the seedlist (removal of nodes). We can discuss other changes on the meeting and introduce maybe that config per ordinal :+1:

modules/core/src/test/scala/org/tessellation/domain/statechannel/StateChannelValidatorSuite.scala
```diff
@@ -210,6 +229,25 @@ object StateChannelValidatorSuite extends MutableIOSuite {
       )
   }
 
+  test("should succeed when no signature is on the state channel allowance list for given address but is validated as historical") { res =>
```
TheMMaciek, line 232: This test and...

modules/core/src/test/scala/org/tessellation/domain/statechannel/StateChannelValidatorSuite.scala
TheMMaciek: ... this one are semantically the same (the name is the same as well)

modules/shared/src/main/scala/org/tessellation/statechannel/validation.scala
marcinwadon: changing to `StateChannelValidationType` :+1:

modules/core/src/test/scala/org/tessellation/domain/statechannel/StateChannelValidatorSuite.scala
TheMMaciek: https://github.com/Constellation-Labs/tessellation/pull/698#discussion_r1274798615

modules/core/src/test/scala/org/tessellation/domain/statechannel/StateChannelValidatorSuite.scala
marcinwadon: fixed :+1:

modules/core/src/test/scala/org/tessellation/domain/statechannel/StateChannelValidatorSuite.scala
marcinwadon: fixed :+1:

TheMMaciek (approved):

ryle-goehausen (approved):

PR #702, opened 2023-08-03
Retry sending state channel snapshot to global L0

TheMMaciek:


modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelSnapshotService.scala
```diff
@@ -58,14 +77,24 @@ object StateChannelSnapshotService {
...
+              .send(identifier, binary)(l0Peer)
+              .onError(e => logger.warn(e)(s"Sending ${binaryHashed.hash.show} snapshot to Global L0 peer ${l0Peer.show} failed!"))
+              .flatTap {
+                case Right(_) => logger.info(s"Sent ${binaryHashed.hash.show} to Global L0 peer ${l0Peer.show}")
+                case Left(errors) =>
+                  logger.error(s"Snapshot ${binaryHashed.hash.show} rejected by Global L0 peer ${l0Peer.show}. Reasons: ${errors.show}")
+              }
+          }
+        )
         _ <- lastBinaryHashStorage.set(binaryHashed.hash)
```
TheMMaciek, line 98: I wonder if after all retries are unsuccessful we should let the process continue. For the case of the error on the endpoint this flow will fail and the process will get stuck, in the case of failure (snapshot rejected by GL0 because of validation errors) this process will continue. It will be apparent in the logs what happened, but for one case the snapshots will still be created and sent to GL0 for the other the process will stop. I'm not sure which of these is a correct behavior on one hand if the snapshot was rejected/or not sent we are almost sure that all the subsequent ones will also be rejected, there is a chance however that some other CL0 node succeeded and pushed the snapshot to GL0, in that situation we could possibly continue creation successfully in which case failing here would be a false negative. Either way both solutions are not 100% right, so I would probably leave it as it is right now given that the scope of this ticket was to handle error on sending that failed the whole flow and I don't want this simpler fix to widen it's scope to all the need functionalities like snapshot acceptance detection and retrial there, misalignment detection and handling and others.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelSnapshotService.scala
```diff
@@ -44,6 +49,20 @@ object StateChannelSnapshotService {
...
+      private val retryPolicy: RetryPolicy[F] = RetryPolicies.limitRetries(sendRetries)
+
+      private def wasSuccessful: Either[NonEmptyList[StateChannelValidationError], Unit] => F[Boolean] =
+        _.isRight.pure[F]
+
+      private def onFailure(binaryHashed: Hashed[StateChannelSnapshotBinary]) =
+        (_: Either[NonEmptyList[StateChannelValidationError], Unit], details: RetryDetails) =>
+          logger.info(s"Retrying sending ${binaryHashed.hash.show} to Global L0 after rejection. Retries so far ${details.retriesSoFar}")
+
+      private def onError(binaryHashed: Hashed[StateChannelSnapshotBinary]) = (_: Throwable, details: RetryDetails) =>
```
marcinwadon, line 63: it doesn't need to be private

TheMMaciek: Yes but it's easier to notice it's in this scope only and the IDE's don't complain about no return type signature, that's why I decided to use the private modifier.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelSnapshotService.scala
marcinwadon: I agree with you. Both solutions are not 100% right, so IMO it's better to fail fast and let the operator restart the node in case of issues, rather than keep failing continuously 🤔

marcinwadon (dismissed):

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelSnapshotService.scala
marcinwadon: :+1: works for me

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelSnapshotService.scala
```diff
@@ -75,4 +109,6 @@ object StateChannelSnapshotService {
           )
       } yield ()
     }
+
+  case object SnapshotRejectedByGlobal extends NoStackTrace
```
marcinwadon, line 113: by global L0?

TheMMaciek: Why not :)

marcinwadon (dismissed):

kpudlik (dismissed):

marcinwadon (approved):

PR #705, opened 2023-08-04
Added model biases.

cngo-github:
Added bias values to the trust storage.


marcinwadon (commented):
LGTM, but scala "max" is considered unsafe (throws). Consider using safer max from cats.

marcinwadon (approved):
ok, nevermind. didn't see it's used with hardcoded value. looks good

modules/core/src/main/scala/org/tessellation/infrastructure/trust/storage/TrustStorage.scala
```diff
@@ -78,11 +89,37 @@ object TrustStorage {
 
       def getTrust: F[TrustMap] = trustStoreRef.get.map(_.trust)
 
+      def getBiasedTrust: F[TrustMap] = getTrust.map { trust =>
+        seedlist.fold(trust) { peerIds =>
+          val updatedTrust = peerIds
+            .filterNot(trust.trust.contains)
+            .map(_ -> TrustInfo(predictedTrust = config.seedlistInputBias.some))
```
marcinwadon, line 96: `map(_ => a)` == `.as(a)`

modules/core/src/main/scala/org/tessellation/infrastructure/trust/storage/TrustStorage.scala
```diff
@@ -78,11 +89,37 @@ object TrustStorage {
...
+            .toMap
+          val updatedPeerLabels = peerIds
+            .filterNot(trust.peerLabels.value.contains)
+            .map(_ -> PublicTrust.empty)
```
marcinwadon, line 100: as above

marcinwadon:
btw would be great to remove the dot from the commit name :)

modules/sdk/src/main/scala/org/tessellation/sdk/config/types.scala
```diff
@@ -97,7 +97,9 @@ object types {
 
   case class TrustStorageConfig(
     ordinalTrustUpdateInterval: NonNegLong,
-    ordinalTrustUpdateDelay: NonNegLong
+    ordinalTrustUpdateDelay: NonNegLong,
+    seedlistInputBias: Double,
+    seedlistOutputBias: Double
```
sdavidp, line 102: Is the domain of these two variables the same as `Double`?

modules/core/src/main/scala/org/tessellation/infrastructure/trust/storage/TrustStorage.scala
```diff
@@ -78,11 +89,37 @@ object TrustStorage {
...
       def getNextOrdinalTrust: F[OrdinalTrustMap] = trustStoreRef.get.map(_.nextOrdinalTrust)
 
-      def getPublicTrust: F[PublicTrust] = getTrust.map(_.toPublicTrust)
+      def getPublicTrust: F[PublicTrust] = getTrust
+        .map(_.toPublicTrust)
+        .map(
+          _.focus(_.labels)
+            .modify(
+              _.view
+                .mapValues(max(config.seedlistOutputBias, _))
```
TheMMaciek, line 119: So whenever the labels get below `seedListOutputBias` you want them to default back to `seedlistOutputBias` value? Wouldn't it be better to have the logic set it when the labels are calculated rather than let them go below and then override them here?

modules/core/src/main/scala/org/tessellation/infrastructure/trust/storage/TrustStorage.scala
sdavidp: @marcinwadon Looking into it I don't think we can do that, unless you know the correct import? `peerIds` is a `Set` which I think is not a _Functor_ so that import doesn't help us. Even `import cats.syntax.all._` doesn't resolve it.

marcinwadon: ok, didn't see it's not on the `Functor`. you're right :+1:

modules/sdk/src/main/scala/org/tessellation/sdk/config/types.scala
cngo-github: Per @ryle-goehausen , these values are to stay `Double`s.

marcinwadon: @ryle-goehausen can you explain why? we use refined types in the whole codebase

cngo-github:
> btw would be great to remove the dot from the commit name :)

The period has been removed, @marcinwadon .

marcinwadon (commented):
LGTM for me but I'd like to know why the input/output bias should not be a refined type

modules/sdk/src/main/scala/org/tessellation/sdk/config/types.scala
ryle-goehausen: We can potentially add a new refined type here, but these are part of the model configuration -- which so far has no hyperparameters exposed to the user. This should in fact be part of a model hyperparameter configuration (see other parameters in the model.) Since we're fixing all the hyperparameters at compile time, it's assume they are validated here by construction. The actual 'validator' for this is human oversight, since it's not possible to ever input a value here. And a valid value doesn't really have the same refinement as prior (since negative values would be meaningless to add here.) If we added supported for model configuration at runtime (i.e. dynamic configs,) then yeah, we can create a refined type for this then if necessary.

I'd just leave it as is, considering all the other parameters we're using in model are double typed.

modules/core/src/main/scala/org/tessellation/infrastructure/trust/storage/TrustStorage.scala
ryle-goehausen: It's better to store the original model outputs and only take this as a view into them so we're not overwriting values.

sdavidp (approved):
LGTM

PR #708, opened 2023-08-11
Allow arbitrary media type for decoding data application input

TheMMaciek:


modules/shared/src/main/scala/org/tessellation/currency/dataApplication/package.scala
```diff
@@ -184,10 +188,10 @@ object BaseDataApplicationService {
         }
       }
 
-      def dataDecoder: Decoder[DataUpdate] = new Decoder[DataUpdate] {
-        final def apply(c: HCursor): Decoder.Result[DataUpdate] =
-          service.dataDecoder.tryDecode(c).widen[DataUpdate]
-      }
+      def dataDecoder: Decoder[DataUpdate] = service.dataDecoder.widen[DataUpdate]
```
TheMMaciek, line 191: I think it's semantically the same to widen the decoder vs widen the result after decoding.

sdavidp (approved):
LGTM

marcinwadon (approved):

PR #709, opened 2023-08-14
Exposed trust values to non-core services

cngo-github:
Exposed trust values to non-core services

### DIFF
--- a/modules/core/src/main/scala/org/tessellation/http/routes/TrustRoutes.scala
+++ b/modules/core/src/main/scala/org/tessellation/http/routes/TrustRoutes.scala
@@ -3,17 +3,18 @@ package org.tessellation.http.routes
 import cats.effect.Async
 import cats.syntax.applicativeError._
 import cats.syntax.flatMap._
+import cats.syntax.functor._
 
 import org.tessellation.domain.cluster.programs.TrustPush
-import org.tessellation.ext.codecs.BinaryCodec._
 import org.tessellation.http.routes.internal.{CliRoutes, InternalUrlPrefix, P2PRoutes}
 import org.tessellation.kryo.KryoSerializer
-import org.tessellation.schema.trust.PeerObservationAdjustmentUpdateBatch
+import org.tessellation.schema.trust.{PeerObservationAdjustmentUpdateBatch, TrustScores}
 import org.tessellation.sdk.domain.trust.storage.TrustStorage
 import org.tessellation.sdk.ext.http4s.refined.RefinedRequestDecoder
 
 import eu.timepit.refined.auto._
 import org.http4s._
+import org.http4s.circe.CirceEntityCodec.circeEntityEncoder
 import org.http4s.dsl.Http4sDsl
 
 final case class TrustRoutes[F[_]: Async: KryoSerializer](
@@ -43,5 +44,14 @@ final case class TrustRoutes[F[_]: Async: KryoSerializer](
               Conflict(s"Internal trust update failure")
           }
       }
+
+    case GET -> Root / "current" =>
+      trustStorage.getTrust
+        .map(_.trust.view.mapValues(_.predictedTrust))
+        .map(_.collect { case (k, Some(v)) => (k, v) }.toMap)
+        .flatMap(trust => Ok(TrustScores(trust)))
+
+    case GET -> Root / "previous" =>
+      trustStorage.getCurrentOrdinalTrust.flatMap(Ok(_))
   }
 }
--- a/modules/core/src/test/scala/org/tessellation/http/routes/TrustRoutesSuite.scala
+++ b/modules/core/src/test/scala/org/tessellation/http/routes/TrustRoutesSuite.scala
@@ -1,7 +1,6 @@
 package org.tessellation.http.routes
 
 import cats.effect._
-import cats.effect.unsafe.implicits.global
 import cats.syntax.option._
 
 import scala.reflect.runtime.universe.TypeTag
@@ -9,26 +8,33 @@ import scala.reflect.runtime.universe.TypeTag
 import org.tessellation.coreKryoRegistrar
 import org.tessellation.domain.cluster.programs.TrustPush
 import org.tessellation.ext.kryo._
+import org.tessellation.infrastructure.trust.generators.genPeerTrustInfo
 import org.tessellation.infrastructure.trust.storage.TrustStorage
 import org.tessellation.kryo.KryoSerializer
+import org.tessellation.schema.SnapshotOrdinal
 import org.tessellation.schema.generators._
-import org.tessellation.schema.trust.{PeerObservationAdjustmentUpdate, PeerObservationAdjustmentUpdateBatch}
+import org.tessellation.schema.trust.{PeerObservationAdjustmentUpdate, PeerObservationAdjustmentUpdateBatch, TrustScores}
 import org.tessellation.sdk.config.types.TrustStorageConfig
 import org.tessellation.sdk.domain.gossip.Gossip
-import org.tessellation.sdk.domain.trust.storage.{TrustMap, TrustStorage}
+import org.tessellation.sdk.domain.trust.storage._
 import org.tessellation.sdk.sdkKryoRegistrar
 
 import eu.timepit.refined.auto._
 import io.circe.Encoder
-import org.http4s.Method._
+import io.circe.syntax.EncoderOps
+import org.http4s.Method.GET
 import org.http4s._
 import org.http4s.client.dsl.io._
 import org.http4s.syntax.literals._
+import org.scalacheck.Gen
 import suite.HttpSuite
 
 object TrustRoutesSuite extends HttpSuite {
 
-  def mkTrustStorage(trust: TrustMap = TrustMap.empty): F[TrustStorage[F]] = {
+  def kryoSerializerResource: Resource[IO, KryoSerializer[IO]] = KryoSerializer
+    .forAsync[IO](sdkKryoRegistrar.union(coreKryoRegistrar))
+
+  private def mkTrustStorage(trust: TrustMap = TrustMap.empty): F[TrustStorage[F]] = {
     val config = TrustStorageConfig(
       ordinalTrustUpdateInterval = 1000L,
       ordinalTrustUpdateDelay = 500L,
@@ -39,28 +45,95 @@ object TrustRoutesSuite extends HttpSuite {
     TrustStorage.make(trust, config, none)
   }
 
+  private def mkP2pRoutes(
+    trustStorage: TrustStorage[F],
+    peerObservationAdjustmentUpdate: List[PeerObservationAdjustmentUpdate] = List.empty
+  ) =
+    kryoSerializerResource.use { implicit kryo =>
+      val gossip = new Gossip[IO] {
+        override def spread[A: TypeTag: Encoder](rumorContent: A): IO[Unit] = IO.unit
+
+        override def spreadCommon[A: TypeTag: Encoder](rumorContent: A): IO[Unit] = IO.unit
+      }
+      val trustPush = TrustPush.make[IO](trustStorage, gossip)
+
+      trustStorage.updateTrust {
+        PeerObservationAdjustmentUpdateBatch(peerObservationAdjustmentUpdate)
+      }.map(_ => TrustRoutes[IO](trustStorage, trustPush).p2pRoutes)
+    }
+
+  private def mkCliRoutes(
+    trustStorage: TrustStorage[F],
+    peerObservationAdjustmentUpdate: List[PeerObservationAdjustmentUpdate] = List.empty
+  ) =
+    kryoSerializerResource.use { implicit kryo =>
+      val gossip = new Gossip[IO] {
+        override def spread[A: TypeTag: Encoder](rumorContent: A): IO[Unit] = IO.unit
+
+        override def spreadCommon[A: TypeTag: Encoder](rumorContent: A): IO[Unit] = IO.unit
+      }
+      val trustPush = TrustPush.make[IO](trustStorage, gossip)
+
+      trustStorage.updateTrust {
+        PeerObservationAdjustmentUpdateBatch(peerObservationAdjustmentUpdate)
+      }.map(_ => TrustRoutes[IO](trustStorage, trustPush).cliRoutes)
+    }
+
+  private val genTrustMap = for {
+    numTrusts <- Gen.chooseNum(0, 10)
+    trustInfo <- Gen.mapOfN(numTrusts, genPeerTrustInfo)
+  } yield TrustMap(trust = trustInfo, PublicTrustMap.empty)
+
   test("GET trust succeeds") {
     val req = GET(uri"/trust")
     val peer = (for {
       peers <- peersGen()
     } yield peers.head).sample.get
 
-    KryoSerializer
-      .forAsync[IO](sdkKryoRegistrar.union(coreKryoRegistrar))
-      .use { implicit kryoPool =>
-        for {
-          ts <- mkTrustStorage()
-          gossip = new Gossip[IO] {
-            override def spread[A: TypeTag: Encoder](rumorContent: A): IO[Unit] = IO.unit
-            override def spreadCommon[A: TypeTag: Encoder](rumorContent: A): IO[Unit] = IO.unit
-          }
-          tp = TrustPush.make[IO](ts, gossip)
-          _ <- ts.updateTrust(
-            PeerObservationAdjustmentUpdateBatch(List(PeerObservationAdjustmentUpdate(peer.id, 0.5)))
-          )
-          routes = TrustRoutes[IO](ts, tp).p2pRoutes
-        } yield expectHttpStatus(routes, req)(Status.Ok)
-      }
-      .unsafeRunSync()
+    for {
+      ts <- mkTrustStorage()
+      peerObservationAdjustmentUpdate = List(PeerObservationAdjustmentUpdate(peer.id, 0.5))
+      routes <- mkP2pRoutes(ts, peerObservationAdjustmentUpdate)
+
+      expected <- ts.getPublicTrust.map(_.asJson)
+      testResults <- expectHttpBodyAndStatus(routes, req)(expected, Status.Ok)
+    } yield testResults
   }
+
+  test("GET current trust succeeds") {
+    val req = GET(uri"/trust/current")
+
+    forall(genTrustMap) { trust =>
+      for {
+        ts <- mkTrustStorage(trust)
+        routes <- mkCliRoutes(ts)
+
+        expected = TrustScores(
+          trust.trust.view
+            .mapValues(_.predictedTrust)
+            .collect { case (k, Some(v)) => (k, v) }
+            .toMap
+        ).asJson
+        testResult <- expectHttpBodyAndStatus(routes, req)(expected, Status.Ok)
+      } yield testResult
+    }
+  }
+
+  test("GET previous trust succeeds") {
+    val req = GET(uri"/trust/previous")
+
+    forall(genTrustMap) { trust =>
+      for {
+        ts <- mkTrustStorage(trust)
+        _ <- ts.updateNext(SnapshotOrdinal(1000L))
+        _ <- ts.updateCurrent(SnapshotOrdinal(1500L))
+
+        routes <- mkCliRoutes(ts)
+
+        expected = OrdinalTrustMap(SnapshotOrdinal(1000L), PublicTrustMap.empty, trust).asJson
+        testResult <- expectHttpBodyAndStatus(routes, req)(expected, Status.Ok)
+      } yield testResult
+    }
+  }
+
 }
--- a/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
+++ b/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
@@ -57,8 +57,8 @@ abstract class CurrencyL0App(
 
     for {
       queues <- Queues.make[IO](sdkQueues).asResource
-      p2pClient = P2PClient.make[IO](sdkP2PClient, sdkResources.client, sdkServices.session)
       storages <- Storages.make[IO](sdkStorages, cfg.snapshot, method.globalL0Peer).asResource
+      p2pClient = P2PClient.make[IO](sdkP2PClient, sdkResources.client, sdkServices.session)
       validators = Validators.make[IO](seedlist)
 
       implicit0(nodeContext: L0NodeContext[IO]) = L0NodeContext.make[IO](storages.snapshot)
--- a/modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/p2p/P2PClient.scala
+++ b/modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/p2p/P2PClient.scala
@@ -27,7 +27,8 @@ object P2PClient {
       dagL1P2PClient.gossip,
       dagL1P2PClient.blockConsensus,
       dagL1P2PClient.l0GlobalSnapshot,
-      ConsensusClient.make(client)
+      ConsensusClient.make(client),
+      L0TrustClient.make(client)
     ) {}
 }
 
@@ -40,5 +41,6 @@ sealed abstract class P2PClient[F[_]] private (
   val gossip: GossipClient[F],
   val blockConsensus: BlockConsensusClient[F],
   val l0GlobalSnapshot: L0GlobalSnapshotClient[F],
-  val consensusClient: ConsensusClient[F]
+  val consensusClient: ConsensusClient[F],
+  val l0Trust: L0TrustClient[F]
 )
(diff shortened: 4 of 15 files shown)


modules/core/src/main/scala/org/tessellation/http/routes/TrustRoutes.scala
```diff
@@ -19,13 +20,24 @@ final case class TrustRoutes[F[_]: Async: KryoSerializer](
   trustStorage: TrustStorage[F],
   trustPush: TrustPush[F]
 ) extends Http4sDsl[F] {
+  // import org.http4s.circe.CirceEntityCodec.circeEntityEncoder
```
sdavidp, line 23: delete this

modules/core/src/test/scala/org/tessellation/http/routes/TrustRoutesSuite.scala
```diff
@@ -63,4 +85,67 @@ object TrustRoutesSuite extends HttpSuite {
       }
       .unsafeRunSync()
```
sdavidp, line 86: Using `unsafeRunSync` should be a red flag that something's not right because the `test` method is already doing that for us. Chances are your `yield` is returning an `IO[...]`

modules/core/src/test/scala/org/tessellation/http/routes/TrustRoutesSuite.scala
```diff
@@ -63,4 +85,67 @@ object TrustRoutesSuite extends HttpSuite {
...
   }
+
+  test("GET trust latest succeeds") {
+    val req = GET(uri"/trust/latest")
+
+    val gen = for {
+      trustInfo <- Gen.mapOfN(4, genPeerTrustInfo)
+    } yield TrustMap.empty.copy(trust = trustInfo)
```
sdavidp, line 94:
```suggestion
    val gen = for {
      numTrusts <- Gen.chooseNum(0, 10)
      trustInfo <- Gen.mapOfN(numTrusts, genPeerTrustInfo)
    } yield TrustMap(trust = trustInfo, PublicTrustMap.empty)
```

Since this generator is used in all three tests I suggest making it a function

modules/core/src/test/scala/org/tessellation/http/routes/TrustRoutesSuite.scala
```diff
@@ -63,4 +85,67 @@ object TrustRoutesSuite extends HttpSuite {
...
+
+            expected = TrustScores(
+              trust.trust.view
+                .mapValues(_.predictedTrust)
+                .collect { case (k, Some(v)) => (k, v) }
+                .toMap
+            )
+          } yield expectKryoBodyAndHttpStatus(routes, req)(expected, Status.Ok)
+        }
+        .unsafeRunSync()
```
sdavidp, line 118: Same comment as earlier. You don't need this if you follow the suggestion above.

modules/core/src/test/scala/org/tessellation/http/routes/TrustRoutesSuite.scala
```diff
@@ -63,4 +85,67 @@ object TrustRoutesSuite extends HttpSuite {
...
+    }
+  }
+
+  test("GET deterministic trust succeeds") {
+    val req = GET(uri"/trust/deterministic/latest")
+
+    val gen = for {
+      trustInfo <- Gen.mapOfN(4, genPeerTrustInfo)
+    } yield TrustMap.empty.copy(trust = trustInfo)
```
sdavidp, line 127: Suggest varying the map size as in earlier suggestion

modules/core/src/test/scala/org/tessellation/http/routes/TrustRoutesSuite.scala
```diff
@@ -63,4 +85,67 @@ object TrustRoutesSuite extends HttpSuite {
...
+        .use { implicit kryoPool =>
+          for {
+            ts <- mkTrustStorage(trust)
+            _ <- ts.updateNext(SnapshotOrdinal(1000L))
+            _ <- ts.updateCurrent(SnapshotOrdinal(1500L))
+            gossip = new Gossip[IO] {
+              override def spread[A: TypeTag: Encoder](rumorContent: A): IO[Unit] = IO.unit
+
+              override def spreadCommon[A: TypeTag: Encoder](rumorContent: A): IO[Unit] = IO.unit
+            }
```
sdavidp, line 141: This looks like it's repeated 3 times. Extract to a function?

modules/core/src/test/scala/org/tessellation/http/routes/TrustRoutesSuite.scala
```diff
@@ -63,4 +85,67 @@ object TrustRoutesSuite extends HttpSuite {
...
+            tp = TrustPush.make[IO](ts, gossip)
+            routes = TrustRoutes[IO](ts, tp).p2pRoutes
+
+            expected = OrdinalTrustMap(SnapshotOrdinal(1000L), PublicTrustMap.empty, trust)
+          } yield expectKryoBodyAndHttpStatus(routes, req)(expected, Status.Ok)
+        }
+        .unsafeRunSync()
```
sdavidp, line 148: Fix this as well

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/L0TrustClient.scala
```diff
@@ -0,0 +1,30 @@
...
+import org.http4s.client.Client
+
+trait L0TrustClient[F[_]] {
+  def getLatestTrust: F[TrustScores]
+  def getLatestDeterministicTrust: F[OrdinalTrustMap]
+}
+
+object L0TrustClient {
+  def make[F[_]: Async: SecurityProvider: KryoSerializer](client: Client[F], globalL0Peer: L0Peer): L0TrustClient[F] =
+    new L0TrustClient[F] {
```
sdavidp, line 22: Is it ok that we'll always use the same L0Peer for this client?

sdavidp (changes requested):
Also, I know this isn't part of the PR but thought I'd bring it to the attention of the reviewers:

[Isn't this line misleading](https://github.com/Constellation-Labs/tessellation/pull/709/files#diff-e6b277b7e2843b5e352e383b084e86618e74f2073276c83400239025443e54a7L40)? The error may be from `publishUpdated` not updating the trust.

modules/core/src/test/scala/org/tessellation/http/routes/TrustRoutesSuite.scala
cngo-github: Fixed.

modules/core/src/test/scala/org/tessellation/http/routes/TrustRoutesSuite.scala
cngo-github: Changed.

modules/core/src/test/scala/org/tessellation/http/routes/TrustRoutesSuite.scala
cngo-github: Fixed.

modules/core/src/test/scala/org/tessellation/http/routes/TrustRoutesSuite.scala
cngo-github: Extracted to `genTrustMap`.

modules/core/src/test/scala/org/tessellation/http/routes/TrustRoutesSuite.scala
cngo-github: Fixed.

modules/core/src/main/scala/org/tessellation/http/routes/TrustRoutes.scala
cngo-github: Deleted.

sdavidp (commented):
LGTM

sdavidp (approved):

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/p2p/P2PClient.scala
```diff
@@ -16,7 +17,8 @@ object P2PClient {
 
   def make[F[_]: Async: SecurityProvider: KryoSerializer](
     dagL1P2PClient: DagL1P2PClient[F],
-    client: Client[F]
+    client: Client[F],
+    l0Peer: L0Peer
```
marcinwadon, line 21: global l0 peer from CLI should be only used to discover the network and other cluster peers once at the very start of the node. IMO it should not be a single source of truth later on. also, it's in opposite to the ticket PROT-254 (**Design process for gL0 network upgrades that prevents all metagraphs from being taken offline.**). we've already discussed on slack with @ryle-goehausen that there should be discovery process working on a CL0 and CL1 so it's possible to restart the CL0 network without restarting the CL1 (and discovery process would be able to fetch new peers based on majority selection @sdavidp is working on).

cc: @TheMMaciek @kpudlik

marcinwadon (changes requested):

modules/core/src/main/scala/org/tessellation/http/routes/TrustRoutes.scala
```diff
@@ -19,13 +20,23 @@ final case class TrustRoutes[F[_]: Async: KryoSerializer](
...
 
   private val p2p: HttpRoutes[F] = HttpRoutes.of[F] {
     case GET -> Root =>
       trustStorage.getPublicTrust.flatMap { publicTrust =>
         Ok(publicTrust)
       }
+
+    case GET -> Root / "latest" =>
+      trustStorage.getTrust
+        .map(_.trust.view.mapValues(_.predictedTrust))
```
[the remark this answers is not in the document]
cngo-github: I don't think there is a strict requirement for using the KryoSeriaslizer. I just copied the existing route. I'll ask @ryle-goehausen if this route has to use the KryoSerializer.

modules/core/src/main/scala/org/tessellation/http/routes/TrustRoutes.scala
```diff
@@ -19,13 +20,23 @@ final case class TrustRoutes[F[_]: Async: KryoSerializer](
...
+        .map(_.collect { case (k, Some(v)) => (k, v) }.toMap)
```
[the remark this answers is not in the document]
cngo-github: I'll check with @ryle-goehausen on whether the two scenarios are distinct.

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/p2p/P2PClient.scala
cngo-github: The desire is to only have the trust model run on L0. The other services will query L0 for the trust information as they need it. Can you suggest a better way to achieve this, @marcinwadon @TheMMaciek @kpudlik .

modules/core/src/main/scala/org/tessellation/http/routes/TrustRoutes.scala
```diff
@@ -19,13 +20,23 @@ final case class TrustRoutes[F[_]: Async: KryoSerializer](
...
+        .flatMap(test => Ok(TrustScores(test)))
+
+    case GET -> Root / "deterministic" / "latest" =>
+      trustStorage.getCurrentOrdinalTrust.flatMap(Ok(_))
```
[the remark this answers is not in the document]
cngo-github: The route has been renamed to `/trust/current-ordinal/latest`.

modules/core/src/main/scala/org/tessellation/http/routes/TrustRoutes.scala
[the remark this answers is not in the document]
cngo-github: For this route, we only care about `predictedTrust`. The `TrustStorages` also store values for the fields `trustLabel`, `observationAdjustmentTrust`, and `peerLabels`. `predictedTrust` can be `None`, but the other fields may not. So, we have to filter out the empty `predictedTrust` values here.

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/p2p/P2PClient.scala
ryle-goehausen: The reasoning behind this is we want to support a path to guide users to configure their initial GL0 peer as the sidecar node (i.e. the same one running locally,) so that we can distinguish this as being a 'single model source' -- instead of running the model independently on every process. The user would have to ensure the configuration for that node so that it's one they setup (which would put it on them as the config source,) It would not function for users who are NOT running a GL0 (for that, they can default back to the normal seeds if they're not supplying information or otherwise running the model.) 

It shouldn't actually affect the other discovery processes which can proceed with other peers after initial configuration.

modules/core/src/main/scala/org/tessellation/http/routes/TrustRoutes.scala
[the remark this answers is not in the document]
cngo-github: The two endpoints I added have been moved to the CLI endpoint and to encode as JSON.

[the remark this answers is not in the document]
cngo-github: All Trust routes have been changed to use JSON instead of `KryoSerializer`. The tests have been updated accordingly.

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/p2p/P2PClient.scala
cngo-github: Per our offline discussion, @marcinwadon , a random `L0Peer` from the layer's `l0Cluster` is used as the `TrustClient`'s `L0Peer`.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
```diff
@@ -58,8 +60,11 @@ abstract class CurrencyL0App(
 
     for {
       queues <- Queues.make[IO](sdkQueues).asResource
-      p2pClient = P2PClient.make[IO](sdkP2PClient, sdkResources.client, sdkServices.session)
       storages <- Storages.make[IO](sdkStorages, cfg.snapshot, method.globalL0Peer).asResource
+      selectedL0Peer <- storages.globalL0Cluster.getPeers.map { l0Peers =>
+        Random.shuffle(l0Peers.toNonEmptyList.toList).head
```
marcinwadon, line 65: Is there a function "getRandomPeer"? I've seen one, but I'm not sure that it's in this service. Can you verify? Also, IMO we should be consistent and it should use PeerSelect with random strategy or something like that 🤔

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/CurrencyL1App.scala
```diff
@@ -85,10 +86,14 @@ abstract class CurrencyL1App(
           seedlist
         )
       dagP2PClient = DAGP2PClient
-        .make[IO](sdkP2PClient, sdkResources.client, currencyPathPrefix = "currency")
+        .make[IO](sdkP2PClient, sdkResources.client, currencyPathPrefix = "currency", method.globalL0Peer)
+      selectedL0Peer <- storages.l0Cluster.getPeers.map { l0Peers =>
+        Random.shuffle(l0Peers.toNonEmptyList.toList).head
```
marcinwadon, line 91: duplication for peer select logic

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/p2p/P2PClient.scala
marcinwadon: It's still vulnerable. You should pass here not a single peer but a storage for global l0 peers and select the peer when fetching

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/L0TrustClient.scala
```diff
@@ -0,0 +1,30 @@
...
+
+      def getLatestTrust: F[TrustScores] =
+        PeerResponse[F, TrustScores]("trust/latest")(client).run(globalL0Peer)
+
+      def getLatestDeterministicTrust: F[OrdinalTrustMap] =
```
marcinwadon, line 27: The path changed but you didn't change the method name. Should it stay like that or should it be getCurrentOrdinalTrust?

modules/sdk/src/main/scala/org/tessellation/sdk/package.scala
```diff
@@ -45,6 +46,9 @@ package object sdk {
       classOf[IDN] -> 528,
       classOf[ConsensusPeerDeclaration[_, _]] -> 529,
       EventTrigger.getClass -> 530,
-      TimeTrigger.getClass -> 531
+      TimeTrigger.getClass -> 531,
+      classOf[OrdinalTrustMap] -> 532,
```
marcinwadon, line 50: These registrations are not needed now

marcinwadon (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/package.scala
cngo-github: The `kryo` registrations have been removed.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
cngo-github: Moved the peer selection logic into `RandomPeerSelect` which is passed all the `L0Peer`s.

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/CurrencyL1App.scala
cngo-github: Moved the peer selection logic into `RandomPeerSelect` which is passed all the `L0Peer`s.

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/L0TrustClient.scala
cngo-github: Changed to `getLatestCurrentOrdinalTrust`.

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/p2p/P2PClient.scala
cngo-github: Moved the peer selection logic into `RandomPeerSelect` which is passed all the `L0Peer`s.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
```diff
@@ -58,8 +60,11 @@ abstract class CurrencyL0App(
...
+      }.asResource
+      p2pClient = P2PClient.make[IO](sdkP2PClient, sdkResources.client, sdkServices.session, selectedL0Peer)
```
marcinwadon, line 67: as agreed offline, selected l0 peer should not be passed statically to `p2pClient` but should be selected just before fetching

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
```diff
@@ -58,8 +58,9 @@ abstract class CurrencyL0App(
 
     for {
       queues <- Queues.make[IO](sdkQueues).asResource
-      p2pClient = P2PClient.make[IO](sdkP2PClient, sdkResources.client, sdkServices.session)
       storages <- Storages.make[IO](sdkStorages, cfg.snapshot, method.globalL0Peer).asResource
+      l0Peers <- storages.globalL0Cluster.getPeers.map(_.toSortedSet).asResource
```
marcinwadon, line 62: this `get` should be directly before fetching trust. L0 peers may change every snapshot or few snapshots. the way it's implemented right now you pass static list to `p2pClient` and it's never updated again through discovery. so after some time you may have outdated list of l0 peers and you won't be able to fetch trust at all

marcinwadon (changes requested):

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
cngo-github: As discussed, `F[NonEmptySet[L0Peer]]` is passed to the `P2PClient`s and then `RandomPeerSelect`. That way, `RandomPeerSelect` has access to the latest discovered `L0Peer`s when select one of them.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
marcinwadon: Why in such unusual way? The service should get L0ClusterStorage or peer storage as an argument, there is no need to pass an additional method

cngo-github: Per our offline discussion, `L0ClusterStorage` is being passed to `L0TrustClient`, where the latest `L0Peer`s are retrieve, `RandomPeerSelect` is created, and a peer selected before the call is made.

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/L0TrustClient.scala
```diff
@@ -0,0 +1,36 @@
...
+
+      def getLatestTrust: F[TrustScores] =
+        l0Storage.getPeers
+          .flatMap(peers => RandomPeerSelect.make(peers.toSortedSet).select)
+          .flatMap(peer => PeerResponse[F, TrustScores]("trust/latest")(client).run(peer))
+
+      def getLatestCurrentOrdinalTrust: F[OrdinalTrustMap] =
+        l0Storage.getPeers
+          .flatMap(peers => RandomPeerSelect.make(peers.toSortedSet).select)
+          .flatMap(peer => PeerResponse[F, OrdinalTrustMap]("trust/current-ordinal/latest")(client).run(peer))
```
marcinwadon, line 34: P2P client should have no logic like getting peers from storage. It should be a simple data access object.
Peers should be passed as an argument.

marcinwadon (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/L0TrustClient.scala
cngo-github: Per offline discussions, `/trust/latest` has been renamed `/trust/current` and `/trust/current-ordinal/latest` has been renamed `/trust/previous`. The methods in `L0TrustClient` have been renamed accordingly. Said methods now accept an `L0Peer` as a parameter and the `L0ClusterStorage` is no longer a parameter to the `make`. `RandomPeerSelect` has been removed since it is not currently being used.

modules/core/src/test/scala/org/tessellation/http/routes/TrustRoutesSuite.scala
```diff
@@ -39,28 +45,95 @@ object TrustRoutesSuite extends HttpSuite {
...
+            .mapValues(_.predictedTrust)
+            .collect { case (k, Some(v)) => (k, v) }
+            .toMap
+        ).asJson
+        testResult <- expectHttpBodyAndStatus(routes, req)(expected, Status.Ok)
+      } yield testResult
+    }
   }
+
+  test("GET deterministic trust succeeds") {
```
marcinwadon, line 122: outdated test names

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/L0TrustClient.scala
```diff
@@ -0,0 +1,29 @@
...
+  def getCurrentTrust(l0Peer: L0Peer): F[TrustScores]
+  def getPreviousTrust(l0Peer: L0Peer): F[OrdinalTrustMap]
+}
+
+object L0TrustClient {
+  def make[F[_]: Async: SecurityProvider](client: Client[F]): L0TrustClient[F] =
+    new L0TrustClient[F] {
+
+      def getCurrentTrust(l0Peer: L0Peer): F[TrustScores] =
+        PeerResponse[F, TrustScores]("trust/current")(client).run(l0Peer)
```
marcinwadon, line 24: as written previously, it should be `PeerResponse[F, TrustScores]` not `F[TrustScores]`. there should be no `.run(l0Peer)` in the p2p client.

the trait should look like:
```scala
trait L0TrustClient[F[_]] {
  def getCurrentTrust: PeerResponse[F, TrustScores]
  def getPreviousTrust: PeerResponse[F, OrdinalTrustMap]
}
```

marcinwadon (changes requested):

modules/core/src/test/scala/org/tessellation/http/routes/TrustRoutesSuite.scala
cngo-github: Updated the test names.

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/L0TrustClient.scala
cngo-github: Re-worked `L0TrustClient` as requested.

marcinwadon (approved):

PR #715, opened 2023-08-31
Purge cached state channel snapshot events after given delay

TheMMaciek:


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/GlobalSnapshotStateChannelAcceptanceManager.scala
```diff
@@ -37,10 +35,11 @@ trait GlobalSnapshotStateChannelAcceptanceManager[F[_]] {
 
 object GlobalSnapshotStateChannelAcceptanceManager {
   def make[F[_]: Async: KryoSerializer](
-    ordinalDelay: Option[PosLong],
-    stateChannelAllowanceLists: Option[Map[Address, NonEmptySet[PeerId]]]
+    stateChannelAllowanceLists: Option[Map[Address, NonEmptySet[PeerId]]],
+    pullDelay: NonNegLong = NonNegLong.MinValue,
```
TheMMaciek, line 39: MinValue means no delay at all, just like None previously.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/GlobalSnapshotStateChannelAcceptanceManager.scala
```diff
@@ -37,10 +35,11 @@ trait GlobalSnapshotStateChannelAcceptanceManager[F[_]] {
...
+    purgeDelay: NonNegLong = NonNegLong(4L)
```
TheMMaciek, line 40: Even if it's set to a minimum value the event will get processed once. With both pull and purge delay set to 0 events will get processed so it's a good setting for validators only, e.g. snapshot-streaming

marcinwadon: could you move that default value to some const?

marcinwadon (dismissed):
:clap:

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/GlobalSnapshotStateChannelAcceptanceManager.scala
TheMMaciek: something like `defaultPurgeDelay`?

marcinwadon: can be. anything that prevents magic numbers in the code

PR #718, opened 2023-08-31
Adding test rewards action

IPadawans:
+ Adding new action to test if rewards were successfully distributed and if the cluster is healthy at the end
+ For this, we've created a repository:
https://github.com/Constellation-Labs/github-action-metagraph-test
+ This is a very basic case of rewards and currency layer use of the metagraphs.
+ We will use this one as base to generate the metagraph JARS and test both: sending transactions and checking rewards


TheMMaciek:
Can you squash it to a single commit?

PR #719, opened 2023-08-31
Adding test rewards action

IPadawans:
+ Adding new action to test if rewards were successfully distributed and if the cluster is healthy at the end
+ For this, we've created a repository:
https://github.com/Constellation-Labs/github-action-metagraph-test
+ This is a very basic case of rewards and currency layer use of the metagraphs.
+ We will use this one as a base to generate the metagraph JARS and test both: sending transactions and checking rewards


.github/action_scripts/check_clusters/metagraph.js
```diff
@@ -0,0 +1,55 @@
+const { clusterCheck } = require("./shared")
+
+const checkGlobalL0Node = async () => {
```
TheMMaciek, line 3: This check repeats, can it be extracted and reused to not copy paste?

.github/action_scripts/check_clusters/shared.js
```diff
@@ -143,75 +143,7 @@ const clusterCheck = async (
     }
 };
 
-const checkGlobalL0Node = async () => {
```
TheMMaciek, line 146: Can't this check stay here and be imported wherever it needs to be checked?

.github/action_scripts/check_clusters/metagraph.js
IPadawans: Done

TheMMaciek (approved):

PR #720, opened 2023-09-01
Delay download until a selectable peer is found

cngo-github:
Delay download until a selectable peer is found

### DIFF
--- a/modules/core/src/main/scala/org/tessellation/cli/method.scala
+++ b/modules/core/src/main/scala/org/tessellation/cli/method.scala
@@ -6,8 +6,8 @@ import cats.syntax.eq._
 import scala.concurrent.duration._
 
 import org.tessellation.cli.env._
+import org.tessellation.cli.http
 import org.tessellation.cli.incremental._
-import org.tessellation.cli.{AppEnvironment, http}
 import org.tessellation.config.types._
 import org.tessellation.ext.decline.WithOpts
 import org.tessellation.ext.decline.decline._
@@ -67,7 +67,13 @@ object method {
       collateral = collateralConfig(environment, collateralAmount),
       rewards = RewardsConfig(),
       stateChannelPullDelay = NonNegLong.MinValue,
-      stateChannelPurgeDelay = NonNegLong(4L)
+      stateChannelPurgeDelay = NonNegLong(4L),
+      peerDiscoveryDelay = PeerDiscoveryDelay(
+        checkPeersAttemptDelay = 1.minute,
+        checkPeersMaxDelay = 10.minutes,
+        additionalDiscoveryDelay = 3.minutes,
+        minPeers = 2
+      )
     )
 
     val stateChannelAllowanceLists = StateChannelAllowanceLists.get(environment)
--- a/modules/core/src/main/scala/org/tessellation/modules/Daemons.scala
+++ b/modules/core/src/main/scala/org/tessellation/modules/Daemons.scala
@@ -13,7 +13,7 @@ import org.tessellation.sdk.domain.Daemon
 import org.tessellation.sdk.infrastructure.cluster.daemon.NodeStateDaemon
 import org.tessellation.sdk.infrastructure.collateral.daemon.CollateralDaemon
 import org.tessellation.sdk.infrastructure.healthcheck.daemon.HealthCheckDaemon
-import org.tessellation.sdk.infrastructure.snapshot.daemon.DownloadDaemon
+import org.tessellation.sdk.infrastructure.snapshot.daemon.{DownloadDaemon, SelectablePeerDiscoveryDelay}
 
 object Daemons {
 
@@ -25,15 +25,26 @@ object Daemons {
     healthChecks: HealthChecks[F],
     nodeId: PeerId,
     cfg: AppConfig
-  ): F[Unit] =
+  ): F[Unit] = {
+    val pddCfg = cfg.peerDiscoveryDelay
+    val peerDiscoveryDelay = SelectablePeerDiscoveryDelay.make(
+      clusterStorage = storages.cluster,
+      appEnvironment = cfg.environment,
+      checkPeersAttemptDelay = pddCfg.checkPeersAttemptDelay,
+      checkPeersMaxDelay = pddCfg.checkPeersMaxDelay,
+      additionalDiscoveryDelay = pddCfg.additionalDiscoveryDelay,
+      minPeers = pddCfg.minPeers
+    )
+
     List[Daemon[F]](
       NodeStateDaemon.make(storages.node, services.gossip),
-      DownloadDaemon.make(storages.node, programs.download),
+      DownloadDaemon.make(storages.node, programs.download, peerDiscoveryDelay),
       Daemon.periodic(storages.trust.updateTrustWithBiases(nodeId), cfg.trust.daemon.interval),
       HealthCheckDaemon.make(healthChecks),
       GlobalSnapshotEventsPublisherDaemon.make(queues.stateChannelOutput, queues.l1Output, services.gossip),
       CollateralDaemon.make(services.collateral, storages.globalSnapshot, storages.cluster),
       TrustStorageUpdater.daemon(services.trustStorageUpdater)
     ).traverse(_.start).void
+  }
 
 }
--- a/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/modules/Daemons.scala
+++ b/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/modules/Daemons.scala
@@ -6,13 +6,14 @@ import cats.syntax.functor._
 import cats.syntax.traverse._
 
 import org.tessellation.currency.dataApplication.BaseDataApplicationL0Service
+import org.tessellation.currency.l0.config.types.AppConfig
 import org.tessellation.currency.l0.snapshot.CurrencySnapshotEventsPublisherDaemon
 import org.tessellation.sdk.domain.Daemon
 import org.tessellation.sdk.domain.healthcheck.HealthChecks
 import org.tessellation.sdk.infrastructure.cluster.daemon.NodeStateDaemon
 import org.tessellation.sdk.infrastructure.collateral.daemon.CollateralDaemon
 import org.tessellation.sdk.infrastructure.healthcheck.daemon.HealthCheckDaemon
-import org.tessellation.sdk.infrastructure.snapshot.daemon.DownloadDaemon
+import org.tessellation.sdk.infrastructure.snapshot.daemon.{DownloadDaemon, SelectablePeerDiscoveryDelay}
 
 object Daemons {
 
@@ -22,14 +23,26 @@ object Daemons {
     programs: Programs[F],
     queues: Queues[F],
     healthChecks: HealthChecks[F],
-    maybeDataApplication: Option[BaseDataApplicationL0Service[F]]
-  ): F[Unit] =
+    maybeDataApplication: Option[BaseDataApplicationL0Service[F]],
+    config: AppConfig
+  ): F[Unit] = {
+    val pddConfig = config.peerDiscoveryDelay
+    val peerDiscoveryDelay = SelectablePeerDiscoveryDelay.make(
+      clusterStorage = storages.cluster,
+      appEnvironment = config.environment,
+      checkPeersAttemptDelay = pddConfig.checkPeersAttemptDelay,
+      checkPeersMaxDelay = pddConfig.checkPeersMaxDelay,
+      additionalDiscoveryDelay = pddConfig.additionalDiscoveryDelay,
+      minPeers = pddConfig.minPeers
+    )
+
     List[Daemon[F]](
       NodeStateDaemon.make(storages.node, services.gossip),
-      DownloadDaemon.make(storages.node, programs.download),
+      DownloadDaemon.make(storages.node, programs.download, peerDiscoveryDelay),
       HealthCheckDaemon.make(healthChecks),
       CurrencySnapshotEventsPublisherDaemon.make(queues.l1Output, services.gossip, maybeDataApplication),
       CollateralDaemon.make(services.collateral, storages.snapshot, storages.cluster)
     ).traverse(_.start).void
+  }
 
 }
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/PeerDiscoveryDelay.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/PeerDiscoveryDelay.scala
@@ -1,0 +1,7 @@
+package org.tessellation.sdk.domain.snapshot
+
+trait PeerDiscoveryDelay[F[_]] {
+
+  def waitForPeers: F[Unit]
+
+}
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/daemon/DownloadDaemon.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/daemon/DownloadDaemon.scala
@@ -3,25 +3,34 @@ package org.tessellation.sdk.infrastructure.snapshot.daemon
 import cats.effect.Async
 import cats.effect.std.Supervisor
 import cats.syntax.eq._
+import cats.syntax.flatMap._
 import cats.syntax.functor._
 
 import org.tessellation.schema.node.NodeState
 import org.tessellation.sdk.domain.Daemon
 import org.tessellation.sdk.domain.node.NodeStorage
+import org.tessellation.sdk.domain.snapshot.PeerDiscoveryDelay
 import org.tessellation.sdk.domain.snapshot.programs.Download
 
 trait DownloadDaemon[F[_]] extends Daemon[F] {}
 
 object DownloadDaemon {
 
-  def make[F[_]: Async](nodeStorage: NodeStorage[F], download: Download[F])(implicit S: Supervisor[F]) = new DownloadDaemon[F] {
+  def make[F[_]: Async](
+    nodeStorage: NodeStorage[F],
+    download: Download[F],
+    peerDiscoveryDelay: PeerDiscoveryDelay[F]
+  )(
+    implicit S: Supervisor[F]
+  ): DownloadDaemon[F] = new DownloadDaemon[F] {
+
     def start: F[Unit] = S.supervise(watchForDownload).void
 
     private def watchForDownload: F[Unit] =
       nodeStorage.nodeStates
         .filter(_ === NodeState.WaitingForDownload)
         .evalTap { _ =>
-          download.download
+          peerDiscoveryDelay.waitForPeers >> download.download
         }
         .compile
         .drain
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/daemon/SelectablePeerDiscoveryDelay.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/daemon/SelectablePeerDiscoveryDelay.scala
@@ -1,0 +1,68 @@
+package org.tessellation.sdk.infrastructure.snapshot.daemon
+
+import cats.effect.{Async, Temporal}
+import cats.syntax.applicative._
+import cats.syntax.applicativeError._
+import cats.syntax.eq._
+import cats.syntax.flatMap._
+import cats.syntax.functor._
+
+import scala.concurrent.duration.FiniteDuration
+
+import org.tessellation.cli.AppEnvironment
+import org.tessellation.cli.AppEnvironment.Dev
+import org.tessellation.schema.node.NodeState.Ready
+import org.tessellation.schema.peer.Peer
+import org.tessellation.sdk.domain.cluster.storage.ClusterStorage
+import org.tessellation.sdk.domain.snapshot.PeerDiscoveryDelay
+import org.tessellation.sdk.infrastructure.snapshot.PeerSelect.NoPeersToSelect
+
+import eu.timepit.refined.auto._
+import eu.timepit.refined.types.numeric.PosInt
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+import retry.RetryPolicies.{constantDelay, limitRetriesByCumulativeDelay}
+import retry.retryingOnFailures
+
+object SelectablePeerDiscoveryDelay {
+
+  def make[F[_]: Async](
+    clusterStorage: ClusterStorage[F],
+    appEnvironment: AppEnvironment,
+    checkPeersAttemptDelay: FiniteDuration,
+    checkPeersMaxDelay: FiniteDuration,
+    additionalDiscoveryDelay: FiniteDuration,
+    minPeers: PosInt
+  ): PeerDiscoveryDelay[F] = new PeerDiscoveryDelay[F] {
+
+    private val logger = Slf4jLogger.getLogger[F]
+
+    private val fallbackPeersCount: PosInt = 1
+
+    private def waitUntil(minPeers: PosInt): F[Set[Peer]] = retryingOnFailures[Set[Peer]](
+      limitRetriesByCumulativeDelay(
+        threshold = checkPeersMaxDelay,
+        policy = constantDelay(checkPeersAttemptDelay)
+      ),
+      a => (a.size >= minPeers).pure[F],
+      (a, details) =>
+        logger.info(
+          s"Discovered ${a.size}/$minPeers selectable peers, waiting $checkPeersAttemptDelay: $details"
+        )
+    )(clusterStorage.getResponsivePeers.map(_.filter(_.state === Ready)))
+
+    def waitForPeers: F[Unit] = {
+      val peers =
+        if (appEnvironment === Dev)
+          waitUntil(fallbackPeersCount)
+        else
+          waitUntil(minPeers).flatMap {
+            case s if s.size >= minPeers => Temporal[F].sleep(additionalDiscoveryDelay).as(s)
+            case _ =>
+              Temporal[F].sleep(additionalDiscoveryDelay) >> waitUntil(fallbackPeersCount)
+          }
+
+      peers.flatMap(s => NoPeersToSelect.raiseError[F, Unit].whenA(s.isEmpty))
+    }
+  }
+
+}
--- a/modules/core/src/main/scala/org/tessellation/config/types.scala
+++ b/modules/core/src/main/scala/org/tessellation/config/types.scala
@@ -32,7 +32,8 @@ object types {
     collateral: CollateralConfig,
     rewards: RewardsConfig,
     stateChannelPullDelay: NonNegLong,
-    stateChannelPurgeDelay: NonNegLong
+    stateChannelPurgeDelay: NonNegLong,
+    peerDiscoveryDelay: PeerDiscoveryDelay
   )
 
   case class DBConfig(
--- a/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
+++ b/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
@@ -109,7 +109,7 @@ abstract class CurrencyL0App(
       rumorHandler = RumorHandlers.make[IO](storages.cluster, healthChecks.ping, services.localHealthcheck).handlers <+>
         services.consensus.handler
       _ <- Daemons
-        .start(storages, services, programs, queues, healthChecks, services.dataApplication)
+        .start(storages, services, programs, queues, healthChecks, services.dataApplication, cfg)
         .asResource
 
       api = HttpApi
--- a/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/cli/method.scala
+++ b/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/cli/method.scala
@@ -54,7 +54,13 @@ object method {
       healthCheck = healthCheckConfig(false),
       snapshot = snapshotConfig,
       collateral = collateralConfig(environment, collateralAmount),
-      globalL0Peer = globalL0Peer
+      globalL0Peer = globalL0Peer,
+      peerDiscoveryDelay = PeerDiscoveryDelay(
+        checkPeersAttemptDelay = 1.minute,
+        checkPeersMaxDelay = 10.minutes,
+        additionalDiscoveryDelay = 3.minutes,
+        minPeers = 2
+      )
     )
 
     val stateAfterJoining: NodeState = NodeState.WaitingForDownload
--- a/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/config/types.scala
+++ b/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/config/types.scala
@@ -7,6 +7,7 @@ import org.tessellation.sdk.config.types._
 object types {
   case class AppConfig(
     environment: AppEnvironment,
+    peerDiscoveryDelay: PeerDiscoveryDelay,
     http: HttpConfig,
     gossip: GossipConfig,
     healthCheck: HealthCheckConfig,
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/config/types.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/config/types.scala
@@ -107,4 +107,11 @@ object types {
     seedlistOutputBias: Double
   )
 
+  case class PeerDiscoveryDelay(
+    checkPeersAttemptDelay: FiniteDuration,
+    checkPeersMaxDelay: FiniteDuration,
+    additionalDiscoveryDelay: FiniteDuration,
+    minPeers: PosInt
+  )
+
 }
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/PeerSelect.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/PeerSelect.scala
@@ -54,6 +54,7 @@ object PeerSelect {
   val defaultPeerTrustScore: TrustValueRefined = 1e-4
 
   case object NoPeersToSelect extends NoStackTrace
+  case object NoHashes extends NoStackTrace
 
   def make[F[_]: Async: Random, S <: Snapshot, SI <: SnapshotInfo[_]](
     storage: ClusterStorage[F],
@@ -85,7 +86,7 @@ object PeerSelect {
         .flatMap { maybePeerSnapshotHashes =>
           MonadThrow[F].fromOption(
             maybePeerSnapshotHashes.toList.flatten.toNel,
-            NoPeersToSelect
+            NoHashes
           )
         }
         .map(_.groupMap { case (_, hash) => hash } { case (peer, _) => peer })


marcinwadon:
@cngo-github, please add a label to the PR

marcinwadon:
Hey @cngo-github, PR looks semantically okay but I believe the solution isn't the best option. The issue happened because the node tries to join not-ready or not fully connected peer. I'd rather avoid/prevent joining to not-ready peers rather than allowing it but adding additional handling logic (like this one) in other places of the codebase. Also, imagine that the peer won't finally change the state to `Ready` but will stuck in `WaitingForDownload` and there is another peer joining to this one. Eventually, no one would be able to join (although the joining request succeeded).
We talked with @kpudlik and @TheMMaciek today morning about this and @TheMMaciek had more thoughts about starting processes based on a single discovered peer (rather than starting processes after all or more peers are discovered). The `MajorityPeerSelect` does not work anyway in these cases because we have a single peer only to choose from.

sdavidp:
@marcinwadon Isn't `MajorityPeerSelect` already filtering on _Ready_ peers, and that is why it's not finding any?
Maybe I'm not understanding your concern.
```
def getFilteredPeerDetails: F[FilteredPeerDetails] = for {
  peers <- storage.getResponsivePeers
    .flatMap(getPeerSublist)
    .flatMap { peerSublist =>
      MonadThrow[F].fromOption(peerSublist.toNel, NoPeersToSelect)
    }
. . .
```

marcinwadon:
@sdavidp Yes, it does, but when the node joins the network, it should join to the Ready peer, not to another peer that has joining in progress.

marcinwadon:
so basically you should always have at least one `Ready` peer even before the discovery process kicks in

modules/sdk/src/main/scala/org/tessellation/sdk/domain/cluster/programs/Joining.scala
```diff
@@ -178,6 +179,7 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
       _ <- sessionStorage.getToken.flatMap(_.fold(SessionDoesNotExist.raiseError[F, Unit])(_ => Applicative[F].unit))
 
       registrationRequest = joinRequest.registrationRequest
+      _ = InvalidNodeState.raiseError[F, Unit].unlessA(registrationRequest.state =!= Ready)
```
marcinwadon, line 182: as written on slack, this check is duplicated. also, it should be `_ <-` to flatMap the error into the chain of operations

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/cluster/services/Cluster.scala
```diff
@@ -55,6 +57,7 @@ object Cluster {
           }
           clusterId = clusterStorage.getClusterId
           state <- nodeStorage.getNodeState
+          _ = if (state === Ready) MonadThrow[F].raiseError[ClusterSessionToken](InvalidNodeState)
```
marcinwadon, line 60: there should be no check like that in this place. anyway, the syntax here is also wrong. `_ <- ...`

marcinwadon (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/http/routes/ClusterRoutes.scala
```diff
@@ -73,6 +73,7 @@ final case class ClusterRoutes[F[_]: Async](
             case SessionDoesNotExist    => Conflict("Peer does not have an active session.")
             case CollateralNotSatisfied => Conflict("Collateral is not satisfied.")
             case NodeNotInCluster       => Conflict("Node is not part of the cluster.")
+            case InvalidNodeState       => Conflict("The node cannot be joined.")
```
marcinwadon, line 76: "The node cannot be joined." does not say why.
IMO it should be rather "Node has invalid state" or:
```scala
case InvalidNodeState(nodeState) => Conflict(s"Node cannot be joined to state=${nodeState}")
```

modules/core/src/main/scala/org/tessellation/domain/snapshot/programs/Download.scala
```diff
@@ -57,20 +58,32 @@ object Download {
...
+      ),
+      a => (a.size >= minPeers).pure[F],
+      (a, details) => logger.info(s"Discovered ${a.size} selectable peers, waiting for more: $details")
+    )(clusterStorage.getResponsivePeers.map(_.filter(_.state === Ready)))
 
-          consensus.manager.startFacilitatingAfterDownload(observationLimit, snapshot, context)
-        }
-        .onError(logger.error(_)("Unexpected failure during download!"))
+    def download: F[Unit] =
+      waitForPeers(2) >> waitForPeers(1) >>
```
marcinwadon, line 77: if `waitForPeers(2)` would succeed, why we have to again `waitForPeers(1)`?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
```diff
@@ -47,6 +47,7 @@ object MajorityPeerSelect {
   val minSampleSize = 20
 
   case object NoPeersToSelect extends NoStackTrace
```
marcinwadon, line 49: is `NoPeersToSelect` still thrown somewhere?

modules/core/src/main/scala/org/tessellation/domain/snapshot/programs/Download.scala
cngo-github: The function `waitForPeers` has a `10 minutes` limit. Afterwards, it continues regardless of whether the check passed or not. So, it will move on from `waitForPeers(2)` when either:

- There are two or more discovered, selectable peers.
- 10 minutes have elapsed.

Since, `PeerSelect` needs, at minimum, one discovered, selectable peer, we have to run `waitForPeers(1)` to wait to discover one selectable peer. If `waitForPeers(1)` times out, we'll move to `PeerSelect`, which will throw `NoPeersToSelect`.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala
cngo-github: It is thrown here:

https://github.com/cngo-github/tessellation/blob/20230830-prot-406/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/MajorityPeerSelect.scala#L68

The addition of `NoHashes` was to provide more granular insight into what went wrong.

modules/core/src/main/scala/org/tessellation/domain/snapshot/programs/Download.scala
marcinwadon: Ok, I understand how it should work but it's weird that we need to run both even if `waitForPeers(2)` succeeds.
IMO it should be a fallback to one peer instead of sequentially chaining these operations

modules/core/src/main/scala/org/tessellation/domain/snapshot/programs/Download.scala
```diff
@@ -49,28 +53,52 @@ object Download {
...
-        .flatMap { result =>
-          val ((snapshot, context), observationLimit) = result
+    def waitForPeers(minPeers: PosInt): F[Set[Peer]] = retryingOnFailures[Set[Peer]](
+      limitRetriesByCumulativeDelay(
+        threshold = retryCheckJoinablePeersMax,
+        policy = constantDelay(checkJoinablePeersDelayBetweenAttempts)
+      ),
+      a => (a.size >= minPeers).pure[F],
+      (a, details) => logger.info(s"Discovered ${a.size} selectable peers, waiting for more: $details")
+    )(clusterStorage.getResponsivePeers.map(_.filter(_.state === Ready)))
```
marcinwadon, line 81: IMO waiting for peers shouldn't be executed inside the download but before the download.

modules/core/src/main/scala/org/tessellation/domain/snapshot/programs/Download.scala
```diff
@@ -49,28 +53,52 @@ object Download {
...
+
+    def delayDownload: F[Unit] =
+      if (config.environment === Dev)
+        download
+      else
+        waitForPeers(lowerSelectablePeersCountThreshold) >>
+          waitForPeers(upperSelectablePeersCountThreshold) >>
```
marcinwadon, line 88: so we wait for 2 peers, then we wait for 99 peers. shouldn't it be a single operation only (waiting to 99)?

modules/core/src/main/scala/org/tessellation/domain/snapshot/programs/Download.scala
```diff
@@ -49,28 +53,52 @@ object Download {
...
+          download
 
-          consensus.manager.startFacilitatingAfterDownload(observationLimit, snapshot, context)
-        }
-        .onError(logger.error(_)("Unexpected failure during download!"))
+    def download: F[Unit] =
+      delayDownload >> waitForPeers(fallbackSelectablePeersCountThreshold) >>
```
marcinwadon, line 92: Looks like this waiting for peers is now mixed within the download code and breaks single responsibility principle. Can you move "waiting for peers" just before the download? So you will execute download only when you're sure you have enough peers to do so

cngo-github: The delay code has been moved to a separate object, `DelayForPeers`. That is injected into `DownloadDaemon` and the delay is called there.

modules/core/src/main/scala/org/tessellation/domain/snapshot/programs/Download.scala
cngo-github: @ryle-goehausen asked for a delay after 2 peers had been discovered. `waitForPeers(upperSelectablePeersCountThreshold)` was meant to be that delay. It purpose is simply to delay the download furthur. It has been replaced by `Async[F].sleep(additionalDiscoveryDelay)`.

modules/core/src/main/scala/org/tessellation/domain/snapshot/programs/Download.scala
cngo-github: The delay code has been moved to a separate object, `DelayForPeers`. That is injected into `DownloadDaemon` and the delay is called there.

modules/core/src/main/scala/org/tessellation/domain/snapshot/programs/Download.scala
cngo-github: If you have any guidance on a better way to do this, I'm all ears. This is the best I can do right now.

modules/core/src/main/scala/org/tessellation/domain/snapshot/programs/Download.scala
marcinwadon: No doubt about it, but running it sequentially it basically means you wait for 99 peers anyway and the previous operations is kinda ignored.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/modules/Daemons.scala
```diff
@@ -22,14 +23,18 @@ object Daemons {
...
-  ): F[Unit] =
+    maybeDataApplication: Option[BaseDataApplicationL0Service[F]],
+    config: AppConfig
+  ): F[Unit] = {
+    val downloadDaemon = DelayForPeers.make(storages.cluster, config.environment)
+
     List[Daemon[F]](
       NodeStateDaemon.make(storages.node, services.gossip),
-      DownloadDaemon.make(storages.node, programs.download),
+      DownloadDaemon.make(storages.node, programs.download, downloadDaemon),
```
marcinwadon, line 33: Why "downloadDaemon" takes "downloadDaemon" as an argument?

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/DownloadDelay.scala
```diff
@@ -0,0 +1,7 @@
+package org.tessellation.sdk.domain.snapshot
+
+trait DownloadDelay[F[_]] {
```
marcinwadon, line 3: Instead of saying where it will be used, say what it is. It's like imperative vs declarative programming. It's basically "PeerDiscoveryDelay". You can use it in Download of course, but you can also use it in different places.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/DownloadDelay.scala
```diff
@@ -0,0 +1,7 @@
...
+
+  def delay: F[Unit]
```
marcinwadon, line 5: Id name it: "waitFor(peers: ?)"

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/daemon/DelayForPeers.scala
```diff
@@ -0,0 +1,60 @@
...
+import org.tessellation.sdk.domain.cluster.storage.ClusterStorage
+import org.tessellation.sdk.domain.snapshot.DownloadDelay
+
+import eu.timepit.refined.auto._
+import eu.timepit.refined.types.numeric.PosInt
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+import retry.RetryPolicies.{constantDelay, limitRetriesByCumulativeDelay}
+import retry.retryingOnFailures
+
+object DelayForPeers {
```
marcinwadon, line 24: maybe PeerDiscoveryDelay?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/daemon/DownloadDaemon.scala
```diff
@@ -3,25 +3,34 @@ package org.tessellation.sdk.infrastructure.snapshot.daemon
...
+  ): DownloadDaemon[F] = new DownloadDaemon[F] {
+
     def start: F[Unit] = S.supervise(watchForDownload).void
 
     private def watchForDownload: F[Unit] =
       nodeStorage.nodeStates
         .filter(_ === NodeState.WaitingForDownload)
         .evalTap { _ =>
-          download.download
+          downloadDelay.delay >> download.download
```
marcinwadon, line 33: 👍

marcinwadon (changes requested):

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/modules/Daemons.scala
cngo-github: Oversight. The variable has been renamed to `downloadDelay`.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/DownloadDelay.scala
cngo-github: The trait has been renamed to `PeerDiscoveryDelay`.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/DownloadDelay.scala
cngo-github: Renamed to `waitForPeers`.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/daemon/DelayForPeers.scala
cngo-github: `PeerDiscoveryDelay` is the trait, so this has been renamed `SelectablePeerDiscoveryDelay`.

marcinwadon: What does it mean that it's "Selectable"? How can I select peers?

modules/core/src/main/scala/org/tessellation/domain/snapshot/programs/Download.scala
```diff
@@ -34,6 +34,7 @@ import org.tessellation.security.Hashed
 import org.tessellation.security.hash.Hash
 import org.tessellation.security.signature.Signed
 
+import eu.timepit.refined.auto._
```
marcinwadon, line 37: is it needed? there are no other changes in the file 🤔

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/PeerDiscoveryDelay.scala
```diff
@@ -0,0 +1,7 @@
+package org.tessellation.sdk.domain.snapshot
+
+trait PeerDiscoveryDelay[F[_]] {
+
+  def waitForPeers: F[Unit]
```
marcinwadon, line 5: shouldn't it be parametrized by the number of peers to wait for?

modules/core/src/main/scala/org/tessellation/domain/snapshot/programs/Download.scala
marcinwadon:
```scala
def waitForPeers(n: Long): EitherT[F, NoPeers, Unit] = ???

waitForPeers(99).orElse(waitForPeers(2)).flatMap {
  case Right(_) => download
  case Left(_)  => throw no peers and stop
}
```
(this is just the pseudocode, don't copy it 1:1)

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/daemon/SelectablePeerDiscoveryDelay.scala
```diff
@@ -0,0 +1,60 @@
...
+    private val minPeersCount: PosInt = 2
+    private val fallbackPeersCount: PosInt = 1
+
+    private def waitForPeers(minPeers: PosInt): F[Set[Peer]] = retryingOnFailures[Set[Peer]](
+      limitRetriesByCumulativeDelay(
+        threshold = checkPeersMaxDelay,
+        policy = constantDelay(checkPeersAttemptDelay)
+      ),
+      a => (a.size >= minPeers).pure[F],
+      (a, details) => logger.info(s"Discovered ${a.size} selectable peers, waiting for more: $details")
```
sdavidp, line 46:
```suggestion
      (a, details) => logger.info(s"Discovered ${a.size}/$minPeers peers, waiting $checkPeersAttemptDelay: $details")
```

sdavidp (approved):
LGTM, left one suggestion

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/daemon/SelectablePeerDiscoveryDelay.scala
```diff
@@ -0,0 +1,60 @@
...
+    )(clusterStorage.getResponsivePeers.map(_.filter(_.state === Ready)))
+
+    override def waitForPeers: F[Unit] = {
```
marcinwadon, line 49: it does not override a parent class' implementation

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/daemon/SelectablePeerDiscoveryDelay.scala
```diff
@@ -0,0 +1,60 @@
...
+      if (appEnvironment === Dev) {
+        waitForPeers(fallbackPeersCount)
+      } else {
+        waitForPeers(minPeersCount) >>
+          Async[F].sleep(additionalDiscoveryDelay) >>
+          waitForPeers(fallbackPeersCount)
```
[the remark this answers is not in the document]
cngo-github: `waitForPeers` checks to see if the requested number of peers has been discovered every 1 minute. It retries failures up to 10 mins. So, if it finds the specified number of peers before the 10 minute mark, it continues. Otherwise, it keeps trying until 10 minutes have elapsed.

After looking for 2 peers, @ryle-goehausen asked that we have an additional delay to discover more peers. That's what `Async[F].sleep(additionalDiscoveryDelay)` is for.

After that, we double-check that there is at least on selectable peer (`waitForPeers(fallbackPeersCount)`) so that `PeerSelect` doesn't blow-up. After that, `PeerSelect` throws `NoPeersToSelect` if there are no selectable peers.

The `additionalDiscoveryDelay` is being reduced to 3 minutes after talking with @ryle-goehausen .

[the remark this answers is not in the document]
marcinwadon: Ok, why din't you add those 3 minutes to either first or second "wait for peers"? You'd wait the same amount of time but eod you avoid blocking

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/daemon/DelayForPeers.scala
cngo-github: A "selectable" peer is a responsive peer in the `Ready` state.

marcinwadon: ok, makes sense :+1:

modules/core/src/main/scala/org/tessellation/domain/snapshot/programs/Download.scala
cngo-github: The purpose of `waitForPeers(upperSelectablePeersCountThreshold)` was to simply delay for 10 minutes. It has been replaced by `Async[F].sleep(additionalDiscoveryDelay)`.

modules/sdk/src/main/scala/org/tessellation/sdk/http/routes/ClusterRoutes.scala
cngo-github: This is no longer relevant.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/daemon/SelectablePeerDiscoveryDelay.scala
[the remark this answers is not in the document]
cngo-github: `waitForPeers`, the private method, does not necessarily wait the full 10 minutes. It waits until the specified number of selectable peers is discovered or 10 minutes, whichever comes first. It checks for selectable peers every 1 minute. So, in theory, it can finish right away, but it won't delay longer than 10 minutes. The `additionalDiscoveryDelay` is meant to wait for the full period of time. I also realized that `waitForPeers` is the name of both the public and private method. I've renamed the private method to `waitUntil`.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/daemon/SelectablePeerDiscoveryDelay.scala
cngo-github: Changed the logged message as requested.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/PeerDiscoveryDelay.scala
cngo-github: `checkPeersAttemptDelay`, `checkPeersMaxDelay`, `additionalDiscoveryDelay`, and `minPeersCount` have been added as parameters of the `make` and their values added to `AppConfig`.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/daemon/SelectablePeerDiscoveryDelay.scala
cngo-github: Removed the `override`.

modules/core/src/main/scala/org/tessellation/domain/snapshot/programs/Download.scala
cngo-github: It is not needed and has been removed.

modules/core/src/main/scala/org/tessellation/domain/snapshot/programs/Download.scala
cngo-github: Changed to

```
    def waitForPeers: F[Unit] = {
      val peers = if (appEnvironment === Dev) {
        waitUntil(fallbackPeersCount)
      } else {
        waitUntil(minPeersCount) >>
          Async[F].sleep(additionalDiscoveryDelay) >>
          waitUntil(fallbackPeersCount)
      }

      peers.flatMap {
        case s if s.nonEmpty => ().pure
        case _               => NoPeersToSelect.raiseError[F, Unit]
      }
    }
```

marcinwadon: @cngo-github what has been changed? I see no logical difference between this and the previous code, still `waitUntil(fallbackPeersCount)` is executed even if `minPeersCount` succeeded. Also, `sleep` is executed always, even if `minPeersCount` has been reached immediately.

modules/core/src/main/scala/org/tessellation/domain/snapshot/programs/Download.scala
marcinwadon: but `waitForPeers` didn't simply delay for 10 minutes but also check for the number of peers and continued if reached

modules/core/src/main/scala/org/tessellation/cli/method.scala
```diff
@@ -38,6 +38,14 @@ object method {
       environment = environment,
       http = httpConfig,
       db = dbConfig,
+      download = Download(
+        PeerDiscoveryDelay(
+          checkPeersAttemptDelay = 1.minute,
+          checkPeersMaxDelay = 10.minutes,
+          additionalDiscoveryDelay = 3.minutes,
+          minPeersCount = 2
+        )
```
marcinwadon, line 47: peer discovery delay should be separate from the download (as we discussed to take it off the download and make it uncoupled)

modules/core/src/main/scala/org/tessellation/modules/Daemons.scala
```diff
@@ -25,15 +25,26 @@ object Daemons {
     healthChecks: HealthChecks[F],
     nodeId: PeerId,
     cfg: AppConfig
-  ): F[Unit] =
+  ): F[Unit] = {
+    val c = cfg.download.peerDiscoveryDelay
```
marcinwadon, line 29: can you use a better name for this? it's not within a simple function scope, so it would be better to name it

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/daemon/DownloadDaemon.scala
```diff
@@ -3,25 +3,34 @@ package org.tessellation.sdk.infrastructure.snapshot.daemon
...
 
 trait DownloadDaemon[F[_]] extends Daemon[F] {}
 
 object DownloadDaemon {
 
-  def make[F[_]: Async](nodeStorage: NodeStorage[F], download: Download[F])(implicit S: Supervisor[F]) = new DownloadDaemon[F] {
+  def make[F[_]: Async](
+    nodeStorage: NodeStorage[F],
+    download: Download[F],
+    downloadDelay: PeerDiscoveryDelay[F]
```
marcinwadon, line 22: it's a peer discovery delay (executed before download). I'd not name it `download` to keep it uncoupled

modules/core/src/main/scala/org/tessellation/domain/snapshot/programs/Download.scala
cngo-github: It has been replaced by `Async[F].sleep(additionalDiscoveryDelay)`.

modules/core/src/main/scala/org/tessellation/modules/Daemons.scala
cngo-github: Renamed to `pddConfig`.

modules/core/src/main/scala/org/tessellation/cli/method.scala
cngo-github: Moved `PeerDiscoveryDelay` out of `Download`.

modules/core/src/main/scala/org/tessellation/domain/snapshot/programs/Download.scala
cngo-github: @marcinwadon , I believe that according to our offline discussion, this is no longer an issue.

marcinwadon (dismissed):

modules/core/src/main/scala/org/tessellation/domain/snapshot/programs/Download.scala
marcinwadon: IMO the case is still valid. The additional delay should be there as we discussed, but `fallbackPeersCount` will be executed one way or another, but IMO it shouldn't if the first `waitUntil` succeeds and there are more peers already found

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/daemon/SelectablePeerDiscoveryDelay.scala
```diff
@@ -0,0 +1,69 @@
...
+        )
+    )(clusterStorage.getResponsivePeers.map(_.filter(_.state === Ready)))
+
+    def waitForPeers: F[Unit] = {
+      val peers = if (appEnvironment === Dev) {
+        waitUntil(fallbackPeersCount)
+      } else {
+        waitUntil(minPeersCount) >>
+          Async[F].sleep(additionalDiscoveryDelay) >>
+          waitUntil(fallbackPeersCount)
```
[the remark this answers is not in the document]
cngo-github: So, the purpose of

```
waitUntil(minPeersCount) >>
          Async[F].sleep(additionalDiscoveryDelay)
```

is simply to wait for more than 1 peer. Since we don't care about the exact peer count, that has been simplified to just a single delay. Afterwards, we check for 1 peer since `PeerSelect` needs at least 1 peer.

TLDR:

The function has been simplified to

```
    def waitForPeers: F[Unit] =
      Async[F].sleep(additionalDiscoveryDelay).unlessA(appEnvironment === Dev) >>
        waitUntil(fallbackPeersCount).flatMap {
          case s if s.nonEmpty => ().pure
          case _               => NoPeersToSelect.raiseError[F, Unit]
        }
```

modules/core/src/main/scala/org/tessellation/domain/snapshot/programs/Download.scala
cngo-github: So, the purpose of

```
waitUntil(minPeersCount) >>
          Async[F].sleep(additionalDiscoveryDelay)
```

is simply to wait for more than 1 peer. Since we don't care about the exact peer count, that has been simplified to just a single delay. Afterwards, we check for 1 peer since `PeerSelect` needs at least 1 peer.

TLDR:

The function has been simplified to

```
    def waitForPeers: F[Unit] =
      Async[F].sleep(additionalDiscoveryDelay).unlessA(appEnvironment === Dev) >>
        waitUntil(fallbackPeersCount).flatMap {
          case s if s.nonEmpty => ().pure
          case _               => NoPeersToSelect.raiseError[F, Unit]
        }
```

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/daemon/SelectablePeerDiscoveryDelay.scala
```diff
@@ -0,0 +1,71 @@
...
+        )
+    )(clusterStorage.getResponsivePeers.map(_.filter(_.state === Ready)))
+
+    def waitForPeers: F[Unit] = {
+      val peers =
+        if (appEnvironment === Dev)
+          waitUntil(fallbackPeersCount)
+        else
+          waitUntil(minPeers).flatMap {
+            case s if s.size >= minPeers => Async[F].sleep(additionalDiscoveryDelay).as(s)
```
marcinwadon, line 59: `Temporal[F].sleep`

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/daemon/SelectablePeerDiscoveryDelay.scala
```diff
@@ -0,0 +1,71 @@
...
+
+object SelectablePeerDiscoveryDelay {
+
+  def make[F[_]: Async](
+    clusterStorage: ClusterStorage[F],
+    appEnvironment: AppEnvironment,
+    checkPeersAttemptDelay: FiniteDuration,
+    checkPeersMaxDelay: FiniteDuration,
+    additionalDiscoveryDelay: FiniteDuration,
+    minPeers: PosInt
```
marcinwadon, line 34: so, I can pass `minPeers = 10` but the function `waitForPeers` would succeed without any error if it would find 1 peer only. is that correct behavior?

cngo-github: That is the correct behavior. On `Dev` environments, the delay only waits for 1 peer. On non-`Dev` environments, it waits for `minPeers` for up to 10 minutes. If `minPeers` peers are found, an additional 3 minute delay is executed. Otherwise, a 3 minute delay is executed followed by waiting for 1 peer for up to 10 minutes.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/daemon/SelectablePeerDiscoveryDelay.scala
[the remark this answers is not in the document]
cngo-github: Per our discussion offline, this has been re-factored ro

```
    def waitForPeers: F[Unit] = {
      val peers =
        if (appEnvironment === Dev)
          waitUntil(fallbackPeersCount)
        else
          waitUntil(minPeers).flatMap {
            case s if s.size >= minPeers => Temporal[F].sleep(additionalDiscoveryDelay).as(s)
            case _ =>
              Temporal[F].sleep(additionalDiscoveryDelay) >> waitUntil(fallbackPeersCount)
          }

      peers.flatMap {
        case s if s.nonEmpty => ().pure
        case _               => NoPeersToSelect.raiseError[F, Unit]
      }
    }
```

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/daemon/SelectablePeerDiscoveryDelay.scala
cngo-github: Changed to `Temporal[F].sleep`.

sdavidp (commented):
LGTM. Left one suggestion

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/daemon/SelectablePeerDiscoveryDelay.scala
```diff
@@ -0,0 +1,71 @@
...
+          waitUntil(minPeers).flatMap {
+            case s if s.size >= minPeers => Temporal[F].sleep(additionalDiscoveryDelay).as(s)
+            case _ =>
+              Temporal[F].sleep(additionalDiscoveryDelay) >> waitUntil(fallbackPeersCount)
+          }
+
+      peers.flatMap {
+        case s if s.nonEmpty => ().pure
+        case _               => NoPeersToSelect.raiseError[F, Unit]
+      }
```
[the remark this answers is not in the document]
cngo-github: Changed to

```
peers.flatMap(s => NoPeersToSelect.raiseError[F, Unit].whenA(s.isEmpty))
```

sdavidp (approved):

PR #721, opened 2023-09-04
Persist genesis and first incremental during download

TheMMaciek:


modules/core/src/main/scala/org/tessellation/domain/snapshot/storages/SnapshotDownloadStorage.scala
```diff
@@ -19,5 +19,7 @@ trait SnapshotDownloadStorage[F[_]] {
   def moveTmpToPersisted(snapshot: Signed[GlobalIncrementalSnapshot]): F[Unit]
 
   def readGenesis(ordinal: SnapshotOrdinal): F[Option[Signed[GlobalSnapshot]]]
+  def writeGenesis(genesis: Signed[GlobalSnapshot]): F[Unit]
```
marcinwadon, line 22: IMO this should be done once only. If there is already a genesis stored and it's different than the network genesis, then the process should stop and user should manually remove the genesis. It will prevent human error joining to a different network or sth

marcinwadon (approved):
Looks good. I left one comment to consider about a genesis snapshot.

modules/core/src/main/scala/org/tessellation/domain/snapshot/storages/SnapshotDownloadStorage.scala
TheMMaciek: It's not in plain sight here, but writing a "second" genesis actually will throw an error because of the repeated ordinal. Not sure if that's enough but seems to be given that we specify the genesis's ordinal. We could also change the tactic to what I proposed some time ago - if you want to join a network you need to supply the correct genesis manually yourself, that way there will be no possibillity of overwriting it and you will be sure that you are joining the network with the same genesis. We won't have the genesis download part in the logic then as well.

PR #726, opened 2023-09-14
Recursively limit proposal size by returning excess blocks back to pool

kpudlik:


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotCreator.scala
```diff
@@ -76,116 +80,142 @@ object CurrencySnapshotCreator {
...
-          acceptanceResult,
-          currentOrdinal
-        )
+          maybeLastDataState <- (maybeLastDataApplication, maybeDataApplicationService).mapN {
+            case (lastDataApplication, (_, service)) =>
+              service
+                .deserializeState(lastDataApplication)
+                .flatTap {
+                  case Left(err) => logger.warn(err)(s"Cannot deserialize custom state")
+                  case Right(a)  => logger.info(s"Deserialized state: $a")
```
marcinwadon, line 106: do we want to log the whole deserialized state? especially as `info`?

marcinwadon: I see now it was in the code already. but just spotted this and probably we should remove this

marcinwadon (dismissed):
lgtm, although I'm not sure we should limit the number of signatures 🤔 anyone can add few more signatures and the snapshot would be incorrect then 🤔

kpudlik:
> lgtm, although I'm not sure we should limit the number of signatures 🤔 anyone can add few more signatures and the snapshot would be incorrect then 🤔

In 100 kB we should fit ~300 signatures. Maybe we should confront the config with the loaded whitelist in the following way:

```
if (whitelistedNodesCount * 0,33kB >= 100 kB) { throw new Exception("Signatures may potentially exceed the size limit. Increase the ratio between signatures size and data size in proposal")
```

modules/sdk/src/main/scala/org/tessellation/sdk/cli/CliMethod.scala
```diff
@@ -86,6 +87,10 @@ trait CliMethod {
       ensureCheckInterval = 10.seconds
     )
   )
+  val snapshotSizeConfig: SnapshotSizeConfig = SnapshotSizeConfig(
+    maxSignaturesSizeInBytes = 102_400L,
```
TheMMaciek, line 91: Why can't we derive the maximum size of a single signature and then calculate how much they will take based on the number of facilitators after facility step ends? After it ends there can't be any more facilitators added. Hardcoding to 100kb or 300 signers seems like wasting possibly available space.

marcinwadon: IMO forbidding to sign the snapshot by additional peer is no-go and this is what you and Krystian proposes 🤔

TheMMaciek: The maximum amount of signatures is known during consensus and no more will be accepted anyway, so leaving it open to accept more signatures given that it will never be utilized is pointless. The count of facilitators calculated at the end of `Facility` step is effectively the max count of signatures for that consensus round and in consequence a snapshot.

marcinwadon: ok, I got your point and I agree with you. since we know the maximum possible number of peers that would sign the proposal, we can limit the size of signatures to this and not to the maximum possible number taken from seedlist/allowance list.

kpudlik: I agree. I checked the size of single `SignatureProof` and I multiply it by facilitators number from `Facility` step. So the max cap depends on number of facilitators. We can of course have less because some facilitators may fail at next steps but at least we won't waste size in a hardcoded way.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotCreator.scala
kpudlik: What should I do then? Change the log-level or remove that line completely?

marcinwadon: It's the whole deserialized state so I'd remove this log completly

marcinwadon (dismissed):

modules/shared/src/test/scala/org/tessellation/schema/generators.scala
```diff
@@ -118,7 +121,11 @@ object generators {
       txnSalt <- transactionSaltGen
     } yield Transaction(src, dst, txnAmount, txnFee, txnReference, txnSalt)
 
-  val signatureGen: Gen[Signature] = nesGen(str => Signature(Hex(str)))
+  val signatureGen: Gen[Signature] =
+    Gen
+      .listOfN(142, Gen.alphaChar)
```
TheMMaciek, line 126: Are we sure that the signature is always 142 chars long? I remember that sometimes signatures seemed to have different length.

Now that I quickly examine the snapshot signatures some of them are longer e.g. 144

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotCreatorSuite.scala
```diff
@@ -0,0 +1,28 @@
...
+import weaver.scalacheck.Checkers
+
+object CurrencySnapshotCreatorSuite extends MutableIOSuite with Checkers {
+  type Res = KryoSerializer[IO]
+
+  def sharedResource: Resource[IO, KryoSerializer[IO]] =
+    KryoSerializer.forAsync[IO](sdkKryoRegistrar)
+
+  test("single signature size matches the size limit") { implicit kryo =>
+    forall(signatureProofGen) { proof =>
```
TheMMaciek, line 21: I don't think you can really prove the size always matches if you are mocking the values. In reality the signature isn't always 142 chars long.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotCreator.scala
```diff
@@ -76,111 +86,144 @@ object CurrencySnapshotCreator {
...
+                }
+                .map(_.toOption)
+                .handleErrorWith(err =>
+                  logger.error(err)(s"Unhandled exception during deserialization data application, fallback to empty state") >>
+                    none[DataState].pure[F]
+                )
+          }.flatSequence
+
+          maybeNewDataState <- (maybeDataApplicationService, maybeLastDataState).mapN {
+            case (((nodeContext, service), lastState)) =>
```
TheMMaciek, line 124: Why additional () here?

modules/shared/src/test/scala/org/tessellation/schema/generators.scala
kpudlik: Interesting. I checked many signatures and it was always `142`. If the signature size is not fixed then the whole signature size prediction makes no sense. Eventually we can take `max(allSignatureSizes)` 🤷🏻‍♂️

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotCreatorSuite.scala
kpudlik: Yeah. I would need to write an effectful generator to actually sign snapshots which isn't that trivial and needs some workarounds.

kpudlik: https://stackoverflow.com/questions/52122705/java-ecdsawithsha256-signature-with-inconsistent-length

We use `val defaultSignFunc = "SHA512withECDSA"` which uses `ASN.1 DER` (variable length) encoding. I checked snapshots and signatures vary between `140` and `144` chars so for now I would suggest setting `144` and then trying `val defaultSignFunc = "SHA512withPLAIN-ECDSA"` taking backward compatibility into account.

modules/sdk/src/main/scala/org/tessellation/sdk/cli/CliMethod.scala
```diff
@@ -16,31 +16,14 @@ import org.tessellation.sdk.PriorityPeerIds
...
-  val trustRatingsPath: Option[Path]
-
-  val httpConfig: HttpConfig
-
-  val stateAfterJoining: NodeState
-
-  val collateralAmount: Option[Amount]
+object CliMethod {
+  val snapshotSizeConfig: SnapshotSizeConfig = SnapshotSizeConfig(
+    singleSignatureSizeInBytes = 296L,
```
TheMMaciek, line 24: How is that value calculated?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -269,7 +270,8 @@ object ConsensusStateUpdater {
                     state.lastOutcome.status.context,
                     majorityTrigger,
                     resources,
-                    scoredProposalHashes
+                    scoredProposalHashes,
+                    candidates
```
TheMMaciek, line 274: Why `candidates` rather than `facilitators`?

kpudlik: It has been already named `candidates` in `ConsensusStateUpdater`by Tomek. I just passed it to the function.

TheMMaciek: I see that it's passed and I think it may be the wrong value, didn't you mean `state.facilitators` here

kpudlik: Good catch as `candidates` is just a diff for nodes we should add to the consensus can be empty. I will change it to `state.facilitators`.

modules/sdk/src/main/scala/org/tessellation/sdk/config/types.scala
```diff
@@ -23,9 +23,15 @@ object types {
...
     trustStorage: TrustStorageConfig,
-    priorityPeerIds: Option[NonEmptySet[PeerId]]
+    priorityPeerIds: Option[NonEmptySet[PeerId]],
+    snapshotSizeConfig: SnapshotSizeConfig
   )
 
+  case class SnapshotSizeConfig(
+                                 singleSignatureProofSize: PosLong,
+                                 maxStateChannelSnapshotBinarySizeInBytes: PosLong
+  ) {}
```
TheMMaciek, line 33: Formatting

TheMMaciek: And no need for `{}`

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotCreator.scala
```diff
@@ -76,111 +91,146 @@ object CurrencySnapshotCreator {
...
+                  service
+                    .validateData(lastState, updates)
+                    .flatTap { validated =>
+                      logger.warn(s"Data application is invalid, errors: ${validated.toString}").unlessA(validated.isValid)
+                    }
+                    .map(_.isValid)
+                    .handleErrorWith(err =>
+                      logger.error(err)(s"Unhandled exception during validating data application, assumed as invalid").as(false)
+                    )
+                    .ifM(
```
TheMMaciek, line 144: Aren't you changing semantics here? I mean previously if the data was invalid you did run the combine method but with empty list of updates. Now you will not run it at all if I'm not mistaken. I'm not sure if the combine method is not incrementing something or performing some other update even though the list of updates is empty e.g. which we are preventing right now.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotCreator.scala
```diff
@@ -76,111 +91,146 @@ object CurrencySnapshotCreator {
...
+            if (size <= maxUnsignedProposalSizeInBytes) {
+              CurrencySnapshotCreationResult(
+                artifact,
+                CurrencySnapshotContext(lastContext.address, snapshotInfo),
+                awaitingBlocks.toSet ++ sizeExcessiveBlocks,
+                rejectedBlocks.toSet
+              ).pure[F]
+            } else {
+              acceptanceResult.accepted match {
+                case _ :+ _ :+ excessivePair =>
```
TheMMaciek, line 218: Why not:
```
case acceptedBlocks :+ excessivePair =>
  excessivePair match {
    case (excessiveBlock, _) =>
      val remainingBlocks = acceptedBlocks.map(_._1)
      val excessiveBlocks = sizeExcessiveBlocks + excessiveBlock
      createProposalWithSizeLimit(remainingBlocks, excessiveBlocks)
  }
```
do we consider creating a snapshot without a single event an unacceptable case?

If so we could still pass the currently accepted except the last one, maybe like this:
```
case acceptedBlocks @ (_ :+ _) :+ excessivePair =>
```

I'm not considering it the essential update, just rerunning with possibly ignored blocks we can have in `blocksForAcceptance` seems wasteful.

modules/shared/src/test/scala/org/tessellation/schema/generators.scala
```diff
@@ -46,7 +46,10 @@ object generators {
     Gen.oneOf(Responsive, Unresponsive)
 
   val idGen: Gen[Id] =
-    nesGen(str => Id(Hex(str)))
+    Gen
+      .listOfN(128, Gen.alphaChar)
+      .map(_.mkString)
+      .map(str => Id(Hex(str)))
```
sdavidp, line 52: I made this suggestion in another PR -- I believe `listOfN` can generate an empty list, which is what `nesGen` avoids. Some tests may break if an empty string is used.

What about a proper `Hex` generator? Then we can implement others in terms of it:
```
  
  def hexGen(n: Int): Gen[Hex] = Gen.stringOfN(n, Gen.hexChar).map(Hex(_))

  val peerIdGen: Gen[PeerId] = hexGen(128).map(PeerId(_))
  val idGen: Gen[Id] = hexGen(128).map(Id(_))
  val signatureGen: Gen[Signature] = Gen.chooseNum(140, 144).flatMap(hexGen).map(Signature(_))
```

PR #727, opened 2023-09-19
Add json with brotli compression serializer

TheMMaciek:
A PR introducing the compression for metagraph snapshot binary. Here the implementation with Brotli algorithm. There is a second PR built on top of this one changing the compression algorithm to Gzip.

The reason there is a second one is that there is one caveat with Brotli in java.
The library used in this PR namely brotli4j (https://github.com/hyperxpro/Brotli4j) is the only one I found providing encoding/compressing functionality with Brotli. The way it works though is by using native brotli implementation provided by the host platform and by using JNI (Java Native Interface). The implementation works on macOS for sure and is tested by the library owner on other platforms like Ubuntu 18.04 or Windows Server 2022. We could run some cross platform test with GitHub Actions to check the compatibility with selected platforms and if the compatibility looks widespread enough we could go with this implementation I think. If the compatibility is not sufficient we can go with Gzip (I would then locally squash both commits into one and then merge a single commit).

In general Brotli is much more performant than Gzip, with compression time being much lower and ratio being higher. So from the performance point of view it's better to use Brotli given that it seems widely supported in other popular programming languages. The only question is if it's safe and compatible to a sufficient degree with mutliple platforms to use it.

### DIFF
--- a/modules/shared/src/main/scala/org/tessellation/json/JsonBrotliBinarySerializer.scala
+++ b/modules/shared/src/main/scala/org/tessellation/json/JsonBrotliBinarySerializer.scala
@@ -1,0 +1,36 @@
+package org.tessellation.json
+
+import cats.effect.kernel.Sync
+import cats.syntax.functor._
+
+import com.aayushatharva.brotli4j.Brotli4jLoader
+import com.aayushatharva.brotli4j.decoder.{Decoder => BrotliDecoder}
+import com.aayushatharva.brotli4j.encoder.Encoder.Parameters
+import com.aayushatharva.brotli4j.encoder.{Encoder => BrotliEncoder}
+import io.circe.jawn.JawnParser
+import io.circe.syntax._
+import io.circe.{Decoder, Encoder}
+
+trait JsonBrotliBinarySerializer[F[_]] {
+  def serialize[A: Encoder](content: A): F[Array[Byte]]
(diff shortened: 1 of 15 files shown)


modules/shared/src/main/scala/org/tessellation/json/JsonBrotliBinarySerializer.scala
```diff
@@ -0,0 +1,36 @@
...
+trait JsonBrotliBinarySerializer[F[_]] {
+  def serialize[A: Encoder](content: A): F[Array[Byte]]
+  def deserialize[A: Decoder](content: Array[Byte]): F[Either[Throwable, A]]
+}
+
+object JsonBrotliBinarySerializer {
+  private val compressionLevel = 2
+
+  def make[F[_]: Sync](): F[JsonBrotliBinarySerializer[F]] =
+    Sync[F].delay(Brotli4jLoader.ensureAvailability()).map { _ =>
```
TheMMaciek, line 23: This is the part were availability of native brotli implementation is checked.

TheMMaciek: AFAIK it can throw if the native library is not available.

marcinwadon (approved):

marcinwadon:
We've internally tested this with latest 2 versions of ubuntu and agreed we'd like to merge it. If needed, we'll make more community tests.

PR #728, opened 2023-09-19
Replace brotli compression with gzip

TheMMaciek:


modules/currency-l0/src/main/scala/org/tessellation/currency/l0/modules/Programs.scala
```diff
@@ -17,9 +17,11 @@ import org.tessellation.sdk.infrastructure.snapshot.{CurrencySnapshotContextFunc
 import org.tessellation.sdk.modules.SdkPrograms
 import org.tessellation.security.SecurityProvider
 
+import fs2.io.file.Files
+
 object Programs {
 
-  def make[F[_]: Async: Random: KryoSerializer: SecurityProvider](
+  def make[F[_]: Async: Random: KryoSerializer: SecurityProvider: Files](
```
TheMMaciek, line 24: FYI adding `FIles` and `Network` capabilities is required because of fs2 version being bumped up implicitly by the `fs2-compression` library it's a minor version update so it seems it should be fine. Though maybe it would be better to perform an update in a separate PR if we will go with this change. We could also use earlier version of `fs2-compression` which uses a fs2 not requiring these udpates yet.

marcinwadon:
We internally agreed to try brotli, so I'm closing the PR

PR #731, opened 2023-09-25
Use trust scores in peer declarations selection

cngo-github:
Use trust scores in peer declarations selection

### DIFF
--- a/modules/core/src/main/scala/org/tessellation/infrastructure/trust/storage/TrustStorage.scala
+++ b/modules/core/src/main/scala/org/tessellation/infrastructure/trust/storage/TrustStorage.scala
@@ -13,6 +13,7 @@ import org.tessellation.schema.SnapshotOrdinal
 import org.tessellation.schema.peer.PeerId
 import org.tessellation.schema.trust._
 import org.tessellation.sdk.config.types.TrustStorageConfig
+import org.tessellation.sdk.domain.seedlist.SeedlistEntry
 import org.tessellation.sdk.domain.trust.storage._
 
 import derevo.circe.magnolia.encoder
@@ -30,13 +31,13 @@ object TrustStorage {
 
   def make[F[_]: Monad: Ref.Make](
     config: TrustStorageConfig,
-    seedlist: Option[Set[PeerId]]
+    seedlist: Option[Set[SeedlistEntry]]
   ): F[TrustStorage[F]] = make(TrustMap.empty, config, seedlist)
 
   def make[F[_]: Monad: Ref.Make](
     trustInfo: TrustMap,
     config: TrustStorageConfig,
-    seedlist: Option[Set[PeerId]]
+    seedlist: Option[Set[SeedlistEntry]]
   ): F[TrustStorage[F]] =
     Ref[F]
       .of[TrustStore](
@@ -53,16 +54,24 @@ object TrustStorage {
   def make[F[_]: Monad: Ref.Make](
     trustUpdates: Option[PeerObservationAdjustmentUpdateBatch],
     config: TrustStorageConfig,
-    seedlist: Option[Set[PeerId]]
+    seedlist: Option[Set[SeedlistEntry]]
   ): F[TrustStorage[F]] =
     trustUpdates.fold(make(config, seedlist))(batch => make(config, seedlist).flatTap(_.updateTrust(batch)))
 
   def make[F[_]: Monad](
     trustStoreRef: Ref[F, TrustStore],
     config: TrustStorageConfig,
-    seedlist: Option[Set[PeerId]]
+    seedlist: Option[Set[SeedlistEntry]]
   ): TrustStorage[F] =
     new TrustStorage[F] {
+
+      val seedlistScores: Option[Map[PeerId, TrustValueRefined]] =
+        seedlist
+          .map(
+            _.map { e =>
+              e.peerId -> e.bias
+            }.collect { case (peerId, Some(s)) if s.value > 0.0 => peerId -> s }.toMap
+          )
 
       def updateTrust(trustUpdates: PeerObservationAdjustmentUpdateBatch): F[Unit] =
         trustStoreRef.update { store =>
@@ -84,6 +93,26 @@ object TrustStorage {
         }.toMap)
         .map(_.collect { case (peerId, Some(trustScore)) => peerId -> trustScore })
         .map(TrustScores(_))
+
+      def getBiasedSeedlistOrdinalPeerLabels: F[Option[Map[PeerId, Double]]] = getCurrentOrdinalTrust.map { ordinalTrustMap =>
+        seedlistScores.map { scores =>
+          val publicTrusts = ordinalTrustMap.peerLabels.value.flatMap {
+            case (peerId, trust) =>
+              scores.get(peerId).map(s => trust.labels.view.mapValues(_ * s.value).toMap)
+          }
+
+          publicTrusts
+            .foldLeft(Map.empty[PeerId, List[Double]]) {
+              case (acc, pt) =>
+                val normalized = pt.map { case (id, d) => id -> (d :: acc.getOrElse(id, Nil)) }
+
+                acc ++ normalized
+            }
+            .view
+            .mapValues(s => s.sum / s.length)
+            .toMap
+        }
+      }
 
       def updateTrustWithBiases(selfPeerId: PeerId): F[Unit] =
         trustStoreRef.update { store =>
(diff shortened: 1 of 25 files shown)


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/ProposalHashSelect.scala
```diff
@@ -0,0 +1,43 @@
...
+import org.tessellation.schema.trust.{TrustScores, defaultPeerTrustScore}
+import org.tessellation.sdk.domain.snapshot.ProposalSelect
+import org.tessellation.sdk.domain.snapshot.ProposalSelect.occuranceTrustScoreOrdering
+import org.tessellation.sdk.infrastructure.consensus.PeerDeclarations
+import org.tessellation.sdk.infrastructure.snapshot.WeightedSampling.WeightRefinement
+
+import eu.timepit.refined.numeric.Positive
+import eu.timepit.refined.refineV
+
+object ProposalHashSelect {
```
marcinwadon, line 16: shouldn't it be next to the `ProposalSelect`? Will be there any other kind of `ProposalSelect` so you would use the strategy pattern here (unlike the PeerSelect)?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/ProposalHashSelect.scala
```diff
@@ -0,0 +1,43 @@
...
+        }
+        .collect { case (hash, Right(score)) => hash -> score.value }
+        .groupMapReduce { case (hash, _) => hash } { case (_, score) => score }(
+          _ + _
+        )
+        .view
+        .mapValues(refineV[Positive](_))
+        .collect {
+          case (hash, Right(score)) => hash -> score
+        }
```
sdavidp, line 39: Could we not use the unsafe method here since we're guaranteed the value will be positive? E.g. Here's a possible alternate implementation that also makes the newly created _Ordering_ unnecessary.
```
declarations.map {
  case (peerId, pd) =>
    pd.proposal
      .map(_.hash -> trustScores.getOrElse(peerId, defaultPeerTrustScore.value))
      .filter(_._2 > 0.0)
}
  .flatten
  .groupMapReduce { case (hash, _) => hash } { case (_, score) => score }(_ + _)
  .toList
  .sortBy(_._2)(Ordering[Double].reverse)
  .map { case (h, d) => h -> Refined.unsafeApply(d) }
```

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/ProposalHashSelect.scala
```diff
@@ -0,0 +1,43 @@
...
+  ): ProposalSelect[F] = (declarations: Map[PeerId, PeerDeclarations]) =>
+    getCurrentTrust.map(_.scores).map { trustScores =>
+      declarations.view
+        .mapValues(_.proposal)
+        .toList
+        .collect {
+          case (peerId, Some(proposal)) =>
+            val score = trustScores.getOrElse(peerId, defaultPeerTrustScore.value)
+
+            proposal.hash -> refineV[WeightRefinement](score)
```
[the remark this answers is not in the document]
cngo-github: I used the refined type to allow for the reverse sorting since it's ordering was created.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/ProposalHashSelect.scala
cngo-github: Changed to

```
    getCurrentTrust.map(_.scores).map { trustScores =>
      declarations.map {
        case (peerId, declarations) =>
          declarations.proposal.map {
            _.hash -> trustScores.getOrElse(peerId, defaultPeerTrustScore.value)
          }.filter { case (_, s) => s > 0.0 }
      }.flatten.groupMapReduce { case (hash, _) => hash } { case (_, score) => score }(_ + _).toList.sortBy { case (_, score) => score }(
        Order[Double].toOrdering.reverse
      ).map { case (h, s) => h -> Refined.unsafeApply(s) }
```

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/ProposalSelect.scala
```diff
@@ -0,0 +1,26 @@
...
+
+trait ProposalSelect[F[_]] {
+
+  def score(declarations: Map[PeerId, PeerDeclarations]): F[List[(Hash, OccuranceTrustScore)]]
+
+}
+
+object ProposalSelect {
+
+  type OccuranceTrustScore = Double Refined Positive
```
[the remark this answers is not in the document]
cngo-github: Switched to using `PosDouble`.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/ProposalHashSelect.scala
cngo-github: Renamed to `ProposalSelect`.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/ProposalHashSelect.scala
sdavidp: Can the formatting be cleaned up a bit? I find this harder to read than the sample I provided.

cngo-github: What do you recommend?

cngo-github: Per our offline discussion, this is no longer relevant.

modules/core/src/main/scala/org/tessellation/infrastructure/trust/storage/TrustStorage.scala
```diff
@@ -53,17 +54,25 @@ object TrustStorage {
...
-    seedlist: Option[Set[PeerId]]
+    seedlist: Option[Set[SeedlistEntry]]
   ): TrustStorage[F] =
     new TrustStorage[F] {
 
+      val seedlistScores: Option[Map[PeerId, TrustValueRefined]] =
+        seedlist
+          .map(
+            _.filter(_.bias.exists(_.value > 0.0))
+              .map(e => e.peerId -> e.bias.getOrElse(defaultPeerTrustScore))
```
kpudlik, line 72: That `getOrElse` never gets called because you have already filtered all the `None` in line `71`.

If you have such seedlist entries:

```
A 0.0
B 1.0
C -1.0
```

then your code makes such Map:

```
Map(B -> 1.0)
```

If you want to keep values which are `<= 0.0` but just default them to `defaultPeerTrustScore` then you should map over entries and convert all the `<=` to `defaultPeerTrustScore`:

```scala
seedlist.map(
  _.map(_.bias.filter(_.value > 0.0).getOrElse(defaultPeerTrustScore))
)
```

kpudlik (changes requested):

modules/core/src/main/scala/org/tessellation/infrastructure/trust/storage/TrustStorage.scala
cngo-github: Changed to

```
        seedlist
          .map(
            _.map { e =>
              e.peerId -> e.bias.filter(_.value > 0.0).getOrElse(defaultPeerTrustScore)
            }.toMap
          )
```

sdavidp: Doesn't the `getOrElse` mean that the lowest "score" someone can have is the default value? In other words, if the bias is determined to be the lowest value (-1?) then it automatically gets upgraded to the default. Is that what we want?

ryle-goehausen: Right now we're not really using any seed list entry to denote a negative score (i.e. the equivalent of downranking a node due to negative behavior.) The getOrElse here should be supplying a value for all entries on the seed list that otherwise have no score, so the interpretation of this is that we haven't updated old entries in the seed list to have specified scores, so if no information is present they should default to a preset default.

It might be useful to include negative scores in the future (for some sort of slasher like mechanism,) but for now it's not used at all. Still, if a score WAS negative, this function would be incorrect, because we'd just be filling it with a default.

The correct behavior would be to disregard a negative score, take the value from the seedlist if filled, otherwise default to constant.

cngo-github: Changed to

```
        seedlist
          .map(
            _.map { e =>
              e.peerId -> e.bias
            }.collect { case (peerId, Some(s)) if s.value > 0.0 => peerId -> s }.toMap
          )
```

kpudlik (approved):

PR #732, opened 2023-09-25
Add CurrencySnapshotEvents parameter to Rewards.distribute

sdavidp:


modules/sdk/src/main/scala/org/tessellation/sdk/domain/rewards/Rewards.scala
```diff
@@ -7,13 +7,15 @@ import org.tessellation.schema.balance.Balance
...
 import org.tessellation.security.signature.Signed
 
 trait Rewards[F[_], P <: StateProof, S <: IncrementalSnapshot[P]] {
   def distribute(
     lastArtifact: Signed[S],
     lastBalances: SortedMap[Address, Balance],
     acceptedTransactions: SortedSet[Signed[Transaction]],
-    trigger: ConsensusTrigger
+    trigger: ConsensusTrigger,
+    events: Set[CurrencySnapshotEvent]
```
marcinwadon, line 19: looks like it leaks the implementation to the declaration. at this point we know nothing about `S` being a global snapshot or currency snapshot. I think you should parametrize `events` and pass global snapshot events on GL0 and currency snapshot events on Currency L0

marcinwadon (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/domain/rewards/Rewards.scala
sdavidp: @ryle-goehausen ☝️ FYI

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -120,7 +120,7 @@ object GlobalSnapshotConsensusFunctions {
             snapshotContext,
             lastActiveTips,
             lastDeprecatedTips,
-            rewards.distribute(lastArtifact, snapshotContext.balances, _, trigger),
+            rewards.distribute(lastArtifact, snapshotContext.balances, _, trigger, Set.empty),
```
marcinwadon, line 123: imo you should pass `events` here, not the `Set.empty`

marcinwadon (changes requested):

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
sdavidp: I think @ryle-goehausen asked me to pass empty set...

marcinwadon: isn't it misleading that I expect it to be a set of events and I got an empty list? seems to be a bug 🤔

ryle-goehausen: Right now the only purpose of this event set is for use in an application rewards function -- we're not calculating any rewards from any other events yet -- if in the future we do so we can pass something here.

ryle-goehausen: If this events is easily in scope here however, there's no problem passing it so long as it's not actually affecting any code. I was fine with empty set when we were talking about the other object typed differently. Here now after updating the types its fine to pass in (so long as we aren't using it.)

sdavidp: Done

ryle-goehausen (approved):
requested changes implemented.

marcinwadon (approved):

PR #733, opened 2023-09-26
Always call combine when creating proposal artifact

sdavidp:
It's expected that the metagraph implementations modify their implementation of `combine` to support receiving `List` instead of `NonEmptyList`


ryle-goehausen (approved):
LGTM, this is a requirement for the dor metagraph since it's storing the incremental data updates per snapshot in the state object since not present in the currency snapshot.

scasplte2 (approved):
🤙 looks good

PR #736, opened 2023-10-05
Added proposal trust select to other layers.

cngo-github:
Added proposal trust select to other (non-core) layers (i.e., currency).

### DIFF
--- a/modules/core/src/main/scala/org/tessellation/infrastructure/trust/storage/TrustStorage.scala
+++ b/modules/core/src/main/scala/org/tessellation/infrastructure/trust/storage/TrustStorage.scala
@@ -31,13 +31,13 @@ object TrustStorage {
 
   def make[F[_]: Monad: Ref.Make](
     config: TrustStorageConfig,
-    seedlist: Option[Set[SeedlistEntry]]
+    seedlist: Set[SeedlistEntry]
   ): F[TrustStorage[F]] = make(TrustMap.empty, config, seedlist)
 
   def make[F[_]: Monad: Ref.Make](
     trustInfo: TrustMap,
     config: TrustStorageConfig,
-    seedlist: Option[Set[SeedlistEntry]]
+    seedlist: Set[SeedlistEntry]
   ): F[TrustStorage[F]] =
     Ref[F]
       .of[TrustStore](
@@ -54,24 +54,21 @@ object TrustStorage {
   def make[F[_]: Monad: Ref.Make](
     trustUpdates: Option[PeerObservationAdjustmentUpdateBatch],
     config: TrustStorageConfig,
-    seedlist: Option[Set[SeedlistEntry]]
+    seedlist: Set[SeedlistEntry]
   ): F[TrustStorage[F]] =
     trustUpdates.fold(make(config, seedlist))(batch => make(config, seedlist).flatTap(_.updateTrust(batch)))
 
   def make[F[_]: Monad](
     trustStoreRef: Ref[F, TrustStore],
     config: TrustStorageConfig,
-    seedlist: Option[Set[SeedlistEntry]]
+    seedlist: Set[SeedlistEntry]
   ): TrustStorage[F] =
     new TrustStorage[F] {
 
-      val seedlistScores: Option[Map[PeerId, TrustValueRefined]] =
-        seedlist
-          .map(
-            _.map { e =>
-              e.peerId -> e.bias
-            }.collect { case (peerId, Some(s)) if s.value > 0.0 => peerId -> s }.toMap
-          )
+      val seedlistScores: Map[PeerId, TrustValueRefined] =
+        seedlist.map { e =>
+          e.peerId -> e.bias
+        }.collect { case (peerId, Some(s)) if s.value > 0.0 => peerId -> s }.toMap
 
       def updateTrust(trustUpdates: PeerObservationAdjustmentUpdateBatch): F[Unit] =
         trustStoreRef.update { store =>
@@ -94,25 +91,24 @@ object TrustStorage {
         .map(_.collect { case (peerId, Some(trustScore)) => peerId -> trustScore })
         .map(TrustScores(_))
 
-      def getBiasedSeedlistOrdinalPeerLabels: F[Option[Map[PeerId, Double]]] = getCurrentOrdinalTrust.map { ordinalTrustMap =>
-        seedlistScores.map { scores =>
-          val publicTrusts = ordinalTrustMap.peerLabels.value.flatMap {
-            case (peerId, trust) =>
-              scores.get(peerId).map(s => trust.labels.view.mapValues(_ * s.value).toMap)
-          }
-
-          publicTrusts
-            .foldLeft(Map.empty[PeerId, List[Double]]) {
-              case (acc, pt) =>
-                val normalized = pt.map { case (id, d) => id -> (d :: acc.getOrElse(id, Nil)) }
-
-                acc ++ normalized
-            }
-            .view
-            .mapValues(s => s.sum / s.length)
-            .toMap
+      def getBiasedSeedlistOrdinalPeerLabels: F[TrustLabels] = getCurrentOrdinalTrust.map { ordinalTrustMap =>
+        val publicTrusts = ordinalTrustMap.peerLabels.value.flatMap {
+          case (peerId, trust) =>
+            seedlistScores.get(peerId).map(s => trust.labels.view.mapValues(_ * s.value).toMap)
         }
+
+        publicTrusts
+          .foldLeft(Map.empty[PeerId, List[Double]]) {
+            case (acc, pt) =>
+              val normalized = pt.map { case (id, d) => id -> (d :: acc.getOrElse(id, Nil)) }
+
+              acc ++ normalized
+          }
+          .view
+          .mapValues(s => s.sum / s.length)
+          .toMap
       }
+        .map(TrustLabels(_))
 
       def updateTrustWithBiases(selfPeerId: PeerId): F[Unit] =
         trustStoreRef.update { store =>
@@ -205,23 +201,20 @@ object TrustStorage {
           } else store
         }
 
-      private def getBiasedTrust(trust: TrustMap): TrustMap =
-        seedlistScores.fold(trust) { e =>
-          val peerIds = e.keys
+      private def getBiasedTrust(trust: TrustMap): TrustMap = {
+        val peerIds = seedlistScores.keys
 
-          val updatedTrust = peerIds
-            .filterNot(trust.trust.contains)
-            .map(_ -> TrustInfo(predictedTrust = config.seedlistInputBias.some))
-            .toMap
-          val updatedPeerLabels = peerIds
-            .filterNot(trust.peerLabels.value.contains)
-            .map(_ -> PublicTrust.empty)
-            .toMap
+        val updatedTrust = peerIds
+          .filterNot(trust.trust.contains)
+          .map(_ -> TrustInfo(predictedTrust = config.seedlistInputBias.some))
+        val updatedPeerLabels = peerIds
+          .filterNot(trust.peerLabels.value.contains)
+          .map(_ -> PublicTrust.empty)
 
-          TrustMap(
-            trust.trust ++ updatedTrust,
-            PublicTrustMap(trust.peerLabels.value ++ updatedPeerLabels)
-          )
-        }
+        TrustMap(
+          trust.trust ++ updatedTrust,
+          PublicTrustMap(trust.peerLabels.value ++ updatedPeerLabels)
+        )
+      }
     }
 }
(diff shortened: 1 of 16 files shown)


modules/core/src/main/scala/org/tessellation/infrastructure/trust/storage/TrustStorage.scala
```diff
@@ -113,6 +113,8 @@ object TrustStorage {
             .toMap
         }
       }
+        .map(_.getOrElse(Map.empty[PeerId, Double]))
```
sdavidp, line 116: It looks like there's no longer a distinction `seedlist` being `None` vs `Some(empty set)`
If so we should refactor

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/modules/Services.scala
```diff
@@ -81,7 +84,19 @@ object Services {
         maybeDataApplication
       )
 
-      proposalSelect = ProposalOccurrenceSelect.make()
+      getTrusts = Applicative[F].product(
+        p2PClient.l0Trust.getPreviousSeedlistPeerLabels.run(cfg.globalL0Peer),
+        p2PClient.l0Trust.getCurrentTrust.run(cfg.globalL0Peer)
```
sdavidp, line 89: Move into the following `else` since it's only. used there?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/ProposalTrustSelect.scala
```diff
@@ -19,19 +18,18 @@ import eu.timepit.refined.numeric.Positive
...
-      relative = r.trust.view
-        .mapValues(_.predictedTrust)
-        .collect { case (peerId, Some(s)) if s > 0 && declarations.contains(peerId) => peerId -> s }
-        .toMap
+      deterministic = d.value.filter {
+        case (peerId, s) => s > 0 && declarations.contains(peerId)
+      }
+
+      relative = r.value.collect {
+        case (peerId, s) if s > 0.0 && declarations.contains(peerId) => peerId -> s
```
sdavidp, line 31: Isn't this also a "filter" ?
If so you can make it a function and clean up its two uses.

modules/core/src/main/scala/org/tessellation/modules/Services.scala
```diff
@@ -65,7 +65,10 @@ object Services {
         .pure[F]
 
       trustStorage = storages.trust
-      getTrusts = Applicative[F].product(trustStorage.getBiasedSeedlistOrdinalPeerLabels, trustStorage.getTrust)
+      getTrusts = Applicative[F].product(
+        trustStorage.getBiasedSeedlistOrdinalPeerLabels,
+        trustStorage.getBiasedTrustScores
+      )
```
marcinwadon, line 71: isn't it just a tuple?

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/L0TrustClient.scala
```diff
@@ -14,16 +15,21 @@ import org.http4s.client.Client
...
+  def make[F[_]: Async: SecurityProvider: KryoSerializer](client: Client[F]): L0TrustClient[F] =
     new L0TrustClient[F] {
 
       def getCurrentTrust: PeerResponse[F, TrustScores] =
         PeerResponse[F, TrustScores]("trust/current")(client)
 
       def getPreviousTrust: PeerResponse[F, OrdinalTrustMap] =
         PeerResponse[F, OrdinalTrustMap]("trust/previous")(client)
+
+      def getPreviousSeedlistPeerLabels: PeerResponse[F, TrustLabels] =
```
marcinwadon, line 32: what does `seedlist` mean in this context? shouldn't it be just `getPreviousPeerLabels` as this is the endpoint you query for?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/ProposalTrustSelect.scala
```diff
@@ -19,19 +18,18 @@ import eu.timepit.refined.numeric.Positive
 object ProposalTrustSelect {
 
   def make[F[_]: MonadThrow](
-    getTrusts: F[(Option[Map[PeerId, Double]], TrustMap)],
+    getTrusts: F[(TrustLabels, TrustScores)],
```
marcinwadon, line 21: Can you pass `trustStorage` here? It does not make sense to pass all methods separately if you in fact use `trustStorage` in here and you wanted to make a single implementation only and iterate it over creating new strategies. 🤔

marcinwadon (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/ProposalTrustSelect.scala
cngo-github: We cannot pass the `TrustStorage`. `TrustStorage` is only available on the GL0. We want to use trust information other nodes (i.e, `currency`) as well. So, on `GL0`, `getTrusts` is created by wrapping calls to functions of `TrustStorage`. Where `TrustStorage` is unavailable, we populate `getTrust` from the `TrustRoutes`.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/ProposalTrustSelect.scala
cngo-github: Refactored to

```
      proposalSelect =
        if (cfg.environment === Dev)
          ProposalOccurrenceSelect.make()
        else {
          val getTrusts = Applicative[F].product(
            p2PClient.l0Trust.getPreviousSeedlistPeerLabels.run(cfg.globalL0Peer),
            p2PClient.l0Trust.getCurrentTrust.run(cfg.globalL0Peer)
          )

          val primary = ProposalTrustSelect.make(getTrusts, cfg.proposalSelect)
          val secondary = ProposalOccurrenceSelect.make()

          ProposalSelectWithFallback.make(primary, secondary)
        }
```

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/L0TrustClient.scala
cngo-github: Whenever peers update their deterministic trust map, they gossip it out as `SnapshotOrdinalPublicTrust`s. In other words, they gossip out a `Map[PeerId, Double]`, which is a mapping of how trustworthy they perceive other peers to be.

Nodes take these gossiped `PublicTrust`s and store them, associated to the peer that sent them, in their deterministic trust map as peerLabels. What this endpoint returns is:

Starting with the peerLabels, we grab the `labels` from the `PublicTrust` for each seedlist peer entry. In other words, we only use `PublicTrust`s from the node's seedlist peers. We then weight the trust values of each `labels` by how much we trust the peer that sent the `PublicTrust`. So we have an `Iterable` of weighted trust maps from each seedlist peer.

Then we combine the trust maps by averaging the trust scores for each entry. So, we are left with a trust map containing all the `PeerId`s gossiped to us by our seedlist peers during the last deterministic trust map update cycle, as well as their averaged, weighted trust score.

The seedlist peers are the peers designated upon node start up as its seedlist. The trustworthiness of each seedlist peer is loaded the same way. This functionality was added in PR https://github.com/Constellation-Labs/tessellation/pull/666.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/modules/Services.scala
cngo-github: Re-factored to

```
      proposalSelect =
        if (cfg.environment === Dev)
          ProposalOccurrenceSelect.make()
        else {
          val getTrusts = Applicative[F].product(
            p2PClient.l0Trust.getPreviousSeedlistPeerLabels.run(cfg.globalL0Peer),
            p2PClient.l0Trust.getCurrentTrust.run(cfg.globalL0Peer)
          )

          val primary = ProposalTrustSelect.make(getTrusts, cfg.proposalSelect)
          val secondary = ProposalOccurrenceSelect.make()

          ProposalSelectWithFallback.make(primary, secondary)
        }
```

modules/core/src/main/scala/org/tessellation/infrastructure/trust/storage/TrustStorage.scala
cngo-github: Changed the seedlist to `Set[SeedlistEntry]` instead of `Option[Set[SeedlistEntry]]`.

modules/core/src/main/scala/org/tessellation/infrastructure/trust/storage/TrustStorage.scala
```diff
@@ -205,23 +201,22 @@ object TrustStorage {
...
-            PublicTrustMap(trust.peerLabels.value ++ updatedPeerLabels)
-          )
-        }
+      private def getBiasedTrust(trust: TrustMap): TrustMap = {
+        val peerIds = seedlistScores.keys
+
+        val updatedTrust = peerIds
+          .filterNot(trust.trust.contains)
+          .map(_ -> TrustInfo(predictedTrust = config.seedlistInputBias.some))
+          .toMap
```
sdavidp, line 210: Not necessary. `Map.++` only requires an `Iterable[_]` which is what you have before the `.toMap`

modules/core/src/main/scala/org/tessellation/infrastructure/trust/storage/TrustStorage.scala
```diff
@@ -205,23 +201,22 @@ object TrustStorage {
...
+        val updatedPeerLabels = peerIds
+          .filterNot(trust.peerLabels.value.contains)
+          .map(_ -> PublicTrust.empty)
+          .toMap
```
sdavidp, line 214: Not necessary

modules/core/src/test/scala/org/tessellation/http/routes/TrustRoutesSuite.scala
```diff
@@ -136,4 +141,34 @@ object TrustRoutesSuite extends HttpSuite {
...
+          _ <- ts.updateNext(peerId, snapshotOrdinalPublicTrust)
+          _ <- ts.updateCurrent(snapshotOrdinal.plus(500L))
+
+          routes <- mkPublicRoutes(ts)
+
+          expected <- ts.getBiasedSeedlistOrdinalPeerLabels.map(_.asJson)
+          testResult <- expectHttpBodyAndStatus(routes, req)(expected, Status.Ok)
+        } yield testResult
+    }
+  }
```
sdavidp, line 172: I'm concerned about the hard-coded constants. Shouldn't be using the constants defined in `config` ?

cngo-github: This is only a test. Does it matter?

sdavidp: I believe it does because with the way it's written now one has to wonder if those values are the same as those in the config or if it's coincidence. Since the test relies on it matching those values it makes sense to pull the config out of your make function so you can access it and use them here.

modules/core/src/main/scala/org/tessellation/infrastructure/trust/storage/TrustStorage.scala
cngo-github: Removed.

modules/core/src/main/scala/org/tessellation/infrastructure/trust/storage/TrustStorage.scala
cngo-github: Removed.

modules/core/src/test/scala/org/tessellation/http/routes/TrustRoutesSuite.scala
cngo-github: All of the magic numbers have been replaced by their `config` counterparts.

kpudlik (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/http/p2p/clients/L0TrustClient.scala
marcinwadon: Ok, I see. Nevertheless, you query for previous peer labels, so IMO it's how the client should be named like. The info about the seedlist is imo the implementation detail that should be hidden behind the API call.

cngo-github: Renamed the function to `getPreviousPeerLabels`.

cngo-github:
This is currently blocked.

PR #737, opened 2023-10-06
Changed min. peers for selection to 1.

cngo-github:
Changed min. peers for selection to 1.

### DIFF
--- a/modules/core/src/main/scala/org/tessellation/cli/method.scala
+++ b/modules/core/src/main/scala/org/tessellation/cli/method.scala
@@ -71,8 +71,8 @@ object method {
       peerDiscoveryDelay = PeerDiscoveryDelay(
         checkPeersAttemptDelay = 1.minute,
         checkPeersMaxDelay = 10.minutes,
-        additionalDiscoveryDelay = 3.minutes,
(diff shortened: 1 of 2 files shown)


modules/core/src/main/scala/org/tessellation/cli/method.scala
```diff
@@ -71,8 +71,8 @@ object method {
       peerDiscoveryDelay = PeerDiscoveryDelay(
         checkPeersAttemptDelay = 1.minute,
         checkPeersMaxDelay = 10.minutes,
-        additionalDiscoveryDelay = 3.minutes,
-        minPeers = 2
+        additionalDiscoveryDelay = 0.minutes,
```
TheMMaciek, line 74: Aren't we effectively disabling this whole functionality? If so shouldn't we just remove/revert it (the delay functionality)? There is no point in maintaining obsolete not used code.

cngo-github: Per our offline discussion, we will proceed with disabling this functionality for now.

ryle-goehausen (approved):

PR #740, opened 2023-10-10
Rollback does not use majority hash checks for 1st snapshot pull

sdavidp:


TheMMaciek (changes requested):
see comments

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/programs/Rollback.scala
```diff
@@ -44,7 +44,7 @@ object Rollback {
     private val logger = Slf4jLogger.getLogger[F]
 
     def rollback: F[Unit] = for {
-      (globalSnapshot, globalSnapshotInfo) <- globalL0Service.pullLatestSnapshot
+      (globalSnapshot, globalSnapshotInfo) <- globalL0Service.pullLatestSnapshot(skipMajorityCheck = true)
```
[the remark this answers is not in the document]
sdavidp: @TheMMaciek because those are private, i.e. not part of the interface. This solution is one approach but I'm open to changing the interface if everyone's ok with that.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/services/GlobalL0Service.scala
```diff
@@ -34,6 +34,7 @@ import org.typelevel.log4cats.slf4j.Slf4jLogger
 trait GlobalL0Service[F[_]] {
   type LatestSnapshotTuple = (Hashed[GlobalIncrementalSnapshot], GlobalSnapshotInfo)
   def pullLatestSnapshot: F[LatestSnapshotTuple]
+  def pullLatestSnapshotFromRandomPeer: F[LatestSnapshotTuple]
```
marcinwadon, line 37: if you add `pullLatestSnapshotFromRandomPeer` to the trait itself, it makes me thinking what `pullLatestSnapshot` does. IMO it makes sense to either change the `pullLatestSnapshot` name as well, or parametrize the function instead of creating next one, like:

```scala
def pullLatestSnapshot(useRandomPeer: Boolean): ...
```

sdavidp: @marcinwadon I initially did that but previous PR comment suggested to make the call explicit instead

TheMMaciek: `pullLatestSnapshot` uses random or majority based on the configuration, we don't know the configuration at the call site. On the other hand we want to explicitly pull from random peer here.

TheMMaciek: At the same time `useRandomPeer` sounds much better than `skipMajorityCheck` so we could go with such a param I think if you guys like that solution more. `skipMajorityCheck` was a bit deceiving because it didn't indicate we are simply using random peer to pulll

TheMMaciek (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/services/GlobalL0Service.scala
marcinwadon: ok, I agree with @TheMMaciek then

PR #741, opened 2023-10-10
globalL0SnapshotProcessing uses setInitial for Left result

sdavidp:


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/storage/LastSnapshotStorage.scala
```diff
@@ -37,6 +37,7 @@ object LastSnapshotStorage {
 
       def set(snapshot: Hashed[S], state: SI): F[Unit] =
         snapshotR.modify {
+          case None => ((snapshot, state).some, Applicative[F].unit)
```
TheMMaciek, line 40: There is a `setInitial` function that performs that task. It's done like that on purpose so that setting first snapshot is   explicit and deliberate.

sdavidp: @TheMMaciek Yes, that's where I took that line from. I wasn't sure if it was necessary that we be explicit. I have to look further at the code but sometimes it seems like it's not easy to determine that.

Can we talk more about this task so I know how to go about fixing it?

TheMMaciek:
I would improve the commit message to be more reader friendly rather than mentioning class names and scala internals. We put these as changelog. Other than that the PR looks good.

sdavidp:
> I would improve the commit message to be more reader friendly rather than mentioning class names and scala internals. We put these as changelog. Other than that the PR looks good.

Is the new commit message better?

TheMMaciek (approved):

PR #742, opened 2023-10-11
Respect pull limit for majority hash validation

sdavidp:


modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/services/GlobalL0Service.scala
```diff
@@ -180,16 +182,23 @@ object GlobalL0Service {
             )
         }
 
+      private def calculateLastOrdinal(nextOrdinal: SnapshotOrdinal, latestOrdinal: SnapshotOrdinal): SnapshotOrdinal =
+        SnapshotOrdinal.unsafeApply(
```
marcinwadon, line 186: are we 100% sure it will always satisfy `SnapshotOrdinal` refinement?

sdavidp: @marcinwadon I believe this should always hold: `nextOrdinal < calculatedResult <= latestOrdinal`

marcinwadon: do I understand you correctly that the `nextOrdinal` is always lower than the `latestOrdinal`?

sdavidp: If not then the result is `latestOrdinal`.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/snapshot/services/GlobalL0Service.scala
```diff
@@ -168,8 +168,10 @@ object GlobalL0Service {
           } { lastStoredOrdinal =>
             for {
               l0Peer <- globalL0ClusterStorage.getRandomPeer
-              lastOrdinal <- l0GlobalSnapshotClient.getLatestOrdinal.run(l0Peer)
-              pulled <- pullSnapshots(l0Peer, lastStoredOrdinal.next, lastOrdinal)
+              latestOrdinal <- l0GlobalSnapshotClient.getLatestOrdinal.run(l0Peer)
+              nextOrdinal = lastStoredOrdinal.next
+              lastOrdinal = calculateLastOrdinal(nextOrdinal, latestOrdinal)
```
TheMMaciek, line 173: I would find a better name for these, latest refers to the latest on the node we query, last is the last one we want to download (limit?). The names themselves don't suggest which is which.

TheMMaciek (approved):
LGTM

PR #743, opened 2023-10-12
Data application calculated state with updated rewards fn

marcinwadon:


modules/currency-l0/src/main/scala/org/tessellation/currency/l0/http/routes/DataBlockRoutes.scala
```diff
@@ -38,4 +41,11 @@ final case class DataBlockRoutes[F[_]: Async](
           case Right(_) => Ok()
         }
   }
+
+  protected val p2p: HttpRoutes[F] = HttpRoutes.of[F] {
+    case GET -> Root / "state" / "off-chain" =>
```
ryle-goehausen, line 46: Using off-chain state might be ambiguous nomenclature, since off-chain is often used to describe data which has a hash reference on chain is stored separately entirely off-chain. A better name might be reflective of the fact that this is information that can only be reconstructed by calculating all operations embedded within the chain data. So it's entirely 'available' within the data data, but only if you reconstruct or recalculate from genesis.

Maybe calculatedState or something reflective of that ? To distinguish it from data which is completely isolated from the chain data?

marcinwadon: done :+1:

kpudlik (approved):

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/programs/Download.scala
```diff
@@ -65,8 +67,23 @@ object Download {
...
+          def setCalculatedState = maybeDataApplication.map { da =>
+            implicit val d = da.calculatedStateDecoder
+
+            clusterStorage.getResponsivePeers
+              .map(NodeState.ready)
+              .map(_.toList)
+              .flatMap(Random[F].shuffleList)
+              .flatMap {
+                case Nil       => (new Throwable(s"No peers to fetch off-chain state from")).raiseError[F, DataCalculatedState]
+                case peer :: _ => p2pClient.dataApplication.getCalculatedState.run(peer)
```
TheMMaciek, line 79: There is no way of telling whether this offChain state corresponds to the currency snapshot set here or if it's the previous/next one. Seems like this could be an issue. My intuition tells me that we should be able to tie the off-chain state to the snapshot e.g. by using ordinal.

TheMMaciek: I think we should also store a hash of the off chain state in the currency snapshot so that when the state is downloaded it can be hashed and verified.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotCreator.scala
```diff
@@ -101,29 +101,36 @@ object CurrencySnapshotCreator {
...
                   .ifF(updates.toList, List.empty[Signed[DataUpdate]])
               }
               .getOrElse(List.empty[Signed[DataUpdate]].pure[F])
-              .flatMap(service.combine(lastState, _))
-              .flatMap(service.serializeState)
+              .flatMap { updates =>
+                service.getCalculatedState
+                  .map(DataState(lastState, _))
+                  .flatMap(service.combine(_, updates))
+                  .flatTap(state => service.setCalculatedState(state.calculated))
```
TheMMaciek, line 131: Should we set state in the middle of the function run? I mean what if the function fails? The state will be set. Shouldn't we return the off chain state as the output of this function?

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/programs/Download.scala
marcinwadon: done :+1:

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotCreator.scala
marcinwadon: Let me revisit this later on. I agree with you - in case of partial failure, we may end up with corrupted state. Anyway, we're safe because of other checks (ordinal and hash proof), so nodes should not progress with corrupted state.

kpudlik (approved):

PR #744, opened 2023-10-12
Return rejected/awaited DataApplicationBlocks back to the consensus pool

kpudlik:
First: https://github.com/Constellation-Labs/tessellation/pull/726

### DIFF
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/cli/CliMethod.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/cli/CliMethod.scala
@@ -12,10 +12,10 @@ import org.tessellation.schema.address.Address
 import org.tessellation.schema.balance.Amount
 import org.tessellation.schema.node.NodeState
 import org.tessellation.schema.peer.PeerId
-import org.tessellation.sdk.PriorityPeerIds
 import org.tessellation.sdk.config.types._
 
 import eu.timepit.refined.auto._
+import eu.timepit.refined.types.all.PosLong
 import fs2.io.file.Path
 
 trait CliMethod {
@@ -89,6 +89,10 @@ trait CliMethod {
       ensureCheckInterval = 10.seconds
     )
   )
+  val snapshotSizeConfig: SnapshotSizeConfig = SnapshotSizeConfig(
+    maxSignaturesSizeInBytes = 102_400L,
+    maxStateChannelSnapshotBinarySizeInBytes = PosLong.unsafeFrom(512_000L)
+  )
 
   lazy val sdkConfig: SdkConfig = SdkConfig(
     environment,
@@ -98,7 +102,8 @@ trait CliMethod {
     stateAfterJoining,
     collateralConfig(environment, collateralAmount),
     trustStorageConfig,
-    PriorityPeerIds.get(environment)
+    PriorityPeerIds.get(environment),
+    snapshotSizeConfig
   )
 
 }
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/cli/CliMethod.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/cli/CliMethod.scala
@@ -12,35 +12,18 @@ import org.tessellation.schema.address.Address
 import org.tessellation.schema.balance.Amount
 import org.tessellation.schema.node.NodeState
 import org.tessellation.schema.peer.PeerId
+import org.tessellation.sdk.PriorityPeerIds
 import org.tessellation.sdk.config.types._
 
 import eu.timepit.refined.auto._
 import eu.timepit.refined.types.all.PosLong
 import fs2.io.file.Path
 
-trait CliMethod {
-
-  val keyStore: StorePath
-  val alias: KeyAlias
-  val password: Password
-
-  val environment: AppEnvironment
-
-  val seedlistPath: Option[SeedListPath]
-
-  val l0SeedlistPath: Option[SeedListPath]
-
-  val prioritySeedlistPath: Option[SeedListPath]
-
-  val stateChannelAllowanceLists: Option[Map[Address, NonEmptySet[PeerId]]]
-
-  val trustRatingsPath: Option[Path]
-
-  val httpConfig: HttpConfig
-
-  val stateAfterJoining: NodeState
-
-  val collateralAmount: Option[Amount]
+object CliMethod {
+  val snapshotSizeConfig: SnapshotSizeConfig = SnapshotSizeConfig(
+    singleSignatureSizeInBytes = 296L,
+    maxStateChannelSnapshotBinarySizeInBytes = PosLong(512_000L)
+  )
 
   val collateralConfig = (environment: AppEnvironment, amount: Option[Amount]) =>
     CollateralConfig(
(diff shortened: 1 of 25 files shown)


modules/sdk/src/main/scala/org/tessellation/sdk/cli/CliMethod.scala
```diff
@@ -16,31 +16,14 @@ import org.tessellation.sdk.PriorityPeerIds
...
-
-  val httpConfig: HttpConfig
-
-  val stateAfterJoining: NodeState
-
-  val collateralAmount: Option[Amount]
+object CliMethod {
+  val snapshotSizeConfig: SnapshotSizeConfig = SnapshotSizeConfig(
+    singleSignatureSizeInBytes = 296L,
+    maxStateChannelSnapshotBinarySizeInBytes = PosLong.unsafeFrom(512_000L)
```
sdavidp, line 25: Since we're constructing from a literal, the compiler can validate for us rather than at runtime. WDYT?

```suggestion
    maxStateChannelSnapshotBinarySizeInBytes = PosLong(512_000L)
```

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotCreator.scala
```diff
@@ -50,137 +54,181 @@ trait CurrencySnapshotCreator[F[_]] {
...
 object CurrencySnapshotCreator {
   def make[F[_]: Async: KryoSerializer: SecurityProvider](
     currencySnapshotAcceptanceManager: CurrencySnapshotAcceptanceManager[F],
-    maybeDataApplicationService: Option[(L0NodeContext[F], BaseDataApplicationL0Service[F])]
+    maybeDataApplicationService: Option[(L0NodeContext[F], BaseDataApplicationL0Service[F])],
+    snapshotSizeConfig: SnapshotSizeConfig
   ): CurrencySnapshotCreator[F] = new CurrencySnapshotCreator[F] {
     private val logger = Slf4jLogger.getLogger
 
+    private def maxProposalSizeInBytes(facilitators: Set[PeerId]) =
```
sdavidp, line 70: What's the return type? Will this function return value < 0?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotCreator.scala
```diff
@@ -50,137 +54,181 @@ trait CurrencySnapshotCreator[F[_]] {
...
-          awaitingBlocks.toSet,
-          rejectedBlocks.toSet
-        )
+          maybeLastDataState <- (maybeLastDataApplication, maybeDataApplicationService).mapN {
+            case (lastDataApplication, (_, service)) =>
+              service
+                .deserializeState(lastDataApplication)
+                .flatTap {
+                  case Left(err) => logger.warn(err)(s"Cannot deserialize custom state")
+                  case Right(a)  => logger.info(s"Deserialized state: $a")
```
sdavidp, line 114: How often will this log? This feels like debug or trace level.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotCreator.scala
```diff
@@ -50,137 +54,181 @@ trait CurrencySnapshotCreator[F[_]] {
...
+                }
+                .map(_.toOption)
+                .handleErrorWith(err =>
+                  logger.error(err)(s"Unhandled exception during deserialization data application, fallback to empty state") >>
```
sdavidp, line 118: Is this something that needs our attention? I.e. if one day we scrape the logs for ERROR messages is this something we need to address, or is it really a WARNing?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotCreator.scala
```diff
@@ -50,137 +54,181 @@ trait CurrencySnapshotCreator[F[_]] {
...
+            case (((nodeContext, service), lastState)) =>
+              implicit val context: L0NodeContext[F] = nodeContext
+              NonEmptyList
+                .fromList(dataBlocks.distinctBy(_.value.roundId))
+                .map(_.flatMap(_.value.updates))
+                .map { updates =>
+                  service
+                    .validateData(lastState, updates)
+                    .flatTap { validated =>
+                      Applicative[F].unlessA(validated.isValid)(logger.warn(s"Data application is invalid, errors: ${validated.toString}"))
```
sdavidp, line 133: Can this be simplified to
```suggestion
                      logger.warn(s"Data application is invalid, errors: ${validated.toString}").unlessA(validated.isValid)
```

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotCreator.scala
```diff
@@ -50,137 +54,181 @@ trait CurrencySnapshotCreator[F[_]] {
...
+                    }
+                    .map(_.isValid)
+                    .handleErrorWith(err =>
+                      logger.error(err)(s"Unhandled exception during validating data application, assumed as invalid") >> false.pure[F]
```
sdavidp, line 137:
```suggestion
                      logger.error(err)("Unhandled exception during validating data application, assumed as invalid").as(false)
```
Also, should this also be a WARN instead of ERROR?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotCreator.scala
```diff
@@ -50,137 +54,181 @@ trait CurrencySnapshotCreator[F[_]] {
...
+                      service
+                        .combine(lastState, updates.toList)
+                        .flatMap { state =>
+                          service.serializeState(state)
+                        }
+                        .map(_.some)
+                        .handleErrorWith(err =>
+                          logger.error(err)(
+                            s"Unhandled exception during combine and serialize data application, fallback to last data application"
+                          ) >> maybeLastDataApplication.pure[F]
```
sdavidp, line 149: Same as above. Use `.as` and wonder if this should be WARN

sdavidp (changes requested):
I'll finish review in a bit but take note of where `.as` can be used instead of `>> x.pure[F]`

sdavidp (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotCreator.scala
```diff
@@ -50,137 +54,181 @@ trait CurrencySnapshotCreator[F[_]] {
...
+              acceptanceResult.accepted match {
+                case _ :+ _ :+ excessivePair =>
+                  excessivePair match {
+                    case (excessiveBlock, _) =>
+                      val remainingBlocks = blocksForAcceptance.filterNot(_ =!= excessiveBlock)
+                      val excessiveBlocks = sizeExcessiveBlocks + excessiveBlock
+                      createProposalWithSizeLimit(remainingBlocks, excessiveBlocks)
+                  }
+                case _ =>
+                  new Throwable(s"Proposal doesn't fit the size limit after removing all the excessive blocks!")
```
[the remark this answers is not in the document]
marcinwadon: > IMO we should never construct Throwable objects.

@sdavidp can you explain why, so we would understand it better why it's bad?

kpudlik:
@sdavidp please review https://github.com/Constellation-Labs/tessellation/pull/726 first because that one is based on https://github.com/Constellation-Labs/tessellation/pull/726. Otherwise I will have comments here but I will be forced to fix another PR.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotCreator.scala
kpudlik: @marcinwadon - it is your part. Do you agree?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotCreator.scala
kpudlik: @marcinwadon ^ same question regarding severity

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotCreator.scala
kpudlik: I removed that log as it can easily pollute the logs.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotCreator.scala
kpudlik: Max proposal size can't be `<=  0` so I would set it to `PosLong`.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotCreator.scala
marcinwadon: seems like the comment is pointing the code that has been already merged and the change is only in indentation. can we create a ticket for that and discuss these changes separately @sdavidp ?

sdavidp:
Oops, I didn't see this until now. I'm in the middle of reviewing PR#750.
Do you want me to pause and go to 726?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/CurrencySnapshotCreator.scala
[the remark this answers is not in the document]
sdavidp: @marcinwadon `Throwable` is the top-level "error" type with two subclasses: `Error` and `Exception`. `Error` is thrown by the JVM and is almost always a fatal error that we cannot recover from: We should almost never catch an `Error`.

`Exception` is the supertype for all application level "errors" and is what we should be restricting ourselves to.

For reasons I'm not aware of _Cats_ uses `Throwable` as the expected type for all errors. However, IMHO it is not good form to be doing `new Throwable` for many reasons. An "ok" compromise is to do `new Exception`, esp. in tests, but in production I think it's cleaner to use existing exceptions if applicable or to create our own, as we do in many places in the codebase.

Does that help?

PR #749, opened 2023-10-16
Added fork info handler and storage

cngo-github:
Added fork info handler and storage.

### DIFF
--- a/modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
+++ b/modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
@@ -20,6 +20,7 @@ import org.tessellation.schema.balance.{Amount, Balance}
 import org.tessellation.schema.peer.PeerId
 import org.tessellation.sdk.domain.block.processing._
 import org.tessellation.sdk.domain.consensus.ConsensusFunctions.InvalidArtifact
+import org.tessellation.sdk.domain.gossip.Gossip
 import org.tessellation.sdk.domain.rewards.Rewards
 import org.tessellation.sdk.domain.snapshot.storage.SnapshotStorage
 import org.tessellation.sdk.infrastructure.consensus.trigger.{ConsensusTrigger, EventTrigger, TimeTrigger}
@@ -32,7 +33,7 @@ import org.tessellation.statechannel.{StateChannelOutput, StateChannelValidation
 import eu.timepit.refined.auto._
 import org.typelevel.log4cats.slf4j.Slf4jLogger
 
-abstract class GlobalSnapshotConsensusFunctions[F[_]: Async: SecurityProvider]
+abstract class GlobalSnapshotConsensusFunctions[F[_]: Async: SecurityProvider: KryoSerializer]
     extends SnapshotConsensusFunctions[
       F,
       GlobalSnapshotEvent,
@@ -47,7 +48,8 @@ object GlobalSnapshotConsensusFunctions {
     globalSnapshotStorage: SnapshotStorage[F, GlobalSnapshotArtifact, GlobalSnapshotContext],
     globalSnapshotAcceptanceManager: GlobalSnapshotAcceptanceManager[F],
     collateral: Amount,
-    rewards: Rewards[F, GlobalSnapshotStateProof, GlobalIncrementalSnapshot, GlobalSnapshotEvent]
+    rewards: Rewards[F, GlobalSnapshotStateProof, GlobalIncrementalSnapshot, GlobalSnapshotEvent],
+    gossip: Gossip[F]
   ): GlobalSnapshotConsensusFunctions[F] = new GlobalSnapshotConsensusFunctions[F] {
 
     private val logger = Slf4jLogger.getLoggerFromClass(GlobalSnapshotConsensusFunctions.getClass)
@@ -62,7 +64,8 @@ object GlobalSnapshotConsensusFunctions {
         .ifM(
           metrics.globalSnapshot(signedArtifact),
           logger.error("Cannot save GlobalSnapshot into the storage")
-        )
+        ) >>
+        gossipForkInfo(gossip, signedArtifact)
 
     override def validateArtifact(
       lastSignedArtifact: Signed[GlobalSnapshotArtifact],
--- a/modules/core/src/test/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctionsSuite.scala
+++ b/modules/core/src/test/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctionsSuite.scala
@@ -2,15 +2,17 @@ package org.tessellation.infrastructure.snapshot
 
 import cats.data.NonEmptyList
 import cats.effect.std.Supervisor
-import cats.effect.{IO, Resource}
+import cats.effect.{IO, Ref, Resource}
 import cats.syntax.applicative._
 import cats.syntax.either._
 import cats.syntax.list._
 
 import scala.collection.immutable.{SortedMap, SortedSet}
+import scala.reflect.runtime.universe.TypeTag
 
 import org.tessellation.currency.schema.currency.SnapshotFee
 import org.tessellation.ext.cats.syntax.next._
+import org.tessellation.ext.crypto._
 import org.tessellation.kryo.KryoSerializer
 import org.tessellation.schema.address.Address
 import org.tessellation.schema.balance.Amount
@@ -18,6 +20,8 @@ import org.tessellation.schema.epoch.EpochProgress
 import org.tessellation.schema.peer.PeerId
 import org.tessellation.schema.{Block, _}
 import org.tessellation.sdk.domain.block.processing._
+import org.tessellation.sdk.domain.fork.ForkInfo
+import org.tessellation.sdk.domain.gossip.Gossip
 import org.tessellation.sdk.domain.rewards.Rewards
 import org.tessellation.sdk.domain.snapshot.storage.SnapshotStorage
 import org.tessellation.sdk.infrastructure.consensus.trigger.EventTrigger
@@ -32,12 +36,35 @@ import org.tessellation.security.{KeyPairGenerator, SecurityProvider}
 import org.tessellation.statechannel.{StateChannelOutput, StateChannelSnapshotBinary, StateChannelValidationType}
 import org.tessellation.syntax.sortedCollection._
 
+import io.circe.Encoder
 import weaver.MutableIOSuite
 import weaver.scalacheck.Checkers
 
 object GlobalSnapshotConsensusFunctionsSuite extends MutableIOSuite with Checkers {
 
   type Res = (Supervisor[IO], KryoSerializer[IO], SecurityProvider[IO], Metrics[IO])
+
+  def mkMockGossip[B](spreadRef: Ref[IO, List[B]]): Gossip[IO] =
+    new Gossip[IO] {
+      override def spread[A: TypeTag: Encoder](rumorContent: A): IO[Unit] =
+        spreadRef.update(rumorContent.asInstanceOf[B] :: _)
+
+      override def spreadCommon[A: TypeTag: Encoder](rumorContent: A): IO[Unit] =
+        IO.raiseError(new Exception("spreadCommon: Unexpected call"))
+    }
+
+  def mkSignedArtifacts()(
+    implicit ks: KryoSerializer[IO],
+    sp: SecurityProvider[IO]
+  ): IO[(Signed[GlobalSnapshotArtifact], Signed[GlobalSnapshot])] = for {
+    keyPair <- KeyPairGenerator.makeKeyPair[IO]
+
+    genesis = GlobalSnapshot.mkGenesis(Map.empty, EpochProgress.MinValue)
+    signedGenesis <- Signed.forAsyncKryo[IO, GlobalSnapshot](genesis, keyPair)
+
+    lastArtifact <- GlobalIncrementalSnapshot.fromGlobalSnapshot(signedGenesis.value)
+    signedLastArtifact <- Signed.forAsyncKryo[IO, GlobalIncrementalSnapshot](lastArtifact, keyPair)
+  } yield (signedLastArtifact, signedGenesis)
 
   override def sharedResource: Resource[IO, Res] =
     Supervisor[IO].flatMap { supervisor =>
@@ -54,6 +81,7 @@ object GlobalSnapshotConsensusFunctionsSuite extends MutableIOSuite with Checker
       override def prepend(snapshot: Signed[GlobalIncrementalSnapshot], state: GlobalSnapshotInfo): IO[Boolean] = ???
 
       override def head: IO[Option[(Signed[GlobalIncrementalSnapshot], GlobalSnapshotInfo)]] = ???
+
       override def headSnapshot: IO[Option[Signed[GlobalIncrementalSnapshot]]] = ???
 
       override def get(ordinal: SnapshotOrdinal): IO[Option[Signed[GlobalIncrementalSnapshot]]] = ???
@@ -83,7 +111,7 @@ object GlobalSnapshotConsensusFunctionsSuite extends MutableIOSuite with Checker
 
   }
 
-  val scProcessor = new GlobalSnapshotStateChannelEventsProcessor[IO] {
+  val scProcessor: GlobalSnapshotStateChannelEventsProcessor[IO] = new GlobalSnapshotStateChannelEventsProcessor[IO] {
     def process(
       ordinal: SnapshotOrdinal,
       lastGlobalSnapshotInfo: GlobalSnapshotInfo,
@@ -110,94 +138,118 @@ object GlobalSnapshotConsensusFunctionsSuite extends MutableIOSuite with Checker
   val rewards: Rewards[F, GlobalSnapshotStateProof, GlobalIncrementalSnapshot, GlobalSnapshotEvent] =
     (_, _, _, _, _, _) => IO(SortedSet.empty)
 
-  def mkGlobalSnapshotConsensusFunctions()(implicit ks: KryoSerializer[IO], sp: SecurityProvider[IO], m: Metrics[IO]) = {
-    val snapshotAcceptanceManager = GlobalSnapshotAcceptanceManager.make[IO](bam, scProcessor, collateral)
-    GlobalSnapshotConsensusFunctions.make[IO](gss, snapshotAcceptanceManager, collateral, rewards)
+  def mkGlobalSnapshotConsensusFunctions(gossip: Gossip[F])(
+    implicit ks: KryoSerializer[IO],
+    sp: SecurityProvider[IO],
+    m: Metrics[IO]
+  ): GlobalSnapshotConsensusFunctions[IO] = {
+    val snapshotAcceptanceManager: GlobalSnapshotAcceptanceManager[IO] =
+      GlobalSnapshotAcceptanceManager.make[IO](bam, scProcessor, collateral)
+
+    GlobalSnapshotConsensusFunctions
+      .make[IO](
+        gss,
+        snapshotAcceptanceManager,
+        collateral,
+        rewards,
+        gossip
+      )
   }
+
+  def getTestData(
+    implicit sp: SecurityProvider[F],
+    kryo: KryoSerializer[F],
+    m: Metrics[F]
+  ): IO[(GlobalSnapshotConsensusFunctions[IO], Set[PeerId], Signed[GlobalSnapshotArtifact], Signed[GlobalSnapshot], StateChannelEvent)] =
+    for {
+      gossip <- Ref.of(List.empty[ForkInfo]).map(mkMockGossip)
+      gscf = mkGlobalSnapshotConsensusFunctions(gossip)
+      facilitators = Set.empty[PeerId]
+
+      keyPair <- KeyPairGenerator.makeKeyPair[F]
+
+      genesis = GlobalSnapshot.mkGenesis(Map.empty, EpochProgress.MinValue)
+      signedGenesis <- Signed.forAsyncKryo[F, GlobalSnapshot](genesis, keyPair)
+
+      lastArtifact <- GlobalIncrementalSnapshot.fromGlobalSnapshot(signedGenesis.value)
+      signedLastArtifact <- Signed.forAsyncKryo[IO, GlobalIncrementalSnapshot](lastArtifact, keyPair)
+
+      scEvent <- mkStateChannelEvent()
+    } yield (gscf, facilitators, signedLastArtifact, signedGenesis, scEvent)
 
   test("validateArtifact - returns artifact for correct data") { res =>
     implicit val (_, ks, sp, m) = res
 
-    val gscf = mkGlobalSnapshotConsensusFunctions()
+    for {
+      (gscf, facilitators, signedLastArtifact, signedGenesis, scEvent) <- getTestData
 
-    val facilitators = Set.empty[PeerId]
+      (artifact, _, _) <- gscf.createProposalArtifact(
+        SnapshotOrdinal.MinValue,
+        signedLastArtifact,
+        signedGenesis.value.info,
+        EventTrigger,
+        Set(scEvent.asLeft[DAGEvent]),
+        facilitators
+      )
+      result <- gscf.validateArtifact(
+        signedLastArtifact,
+        signedGenesis.value.info,
+        EventTrigger,
+        artifact,
+        facilitators
+      )
 
-    KeyPairGenerator.makeKeyPair[IO].flatMap { keyPair =>
-      val genesis = GlobalSnapshot.mkGenesis(Map.empty, EpochProgress.MinValue)
-      Signed.forAsyncKryo[IO, GlobalSnapshot](genesis, keyPair).flatMap { signedGenesis =>
-        GlobalIncrementalSnapshot.fromGlobalSnapshot(signedGenesis.value).flatMap { lastArtifact =>
-          Signed.forAsyncKryo[IO, GlobalIncrementalSnapshot](lastArtifact, keyPair).flatMap { signedLastArtifact =>
-            mkStateChannelEvent().flatMap { scEvent =>
-              gscf
-                .createProposalArtifact(
-                  SnapshotOrdinal.MinValue,
-                  signedLastArtifact,
-                  signedGenesis.value.info,
-                  EventTrigger,
-                  Set(scEvent.asLeft[DAGEvent]),
-                  facilitators
-                )
-                .flatMap {
-                  case (artifact, _, _) =>
-                    gscf.validateArtifact(signedLastArtifact, signedGenesis.value.info, EventTrigger, artifact, facilitators).map {
-                      result =>
-                        expect.same(result.isRight, true) && expect
-                          .same(
-                            result.map(_._1.stateChannelSnapshots(scEvent.address)),
-                            Right(NonEmptyList.one(scEvent.snapshotBinary))
-                          )
-                    }
-                }
-            }
-          }
-        }
-      }
-    }
+      expected = Right(NonEmptyList.one(scEvent.snapshotBinary))
+      actual = result.map(_._1.stateChannelSnapshots(scEvent.address))
+      expectation = expect.same(true, result.isRight) && expect.same(expected, actual)
+    } yield expectation
   }
 
   test("validateArtifact - returns invalid artifact error for incorrect data") { res =>
     implicit val (_, ks, sp, m) = res
 
-    val gscf = mkGlobalSnapshotConsensusFunctions()
+    for {
+      (gscf, facilitators, signedLastArtifact, signedGenesis, scEvent) <- getTestData
 
-    val facilitators = Set.empty[PeerId]
-
-    KeyPairGenerator.makeKeyPair[IO].flatMap { keyPair =>
-      val genesis = GlobalSnapshot.mkGenesis(Map.empty, EpochProgress.MinValue)
-      Signed.forAsyncKryo[IO, GlobalSnapshot](genesis, keyPair).flatMap { signedGenesis =>
-        GlobalIncrementalSnapshot.fromGlobalSnapshot(signedGenesis.value).flatMap { lastArtifact =>
-          Signed.forAsyncKryo[IO, GlobalIncrementalSnapshot](lastArtifact, keyPair).flatMap { signedLastArtifact =>
-            mkStateChannelEvent().flatMap { scEvent =>
-              gscf
-                .createProposalArtifact(
-                  SnapshotOrdinal.MinValue,
-                  signedLastArtifact,
-                  signedGenesis.value.info,
-                  EventTrigger,
-                  Set(scEvent.asLeft[DAGEvent]),
-                  facilitators
-                )
-                .flatMap { proposalArtifact =>
-                  gscf
-                    .validateArtifact(
-                      signedLastArtifact,
-                      signedGenesis.value.info,
-                      EventTrigger,
-                      proposalArtifact._1.copy(ordinal = proposalArtifact._1.ordinal.next),
-                      facilitators
-                    )
-                    .map { result =>
-                      expect.same(result.isLeft, true)
-                    }
-                }
-            }
-          }
-        }
-      }
-    }
+      (artifact, _, _) <- gscf.createProposalArtifact(
+        SnapshotOrdinal.MinValue,
+        signedLastArtifact,
+        signedGenesis.value.info,
+        EventTrigger,
+        Set(scEvent.asLeft[DAGEvent]),
+        facilitators
+      )
+      result <- gscf.validateArtifact(
+        signedLastArtifact,
+        signedGenesis.value.info,
+        EventTrigger,
+        artifact.copy(ordinal = artifact.ordinal.next),
+        facilitators
+      )
+    } yield expect.same(true, result.isLeft)
   }
 
-  def mkStateChannelEvent()(implicit S: SecurityProvider[IO], K: KryoSerializer[IO]) = for {
+  test("gossip signed artifacts") { res =>
+    implicit val (_, ks, sp, m) = res
+
+    for {
+      gossiped <- Ref.of(List.empty[ForkInfo])
+      mockGossip = mkMockGossip(gossiped)
+
+      gscf = mkGlobalSnapshotConsensusFunctions(mockGossip)
+      (signedLastArtifact, _) <- mkSignedArtifacts()
+
+      _ <- gscf.gossipForkInfo(mockGossip, signedLastArtifact)
+
+      expected = signedLastArtifact.hash.map { h =>
+        List(ForkInfo(signedLastArtifact.value.ordinal, h))
+      }
+        .getOrElse(List.empty)
+      actual <- gossiped.get
+    } yield expect.eql(expected, actual)
+  }
+
+  def mkStateChannelEvent()(implicit S: SecurityProvider[IO], K: KryoSerializer[IO]): IO[StateChannelEvent] = for {
     keyPair <- KeyPairGenerator.makeKeyPair[IO]
     binary = StateChannelSnapshotBinary(Hash.empty, "test".getBytes, SnapshotFee.MinValue)
     signedSC <- forAsyncKryo(binary, keyPair)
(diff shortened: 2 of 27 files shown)


modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/fork/ForkInfoStorageSuite.scala
```diff
@@ -0,0 +1,90 @@
...
+    val config = ForkInfoStorageConfig(queueSize)
+
+    Ref[IO].of(queue).map(ForkInfoStorage.make(_, config))
+  }
+
+  test("initial store is empty") {
+    val gen = for {
+      queueSize <- Gen.chooseNum(2, 10)
+      refinedQueueSize = Refined.unsafeApply[Int, Positive](queueSize)
+    } yield refinedQueueSize
```
sdavidp, line 38: Do you really need a `forall` for this? checking for empty queue is not dependent on queue size

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/fork/ForkInfoStorageSuite.scala
```diff
@@ -0,0 +1,90 @@
...
+        for {
+          store <- mkForkInfoStorage(Queue.from(entries), queueSize)
+
+          _ <- store.add(entry)
+
+          expected: List[StoredForkInfo] = entries.tail :+ entry
+          actual <- store.getForkInfo.map(_.toList)
+        } yield expect.eql(expected, actual)
+    }
+  }
```
sdavidp, line 89: Food for thought: Do we need two tests for this?

Couldn't we do this with a single test where the number of entries generated is anywhere from 0 to _n * queueSize_ where _n = 2 or 3_ ?  (we really don't need anything big)

(Incidentally, shouldn't that variable be called `storageSize`?)

To verify the storage is working as expected you'd check that the sequence returned from the storage contains the last `storageSize` values input, which covers but under and overflow.

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/fork/HandlerSuite.scala
```diff
@@ -0,0 +1,51 @@
...
+import eu.timepit.refined.auto._
+import io.circe.syntax.EncoderOps
+import weaver.SimpleIOSuite
+import weaver.scalacheck.Checkers
+
+object HandlerSuite extends SimpleIOSuite with Checkers {
+
+  test("recieved gossip gets added to the storage") {
+    val gen = for {
+      forkInfo <- genStoredForkInfoEntry.map(_.forkInfo)
```
sdavidp, line 22: I don't think you should be accessing a `val` in another suite. Probably better to create a `generators` object and define `genStoredForkInfoEntry` and `genForkInfo` in there.

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/fork/HandlerSuite.scala
```diff
@@ -0,0 +1,51 @@
...
+            Ordinal(Generation.MinValue, Counter.MinValue),
+            forkInfo.asJson,
+            ContentType.of[ForkInfo]
+          )
+
+          _ <- handler
+            .run((rawRumor, peerId2))
+            .getOrElseF(
+              IO.raiseError(new Exception("failed to handle rumor"))
+            )
```
sdavidp, line 43:
```suggestion
          _ <- handler.run((rawRumor, peerId2)).value
```

sdavidp (changes requested):

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/fork/HandlerSuite.scala
sdavidp: @cngo-github Perhaps a refactor closer to what you had is
```
          _ <- handler
            .run((rawRumor, peerId2))
            .getOrRaise(new Exception("failed to handle rumor"))
```
if `None` is considered an error condition.

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/fork/HandlerSuite.scala
```diff
@@ -0,0 +1,51 @@
...
+import org.tessellation.sdk.infrastructure.fork.handler.forkInfoHandler
+
+import eu.timepit.refined.auto._
+import io.circe.syntax.EncoderOps
+import weaver.SimpleIOSuite
+import weaver.scalacheck.Checkers
+
+object HandlerSuite extends SimpleIOSuite with Checkers {
+
+  test("recieved gossip gets added to the storage") {
```
sdavidp, line 20:
```suggestion
  test("received gossip gets added to the storage") {
```

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/fork/HandlerSuite.scala
cngo-github: Changed to

```
          _ <- handler
            .run((rawRumor, peerId2))
            .getOrRaise(new Exception("failed to handle rumor"))
```

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/fork/HandlerSuite.scala
cngo-github: Corrected the typo.

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/fork/HandlerSuite.scala
cngo-github: Moved `genStoredForkInfoEntry` to a generators file.

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/fork/ForkInfoStorageSuite.scala
cngo-github: Consolidated the two tests.

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/fork/ForkInfoStorageSuite.scala
cngo-github: Replaced the generator with `Random.between`.

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -65,6 +69,13 @@ object GlobalSnapshotConsensusFunctions {
           metrics.globalSnapshot(signedArtifact),
           logger.error("Cannot save GlobalSnapshot into the storage")
         )
+        .flatMap { _ =>
+          signedArtifact.hash.toOption.map { hash =>
+            gossip.spread(
+              ForkInfo(signedArtifact.value.ordinal, hash)
+            )
+          }.getOrElse(().pure)
+        }
```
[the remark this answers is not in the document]
cngo-github: Changed to

```
        .flatTap { _ =>
          signedArtifact.hash.liftTo[F].flatMap { h =>
            gossip.spread(ForkInfo(signedArtifact.value.ordinal, h))
          }
```

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
```diff
@@ -52,7 +58,13 @@ object CurrencySnapshotConsensusFunctions {
...
       context: CurrencySnapshotContext
     ): F[Unit] =
-      stateChannelSnapshotService.consume(signedArtifact, context)
+      stateChannelSnapshotService.consume(signedArtifact, context).flatMap { _ =>
+        signedArtifact.hash.toOption.map { hash =>
+          gossip.spread(
+            ForkInfo(signedArtifact.value.ordinal, hash)
+          )
+        }.getOrElse(().pure)
+      }
```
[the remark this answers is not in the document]
cngo-github: A convenience method was added to `SnapshotConsensusFunctions`.

cngo-github:
> @cngo-github Have you checked the current implementation for fork detection whether it can be used to discover new forks? It's already in the codebase and @TheMMaciek mentioned this to you

I talked over this with @ryle-goehausen and the short answer is no.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/fork/ForkInfoStorage.scala
```diff
@@ -0,0 +1,35 @@
...
+object ForkInfoStorage {
+
+  type ForkInfoMap = Map[PeerId, ForkInfoQueue]
+
+  def make[F[_]: Monad: Ref.Make](config: ForkInfoStorageConfig): F[ForkInfoStorage[F]] =
+    Ref[F]
+      .of(Map.empty[PeerId, ForkInfoQueue])
+      .map(make(_, config))
+
+  def make[F[_]: Monad](
```
scasplte2, line 22: Is it common for us to overload functions that return different types?

modules/sdk/src/main/scala/org/tessellation/sdk/domain/fork/ForkInfoQueue.scala
```diff
@@ -0,0 +1,23 @@
...
+    ForkInfoQueue(updated, maxSize)
+  }
+}
+
+object ForkInfoQueue {
+  def empty(maxSize: PosInt): ForkInfoQueue = ForkInfoQueue(value = Queue.empty, maxSize = maxSize)
+
+  def fromList(entries: List[ForkInfo], maxSize: PosInt): ForkInfoQueue =
+    entries.foldLeft(ForkInfoQueue.empty(maxSize)) { case (acc, entry) => acc.add(entry) }
+}
```
sdavidp, line 23: 1. I don't think you should name your class based on the internal implementation. Also, your constructor allows the caller to create your object with an invalid state, a queue with more than `maxSize` elements.
2. Why are you passing in a Queue? Isn't that an internal (hidden) implementation detail? Your constructor should probably be equivalent to calling `empty`.
3. Is this really a `case class` ? Feels more like `class` to me.
4. `fromList` is too specific. Look at the interface for `Queue` and you can reduce the constraint on the inpur argument and probably also simplify your implementation.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/fork/ForkInfoStorage.scala
```diff
@@ -0,0 +1,35 @@
...
+    storeRef: Ref[F, ForkInfoMap],
+    config: ForkInfoStorageConfig
+  ): ForkInfoStorage[F] = new ForkInfoStorage[F] {
+    def add(peerId: PeerId, entry: ForkInfo): F[Unit] = storeRef.update { s =>
+      val updated = s.getOrElse(peerId, ForkInfoQueue.empty(config.maxSize)).add(entry)
+
+      s + (peerId -> updated)
+    }
+
```
sdavidp, line 31: It is probably faster and cleaner to use `Map.updatedWith`

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/fork/generators.scala
```diff
@@ -0,0 +1,17 @@
...
+import org.tessellation.sdk.domain.fork.ForkInfo
+import org.tessellation.security.hash.Hash
+
+import org.scalacheck.Gen
+
+object generators {
+  val genStoredForkInfoEntry: Gen[(PeerId, ForkInfo)] = for {
+    peerId <- peerIdGen
+    ordinal <- snapshotOrdinalGen
+    hashString <- Gen.alphaStr
```
sdavidp, line 14: I would use the hash generator, or create one if it doesn't exist.

sdavidp (changes requested):

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -65,6 +67,9 @@ object GlobalSnapshotConsensusFunctions {
           metrics.globalSnapshot(signedArtifact),
           logger.error("Cannot save GlobalSnapshot into the storage")
         )
+        .flatTap { _ =>
+          gossipForkInfo(gossip = gossip, signed = signedArtifact)
```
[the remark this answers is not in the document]
cngo-github: Removed the argument names.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
```diff
@@ -52,7 +55,9 @@ object CurrencySnapshotConsensusFunctions {
       signedArtifact: Signed[CurrencyIncrementalSnapshot],
       context: CurrencySnapshotContext
     ): F[Unit] =
-      stateChannelSnapshotService.consume(signedArtifact, context)
+      stateChannelSnapshotService.consume(signedArtifact, context).flatMap { _ =>
+        gossipForkInfo(gossip = gossip, signed = signedArtifact)
```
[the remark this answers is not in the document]
cngo-github: Removed the argument names.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/fork/ForkInfoQueue.scala
cngo-github: 1. What do you recommend?
2. The case class has been made a sealed case class and the only way to make it is with the `empty` and `from` functions.
3. Implementing it as a class gave me too many issues. Making it a sealed case class seems to resolve your other concerns.
4. Generalized to `from` that works with any `IterableOnce`.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/fork/ForkInfoStorage.scala
cngo-github: I made an implementation with `updatedWith` and spoke with @scasplte2 about it. We ultimately decided to revert to this.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/fork/ForkInfoEntries.scala
```diff
@@ -0,0 +1,35 @@
...
+  def empty(maxSize: PosInt): ForkInfoEntries = new ForkInfoEntries(maxSize, Queue.empty)
+
+  def from(maxSize: PosInt, entries: IterableOnce[ForkInfo]): Option[ForkInfoEntries] =
+    if (entries.iterator.size > maxSize)
+      none
+    else
+      entries.iterator
+        .foldLeft(ForkInfoEntries.empty(maxSize)) { case (acc, entry) => acc.add(entry) }
+        .some
+}
```
sdavidp, line 35: This doesn't address my concern of being able to create an invalid object. Here's an alternate implementation to consider:
```
sealed trait ForkInfoEntries {
  def add(entry: ForkInfo): ForkInfoEntries
  def getEntries: Iterable[ForkInfo]
}

object ForkInfoEntries {
  private case class ForkInfoEntriesImpl(maxSize: PosInt, value: Queue[ForkInfo]) extends ForkInfoEntries {
    def add(entry: ForkInfo): ForkInfoEntries = {
      val updated = (if (value.size === maxSize) value.tail else value) :+ entry

      copy(value = updated)
    }

    def getEntries: Iterable[ForkInfo] = value
  }

  def empty(maxSize: PosInt): ForkInfoEntries = ForkInfoEntriesImpl(maxSize, Queue.empty)

  def from(maxSize: PosInt, entries: Iterable[ForkInfo]): ForkInfoEntries =
    ForkInfoEntriesImpl(maxSize, Queue.from(entries.takeRight(maxSize)))
}
```

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/fork/generators.scala
```diff
@@ -0,0 +1,16 @@
...
+import org.tessellation.sdk.domain.fork.ForkInfo
+import org.tessellation.security.hash.Hash
+
+import org.scalacheck.Gen
+
+object generators {
+  val genStoredForkInfoEntry: Gen[(PeerId, ForkInfo)] = for {
+    peerId <- peerIdGen
+    ordinal <- snapshotOrdinalGen
+    hash <- Hash.arbitrary.arbitrary
```
sdavidp, line 14:
```suggestion
    hash <- Arbitrary.arbitrary[Hash]
```

sdavidp (changes requested):

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -62,6 +64,9 @@ object GlobalSnapshotConsensusFunctions {
           metrics.globalSnapshot(signedArtifact),
           logger.error("Cannot save GlobalSnapshot into the storage")
         )
+        .flatTap { _ =>
+          gossipForkInfo(gossip, signedArtifact)
+        }
```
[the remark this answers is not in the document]
marcinwadon: @sdavidp  why such change? IMO changing the order of execution may drastically obfuscate the code. Also, there is a difference between *> and >> and we prefer the later as it chain operations

modules/sdk/src/main/scala/org/tessellation/sdk/domain/fork/ForkInfoQueue.scala
marcinwadon: > Implementing it as a class gave me too many issues.
@cngo-github can you describe some of them? `case class` is kind of sugar syntax that allows you to pattern match and I doubt you would use pattern matching for the queue

> I don't think you should name your class based on the internal implementation.
@sdavidp IMO it depends what level of abstraction you need. if that's a more specialized queue, then it's fine to name it a Queue. if it's a domain part, then it leaks the implementation in the name indeed

modules/sdk/src/main/scala/org/tessellation/sdk/domain/fork/ForkInfo.scala
```diff
@@ -0,0 +1,11 @@
...
+
+import org.tessellation.schema.SnapshotOrdinal
+import org.tessellation.security.hash.Hash
+
+import derevo.cats.{eqv, show}
+import derevo.circe.magnolia.{decoder, encoder}
+import derevo.derive
+
+@derive(eqv, decoder, encoder, show)
+case class ForkInfo(ordinal: SnapshotOrdinal, hash: Hash)
```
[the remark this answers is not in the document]
marcinwadon: question: maybe it should be: ```case class ForkInfo(ordinal: MajoritySnapshotOrdinal)```

modules/sdk/src/main/scala/org/tessellation/sdk/domain/fork/ForkDetect.scala
```diff
@@ -0,0 +1,5 @@
+package org.tessellation.sdk.domain.fork
+
+trait ForkDetect[F[_]] {
+  def getMajorityFork: F[ForkInfo]
```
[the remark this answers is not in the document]
marcinwadon: @sdavidp can you elaborate on that? currently, we use no-args fn for getters, and `()` if the fn creates/modifies something.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/fork/ForkInfo.scala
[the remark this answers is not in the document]
sdavidp: @marcinwadon Would would be the distinction between the two in this case? They both represent a _ordinal-hash_ pair?

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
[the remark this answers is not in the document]
sdavidp: @marcinwadon You're right. I thought it was equivalent to _flatTap_ but it's not, it's still calling _productL_ which is not the same. `<<` is deprecated which is why I used `<*`.

I think the deprecation message is misleading because the operations are not analagous:
`>>` is an alias for `flatMap` but `<<` is NOT an alias for `flatTap` as I expected.

I'm sorry @cngo-github, @marcinwadon  is correct. Ignore those two suggestions to use `<*`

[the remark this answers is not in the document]
cngo-github: Changed the `flatTap`s to `>>`.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/fork/ForkInfoEntries.scala
cngo-github: Changed to

```
sealed trait ForkInfoEntries {
  def add(entry: ForkInfo): ForkInfoEntries
  def getEntries: Iterable[ForkInfo]
}

object ForkInfoEntries {
  implicit val eq: Eq[ForkInfoEntries] = Eq.instance[ForkInfoEntries] { (e1, e2) =>
    e1.getEntries.toSeq === e2.getEntries.toSeq
  }

  implicit val show: Show[ForkInfoEntries] = Show.show[ForkInfoEntries] { e =>
    e.getEntries.mkString("[", ",", "]")
  }

  private case class ForkInfoEntriesImpl(maxSize: PosInt, value: Queue[ForkInfo]) extends ForkInfoEntries {
    def add(entry: ForkInfo): ForkInfoEntries = {
      val updated = (if (value.size === maxSize) value.tail else value) :+ entry

      copy(value = updated)
    }

    def getEntries: Iterable[ForkInfo] = value
  }

  def empty(maxSize: PosInt): ForkInfoEntries = ForkInfoEntriesImpl(maxSize, Queue.empty)

  def from(maxSize: PosInt, entries: Iterable[ForkInfo]): ForkInfoEntries =
    ForkInfoEntriesImpl(maxSize, Queue.from(entries.takeRight(maxSize)))

}
```

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/fork/generators.scala
cngo-github: Changed to:

```
    hash <- Arbitrary.arbitrary[Hash]
```

modules/sdk/src/main/scala/org/tessellation/sdk/domain/fork/ForkInfoQueue.scala
cngo-github: @marcinwadon , I had problems with `eqv` and `show`. This has been resolved in [this comment](https://github.com/Constellation-Labs/tessellation/pull/749#discussion_r1369440934).

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/fork/handler.scala
```diff
@@ -0,0 +1,25 @@
...
+
+import org.tessellation.schema.gossip.PeerRumor
+import org.tessellation.sdk.domain.fork.{ForkInfo, ForkInfoStorage}
+import org.tessellation.sdk.infrastructure.gossip.{IgnoreSelfOrigin, RumorHandler, rumorLoggerName}
+
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+
+object handler {
+
+  def forkInfoHandler[F[_]: Async](storage: ForkInfoStorage[F]): RumorHandler[F] = {
```
sdavidp, line 15: Would this be more clear if the object was called `ForkInfoRumorHandler` and the function called `make`?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/fork/ForkDetect.scala
```diff
@@ -0,0 +1,48 @@
+package org.tessellation.sdk.infrastructure.fork
+
+import cats.Monad
+import cats.syntax.flatMap._
+import cats.syntax.functor._
```
sdavidp, line 5:
```suggestion
import cats.syntax.contravariantSemigroupal._
```
replace with this import if you go with `mapN` below

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/fork/ForkDetect.scala
```diff
@@ -0,0 +1,48 @@
...
+  def make[F[_]: Monad](
+    getTrustScores: F[TrustScores],
+    getForkInfo: F[ForkInfoMap]
+  ): ForkDetect[F] = new ForkDetect[F] {
+
+    def getMajorityFork: F[ForkInfo] =
+      getTrustScores.flatMap { trust =>
+        val scores = trust.scores.filter { case (_, s) => s > 0.0 }
+
+        getForkInfo.map { info =>
```
sdavidp, line 23:
```suggestion
      (getTrustScores, getForkInfo).mapN { case (trust, info) =>
        val scores = trust.scores.filter { case (_, s) => s > 0.0 }
```
If you take this suggestion than also use suggestion re. the import

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/fork/ForkDetect.scala
```diff
@@ -0,0 +1,48 @@
...
+          val majorityOrdinal = scoredOrdinals.maxBy { case (_, s) => s }._1
+          val majorityHash =
+            groupedOrdinals(majorityOrdinal).groupMapReduce { case (h, _) => h } { case (_, s) => s }(_ + _).maxBy {
+              case (_, s) => s
+            }._1
+
+          ForkInfo(majorityOrdinal, majorityHash)
+        }
+      }
+  }
```
sdavidp, line 46:
```suggestion
```
Remove a closing brace if you go with `mapN`

modules/sdk/src/main/scala/org/tessellation/sdk/domain/fork/ForkInfo.scala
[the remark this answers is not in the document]
sdavidp: To me there's nothing about the the contents of this class that identifies it as "fork info"; it seems to identify a snapshot. In fact, it's a subset of _snapshot.scala:SnapshotMetadata_
```
  case class SnapshotMetadata(ordinal: SnapshotOrdinal, hash: Hash, lastSnapshotHash: Hash)
```

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/fork/handler.scala
cngo-github: I am following the convention in the other `handler` files.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/fork/ForkDetect.scala
cngo-github: Changed to

```
      (getTrustScores, getForkInfo).mapN {
```

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/fork/ForkDetect.scala
cngo-github: Removed.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/fork/ForkDetect.scala
cngo-github: Imported `contravariantSemigroupal`.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/fork/handler.scala
sdavidp: There are also examples of a well-named `object` and `make` or other descriptive name. I think that's a better pattern to follow.

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/fork/HandlerSuite.scala
```diff
@@ -0,0 +1,51 @@
...
+          rawRumor = PeerRumorRaw(
+            peerId1,
+            Ordinal(Generation.MinValue, Counter.MinValue),
+            forkInfo.asJson,
+            ContentType.of[ForkInfo]
+          )
+
+          _ <- handler
+            .run((rawRumor, peerId2))
+            .getOrRaise(new Exception("failed to handle rumor"))
```
sdavidp, line 38: Is this is correct?
Doesn't the fact that `RumorHandler.run` is a Kleisli on `OptionT` mean we have 3 states?

- F[Some(Unit)]
- F[None]
- F raised as error

It seems to me that in this case we're raising an error on the `None` case but from what I can tell by the implementation that (a) would never happen, and (b) isn't really an error since we're throwing the option away.

So isn the real error case when there's an error in `F` ?
Or put another way, do we just need this for the happy path:
```suggestion
          _ <- handler.run((rawRumor, peerId2)).value
```
and then have a test to show what is the error case?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/fork/forkInfoHandler.scala
```diff
@@ -0,0 +1,25 @@
...
+import cats.syntax.flatMap._
+import cats.syntax.show._
+
+import org.tessellation.schema.gossip.PeerRumor
+import org.tessellation.sdk.domain.fork.{ForkInfo, ForkInfoStorage}
+import org.tessellation.sdk.infrastructure.gossip.{IgnoreSelfOrigin, RumorHandler, rumorLoggerName}
+
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+
+object forkInfoHandler {
```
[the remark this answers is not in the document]
cngo-github: Changed to `ForkInfoHandler`.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/fork/ForkInfoEntries.scala
```diff
@@ -0,0 +1,40 @@
...
+    def add(entry: ForkInfo): ForkInfoEntries = {
+      val updated = (if (value.size === maxSize) value.tail else value) :+ entry
+
+      copy(value = updated)
+    }
+
+    def getEntries: Iterable[ForkInfo] = value
+  }
+
+  def empty(maxSize: PosInt): ForkInfoEntries = ForkInfoEntriesImpl(maxSize, Queue.empty)
```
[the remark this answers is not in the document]
cngo-github: Changed to `apply`.

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/fork/HandlerSuite.scala
```diff
@@ -0,0 +1,51 @@
...
+import org.tessellation.sdk.config.types.ForkInfoStorageConfig
+import org.tessellation.sdk.domain.fork.{ForkInfo, ForkInfoEntries, ForkInfoMap}
+import org.tessellation.sdk.infrastructure.fork.generators.genStoredForkInfoEntry
+
+import eu.timepit.refined.auto._
+import io.circe.syntax.EncoderOps
+import weaver.SimpleIOSuite
+import weaver.scalacheck.Checkers
+
+object HandlerSuite extends SimpleIOSuite with Checkers {
```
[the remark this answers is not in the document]
cngo-github: Changed to `ForkInfoHandlerSuite`.

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/fork/HandlerSuite.scala
cngo-github: `None` is irrelevant and the `getOrRaise` is replaced by `value`.

sdavidp (dismissed):
LGTM, though I think you should get an approval from someone else as well since I helped write some of this.

modules/sdk/src/main/scala/org/tessellation/sdk/domain/fork/ForkInfoEntries.scala
```diff
@@ -0,0 +1,40 @@
...
+    e1.getEntries.toSeq === e2.getEntries.toSeq
+  }
+
+  implicit val show: Show[ForkInfoEntries] = Show.show[ForkInfoEntries] { e =>
+    e.getEntries.mkString("[", ",", "]")
+  }
+
+  private case class ForkInfoEntriesImpl(maxSize: PosInt, value: Queue[ForkInfo]) extends ForkInfoEntries {
+    def add(entry: ForkInfo): ForkInfoEntries = {
+      val updated = (if (value.size === maxSize) value.tail else value) :+ entry
```
kpudlik, line 27: So in case of reaching the size limit we cut the `head` and `append` the `entry`. It sounds like a combination of `dequeue` + `enqueue` of `FIFO` but you use neither of `Queue` specific methods.

So maybe instead of calling `tail` and `append` we can use the fact that both `dequeue` and `enqueue` have `O(1)` complexity which makes `O(1) + O(1) = O(1)`

```scala
val updated = if (value.size === maxSize) {
  value.dequeueOption
     .map { case (_, q) => q }
     .getOrElse(value)
     .enqueue(entry)
} else {
  value.enqueue(entry)
}
```

kpudlik (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/domain/fork/ForkInfoEntries.scala
cngo-github: Changed to

```
    def add(entry: ForkInfo): ForkInfoEntries = {
      val updated = if (value.size === maxSize) {
        value.dequeueOption.map { case (_, q) => q }
          .getOrElse(value)
          .enqueue(entry)
      } else {
        value.enqueue(entry)
      }

      copy(value = updated)
    }
```

sdavidp: I can see the logic of using `enqueue` and `dequeue` but I think the implementation can be simplified to
```
if (value.size === maxSize) value.dequeue._2.enqueue(entry) else value.enqueue(entry)
```
since `maxSize` less than should not be permitted as it creates an unusable object (i.e. starting with an invalid state)

cngo-github: Changed to:

```
    def add(entry: ForkInfo): ForkInfoEntries = {
      val updated =
        if (value.size === maxSize)
          value.dequeue._2.enqueue(entry)
        else
          value.enqueue(entry)

      copy(value = updated)
    }
```

kpudlik: @sdavidp @cngo-github Ok I see. `value.dequeue` can throw `NoSuchElementException` for empty `Queue`, so I decided to go for `dequeueOption` but since `maxSize` is `PosLong` then it is secured by `value.size === maxSize` check. :+1:

modules/core/src/test/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctionsSuite.scala
```diff
@@ -110,44 +138,59 @@ object GlobalSnapshotConsensusFunctionsSuite extends MutableIOSuite with Checker
...
+                        result =>
+                          expect.same(result.isRight, true) && expect
+                            .same(
+                              result.map(_._1.stateChannelSnapshots(scEvent.address)),
+                              Right(NonEmptyList.one(scEvent.snapshotBinary))
+                            )
+                      }
+                  }
+              }
             }
```
kpudlik, line 194: Since we have a lot of nested `flatMap`s then we should consider for-comp:

```scala
for {
  keyPair <- KeyPairGenerator.makeKeyPair[IO]
  genesis = GlobalSnapshot.mkGenesis(Map.empty, EpochProgress.MinValue)
  signedGenesis <- Signed.forAsyncKryo[IO, GlobalSnapshot](genesis, keyPair)
  ... // etc etc
} yield ()
```

kpudlik (dismissed):

modules/core/src/test/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotConsensusFunctionsSuite.scala
cngo-github: Re-factored to use `for`-comprehension.

PR #757, opened 2023-10-24
Separate mode to generate genesis with address metadata

kpudlik:


modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/programs/Genesis.scala
```diff
@@ -96,24 +102,44 @@ object Genesis {
...
+        genesis = CurrencySnapshot.mkGenesis(balances, dataApplicationPart)
+        signedGenesis <- genesis.sign(keyPair)
+        signedBinary <- stateChannelSnapshotService.createGenesisBinary(signedGenesis)
+        identifier = signedBinary.value.toAddress
+      } yield (signedGenesis, identifier)
+
+      createGenesis.flatTap {
+        case (signedGenesis, identifier) =>
+          genesisLoader.write(signedGenesis, identifier, genesisOutputPath).void
+      }
```
sdavidp, line 142:
```suggestion
        _ <- genesisLoader.write(signedGenesis, identifier, genesisOutputPath)
      } yield (signedGenesis, identifier)
```
then you can discard `createGenesis` if you like and just have the for-comp.

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/genesis/GenesisFSSuite.scala
```diff
@@ -0,0 +1,70 @@
...
+  } yield (balancesMap, identifier)
+
+  test("writes and loads genesis") { res =>
+    implicit val (ks, sp) = res
+    val loader = GenesisFS.make[IO, CurrencySnapshot]
+
+    forall(gen) {
+      case (balances, identifier) =>
+        for {
+          tempDir <- Files[F].createTempFile
```
sdavidp, line 39: `createTempFile` won't clean up after its done.

```suggestion
        Files[F].tempFile.use { tmpPath =>
          for {
```
and add closing brace. I also renamed the variable, if you agree it's clearer.

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/genesis/GenesisFSSuite.scala
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  test("writes and loads metadata") { res =>
+    implicit val (ks, sp) = res
+    val loader = GenesisFS.make[IO, CurrencySnapshot]
+
+    forall(gen) {
+      case (balances, identifier) =>
+        for {
+          tempDir <- Files[F].createTempFile
```
sdavidp, line 56:
```suggestion
        Files[F].tempFile.use { tmpPath =>
          for {
```

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/programs/Genesis.scala
```diff
@@ -96,24 +102,44 @@ object Genesis {
...
-      dataApplication: Option[BaseDataApplicationL0Service[F]] = None
-    )(implicit context: L0NodeContext[F]): F[Unit] = {
+    override def accept(dataApplication: Option[BaseDataApplicationL0Service[F]])(genesisPath: Path)(
+      implicit context: L0NodeContext[F]
+    ): F[Unit] = genesisLoader
+      .loadSignedGenesis(genesisPath)
+      .flatTap { genesis =>
+        dataApplication
+          .map(app => app.setCalculatedState(genesis.ordinal, app.genesis.calculated))
+          .getOrElse(false.pure[F])
```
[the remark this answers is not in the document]
kpudlik: `flatTap` expects `F[_]` to be returned but `dataApplication.map(...)` is `Option[F[_]]` so we need to get rid of `Option` to make the compiler happy. I think we can just use `traverse` instead of `getOrElse`.

```
      _ <- dataApplication.traverse { app =>
            app.setCalculatedState(firstIncrementalSnapshot.ordinal, app.genesis.calculated)
          }
```

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/programs/Genesis.scala
kpudlik: Sure. I had it extracted before for some reasons but implementation changed so I see that it can just single for-comp.

modules/sdk/src/test/scala/org/tessellation/sdk/infrastructure/genesis/GenesisFSSuite.scala
kpudlik: Good point. Using Resource is much better

sdavidp (approved):

PR #758, opened 2023-10-25
Link snapshot file to ordinal filename if missing

sdavidp:
Currently when we try to write a snapshot we raise an error if a file exists with the hash as filename or ordinal. This PR adjusts the logic so that if the hash filename exists but the ordinal is missing we create the link instead of raising error.

### DIFF
--- a/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
+++ b/modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
@@ -2,15 +2,19 @@ package org.tessellation.sdk.infrastructure.snapshot.storage
 
 import cats.Applicative
 import cats.effect.Async
+import cats.syntax.applicative._
 import cats.syntax.applicativeError._
 import cats.syntax.contravariantSemigroupal._
 import cats.syntax.flatMap._
 import cats.syntax.functor._
 
+import scala.util.control.NoStackTrace
+
 import org.tessellation.ext.crypto._
 import org.tessellation.kryo.KryoSerializer
 import org.tessellation.schema.SnapshotOrdinal
 import org.tessellation.schema.snapshot.Snapshot
+import org.tessellation.sdk.infrastructure.snapshot.storage.SnapshotLocalFileSystemStorage.UnableToPersisSnapshot
 import org.tessellation.security.hash.Hash
 import org.tessellation.security.signature.Signed
 import org.tessellation.storage.LocalFileSystemStorage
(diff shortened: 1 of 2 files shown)


ryle-goehausen (approved):
I think this is okay, it'd be a little more contentious if we were dealing with the other case (missing hash but ordinal found,) which might indicate a chain integrity error due to user error, but since we have hash validation on chain references based on prior hashes, this should fix errors where ordinal has been removed by accident without introducing an error

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
```diff
@@ -18,24 +18,31 @@ import org.tessellation.storage.LocalFileSystemStorage
...
+        case (true, false) =>
+          new Exception(s"Snapshot ordinal file $ordinalName exists but hash missing").raiseError[F, Unit]
+
+        case (false, true) =>
+          logger.warn(s"Snapshot hash file $hashName exists but ordinal missing; linking to $ordinalName") >>
+            link(hashName, ordinalName)
+
+        case (false, false) =>
+          write(hashName, snapshot) >> link(hashName, ordinalName)
+      }
```
TheMMaciek, line 44: I agree with the improved logic for logging. I would try though to log in a single message with ordinal, shortened hash name and two conditions s"savedAsOrdinal=$ordinalExists savedAsHash=$hashExists".

Regarding the fixing logic I think it will maybe fix a very particular small case where we successfully save under hash but then for some reason fail to link. If we have corrupted state under hash we will link but the snapshot under hash will not be fixed. Also we are checking if the snapshot is saved under ordinal and hash without locking so theoretically a state could change - still I think linking in such situation should simply fail if a file already exists, so maybe that's not a problem. Overwriting snapshot by hash in such a situation as well could be a better improvement. But then even better solution would be to handle the cleanup after failed writing attempt (I think such cleanup is not happening automatically but best to check) to try to not end in such a state at all, we would also need to synchronously lock both files I think. Even better would be to only save snapshots on disk by hash and have some transactional storage on disk for the ordinal -> hash mapping (overwriting snapshot by hash is safe, overwriting by ordinal isn't hence the need for transactional update for ordinal).

So in general I'm ambivalent regarding that change as I think it will probably not have a substantial impact and won't totally solve any of the issues we have.

ryle-goehausen (approved):
Agree with Maciej that this may not really solve all the underlying problems. It will at least fix the one issue we saw in the past, so it's okay with me to merge.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
Proph151Music: > I agree with the improved logic for logging. I would try though to log in a single message with ordinal, shortened hash name and two conditions s"savedAsOrdinal=$ordinalExists savedAsHash=$hashExists".
> 
> Regarding the fixing logic I think it will maybe fix a very particular small case where we successfully save under hash but then for some reason fail to link. If we have corrupted state under hash we will link but the snapshot under hash will not be fixed. Also we are checking if the snapshot is saved under ordinal and hash without locking so theoretically a state could change - still I think linking in such situation should simply fail if a file already exists, so maybe that's not a problem. Overwriting snapshot by hash in such a situation as well could be a better improvement. But then even better solution would be to handle the cleanup after failed writing attempt (I think such cleanup is not happening automatically but best to check) to try to not end in such a state at all, we would also need to synchronously lock both files I think. Even better would be to only save snapshots on disk by hash and have some transactional storage on disk for the ordinal -> hash mapping (overwriting snapshot by hash is safe, overwriting by ordinal isn't hence the need for transactional update for ordinal).
> 
> So in general I'm ambivalent regarding that change as I think it will probably not have a substantial impact and won't totally solve any of the issues we have.

Sorry if this is completely incorrect with what you are trying to accomplish... I think I understand what you guys are talking about... but it's possible I'm missing the entire context.  But here is an idea that maybe could help...

Could it possibly be helpful to rename the file to something like "{ordinal}-{hash}" after the snapshot has completed verification and linking. I could see this being useful in the fact that you would only have a properly named file once it has completely been verified and linked. If it hasn't been verified it would only be named "{ordinal}" without a dash and no hash, which would instantly indicate a file that was not fully validated.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
```diff
@@ -92,6 +102,10 @@ final class SnapshotLocalFileSystemStorage[F[_]: Async: KryoSerializer, S <: Sna
 
 object SnapshotLocalFileSystemStorage {
 
+  case class UnableToPersisSnapshot(ordinalName: String, hashName: String, hashFileExists: Boolean) extends NoStackTrace {
```
TheMMaciek, line 105: You have a small typo here.

PR #760, opened 2023-10-26
Enable state channel public routes

scasplte2:
## Overview
Enable state channel public routes

## Approach
- Removed flag checking environment in `HttpApi`

## Testing
- Unit test passing
- CI passed

## Tickets
- PROT-464


ryle-goehausen (approved):
May not be all of ticket but shouldn't break anything

PR #763, opened 2023-10-30
fix path portability for windows

### LINKED ISSUE #762: void

abebeos:
closes #762

This is a quick-fix that took several hours of trial&error to come up.

It looks like only `Paths.get(URI)` returns a correct path that starts with the drive-letter on windows.

I could not find `Paths.get` functionality within `fs2 Paths`, nor any other method that returns the correct path. Thus using `Paths.get()`.

(This should be possibly moved into an helper/shared function to avoid the duplication. But this PR works as-is fine on windows, tests pass now.)

### General Notes

An object of the class `Path` should naturally be working on the platform. This feels like a flaw/bug in the underlying java library.


ryle-goehausen (commented):
I don't think this will cause any problems, and if it fixes a windows issue it's a small enough change to be fine, running tests to verify now.

PR #765, opened 2023-11-01
Limit rewards to seedlist if desired

cngo-github:
Limit rewards to seedlist addresses only if desired

### DIFF
--- a/modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
+++ b/modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
@@ -10,6 +10,7 @@ import cats.syntax.either._
 import cats.syntax.eq._
 import cats.syntax.flatMap._
 import cats.syntax.functor._
+import cats.syntax.set._
 import cats.syntax.show._
 
 import scala.collection.immutable.{SortedMap, SortedSet}
@@ -22,6 +23,7 @@ import org.tessellation.schema.ID.Id
 import org.tessellation.schema.address.Address
 import org.tessellation.schema.balance.{Amount, Balance}
 import org.tessellation.schema.epoch.EpochProgress
+import org.tessellation.schema.peer.PeerId
 import org.tessellation.schema.transaction.{RewardTransaction, Transaction, TransactionAmount}
 import org.tessellation.schema.{GlobalIncrementalSnapshot, GlobalSnapshotStateProof, SnapshotOrdinal}
 import org.tessellation.sdk.domain.rewards.Rewards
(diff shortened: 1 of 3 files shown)


TheMMaciek (changes requested):
This PR is completely wrong so please close it.
You are taking the configuration of rewards distribution and mixing it up with seedlist which are two completely different things. You are actually filtering out addresses from config like stardust addresses (that get a constant 5% share of rewards) if they are not on the seedlist - they aren't, never were on that seedlist and don't need to be there. If your goal was to not distribute rewards to peers not present on the seedlist than nothing needs to be added AFAIK as the node running the GL0 consensus is not adding peers from outside of the seedlist as facilitators so effectively will not attempt distributing the rewards to such peers.

cngo-github:
> This PR is completely wrong so please close it. You are taking the configuration of rewards distribution and mixing it up with seedlist which are two completely different things. You are actually filtering out addresses from config like stardust addresses (that get a constant 5% share of rewards) if they are not on the seedlist - they aren't, never were on that seedlist and don't need to be there. If your goal was to not distribute rewards to peers not present on the seedlist than nothing needs to be added AFAIK as the node running the GL0 consensus is not adding peers from outside of the seedlist as facilitators so effectively will not attempt distributing the rewards to such peers.

Re-worked this PR per our offline discussion.

modules/core/src/main/scala/org/tessellation/modules/Services.scala
```diff
@@ -54,13 +57,16 @@ object Services {
     selfId: PeerId,
     keyPair: KeyPair,
     cfg: AppConfig
-  ): F[Services[F]] =
+  ): F[Services[F]] = {
+    val seedlistPeerIds = seedlist.map(_.map(_.peerId)).flatMap(SortedSet.from(_).toNes)
```
sdavidp, line 61:
```suggestion
    val seedlistPeerIds = seedlist.flatMap(s => SortedSet.from(s.map(_.peerId)).toNes)
```

cngo-github: Changed to

```    
    val seedlistPeerIds = seedlist.flatMap(s =>
      SortedSet
        .from(s.map(_.peerId))
        .toNes
    )
```

modules/core/src/main/scala/org/tessellation/infrastructure/rewards/Rewards.scala
```diff
@@ -56,7 +59,11 @@ object Rewards {
         events: Set[GlobalSnapshotEvent],
         maybeCalculatedState: Option[DataCalculatedState] = None
       ): F[SortedSet[RewardTransaction]] = {
-        val facilitators = lastArtifact.proofs.map(_.id)
+        val artifactFacilitators = lastArtifact.proofs.map(_.id)
+
+        val facilitators = seedlist.flatMap { s =>
+          artifactFacilitators.filter(id => s.contains(id.toPeerId)).toNes
+        }.getOrElse(artifactFacilitators)
```
kpudlik, line 66: It is wrong imo because you are filtering out the node that actually signed `lastArtifact` but is not on the seedlist anymore. Imo we should allow rewarding historical commitment.

kpudlik (changes requested):

cngo-github:
This is blocked for now.

PR #767, opened 2023-11-02
Add message to error classes that have data

sdavidp:


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/healthcheck/LocalHealthcheck.scala
```diff
@@ -31,7 +31,10 @@ object LocalHealthcheck {
     mkPeersR.map(make(_, retryPolicy, nodeClient, clusterStorage))
   }
 
-  case class PeerUnresponsive(id: PeerId) extends NoStackTrace
+  case class PeerUnresponsive(id: PeerId) extends NoStackTrace {
+    implicit val show = PeerId.show
+    override val getMessage = id.show
```
TheMMaciek, line 36: If we are adding the `getMessage` definition, maybe we should make it whole sentence? Like:
`Peer with id ${id.show} is unresponsive`

TheMMaciek: That's what I would vote for regarding every message in this PR.

TheMMaciek, line 36: Regarding `implicit val show = PeerId.show` maybe you can pass the show instance explicitly to `show` function? 🤔

modules/shared/src/main/scala/org/tessellation/security/signature/Signed.scala
```diff
@@ -35,6 +35,7 @@ import org.scalacheck.Arbitrary.arbitrary
 case class Signed[+A](value: A, proofs: NonEmptySet[SignatureProof])
 
 object Signed {
+  // TODO what part of signed can we write to the log for this error?
```
TheMMaciek, line 38: Not sure if with the current input there is something viable to be put in the error message. Please remove the TODO before the final version of this PR is ready (we tend to create tickets instead of TODOs).

sdavidp: @TheMMaciek I wondered... so if there isn't anything we can print then maybe the arg has no value?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/healthcheck/LocalHealthcheck.scala
sdavidp: @TheMMaciek I don't think `.show` takes an implicit arg. Is there a way to do it that I'm not seeing?

Also, thank you for the comment -- that's the wrong implicit; I wanted `shortShow`

sdavidp: @TheMMaciek I can do that; the reason I didn't is that the type name of the error already tells us that, and there's usually another message logged as well.

modules/shared/src/main/scala/org/tessellation/security/signature/Signed.scala
TheMMaciek: Depends how it's used, maybe some handling logic uses that value?

TheMMaciek (approved):

PR #768, opened 2023-11-03
Improve rollback flow and logging

TheMMaciek:
Improved logging to get a clear error when loading snapshot failed and for which hash that happened. Used `tailRecM` in hashes chain discovery in order to not propagate the error up through the call stack failing all the enclosing invocations and making the logs impossible to read. Added log with terminal discovered hash treated as full snapshot hash.

### DIFF
--- a/modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotTraverse.scala
+++ b/modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotTraverse.scala
@@ -1,18 +1,24 @@
 package org.tessellation.infrastructure.snapshot
 
-import cats._
 import cats.data.NonEmptyChain
+import cats.effect.Sync
+import cats.syntax.applicativeError._
+import cats.syntax.either._
 import cats.syntax.flatMap._
 import cats.syntax.foldable._
 import cats.syntax.functor._
 import cats.syntax.option._
 import cats.syntax.show._
-import cats.syntax.traverse._
 
+import scala.util.control.NoStackTrace
+
+import org.tessellation.ext.cats.syntax.partialPrevious.catsSyntaxPartialPrevious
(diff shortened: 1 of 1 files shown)


modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotTraverse.scala
```diff
@@ -1,50 +1,73 @@
...
           }
 
         for {
-          rollbackInc <- loadIncOrErr(rollbackHash)
-          (globalHash, incHashesNec) <- hashChain(rollbackInc.lastSnapshotHash).map { nec =>
-            (nec.head, NonEmptyChain.fromChainAppend(nec.tail, rollbackHash))
-          }
-          global <- loadFullOrErr(globalHash)
+          (globalHashCandidate, incHashesNec) <- discoverHashesChain(rollbackHash)
+          global <- loadFullOrErr(globalHashCandidate)
```
sdavidp, line 70: Am I correct that the order of these operations has changed? Is that ok?

sdavidp (commented):
Left one comment, otherwise LGTM

modules/core/src/main/scala/org/tessellation/infrastructure/snapshot/GlobalSnapshotTraverse.scala
TheMMaciek: I don't think the order changed. What is different here is that we don't need to try loading the rollback incremental hash first to get an error if it's not present. That will now happen in the `discoverHashesChain` function - a case where only one hash is there at the end in the NEC which effectively means that we didn't find rollback snapshot and that's why we don't have it's predecessor. Other than that I changed names a bit, because the globalHash is a candidate rather than a sure thing.

sdavidp (approved):

PR #772, opened 2023-11-09
Add method to process accepted snapshots on metagraph

sdavidp:
This PR also fixes a typo introduced in a previous merge.


modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
```diff
@@ -49,7 +52,8 @@ object CurrencySnapshotConsensusFunctions {
       signedArtifact: Signed[CurrencyIncrementalSnapshot],
       context: CurrencySnapshotContext
     ): F[Unit] =
-      stateChannelSnapshotService.consume(signedArtifact, context)
+      stateChannelSnapshotService.consume(signedArtifact, context) >>
+        maybeDataApplication.traverse_(_.onSnapshotAcceptedLive(signedArtifact))
```
marcinwadon, line 56: it should be rather `onConsensusResult` or `consumeSnapshotAfterConsensus` or something like that. it isn't equal to snapshot acceptance.

modules/shared/src/main/scala/org/tessellation/currency/dataApplication/package.scala
```diff
@@ -87,6 +87,8 @@ trait BaseDataApplicationL0Service[F[_]] extends BaseDataApplicationService[F] w
   def genesis: DataState.Base
 
   final def serializedOnChainGenesis: F[Array[Byte]] = serializeState(genesis.onChain)
+
+  def onSnapshotAcceptedLive(snapshot: Signed[CurrencyIncrementalSnapshot]): F[Unit]
```
marcinwadon, line 91: IMO this function should be optional to be implemented, with fallback to `noop` operation

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
sdavidp: @marcinwadon I can't find reference to either `onConsensusResult` or `consumeSnapshot`. Where are you suggesting this change go?

marcinwadon: I meant the function name

marcinwadon: onSnapshotAcceptedLive seems odd to me. It's not triggered on snapshot acceptance, and what does "live" mean in this context?

modules/shared/src/main/scala/org/tessellation/currency/dataApplication/package.scala
sdavidp: @marcinwadon Where do you suggest the default implementation go?

marcinwadon: what do you mean @sdavidp ?

marcinwadon: you mean this?

```scala
def onSnapshotAcceptedLive(snapshot: Signed[CurrencyIncrementalSnapshot]): F[Unit] = Applicative[F].unit
```

sdavidp: @marcinwadon Yes but I can't do that because `Applicative` is not available to me on the trait.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
sdavidp: @ryle-goehausen  ☝️

TheMMaciek: I agree that the name seems odd. Could we follow the convention we already have? So maybe: `consumeMajorityArtifact`? Especially that it takes the artifact as input. `onSnapshotAcceptedLive` sounds new and exotic but maybe it should sound plain and keep to the convention rather.

modules/shared/src/main/scala/org/tessellation/currency/dataApplication/package.scala
TheMMaciek: Not sure if I understand correctly but for me we can require the user to supply this implementation and it can be a simple unit like in the example above if nothing needs to be done which is an equivalent of noop.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
sdavidp: Renamed function to `consumeMajorityArtifact`

modules/shared/src/main/scala/org/tessellation/currency/dataApplication/package.scala
sdavidp: @TheMMaciek That was my expectation as well so I don't understand the original comment. @marcinwadon Can you please be more specific?

marcinwadon: Yep. We may specific more "handlers", like:
- onSnapshotAcceptance,
- onConsumeSnapshot,
- onArtifactCreated,
and so on.

They all should be optional to be implemented by the end user. One option would be to specify them as Option but it does not make sense for me. It's better to deliver default implementation "Noop", and if anyone wants to bind to the handler, one can provide own implementation. It does make not sense to require end user to provide all implementations for all the handlers. The easiest way would be to require the Applicative (can be directly for that function) which specifies "pure" method. For someone using IO it would be IO[Unit], for someone else it may be Id so just ().

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
marcinwadon: @TheMMaciek @sdavidp i would stick to the convention "onSomething" for all the handlers. We may add more in the future. What do you think?

The same way like you can find on web: onChange, onClick, onDoubleClick, and so on

modules/shared/src/main/scala/org/tessellation/currency/dataApplication/package.scala
sdavidp: @marcinwadon Yes I understand all that. My question is where do you want the default _noop_ implementation to live? Context bounds are not permitted in a `trait` so it can't be there. The only options I can think of is to turn one of the traits to an abstract class or to provide the context bounds on an _apply_.

marcinwadon: Not apply but Applicative. You can add it directly to the function that requires it, so the one you've just added

modules/shared/src/main/scala/org/tessellation/currency/dataApplication/package.scala
```diff
@@ -130,10 +132,12 @@ trait DataApplicationContextualOps[F[_], D <: DataUpdate, DON <: DataOnChainStat
   def routesPrefix: ExternalUrlPrefix = "/data-application"
 }
 
-trait DataApplicationL0Service[F[_], D <: DataUpdate, DON <: DataOnChainState, DOF <: DataCalculatedState]
+abstract class DataApplicationL0Service[F[_]: Applicative, D <: DataUpdate, DON <: DataOnChainState, DOF <: DataCalculatedState]
```
marcinwadon, line 135: It should stay a trait. Please see me previous comment

marcinwadon (changes requested):

modules/shared/src/main/scala/org/tessellation/currency/dataApplication/package.scala
sdavidp: @marcinwadon If it's a trait then we can't have the `Applicative[F].unit` implementation here, which is why I asked where you want the implementation to go.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
ryle-goehausen: The reason for naming it 'snapshot' is because, from the perspective of the DataApplication (the only place we currently need this,) it's always going to be a currency snapshot. So it's less abstract than an arbitrary artifact since this is a specific implementation. (Since we're making this change in CurrencySnapshotConsensusFns.)

Consume is also okay as nomenclature, I suggested inserting it here because it seemed like an easy place (to avoid any interactions with downloads or overlap in other 'acceptance' style logic,) so maybe that's a better name than acceptedLive? The live was meant to distinguish between the case where we're accepting a snapshot due to a download or rollback or anything else -- other than a live result of a consensus process (this is important in the application context.)

'Consensus' is an appropriate renaming of 'live', `onConsensusResult` is fine here, I do agree with the `on` standard for application handlers (especially when they're so isolated like this.) But you could also insert snapshot in the name since it's not a specific artifact. I.e. `onSnapshotConsensusResult`

modules/shared/src/main/scala/org/tessellation/currency/dataApplication/package.scala
ryle-goehausen: Yeah we looked over this, if we have another type argument in the function, then we need to translate from G (function arg) to F anyways (hence the problem with the trait). I'm not seeing a way around this otherwise? I do agree with a default from a user perspective being easier (see the example route having a specific impl in the trait,) but the problem here is with the type bounds.

I'm fine with abstract class here if we can't find a cleaner way around this.

ryle-goehausen: If you can supply another alternative marcin happy to use this here (which translates from the type bounds on the function type param to the trait param.)

marcinwadon: @ryle-goehausen I've already written it twice somewhere in comments. In general, it's as simple as:
```
trait DataApplicationL0Service[F[_], D <: DataUpdate, ...] {

  def onConsumeArtifact(snapshot: Signed[CurrencyIncrementalSnapshot])(implicit A: Applicative[F]): F[Unit] = A.unit

}
```

marcinwadon: @sdavidp /\

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
marcinwadon: > The reason for naming it 'snapshot' is because, from the perspective of the DataApplication (the only place we currently need this,) it's always going to be a currency snapshot. So it's less abstract than an arbitrary artifact since this is a specific implementation. (Since we're making this change in CurrencySnapshotConsensusFns.)

No problem, it can be a `snapshot`.

> The live was meant to distinguish between the case where we're accepting a snapshot due to a download or rollback or anything else -- other than a live result of a consensus process (this is important in the application context.)

I'd keep using the unified nomenclature across the codebase. We've never used `Live` for such distinguish and it will be misleading.

@ryle-goehausen I agree with you. I'm fine with both `onConsensusResult` or `onSnapshotConsensusResult`, whatever :) just not `acceptance` since it's not triggered from/during acceptance logic and not `live` as it may be misleading.

marcinwadon: @sdavidp /\

modules/shared/src/main/scala/org/tessellation/currency/dataApplication/package.scala
```diff
@@ -134,6 +136,8 @@ trait DataApplicationL0Service[F[_], D <: DataUpdate, DON <: DataOnChainState, D
     extends DataApplicationService[F, D, DON, DOF]
     with DataApplicationContextualOps[F, D, DON, DOF, L0NodeContext[F]] {
   def genesis: DataState[DON, DOF]
+
+  def onConsumeArtifact(snapshot: Signed[CurrencyIncrementalSnapshot])(implicit G: Applicative[F]): F[Unit] = G.unit
```
marcinwadon, line 140: Please name it "A" as Applicative or just "applicative: Applicative". The same pattern with M: Monad, and so on. F and G may be misleading as F and G are usually used for generic type bounds

modules/shared/src/main/scala/org/tessellation/currency/dataApplication/package.scala
```diff
@@ -134,6 +136,8 @@ trait DataApplicationL0Service[F[_], D <: DataUpdate, DON <: DataOnChainState, D
     extends DataApplicationService[F, D, DON, DOF]
     with DataApplicationContextualOps[F, D, DON, DOF, L0NodeContext[F]] {
   def genesis: DataState[DON, DOF]
+
+  def onSnapshotConsensusResult(snapshot: Signed[CurrencyIncrementalSnapshot])(implicit A: Applicative[F]): F[Unit] = A.unit
```
scasplte2, line 140: @ryle-goehausen just to confirm, is this where you want me to eventually trigger the snapshot push to the external EVM?

TheMMaciek (approved):
LGTM

marcinwadon (approved):

PR #778, opened 2023-11-16
Added double-sign detection.

cngo-github:
Added double-sign detection mechanism.

### DIFF
--- a/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/modules/Daemons.scala
+++ b/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/modules/Daemons.scala
@@ -10,6 +10,7 @@ import org.tessellation.currency.l0.config.types.AppConfig
 import org.tessellation.currency.l0.snapshot.CurrencySnapshotEventsPublisherDaemon
 import org.tessellation.node.shared.domain.Daemon
 import org.tessellation.node.shared.domain.healthcheck.HealthChecks
+import org.tessellation.node.shared.domain.snapshot.DoubleSignDetect
 import org.tessellation.node.shared.infrastructure.cluster.daemon.NodeStateDaemon
 import org.tessellation.node.shared.infrastructure.collateral.daemon.CollateralDaemon
 import org.tessellation.node.shared.infrastructure.healthcheck.daemon.HealthCheckDaemon
@@ -41,7 +42,8 @@ object Daemons {
       DownloadDaemon.make(storages.node, programs.download, peerDiscoveryDelay),
       HealthCheckDaemon.make(healthChecks),
       CurrencySnapshotEventsPublisherDaemon.make(queues.l1Output, services.gossip, maybeDataApplication),
-      CollateralDaemon.make(services.collateral, storages.snapshot, storages.cluster)
+      CollateralDaemon.make(services.collateral, storages.snapshot, storages.cluster),
+      DoubleSignDetect.daemon(services.doubleSignDetect, config.doubleSignDetectDaemon)
     ).traverse(_.start).void
   }
 
--- a/modules/currency-l1/src/main/scala/org/tessellation/currency/l1/cli/method.scala
+++ b/modules/currency-l1/src/main/scala/org/tessellation/currency/l1/cli/method.scala
@@ -60,7 +60,10 @@ object method {
         pullTxsCount = 100L
       ),
       healthCheck = healthCheckConfig(false),
-      collateral = collateralConfig(environment, collateralAmount)
+      collateral = collateralConfig(environment, collateralAmount),
+      doubleSignDetectDaemon = DoubleSignDetectDaemonConfig(
+        runInterval = 1.minute
+      )
     )
 
     val stateChannelAllowanceLists = None
--- a/modules/dag-l0/src/main/scala/org/tessellation/dag/l0/modules/Daemons.scala
+++ b/modules/dag-l0/src/main/scala/org/tessellation/dag/l0/modules/Daemons.scala
@@ -9,6 +9,7 @@ import org.tessellation.dag.l0.config.types.AppConfig
 import org.tessellation.dag.l0.infrastructure.snapshot.GlobalSnapshotEventsPublisherDaemon
 import org.tessellation.dag.l0.infrastructure.trust.TrustStorageUpdater
 import org.tessellation.node.shared.domain.Daemon
+import org.tessellation.node.shared.domain.snapshot.DoubleSignDetect
 import org.tessellation.node.shared.infrastructure.cluster.daemon.NodeStateDaemon
 import org.tessellation.node.shared.infrastructure.collateral.daemon.CollateralDaemon
 import org.tessellation.node.shared.infrastructure.healthcheck.daemon.HealthCheckDaemon
@@ -43,7 +44,8 @@ object Daemons {
       HealthCheckDaemon.make(healthChecks),
       GlobalSnapshotEventsPublisherDaemon.make(queues.stateChannelOutput, queues.l1Output, services.gossip),
       CollateralDaemon.make(services.collateral, storages.globalSnapshot, storages.cluster),
-      TrustStorageUpdater.daemon(services.trustStorageUpdater)
+      TrustStorageUpdater.daemon(services.trustStorageUpdater),
+      DoubleSignDetect.daemon(services.doubleSignDetect, cfg.doubleSignDetectDaemon)
     ).traverse(_.start).void
   }
 
--- a/modules/node-shared/src/main/scala/org/tessellation/node/shared/cli/CliMethod.scala
+++ b/modules/node-shared/src/main/scala/org/tessellation/node/shared/cli/CliMethod.scala
@@ -61,6 +61,10 @@ object CliMethod {
 
   val forkInfoStorageConfig: ForkInfoStorageConfig = ForkInfoStorageConfig(10)
 
+  val doubleSignDetectConfig: DoubleSignDetectConfig = DoubleSignDetectConfig(
+    maxDetectionDelta = 10L
+  )
+
   val leavingDelay = 30.seconds
 
   def healthCheckConfig(pingEnabled: Boolean) = HealthCheckConfig(
@@ -114,6 +118,8 @@ trait CliMethod {
 
   val forkInfoStorageConfig: ForkInfoStorageConfig = CliMethod.forkInfoStorageConfig
 
+  val doubleSignDetectConfig: DoubleSignDetectConfig = CliMethod.doubleSignDetectConfig
+
   lazy val nodeSharedConfig: SharedConfig = SharedConfig(
     environment,
     gossipConfig,
@@ -124,7 +130,8 @@ trait CliMethod {
     trustStorageConfig,
     PriorityPeerIds.get(environment),
     snapshotSizeConfig,
-    forkInfoStorageConfig
+    forkInfoStorageConfig,
+    doubleSignDetectConfig
   )
 
 }
--- a/modules/node-shared/src/main/scala/org/tessellation/node/shared/config/types.scala
+++ b/modules/node-shared/src/main/scala/org/tessellation/node/shared/config/types.scala
@@ -25,7 +25,8 @@ object types {
     trustStorage: TrustStorageConfig,
     priorityPeerIds: Option[NonEmptySet[PeerId]],
     snapshotSizeConfig: SnapshotSizeConfig,
-    forkInfoStorage: ForkInfoStorageConfig
+    forkInfoStorage: ForkInfoStorageConfig,
+    doubleSignDetect: DoubleSignDetectConfig
   )
 
   case class SnapshotSizeConfig(
@@ -125,5 +126,11 @@ object types {
   case class ForkInfoStorageConfig(
     maxSize: PosInt
   )
+
+  case class DoubleSignDetectConfig(
+    maxDetectionDelta: PosLong
+  )
+
+  case class DoubleSignDetectDaemonConfig(runInterval: FiniteDuration)
 
 }
--- a/modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/snapshot/DetectedDoubleSign.scala
+++ b/modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/snapshot/DetectedDoubleSign.scala
@@ -1,0 +1,18 @@
+package org.tessellation.node.shared.domain.snapshot
+
+import org.tessellation.schema.SnapshotOrdinal
+import org.tessellation.schema.peer.PeerId
+import org.tessellation.security.hash.Hash
+
+import derevo.cats.{eqv, show}
+import derevo.derive
+import eu.timepit.refined.cats._
+import eu.timepit.refined.types.numeric.NonNegLong
+
+@derive(eqv, show)
+case class DetectedDoubleSign(
+  peerId: PeerId,
+  ordinal: SnapshotOrdinal,
+  delta: NonNegLong,
+  hashes: Seq[Hash]
+)
--- a/modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/snapshot/DoubleSignDetect.scala
+++ b/modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/snapshot/DoubleSignDetect.scala
@@ -1,0 +1,81 @@
+package org.tessellation.node.shared.domain.snapshot
+
+import cats.effect.Async
+import cats.effect.std.Supervisor
+import cats.syntax.all._
+
+import org.tessellation.node.shared.config.types.{DoubleSignDetectConfig, DoubleSignDetectDaemonConfig}
+import org.tessellation.node.shared.domain.Daemon
+import org.tessellation.node.shared.domain.fork.{ForkInfo, ForkInfoEntries, ForkInfoStorage}
+import org.tessellation.node.shared.domain.snapshot.DoubleSignDetect.IndexedForkInfo
+import org.tessellation.schema.peer.PeerId
+
+import eu.timepit.refined.api.Refined
+import eu.timepit.refined.auto._
+import eu.timepit.refined.numeric.NonNegative
+import eu.timepit.refined.types.numeric.NonNegLong
+import org.typelevel.log4cats.SelfAwareStructuredLogger
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+
+object DoubleSignDetect {
+  case class IndexedForkInfo(info: ForkInfo, idx: NonNegLong)
+
+  def daemon[F[_]: Async: Supervisor](
+    doubleSignDetect: DoubleSignDetect[F],
+    config: DoubleSignDetectDaemonConfig
+  ): Daemon[F] =
+    Daemon.periodic(doubleSignDetect.run, config.runInterval)
+}
+
+class DoubleSignDetect[F[_]: Async](
+  storage: ForkInfoStorage[F],
+  config: DoubleSignDetectConfig
+) {
+  private val logger: SelfAwareStructuredLogger[F] = Slf4jLogger.getLogger
+
+  private def isDoubleSign(previous: IndexedForkInfo, current: IndexedForkInfo): Boolean =
+    (current.idx - previous.idx <= config.maxDetectionDelta) && (previous.info.hash =!= current.info.hash)
+
+  private def hasDoubleSign(info: Iterable[IndexedForkInfo]): Boolean =
+    info.toList
+      .sortBy(_.idx.value)
+      .tailRecM[Option, Boolean] {
+        case Nil | _ :: Nil                                => false.asRight.some
+        case prev :: curr :: _ if isDoubleSign(prev, curr) => true.asRight.some
+        case _ :: tail                                     => tail.asLeft.some
+      }
+      .contains(true)
+
+  def validate(peerId: PeerId, entries: ForkInfoEntries): Option[DetectedDoubleSign] = {
+    val grouped: Iterable[Iterable[IndexedForkInfo]] =
+      entries.getEntries.zipWithIndex.map {
+        case (info, idx) =>
+          val refinedIdx = Refined.unsafeApply[Long, NonNegative](idx.toLong)
+
+          IndexedForkInfo(info, refinedIdx)
+      }
+        .groupBy(_.info.ordinal)
+        .values
+
+    grouped.find(hasDoubleSign).map { found =>
+      val deltaSum = found.lazyZip(found.tail).map { case (x, y) => y.idx - x.idx }.sum
+
+      DetectedDoubleSign(
+        peerId = peerId,
+        ordinal = found.head.info.ordinal,
+        delta = Refined.unsafeApply[Long, NonNegative](deltaSum),
+        hashes = found.map(_.info.hash).toSeq
+      )
+    }
+  }
+
+  def run: F[Unit] = storage.getForkInfo.flatMap {
+    _.forks.toList.traverse_ {
+      case (id, entries) =>
+        validate(id, entries).traverse_ { detection =>
+          logger.warn(s"Found a double signing: ${detection.show}")
+        }
+    }
+  }
+
+}
--- a/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/cli/method.scala
+++ b/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/cli/method.scala
@@ -61,7 +61,10 @@ object method {
         additionalDiscoveryDelay = 0.minutes,
         minPeers = 1
       ),
-      snapshotSizeConfig = snapshotSizeConfig
+      snapshotSizeConfig = snapshotSizeConfig,
+      doubleSignDetectDaemon = DoubleSignDetectDaemonConfig(
+        runInterval = 1.minute
+      )
     )
 
     val stateAfterJoining: NodeState = NodeState.WaitingForDownload
--- a/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/config/types.scala
+++ b/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/config/types.scala
@@ -14,6 +14,7 @@ object types {
     snapshot: SnapshotConfig,
     collateral: CollateralConfig,
     globalL0Peer: L0Peer,
-    snapshotSizeConfig: SnapshotSizeConfig
+    snapshotSizeConfig: SnapshotSizeConfig,
+    doubleSignDetectDaemon: DoubleSignDetectDaemonConfig
   )
 }
--- a/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/modules/Services.scala
+++ b/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/modules/Services.scala
@@ -22,6 +22,7 @@ import org.tessellation.node.shared.domain.gossip.Gossip
 import org.tessellation.node.shared.domain.healthcheck.LocalHealthcheck
 import org.tessellation.node.shared.domain.rewards.Rewards
 import org.tessellation.node.shared.domain.seedlist.SeedlistEntry
+import org.tessellation.node.shared.domain.snapshot.DoubleSignDetect
 import org.tessellation.node.shared.domain.snapshot.services.{AddressService, GlobalL0Service}
 import org.tessellation.node.shared.infrastructure.collateral.Collateral
 import org.tessellation.node.shared.infrastructure.metrics.Metrics
@@ -128,7 +129,8 @@ object Services {
         snapshotContextFunctions = sharedServices.currencySnapshotContextFns,
         dataApplication = maybeDataApplication,
         globalSnapshotContextFunctions = globalSnapshotContextFns,
-        sentStateChannelBinaryTrackingService = sentStateChannelBinaryTrackingService
+        sentStateChannelBinaryTrackingService = sentStateChannelBinaryTrackingService,
+        doubleSignDetect = sharedServices.doubleSignDetect
       ) {}
 }
 
@@ -150,5 +152,6 @@ sealed abstract class Services[F[_]] private (
   val snapshotContextFunctions: CurrencySnapshotContextFunctions[F],
   val dataApplication: Option[BaseDataApplicationL0Service[F]],
   val globalSnapshotContextFunctions: GlobalSnapshotContextFunctions[F],
-  val sentStateChannelBinaryTrackingService: SentStateChannelBinaryTrackingService[F]
+  val sentStateChannelBinaryTrackingService: SentStateChannelBinaryTrackingService[F],
+  val doubleSignDetect: DoubleSignDetect[F]
 )
--- a/modules/currency-l1/src/main/scala/org/tessellation/currency/l1/CurrencyL1App.scala
+++ b/modules/currency-l1/src/main/scala/org/tessellation/currency/l1/CurrencyL1App.scala
@@ -148,7 +148,7 @@ abstract class CurrencyL1App(
         blockRumorHandler[IO](queues.peerBlock)
 
       _ <- DAGL1Daemons
-        .start(storages, services, healthChecks)
+        .start(storages, services, healthChecks, cfg)
         .asResource
 
       implicit0(nodeContext: L1NodeContext[IO]) = L1NodeContext.make[IO](storages.lastGlobalSnapshot, storages.lastSnapshot)
--- a/modules/currency-l1/src/main/scala/org/tessellation/currency/l1/modules/Services.scala
+++ b/modules/currency-l1/src/main/scala/org/tessellation/currency/l1/modules/Services.scala
@@ -56,6 +56,7 @@ object Services {
       val transaction = TransactionService.make[F](storages.transaction, validators.transactionContextual)
       val collateral = Collateral.make[F](cfg.collateral, storages.lastSnapshot)
       val dataApplication = maybeDataApplication
+      val doubleSignDetect = sharedServices.doubleSignDetect
     }
 }
 
--- a/modules/dag-l0/src/main/scala/org/tessellation/dag/l0/cli/method.scala
+++ b/modules/dag-l0/src/main/scala/org/tessellation/dag/l0/cli/method.scala
@@ -73,6 +73,9 @@ object method {
         checkPeersMaxDelay = 10.minutes,
         additionalDiscoveryDelay = 0.minutes,
         minPeers = 1
+      ),
+      doubleSignDetectDaemon = DoubleSignDetectDaemonConfig(
+        runInterval = 1.minute
       )
     )
 
--- a/modules/dag-l0/src/main/scala/org/tessellation/dag/l0/config/types.scala
+++ b/modules/dag-l0/src/main/scala/org/tessellation/dag/l0/config/types.scala
@@ -34,7 +34,8 @@ object types {
     rewards: RewardsConfig,
     stateChannelPullDelay: NonNegLong,
     stateChannelPurgeDelay: NonNegLong,
-    peerDiscoveryDelay: PeerDiscoveryDelay
+    peerDiscoveryDelay: PeerDiscoveryDelay,
+    doubleSignDetectDaemon: DoubleSignDetectDaemonConfig
   )
 
   case class DBConfig(
--- a/modules/dag-l0/src/main/scala/org/tessellation/dag/l0/modules/Services.scala
+++ b/modules/dag-l0/src/main/scala/org/tessellation/dag/l0/modules/Services.scala
@@ -22,6 +22,7 @@ import org.tessellation.node.shared.domain.gossip.Gossip
 import org.tessellation.node.shared.domain.healthcheck.LocalHealthcheck
 import org.tessellation.node.shared.domain.rewards.Rewards
 import org.tessellation.node.shared.domain.seedlist.SeedlistEntry
+import org.tessellation.node.shared.domain.snapshot.DoubleSignDetect
 import org.tessellation.node.shared.domain.snapshot.services.AddressService
 import org.tessellation.node.shared.infrastructure.collateral.Collateral
 import org.tessellation.node.shared.infrastructure.consensus.Consensus
@@ -96,7 +97,8 @@ object Services {
         collateral = collateralService,
         rewards = rewards,
         stateChannel = stateChannelService,
-        trustStorageUpdater = trustUpdaterService
+        trustStorageUpdater = trustUpdaterService,
+        doubleSignDetect = sharedServices.doubleSignDetect
       ) {}
 }
 
@@ -110,5 +112,6 @@ sealed abstract class Services[F[_]] private (
   val collateral: Collateral[F],
   val rewards: Rewards[F, GlobalSnapshotStateProof, GlobalIncrementalSnapshot, GlobalSnapshotEvent],
   val stateChannel: StateChannelService[F],
-  val trustStorageUpdater: TrustStorageUpdater[F]
+  val trustStorageUpdater: TrustStorageUpdater[F],
+  val doubleSignDetect: DoubleSignDetect[F]
 )
--- a/modules/dag-l1/src/main/scala/org/tessellation/dag/l1/Main.scala
+++ b/modules/dag-l1/src/main/scala/org/tessellation/dag/l1/Main.scala
@@ -106,7 +106,7 @@ object Main
         blockRumorHandler(queues.peerBlock)
 
       _ <- Daemons
-        .start(storages, services, healthChecks)
+        .start(storages, services, healthChecks, cfg)
         .asResource
 
       api = HttpApi
--- a/modules/dag-l1/src/main/scala/org/tessellation/dag/l1/cli/method.scala
+++ b/modules/dag-l1/src/main/scala/org/tessellation/dag/l1/cli/method.scala
@@ -55,7 +55,10 @@ object method {
         pullTxsCount = 100L
       ),
       healthCheck = healthCheckConfig(false),
-      collateral = collateralConfig(environment, collateralAmount)
+      collateral = collateralConfig(environment, collateralAmount),
+      doubleSignDetectDaemon = DoubleSignDetectDaemonConfig(
+        runInterval = 1.minute
+      )
     )
 
     val stateChannelAllowanceLists = None
--- a/modules/dag-l1/src/main/scala/org/tessellation/dag/l1/config/types.scala
+++ b/modules/dag-l1/src/main/scala/org/tessellation/dag/l1/config/types.scala
@@ -15,7 +15,8 @@ object types {
     gossip: GossipConfig,
     consensus: ConsensusConfig,
     healthCheck: HealthCheckConfig,
-    collateral: CollateralConfig
+    collateral: CollateralConfig,
+    doubleSignDetectDaemon: DoubleSignDetectDaemonConfig
   )
 
   case class DBConfig(
--- a/modules/dag-l1/src/main/scala/org/tessellation/dag/l1/modules/Daemons.scala
+++ b/modules/dag-l1/src/main/scala/org/tessellation/dag/l1/modules/Daemons.scala
@@ -5,8 +5,10 @@ import cats.effect.std.Supervisor
 import cats.syntax.functor._
 import cats.syntax.traverse._
 
+import org.tessellation.dag.l1.config.types.AppConfig
 import org.tessellation.dag.l1.infrastructure.healthcheck.HealthCheckDaemon
 import org.tessellation.node.shared.domain.Daemon
+import org.tessellation.node.shared.domain.snapshot.DoubleSignDetect
 import org.tessellation.node.shared.infrastructure.cluster.daemon.NodeStateDaemon
 import org.tessellation.node.shared.infrastructure.collateral.daemon.CollateralDaemon
 import org.tessellation.schema.snapshot.{Snapshot, SnapshotInfo, StateProof}
@@ -21,12 +23,14 @@ object Daemons {
   ](
     storages: Storages[F, P, S, SI],
     services: Services[F, P, S, SI],
-    healthChecks: HealthChecks[F]
+    healthChecks: HealthChecks[F],
+    cfg: AppConfig
   ): F[Unit] =
     List[Daemon[F]](
       NodeStateDaemon.make(storages.node, services.gossip),
       CollateralDaemon.make(services.collateral, storages.lastSnapshot, storages.cluster),
-      HealthCheckDaemon.make(healthChecks)
+      HealthCheckDaemon.make(healthChecks),
+      DoubleSignDetect.daemon(services.doubleSignDetect, cfg.doubleSignDetectDaemon)
     ).traverse(_.start).void
 
 }
--- a/modules/dag-l1/src/main/scala/org/tessellation/dag/l1/modules/Services.scala
+++ b/modules/dag-l1/src/main/scala/org/tessellation/dag/l1/modules/Services.scala
@@ -12,6 +12,7 @@ import org.tessellation.node.shared.domain.cluster.storage.L0ClusterStorage
 import org.tessellation.node.shared.domain.collateral.Collateral
 import org.tessellation.node.shared.domain.gossip.Gossip
 import org.tessellation.node.shared.domain.healthcheck.LocalHealthcheck
+import org.tessellation.node.shared.domain.snapshot.DoubleSignDetect
 import org.tessellation.node.shared.domain.snapshot.services.GlobalL0Service
 import org.tessellation.node.shared.domain.snapshot.storage.LastSnapshotStorage
 import org.tessellation.node.shared.infrastructure.block.processing.BlockAcceptanceManager
@@ -55,6 +56,7 @@ object Services {
       val session = sharedServices.session
       val transaction = TransactionService.make[F](storages.transaction, validators.transactionContextual)
       val collateral = Collateral.make[F](cfg.collateral, storages.lastSnapshot)
+      val doubleSignDetect = sharedServices.doubleSignDetect
     }
 }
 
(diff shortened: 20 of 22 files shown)


modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/security/DoubleSignDetectRunner.scala
```diff
@@ -0,0 +1,37 @@
...
+object DoubleSignDetectRunner {
+
+  val doubleSignDetectLoggerName = "DoubleSignDetectLogger"
+
+  def make[F[_]: Async](storage: ForkInfoStorage[F], detect: DoubleSignDetect[F]): DoubleSignDetectRunner[F] =
+    new DoubleSignDetectRunner[F] {
+
+      val logger = Slf4jLogger.getLoggerFromName[F](doubleSignDetectLoggerName)
+
+      override def run: F[Unit] = storage.getForkInfo.flatMap {
```
marcinwadon, line 25: what it overrides?

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/security/DoubleSignDetectRunner.scala
```diff
@@ -0,0 +1,37 @@
...
+        _.forks.toList.traverse_ {
+          case (id, entries) =>
+            detect.detect(id, entries).traverse_ { detection =>
```
marcinwadon, line 28: `detect.detect` should be imo `doubleSignDetector.detect` or better `doubleSignValidator.validate`

cngo-github: Changed to

```
      override def run: F[Unit] = storage.getForkInfo.flatMap {
        _.forks.toList.traverse_ {
          case (id, entries) =>
            doubleSignValidator.validate(id, entries).traverse_ { detection =>
              logger.warn(s"Found a double signing: $detection")
            }
        }
      }
```

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/security/DoubleSignDetectRunner.scala
cngo-github: Removed `override`.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/security/DoubleSignDetectRunner.scala
```diff
@@ -0,0 +1,37 @@
...
+
+import scala.concurrent.duration.DurationInt
+
+import org.tessellation.sdk.domain.Daemon
+import org.tessellation.sdk.domain.fork.ForkInfoStorage
+import org.tessellation.sdk.domain.security.{DoubleSignDetect, DoubleSignDetectRunner}
+
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+
+object DoubleSignDetectRunner {
```
marcinwadon, line 16: what does `Runner` suffix mean? is it a program, or a service? looks like a program

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/security/DoubleSignDetectRunner.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  val doubleSignDetectLoggerName = "DoubleSignDetectLogger"
```
TheMMaciek, line 18: Isn't the place of logger creation making it easy enough to identify so that we don't have to name it explicitly?

kpudlik (changes requested):

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/security/DoubleSignDetect.scala
```diff
@@ -0,0 +1,52 @@
...
+    config: DoubleSignDetectConfig
+  ): DoubleSignDetect[F] = new DoubleSignDetect[F] {
+    case class IndexedForkInfo(info: ForkInfo, idx: Int)
+
+    def isDoubleSign(previous: IndexedForkInfo, current: IndexedForkInfo): Boolean =
+      (current.idx - previous.idx <= config.minDistance) && (previous.info.hash =!= current.info.hash)
+
+    def hasDoubleSign(infos: Iterable[IndexedForkInfo]): Boolean =
+      infos.toList
+        .sortBy(_.idx)
```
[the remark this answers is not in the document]
TheMMaciek: The index is used in the function determining whether the difference is to be treated as malicious action or not. Index is used to calculate a "distance" between fork info declarations and based on that distance the found difference in hash at a given ordinal will be interpreted as malicious action or not. There are other problems with using such an index for such functionality (e.g. it is controlled by the node sending messages), but extracting it and using for sorting is required in the implementation as it is now.

TheMMaciek (changes requested):
After reviewing the semantics of this change I think that it doesn't achieve the intended functionality. I will post further explanation why I think that. I just need to put the more evolved explanation together.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/security/DoubleSignDetectRunner.scala
```diff
@@ -0,0 +1,37 @@
...
+          case (id, entries) =>
+            doubleSignValidator.validate(id, entries).traverse_ { detection =>
+              logger.warn(s"Found a double signing: $detection")
+            }
+        }
+      }
+    }
+
+  def daemon[F[_]: Async: Supervisor](runner: DoubleSignDetectRunner[F]): Daemon[F] =
+    Daemon.periodic(runner.run, 1.minute)
```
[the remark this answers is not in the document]
cngo-github: Moved to the appropriate `AppConfig`s.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/security/DoubleSignDetectRunner.scala
cngo-github: Changed to `DoubleSignDetectProgram`.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/security/DoubleSignDetectRunner.scala
cngo-github: Changed to

```
val logger = Slf4jLogger.getLogger
```

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/security/DoubleSignDetectRunner.scala
```diff
@@ -0,0 +1,37 @@
...
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+
+object DoubleSignDetectRunner {
+
+  val doubleSignDetectLoggerName = "DoubleSignDetectLogger"
+
+  def make[F[_]: Async](storage: ForkInfoStorage[F], doubleSignValidator: DoubleSignDetect[F]): DoubleSignDetectRunner[F] =
+    new DoubleSignDetectRunner[F] {
+
+      val logger = Slf4jLogger.getLoggerFromName[F](doubleSignDetectLoggerName)
```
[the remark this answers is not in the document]
cngo-github: Changed to

```
val logger = Slf4jLogger.getLogger
```

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/security/DoubleSignDetect.scala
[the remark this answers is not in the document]
cngo-github: Changed to

```
    def hasDoubleSign(infos: Iterable[IndexedForkInfo]): Boolean =
      infos.toList
        .tailRecM[Option, Boolean] {
          case Nil | _ :: Nil                                => false.asRight.some
          case prev :: curr :: _ if isDoubleSign(prev, curr) => true.asRight.some
          case _ :: tail                                     => tail.asLeft.some
        }
        .contains(true)
```

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/security/DoubleSignDetectRunner.scala
marcinwadon: Please check how other programs in the repository are constructed/named

TheMMaciek (dismissed):
Explanation on why I think the semantics of this change are not fulfilling the intended functionality:

The fork info declarations are sent by the node in question itself. Hence the malicious node can decide to simply not send any declarations and having that we will not detect anything. Even if the malicious node decided to send declarations it could send them in such a way that would let it manipulate the result on the nodes checking them. That results from the reliance on order of declarations put into the storage and the storages capacity itself.

For sure such a functionality cannot be implemented by inspecting the declarations from the node we are checking. I think such a functionality needs to be based on inspection of artifacts spread by the peers. By inspecting these artifacts we can detect whether there are two or more artifacts with the same ordinal belonging to the same chain (same previous hash) that contain signatures from the same node (so the node signed both). I don't think we need to worry about "distance" between declarations or anything like that because node is signing the artifact during consensus round where the consensus is a synchronous process and there is no chance of signing two artifacts at the same time - it would actually require a node to run two processes at the same time or in fact be malicious. The rejoin case after forking non maliciously should not trigger false positive here because a node will not sign an artifact that was already created while this node created a different artifact (will not add a signature post consensus). On following snapshots the last hash should differ I think indicating that this node forked or picked a different fork which I don't consider as the indication of being malicious. I also think that by inspecting signatures we only catch a certain kind of malicious action, we don't check the proposals that node spreads trying to fork the network. I think we could maybe change the consensus resources structure to allow for mutliple proposals to be persisted for every node (only first one will be taken into account when running consensus just like it is like now) having that we will be able to catch that case as well where a node spreads multiple snapshot proposals for the same consensus round. This case may not be visible based on the created artifact, because a node may choose to spread concurrent proposals but then only sign a single artifact.

Another thing which in my opinion makes sense is to consider the following: removing the fork info storage and only relying on the consensus resources and running all checks against those resoures - they already contain most of the information aggregated in fork info storage. The changes we would need to perform would be to keep those resources for a longer time, we would need to allow to update the resources post consensus e.g. adding additional proposals from nodes if such proposals are received after consensus ends and additionaly aggregate the artifacts for the same ordinal. It seems it would be better to have a single source of truth than aggregate information multiple times once in consensus resources, second time in fork info storage and in the future maybe many more. I would investigate this proposal a bit as it seems a reasonable solution unless we find a major flaw with it.

ryle-goehausen:
> The fork info declarations are sent by the node in question itself. Hence the malicious node can decide to simply not send any declarations and having that we will not detect anything.

We're using this information elsewhere in determining current active majority fork (apart from artifact formation,) for resetting the node -- which means it is a place you'd want to 'fake' or attack the data. So we had this discussion in slack already when marcin pointed this out. The issue is we looked into modifying the gossip routine on the final 'consumable' artifact containing signatures, which would require a much more major change to the gossip routine to listen in on it. I think we should make that change eventually, so that we know who gossiped each artifact (signed authenticated info on artifact origin,) who signed each artifact, and the receipt order for when we received it.) So while this doesn't cover the entire proper flow, it does at least cover the fork detection code.

Right now, the information we're assembling here, is identical to that info, and we can later modify the main gossip routine to produce that same information and put it into these same case classes. So it doesn't change the underlying behavior of what we want to do, just the origin of the data. That change can be made independently of this, and transformed into this structure. (I'm just re-stating what we discussed on slack earlier also here.)

>I don't think we need to worry about "distance" between declarations or anything like that because node is signing the artifact during consensus round where the consensus is a synchronous process and there is no chance of signing two artifacts at the same time

The purpose of this is to distinguish between rollback events, where nodes CAN legitimately sign the same ordinal twice due to a reversion -- versus nodes signing the same ordinal twice on purpose.

> The changes we would need to perform would be to keep those resources for a longer time, we would need to allow to update the resources post consensus e.g. adding additional proposals from nodes if such proposals are received after consensus ends and additionaly aggregate the artifacts for the same ordinal. It seems it would be better to have a single source of truth than aggregate information multiple times once in consensus resources, second time in fork info storage and in the future maybe many more.

On here, we don't actually need the artifact information -- only the data collected. So while I agree with your comment that the source of this data should be derived from the gossiped signed artifact, the fields (gossipPeerId, signerPeerId, ordinal, hash) are the only values required to be persisted, we don't actually care about what the artifact is, or any of the other steps or proposals, only the final signed value, its origin and ordinal/hash.

TheMMaciek:
> We're using this information elsewhere in determining current active majority fork (apart from artifact formation,) for resetting the node -- which means it is a place you'd want to 'fake' or attack the data. So we had this discussion in slack already when marcin pointed this out. The issue is we looked into modifying the gossip routine on the final 'consumable' artifact containing signatures, which would require a much more major change to the gossip routine to listen in on it. I think we should make that change eventually, so that we know who gossiped each artifact (signed authenticated info on artifact origin,) who signed each artifact, and the receipt order for when we received it.) So while this doesn't cover the entire proper flow, it does at least cover the fork detection code.

I know it's being used in different functionality for determining the fork. For that fork detection functionality I think that everything that needs to be known is already gathered by the consensus itself i.e. all proposals are there and we could determine whether we are forking and what is the majority fork at this stage during consensus. The reason we are not fixing it in consensus is because the fork happens in terms of signatures of the previous snapshot (and btw the fork detection functionality will not catch that right now). Forked signatures influence the rewards and as a result, even though we detect the fork during the consensus, the majority artifact is not valid in terms of the last artifact we accepted. Effectively, we accept the minority artifact (at this point we already know we choose minority artifact, so no need for additional detection mechanism). Anyway, leaving that aside, using fork info for the double sign detection is simply wrong. If a node is truly malicious then it will not spread the declarations or will spread such a declaration that we will not detect anything. For such a functionality we need to rely on the observations of other nodes in the cluster (non malicious nodes rather than a malicious one). We can inspect the signed majority artifact and draw all of our conclusions based on it. We can spread it with a gossip common rumor and draw conclusions in a way I described in the previous comment.

> The purpose of this is to distinguish between rollback events, where nodes CAN legitimately sign the same ordinal twice due to a reversion -- versus nodes signing the same ordinal twice on purpose.

Signing the same ordinal is not a proof that someone is or isn't malicious. In my opinion, the only indication of being malicious is detecting that someone signed an artifact with the same ordinal and previous hash but different contents. And therefore we should discover if there are any corner cases (if such a behaviour can be performed by a non malicious node). I think that this is the only valid case for calling a node malicious because if a node forks and then resets it's ok for it to sign an artifact at the ordinal it already processed in a different fork (especially a faster fork). For such cases an aribtrary distance is not gonna help as it will mark nodes in faster forks as malicious vs those in slower forks as not malicious and it's not right. In this scenario probably none of those nodes are malicious. Coming back to the case I consider malicious I can imagine such a case happening for an initial rollback node, but then at the same time a rollback is a network reset so we shouldn't care about what was happening before that reset (in the previous run) and we shouldn't have interoperability/communication between different network runs (and I think we don't have such interoperability). For non rollback nodes if they rejoin after rollback they need to reset their observations, so effectively they won't detect an initial rollback node or any other node that already rejoined as maliciously recreating artifacts. Now the case when a node forks, detects that after some time, and then rejoins to a different fork. The question is if there is a legitimate case where such a node can find itself joining to a consensus at ordinal it already created at a different fork and especially with the previous hash being the same. I would need to think about it a bit more, but it for sure sounds highly unlikely. Why would the node detect that it's forked if not because it created different artifact at a given ordinal. If it did detect that, why would it rejoin to a fork and somehow join a consensus with the same ordinal and previous hash as it had in the original fork (the node is not let in on the running consensus and - moreover - when joining it stays in observing mode for a couple of snapshots).

Regarding the nodes malicious behaviour detection we should also detect nodes behaviour in regards to distributed proposals (i.e. different proposal hashes sent to different groups of nodes). This shouldn't happen for the non-malicious actor as a node should only spread a single proposal per round. Such behaviour should also be checked because it's a legitimate attempt to fork the network - although because of gossip protocol it's hard for the attacking node to control which group of nodes receive which proposal first, but I think with some resources it's doable (parallel intial gossip to multiple nodes).

> On here, we don't actually need the artifact information -- only the data collected. So while I agree with your comment that the source of this data should be derived from the gossiped signed artifact, the fields (gossipPeerId, signerPeerId, ordinal, hash) are the only values required to be persisted, we don't actually care about what the artifact is, or any of the other steps or proposals, only the final signed value, its origin and ordinal/hash.

The only thing I can add here is that spreading the signed artifact with a common rumor seems like a natural thing to do and it's actually pretty performant. Given that it is a common rumor, if most of the nodes have the same signed artifact then they won't download it through gossip. Only nodes that miss the given artifact will receive it through gossip. Additionally, by having the artifact itself we can check much more and we don't rely on declarations (who sent what and when) because we have a proof upfront that someone signed an artifact. If we find two such artifacts, according to rules I formulated above, then we can detect a malicious behaviour right away (unless my assumptions turn out to be wrong). What is more, by having the actual artifact we may be able to detect much more than this singular malicious behaviour, e.g. signing an artifact that doesn't pass consistency validation.

ryle-goehausen:
>How it matters where this information is used knowing it can be faked or basically no information can be sent? I agree with @TheMMaciek that the information should be spread not by the owner but by other peers spreading the view of other nodes.

It does matter since we’re using that as a security check elsewhere, so it would have to be spread to act maliciously. It’s currently substituting as the artifact signature, which we’ll make a separate ticket to unify the source of the data, but again the logic after the fact is identical no matter the original source.

>It does not require changes to the gossip at all.

Look, this is out of scope of this discussion as we previously implemented the fork detection change due to missing data from the prior. I’d have to review again the details for why, but I believe we were trying to distinguish between the final majority accepted artifact, and the one that we had produced a signature for — we were missing information about the final artifact that was chosen, since we were only gossiping and collecting data about what was signed, not which majority was chosen. That was the main point of contention, but it’s irrelevant to this discussion, as we can make another PR to determine the correct source of data, this PR is only about what to do with the data once it’s produced in the form that we’re using here.

>As written above - there is nothing to change with the gossip. Both spread and spreadCommon are enough for all the cases.

It’s the difference between data that is currently used in consensus versus what is accepted in majority.

>How "signing the same ordinal twice" differs from "signing two artifacts at the same time"?

That is the entire purpose of this PR, to distinguish between these two types of events.

Signing the same ordinal twice would be a sequence of received events like this:

1, 2, 3, 4, 1, 2, 3, 4,

Versus events like this,

1, 1, 2, 2, 3, 3, 4, 4

(With different hashes received at each ordinal,)

The first case would be a rollback, where we’re not concurrently signing similar artifacts in receipt order, the second would be a series of concurrently signed events.

>Why? Shouldn't the misaligned node (misaligned after rollback) rejoin to the correct fork? If so, artifacts also needs to be persisted I think.

Rejoining to the fork is completely unrelated to detecting the event (this PR.)

There’s no reason yet to persist the actual artifacts because we’re not doing anything with them, the only information we need is what is persisted here to calculate if there’s an event detected, so we can take action on that event.


>For that fork detection functionality I think that everything that needs to be known is already gathered by the consensus itself i.e. all proposals are there and we could determine whether we are forking and what is the majority fork at this stage during consensus.

So yes, it’s ideal to actually use this information as well, but the problem is we have so many legitimate (normal,) events of forking due to other issues that if we penalize that behavior it’ll lead to too many false positives right now. This attack is far easier to detect accurately, since it’s extremely unlikely to be caused by a ‘regular’ issue.

>The reason we are not fixing it in consensus is because the fork happens in terms of signatures of the previous snapshot

So again, this is the main reason we’re using the final produced artifact for the fork Info, instead of the gossiped signature stage. We want to know what the fork was in terms of the FINAL accepted artifact, since collecting gossip information at the signature stage is insufficient due to the reason you just posted. This is what I mean by modifying the gossip.

>Anyway, leaving that aside, using fork info for the double sign detection is simply wrong. If a node is truly malicious then it will not spread the declarations or will spread such a declaration that we will not detect anything. 

This isn’t true if we’re using this for resetting the majority fork, if we’re making a decision about what fork to persist based on the final gossip forkInfo class, irrespective of what the artifacts are saying, the nodes will reset based on that view of the data, meaning to attack the network you would have to forge that data. I’m not going to argue that it’s better to not also gossip the artifact info and track that as well, but I don’t agree that this is “simply wrong”. If it’s used for determining the majority, then it’s useful. It’s also trivial to later modify the gossip to introduce the final artifact to account for signature differences — and then it is correct and this PR doesn’t change when we make that modification separately.

>For such a functionality we need to rely on the observations of other nodes in the cluster (non malicious nodes rather than a malicious one). We can inspect the signed majority artifact and draw all of our conclusions based on it. We can spread it with a gossip common rumor and draw conclusions in a way I described in the previous comment.

This is exactly what I’d like to do, in a subsequent PR — for now we only gossiped the information of interest which is exactly the same data we’ll use eventually. 

>Signing the same ordinal is not a proof that someone is or isn't malicious.

It is if you’re concurrently producing data of the same ordinal, see the above posted example

“
1, 2, 3, 4, 1, 2, 3, 4,

Versus events like this,

1, 1, 2, 2, 3, 3, 4, 4

“

>And therefore we should discover if there are any corner cases (if such a behaviour can be performed by a non malicious node)

This is the purpose of introducing a detection function to rule out cases of false positives with a threshold. We’d need to inspect metrics and log statements to verify correctness.

>I think that this is the only valid case for calling a node malicious because if a node forks and then resets it's ok for it to sign an artifact at the ordinal it already processed in a different fork

See the example above again, this is exactly what the PR is trying to accomplish.

>Regarding the nodes malicious behaviour detection we should also detect nodes behaviour in regards to distributed proposals (i.e. different proposal hashes sent to different groups of nodes)

I agree with the spirit of this, in general we’d want to have a very large number of different signals during consensus of a potential malicious action, the problem right now we’re dealing with is it’s hard to distinguish between a legitimate error and a malicious action, I think that’s actually a good suggestion of a next potential improvement since duplicate proposals are pretty unlikely to occur normally and have a high signal to noise ratio.


>The only thing I can add here is that spreading the signed artifact with a common rumor seems like a natural thing to do and it's actually pretty performant. Given that it is a common rumor, if most of the nodes have the same signed artifact then they won't download it through gossip. Only nodes that miss the given artifact will receive it through gossip. Additionally, by having the artifact itself we can check much more and we don't rely on declarations

So this whole discussion we can figure out in the next PR. I’d like to split this into a separate issue, the immediate purpose was to gather data (hence chuong introducing that ‘final gossip round’ on the consume handler where we’re gossiping fork info.) I also didn’t want to impact performance in any way by sending the large final produced artifact (containing signatures etc.) another time. We can discuss implementation on exact PR, separate from discussion of the signal / detection introduced here.

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/cli/method.scala
```diff
@@ -60,7 +60,8 @@ object method {
         pullTxsCount = 100L
       ),
       healthCheck = healthCheckConfig(false),
-      collateral = collateralConfig(environment, collateralAmount)
+      collateral = collateralConfig(environment, collateralAmount),
+      doubleSignDetectProgram = DoubleSignDetectProgramConfig(1.minute)
```
marcinwadon, line 64: Please follow the naming we already have in the project

marcinwadon (changes requested):

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/security/DoubleSignDetectRunner.scala
```diff
@@ -0,0 +1,7 @@
+package org.tessellation.node.shared.domain.security
+
+trait DoubleSignDetectRunner[F[_]] {
```
marcinwadon, line 3: Haven't you changed the *Runner name?

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/security/DoubleSignDetectProgram.scala
```diff
@@ -0,0 +1,34 @@
...
+  def make[F[_]: Async](storage: ForkInfoStorage[F], doubleSignValidator: DoubleSignDetect[F]): DoubleSignDetectRunner[F] =
+    new DoubleSignDetectRunner[F] {
+
+      val logger = Slf4jLogger.getLogger
+
+      def run: F[Unit] = storage.getForkInfo.flatMap {
+        _.forks.toList.traverse_ {
+          case (id, entries) =>
+            doubleSignValidator.validate(id, entries).traverse_ { detection =>
+              logger.warn(s"Found a double signing: $detection")
```
marcinwadon, line 26: Please use show instance wherever possible

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/security/DoubleSignDetectProgram.scala
```diff
@@ -0,0 +1,34 @@
...
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+
+object DoubleSignDetectProgram {
+
+  def make[F[_]: Async](storage: ForkInfoStorage[F], doubleSignValidator: DoubleSignDetect[F]): DoubleSignDetectRunner[F] =
+    new DoubleSignDetectRunner[F] {
+
+      val logger = Slf4jLogger.getLogger
+
+      def run: F[Unit] = storage.getForkInfo.flatMap {
```
marcinwadon, line 22: Does "getForkInfo" ensure ordering? I'd use SortedSet to make sure

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/cli/method.scala
cngo-github: `DoubleSignDetectProgramConfig` has been renamed `DoubleSignDetect`. `doubleSignDetectProgram` has been renamed `doubleSignDetect`.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/security/DoubleSignDetectRunner.scala
cngo-github: `DoubleSignDetectRunner` has been renamed to `DoubleSignDetect`.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/security/DoubleSignDetectProgram.scala
cngo-github: Changed to using the `show` instance.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/security/DoubleSignDetectProgram.scala
cngo-github: `getForkInfo` returns a `F[ForkInfoMap]`. `ForkInfoMap` is `Map[PeerId, ForkInfoEntries]`. `ForkInfoEntries` is a case class around `PosInt` and `Queue[ForkInfo]`. So, yes, ordering is ensured.

modules/sdk/src/main/scala/org/tessellation/sdk/infrastructure/security/DoubleSignDetectRunner.scala
cngo-github: Combined `DoubleSignDetect` and `DoubleSignDetectProgram`.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
```diff
@@ -115,7 +115,7 @@ abstract class CurrencyL0App(
         .handlers <+>
         services.consensus.handler
       _ <- Daemons
-        .start(storages, services, programs, queues, healthChecks, services.dataApplication, cfg)
+        .start(storages, services, programs, queues, healthChecks, services.dataApplication, cfg, method.nodeSharedConfig.doubleSignDetect)
```
marcinwadon, line 118: `isn't the config already passed through the `cfg`?

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/modules/Daemons.scala
```diff
@@ -24,7 +26,8 @@ object Daemons {
     queues: Queues[F],
     healthChecks: HealthChecks[F],
     maybeDataApplication: Option[BaseDataApplicationL0Service[F]],
-    config: AppConfig
+    config: AppConfig,
+    doubleSignDetectConfig: DoubleSignDetectConfig
```
marcinwadon, line 30: you already have or should have all the config above, in: `config: AppConfig`

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/security/DoubleSignDetect.scala
```diff
@@ -0,0 +1,12 @@
+package org.tessellation.node.shared.domain.security
+
+import org.tessellation.node.shared.domain.fork.ForkInfoEntries
+import org.tessellation.schema.peer.PeerId
+
+trait DoubleSignDetect[F[_]] {
```
marcinwadon, line 6: If it is a `service`, then it's fine to have a trait, but I'd move it from `domain/security` to something else, cluster or snapshot related, and to `service` directory. If it is a `program`, then there is no need to write a trait.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/security/DoubleSignDetection.scala
```diff
@@ -0,0 +1,16 @@
...
+
+import org.tessellation.schema.SnapshotOrdinal
+import org.tessellation.schema.peer.PeerId
+import org.tessellation.security.hash.Hash
+
+import derevo.cats.{eqv, show}
+import derevo.derive
+
+@derive(eqv, show)
+case class DoubleSignDetection(
```
marcinwadon, line 11: maybe `DetectedDoubleSign`?

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/security/DoubleSignDetect.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  def make[F[_]: Async](
+    storage: ForkInfoStorage[F],
+    config: DoubleSignDetectConfig
+  ): DoubleSignDetect[F] =
+    new DoubleSignDetect[F] {
+
+      val logger = Slf4jLogger.getLogger
+
+      case class IndexedForkInfo(info: ForkInfo, idx: Int)
```
marcinwadon, line 28: I believe you should use `Long` here

modules/node-shared/src/test/scala/org/tessellation/node/shared/infrastructure/security/DoubleSignDetectSuite.scala
```diff
@@ -0,0 +1,123 @@
...
+    } yield (peerId, entries)
+
+    forall(gen) {
+      case (peerId, entries) =>
+        val detect = mkDoubleSignDetect(threshold)
+
+        val expected: Option[DoubleSignDetection] = none
+        val actual = detect.validate(peerId, entries)
+
+        expect.eql(expected, actual)
```
marcinwadon, line 71: maybe just:
```
val result = result.validate(peerId, entries)

expect(result.isEmpty)
```
?

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/modules/Daemons.scala
cngo-github: Moved all configuration for the daemon to `DoubleSignDetectDaemonConfig`, which are included in the various `AppConfig`s.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
cngo-github: Moved all configuration for the daemon to `DoubleSignDetectDaemonConfig`, which are included in the various `AppConfig`s.

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/modules/Daemons.scala
```diff
@@ -24,7 +26,8 @@ object Daemons {
     queues: Queues[F],
     healthChecks: HealthChecks[F],
     nodeId: PeerId,
-    cfg: AppConfig
+    cfg: AppConfig,
+    doubleSignDetectConfig: DoubleSignDetectConfig
```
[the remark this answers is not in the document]
cngo-github: Moved all configuration for the daemon to `DoubleSignDetectDaemonConfig`, which are included in the various `AppConfig`s.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/security/DoubleSignDetect.scala
cngo-github: Per our offline discussion, it is a Program. The trait has been removed, `DoubleSignDetect`'s implementation has been remade into a class and moved out of `infrastructure` and into `domain`.

modules/node-shared/src/test/scala/org/tessellation/node/shared/infrastructure/security/DoubleSignDetectSuite.scala
cngo-github: Changed to

```
val result = detect.validate(peerId, entries)

expect(result.empty)
```

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/security/DoubleSignDetect.scala
cngo-github: Changed to `Long`

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/security/DoubleSignDetection.scala
cngo-github: Renamed to `DetectedDoubleSign` and moved to `domain.snapshot`.

modules/node-shared/src/main/scala/org/tessellation/node/shared/cli/CliMethod.scala
```diff
@@ -61,6 +61,8 @@ object CliMethod {
 
   val forkInfoStorageConfig: ForkInfoStorageConfig = ForkInfoStorageConfig(10)
 
+  val doubleSignDetectConfig: DoubleSignDetectConfig = DoubleSignDetectConfig(10, 1.minute)
```
marcinwadon, line 64: it would be great to explicitly name arguments (as you can see above in the file)

modules/node-shared/src/main/scala/org/tessellation/node/shared/config/types.scala
```diff
@@ -126,4 +127,11 @@ object types {
     maxSize: PosInt
   )
 
+  case class DoubleSignDetectConfig(
+    minDistance: PosInt,
```
marcinwadon, line 131: what does it mean? is it a min distance to start the detection, max distance after which double signs are ignored?

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/security/DoubleSignDetect.scala
```diff
@@ -0,0 +1,72 @@
+package org.tessellation.node.shared.domain.security
```
marcinwadon, line 1: I'd upvote my previous comment re: package name. IMO it should be snapshot or consensus-related, not in the general security

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/snapshot/DetectedDoubleSign.scala
```diff
@@ -0,0 +1,16 @@
...
+import org.tessellation.security.hash.Hash
+
+import derevo.cats.{eqv, show}
+import derevo.derive
+
+@derive(eqv, show)
+case class DetectedDoubleSign(
+  peerId: PeerId,
+  ordinal: SnapshotOrdinal,
+  delta: Long,
```
marcinwadon, line 14: is delta a NonNegLong?

modules/node-shared/src/main/scala/org/tessellation/node/shared/cli/CliMethod.scala
cngo-github: Added names to the arguments.

modules/node-shared/src/main/scala/org/tessellation/node/shared/config/types.scala
```diff
@@ -126,4 +127,11 @@ object types {
...
+    runInterval: FiniteDuration
```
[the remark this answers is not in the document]
cngo-github: `runInterval` is for the daemon only. It has been removed from here.

modules/node-shared/src/main/scala/org/tessellation/node/shared/config/types.scala
cngo-github: It is the maximum amount of "time" that two snapshots with the same ordinal, but with different hashes, can be spaced apart and still be considered a Double Signing event. It has been renamed to `maxDetectionDelta`.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/security/DoubleSignDetect.scala
cngo-github: Moved `DoubleSignDetect` to the `org.tessellation.node.shared.domain.snapshot` package.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/snapshot/DetectedDoubleSign.scala
cngo-github: Changed to a `NonNegLong`.

modules/node-shared/src/main/scala/org/tessellation/node/shared/cli/CliMethod.scala
```diff
@@ -61,6 +61,10 @@ object CliMethod {
 
   val forkInfoStorageConfig: ForkInfoStorageConfig = ForkInfoStorageConfig(10)
 
+  val doubleSignDetectConfig: DoubleSignDetectConfig = DoubleSignDetectConfig(
+    maxDetectionDelta = 10L
```
cngo-github, line 65: @ryle-goehausen , please verify that this is the desired value for the double signing-rollback cutoff.

ryle-goehausen: Yes that's fine for now, we can adjust after viewing metrics.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/snapshot/DoubleSignDetect.scala
```diff
@@ -0,0 +1,81 @@
...
+}
+
+class DoubleSignDetect[F[_]: Async](
+  storage: ForkInfoStorage[F],
+  config: DoubleSignDetectConfig
+) {
+  private val logger: SelfAwareStructuredLogger[F] = Slf4jLogger.getLogger
+
+  private def isDoubleSign(previous: IndexedForkInfo, current: IndexedForkInfo): Boolean =
+    (current.idx - previous.idx <= config.maxDetectionDelta) && (previous.info.hash =!= current.info.hash)
```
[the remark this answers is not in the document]
kpudlik: Also consider such a case that node is forked and sends you proposal `AAA`  for ordinal `100` and then rejoins and sends another proposal `BBB` for ordinal `100`. You end up with a false-positive and you treat such node as malicious. For such a case the observing state probably helps as node "waits" with participating in consensus after the rejoin.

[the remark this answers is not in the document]
cngo-github: > I am also wondering why `<= config.maxDetectionDelta` is considered as a double sign but `> maxDetectionDelta` is not 🤔 . What is the reasoning behind considering one delta as "safe" and another as "unsafe"?

Regarding this, @kpudlik , the reasoning is that if a node does a rollback, it will also transmit different snapshot hashes for the same ordinals. For example, it sends hash `BB` for ordinal `100` when it previously sent hash `AA` for that ordinal. The reasoning is that rollbacks will cause conflicting hashes to be detected, but they will be beyond `maxDetectionDelta`. So, we care about conflicts when they are within `maxDetectionDelta`.

modules/node-shared/src/main/scala/org/tessellation/node/shared/cli/CliMethod.scala
marcinwadon: @ryle-goehausen I think you should run the cluster and observe the metrics, then update the PR with the correct value. Otherwise, the change with incorrect values may enter mainnet after merging this PR.

marcinwadon:
I'm curious how you tested that change @cngo-github since it looks a major one? Did you run an AWS cluster?

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/snapshot/DoubleSignDetect.scala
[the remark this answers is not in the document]
ryle-goehausen: >So the functionality is as following:
>Node signs snapshot of hash AAA at ordinal 100 and then within >maxDetectionDelta signs snapshot of hash BBB also at ordinal >100 and such behavior is considered as a double-sign. I doubt it is >enough to consider it a double sign because lastSnapshotHash >may differ.

It is enough, because we’re assuming here that somewhere else in the ordinal that lastSnapshotHash will be treated as another hash in the iterable, same logic we’re using elsewhere to determine a single hash for ForkInfo as being different constituting a fork.

> The only case when the double sign is obvious is when:
> node proposed 2 proposals or more
> lastSnapshotHashes are equal
> ordinals are equal
> all the proposals were sent during the same consensus round (but > here check happens every config.runInterval for hardcoded delta > which may consider the proposal sent outside the consensus > round as double-sign)

This has already been addressed earlier, the main problem with considering only the proposals of the current round as opposed to a time series window is we’re currently dealing with a large number of honest forks due to unrelated issue which has not yet been fixed. We can’t aggressively penalize a behavior due to a separate underlying problem, hence the use of a more strict check here over a longer period of time.

> I am also wondering why <= config.maxDetectionDelta is > >
> considered as a double sign but > maxDetectionDelta is not 🤔 . 

Reference the earlier function before where this is invoked, the delta is on the gossip receipt index, NOT the ordinal. We’re grouping by the ordinal. The distinction here we’re trying to detect is whether or not we’re receiving similar ordinals at similar receipt order, which is NOT the case that happens during a rollback.

> What is the reasoning behind considering one delta as "safe" and > another as "unsafe"? It can cause "faster" forks to be treated as > malicious but "slower" ones as correct. Relying on magic number > and some "distance" is risky.

Again, this is the wrong interpretation, because we’re doing a groupBy ordinal and delta on receipt order. NOT a delta on ordinal, hence it’s not detecting a ‘faster’ fork’, but rather a concurrently repeating double fork.

I still need to review more to determine if the rest of this is correctly implemented, but in plain English we’re attempting to find similar ordinals with differing hashes that are received in close gossip order. Ideally we’d be expecting some calculation like this to occur over multiple iterations, but for now I think it’s just set to detect a single event. A better (future) implementation would involve detecting this over a longer period of time.


>Also even if we find the perfect maxDetectionDelta and we check > the lastSnapshotHash as well then it still doesn't solve the > double-sign problem. 

This PR doesn’t attempt to solve every case, just one to start with, see earlier comments regarding how we can later make improvements to it.

> Consider such an example:
> A malicious node creates a proposal of hash BBB at ordinal 100 and creates a proposal of hash AAA at ordinal 100. Then he can either:
>”send both, sign one"

This has been addressed earlier, we can derive the information from the signed artifacts which can be done as a separate PR.

> “sign both, send both"

Detected by forkInfo


> and additionally
> “lie in fork-info about which one has been signed"

We’re using forkInfo to detect forks and reset the node against it (or will with nodectl), so they’ll ignore the signed artifact if we don’t include them in that calculation.

> “send many fork-infos to reach the max-size of storage"

Part of a DDOS related problem that can be addressed as part of efforts to eliminate DDOS / spam, this is not the only place where this is an issue, but actually Chuong this is an appropriate place where we could impose a per-node history limit (need to check if that’s already in here.)


>Let's say that such a node decided that BBB is proposed during a proposal round and then lied that he proposed AAA in a fork-info. >Or even didn't not send the fork-info which is completely legal as you do not check if everyone sent it.

We’re using it as a basis for a check which can potentially reset the node, so it’s used.

> So basically the problem with your solution is that each node sends 2 things independently: Proposal and ForkInfo where in theory both should match but in reality malicious node can send the malformed ForkInfo or not send it at all and nothing will happen because gossiping is a step after the consensus.

I’ve already responded to this in detail in Maciej’s comment, these issues can be treated independently and we can derive the exact same info from the other step.


> I think it is the same as Maciek said:
The fork info declarations are sent by the node in question itself. Hence the malicious node can decide to simply not send any declarations and having that we will not detect anything. 

Again, just repeating my earlier comments here, if we’re using it to reset the node to the correct fork, it’s fundamentally a requirement, even then on a shorter term basis, we’ll change the derivation source of the information so it’s derived from the signatures, which is a separate change.

> Even if the malicious node decided to send declarations it could send them in such a way that would let it manipulate the result on the nodes checking them. That results from the reliance on the order of declarations put into the storage and the storage capacity itself.

Any node switching chains still needs to validate the underlying data, this is not any different from saying you can ‘manipulate’ the active fork by a majority operation so long as it’s valid, that’s always a concern. The storage problem I agree with should potentially be solved in here to limit the amount of info per node.

> That's why I think that instead of having a separate time-based step that relies on magic distance we should just run the detection as a part of the consensus round 

We can’t do that as already explained, considering how many honest short-time forks we see it’d penalize nodes that are absolutely not malicious until that is solved separately.

> (so you get sync with consensus for free). Currently, consensus only stores the very first proposal sent by the node and all next proposals (which may be double-sign) are just ignored. 

I agree that should be fixed, but there’s no reason not to consider a time-window of this data to be of importance. We’ve discussed this earlier, and even if we stored a time window inside the consensus proposal classes, it would essentially be the exact same data model proposed in this, since we don’t care about the underlying data in the proposals, but rather just the hashes. So either way, it would recreate what we’re doing here, it’s just a matter of organization.

A later step (after solving some of the meaningless forking issues,) would in fact be to ALSO include the information about the latest proposal, to distinguish fork events that are immediate versus those over a time window.

> What we could do is just to store all the proposals sent by the node instead of the first one and then after collecting them just check if any of the nodes sent the proposal for the same ordinal and lastSnapshotHash twice or more. So instead of proposal: Option[Proposal], just store proposal: Option[NonEmptyList[Proposal]] and all the nodes who have more than 2 proposals for the same ordinal and lastSnapshotHash are double-signers.

Already addressed earlier, same problem as before where we get a ton of false positives.

> No relying on time -> no race conditions

We’re not relying on explicit time here, but receipt order.
 > No relying on distance -> no favoring slower forks and no need for hyperparameter optimization

In general, any detection function would be best considering a window of events, rather than the most immediate event. It’s useful to distinguish the severity of the malicious behavior, where a node acting over a longer period of events is considered ‘worse’ than an immediate operation, which could simply be due to connectivity issues or networking related problems.
 > Relying on consensus round -> sync with consensus and be enclosed within the context of a single round with access to all the data to detect double-sign

Again, the context of a single round is strictly speaking less information than a window of events over multiple rounds, which allows us to detect and distinguish between different types of situations.


> Also consider such a case that node is forked and sends you proposal AAA for ordinal 100 and then rejoins and sends another proposal BBB for ordinal 100 (the correct one this time because it is not forked anymore). You end up with a false-positive and you treat such node as malicious. For such a case the observing state probably helps as node "waits" with participating in consensus after the rejoin.

 Since we’re planning on resets through restarts, I don’t think this can happen easily. By the time the node resets, it would have NOT done a new proposal round for round 100, but would instead ‘rejoin’ at round 105 or so. It would redownload against the ‘correct’ chain, and then attempt proposals at the ‘later’ round. So actually I don’t think it would even happen, because of the way we reset nodes.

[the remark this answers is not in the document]
kpudlik: Ok, I went through the answers and I think that that PR still doesn't completely cover the double-sign problem but may solve the forking problem. On the other hand, I agree that stricter mechanisms may lead to false penalties due to the forking issues we have. 
>This has already been addressed earlier, the main problem with considering only the proposals of the current round as opposed to a time series window is we’re currently dealing with a large number of honest forks due to unrelated issue which has not yet been fixed

^ this is also the reason why it needs to be implemented outside of single consensus round context.

Syntactically everything is ok, but I am still wondering if it works properly semantically so if you tested it and you are planning to create more PRs to cover more cases then feel free to merge that.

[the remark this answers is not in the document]
cngo-github: I've created a separate ticket to create an attack harness to rigorously test double signing detection. For the purposes of this PR, the feature is only being used for logging. So, we only want to ensure that it doesn't break anything.

cngo-github:
> I'm curious how you tested that change @cngo-github since it looks a major one? Did you run an AWS cluster?

@marcinwadon , I've created a separate ticket to create an attack harness to rigorously test double signing detection. For the purposes of this PR, the feature is only being used for logging. So, we only want to ensure that it doesn't break anything. I believe that the Github Actions tests are sufficient to show that nothing is broken.

marcinwadon:
The change is already outdated and won't be updated soon, so I'm closing the PR.

PR #779, opened 2023-11-17
Change Data Application hook to return optional Resource

sdavidp:


modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
```diff
@@ -45,7 +45,7 @@ abstract class CurrencyL0App(
   val kryoRegistrar: Map[Class[_], KryoRegistrationId[KryoRegistrationIdRange]] =
     sdkKryoRegistrar
 
-  def dataApplication: Option[BaseDataApplicationL0Service[IO]]
+  def dataApplication: IO[Option[BaseDataApplicationL0Service[IO]]] = none.pure[IO]
```
marcinwadon, line 48: The data application should be wrapped into IO and the whole implementation should be optional, so it should be: Option[IO

marcinwadon (changes requested):

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
sdavidp: Why is that the desired interface?

kpudlik: Because providing the data application is optional for metagraph developer. One can just create a metagraph with just regular currency transactions without any custom data. So the `Option` should be the base effect for that def.

sdavidp: changed

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/Main.scala
```diff
@@ -20,8 +19,6 @@ object Main
       version = BuildInfo.version
     ) {
 
-  def dataApplication: Option[BaseDataApplicationL0Service[IO]] = None
```
marcinwadon, line 23: Are you sure about removing that one? It was left to be an example that one can implement a data application and provide it here

marcinwadon: You removed both actually. I know it defaults to None. Ok, nevermind. It should be already covered by docs I believe. Can you confirm that?

marcinwadon (dismissed):

marcinwadon (approved):

PR #782, opened 2023-11-23
Fix DataApplication traverse recursion loop

marcinwadon:


IPadawans (approved):
LGTM

modules/shared/src/main/scala/org/tessellation/currency/dataApplication/DataApplicationTraverse.scala
```diff
@@ -162,11 +162,9 @@ object DataApplicationTraverse {
                     }
               } >>= {
               case (stepCache, Some((state, ordinal))) =>
-                (cache ++ stepCache, (state, ordinal).some).asRight[Acc].pure[F]
+                (cache ++ stepCache, (state, ordinal).some).asRight[Cache].pure[F]
```
TheMMaciek, line 165: I don't think you need to add `stepCache` to `cache` again here, check if I'm right but it seems `stepCache` contains `cache` at this point already.

Also you could have a bit less repackaging with:
```
case cacheAndLastState @ (_, Some(_)) =>
  cacheAndLastState.asRight[Cache].pure[F]
case (stepCache, _) =>
  stepCache.asLeft[Output].pure[F]
```

PR #795, opened 2023-12-05
Fix check for snapshot info directory

marcinwadon:


modules/sdk/src/main/scala/org/tessellation/sdk/cli/snapshot.scala
```diff
@@ -29,14 +29,16 @@ object snapshot {
...
-    case _ =>
-      "Paths for global snapshot and incremental snapshot (both persisted and tmp) must be different.".invalidNel[(Path, Path, Path, Path)]
+  val pathsOpts = (snapshotPath, incrementalPersistedSnapshotPath, incrementalTmpSnapshotPath, snapshotInfoPath)
+  val pathsErr = "Paths for file storages must differ.".invalidNel[(Path, Path, Path, Path)]
+
+  val opts = pathsOpts.tupled.mapValidated {
+    case (snapshotPath, incrementalPersistedSnapshotPath, incrementalTmpSnapshotPath, snapshotInfoPath) =>
+      val paths = List(snapshotPath, incrementalPersistedSnapshotPath, incrementalTmpSnapshotPath, snapshotInfoPath)
+      if (paths.size === paths.toSet.size)
+        (snapshotPath, incrementalPersistedSnapshotPath, incrementalTmpSnapshotPath, snapshotInfoPath).validNel[String]
```
sdavidp, line 39: Does this refactor work? I'm trying to avoid the same tuple repeated 3 times
```suggestion
    case pathsTupled =>
       if (pathsTupled == pathsTupled.distinct)
         pathsTupled.validNel[String]
```

IPadawans (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/cli/snapshot.scala
sdavidp: I forgot the `.toList` part in my suggestion.

marcinwadon: there is no `.toList` on tuple. I've used the shapeless way right now. please re-review
image

kpudlik (approved):

modules/sdk/src/main/scala/org/tessellation/sdk/cli/snapshot.scala
sdavidp: Looks like my suggestion will only work for Scala 3. Disregard...

PR #798, opened 2023-12-06
Initial SDK - SecurityProvider and Keypair generator as examples

marcinwadon:


sdavidp:
@marcinwadon To fix the failing tests, do you need to add `sdk/publishM2` to the correct action in the 2 .yml files?

marcinwadon:
No, @sdavidp. I need to change the template repository https://github.com/Constellation-Labs/github-action-metagraph-test that we use for github action tests. IMO it should be a part of this repository. Otherwise it looks like kinda circular dependency. Anyway, the template is global for this repository, so changing it now would break previous PRs.

kpudlik (approved):

A1811-qbite (approved):

PR #801, opened 2023-12-13
Fix rosetta publish error

sdavidp:
When executing `publishLocal` the sbt task ends due to a scaladoc error in _rosetta_, so not all the modules are published. This PR

1. Fixes the error
2. Reorders the _aggregate_ list to closer match dependencies so the error wouldn't affect publishing of required artifacts

### DIFF
--- a/modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/Amount.scala
+++ b/modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/Amount.scala
@@ -11,15 +11,13 @@ import eu.timepit.refined._
 import eu.timepit.refined.api.Refined
 import eu.timepit.refined.auto._
 import eu.timepit.refined.cats._
-import eu.timepit.refined.numeric.Interval.OpenClosed
-import eu.timepit.refined.numeric.{NonNegative, Positive}
+import eu.timepit.refined.numeric.{Negative, NonNegative, Positive}
 import eu.timepit.refined.predicates.all.Or
 import io.circe.refined._
 import io.estatico.newtype.macros.newtype
 
 object amount {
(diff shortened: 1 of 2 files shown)


TheMMaciek:
@sdavidp you need a label

TheMMaciek (approved):

modules/rosetta/src/main/scala/org/tessellation/rosetta/domain/Amount.scala
```diff
@@ -11,15 +11,13 @@ import eu.timepit.refined._
...
-import eu.timepit.refined.numeric.{NonNegative, Positive}
+import eu.timepit.refined.numeric.{Negative, NonNegative, Positive}
 import eu.timepit.refined.predicates.all.Or
 import io.circe.refined._
 import io.estatico.newtype.macros.newtype
 
 object amount {
-  type NegatedPositive = OpenClosed[W.`Long.MinValue`.T, -1L]
-  type AmountValuePredicate = Positive Or NegatedPositive
+  type AmountValuePredicate = Positive Or Negative
```
marcinwadon, line 20: Does Negative exclude 0?

sdavidp: @marcinwadon No. If you drill into both of those two they're defined as `Greater[0]` and `Less[0]`. Maybe I didn't see it but I couldn't find a "NonZero" refinement, which is what this is.

sdavidp: Update. I mean YES... it excludes zero. I first read "include"

PR #805, opened 2023-12-15
Cut pulled GlobalSnapshotEvents, if necessary

sdavidp:


modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/event/PeerIdEventCutter.scala
```diff
@@ -0,0 +1,9 @@
+package org.tessellation.node.shared.domain.event
+
+import org.tessellation.schema.gossip
+import org.tessellation.schema.peer.PeerId
+
+trait PeerIdEventCutter[F[_], Event] {
+  type PeerIdEvents = Map[PeerId, List[(gossip.Ordinal, Event)]]
```
marcinwadon, line 7: `import org.tessellation.schema.gossip.Ordinal`. should it stay an `Ordinal` or should it be `SnapshotOrdinal`?

sdavidp: @marcinwadon The data structure that I need to work with is returning `gossip.Ordinal` which is not the same as `SnapshotOrdinal`. I can change it import the class directly if you like. I left it like that so that it was obvious that it wasn't the same type of ordinal.

marcinwadon: afaik we always use direct imports.
why you need to rely on the gossip ordering?

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
```diff
@@ -0,0 +1,133 @@
...
+      case class PeerIdHashedEvent(id: PeerId, event: Hashed[Block]) extends PeerIdEvent {
+        val fee: Long = event.signed.value.transactions.toIterable.map(_.value.fee.value.value).sum
+        def hash: Hash = event.hash
+        def toGlobalSnapshotEvent: GlobalSnapshotEvent = event.signed.asRight[StateChannelEvent]
+
+        def isMyChild(that: PeerIdHashedEvent): Boolean =
+          that.event.signed.value.parent.contains_(event.ownReference)
+      }
+
+      def unpackGlobalPeerIdEvents(events: GlobalPeerIdEvents): F[(List[PeerIdStateChannelEvent], List[PeerIdHashedEvent])] =
```
sdavidp, line 70: It might be cleaner to return `Ior[List[PeerIdStateChannelEvent], List[PeerIdHashedEvent]]`

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
```diff
@@ -0,0 +1,133 @@
...
+import org.tessellation.schema.gossip.Ordinal
+import org.tessellation.schema.peer.PeerId
+import org.tessellation.security.Hashed
+import org.tessellation.security.hash._
+
+trait GlobalSnapshotEventCutter[F[_]] extends PeerIdEventCutter[F, GlobalSnapshotEvent]
+
+object GlobalSnapshotEventCutter {
+
+  def make[F[_]: Async: KryoSerializer](maxBinarySizeBytes: Long = 20_000_000L): PeerIdEventCutter[F, GlobalSnapshotEvent] =
```
sdavidp, line 27: Reviewers: I'm thinking of moving the default size to `ConsensusConfig`. WDYT?

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/event/PeerIdEventCutter.scala
marcinwadon: my question was why the gossip is in here. how does it connect with the gossip?

sdavidp: Put another way, I need to act on the result of this call:
```
ConsensusStorage.pullEvents(upperBound: Bound): F[Map[PeerId, List[(Ordinal, Event)]]]
```
and the only type variable is `Event`, which is either a `GlobalSnapshotEvent` or `CurrencySnapshotEvent` -- the others are fixed.

sdavidp: @marcinwadon I could define the trait this way,
```
trait EventCutter[F[_], K, T, Event] {
  def cut(events: Map[K, List[(T, Event)]]): F[Map[K, List[(T, Event)]]]
}
```
and in the _make_ functions where it's passed around I could hard-code _K_ = `PeerId` and _T_ = `Ordinal`.
Would that be better?

marcinwadon: Ok, got it. So, the answer for my question is that the gossip ordinal is used to determine the order of events. That's fine to keep the gossip ordinal hardcoded then.

modules/dag-l0/src/test/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutterSuite.scala
```diff
@@ -0,0 +1,206 @@
...
+      case None => None
+      case Some(globalPeerIdEvent) =>
+        val globalSnapshotEvent = eventToRemove.gse
+        val updatedEvents = globalPeerIdEvent.filterNot { case (_, gse) => gse == globalSnapshotEvent }
+        Option.when(updatedEvents.nonEmpty)(updatedEvents)
+    }
+
+  // Take n events from the two (sorted) lists:
+  // Take lowest head from the lists
+  // If heads are equal, take from the longer list, or first list if lengths are equal
```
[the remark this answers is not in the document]
sdavidp: @marcinwadon Done

modules/dag-l0/src/test/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutterSuite.scala
```diff
@@ -0,0 +1,206 @@
...
+        parent <- parentBlockGen
+        childBlocks <- childrenBlockGen
+      } yield (peerId, parent, childBlocks)
+
+    forall(gen) {
+      case (peerId, parent, childBlocks) =>
+        for {
+          parentRefs <- parent.toHashed[IO].map(_.ownReference).map(NonEmptyList.of(_))
+
+          // safe to remap since generated Signed[Block] not dependent on parent refs
```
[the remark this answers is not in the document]
sdavidp: Done

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
```diff
@@ -0,0 +1,146 @@
...
+              if (h1.fee < h2.fee || (h1.fee == h2.fee && sortedStateEvents.length >= hashedEvents.length))
+                (h1, new OrderedEvents(t1, hashedEvents)).some
+              else
+                (h2, new OrderedEvents(sortedStateEvents, t2 ++ parents)).some
+          }
+        }
+      }
+
+      object OrderedEvents {
+        // Order for removal: NOT parent → Fee → Split by event type → Hash
```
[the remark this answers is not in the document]
sdavidp: Done

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/event/PeerIdEventCutter.scala
sdavidp: Done

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
```diff
@@ -0,0 +1,144 @@
...
+              es.foldLeft(z) {
+                case ((accLeft, accRight), (peerId, (stateChannelEvents, hashedEvents))) =>
+                  val left = stateChannelEvents.map(PeerIdStateChannelEvent(peerId, _))
+                  val right = hashedEvents.map(PeerIdHashedEvent(peerId, _))
+                  (left ++ accLeft, right ++ accRight)
+              }
+            }
+
+        def unpackEither[A, B](es: List[Either[A, B]]): (List[A], List[B]) =
+          es.foldLeft((List.empty[A], List.empty[B])) {
```
TheMMaciek, line 137: I think you can use `.partitionMap(identity)` here

TheMMaciek (commented):
Some initial comments, more coming in

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
```diff
@@ -0,0 +1,144 @@
...
+        def fee: Long
+        def hash: Hash
+        def toGlobalSnapshotEvent: GlobalSnapshotEvent
+      }
+
+      val ordering: Ordering[PeerIdEvent] = Ordering.by[PeerIdEvent, (Long, Hash)](e => (e.fee, e.hash))
+
+      case class PeerIdStateChannelEvent(id: PeerId, event: StateChannelEvent) extends PeerIdEvent {
+        val fee: Long = event.snapshotBinary.value.fee.value.value
+        def hash: Hash = event.snapshotBinary.lastSnapshotHash
```
TheMMaciek, line 73: Why are you identifying `Block` by it's `hash` and `SnapshotBinary` by it's `lastSnapshotHash`

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
```diff
@@ -0,0 +1,144 @@
...
+        def toGlobalSnapshotEvent: GlobalSnapshotEvent = event.asLeft[DAGEvent]
+      }
+
+      case class PeerIdHashedEvent(id: PeerId, event: Hashed[Block]) extends PeerIdEvent {
```
TheMMaciek, line 77: It's a `...HashedBlock` event or `...HashedDagBlock` event. I don't think `HashedEvent` is clear and also it's inconsistent with the naming of `PeerIdStateChannelEvent`

TheMMaciek: You could even drop the `Hashed` I guess as the types require `Hashed` either way

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
```diff
@@ -0,0 +1,144 @@
...
+              new OrderedEvents(stateChannelEvents.sorted(ordering), hashedEvents)
+          }
+
+        def unpackGlobalPeerIdEvents(events: GlobalPeerIdEventMap): F[(List[PeerIdStateChannelEvent], List[PeerIdHashedEvent])] =
+          events.view
+            .mapValues(es => unpackEither(es.map(_._2)))
+            .toList
+            .traverse {
+              case (peerId, (stateChannelEvents, dagEvents)) =>
+                dagEvents.traverse(_.toHashed[F]).map(hashedEvents => peerId -> (stateChannelEvents, hashedEvents))
```
TheMMaciek, line 124: Could you perform the mapping to `PeerIdStateChannelEvent` and `PeerIdHashedEvent` here? And after it just flatten it or add up using a predefined `Monoid`? I think that could work 🤔  There would also be no need to pass the `peerId` and the `foldLeft`  part below aggregating the events won't be necessary. It looks a bit boilerplate'y right now.

TheMMaciek: Something like:
```
            .traverse {
              case (peerId, (stateChannelEvents, dagEvents)) =>
                dagEvents.traverse(_.toHashed[F]).map { hashedEvents =>
                  val scEvents = stateChannelEvents.map(PeerIdStateChannelEvent(peerId, _))
                  val dagEvents = hashedEvents.map(PeerIdHashedEvent(peerId, _))

                  (scEvents, dagEvents)
                }
            }
```
and then the part adding the matching lists together (didn't try to come up with this part yet)

TheMMaciek: I guess this:
```
.map(_.combineAll)
```

TheMMaciek: Check if this implementation is the right one:
```
        def unpackGlobalPeerIdEvents(events: GlobalPeerIdEventMap): F[(List[PeerIdStateChannelEvent], List[PeerIdHashedEvent])] =
          events.view
            .mapValues(_.map(_._2).partitionMap(identity()))
            .toList
            .traverse {
              case (peerId, (stateChannelEvents, dagEvents)) =>
                dagEvents.traverse(_.toHashed[F]).map { hashedEvents =>
                  val scEvents = stateChannelEvents.map(PeerIdStateChannelEvent(peerId, _))
                  val dagEvents = hashedEvents.map(PeerIdHashedEvent(peerId, _))

                  (scEvents, dagEvents)
                }
            }
            .map(_.combineAll)
```

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
```diff
@@ -0,0 +1,144 @@
...
+          }
+        }
+      }
+
+      object OrderedEvents {
+        // NOTE: Order for removal: NOT parent → Fee → Split by event type → Hash
+        def apply(events: GlobalPeerIdEventMap): F[OrderedEvents] =
+          unpackGlobalPeerIdEvents(events).map {
+            case (stateChannelEvents, hashedEvents) =>
+              new OrderedEvents(stateChannelEvents.sorted(ordering), hashedEvents)
```
TheMMaciek, line 115: I don't think `hashedEvents` are ordered at this point (also there is no total ordering on block I think)

kpudlik (changes requested):

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
sdavidp: @TheMMaciek thank you, I was looking for a better way to do that.

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
```diff
@@ -0,0 +1,144 @@
...
+      case class PeerIdHashedEvent(id: PeerId, event: Hashed[Block]) extends PeerIdEvent {
+        val fee: Long = event.signed.value.transactions.toIterable.map(_.value.fee.value.value).sum
+        def hash: Hash = event.hash
+        def toGlobalSnapshotEvent: GlobalSnapshotEvent = event.signed.asRight[StateChannelEvent]
+
+        def isMyChild(that: PeerIdHashedEvent): Boolean =
+          that.event.signed.value.parent.contains_(event.ownReference)
+      }
+
+      class OrderedEvents(sortedStateEvents: List[PeerIdStateChannelEvent], hashedEvents: List[PeerIdHashedEvent]) {
```
TheMMaciek, line 86: Why not make it a regular `case class`? I think it is still possible to create the regular class without the processing in the apply function right? So it's not guarding the manual creation of such a not pre-processed object I think.

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+      trait PeerIdEvent {
+        def id: PeerId
+        def fee: Long
+        def hash: Hash
+        def toGlobalSnapshotEvent: GlobalSnapshotEvent
+      }
+
+      val ordering: Ordering[PeerIdEvent] = Ordering.by[PeerIdEvent, (Long, Hash)](e => (e.fee, e.hash))
```
TheMMaciek, line 69: Ordering by `fee` and `hash` only is faulty. You can have a newer event with higher fee be attempted before the older event that actually needs to be processed first. Hash can only be used to provide deterministicity for events with the same ordering in regards to all other parameters. Blocks don't have a total order per se. Metagraph snapshots do have and ordering in regards to the chain of snapshots, but you need to discover that chain to know the order. Hashes themselves don't preserve the original order at all, they are only a representation of data underneath.

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
```diff
@@ -0,0 +1,144 @@
...
+
+import eu.timepit.refined.types.numeric.PosInt
+
+object GlobalSnapshotEventCutter {
+  type GlobalPeerIdEventMap = Map[PeerId, List[(Ordinal, GlobalSnapshotEvent)]]
+
+  def make[F[_]: Async: KryoSerializer: Hasher](maxBinarySizeBytes: PosInt): EventCutter[F, PeerId, Ordinal, GlobalSnapshotEvent] =
+    new EventCutter[F, PeerId, Ordinal, GlobalSnapshotEvent] {
+
+      def cut(events: GlobalPeerIdEventMap): F[GlobalPeerIdEventMap] =
```
TheMMaciek, line 32: I think this function discards the cut events and unless it's intentionally introduced in this PR we should change it as we have a logic in other places returning such events back to the pool/cache.

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
sdavidp: @ryle-goehausen will need to answer this; my understanding is that ordering by hash is to make cutting deterministic in the event of a tie with the other rules

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
sdavidp: @ryle-goehausen Can you answer this? Do we need to retain the cut events in a separate collection?

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
sdavidp: Not making a case class isn't for guarding; what do we gain by making it a case class?

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
sdavidp: @TheMMaciek I thought of that and checked, and they are still ordered. However, it did bother me that this may not be something we can rely on. I'll look into it again.

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
sdavidp: @TheMMaciek That's much cleaner, thank you.

modules/node-shared/src/main/scala/org/tessellation/node/shared/cli/snapshot.scala
```diff
@@ -44,7 +44,8 @@ object snapshot {
           timeTriggerInterval = 43.seconds,
           declarationTimeout = 50.seconds,
           declarationRangeLimit = 3L,
-          lockDuration = 10.seconds
+          lockDuration = 10.seconds,
+          EventCutterConfig(maxBinarySizeBytes = 20_000_000)
```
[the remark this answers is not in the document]
sdavidp: @TheMMaciek My understanding is the value is arbitrary. @ryle-goehausen ?

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -216,7 +219,7 @@ object ConsensusStateUpdater {
                     Applicative[F].whenA(majorityTrigger === TimeTrigger)(consensusStorage.clearTimeTrigger) >>
                       state.facilitators.hash.flatMap { facilitatorsHash =>
                         for {
-                          peerEvents <- consensusStorage.pullEvents(bound)
+                          peerEvents <- consensusStorage.pullEvents(bound) >>= eventCutter.cut
```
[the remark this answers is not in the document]
sdavidp: @kpudlik @TheMMaciek @ryle-goehausen 
We can do this instead,
```
peerEvents <- consensusStorage.pullEvents(bound)
reducedPeerEvents <- eventCutter.cut(peerEvents)
```
then I'll need some guidance on where each of the results should be plugged in.

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
```diff
@@ -0,0 +1,144 @@
...
+            events.pure[F]
+          )
+
+      def needsCutting(events: GlobalPeerIdEventMap): F[Boolean] =
+        events.toBinary.liftTo[F].map(_.length > maxBinarySizeBytes.value)
+
+      def removeEvents(startingEvents: GlobalPeerIdEventMap, iterator: Iterator[PeerIdEvent]): F[GlobalPeerIdEventMap] =
+        startingEvents.tailRecM { events =>
+          for {
+            eventToRemove <- MonadThrow[F].fromOption(iterator.nextOption(), new IllegalStateException("No GlobalSnapshotEvent events"))
```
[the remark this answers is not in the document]
sdavidp: @TheMMaciek Correct, that should never happen; if it does happen then the cutting size is much too small (configuration error) or something is wrong with the algorithm (programmer error)

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
sdavidp: @TheMMaciek Now that I think about it, the hash is only used to make sorting deterministic, so it shouldn't matter.

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
TheMMaciek: Yes ordering by hash as a deterministic measure is fine, but the only other parameter we sort by here is the `fee`. These events aren't ordered after pull especially that there is no total order for blocks.

TheMMaciek (changes requested):
Generally implementing the cutter based on the ordering by fee and hash will not work well. These events on the input to the cut function are not ordered themselves so there is no way to correctly pick the ones that will pass because ordering by fee and hash doesn't abide to the discrete ordering of the events. E.i. even though one event may have bigger fee than the other it doesn't mean we can accept it before the other one and I think that's what this implementation would do - it would be inefficient as a lot or even most of the events selected here may not make it to the snapshot because they won't pass validation. First we need to determine what is this discrete order of these events: parallel chains of metagraph snapshots and a tree of dag blocks. We can't calculate the order of dag blocks based on their references because the parent <-> child relationship is not the only thing influencing the discrete order of blocks - there may be two blocks unrelated in terms of parents (seeming as parallel blocks) that actually need to be accepted one after the other because of the balance changes that need to happen.

So after determining the parallel chains of metagraph snapshots (single chain for every metagraph identifier) and a tree of dag blocks we could then look at the fees and only let some subset of all these events to be put in a snapshot respecting the discrete order of the events we are processing.

I'm using the "discrete order" term to mean an order that exists but is not definable as a function comparing fields of objects (e.g. fees, ordinals, heights, parents).

I think there is an implementation for currencySnapshot that's closer to what we need here. Not sure if it creates these chains and trees of events and cuts "from the end" but it for sure is based on the create artifact functionality (which abides to this discrete order of events processed by CL0) and then removes events if the size of the created artifact is too big.

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
TheMMaciek: I guess they are easier to handle and they tend to be used as DTOs which is what is this classes role here I think.

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
```diff
@@ -0,0 +1,134 @@
...
+  type GlobalPeerIdEventMap = Map[PeerId, List[(Ordinal, GlobalSnapshotEvent)]]
+
+  def make[F[_]: Async: KryoSerializer: Hasher](maxBinarySizeBytes: PosInt): EventCutter[F, PeerId, Ordinal, GlobalSnapshotEvent] =
+    new EventCutter[F, PeerId, Ordinal, GlobalSnapshotEvent] {
+
+      def cut(events: GlobalPeerIdEventMap): F[GlobalPeerIdEventMap] =
+        needsCutting(events)
+          .ifM(
+            OrderedEvents(events)
+              .map(Iterator.unfold(_)(_.next))
```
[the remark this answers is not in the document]
sdavidp: I don't understand; that's what the iterator is doing, cutting from a pre-calculated order.

Coming up with a definitive order for removal with a one-time sort might be difficult/confusing due to the no. of events rule. Maybe the re-sorting can be eliminated if we rework the algorithm to ignore parents rather than remove them. (The first pass at this was checking the lengths of the resulting list, not the original with the parents.) However, I wonder if any time-savings is worth the added complication used in removal. 🤔

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/consensus/ConsensusStateUpdater.scala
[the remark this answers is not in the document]
marcinwadon: @sdavidp can you move the event cutter to run inside `createProposalArtifact`? 🤔 so basically "within" pull events

modules/node-shared/src/main/scala/org/tessellation/node/shared/cli/snapshot.scala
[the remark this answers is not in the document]
ryle-goehausen: It's arbitrary at this point, shouldn't matter exact value so long as it's referenced from elsewhere. I think it might make sense to pull this out as a constant though, so it could be imported from SDK or otherwise used elsewhere, rather than as a config.

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
ryle-goehausen: As above, it avoids a recalculation of the hash, we could recalculate it here if there's another reason to? I'm not sure it'll matter apart from sorting.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/consensus/ConsensusStateUpdater.scala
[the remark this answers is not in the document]
ryle-goehausen: In general, we're expecting dropped events in other places (particularly regarding rejections and re-submissions elsewhere,) so it's okay to maintain that for now. If it can be easily fixed by moving it into the pull function, then that makes sense to change and implement it here, as there's no reason (yet) to drop them arbitrarily (although there will be in the future due to memory constraints on the pool.)

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
ryle-goehausen: > You can have a newer event with higher fee be attempted before the older event that actually needs to be processed first. 

This is as desired, especially for dynamic fee calculations, as it follows an RBF (replace by fee) policy, which we'd want to achieve for metagraph snapshot submissions. Since we're dealing with 'unordered' data in the sense that it hasn't yet been accepted, there is no 'newer' event from the perspective of the chain data.

> Hash can only be used to provide deterministicity for events with the same ordering in regards to all other parameters.

Agreed, it should always be the last sort parameter.

>Blocks don't have a total order per se. Metagraph snapshots do have and ordering in regards to the chain of snapshots, but you need to discover that chain to know the order. 

For this, we need to rely on validation to reject invalid ordering. So it should be considered invalid by ordinal definitions, NOT by mempool event receipt order. 

Then if we had the chain of events of considering a mg snapshot of ordinal 100 with fee 1 and another with ordinal 100 and fee 2, even if the second event was 'produced later' (in test time,) it should replace the first event, if all other validations are equal.

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
```diff
@@ -0,0 +1,134 @@
...
+
+          (sortedStateEvents, nonParents.sorted(ordering)) match {
+            case (Nil, Nil) =>
+              none
+            case (h :: t, Nil) =>
+              (h, new OrderedEvents(t, Nil)).some
+            case (Nil, h :: t) =>
+              (h, new OrderedEvents(Nil, t ++ parents)).some
+            case (h1 :: t1, h2 :: t2) =>
+              if (h1.fee < h2.fee || (h1.fee == h2.fee && sortedStateEvents.length >= hashedEvents.length))
```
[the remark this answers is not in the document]
ryle-goehausen: > StateChannelEvent removed before DAGEvent events in case of a tie?

This is fine, it's a rare situation for tie and only goal is to have rough equality.

ryle-goehausen:
>These events on the input to the cut function are not ordered themselves so there is no way to correctly pick the ones that will pass because ordering by fee and hash doesn't abide to the discrete ordering of the events

Since we're decoding the currency snapshot and hence have an ordinal, there is a validation ordering that should be imposed. We should rely on the chain data ordering and validation to determine this, and otherwise allow RBF policies.

modules/node-shared/src/main/scala/org/tessellation/node/shared/cli/snapshot.scala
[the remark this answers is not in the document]
marcinwadon: @ryle-goehausen What do you mean by "imported from SDK"? It's indeed pulled out at the edge of the application, so it can be provided per use case now.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/consensus/ConsensusStateUpdater.scala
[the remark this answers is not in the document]
marcinwadon: > we're expecting dropped events in other places 

@ryle-goehausen can you elaborate on that part? afaik, we handle all the places programatically to return the discarded events back to the pool (unless they are clearly invalid)

modules/node-shared/src/main/scala/org/tessellation/node/shared/cli/snapshot.scala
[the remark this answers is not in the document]
TheMMaciek: Yeah, we can make it a constant so that it's not configurable actually (every node needs the same value here). I suggested the value `20971520` because in the implementation we are comparing count of bytes in the array, so effectively this number would be equivalent to 20MB of data.

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
TheMMaciek: If it's only a tie breaker to get deterministic result on every node then I think we can stick with using `hash` for blocks and `lastSnapshotHash` for snapshots. We will for sure have less computing because we won't calculate the hash of the snapshot.

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
TheMMaciek: Yeah replacing one event by another concurring with higher fee is ok and I think we should have such a mechanism. The problem is that it's not the case I'm describing here. Ordering events by fee and hash only inflicts that we can treat a snapshot with ordinal 100 and fee 10 as the first to accept over snapshot 99 with fee 5 and it won't work. For the concurring events so two or more snapshots at ordinal 100 for the same metagraph we should take the one with higher fee I agree with that.

Essentially we need to first find the order in which all the events we want to process need to be processed (which are parallel and which rely on each other) and then we can pick the ones with higher fee first (but we need to respect that primary order of events).

TheMMaciek:
> Since we're decoding the currency snapshot and hence have an ordinal, there is a validation ordering that should be imposed. We should rely on the chain data ordering and validation to determine this, and otherwise allow RBF policies.

FYI not all metagraph snapshots need to be the `CurrencySnapshot` so we can't always rely that we will be able to order by ordinal. `StateChannelSnapshotBinary` has an order based on the `lastSnapshotHash` so we can figure out the order of these events - we have functions doing that as a part of artifact creation. We also accept blocks with a recursive function that effectively builds an order in which blocks can be accepted.

modules/node-shared/src/main/scala/org/tessellation/node/shared/cli/snapshot.scala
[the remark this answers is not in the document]
ryle-goehausen: > @ryle-goehausen What do you mean by "imported from SDK"? It's indeed pulled out at the edge of the application, so it can be provided per use case now.

Available for use in the CL-0 code for submission validation to ensure an invalid snapshot is not submitted, that's all I mean.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/consensus/ConsensusStateUpdater.scala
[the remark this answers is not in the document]
ryle-goehausen: Can discuss in meeting, it was referenced in prior ticket for metagraph related rejections I believe? There was a prior workaround done by marcus for this.

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
ryle-goehausen: >Essentially we need to first find the order in which all the events we want to process need to be processed (which are parallel and which rely on each other) and then we can pick the ones with higher fee first (but we need to respect that primary order of events).

I think the way we impose this order is through the rejection of events considered to be invalid. I.e. the future submission of events, we can discuss how to deal with this in more detail, because I realize it's the same case as a Batch DAGTransaction submission (i.e. chaining ordinals together,) which makes it more complicated. Let's talk about this in the meeting, because as far as I can tell, it's the same type of problem for which we need to rely on either some batch mechanism, or purely validation / sync operation.

ryle-goehausen:
>FYI not all metagraph snapshots need to be the CurrencySnapshot so we can't always rely that we will be able to order by ordinal. StateChannelSnapshotBinary has an order based on the lastSnapshotHash so we can figure out the order of these events - we have functions doing that as a part of artifact creation. We also accept blocks with a recursive function that effectively builds an order in which blocks can be accepted.

I think so long as we're imposing a validator on lastSnapshotHash and enforcing a synchronous acceptance, it should be okay here (rejecting the batch event case for now, or we can discuss in meeting.) And if we decode a currency snapshot, we can enforce an ordinal validation logic, that way any future events will be rejected as invalid until the active GS changes. But again let's discuss sync-validation versus batching strategies in meeting.

modules/node-shared/src/main/scala/org/tessellation/node/shared/cli/snapshot.scala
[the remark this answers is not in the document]
sdavidp: I changed the value to `20971520` -- it will show up when I push changes with state channel event parents test

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
```diff
@@ -0,0 +1,144 @@
...
+        needsCutting(events)
+          .ifM(
+            OrderedEvents(events)
+              .map(Iterator.unfold(_)(_.next))
+              .flatMap(removeEvents(events, _)),
+            events.pure[F]
+          )
+
+      def needsCutting(events: GlobalPeerIdEventMap): F[Boolean] =
+        events.toBinary.liftTo[F].map(_.length > maxBinarySizeBytes.value)
```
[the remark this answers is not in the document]
sdavidp: @TheMMaciek I changed it as suggested.

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
sdavidp: It's now a `case class`

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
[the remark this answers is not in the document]
sdavidp: @TheMMaciek I'll add another comment since I haven't changed the selection function:

We might be able to come up with a one-time sort and then pick events from that. However, I think the sorting algorithm maybe complicated because of the criteria of not removing parents and the no. of each event (i.e. length of the `Left` and `Right` lists.)

My understanding is this cutting shouldn't happen that often so my thought was the keep the code easier to follow.

WDYT?

ryle-goehausen (approved):
I think all comments have been addressed here, and a minimal batch child ordering strategy was added.

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
```diff
@@ -0,0 +1,118 @@
...
+        val fee: Long = event.snapshotBinary.value.fee.value.value
+
+        def toGlobalSnapshotEvent: GlobalSnapshotEvent = event.asLeft[DAGEvent]
+
+        def isMyChild(that: StateChannelEventWithFee): Boolean =
+          that.event.snapshotBinary.lastSnapshotHash === hash
+      }
+      object StateChannelEventWithFee {
+        def apply(event: StateChannelEvent): F[StateChannelEventWithFee] =
+          event.snapshotBinary.content.hash.map(StateChannelEventWithFee(event, _))
```
TheMMaciek, line 69: Look at `StateChannelSnapshotService` and `consume` method to see how the hash is calculated for snapshot binary. Right now the `isMyChild` wouldn't ever match any two snapshots.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/event/EventCutter.scala
```diff
@@ -0,0 +1,5 @@
+package org.tessellation.node.shared.domain.event
+
+trait EventCutter[F[_], A, B] {
+  def cut(scEvents: List[A], dagEvents: List[B]): F[(List[A], List[B])]
```
TheMMaciek, line 4: if these are arbitrary types then maby `aEvents` and `bEvents`?

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
```diff
@@ -0,0 +1,113 @@
...
+      }
+      object StateChannelEventWithFee {
+        def apply(event: StateChannelEvent): F[StateChannelEventWithFee] =
+          event.snapshotBinary.toHashed.map(hashed => StateChannelEventWithFee(event, hashed.hash))
+      }
+
+      case class DAGBlockEventWithFee(event: Hashed[Block]) extends EventWithFee {
+        val fee: Long = event.signed.value.transactions.toIterable.map(_.value.fee.value.value).sum
+
+        val hash: Hash = event.hash
```
TheMMaciek, line 98: Did we settle that it doesn't need to be compared by `proofsHash` (as it is compared by in other places) because this hash is only for deterministicity?

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
```diff
@@ -0,0 +1,113 @@
...
+
+            val scEventsSorted = scEvents.sortWith((left, right) => Ordering[EventWithFee].lt(scEventsMap(left), scEventsMap(right)))
+            val dagEventsSorted = dagEvents.sortWith((left, right) => Ordering[EventWithFee].lt(dagEventsMap(left), dagEventsMap(right)))
+
+            (scEventsSorted, dagEventsSorted).tailRecM {
+              case (Nil, Nil) => MonadThrow[F].raiseError(new IllegalStateException("No events to cut"))
+
+              case (lefts, rights) =>
+                val result = removeOne(lefts, rights)
+                doesNotExceedMaxSize(result._1, result._2).map(flag => Either.cond(flag, result, result))
```
TheMMaciek, line 64: Why is the outcome of `doesNotExceedMaxSize` called `flag`? Maybe just write it as `.map(Either.cond(_, result, result)` and don't introduce any naming.

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
```diff
@@ -0,0 +1,113 @@
...
+                    case n if (n == -1) || (n == 0 && t1.length >= t2.length) => (t1, dagEvents)
+                    case _                                                    => (scEvents, t2)
+                  }
+              }
+
+            val scEventsSorted = scEvents.sortWith((left, right) => Ordering[EventWithFee].lt(scEventsMap(left), scEventsMap(right)))
+            val dagEventsSorted = dagEvents.sortWith((left, right) => Ordering[EventWithFee].lt(dagEventsMap(left), dagEventsMap(right)))
+
+            (scEventsSorted, dagEventsSorted).tailRecM {
+              case (Nil, Nil) => MonadThrow[F].raiseError(new IllegalStateException("No events to cut"))
```
TheMMaciek, line 60: Maybe it should just return empty lists? You could also log it, but maybe throwing is not neccessary? Although I get it that you don't expect it to ever fail with a correct config, but take into account that someone can use really low numbers and no events would pass the limit - in this situation empty lists are correct results right?

TheMMaciek:
Given that we only want a rough ordering, the PR seems mostly fine. I guess the orderings make sense - correct me if I'm wrong but least significant events are at the head of the list right? Only few stylistic comments added.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/event/EventCutter.scala
sdavidp: done

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
sdavidp: done

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
sdavidp: @TheMMaciek I believe so. @ryle-goehausen, do you want to comment?

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
sdavidp: Is it really valid to remove events to the point they're all gone? I thought the objective was to reduce the no. of events when they're unreasonably large, so if the max size is small then the configuration is wrong.

If you think that's better I can have it return empty list, but if we do I think we should have a larger minimum for the size than 0.

TheMMaciek: I think we can have this function to be a more generic thing. If someone really sets the limit too low for their use case than that's their responsibility. Anyway I wouldn't throw if the result is empty lists - the function did what it was told to do, so the error is not it's fault. If we need some validation on the limit then I would impose it with a custom refined type used in the config.

TheMMaciek: And I would still leave the cutter parameter `maxBinarySizeBytes ` to be a `PosInt`.

sdavidp: @TheMMaciek Changed to not throw.

TheMMaciek (approved):
I think it looks ok given the assumptions for this functionality. One last thing I could suggest is to merge the `cut` and `removeEvents` functions into one `cut` function. You could move the size check to happen first in the recursion and only if the size exceeds the limit cut one element and start another loop. Also to avoid unnecessary sorting you would need to make `scEventsSorted` and `dagEventsSorted` to be lazy vals. This could make this functionality a bit simpler, but it's only a suggestion.

sdavidp:
> I think it looks ok given the assumptions for this functionality. One last thing I could suggest is to merge the `cut` and `removeEvents` functions into one `cut` function. You could move the size check to happen first in the recursion and only if the size exceeds the limit cut one element and start another loop. Also to avoid unnecessary sorting you would need to make `scEventsSorted` and `dagEventsSorted` to be lazy vals. This could make this functionality a bit simpler, but it's only a suggestion.
> 
> EDIT: I think I'm wrong about the sorted events being lazy vals, they will either way be evaluated immediately even if the size doesn't exceed the limit - so it won't give us anything and the sorting will always happen which is not what we want. Also we will always build the `EventWithFee` collections even if it's not necessary in the end. So I guess the current implementation is probably optimal.

My intention was to structure the calls to mimic a _do-while_ loop so that we don't build the intermediate data structures or sort unless we know there's going to be some cutting.

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
```diff
@@ -0,0 +1,116 @@
...
+  def make[F[_]: Async: KryoSerializer: Hasher](maxBinarySizeBytes: PosInt): EventCutter[F, StateChannelEvent, DAGEvent] =
+    new EventCutter[F, StateChannelEvent, DAGEvent] {
+
+      def cut(scEvents: List[StateChannelEvent], dagEvents: List[DAGEvent]): F[(List[StateChannelEvent], List[DAGEvent])] =
+        doesNotExceedMaxSize(scEvents, dagEvents)
+          .ifM((scEvents, dagEvents).pure[F], removeEvents(scEvents, dagEvents))
+
+      def doesNotExceedMaxSize(scEvents: List[StateChannelEvent], dagEvents: List[DAGEvent]): F[Boolean] =
+        (scEvents.toBinary.liftTo[F], dagEvents.toBinary.liftTo[F])
+          .mapN((a, b) => a.length + b.length <= maxBinarySizeBytes.value)
```
kpudlik, line 35: I see that you only compare the size of `List[StateChannelEvent] + List[DAGEvent]` instead of comparing the size of the whole produced `GlobalSnapshot` proposal. Is it by design?

I am asking because `CurrencyEventsCutter` compares the whole `CurrencySnapshot` proposal by recursively recreating it until the whole proposal fits the limit (not just the blocks part of it).

kpudlik (changes requested):

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/GlobalSnapshotEventCutter.scala
sdavidp: Where this is plugged in the `GlobalSnapshot` has already been separated into two parts, and some filtering of invalid values removed so I had to cutter work on those two parts. The max size is effectively arbitrary; does it matter if we're checking the size of the resulting parts rather than the original data structure?

cc: @TheMMaciek @ryle-goehausen

TheMMaciek: I think it doesn't matter as long as we don't want the whole snapshot to not exceed some limit (and I guess it's not the case). Also if we impose validation we will need to only validate the size of the events and not a size of a whole snapshot.

PR #807, opened 2023-12-19
Tracking pending state channel snapshot binaries with retry

kpudlik:


modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/SentStateChannelBinaryTrackingService.scala
```diff
@@ -0,0 +1,76 @@
...
+
+trait SentStateChannelBinaryTrackingService[F[_]] {
+  def setPending(binary: Signed[StateChannelSnapshotBinary]): F[Unit]
+  def getPending: F[ListMap[Signed[StateChannelSnapshotBinary], NonNegInt]]
+  def getRetriable: F[Option[Signed[StateChannelSnapshotBinary]]]
+  def updateByGlobalSnapshot(globalSnapshot: GlobalIncrementalSnapshot): F[Unit]
+}
+
+object SentStateChannelBinaryTrackingService {
+  private val maxChecks: PosInt = 3
```
kpudlik, line 36: I am wondering if we should move it to config and what should be the correct number. It basically says for how many ordinals it waits until it retries.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/modules/Services.scala
```diff
@@ -89,6 +92,8 @@ object Services {
         maybeDataApplication
       )
 
+      sentStateChannelBinaryTrackingService <- SentStateChannelBinaryTrackingService.make[F](storages.identifier)
```
TheMMaciek, line 95: It seems that it is created twice in the same flow, why? 🤔

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/StateChannel.scala
```diff
@@ -31,7 +32,9 @@ object StateChannel {
         .awakeEvery[F](awakePeriod)
         .evalMap(_ => services.globalL0.pullGlobalSnapshots)
         .evalMap {
-          case Left((snapshot, state)) => storages.lastGlobalSnapshot.setInitial(snapshot, state)
+          case Left((snapshot, state)) =>
+            storages.lastGlobalSnapshot.setInitial(snapshot, state) >> services.sentStateChannelBinaryTrackingService
+              .updateByGlobalSnapshot(snapshot)
```
TheMMaciek, line 37: Not sure if you need to add it for this case as it only happens the first time a global snapshot is received.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/SentStateChannelBinaryTrackingService.scala
TheMMaciek: Shouldn't it be called something like 'retryOrdinalDelay' or similar?

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/modules/Services.scala
kpudlik: Fair point, leftover with variable shadowing

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/SentStateChannelBinaryTrackingService.scala
```diff
@@ -0,0 +1,64 @@
...
+      def setPending(binary: Signed[StateChannelSnapshotBinary]): F[Unit] =
+        pendingR.update(_.updated(binary, NonNegInt.MinValue))
+
+      def getPending: F[ListMap[Signed[StateChannelSnapshotBinary], NonNegInt]] = pendingR.get
+
+      def getRetriable: F[Option[Signed[StateChannelSnapshotBinary]]] = getPending.map { pending =>
+        pending.find { case (_, checks) => checks >= retryOrdinalDelay }.map(_._1)
+      }
+
+      def updateByGlobalSnapshot(globalSnapshot: GlobalIncrementalSnapshot): F[Unit] =
```
TheMMaciek, line 58: I would account for one race condition that I think may happen, which is: a node finishing a consensus with some delay in regards to other nodes, then setting the pending binary and at the same time processing GL0 snapshot containing the binary (when the pending binary is not yet set). Other non-delayed nodes may propose the binary to GL0 which may add it to the GL0 snapshot. And this global snapshot may be processed before the binary is locally set as pending. Effectively that binary will never be detected as accepted by GL0. What I think we need to do is discard (assume as accepted) all binaries that are below some binary detected as accepted. If binary with hash ABC is accepted on GL0 then every snapshot leading up to ABC had to be accepted as well. I would add such logic when updating accepted binaries e.g. find the highest index of binary found in GL0 snapshot and then take the slice above such index to remain as pending.

TheMMaciek (changes requested):

marcinwadon (approved):

marcinwadon (approved):

marcinwadon (approved):

TheMMaciek (approved):

TheMMaciek (approved):

marcinwadon (approved):

PR #808, opened 2023-12-19
Joining & peer discovery finishes before download

cngo-github:
Joining & peer discovery finishes completely before unlocking download

### DIFF
--- a/modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
+++ b/modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
@@ -1,21 +1,27 @@
 package org.tessellation.node.shared.domain.cluster.programs
 
-import cats.Applicative
+import cats.data.EitherT
 import cats.effect.Async
-import cats.effect.std.Queue
+import cats.effect.std.Supervisor
 import cats.syntax.applicative._
 import cats.syntax.applicativeError._
+import cats.syntax.either._
 import cats.syntax.flatMap._
+import cats.syntax.foldable._
 import cats.syntax.functor._
 import cats.syntax.option._
 import cats.syntax.order._
+import cats.syntax.parallel._
 import cats.syntax.show._
-import cats.syntax.traverse._
+import cats.{Applicative, MonadThrow, Parallel}
+
+import scala.util.control.NoStackTrace
 
 import org.tessellation.effects.GenUUID
 import org.tessellation.env.AppEnvironment
 import org.tessellation.env.AppEnvironment.Dev
 import org.tessellation.ext.crypto._
+import org.tessellation.node.shared.domain.Daemon
 import org.tessellation.node.shared.domain.cluster.services.{Cluster, Session}
 import org.tessellation.node.shared.domain.cluster.storage.{ClusterStorage, SessionStorage}
 import org.tessellation.node.shared.domain.healthcheck.LocalHealthcheck
@@ -25,6 +31,7 @@ import org.tessellation.node.shared.http.p2p.clients.SignClient
 import org.tessellation.schema.ID.Id
 import org.tessellation.schema.cluster._
 import org.tessellation.schema.node.NodeState
+import org.tessellation.schema.node.NodeState.{SessionStarted, inCluster}
 import org.tessellation.schema.peer.Peer.toP2PContext
 import org.tessellation.schema.peer._
 import org.tessellation.security.hash.Hash
@@ -32,110 +39,19 @@ import org.tessellation.security.signature.Signed
 import org.tessellation.security.{Hasher, SecurityProvider}
 
 import com.comcast.ip4s.{Host, IpLiteralSyntax}
-import fs2.{Pipe, Stream}
 import org.typelevel.log4cats.slf4j.Slf4jLogger
 
-object Joining {
-
-  def make[F[_]: Async: GenUUID: SecurityProvider: Hasher](
-    environment: AppEnvironment,
-    nodeStorage: NodeStorage[F],
-    clusterStorage: ClusterStorage[F],
-    signClient: SignClient[F],
-    cluster: Cluster[F],
-    session: Session[F],
-    sessionStorage: SessionStorage[F],
-    localHealthcheck: LocalHealthcheck[F],
-    seedlist: Option[Set[SeedlistEntry]],
-    selfId: PeerId,
-    stateAfterJoining: NodeState,
-    peerDiscovery: PeerDiscovery[F],
-    versionHash: Hash
-  ): F[Joining[F]] =
-    Queue
-      .unbounded[F, P2PContext]
-      .flatMap(
-        make(
-          _,
-          environment,
-          nodeStorage,
-          clusterStorage,
-          signClient,
-          cluster,
-          session,
-          sessionStorage,
-          localHealthcheck,
-          seedlist,
-          selfId,
-          stateAfterJoining,
-          peerDiscovery,
-          versionHash
-        )
-      )
-
-  def make[F[_]: Async: GenUUID: SecurityProvider: Hasher](
-    joiningQueue: Queue[F, P2PContext],
-    environment: AppEnvironment,
-    nodeStorage: NodeStorage[F],
-    clusterStorage: ClusterStorage[F],
-    signClient: SignClient[F],
-    cluster: Cluster[F],
-    session: Session[F],
-    sessionStorage: SessionStorage[F],
-    localHealthcheck: LocalHealthcheck[F],
-    seedlist: Option[Set[SeedlistEntry]],
-    selfId: PeerId,
-    stateAfterJoining: NodeState,
-    peerDiscovery: PeerDiscovery[F],
-    versionHash: Hash
-  ): F[Joining[F]] = {
-
-    val logger = Slf4jLogger.getLogger[F]
-
-    val joining = new Joining(
-      environment,
-      nodeStorage,
-      clusterStorage,
-      signClient,
-      cluster,
-      session,
-      sessionStorage,
-      localHealthcheck,
-      seedlist,
-      selfId,
-      stateAfterJoining,
-      versionHash,
-      joiningQueue
-    ) {}
-
-    def join: Pipe[F, P2PContext, Unit] =
-      in =>
-        in.evalMap { peer =>
-          {
-            joining.twoWayHandshake(peer, none) >>
-              clusterStorage
-                .getPeer(peer.id)
-                .flatMap {
-                  _.fold(Set.empty[P2PContext].pure[F]) { p =>
-                    peerDiscovery.discoverFrom(p).map(_.map(toP2PContext)).handleErrorWith { err =>
-                      logger.error(err)(s"Peer discovery from peer ${peer.show} failed").as(Set.empty)
-                    }
-                  }
-                }
-                .flatMap(_.toList.traverse(joiningQueue.offer(_).void))
-                .void
-          }.handleErrorWith { err =>
-            logger.error(err)(s"Joining to peer ${peer.show} failed")
-          }
-        }
-
-    val process = Stream.fromQueueUnterminated(joiningQueue).through(join).compile.drain
-
-    Async[F].start(process).as(joining)
-  }
+case class FailedToJoin(peer: Peer) extends NoStackTrace {
+  override def getMessage: String = s"Joining to peer ${peer.show} failed"
+}
+case class FailedToDiscoverFrom(peer: Peer, err: Throwable) extends NoStackTrace {
+  override def getMessage: String =
+    s"Peer discovery from peer ${peer.show} failed because ${err.getMessage}"
 }
 
-sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: Hasher] private (
+class Joining[
+  F[_]: Async: GenUUID: SecurityProvider: Hasher: Supervisor: Parallel
+](
   environment: AppEnvironment,
   nodeStorage: NodeStorage[F],
   clusterStorage: ClusterStorage[F],
@@ -148,18 +64,22 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: Hasher] pr
   selfId: PeerId,
   stateAfterJoining: NodeState,
   versionHash: Hash,
-  joiningQueue: Queue[F, P2PContext]
+  peerDiscovery: PeerDiscovery[F]
 ) {
 
   private val logger = Slf4jLogger.getLogger[F]
 
   def join(toPeer: PeerToJoin): F[Unit] =
-    for {
-      _ <- validateJoinConditions()
-      _ <- session.createSession
-
-      _ <- joiningQueue.offer(toPeer)
-    } yield ()
+    validateJoinConditions() >>
+      session.createSession >>
+      Daemon.spawn {
+        joinAll(toPeer) >>
+          clusterStorage.getResponsivePeers.flatMap { peers =>
+            nodeStorage
+              .tryModifyState(SessionStarted, stateAfterJoining)
+              .whenA(peers.nonEmpty)
+          }
+      }.start
 
   def rejoin(withPeer: PeerToJoin): F[Unit] =
     twoWayHandshake(withPeer, None, skipJoinRequest = true).void
@@ -173,7 +93,7 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: Hasher] pr
 
   def joinRequest(hasCollateral: PeerId => F[Boolean])(joinRequest: JoinRequest, remoteAddress: Host): F[Unit] = {
     for {
-      _ <- nodeStorage.getNodeState.map(NodeState.inCluster).flatMap(NodeNotInCluster.raiseError[F, Unit].unlessA)
+      _ <- nodeStorage.getNodeState.map(inCluster).flatMap(NodeNotInCluster.raiseError[F, Unit].unlessA)
       _ <- sessionStorage.getToken.flatMap(_.fold(SessionDoesNotExist.raiseError[F, Unit])(_ => Applicative[F].unit))
 
       registrationRequest = joinRequest.registrationRequest
@@ -188,6 +108,53 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: Hasher] pr
       _ <- twoWayHandshake(withPeer, remoteAddress.some, skipJoinRequest = true)
     } yield ()
   }.onError(err => logger.error(err)(s"Error during join attempt by ${joinRequest.registrationRequest.id.show}"))
+
+  private def joinTo(toJoin: P2PContext): EitherT[F, Throwable, Peer] =
+    twoWayHandshake(toJoin, none).attemptT
+      .flatMapF[Throwable, Peer] { p =>
+        clusterStorage
+          .getPeer(p.id)
+          .map(
+            Either.fromOption(
+              _,
+              FailedToJoin(p)
+            )
+          )
+      }
+
+  private def discoverFrom(peer: Peer): F[Set[Peer]] =
+    peerDiscovery
+      .discoverFrom(peer)
+      .handleErrorWith { e =>
+        MonadThrow[F].raiseError(FailedToDiscoverFrom(peer, e))
+      }
+
+  private def joinAndDiscover(toJoin: P2PContext): EitherT[F, Throwable, Set[P2PContext]] =
+    joinTo(toJoin)
+      .semiflatMap(discoverFrom)
+      .map(_.map(toP2PContext))
+
+  private def joinAll(peerToJoin: P2PContext): F[Unit] = {
+    type Agg = (Set[P2PContext], Set[P2PContext])
+    type Res = Unit
+
+    (Set(peerToJoin), Set.empty[P2PContext]).tailRecM {
+      case (toJoin, _) if toJoin.isEmpty => ().asRight[Agg].pure
+      case (toJoin, attempted) =>
+        for {
+          discovered <- toJoin.toList.parTraverse { peer =>
+            joinAndDiscover(peer).valueOrF { err =>
+              logger
+                .warn(err)(s"Failed to join and discover from ${peer.show}")
+                .as(Set.empty[P2PContext])
+            }
+          }
+          reduced = discovered.combineAll
+          updatedAttempted = attempted ++ toJoin
+          updatedToJoin = reduced.diff(updatedAttempted)
+        } yield (updatedToJoin, updatedAttempted).asLeft[Res]
+    }
+  }
 
   private def twoWayHandshake(
     withPeer: PeerToJoin,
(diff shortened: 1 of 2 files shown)


modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
```diff
@@ -190,6 +76,47 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
...
+      peerDiscovery
+        .discoverFrom(_)
+        .map(_.map(toP2PContext))
+        .handleErrorWith { err =>
+          logger
+            .error(err)(s"Peer discovery from P2PContext ${context.show} failed")
+            .as(Set.empty)
+        }
+    }
+    discovered <- maybeDiscovered.getOrElse(Set.empty[P2PContext].pure)
```
sdavidp, line 101: This might be cleaner if you use `OptionT`.

sdavidp (changes requested):

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
cngo-github: Implemented the suggested re-factor.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
```diff
@@ -190,6 +76,47 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
...
+    for {
+      nodeState <- nodeStorage.getNodeState
+      canJoinCluster <- nodeStorage.canJoinCluster
+      _ <- Applicative[F].unlessA(canJoinCluster)(NodeStateDoesNotAllowForJoining(nodeState).raiseError[F, Unit])
+    } yield ()
+
+  private def joinAndDiscover(context: P2PContext): F[Set[P2PContext]] = for {
+    _ <- twoWayHandshake(context, none).attempt.map(_.left.map { err =>
+      logger.error(err)(s"Joining to P2PContext ${context.show} failed")
+    })
```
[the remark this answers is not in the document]
cngo-github: Per our initial discussion with @TheMMaciek , failure of the `twoWayHandshake` shouldn't stop the overall joining process. If `twoWayHandshake` fails, then `clusterStorage.getPeer(context.id)` returns a `None` and we move on.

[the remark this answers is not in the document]
marcinwadon: IMO if it's an initial joining it should fail, but for consecutive joins the fail should be ignored and discovery should continue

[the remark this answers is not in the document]
TheMMaciek: Initial join being the node we send as a payload on the /join endpoint.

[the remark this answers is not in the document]
TheMMaciek: In the same time I'm not sure if we need a notion of the initial node and joining to it and a separate join to the rest of the nodes, as not being able to join to the initial node will end up with 0 nodes joined and we can easily detect it as a post joining step and have some logic around it.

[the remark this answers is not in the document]
cngo-github: If there are no joined `Peer`s, `PeerSelect` with throw an error, stopping the application. Is that insufficient? Is an additional check needed?

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
```diff
@@ -190,6 +76,47 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
...
+    type Res = Unit
+
+    (Set(toJoin): Set[P2PContext], Set.empty[P2PContext]).tailRecM {
+      case (joining, _) if joining.isEmpty =>
+        ().asRight[Agg].pure
+      case (joining, joined) =>
+        for {
+          discovered <- joining.toList.parTraverse(joinAndDiscover)
+          updatedJoined = joined ++ joining
+          updatedJoining = discovered.flatten.filterNot(joining.contains).toSet
```
TheMMaciek, line 115: Several points:

- I would investigate a bit more how does the discovery work internally to decide whether we need to filter out the nodes here additionally.
- Another thing if my intuition is right you should use `updatedJoined` to filter out discovered peers and not `joining` if filtering is needed.
- Another thing is that not all peers in `joining` will always be joined, so the second parameter called `joined` here should rather be called `attempted`.
- I would also think about maybe moving discovery to the `Joining` class and perform discovery based on the known details in the `joinAll` (the attempted, failed and joined nodes) and only offload the last discovered to the ref that would be available to fetch from.

TheMMaciek (commented):
For now it looks good overall - I don't see any obvious or critical semantic errors. I would consider some of the points I put in the comments to figure out the best solution.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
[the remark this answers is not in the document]
TheMMaciek: I think we could check if we joined to at least a single initial node here (not sure if every layer uses `PeerSelect`). I would also think about aggregating `joined` and `failed to join` peers to have some sanity check before continuing to the next step. For the simplest initial implementation I think we need at least this first check (if the initial node join succeeded.)

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
TheMMaciek: If you need to filter out in this line maybe that syntax would be better:
`updatedJoining = discovered.flatten.toSet -- joining`

cngo-github: > * I would investigate a bit more how does the discovery work internally to decide whether we need to filter out the nodes here additionally.

@TheMMaciek , while `discoverFrom` does filter out known peers, we are joining asynchronously. Additionally, some joinings may fail. So we still have to filter out peers we attempted to join.

> * Another thing if my intuition is right you should use `updatedJoined` to filter out discovered peers and not `joining` if filtering is needed afterall.

@TheMMaciek , please explain.

> * I would also think about maybe moving discovery to the `Joining` class and perform discovery based on the known details in the `joinAll` (the attempted, failed and joined nodes) and only offload the last discovered to the ref that would be available to fetch from.

@TheMMaciek , please elucidate further. I do not understand.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
```diff
@@ -190,6 +76,47 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
...
   }.onError(err => logger.error(err)(s"Error during join attempt by ${joinRequest.registrationRequest.id.show}"))
 
+  private def validateJoinConditions(): F[Unit] =
+    for {
+      nodeState <- nodeStorage.getNodeState
+      canJoinCluster <- nodeStorage.canJoinCluster
+      _ <- Applicative[F].unlessA(canJoinCluster)(NodeStateDoesNotAllowForJoining(nodeState).raiseError[F, Unit])
+    } yield ()
+
+  private def joinAndDiscover(context: P2PContext): F[Set[P2PContext]] = for {
```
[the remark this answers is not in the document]
cngo-github: The return type of `joinAndDiscover` has been changed to `EitherT[F, Throwable, Set[P2PContext]]`. There is no point in differentiating from a joining or peer discovery failure, since no discovered peers, joining failure, or peer discovery failure will be treated the same, as `Set.empty[P2PContext]`. The failures will simply be logged. Both joining and peer discovery failures will return `Throwable` as `Left`.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
cngo-github: > * Another thing is that not all peers in `joining` will always be joined, so the second parameter called `joined` here should rather be called `attempted`.

Changed `joining` to `attempted` and `updatedJoining` to `updatedAttempted`.

> * I would consider if we need another field/parameter for `failed` to join peers - that depends on which road we will choose (see next point)

Joining for each peer will only be attempted once. A `failed` set is not needed.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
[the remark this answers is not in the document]
TheMMaciek: Well so you won't be able to differentiate between successful join with unsuccessful discovery from the unsuccessful join. I'm only pointing it out in terms of the joining functionality and some heuristic we may want to run at the end of joining process where we could determine whether the joining succeeded overall or didn't. For example joining successfully to a single node out of 100 attempted can be treated as unhealthy condition and may be handled with failing the joining process. Not saying we need to do that, but I see the opportunity and a benefit from such check.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
TheMMaciek: > @TheMMaciek , while discoverFrom does filter out known peers, we are joining asynchronously. Additionally, some joinings may fail. So we still have to filter out peers we attempted to join.

I inspected the discovery functionality again and I think previously discovered peers are filtered out there, so I don't think that the additional filtering is necessary here - correct me if I'm wrong, maybe there is some corner case which I didn't get by briefly checking.

TheMMaciek: Currently the peer discovery is a kind of separate parallel being. I think it was created like this because we wanted to have both tracking of already discovered peers (to not rediscover those again and again on subsequent joins) and to be able to return discovered peers to whomever asks for those. Additionally the joining process before this change didn't keep track of the attempted peers. Right now we have the context of the attempted and joined peers so we could simply offload the discovered not joined peers to the Ref here in the `Joining` class. I think it may get simpler, WDYT @marcinwadon @kpudlik @sdavidp @scasplte2

TheMMaciek: > Joining for each peer will only be attempted once. A failed set is not needed.

Yes they will only be attempted once, but it may be beneficial to keep the information of which joined successfully and which didn't - especially if we would like to run some sanity check at the end of the joining process. Maybe the initial implementation could just log that information, so something like "Joining process finished. Joined X peers and failed to join Y peers ({optionally list of peer ids in a short format})" where X and Y are numbers.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
```diff
@@ -149,29 +46,19 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
...
   def join(toPeer: PeerToJoin): F[Unit] =
-    for {
-      _ <- validateJoinConditions()
-      _ <- session.createSession
-
-      _ <- joiningQueue.offer(toPeer)
-    } yield ()
+    validateJoinConditions() >>
+      session.createSession >>
+      Daemon.spawn(joinAll(toPeer)).start
```
TheMMaciek, line 57: It's started like that so that it returns immediately (the call on the endpoint finishes)?

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
```diff
@@ -190,6 +77,50 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
...
+    type Res = Unit
+
+    (Set(toJoin): Set[P2PContext], Set.empty[P2PContext]).tailRecM {
+      case (joining, _) if joining.isEmpty =>
+        ().asRight[Agg].pure
+      case (joining, attempted) =>
+        for {
+          discovered <- joining.toList.parTraverse { peer =>
+            joinAndDiscover(peer).leftMap { err =>
+              logger.warn(err)(s"Joining to and discovering from peer ${peer.show} failed")
```
TheMMaciek, line 111: Seems like this log may never happen given that you are mapping rather than flatMapping (it will be forever delayed I think)

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
```diff
@@ -190,6 +77,50 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
...
+  private def joinAll(toJoin: PeerToJoin): F[Unit] = {
+    type Agg = (Set[P2PContext], Set[P2PContext])
+    type Res = Unit
+
+    (Set(toJoin): Set[P2PContext], Set.empty[P2PContext]).tailRecM {
+      case (joining, _) if joining.isEmpty =>
+        ().asRight[Agg].pure
+      case (joining, attempted) =>
+        for {
+          discovered <- joining.toList.parTraverse { peer =>
```
TheMMaciek, line 109: I think it will always default to an empty collection of discovered peers (as traverse fails the whole collection traversing) on a single error and I don't think you are successfully handling the errors coming from `joinAndDisccover` function right now.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
```diff
@@ -190,6 +77,50 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: KryoSerial
...
+            joinAndDiscover(peer).leftMap { err =>
+              logger.warn(err)(s"Joining to and discovering from peer ${peer.show} failed")
+            }
+          }.value.map(_.foldLeft(List.empty[Set[P2PContext]]) {
```
TheMMaciek, line 113: If I understand correctly this piece of code what it seems you want is `.toOption.getOrElse(List.empty[Set[P2PContext])`. Using `foldLeft` here makes it a bit confusing in my opinion.

TheMMaciek (changes requested):
I pointed some doubts regarding the imlementation

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
cngo-github: Yes, we don't want to block the HTTP request. The request returns after starting the joining process.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
```diff
@@ -189,6 +101,58 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: Hasher] pr
...
+            joinAndDiscover(peer).leftSemiflatMap { err =>
+              logger.warn(err)(s"Failed to join and discover from ${peer.show}")
+            }
+          }.value
+          discovered = result.toOption
+            .getOrElse(List.empty[Set[P2PContext]])
+          reduced = discovered.foldLeft(Set.empty[P2PContext]) {
+            case (acc, curr) =>
+              acc ++ curr
+          }
```
sdavidp, line 147: 1. for `discovered =`, you don't need to convert `result` to an option; `Either.getOrElse` does what you want
2. the above is because `result` is typed with `Unit` on the left of the `Either` -- seems a bit odd
3. for `reduced =`, that can be simplified to `discovered.foldLeft(Set.empty[P2PContext])(_ ++ _)`
 or you can simplify it further as `discovered.combineAll` if you `import cats.syntax.foldable._`

sdavidp: Follow-up to the comment on the type of `result`, assuming you intend to keep the other functions returning an `EitherT` then maybe this is a bit cleaner:
```
discovered <- toJoin.toList.parTraverse { peer =>
  joinAndDiscover(peer).valueOrF(err =>
    logger.warn(err)(s"Failed to join and discover from ${peer.show}").as(Set.empty[P2PContext])
  )
}
```

cngo-github: Changed to

```scala
          discovered <- toJoin.toList.parTraverse { peer =>
            joinAndDiscover(peer).valueOrF { err =>
              logger
                .warn(err)(s"Failed to join and discover from ${peer.show}")
                .as(Set.empty[P2PContext])
            }
          }
          reduced = discovered.combineAll
```

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
cngo-github: Changed to

```scala
        for {
          discovered <- toJoin.toList.parTraverse { peer =>
            joinAndDiscover(peer).valueOrF { err =>
              logger
                .warn(err)(s"Failed to join and discover from ${peer.show}")
                .as(Set.empty[P2PContext])
            }
          }
          reduced = discovered.combineAll
          updatedAttempted = attempted ++ toJoin
          updatedToJoin = reduced.diff(updatedAttempted)
        } yield (updatedToJoin, updatedAttempted).asLeft[Res]
```

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
cngo-github: Changed to

```scala
        for {
          discovered <- toJoin.toList.parTraverse { peer =>
            joinAndDiscover(peer).valueOrF { err =>
              logger
                .warn(err)(s"Failed to join and discover from ${peer.show}")
                .as(Set.empty[P2PContext])
            }
          }
          reduced = discovered.combineAll
          updatedAttempted = attempted ++ toJoin
          updatedToJoin = reduced.diff(updatedAttempted)
        } yield (updatedToJoin, updatedAttempted).asLeft[Res]
```

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
cngo-github: Changed the signature to

```scala
  private def joinAndDiscover(toJoin: P2PContext): EitherT[F, Throwable, Set[P2PContext]]
```

So, the `parTraverse` will go through the entire list, accumulating errors.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
cngo-github: All failed joining and peer discovery attempts are now logged. We do need to filter previously discovered peers in addition to the filtering in the peer discovery function because we are joining and discovering from a list of `P2PContext`s in parrallel with `parTraverse`. So, we have to do a second round of filtering just in case.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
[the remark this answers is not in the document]
cngo-github: Unsuccessful joins return `FailedToJoin` while successful joins but unsuccessful peer discovery returns `FailedToDiscoverFrom`, so we can differentiate between the two cases.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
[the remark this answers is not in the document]
cngo-github: Node state trasition is now gated by a check of the `ClusterStorage`

```scala
...
    } >> clusterStorage.getResponsivePeers.flatMap { peers =>
      nodeStorage.tryModifyState(SessionStarted, stateAfterJoining).whenA(peers.nonEmpty)
    }
```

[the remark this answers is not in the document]
cngo-github: @TheMMaciek , if the initial peer failed to join, then we cannot discover other peers from it, so the joining process ends (due to no new peers to attempt to join to) and node state transition fails because the `ClusterStorage` is empty.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
```diff
@@ -238,7 +205,7 @@ sealed abstract class Joining[F[_]: Async: GenUUID: SecurityProvider: Hasher] pr
           localHealthcheck.cancel(registrationRequest.id),
           PeerAlreadyJoinedWithDifferentRegistrationData(registrationRequest.id).raiseError[F, Unit]
         )
-      _ <- nodeStorage.tryModifyStateGetResult(NodeState.SessionStarted, stateAfterJoining)
+      _ <- nodeStorage.tryModifyStateGetResult(SessionStarted, stateAfterJoining)
```
TheMMaciek, line 208: I think the whole point of this functionality was to postpone this state change until we attempt to join to all discovered peers. With this line of code the switch will still happen after joining to a first node so exactly as it was before. I think you need to remove it.

TheMMaciek (changes requested):

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
marcinwadon: is it a supervised start? It's not obvious looking at `Daemon.spawn` 🤔

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/Joining.scala
cngo-github: @TheMMaciek , you are correct. This line has been removed.

sdavidp (approved):

TheMMaciek (approved):

PR #810, opened 2024-01-04
L0PeerDiscovery filtering out discovered peers who are not last facilitators

kpudlik:


TheMMaciek (approved):
LGTM

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/cluster/programs/L0PeerDiscovery.scala
```diff
@@ -47,7 +46,9 @@ sealed abstract class L0PeerDiscovery[F[_]: Sync: Random] private (
       .map(_.head)
       .flatMap(l0ClusterStorage.getPeer)
       .flatMap(_.fold(Applicative[F].unit) { p =>
-        getPeersFrom(p).flatMap(l0ClusterStorage.setPeers)
+        getPeersFrom(p)
+          .map(_.filter(peer => lastFacilitators.map(_.toId).contains(peer.id.toId)).toSortedSet.toNes)
```
sdavidp, line 50: Isn't `lastFacilitators.map` constructing a temporary (discarded) collection each time the predicate is invoked? Could we avoid that by doing this?
```
.map(_.filter(peer => lastFacilitators.exists(_.toId === peer.id.toId)).toSortedSet.toNes)
```

kpudlik: Actually it can be just:

```scala
.map(_.filter(peer => lastFacilitators.contains(peer.id)).toSortedSet.toNes)
```

I will address that in next PR.

PR #811, opened 2024-01-09
Hasher type class

marcinwadon:


modules/shared/src/main/scala/org/tessellation/security/Hash.scala
```diff
@@ -52,19 +56,34 @@ object hash {
 }
 
 trait Hashable[F[_]] {
-  def hash[A <: AnyRef](data: A): Either[Throwable, Hash]
+  def hash[A: Encoder](data: A): F[Hash]
 }
 
 object Hashable {
```
kpudlik, line 62: I am not sure if `Hashable` is a valid name anymore. Previously it was used on "hashable" things but now it is a type-class so I am thinking about a different name like: "Hasher" or sth. WDYT?

kpudlik (changes requested):

modules/shared/src/main/scala/org/tessellation/security/Hash.scala
```diff
@@ -52,19 +56,34 @@ object hash {
 }
 
 trait Hashable[F[_]] {
-  def hash[A <: AnyRef](data: A): Either[Throwable, Hash]
+  def hash[A: Encoder](data: A): F[Hash]
```
sdavidp, line 59: Shouldn't the typeclass be typed on the object we're hashing rather than a container?
E.g. are we requiring `IO[A]` to be hashable or `A` ?

modules/shared/src/main/scala/org/tessellation/security/Hash.scala
marcinwadon: @TheMMaciek suggested that the name looks fine as it represents some capability. So you add the capability to the effect type. I can change the name if you want, but I cannot think a better alternative tbh

modules/shared/src/main/scala/org/tessellation/security/Hash.scala
marcinwadon: We are requiring any `A` that has an `Encoder` to be hashable under the effect type `F[_]`.
I don't understand your concern though, can you rephrase that?

marcinwadon: We've texted on Slack and tl;dr; - let's follow the pattern used by `Encoder` from `io.circe`. I will reorganize the PR

marcinwadon: tl;dr; that pattern does not allow for dependencies and cannot be used here

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/domain/block/BlockRelationsSuite.scala
```diff
@@ -4,23 +4,31 @@ import cats.effect.{IO, Resource}
...
-    KryoSerializer.forAsync[IO](Main.kryoRegistrar ++ nodeSharedKryoRegistrar)
+  def sharedResource: Resource[IO, Res] = for {
+    implicit0(ks: KryoSerializer[IO]) <- KryoSerializer.forAsync[IO](Main.kryoRegistrar ++ nodeSharedKryoRegistrar)
+    implicit0(j: JsonHashSerializer[IO]) <- JsonHashSerializer.forSync[IO].asResource
+    h = Hashable.forSync[IO]
+  } yield h
 
-  test("when no relation between blocks then block is independent") { implicit ks =>
+  // override def sharedResource: Resource[IO, BlockServiceSuite.Res] =
+  // KryoSerializer.forAsync[IO](Main.kryoRegistrar ++ nodeSharedKryoRegistrar)
```
TheMMaciek, line 29: Leftover

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/domain/block/BlockRelationsSuite.scala
```diff
@@ -4,23 +4,31 @@ import cats.effect.{IO, Resource}
...
 object BlockRelationsSuite extends MutableIOSuite with Checkers {
 
-  type Res = KryoSerializer[IO]
+  type Res = Hashable[IO]
 
-  override def sharedResource: Resource[IO, BlockServiceSuite.Res] =
-    KryoSerializer.forAsync[IO](Main.kryoRegistrar ++ nodeSharedKryoRegistrar)
+  def sharedResource: Resource[IO, Res] = for {
+    implicit0(ks: KryoSerializer[IO]) <- KryoSerializer.forAsync[IO](Main.kryoRegistrar ++ nodeSharedKryoRegistrar)
+    implicit0(j: JsonHashSerializer[IO]) <- JsonHashSerializer.forSync[IO].asResource
```
TheMMaciek, line 24: Is that notation so that the `KryoSerializer` and `JsonHashSerializer` are implicitly available for `Hashable`?

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/domain/snapshot/programs/SnapshotProcessorSuite.scala
```diff
@@ -73,84 +73,92 @@ object SnapshotProcessorSuite extends SimpleIOSuite with TransactionGenerator {
...
-                  jsonBrotliBinarySerializer
-                ),
-              Amount(0L)
-            )
-            globalSnapshotContextFns = GlobalSnapshotContextFunctions.make[IO](globalSnapshotAcceptanceManager)
-            snapshotProcessor = {
-              val addressStorage = new AddressStorage[IO] {
-                def getState: IO[Map[Address, Balance]] =
-                  balancesR.get
+        JsonHashSerializer.forSync[IO].asResource.flatMap { implicit jhs =>
```
TheMMaciek, line 76: Why not use the `implicit0` notation here?

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/domain/transaction/TransactionStorageSuite.scala
```diff
@@ -32,29 +33,32 @@ object TransactionStorageSuite extends SimpleIOSuite with TransactionGenerator {
...
-        for {
-          lastAccepted <- MapRef.ofConcurrentHashMap[IO, Address, LastTransactionReferenceState]().asResource
-          waitingTransactions <- MapRef.ofConcurrentHashMap[IO, Address, NonEmptySet[Hashed[Transaction]]]().asResource
-          transactionStorage = new TransactionStorage[IO](lastAccepted, waitingTransactions, TransactionReference.empty)
-          key1 <- KeyPairGenerator.makeKeyPair.asResource
-          address1 = key1.getPublic.toAddress
-          key2 <- KeyPairGenerator.makeKeyPair.asResource
-          address2 = key2.getPublic.toAddress
-        } yield (transactionStorage, lastAccepted, waitingTransactions, key1, address1, key2, address2, sp, kp)
+        JsonHashSerializer.forSync[IO].asResource.flatMap { implicit jhs =>
```
TheMMaciek, line 42: `implicit0` notation?

modules/shared/src/main/scala/org/tessellation/json/JsonBrotliBinarySerializer.scala
```diff
@@ -19,12 +19,21 @@ trait JsonBrotliBinarySerializer[F[_]] {
...
+  }
+
+  def forSync[F[_]: Sync](printer: Printer): F[JsonBrotliBinarySerializer[F]] =
     Sync[F].delay(Brotli4jLoader.ensureAvailability()).map { _ =>
       new JsonBrotliBinarySerializer[F] {
+
         def serialize[A: Encoder](content: A): F[Array[Byte]] = {
           val params = new Parameters().setQuality(compressionLevel)
-          Sync[F].delay(BrotliEncoder.compress(content.asJson.noSpaces.getBytes("UTF-8"), params))
+          Sync[F].delay(BrotliEncoder.compress(content.asJson.printWith(printer).getBytes("UTF-8"), params))
```
TheMMaciek, line 36: This may be a breaking change.

modules/shared/src/main/scala/org/tessellation/kryo/KryoSerializer.scala
```diff
@@ -13,7 +13,7 @@ import com.twitter.chill._
 
 trait KryoSerializer[F[_]] {
 
-  def serialize(anyRef: AnyRef): Either[Throwable, Array[Byte]]
+  def serialize[A](anyRef: A): Either[Throwable, Array[Byte]]
```
TheMMaciek, line 16: Not an `anyRef` anymore, maybe we should call it `data` or simply `a` 🤔

TheMMaciek: Do you remember why it was an `AnyRef` originally? Wasn't it because of some Kryo's requirement/limitation? Can we change it then?

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/domain/block/BlockRelationsSuite.scala
marcinwadon: yes. we have a better-monadic-for in the dependencies, so we can use it here: https://github.com/oleg-py/better-monadic-for?tab=readme-ov-file#define-implicits-in-for-comprehensions-or-matches

That's way it's easier to read and extend than normal chaining using `.map` and `.flatMap`

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/domain/snapshot/programs/SnapshotProcessorSuite.scala
marcinwadon: well, I can, but the kryo was already created like that, so just added another one

modules/shared/src/main/scala/org/tessellation/json/JsonBrotliBinarySerializer.scala
marcinwadon: it is not. it's using the same configuration, I just used an explicit way to render the json.
I'm using the following printer:
```scala
    def printer = Printer(dropNullValues = false, indent = "")
```

and `noSpaces` is configured as:
```scala
  final def noSpaces: String = Printer.noSpaces.print(this)

  /* --- */

  /**
   * A pretty-printer configuration that inserts no spaces.
   */
  final val noSpaces: Printer = Printer(
    dropNullValues = false,
    indent = ""
  )

```

modules/shared/src/main/scala/org/tessellation/kryo/KryoSerializer.scala
marcinwadon: I don't remember, probably it was a leftover or sth. Everything serializes correctly, even the primitives, so it shouldn't be an issue

modules/shared/src/main/scala/org/tessellation/security/Hash.scala
TheMMaciek: Now that I look at it I think it may make sense to have it called `Hasher`, it would make sense to have it called `Hashable` if we would have it on the object it hashes (A) rather than on `F[_]`. Calling it something like `HashingAble` would reflect the functionality more, but it sounds bad.

modules/dag-l1/src/test/scala/org/tessellation/dag/l1/domain/snapshot/programs/SnapshotProcessorSuite.scala
TheMMaciek: In other places you did a refactor I think. Up to you.

modules/shared/src/main/scala/org/tessellation/json/JsonBrotliBinarySerializer.scala
TheMMaciek: Ok great!

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/DataApplication.scala
```diff
@@ -46,6 +45,8 @@ object DataApplication {
     lastCurrencySnapshotStorage: LastSnapshotStorage[F, CurrencyIncrementalSnapshot, CurrencySnapshotInfo]
   ): Stream[F, Unit] = {
 
+    implicit val dataEncoder = dataApplicationService.dataEncoder
```
TheMMaciek, line 48: Needed in multiple places I guess?

marcinwadon: What exactly and why?

modules/shared/src/main/scala/org/tessellation/security/signature/Signed.scala
```diff
@@ -99,17 +98,19 @@ object Signed {
     def isSignedExclusivelyBy(signers: Set[Id]): Boolean =
       signed.proofs.map(_.id).toSortedSet.unsorted === signers
 
-    def hasValidSignature[F[_]: Async: SecurityProvider: KryoSerializer]: F[Boolean] =
+    def hasValidSignature[F[_]: Async: SecurityProvider: Hasher](implicit encoder: Encoder[A]): F[Boolean] =
       validProofs(None).map(_.isRight)
 
-    def hasValidSignature[F[_]: Async: SecurityProvider: KryoSerializer](toBytes: A => F[Array[Byte]]): F[Boolean] =
+    def hasValidSignature[F[_]: Async: SecurityProvider: Hasher](toBytes: A => F[Array[Byte]]): F[Boolean] = {
+      implicit val encoder: Encoder[A] = Encoder.instance(_ => Json.Null)
```
TheMMaciek, line 105: Can you explain what is this about?

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/DataApplication.scala
marcinwadon: Ah, ok. I thought it's needed in multiple places but looks like it's not. Moved to the specific function :+1:

modules/shared/src/main/scala/org/tessellation/security/signature/Signed.scala
marcinwadon: Yep - I have two cases:
1. check `hasValidSignature` using the default serializer logic (kryo in our case)
2. check `hasValidSignature` using the custom serializer logic (provided fn `toBytes: A => F[Array[Byte]]`

Both calls `validProofs` and because I didn't want to write two `validProofs` methods, one working on kryo and second one working on custom provided fn, I pass the optional argument (default None for default logic). BUT I need to pass an encoder as well for the default logic. So for the second case, I pass the "noop" encoder. 

As I think about this right now, I can require `validProofs` to expect either the encoder (explicitly) or the custom logic fn. Should be better approach. Changing now!

marcinwadon: changed https://github.com/Constellation-Labs/tessellation/compare/93eb312acde435334d89f35406ecc5f56f59b4b5..5ecc90387aba4e8634bd69e513152ad2576ae76b

kpudlik (approved):

PR #812, opened 2024-01-16
Updating mainnet-seedlist

IPadawans:


AlexBrandes:
LGTM - would you edit the description to remove the google docs link though? We shouldn't make that public.

IPadawans:
> LGTM - would you edit the description to remove the google docs link though? We shouldn't make that public.

Done

AlexBrandes (approved):

PR #814, opened 2024-01-23
Add owner/stake to CurrencyIncrementalSnapshot

sdavidp:


marcinwadon (dismissed):
I think this code should not be merged until we finish schema migration ticket. Anyway, LGTM

modules/shared/src/main/scala/org/tessellation/currency/schema/currency.scala
```diff
@@ -105,7 +113,9 @@ object currency {
     stateProof: CurrencySnapshotStateProof,
     epochProgress: EpochProgress,
     dataApplication: Option[DataApplicationPart] = None,
-    version: SnapshotVersion = SnapshotVersion("0.0.1")
+    version: SnapshotVersion = SnapshotVersion("0.0.1"),
+    adminFeeKeyMessage: Option[AdminFeeKeyMessage] = None,
+    stakingMessage: Option[StakingMessage] = None
```
TheMMaciek, line 118: We can't really add it right now (schema change). One thing I was pointing out in a discussion earlier is that maybe it should be a generic (or sealed but extendable) `Event` and a state of such events would be persisted in the currencySnapshotInfo. I don't think we need these fields on the snapshot itself as they seem to remain mostly constant right? Anyway we can't merge it yet as this is a plain schema change (which we are not handling right now).

marcinwadon: I'd agree that instead of rewriting the same data over and over again, it should happen once and be stored for easy access in the offchain data (currency snapshot info). @ryle-goehausen ?

sdavidp: I had the same thought and mentioned it to Ryle yesterday. He mentioned we're adding it here because of key rotation; subsequent runs could change those values. Is there a better place for us to put it where we can achieve the same goal?

@ryle-goehausen please correct me if misunderstood our discussion.

marcinwadon: @sdavidp I don't see why `subsequent runs could change those values` affect the place and how we would store it. Can you elaborate more on that here or on slack?

marcinwadon: If I understand correctly:

Option 1:
-  snapshot A (stakingMessage: Signed(_abcdef_))
-  snapshot B (stakingMessage: Signed(_abcdef_))
**<-- changing the staking message to _zxcvbn_**
-  snapshot C (stakingMessage: Signed(_zxcvbn_))
-  snapshot D (stakingMessage: Signed(_zxcvbn_))

Option 2:

- snapshot A (events: stakingMessage=Signed(_abcdef_)) `last stakingMessage=Signed(_abcdef_)`
- snapshot B () `last stakingMessage=Signed(_abcdef_)`
**<-- changing the staking message to _zxcvbn_**
- snapshot C (events: stakingMessage=Signed(_zxcvbn_)) `last stakingMessage=Signed(_zxcvbn_)`
- snapshot D () `last stakingMessage=Signed(_zxcvbn_)`

where `lastStakingMessage` can be stored in snapshot companion object (snapshot info or equivalent).

The latter is more generic and more flexible as it allows to easily extend `events` type to handle more type of events in the future.

ryle-goehausen: > @sdavidp I don't see why subsequent runs could change those values affect the place and how we would store it. Can you elaborate more on that here or on slack?

We'd expect to be able to change the keys if they're not established in the beginning or for any other reason.

It should look like (for increasing snapshot incrementals,)

ordinal 10 - message:None - state: None
ordinal 11 - message:Signed(x) - state: Signed(x)
ordinal 12 - message:None - state: Signed(x)
ordinal 13 - message:Signed(y) - state: Signed(y)
ordinal 14 - message:None - state: Signed(y)

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/cli/method.scala
```diff
@@ -84,7 +86,9 @@ object method {
     prioritySeedlistPath: Option[SeedListPath],
     collateralAmount: Option[Amount],
     globalL0Peer: L0Peer,
-    trustRatingsPath: Option[Path]
+    trustRatingsPath: Option[Path],
+    adminFeeKeyMessage: Option[String],
```
ryle-goehausen, line 90: These should not be added here, the correct way to implement this is as a new top level command similar to:

```
  object CreateGenesis extends WithOpts[CreateGenesis] {
```

That command is a unique way to create a genesis snapshot, we should implement a new command to produce an admin / staking key message (with a signing function from the wallet.) The options should be removed here, since none of the other commands will ever need to include this value.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/cli/method.scala
```diff
@@ -137,7 +145,9 @@ object method {
         SeedListPath.priorityOpts,
         CollateralAmountOpts.opts,
         GlobalL0PeerOpts.opts,
-        trustRatingsPathOpts
+        trustRatingsPathOpts,
+        KeyMessageOpts.adminFeeKeyMessageOpts,
```
ryle-goehausen, line 149: Remove these from here too

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/cli/method.scala
```diff
@@ -172,7 +184,9 @@ object method {
         CollateralAmountOpts.opts,
         GlobalL0PeerOpts.opts,
         L0TokenIdentifierOpts.opts,
-        trustRatingsPathOpts
+        trustRatingsPathOpts,
+        KeyMessageOpts.adminFeeKeyMessageOpts,
```
ryle-goehausen, line 188: Remove these from here

modules/shared/src/main/scala/org/tessellation/currency/schema/KeyMessage.scala
```diff
@@ -0,0 +1,46 @@
...
+  @derive(eqv, show, encoder, decoder)
+  @newtype
+  case class StakingMessage(value: Signed[Address])
+
+  object StakingMessage {
+    def apply[F[_]: Async: SecurityProvider: Hasher](s: String, keyPair: KeyPair): F[StakingMessage] =
+      toSignedAddress(s, keyPair).map(StakingMessage(_))
+  }
+
+  private def toSignedAddress[F[_]: Async: SecurityProvider: Hasher](s: String, keyPair: KeyPair): F[Signed[Address]] =
```
ryle-goehausen, line 40: This seems correct to me, and this is the output that we want to print out when a user runs the appropriate CLI command.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/cli/method.scala
sdavidp: @ryle-goehausen What is the new command going to be called? `run-genesis-with-signing` or ... ?
And should the new command be a clone of `CreateGenesis` but overriding the two message fields?

ryle-goehausen: Hey, sorry, I missed this comment earlier, the goal would be to create a single function, separate from CreateGenesis, that does not implement any of the functionality of create genesis, so it can be used for an existing metagraph.

I.e. the CLI argument should only take those two parameters (with at least one filled,) and use the wallet utilities / signing functions to produce an output blob which we can later submit to the endpoint.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/cli/method.scala
```diff
@@ -142,6 +144,45 @@ object method {
     }
   }
 
+  case class RunGenesisWithSigning(
```
ryle-goehausen, line 147: So just to be clear from the earlier comment, I'm not sure this mode actually needs to be supported (but it may actually be useful from user perspective so they can use a single command to both build the genesis and sign things.)

The main command we need to support right now is just a single CLI command RunFeeStakingSigning which only accepts those two parameters, and outputs a signed blob corresponding to the keys loaded from the conventional wallet environment variable paths.

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/SigningMessageRoutes.scala
```diff
@@ -0,0 +1,40 @@
...
+import org.tessellation.node.shared.infrastructure.snapshot.services.SigningMessageService
+import org.tessellation.node.shared.infrastructure.snapshot.storage.SigningMessageStorage
+import org.tessellation.routes.internal._
+import org.tessellation.security.{Hasher, SecurityProvider}
+
+import eu.timepit.refined.auto._
+import org.http4s.HttpRoutes
+import org.http4s.dsl.Http4sDsl
+
+// TODO proper location/module for this class?
```
ryle-goehausen, line 16: Currency L0 is sufficient as far as I'm concerned, considering this is a very lightly used event I don't think we need to make another layer of indirection in the APIs in terms of storing on blocks, but open to discussion on that.

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/SigningMessageRoutes.scala
```diff
@@ -0,0 +1,40 @@
...
+// TODO proper name for this class?
```
ryle-goehausen, line 17: AdminFeeStakingSigningRoutes, MetagraphMessageRoutes, or some variation on this, also somewhat okay as is, but it might be ambiguous as is?

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/MetagraphMessageRoutes.scala
```diff
@@ -0,0 +1,38 @@
...
+  service: MetagraphMessageService[F],
+  storage: MetagraphMessageStorage[F]
+) extends Http4sDsl[F]
+    with PublicRoutes[F] {
+
+  protected val prefixPath: InternalUrlPrefix = "/currency"
+
+  protected val public: HttpRoutes[F] = HttpRoutes.of[F] {
+    case POST -> Root / "signing" / "owner" / dagAddress =>
+      service
```
ryle-goehausen, line 26: Hey, so one thing I see missing here is an invocation of a validator, I'm not certain this is implemented in the storage itself, but we need to validate upon the request whether or not it's signed by a majority of the seed list. 

Just for clarification, we can merge the signatures that are produced manually from the CLI method to produce a majority signed message.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/storage/MetagraphMessageStorage.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  def make[F[_]: Functor: Ref.Make](valuesR: Ref[F, StoredTuple]): MetagraphMessageStorage[F] =
+    new MetagraphMessageStorage[F] {
+      def get: F[StoredTuple] =
+        valuesR.get
+
+      def setAdminFeeKeyMessage(newMessage: Option[AdminFeeKeyMessage]): F[Unit] =
+        valuesR.update {
+          case (_, maybeStaking) => (newMessage, maybeStaking)
+        }
```
marcinwadon, line 32: you can store both in two separate Refs 🤔 just an idea as handling tuples may be more complicated

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/MetagraphMessageRoutes.scala
```diff
@@ -0,0 +1,38 @@
...
+class MetagraphMessageRoutes[F[_]: Async: SecurityProvider: Hasher](
+  service: MetagraphMessageService[F],
+  storage: MetagraphMessageStorage[F]
+) extends Http4sDsl[F]
+    with PublicRoutes[F] {
+
+  protected val prefixPath: InternalUrlPrefix = "/currency"
+
+  protected val public: HttpRoutes[F] = HttpRoutes.of[F] {
+    case POST -> Root / "signing" / "owner" / dagAddress =>
```
marcinwadon, line 25: Shouldn't it receive already signed message? 🤔

sdavidp: Implementation changed to process payload

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/storage/MetagraphMessageStorage.scala
sdavidp: done

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/MetagraphMessageRoutes.scala
sdavidp: rewritten

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/storage/MetagraphMessageStorage.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  def make[F[_]: Functor: Semigroupal: Ref.Make](
+    adminFeeKeyR: Ref[F, Option[AdminFeeKeyMessage]],
+    stakingR: Ref[F, Option[StakingMessage]]
+  ): MetagraphMessageStorage[F] =
+    new MetagraphMessageStorage[F] {
+      def getAdminFeeKeyMessage: F[Option[AdminFeeKeyMessage]] =
+        adminFeeKeyR.get
+
+      def setAdminFeeKeyMessage(maybe: Option[AdminFeeKeyMessage]): F[Unit] =
```
TheMMaciek, line 34: Why maybe? At this point we should either set it or not I guess. None may only result from not ever setting this value.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/storage/MetagraphMessageStorage.scala
```diff
@@ -0,0 +1,44 @@
...
+        adminFeeKeyR.set(maybe)
```
TheMMaciek, line 35: Should it just override whatever is in the storage without any checks?

modules/shared/src/main/scala/org/tessellation/schema/metagraphMessage.scala
```diff
@@ -0,0 +1,20 @@
...
+object metagraphMessage {
+  sealed trait SigningMessage {
+    def signed: Signed[Address]
+  }
+
+  @derive(eqv, show, encoder, decoder)
+  case class AdminFeeKeyMessage(signed: Signed[Address]) extends SigningMessage
+
+  @derive(eqv, show, encoder, decoder)
+  case class StakingMessage(signed: Signed[Address]) extends SigningMessage
```
TheMMaciek, line 19: I honestly think that it should rather be these classes that get signed and also I think some ordinal would be required. So that a message can't be replayed and can't overwrite new message with an old one.

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/MetagraphMessageRoutes.scala
```diff
@@ -0,0 +1,57 @@
...
+  protected val public: HttpRoutes[F] = HttpRoutes.of[F] {
+    case req @ POST -> Root / "signing" / "owner" =>
+      for {
+        signedAddress <- req.as[Signed[Address]]
+        response <- service.offerAdminFeeKeyMessage(signedAddress).flatMap(toResponse(signedAddress, _))
+      } yield response
+
+    case req @ POST -> Root / "signing" / "staking" =>
+      for {
+        signedAddress <- req.as[Signed[Address]]
```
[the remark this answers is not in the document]
ryle-goehausen: Yeah, so just to clarify again, it's fine here to introduce a new type for each of Signed values, that contains some essential difference (in data fields,) such that it serializes differently, i.e. EventType.StakingMessage or some equivalent.

Both fields require an ordinal, so to fix this we just need to introduce some piece of data that distinguishes the event type, either that, or a field name difference so it doesn't serialize the same.

modules/shared/src/main/scala/org/tessellation/schema/metagraphMessage.scala
ryle-goehausen: Key note of distinction here, if you serialize them both and the data is equivalent, it's not enough, it needs some differentiator (even besides ordinal.)

modules/tools/src/main/scala/org/tessellation/tools/Main.scala
```diff
@@ -162,6 +165,20 @@ object Main
       _ <- checkLastReferences(client, basicOpts.baseUrl, addressParams.map(_.address))
     } yield ()
 
+  def createSigningMessage[F[_]: Async: Hasher: SecurityProvider: Console](
```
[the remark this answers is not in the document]
ryle-goehausen: After discussion, let's relocate this to the wallet module

ryle-goehausen:
Summary from prior discussion, main change to introduce is to add a TransactionOrdinal equivalent to the message classes, and some distinction associated with their hash construction. This should be a CLI parameter, and passed in manually.

modules/shared/src/main/scala/org/tessellation/schema/metagraphMessage.scala
TheMMaciek: Yes with kryo these classes will be serialized as different binaries. With json though they may be serialized to the same output - hence for the json we need to modify the logic to also include the top level class to be persisted. We could also add a json classes registry used to salt the binary right beofre signing, and to be added to the binary before checking the signature.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/cli/method.scala
sdavidp: removed

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/cli/method.scala
sdavidp: removed

modules/shared/src/main/scala/org/tessellation/schema/address.scala
```diff
@@ -23,6 +25,11 @@ object address {
 
   object Address {
 
+    def fromString(str: String): Either[Exception, Address] =
```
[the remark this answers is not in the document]
sdavidp: removed

modules/tools/src/main/scala/org/tessellation/tools/Main.scala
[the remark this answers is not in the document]
sdavidp: done

modules/shared/src/main/scala/org/tessellation/schema/metagraphMessage.scala
ryle-goehausen: This should be addressed on recent change with field names

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/modules/HttpApi.scala
```diff
@@ -121,6 +121,8 @@ sealed abstract class HttpApi[F[_]: Async: SecurityProvider: KryoSerializer: Has
   private val metricRoutes = MetricRoutes[F]().publicRoutes
   private val targetRoutes = TargetRoutes[F](services.cluster).publicRoutes
 
+  private val metagraphMessageRoutes = new MetagraphMessageRoutes[F](services.metagraphMessageService).publicRoutes
```
marcinwadon, line 124: afaik we use `data application` name within the codebase, not `metagraph` 🤔 is the functionality clearly for metagraphs and you distinguish the metagraph from any other data application?

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/MetagraphMessageRoutes.scala
```diff
@@ -0,0 +1,57 @@
...
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+
+class MetagraphMessageRoutes[F[_]: Async: SecurityProvider: Hasher](
+  service: MetagraphMessageService[F]
+) extends Http4sDsl[F]
+    with PublicRoutes[F] {
+
+  private val logger = Slf4jLogger.getLoggerFromName("MetagraphMessageRoutes")
+
+  protected val prefixPath: InternalUrlPrefix = "/currency"
```
marcinwadon, line 30: why not `/metagraph` if it's clearly for metagraphs? if it's not, then either `CurrencytMessageRoutes` or `DataApplicationMessageRoutes`

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/MetagraphMessageRoutes.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  protected val public: HttpRoutes[F] = HttpRoutes.of[F] {
+    case req @ POST -> Root / "signing" / "owner" =>
```
marcinwadon, line 33: maybe `owner/sign` and `staking/sign`? also, looks like it does not sign it here, so maybe just `owner` and `staking` without `sign` or `signing`?

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/services/MetagraphMessageService.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  def make[F[_]: Async: SecurityProvider: Hasher](
+    validator: SignedValidator[F],
+    seedlist: Option[Set[SeedlistEntry]],
+    storage: MetagraphMessageStorage[F]
+  ): MetagraphMessageService[F] =
+    new MetagraphMessageService[F] {
+      val peers = seedlist.map(_.map(_.peerId))
+
+      def offerAdminFeeKeyMessage(message: AdminFeeKeyMessage): F[MetagraphMessageServiceErrorOr[Unit]] =
```
marcinwadon, line 36: just `offer(message: AdminFeeKeyMessage)` and `offer(message: StakingMessage)` would be enough IMO. the parameter would be self explaining and you would have just `messageService.offer(message)`

modules/shared/src/main/scala/org/tessellation/schema/metagraphMessage.scala
```diff
@@ -0,0 +1,17 @@
...
+import org.tessellation.schema.transaction.TransactionOrdinal
+import org.tessellation.security.signature.Signed
+
+import derevo.cats.{eqv, show}
+import derevo.circe.magnolia.{decoder, encoder}
+import derevo.derive
+
+object metagraphMessage {
+  @derive(eqv, show, encoder, decoder)
+  case class AdminFeeKeyMessage(transactionOrdinal: TransactionOrdinal, adminFeeKeyAddress: Signed[Address])
```
marcinwadon, line 13: Please take a look how `TransactionReference` is constructed. IMO you need to replicate that, so:
```scala
case class MessageReference(ordinal: MessageOrdinal, hash: Hash)
```

Please don't reuse `TransactionOrdinal` since the message isn't the transaction itself.

modules/shared/src/main/scala/org/tessellation/security/signature/SignedValidator.scala
```diff
@@ -105,6 +105,23 @@ object SignedValidator {
           .toNes
       }.map(SignersNotInSeedlist).toInvalidNec(signed)
 
+    def validateSignedByMajority[A](seedlist: Option[Set[PeerId]], signed: Signed[A]): SignedValidationErrorOr[Signed[A]] =
```
marcinwadon, line 108: `ByMajority` is misleading as if there is no seedlist, it's always valid. Either there should be a minimum no of signatures in case of no seedlist, or the fn name should be updated to sth like `validateSignedBySeedlistMajority`

modules/shared/src/main/scala/org/tessellation/security/signature/SignedValidator.scala
```diff
@@ -105,6 +105,23 @@ object SignedValidator {
...
+      seedlist match {
+        case None => signed.validNec
+        case Some(peers) =>
+          val minRequired = peers.size / 2 + 1
+          val signingPeers = signed.proofs
+            .map(_.id.toPeerId)
+            .toSortedSet
```
marcinwadon, line 115: do we need the set to be sorted?

modules/shared/src/test/scala/org/tessellation/security/signature/SignedValidatorSuite.scala
```diff
@@ -92,6 +100,77 @@ object SignedValidatorSuite extends MutableIOSuite {
...
+    forall(signedOf(addressGen)) { signedAddress =>
+      IO {
+        val actual = mkValidator().validateSignedByMajority(None, signedAddress)
+        val expected = signedAddress.validNec
+        expect.same(actual, expected)
+      }
+    }
+  }
+
+  test("validateSignedByMajority should fail on empty seedlist") { res =>
```
marcinwadon, line 115: I think that the empty seedlist is the same case as the next test, but there is no test for no seedlist.

modules/wallet/src/main/scala/org/tessellation/wallet/Main.scala
```diff
@@ -107,6 +118,26 @@ object Main
...
+
+  private def createOwnerMessage[F[_]: Async: SecurityProvider: Hasher: Console](
+    keyPair: KeyPair,
+    address: Address,
+    ordinal: transaction.TransactionOrdinal
+  ): F[Unit] =
+    Signed
+      .forAsyncKryo[F, Address](address, keyPair)
+      .map(AdminFeeKeyMessage(ordinal, _))
+      .flatMap(m => Console[F].println(m.asJson.spaces2))
```
marcinwadon, line 139: Question - should it print the json or should it write it down to some file, to be later sent as a json file?

marcinwadon (changes requested):

modules/shared/src/main/scala/org/tessellation/schema/metagraphMessage.scala
sdavidp: @ryle-goehausen, @marcinwadon Can you get together to discuss this? I'm not sure how this fits with what I was asked to do.

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/MetagraphMessageRoutes.scala
ryle-goehausen: I'm guessing since it needs to distinguish it as corresponding to the currency related snapshot schema. The exact naming here is not extremely important.

modules/shared/src/main/scala/org/tessellation/schema/metagraphMessage.scala
ryle-goehausen: If we want to follow the convention here established for wrapping it in a reference, that's fine, I don't think it will actually make a difference in this case though, since we're not using the previous message as a validation criteria for the new message, so it's irrelevant what the hash reference actually is (and whether or not the nodes agree on it.) The ordinal alone protects against replay attacks, so I don't think this serves a functional purpose other than organization.

Considering that this message acts as a transaction (maybe not our type, but the generic type,) and we want to introduce additional transaction types in the future, I don't think it makes sense to come up with a unique case class for each usage of TransactionOrdinal. If we want to include the hash though for bookkeeping purposes or other traversal or indexing purposes (or in case we want to change the behavior in the future to require some previous data validation), that might make sense though. So to me this seems more about organization, if it's easy to change then I'd rather just re-use transactionReference rather than creating a unique type for each future transaction we introduce.

modules/shared/src/main/scala/org/tessellation/security/signature/SignedValidator.scala
sdavidp: @marcinwadon  we aren't sorting; we're unwrapping the NonEmptySet so I can intersect it with the peer set.

modules/shared/src/test/scala/org/tessellation/security/signature/SignedValidatorSuite.scala
sdavidp: @marcinwadon I don't understand your concern. There is this test, "validateSignedByMajority should succeed when there is no seedlist" -- are you saying it should fail in that case?

modules/shared/src/main/scala/org/tessellation/schema/metagraphMessage.scala
marcinwadon: Ok. If so, I agree that the hash reference is not needed then. For clean organization, I'd definitely rename it to something else than `Transaction`. It's something you can send, but it's more an event rather than transaction. If you want to implement multiple type of transactions (we already have DAG and Currency transactions) it should be based on the same Transaction type.

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/MetagraphMessageRoutes.scala
marcinwadon: isn't it a data application part? or is it for all currencies (even if there is no data application provided)?
the naming here is important because the endpoint would be publicly exposed through the API

modules/shared/src/main/scala/org/tessellation/security/signature/SignedValidator.scala
marcinwadon: ok :+1:

modules/shared/src/test/scala/org/tessellation/security/signature/SignedValidatorSuite.scala
marcinwadon: > There is this test, "validateSignedByMajority should succeed when there is no seedlist"
sorry, I missed that one. it's fine then :+1:

modules/node-shared/src/test/scala/org/tessellation/node/shared/http/routes/CurrencyMessageRoutesSuite$.scala
marcinwadon, line 1: why the suffix `$`?

modules/node-shared/src/test/scala/org/tessellation/node/shared/http/routes/CurrencyMessageRoutesSuite$.scala
```diff
@@ -0,0 +1,94 @@
...
+  test("POST offers a valid CurrencyMessage") {
+    forall(currencyMessageGen) { message =>
+      val service = mockService(offerFn = valid(message))
+      makeCurrencyMessageRoutes(service).use { routes =>
+        expectHttpStatus(routes, route.withEntity(message.asJson))(NoContent)
+      }
+    }
+  }
+
+  test("POST offers a invalid CurrencyMessage") {
```
marcinwadon, line 55: an invalid

modules/node-shared/src/test/scala/org/tessellation/node/shared/infrastructure/snapshot/services/CurrencyMessageServiceSuite$.scala
marcinwadon, line 1: as above, why `$`?

modules/node-shared/src/test/scala/org/tessellation/node/shared/infrastructure/snapshot/storage/CurrencyMessageStorageSuite$.scala
marcinwadon, line 1: aa

modules/shared/src/main/scala/org/tessellation/schema/currencyMessage.scala
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  @derive(decoder, encoder, order, show)
+  @newtype
+  case class MessageOrdinal(value: NonNegLong) {
+    def next: MessageOrdinal = MessageOrdinal(value |+| 1L)
+  }
+
+  @derive(eqv, show, encoder, decoder)
+  case class CurrencyMessage(messageType: MessageType, address: Address, ordinal: MessageOrdinal)
```
marcinwadon, line 34: should we keep the next ordinal here, or reference to the previous? both should work, but for txs we store the reference 🤔

sdavidp: @ryle-goehausen What are your thoughts?

modules/node-shared/src/test/scala/org/tessellation/node/shared/infrastructure/snapshot/services/CurrencyMessageServiceSuite$.scala
sdavidp: 👉 IntelliJ slipped that past me when I wasn't looking. I will rename.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/modules/HttpApi.scala
sdavidp: renamed to use prefix "Currency", hopefully changed all references.

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/MetagraphMessageRoutes.scala
sdavidp: Changed to single endpoint, `/currency/message`

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/services/MetagraphMessageService.scala
sdavidp: changed to single `offer` method

modules/shared/src/main/scala/org/tessellation/schema/metagraphMessage.scala
sdavidp: Created `MessageOrdinal` to address this

modules/shared/src/main/scala/org/tessellation/security/signature/SignedValidator.scala
sdavidp: renamed

modules/wallet/src/main/scala/org/tessellation/wallet/Main.scala
sdavidp: Added `--output` option to write to a file instead of the console, if provided.

modules/node-shared/src/test/scala/org/tessellation/node/shared/http/routes/CurrencyMessageRoutesSuite$.scala
sdavidp: removed

modules/node-shared/src/test/scala/org/tessellation/node/shared/http/routes/CurrencyMessageRoutesSuite$.scala
sdavidp: fixed

modules/node-shared/src/test/scala/org/tessellation/node/shared/infrastructure/snapshot/storage/CurrencyMessageStorageSuite$.scala
sdavidp: removed

modules/shared/src/main/scala/org/tessellation/schema/currencyMessage.scala
ryle-goehausen: We might as well make this consistent with transaction and use reference to previous (ordinal-1), it doesn't particularly matter what the offset is so long as it's changing by 1 each time.

sdavidp: 1. So the message will contain an optional previous ordinal?
2. If given a previous ordinal then does the new one have to be `previous + 1`? (Seems redundant)
3. Does that mean the message storage only accepts a message if there is none present or if ... what currently stored ordinal matches given previous ordinal?
4. Or if we're to mimic the `create transaction` CLI command, it's taking a file to the previous transaction as input. Is that what we need to do? Receive the previous signed message and construct new one based on that?

modules/wallet/src/main/scala/org/tessellation/wallet/fileio.scala
```diff
@@ -0,0 +1,32 @@
...
+
+import cats.effect.Async
+
+import fs2.io.file.{Files, Flags, Path}
+import fs2.{Stream, text}
+import io.circe.fs2._
+import io.circe.syntax._
+import io.circe.{Decoder, Encoder}
+
+object fileio {
```
marcinwadon, line 11: IMO it should be `package org.tessellation.wallet.file` with `object io`

marcinwadon (approved):
LGTM but one comment

modules/shared/src/main/scala/org/tessellation/schema/currencyMessage.scala
marcinwadon: > So the message will contain an optional previous ordinal?

If you look at the transaction, there is no optional type. The previous ordinal should be always present.

> If given a previous ordinal then does the new one have to be previous + 1? (Seems redundant)

Yes, but you don't need to store the current one. You have the `Next` though (again - check transaction)

> Does that mean the message storage only accepts a message if there is none present or if ... what currently stored ordinal matches given previous ordinal?

Yes, that should be very important (validation).

> Or if we're to mimic the create transaction CLI command, it's taking a file to the previous transaction as input. Is that what we need to do? Receive the previous signed message and construct new one based on that?

It's actually up to you how you want the user to provide it. I see no issue with providing it as CLI parameters

modules/shared/src/main/scala/org/tessellation/schema/currencyMessage.scala
```diff
@@ -0,0 +1,35 @@
...
+    val values = findValues
+
+    case object Owner extends MessageType("Owner")
+    case object Staking extends MessageType("Staking")
+  }
+
+  @derive(decoder, encoder, order, show)
+  @newtype
+  case class MessageOrdinal(value: NonNegLong) {
+    def next: MessageOrdinal = MessageOrdinal(value |+| 1L)
```
marcinwadon, line 30: instead of `def next`, please provide the `Next` instance (see SnapshotOrdinal for reference). `TransactionOrdinal` looks to be the oldest code and it does not have this refactored yet

modules/shared/src/main/scala/org/tessellation/schema/currencyMessage.scala
sdavidp: @marcinwadon @ryle-goehausen 

So are you saying you want me to mirror `Transaction` and define a `CurrencyMessageReference` and so define `CurrencyMessage` to contain it, like this?
```
case class CurrencyMessage(messageType: MessageType, address: Address, parent: CurrencyMessageReference)
```
or do you want it to follow more closely with how a `Transaction` is defined? e.g.
```
case class CurrencyMessage(
  messageType: MessageType,
  address: Address,
  parent: CurrencyMessageReference,
  salt: CurrencyMessageSalt
) extends Fiber[CurrencyMessageReference, CurrencyMessageData]
  with Encodable[String] { . . .}
```

modules/wallet/src/main/scala/org/tessellation/wallet/fileio.scala
sdavidp: I can do that, however, I don't understand the point of shadowing existing root prefixes.

modules/shared/src/main/scala/org/tessellation/schema/currencyMessage.scala
ryle-goehausen: Hey, just to disambiguate, I think you can leave the structure the exact same as it is, and just allow the CLI to pass an integer to it.

```case class CurrencyMessage(messageType: MessageType, address: Address, parentOrdinal: MessageOrdinal)```

I parsed Marcin's comment as the question of should the ordinal be set at N-1 or N (either resulting in the same effect depending how we treat it in the code.) Not whether or not we need a common class or Reference class. So we could could just rename it as parentOrdinal and then define the current ordinal as being parentOrdinal.next() or the other way around. I think that's what he's getting at here.

I don't think we need to change the schema anymore, changing the field name along is fine and providing a current / next method is enough.

File: modules/wallet/src/main/scala/org/tessellation/wallet/fileio.scala (hunk 29 above)
Comment (sdavidp) [reply modules/wallet/src/main/scala/org/tessellation/wallet/fileio.scala]: done

File: modules/shared/src/main/scala/org/tessellation/schema/currencyMessage.scala (hunk 28 above)
Comment (sdavidp) [reply modules/shared/src/main/scala/org/tessellation/schema/currencyMessage.scala]: done

File: modules/shared/src/main/scala/org/tessellation/schema/currencyMessage.scala (hunk 30 above)
Comment (sdavidp) [reply modules/shared/src/main/scala/org/tessellation/schema/currencyMessage.scala]: done

File: modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/storage/CurrencyMessageStorage.scala [hunk 31]
```diff
@@ -0,0 +1,34 @@
...
+  def make[F[_]: Concurrent: Ref.Make]: F[CurrencyMessageStorage[F]] =
+    MapRef.ofSingleImmutableMap[F, MessageType, Signed[CurrencyMessage]](Map.empty).map { mapRef =>
+      new CurrencyMessageStorage[F] {
+        def get(messageType: MessageType): F[Option[Signed[CurrencyMessage]]] =
+          mapRef(messageType).get
+
+        def set(message: Signed[CurrencyMessage]): F[Boolean] =
+          mapRef(message.messageType).modify {
+            case None                                                      => (message.some, true)
+            case Some(m) if message.parentOrdinal === m.parentOrdinal.next => (message.some, true)
```
marcinwadon, line 28: it should be:
```
m.ordinal === message.parentOrdinal
```

where `m.ordinal` is defined as `def ordinal = parentOrdinal.next` using Next instance. That way it's easier to follow what we're actually checking

modules/shared/src/main/scala/org/tessellation/schema/currencyMessage.scala
```diff
@@ -0,0 +1,61 @@
...
+    def apply(value: Long): Either[String, MessageOrdinal] =
+      NonNegLong.from(value).map(MessageOrdinal(_))
+
+    implicit val next: Next[MessageOrdinal] = new Next[MessageOrdinal] {
+      def next(a: MessageOrdinal): MessageOrdinal = MessageOrdinal(a.value |+| NonNegLong(1L))
+      def partialOrder: PartialOrder[MessageOrdinal] = Order[MessageOrdinal]
+    }
+
+    val MinValue: MessageOrdinal = MessageOrdinal(NonNegLong.MinValue)
+    val MinIncrementalValue: MessageOrdinal = next.next(MinValue)
```
marcinwadon, line 45: Where do you use `MinIncrementalValue`? 🤔 I think it's copy-paste error

modules/wallet/src/main/scala/org/tessellation/wallet/fileio.scala
marcinwadon: just to be consistent with namings and packages, no need for reinventing new ones 🤔

modules/shared/src/main/scala/org/tessellation/schema/currencyMessage.scala
sdavidp: removed

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/storage/CurrencyMessageStorage.scala
sdavidp: done

marcinwadon (approved):

PR #815, opened 2024-01-24
Enrich snapshot value passed into onSnapshotConsensusResult

scasplte2:
## Summary
Metagraphs that utilize the `onSnapshotConsensusResult` hook may wish to interact with additional storage and third-party APIs using a reproducible hash value. By passing the `Hashed[CurrencyIncrementalSnapshot]` we retain the original functionality without any additional imports needed during the `consumeMajorityArtifact` evaluation.

While it is possible to import the required dependencies within the data application implementations and create the needed `KryoSerializer` to produce the `Hashed[_]` value locally, this is a common use-case where boilerplate infrastructure code may be simplified by incorporation into the underlying framework. 

## Changes
- modified function signature of `onSnapshotConsensusResult` to take an argument of `Hashed[CurrencyIncrementalSnapshot]` 

## Testing
** work in progress **

## Tickets
- none

### DIFF
--- a/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
+++ b/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
@@ -56,7 +56,9 @@ object CurrencySnapshotConsensusFunctions {
     ): F[Unit] =
       stateChannelSnapshotService.consume(signedArtifact, context) >>
         gossipForkInfo(gossip, signedArtifact) >>
-        maybeDataApplication.traverse_(_.onSnapshotConsensusResult(signedArtifact))
+        maybeDataApplication.traverse_ { da =>
+          signedArtifact.toHashed >>= da.onSnapshotConsensusResult
+        }
 
     def validateArtifact(
       lastSignedArtifact: Signed[CurrencySnapshotArtifact],
--- a/modules/shared/src/main/scala/org/tessellation/currency/dataApplication/package.scala
+++ b/modules/shared/src/main/scala/org/tessellation/currency/dataApplication/package.scala
@@ -91,7 +91,7 @@ trait BaseDataApplicationL0Service[F[_]] extends BaseDataApplicationService[F] w
 
   final def serializedOnChainGenesis: F[Array[Byte]] = serializeState(genesis.onChain)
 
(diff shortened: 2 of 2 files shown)


modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
```diff
@@ -56,7 +56,11 @@ object CurrencySnapshotConsensusFunctions {
     ): F[Unit] =
       stateChannelSnapshotService.consume(signedArtifact, context) >>
         gossipForkInfo(gossip, signedArtifact) >>
-        maybeDataApplication.traverse_(_.onSnapshotConsensusResult(signedArtifact))
+        maybeDataApplication.traverse_ { da =>
+          signedArtifact.toHashed.flatMap { hashedArtifact =>
+            da.onSnapshotConsensusResult(hashedArtifact)
+          }
+        }
```
sdavidp, line 63:
```suggestion
        maybeDataApplication.traverse_(da => signedArtifact.toHashed >>= da.onSnapshotConsensusResult)
```

sdavidp (commented):
Left suggested simplification otherwise LGTM

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
marcinwadon: or using `flatMap` it would be:
```scala
signedArtifact.toHashed.flatMap(da.onSnapshotConsensusResult)
```

up to you @scasplte2

marcinwadon (approved):

marcinwadon:
@scasplte2 please remember to add the correct label

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
scasplte2: thank you for the suggestions, I've updated to use the `>>=` since similar operators are used earlier in the function

marcinwadon (dismissed):

marcinwadon (approved):

PR #816, opened 2024-01-25
Adding new metagraph info endpoint

IPadawans:
### Changes
+ Adding new `metagraph/info` endpoint that can be accessed on currencyL0 and currencyL1 modules
+ Here is one print of the response
Captura de Tela 2024-01-25 às 12 37 54
*Ignore the version of tessellation as `adding-metagraph-info`, I've hardcoded this to test on Euclid*

### JIRA
[PROT-567](https://constellationnetwork.atlassian.net/browse/PROT-567)

### DIFF
--- a/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
+++ b/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
@@ -3,7 +3,6 @@ package org.tessellation.currency.l0
 import cats.effect.{IO, Resource}
 import cats.syntax.all._
 
-import org.tessellation.BuildInfo
 import org.tessellation.currency.dataApplication.{BaseDataApplicationL0Service, L0NodeContext}
 import org.tessellation.currency.l0.cli.method
 import org.tessellation.currency.l0.cli.method._
@@ -30,12 +29,13 @@ abstract class CurrencyL0App(
   name: String,
   header: String,
   clusterId: ClusterId,
-  version: String
+  tessellationVersion: String,
+  metagraphVersion: String
 ) extends TessellationIOApp[Run](
       name,
       header,
       clusterId,
-      version = version
+      version = tessellationVersion
     ) {
 
   val opts: Opts[Run] = method.opts
@@ -128,9 +128,10 @@ abstract class CurrencyL0App(
           keyPair.getPrivate,
           cfg.environment,
           nodeShared.nodeId,
-          BuildInfo.version,
+          tessellationVersion,
           cfg.http,
-          services.dataApplication
+          services.dataApplication,
+          metagraphVersion.some
         )
       _ <- MkHttpServer[IO].newEmber(ServerName("public"), cfg.http.publicHttp, api.publicApp)
       _ <- MkHttpServer[IO].newEmber(ServerName("p2p"), cfg.http.p2pHttp, api.p2pApp)
--- a/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
+++ b/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
@@ -21,6 +21,7 @@ import org.tessellation.node.shared.snapshot.currency.CurrencySnapshotEvent
 import org.tessellation.node.shared.{NodeSharedOrSharedRegistrationIdRange, nodeSharedKryoRegistrar}
 import org.tessellation.schema.cluster.ClusterId
 import org.tessellation.schema.node.NodeState
+import org.tessellation.schema.semver.{MetagraphVersion, TessellationVersion}
 import org.tessellation.security.SecurityProvider
 
 import com.monovore.decline.Opts
@@ -29,8 +30,8 @@ abstract class CurrencyL0App(
   name: String,
   header: String,
   clusterId: ClusterId,
-  tessellationVersion: String,
-  metagraphVersion: String
+  tessellationVersion: TessellationVersion,
+  metagraphVersion: MetagraphVersion
 ) extends TessellationIOApp[Run](
       name,
       header,
--- a/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/Main.scala
+++ b/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/Main.scala
@@ -10,5 +10,6 @@ object Main
       "Currency-l0",
       "Currency L0 node",
       ClusterId(UUID.fromString("517c3a05-9219-471b-a54c-21b7d72f4ae5")),
-      version = BuildInfo.version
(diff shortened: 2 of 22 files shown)


modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/MetagraphRoutes.scala
```diff
@@ -0,0 +1,50 @@
...
+          session,
+          clusterSession,
+          httpCfg.externalIp,
+          httpCfg.publicHttp.port,
+          httpCfg.p2pHttp.port,
+          selfId,
+          tessellationVersion,
+          metagraphVersion
+        )
+      }.flatMap(Ok(_))
```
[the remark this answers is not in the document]
IPadawans: Yes, I've tried to replicate the behavior of `node/info` so we should be fine

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/Main.scala
```diff
@@ -10,5 +10,6 @@ object Main
       "Currency-l0",
       "Currency L0 node",
       ClusterId(UUID.fromString("517c3a05-9219-471b-a54c-21b7d72f4ae5")),
-      version = BuildInfo.version
-    ) {}
+      tessellationVersion = BuildInfo.version,
+      metagraphVersion = "0"
```
marcinwadon, line 14: should it be `0` or just empty?

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/MetagraphRoutes.scala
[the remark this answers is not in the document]
marcinwadon: In general, non handled error would be returned as `InternalServerError`. sometimes we want to log the error, so we use `handleErrorWith` and log the error message or sth. In this case I think it's fine as it is

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/Main.scala
IPadawans: Actually can be empty ""

marcinwadon: if it's going to be empty, maybe it's a case for the optional type? `Option[Version]`?

marcinwadon:
Please remember to squash everything locally before pushing final version and merging please

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/Main.scala
IPadawans: It can be empty here on tessellation, but I want to ensure that everyone who implements those classes provides a version.

kpudlik (changes requested):

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
```diff
@@ -30,12 +29,13 @@ abstract class CurrencyL0App(
   name: String,
   header: String,
   clusterId: ClusterId,
-  version: String
+  tessellationVersion: String,
+  metagraphVersion: String
```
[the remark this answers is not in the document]
IPadawans: @kpudlik I've tried, but maybe I've understood wrongly
Captura de Tela 2024-02-06 às 13 52 13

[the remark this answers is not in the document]
kpudlik: `BuildInfo.version` is not a literal and is generated by the compiler so the only way to implement that is to use `unsafe`:

```scala
@derive(encoder, decoder, show, order)
@newtype
case class MetagraphVersion(version: String Refined Semver)

object MetagraphVersion {
  def unsafeFromBuildInfo = unsafeFrom(BuildInfo.version))
  def unsafeFrom(str: String) = MetagraphVersion(refineV[SemVer].unsafeFrom(str))
}
```

and then
```
metagraphVerison = MetagraphVersion.unsafeFromBuildInfo
```

I think it is fine to use unsafe here as it happens before even starting the node.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
```diff
@@ -23,19 +21,20 @@ import org.tessellation.node.shared.{NodeSharedOrSharedRegistrationIdRange, node
...
-
 import com.monovore.decline.Opts
+import org.tessellation.schema.semver.MetagraphVersion
 
 abstract class CurrencyL0App(
   name: String,
   header: String,
   clusterId: ClusterId,
-  version: String
+  tessellationVersion: String,
```
kpudlik, line 31: It should also be `TessellationVersion` accordingly

kpudlik (changes requested):

modules/shared/src/main/scala/org/tessellation/schema/node.scala
```diff
@@ -103,4 +103,16 @@ object node {
...
+  @derive(encoder)
+  case class MetagraphNodeInfo(
+    state: NodeState,
+    session: Option[SessionToken],
+    clusterSession: Option[ClusterSessionToken],
+    host: Host,
+    publicPort: Port,
+    p2pPort: Port,
+    id: PeerId,
+    tessellationVersion: String,
```
kpudlik, line 115: Use TessellationVersion type

modules/shared/src/main/scala/org/tessellation/schema/node.scala
```diff
@@ -103,4 +103,16 @@ object node {
...
+    metagraphVersion: String
```
kpudlik, line 116: Use MetagraphVersion type

kpudlik (changes requested):

modules/shared/src/main/scala/org/tessellation/schema/node.scala
IPadawans: Done

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/CurrencyL1App.scala
```diff
@@ -1,13 +1,13 @@
 package org.tessellation.currency.l1
 
 import cats.effect.{IO, Resource}
+import cats.implicits.catsSyntaxOptionId
```
marcinwadon, line 4: please don't mix `cats.implicits` with `cats.syntax`. you can import `cats.syntax.option._` or just remove all and import `cats.syntax.all._`

marcinwadon (commented):
looks correct for me :+1:

marcinwadon (approved):

PR #822, opened 2024-02-02
Hash calculation using json serializer

marcinwadon:


modules/currency-l0/src/main/scala/org/tessellation/currency/l0/modules/Programs.scala
```diff
@@ -22,7 +23,7 @@ import org.tessellation.security.{Hasher, SecurityProvider}
 
 object Programs {
 
-  def make[F[_]: Async: Random: KryoSerializer: Hasher: SecurityProvider](
+  def make[F[_]: Async: Random: KryoSerializer: JsonHashSerializer: Hasher: SecurityProvider](
```
marcinwadon, line 26: I think a better name is needed, as it's not only for Hash now

scasplte2: In my mind, "serialization" implies binary as one side of the Encode/Decode functionality. Perhaps `JsonSerializer` is enough to imply the intended behavior.

kpudlik (dismissed):

modules/shared/src/main/scala/org/tessellation/security/signature/Signed.scala
```diff
@@ -106,21 +106,31 @@ object Signed {
...
+            },
+          implicit encoder =>
+            Hasher[F]
+              .hashJson(signed.value)
+              .flatMap {
+                signature.verifySignatureProof(_, proof).flatMap {
+                  if (_) true.pure[F]
+                  else
+                    Hasher[F].hashKryo(signed.value) >>= { signature.verifySignatureProof(_, proof) }
+                }
```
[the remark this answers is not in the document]
marcinwadon: done :+1:

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/cli/method.scala
```diff
@@ -172,8 +179,44 @@ object method {
...
+              snapshot,
+              seedlistPath,
+              prioritySeedlistPath,
+              collateralAmount,
+              globalL0Peer,
+              l0TokenIdentifier,
+              trustRatingsPath,
+              lastKryoHash
+            ) =>
+          val lastKH =
```
kpudlik, line 200: I see it repeated many times. How about doing the following?

```scala
  val lastKryoHashOrdinalOpts: Opts[Option[SnapshotOrdinal]] =
    (
      AppEnvironment.opts,
      Opts
        .argument[NonNegLong]("lastKryoHashOrdinal")
        .map(SnapshotOrdinal(_))
        .orNone
    ).mapN {
      case (env, lastKryoHash) =>
        if (env === AppEnvironment.Dev) lastKryoHash else lastKryoHashOrdinal.get(env).orElse(SnapshotOrdinal.MinValue.some)
    }
```

kpudlik (changes requested):

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/cli/method.scala
marcinwadon: @kpudlik I've already fixed that once working on the configuration, so it makes no sense to fix it here again 🤔 I would have to rebase the configuration with these changes, so I can take a second look then once combining both changes. What do you think?

kpudlik (approved):

PR #823, opened 2024-02-08
Resend multiple state channel binaries and increase retry ordinal delay.

kpudlik:


modules/currency-l0/src/test/scala/org/tessellation/currency/l0/snapshot/services/SentStateChannelBinaryTrackingServiceSuite.scala
```diff
@@ -85,6 +87,7 @@ object SentStateChannelBinaryTrackingServiceSuite extends MutableIOSuite with Ch
           _ <- service.updateByGlobalSnapshot(snapshot)
           pending <- pendingR.get
           toRetry <- service.getRetriable
+          _ = println(s"pending: ${pending.size}, toRetry: ${toRetry.size}, shuffled: ${shuffled.size}")
```
marcinwadon, line 90: needed or left from debugging?

marcinwadon (dismissed):

modules/currency-l0/src/test/scala/org/tessellation/currency/l0/snapshot/services/SentStateChannelBinaryTrackingServiceSuite.scala
kpudlik: fixed

marcinwadon (approved):

PR #824, opened 2024-02-13
Customized currency L0 consensus

TheMMaciek:


modules/currency-l0/src/main/scala/org/tessellation/currency/l0/modules/Storages.scala
```diff
@@ -53,7 +52,6 @@ object Storages {
         )
       lastGlobalSnapshotStorage <- LastSnapshotStorage.make[F, GlobalIncrementalSnapshot, GlobalSnapshotInfo]
       globalL0ClusterStorage <- L0ClusterStorage.make[F](globalL0Peer)
-      lastBinaryHashStorage <- LastBinaryHashStorage.make[F]
```
TheMMaciek, line 56: This storage isn't needed anymore. Last binary hash gets passed around as part of consensus outcome.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencyConsensusHandler.scala
```diff
@@ -0,0 +1,41 @@
...
+      CurrencySnapshotStatus,
+      CurrencyConsensusOutcome,
+      CurrencyConsensusKind
+    ](storage, manager, fns)
+
+    all.eventHandler <+>
+      all.facilityHandler <+>
+      all.proposalHandler <+>
+      all.signatureHandler <+>
+      all.binarySignatureHandler <+>
```
TheMMaciek, line 36: The difference between currency and global is using `binarySignatureHandler` for currency and not for global.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
```diff
@@ -36,30 +31,17 @@ abstract class CurrencySnapshotConsensusFunctions[F[_]: Async: SecurityProvider:
...
-    gossip: Gossip[F],
-    maybeDataApplication: Option[BaseDataApplicationL0Service[F]]
+    currencySnapshotValidator: CurrencySnapshotValidator[F]
   ): CurrencySnapshotConsensusFunctions[F] = new CurrencySnapshotConsensusFunctions[F] {
 
     def getRequiredCollateral: Amount = collateral
 
     def getBalances(context: CurrencySnapshotContext): SortedMap[Address, Balance] = context.snapshotInfo.balances
 
-    def consumeSignedMajorityArtifact(
```
TheMMaciek, line 53: Not needed on the interface anymore. Consensus steps are customizable right now, so the consume logic can be implemented (and is in this PR) as the effect on the last step of consensus.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/programs/Genesis.scala
```diff
@@ -79,18 +78,26 @@ object Genesis {
...
-        signedFirstIncrementalSnapshot,
-        CurrencySnapshotContext(identifier, hashedGenesis.info)
+        CurrencyConsensusOutcome(
+          signedFirstIncrementalSnapshot.ordinal,
+          Facilitators(List(nodeId)),
+          RemovedFacilitators.empty,
+          WithdrawnFacilitators.empty,
+          Finished(
+            signedFirstIncrementalSnapshot,
+            incrementalBinaryHash,
```
TheMMaciek, line 94: That's where the lastBinaryHash resides right now - hance no need for the storage.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/consensus/Consensus.scala
```diff
@@ -1,92 +1,14 @@
...
-    keyPair: KeyPair,
-    consensusConfig: ConsensusConfig,
-    seedlist: Option[Set[SeedlistEntry]],
-    clusterStorage: ClusterStorage[F],
-    nodeStorage: NodeStorage[F],
-    client: Client[F],
-    session: Session[F]
-  ): F[Consensus[F, Event, Key, Artifact, Context]] =
-    for {
-      storage <- ConsensusStorage.make[F, Event, Key, Artifact, Context](consensusConfig)
```
TheMMaciek, line 51: All this setup happens now in `CurrencySnapshotConsensus` and `GlobalSnapshotConsensus` classes. Additional layer of abstraction wasn't needed in my opinion and I also needed to pull some of the initializations out from here so in the end this `make` stopped being useful.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/consensus/Consensus.scala
```diff
@@ -1,92 +1,14 @@
...
-import org.tessellation.node.shared.infrastructure.metrics.Metrics
-import org.tessellation.schema.peer.PeerId
-import org.tessellation.security.{Hasher, SecurityProvider}
-
-import io.circe.{Decoder, Encoder}
-import org.http4s.client.Client
-
-object Consensus {
-
-  def make[
```
TheMMaciek, line 32: All this setup happens now in CurrencySnapshotConsensus and GlobalSnapshotConsensus classes. Additional layer of abstraction wasn't needed in my opinion and I also needed to pull some of the initializations out from here so in the end this `make` stopped being useful.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -18,44 +18,51 @@ import org.tessellation.node.shared.infrastructure.consensus.trigger.{ConsensusT
...
-    consensusClient: ConsensusClient[F, Key, Artifact, Context],
-    selfId: PeerId
-  )(implicit S: Supervisor[F]): F[ConsensusManager[F, Key, Artifact, Context]] = {
+    consensusClient: ConsensusClient[F, Key, Outcome]
+  )(
+    implicit S: Supervisor[F],
+    _artifact: Lens[Outcome, Signed[Artifact]],
+    _context: Lens[Outcome, Context],
+    _key: Lens[Outcome, Key],
+    _trigger: Lens[Outcome, ConsensusTrigger]
```
TheMMaciek, line 64: Lenses used to let the developer use an arbitrary `Outcome` type but still let the developer reuse functionality that relies on the particular fields/data like `Key` or `ConsensusTrigger`. I think it's better that way than a common trait.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/consensus/ConsensusRumorHandlers.scala
```diff
@@ -0,0 +1,75 @@
...
+
+class ConsensusRumorHandlers[F[
+  _
+]: Async: SecurityProvider, Event: TypeTag: Decoder, Key: Show: TypeTag: Decoder, Artifact: TypeTag: Decoder, Context: Decoder, Status, Outcome, Kind: Decoder: TypeTag](
+  storage: ConsensusStorage[F, Event, Key, Artifact, Context, Status, Outcome, Kind],
+  manager: ConsensusManager[F, Key, Artifact, Context, Status, Outcome, Kind],
+  fns: ConsensusFunctions[F, Event, Key, Artifact, Context]
+) {
+
+  val eventHandler = RumorHandler.fromPeerRumorConsumer[F, ConsensusEvent[Event]]() { rumor =>
```
TheMMaciek, line 25: Handlers should work exactly the same as they did before. One addition is a `binarySignatureHandler` that's being used for currency but not for global. This is kind of a collection of available handlers that can than be hand picked by the developer (given that the `ConsensusResources` still remain as a common resources class for both consensuses)

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/consensus/ConsensusState.scala
```diff
@@ -1,111 +1,81 @@
...
-import org.tessellation.security.signature.Signed
 
 import derevo.cats.{eqv, show}
 import derevo.circe.magnolia.{decoder, encoder}
 import derevo.derive
+import monocle.Lens
+import monocle.macros.GenLens
+
+@derive(eqv, encoder, decoder)
+case class Facilitators(value: List[PeerId])
```
TheMMaciek, line 19: Refined classes that are meant to represent certain concepts in the codebase. Hence if a function needs `Facilitators` and `WithdrawnFacilitators` which are supposed to mean the same thing across the codebase then only these field can be extracted (with e.g. a lens) from the state rather than the whole state classes would need to be passed. In the future I would also customize the `ConsensusState` class and `ConsensusResources` class and separate currency and global (and other future consensuses) even more. This is a step in this direction.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/consensus/ConsensusState.scala
```diff
@@ -1,111 +1,81 @@
...
-       |withdrawnFacilitators=${cs.withdrawnFacilitators.show},
+       |facilitatorCount=${cs.facilitators.value.size.show},
+       |removedFacilitators=${cs.removedFacilitators.value.show},
+       |withdrawnFacilitators=${cs.withdrawnFacilitators.value.show},
        |spreadAckKinds=${cs.spreadAckKinds.show},
        |status=${cs.status.show}
        |}""".stripMargin.replace(",\n", ", ")
   }
 
-  implicit class ConsensusStateOps[K, A, C](value: ConsensusState[K, A, C]) {
```
TheMMaciek, line 47: The declaration kind and consensus status are now customized by every consensus, so these mappers needed to be implemented elsewhere.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -65,359 +60,212 @@ object ConsensusStateUpdater {
...
   def make[F[
     _
-  ]: Async: KryoSerializer: Hasher: SecurityProvider: Metrics, Event, Key: Show: Order: TypeTag: Encoder, Artifact <: AnyRef: Eq: TypeTag: Encoder, Context <: AnyRef: Eq: TypeTag](
-    consensusFns: ConsensusFunctions[F, Event, Key, Artifact, Context],
-    consensusStorage: ConsensusStorage[F, Event, Key, Artifact, Context],
+  ]: Async: KryoSerializer: Hasher: SecurityProvider: Metrics, Event, Key: Show: Order: TypeTag: Encoder, Artifact <: AnyRef: Eq: TypeTag: Encoder, Context <: AnyRef: Eq: TypeTag, Status: Eq: Show, Outcome: Eq, Kind: Encoder: Eq: Show: TypeTag](
+    consensusStateAdvancer: ConsensusStateAdvancer[F, Key, Artifact, Context, Status, Outcome, Kind],
+    consensusStorage: ConsensusStorage[F, Event, Key, Artifact, Context, Status, Outcome, Kind],
     gossip: Gossip[F],
-    keyPair: KeyPair
```
TheMMaciek, line 72: It isn't clear here but what happened here is mostly pulling out the status handling logic to a separate class (so that we get custom consensus flow). A lot of functionalities are marked as red and then with green (moved rather than changed)

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -18,44 +18,51 @@ import org.tessellation.node.shared.infrastructure.consensus.trigger.{ConsensusT
...
 import retry.syntax.all._
 
-trait ConsensusManager[F[_], Key, Artifact, Context] {
+trait ConsensusManager[F[_], Key, Artifact, Context, Status, Outcome, Kind] {
   def registerForConsensus(observationKey: Key): F[Unit]
 
   def startFacilitatingAfterDownload(observationKey: Key, lastArtifact: Signed[Artifact], lastContext: Context): F[Unit]
 
-  def startFacilitatingAfterRollback(lastKey: Key, lastArtifact: Signed[Artifact], lastContext: Context): F[Unit]
+  def startFacilitatingAfterRollback(lastKey: Key, initialOutcome: Outcome): F[Unit]
```
marcinwadon, line 37: shouldn't be either `initialKey / initialOutcome` or `lastKey / lastOutcome`?
or just `key, outcome`?

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/consensus/ConsensusManager.scala
marcinwadon: nice one! :clap:

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/consensus/ConsensusManager.scala
TheMMaciek: I think lastKey/lastOutcome makes sense - will just verify if that naming reflects the usage of these values.

TheMMaciek: I would actually leave it as it is because these names `lastKey` and `initialOutcome` do convey well the meaning of these values:
- `lastKey` is the ordinal of the snapshot we performed rollback to (or key of first incremental during genesis creation), so the name works well
- `initialOutcome` is the constructed outcome that simulates/mocks the outcome of the presumed previous consensus (for genesis there is no previous outcome, but we still construct it). We can't really call it a previous outcome because the actual previous outcome either didn't exist or may have been different - the outcome is not persisted between restarts. So the name `initialOutcome` also works well.

marcinwadon: you said the outcome is of the "presumed previous consensus". so it's also the `last` one 🤔

marcinwadon: you said the outcome is of the "presumed previous consensus". so it's also the `last` one 🤔

marcinwadon: you said the outcome is of the "presumed previous consensus". so it's also the `last` one 🤔

TheMMaciek: Although it's probably not the same as actual last outcome. The facilitators field is set to the initial (rollback, genesis) facilitator some of the fields are set to default values whereas they could actually be populated in the actual previous outcome. It's different for the key because snapshot ordinal is a singular thing e.i. ordinal of 100 is just that ordinal of 100 (so we can extract it from the snapshot), whereas outcome is a composite thing consisting of multiple things and because we don't persist the outcome between the restart it needs to be reconstructed here. Hence I understand why it was called `initial` rather than `last`- not my change by the way :). I can change it if you think it still makes more sense to call it `last`.

marcinwadon (approved):
rebase is needed @TheMMaciek

marcinwadon (approved):

kpudlik (approved):

PR #825, opened 2024-02-13
Transaction limiting and balance validation on endpoint

kpudlik:


marcinwadon:
@kpudlik needs rebase

marcinwadon (dismissed):

marcinwadon (dismissed):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionService.scala
```diff
@@ -3,37 +3,57 @@ package org.tessellation.dag.l1.domain.transaction
...
     def offer(
       transaction: Hashed[Transaction]
     ): F[Either[NonEmptyList[ContextualTransactionValidationError], Hash]] =
-      contextualTransactionValidator.validate(transaction.signed).flatMap {
-        case Valid(_) =>
-          transactionStorage
-            .put(transaction)
-            .as(transaction.hash.asRight[NonEmptyList[ContextualTransactionValidationError]])
-        case Invalid(e) => e.toNonEmptyList.asLeft[Hash].pure[F]
+      lastSnapshotStorage.getOrdinalStream.zipWithPrevious.mapFilter {
```
TheMMaciek, line 39: I think there is a nice operator for this task: `changes` that takes the `Eq` for compared element and only emits new elements.

marcinwadon: :+1:

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/modules/Validators.scala
```diff
@@ -21,17 +29,36 @@ object Validators {
...
     val transactionValidator = TransactionValidator.make[F](signedValidator)
     val blockValidator =
       BlockValidator.make[F](signedValidator, transactionChainValidator, transactionValidator)
+
+    val context = new TransactionValidationContext[F] {
+      def getBalance(address: Address): F[Balance] = storages.lastSnapshot.getLatestBalances.map {
+        _.flatMap(_.get(address)).getOrElse(Balance.empty)
+      }
+
+      def getLastSnapshotOrdinal: F[SnapshotOrdinal] =
```
TheMMaciek, line 46: A leftover I guess

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionStorage.scala
```diff
@@ -1,92 +1,124 @@
...
+    transactionsR(source).get.map {
+      _.flatMap(getLastProcessedTransaction).getOrElse(getInitialTx)
     }
 
-  def setLastAccepted(lastTxRefs: Map[Address, TransactionReference]): F[Unit] =
-    lastAccepted.clear >>
-      lastTxRefs.toList.traverse {
-        case (address, reference) => lastAccepted(address).set(Majority(reference).some)
-      }.void
+  def getLastProcessedTransaction(maybeStored: Option[SortedMap[TransactionOrdinal, StoredTransaction]]): StoredTransaction =
```
TheMMaciek, line 48: Not sure if this and the following function should be a part of the internal storage implementation. Especially that you fetch the transactions and then take them back to this function inside of the storage just to extract the correct one. I would rather see it as a pure function in the companion object. Or just extract the last tx based on the `StoredTransaction` type wherever it's needed - you are exposing the `StoredTransaction` either way.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/ContextualTransactionValidator.scala
```diff
@@ -0,0 +1,282 @@
...
+              toProcess
+            case Right(value) =>
+              loop(
+                memPool.tail,
+                value,
+                next.next(currentOrdinal),
+                currentOrdinal,
+                updatedTotalConsumedCap,
+                toProcess ++ memPool.headOption.toList
+              )
```
sdavidp, line 207: - If we're concerned about throwing if list is empty (`memPool.headOption`) then aren't we concerned that `memPool.tail` will throw on empty? maybe empty `memPool` should be a loop-exit condition?
- If we're keeping this code, instead of constructing list from `headOption` maybe `memPool.headOption.fold(...)`  and using `:+`?
- Is `toProcess` going to be a long list? If so then maybe appending will be unnecessarily slow.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/ContextualTransactionValidator.scala
```diff
@@ -0,0 +1,282 @@
...
+        txs: SortedMap[TransactionOrdinal, StoredTransaction]
+      ): SnapshotOrdinal =
+        txs.collectFirst { case (_, MajorityTx(_, snapshotOrdinal)) => snapshotOrdinal }.getOrElse(SnapshotOrdinal.MinValue)
+
+      private def getTransactionsAboveMajority(
+        txs: SortedMap[TransactionOrdinal, StoredTransaction]
+      ): SortedMap[TransactionOrdinal, StoredHashedTransaction] =
+        txs.collect {
+          case (ordinal, stored @ (_: WaitingTx | _: ProcessingTx | _: AcceptedTx)) =>
+            (ordinal, stored.asInstanceOf[StoredHashedTransaction])
```
sdavidp, line 261: Is there a way to do this without using `asInstanceOf` ? If we are certain that this will never blow up then doesn't that mean the input type is going to be `StoredHashedTransaction` ?

sdavidp: Is this an option?
```
case (ordinal, stored:StoredHashedTransaction @ (_: WaitingTx | _: ProcessingTx | _: AcceptedTx)) =>
  (ordinal, stored)
```

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/ContextualTransactionValidator.scala
```diff
@@ -0,0 +1,282 @@
...
+          memPool: List[Hashed[Transaction]],
+          balance: Balance,
+          lastTxRefOrdinal: SnapshotOrdinal,
+          currentOrdinal: SnapshotOrdinal,
+          totalConsumedCap: Double,
+          toProcess: List[Hashed[Transaction]]
+        ): List[Hashed[Transaction]] = {
+          val distancePassed = currentOrdinal.value - lastTxRefOrdinal.value
+          val transactionsAllowedWithinInterval = balance.value.toDouble / baseBalance.value.toDouble
+          val distanceBetweenTxsNeeded = ordinalDistanceForBaseBalance.toDouble / transactionsAllowedWithinInterval
```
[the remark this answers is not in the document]
kpudlik: For `transactionsAllowedWithinInterval` you are right. It would be nice to just fallback to `0` so no transactions are allowed. 
For the `distanceBetweenTxsNeeded` I think it is fine to leave it because `ordinalDistanceForBaseBalance.toDouble / 0.0` gives `Infinity` which is kinda valid in the context of that functionality:

If I have `Balance(0)` then I need to wait `Infinity` ordinals to make a transaction (which kinda makes sense because balance 0 will never make a transaction). Then `consumedCap` (cap is the remaining space to "fit" the transaction) is then also `Infinity` so I used the whole available cap immediately.

On the other hand keep in mind that `baseBalance` is something we set in config and setting it to 0 doesn't make sense in the same way as setting `timeTriggerInterval` for snapshots to `0` doesn't make sense because it will trigger consensus every `0` seconds 🤷🏻‍♂️ . Maybe it is worth to introduce the type `NonNegBalance` though or get rid of incorrect states at the config level in general.

Anyway, good catch. I will use one of solutions described above.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/ContextualTransactionValidator.scala
kpudlik: > Is this an option?
> 
> ```
> case (ordinal, stored:StoredHashedTransaction @ (_: WaitingTx | _: ProcessingTx | _: AcceptedTx)) =>
>   (ordinal, stored)
> ```

It doesn't compile. Compilator is not smart enough to figure out that ann the `WaitingTx`, `ProcessingTx`, and `AcceptedTx` are all `StoredTransaction` but also `StoredHashedTransaction` (mixin for `StoredTransaction`)

Actually, that function should return all the transactions above the majority. After thinking twice I think that we can make an assumption that `MajorityTx` can exist only once in storage and it is always the `head` of storage. It is guaranteed by the logic which sets `MajorityTx` - it removes everything which is below `MajorityTx` and the new `MajorityTx` should have a higher ordinal than the previous one.

Then we can just take everything above literally:

```scala
txs.takeRight(txs.size -1)
```

or just

```scala
txs.filterNot { case (_, MajorityTx(_, _)) => true }
```

but it still is not smart enough to know that I narrowed the type by filtering out.

I think that the solution would be to:

```
txs.collect { case tx @ (_, _: StoredHashedTransaction) => tx }
```

kpudlik: >that mean the input type is going to be StoredHashedTransaction?

I didn't want to make that assumption and I decided to be explicit because if at some point we decide to make `MajorityTx` to be `StoredHashedTransaction` as well then that logic will to be incorrect. Actually I think I will rename `StoredHashedTransaction` to just `NonMajorityTx` so name expresses the purpose/state instead of expressing the content.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/ContextualTransactionValidator.scala
kpudlik: >maybe empty memPool should be a loop-exit condition?

It is:
```
case Right(_) if updatedTotalConsumedCap > 1 || memPool.isEmpty =>
  toProcess
```
So I think we can safely use `memPool.head` there as `memPool.isEmpty` quits the loop.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
```diff
@@ -55,10 +55,10 @@ object BlockService {
           addressStorage.getBalance(address).map(_.some)
 
         def getLastTxRef(address: Address): F[Option[transaction.TransactionReference]] =
-          transactionStorage.getLastAcceptedReference(address).map(_.some)
+          transactionStorage.getLastProcessedTransaction(address).map(_.ref.some)
```
TheMMaciek, line 58: The name of this function seems wrong. The function returns last accepted transaction (either Majority or Accepted).

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionService.scala
```diff
@@ -3,37 +3,65 @@ package org.tessellation.dag.l1.domain.transaction
...
-            .as(transaction.hash.asRight[NonEmptyList[ContextualTransactionValidationError]])
-        case Invalid(e) => e.toNonEmptyList.asLeft[Hash].pure[F]
+      lastSnapshotStorage.getCombinedStream.map {
+        case Some((s, si)) => (s.ordinal, si.balances.getOrElse(transaction.source, Balance.empty))
+        case None          => (SnapshotOrdinal.MinValue, Balance.empty)
+      }.changes.switchMap {
+        case (latestOrdinal, balance) =>
+          Stream.eval {
+            transactionValidator
+              .validate(transaction.signed)
```
TheMMaciek, line 51: I guess this validate can happen only once given that it's a static validation. Not sure if worth refactoring

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionStorage.scala
```diff
@@ -1,191 +1,209 @@
...
-            case Some(Accepted(ref)) if ref == txRef =>
-              (Majority(txRef).some, ().asRight)
-            case other =>
-              (other, UnexpectedStateWhenMarkingTxRefAsMajority(source, txRef, other).asLeft)
-          }
-          .flatMap(_.liftTo[F])
-    }.void
-
-  def put(transaction: Hashed[Transaction]): F[Unit] = put(Set(transaction))
+  def put(
```
TheMMaciek, line 130: I think right now that should be called `tryPut`

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/storage/LastSnapshotStorage.scala
```diff
@@ -56,6 +56,11 @@ object LastSnapshotStorage {
 
       def getCombined: F[Option[(Hashed[S], SI)]] = snapshotR.get
 
+      def getCombinedStream: Stream[F, Option[(Hashed[S], SI)]] =
+        Stream
+          .eval[F, Option[(Hashed[S], SI)]](snapshotR.get)
+          .merge(snapshotR.discrete)
```
TheMMaciek, line 62: Can you explain this construct?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionService.scala
```diff
@@ -3,37 +3,65 @@ package org.tessellation.dag.l1.domain.transaction
...
+                case Valid(_) =>
+                  transactionStorage.put(transaction, latestOrdinal, balance)
+                case Invalid(e) =>
+                  e.toNonEmptyList.asLeft[Hash].leftWiden[NonEmptyList[ContextualTransactionValidationError]].pure[F]
+              }
+          }
+      }.head.compile.last.flatMap {
+        case Some(value) => value.pure[F]
+        case None =>
+          (new Exception(s"Unexpected state, stream should always emit the first snapshot"))
```
TheMMaciek, line 63: I think this can happen right now, we don't check whether the node is ready on the endpoint (if the download finished). So it's kind of unexpected but not impossible under normal conditions.

TheMMaciek: Also not sure if node turns to Ready only after download is done or if that can be earlier.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/block/BlockService.scala
kpudlik: Yes.
Transaction has the following lifecycle:

Waiting -> Processing -> Accepted -> Majority

so Processed means either Accepted or Majority or anything which comes after Processing

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionService.scala
kpudlik: Fair point, no need to rerun static validation if stream emits.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/storage/LastSnapshotStorage.scala
kpudlik: @TheMMaciek @marcinwadon according to: https://www.javadoc.io/doc/co.fs2/fs2-docs_3/3.4.0/fs2/concurrent/Signal.html#discrete:Stream[F,A]

We probably could just
```scala
snapshotR.continuous merge snapshotR.discrete
```

image

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/TransactionService.scala
kpudlik: Yeah. We should throw that eventually if node finished the Download but snapshot is not present.
But what if transaction comes if there is no download finished? All the validations which are based on ordinals may provide false negatives if we make a fallback to `SnapshotOrdinal.MinValue` and `Balance.empty`.

On the other hand better false negatives than false positives.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/ContextualTransactionValidator.scala
```diff
@@ -0,0 +1,265 @@
...
+          lastAcceptedWithinCap: Option[Hashed[Transaction]]
+        ): Option[Hashed[Transaction]] =
+          remainingTransactions match {
+            case head :: tail =>
+              val consumedCap =
+                if (head.fee.value >= transactionLimitConfig.minFeeWithoutLimit.value) 0
+                else {
+                  val distancePassed = currentOrdinal.value - lastTxRefOrdinal.value
+                  val transactionsAllowedWithinInterval =
+                    if (balance.value > 0) balance.value.toDouble / transactionLimitConfig.baseBalance.value.toDouble else 0.0
```
TheMMaciek, line 169: `baseBalance` is a `NonNegLong`, shouldn't we for that reason check if it's greater than 0 as well before attempting division?

TheMMaciek: I guess dividing by 0 doesn't throw - if that's the case then this works fine.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/transaction/ContextualTransactionValidator.scala
```diff
@@ -0,0 +1,265 @@
...
+                  val distanceBetweenTxsNeeded = ordinalDistanceForBaseBalance.toDouble / transactionsAllowedWithinInterval
+                  distanceBetweenTxsNeeded / Math.max(distancePassed, 1)
+                }
+
+              val updatedTotalConsumedCap = totalConsumedCap + consumedCap
+              val maybeRemainingBalance = toAmount(head.amount).plus(head.fee).flatMap(balance.minus)
+
+              maybeRemainingBalance match {
+                case Right(remainingBalance) if updatedTotalConsumedCap <= 1 && remainingBalance.value > 0 && tail.nonEmpty =>
```
TheMMaciek, line 178: I think you could drop the last two conditions and just let it execute once more, then the second case will not be required and on a subsequent execution the default case will be triggered for balance = 0 or if there were no more transactions.

TheMMaciek (approved):

PR #827, opened 2024-02-19
Adding READY condition to fetch snapshots using SnapshotRoutes

IPadawans:
### Changes
+ Adding condition that nodes should be in READY state to allow access to fetching snapshots using SnapshotRoutes

### JIRA
[PROT-597](https://constellationnetwork.atlassian.net/browse/PROT-597)

### DIFF
--- a/modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/SnapshotRoutes.scala
+++ b/modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/SnapshotRoutes.scala
@@ -1,35 +1,37 @@
 package org.tessellation.node.shared.http.routes
 
 import cats.effect.Async
-import cats.syntax.flatMap._
-import cats.syntax.functor._
-import cats.syntax.traverse._
+import cats.syntax.all._
 
 import org.tessellation.ext.codecs.BinaryCodec
 import org.tessellation.ext.http4s.HashVar
 import org.tessellation.ext.http4s.headers.negotiation.resolveEncoder
 import org.tessellation.kryo.KryoSerializer
+import org.tessellation.node.shared.domain.node.NodeStorage
 import org.tessellation.node.shared.domain.snapshot.storage.SnapshotStorage
 import org.tessellation.node.shared.ext.http4s.SnapshotOrdinalVar
 import org.tessellation.node.shared.infrastructure.snapshot.storage.SnapshotLocalFileSystemStorage
 import org.tessellation.routes.internal._
 import org.tessellation.schema.GlobalSnapshot
+import org.tessellation.schema.node.NodeState
 import org.tessellation.schema.snapshot.{Snapshot, SnapshotMetadata}
 import org.tessellation.security.Hasher
 import org.tessellation.security.signature.Signed
 
 import io.circe.Encoder
 import io.circe.shapes._
+import org.http4s.circe.CirceEntityCodec.circeEntityEncoder
 import org.http4s.circe.CirceEntityEncoder
 import org.http4s.dsl.Http4sDsl
-import org.http4s.{EntityEncoder, HttpRoutes}
+import org.http4s.{EntityEncoder, HttpRoutes, Response}
 import shapeless.HNil
 import shapeless.syntax.singleton._
 
 final case class SnapshotRoutes[F[_]: Async: Hasher: KryoSerializer, S <: Snapshot: Encoder, C: Encoder](
   snapshotStorage: SnapshotStorage[F, S, C],
   fullGlobalSnapshotStorage: Option[SnapshotLocalFileSystemStorage[F, GlobalSnapshot]],
-  prefixPath: InternalUrlPrefix
+  prefixPath: InternalUrlPrefix,
+  nodeStorage: NodeStorage[F]
 ) extends Http4sDsl[F]
     with PublicRoutes[F]
     with P2PRoutes[F] {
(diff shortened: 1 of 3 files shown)


modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/SnapshotRoutes.scala
```diff
@@ -40,78 +43,102 @@ final case class SnapshotRoutes[F[_]: Async: Hasher: KryoSerializer, S <: Snapsh
   implicit def binaryAndJsonEncoders[A <: AnyRef: Encoder]: List[EntityEncoder[F, A]] =
     List(BinaryCodec.encoder[F, A], CirceEntityEncoder.circeEntityEncoder[F, A])
 
+  private def badRequestNodeNotReady(currentState: NodeState): F[Response[F]] =
+    BadRequest(("message" ->> s"Node should be in READY state before fetching. Current state: $currentState") :: HNil)
```
marcinwadon, line 47: The enum NodeState specifies state as `Ready` not `READY`

IPadawans: Updated

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/SnapshotRoutes.scala
```diff
@@ -14,22 +14,25 @@ import org.tessellation.node.shared.ext.http4s.SnapshotOrdinalVar
...
+import org.http4s.{EntityEncoder, HttpRoutes, Response}
 import shapeless.HNil
 import shapeless.syntax.singleton._
 
 final case class SnapshotRoutes[F[_]: Async: Hasher: KryoSerializer, S <: Snapshot: Encoder, C: Encoder](
   snapshotStorage: SnapshotStorage[F, S, C],
   fullGlobalSnapshotStorage: Option[SnapshotLocalFileSystemStorage[F, GlobalSnapshot]],
-  prefixPath: InternalUrlPrefix
+  prefixPath: InternalUrlPrefix,
+  nodeState: F[NodeState]
```
marcinwadon, line 35: please pass the storage instead of a single function from the store. see: `snapshotStorage` above

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/SnapshotRoutes.scala
```diff
@@ -40,78 +43,102 @@ final case class SnapshotRoutes[F[_]: Async: Hasher: KryoSerializer, S <: Snapsh
   implicit def binaryAndJsonEncoders[A <: AnyRef: Encoder]: List[EntityEncoder[F, A]] =
     List(BinaryCodec.encoder[F, A], CirceEntityEncoder.circeEntityEncoder[F, A])
 
+  private def badRequestNodeNotReady(currentState: NodeState): F[Response[F]] =
+    BadRequest(("message" ->> s"Node should be in Ready state before fetching. Current state: $currentState") :: HNil)
```
TheMMaciek, line 47: Not sure BadRequest is correct - it's supposed to mean a client error and node not being ready yet is not a client error.

TheMMaciek: I would also not expose the actual state. I would just say that node or api is not ready.

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/SnapshotRoutes.scala
```diff
@@ -40,78 +43,102 @@ final case class SnapshotRoutes[F[_]: Async: Hasher: KryoSerializer, S <: Snapsh
...
-      snapshotStorage.headSnapshot.map(_.map(_.ordinal)).flatMap {
-        case Some(ordinal) => Ok(("value" ->> ordinal.value.value) :: HNil)
-        case None          => NotFound()
+      nodeState.flatMap {
+        case NodeState.Ready =>
+          snapshotStorage.headSnapshot.map(_.map(_.ordinal)).flatMap {
+            case Some(ordinal) => Ok(("value" ->> ordinal.value.value) :: HNil)
+            case None          => NotFound()
+          }
+        case currentState => badRequestNodeNotReady(currentState)
```
marcinwadon, line 57: please don't use a check for a specific node state. use `is(ready)` function instead. for `Ready` state it's a simple check, but for node being in cluster, it's more complex as there are multiple states to check against. having it done via the fn, we keep it open for extending (which is one of the SOLID rules)

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/SnapshotRoutes.scala
marcinwadon: I support that /\. `4xx` status code is reserved for client error, but this is a server error

IPadawans: What status do you suggest, I know that should be something different than 200

marcinwadon: I think that `503 Service Unavailable` would fit that purpose

IPadawans: Got it, I agree

marcinwadon: @IPadawans Here is the documentation for `4xx` vs `5xx`

>[6.5](https://www.rfc-editor.org/rfc/rfc7231#section-6.5).  Client Error 4xx
>   The 4xx (Client Error) class of status code indicates that the client
   seems to have erred.  Except when responding to a HEAD request, the
   server SHOULD send a representation containing an explanation of the
   error situation, and whether it is a temporary or permanent
   condition.  These status codes are applicable to any request method.
   User agents SHOULD display any included representation to the user.

> [6.6](https://www.rfc-editor.org/rfc/rfc7231#section-6.6).  Server Error 5xx
>   The 5xx (Server Error) class of status code indicates that the server
   is aware that it has erred or is incapable of performing the
   requested method.  Except when responding to a HEAD request, the
   server SHOULD send a representation containing an explanation of the
   error situation, and whether it is a temporary or permanent

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/SnapshotRoutes.scala
```diff
@@ -40,78 +42,119 @@ final case class SnapshotRoutes[F[_]: Async: Hasher: KryoSerializer, S <: Snapsh
...
+
   protected val httpRoutes: HttpRoutes[F] = HttpRoutes.of[F] {
     case GET -> Root / "latest" / "ordinal" =>
-      import org.http4s.circe.CirceEntityCodec.circeEntityEncoder
-
-      snapshotStorage.headSnapshot.map(_.map(_.ordinal)).flatMap {
-        case Some(ordinal) => Ok(("value" ->> ordinal.value.value) :: HNil)
-        case None          => NotFound()
+      nodeStateF.flatMap { nodeState =>
+        if (validStateForSnapshotReturn(nodeState)) {
```
marcinwadon, line 54: Please avoid naming it `F` at the end. `getNodeState` would be okay IMO. but I think you can also use it directly.
Also, that if statement can be simplified:
```scala
nodeStorage.getNodeState
  .map(validStateForSnapshotReturn)
  .ifM(
    // if true,
    // if false
  )
```

```scala
nodeStorage.getNodeState
  .map(validStateForSnapshotReturn)
  .ifM(
    snapshotStorage.headSnapshot...,
    serviceUnavailableNodeNotReady
  )
```

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/SnapshotRoutes.scala
```diff
@@ -40,78 +42,119 @@ final case class SnapshotRoutes[F[_]: Async: Hasher: KryoSerializer, S <: Snapsh
...
 
+  private val serviceUnavailableNodeNotReady: F[Response[F]] =
+    ServiceUnavailable(("message" ->> s"Node should be in Ready state before fetching") :: HNil)
+
+  private val nodeStateF: F[NodeState] = nodeStorage.getNodeState
+  private def validStateForSnapshotReturn(state: NodeState): Boolean = state === NodeState.Ready
+
   protected val httpRoutes: HttpRoutes[F] = HttpRoutes.of[F] {
     case GET -> Root / "latest" / "ordinal" =>
-      import org.http4s.circe.CirceEntityCodec.circeEntityEncoder
```
marcinwadon, line 45: are you sure you can remove these encoders?

marcinwadon: edit: probably yes 🤔 please double check if the endpoints return the binary format when required (I see there is a resolveEncoder part in some endpoints)

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/SnapshotRoutes.scala
```diff
@@ -40,78 +42,119 @@ final case class SnapshotRoutes[F[_]: Async: Hasher: KryoSerializer, S <: Snapsh
   implicit def binaryAndJsonEncoders[A <: AnyRef: Encoder]: List[EntityEncoder[F, A]] =
     List(BinaryCodec.encoder[F, A], CirceEntityEncoder.circeEntityEncoder[F, A])
 
+  private val serviceUnavailableNodeNotReady: F[Response[F]] =
+    ServiceUnavailable(("message" ->> s"Node should be in Ready state before fetching") :: HNil)
```
TheMMaciek, line 46: Maybe:
"Node not yet ready to handle snapshot API requests" 🤔 
I would not expose the internal implementation e.i. `Ready` state

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/SnapshotRoutes.scala
IPadawans: I haven't removed it, I'm importing it globally now

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/SnapshotRoutes.scala
marcinwadon: @TheMMaciek I think it's important to give info when the client can expect to make a request and success 🤔 Instead of saying explicitly about the state, it can say `Node is not ready yet` or sth like that. Would that be okay for you @TheMMaciek?
cc: @IPadawans

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/SnapshotRoutes.scala
marcinwadon: Yes, I know, but not every endpoint imported it. Having it imported globally makes it available for the other endpoints too, hence - worth checking whether it does not change the output format of those.

TheMMaciek: Yeah I was thinking about it as well. We had a convention that we specify an encoder for every endpoint separately. But I guess we didn't have resolver for encoders back then. So I guess it should be fine, but would be good to check if we are not changing some logic (I guess if there are no clashes of implicits then it's ok)

marcinwadon: actually:
> (I guess if there are no clashes of implicits then it's ok)
it may be that we have to encoders (one imported globally, one resolved) -> @IPadawans check please if the code compiles 🤔

IPadawans: It's compiling and it's working, I've tested
Captura de Tela 2024-02-19 às 10 26 33

marcinwadon: cool. please make sure the endpoints that are resolving the encoder (so pass `Accept: application/json` and `Accept: application/octet-stream` gets json/binary format. If that works, it's fine.

IPadawans: Captura de Tela 2024-02-19 às 10 29 19

Captura de Tela 2024-02-19 às 10 29 35

IPadawans: /\ this is when the node is Ready

marcinwadon: cool. so it works :+1:

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/SnapshotRoutes.scala
IPadawans: Updated

marcinwadon (approved):
Looks good to me. Please remember about extending the API doc.

TheMMaciek (approved):

IPadawans:
> Looks good to me. Please remember about extending the API doc.

https://github.com/Constellation-Labs/apidoc/pull/26

PR #828, opened 2024-02-20
WIP - Configuration

marcinwadon:


modules/dag-l0/src/main/resources/application.conf
```diff
@@ -0,0 +1,50 @@
+state-after-joining = WaitingForDownload
```
scasplte2, line 1: Is the plan that only non-consensus breaking settings will be included in the configuration? As a user, knowing which values I should vs. should not change has always seemed like a pain point. though perhaps it is not such a concern in practice?

modules/node-shared/src/main/scala/org/tessellation/node/shared/package.scala
```diff
@@ -32,20 +30,20 @@ package object shared {
 
   val nodeSharedKryoRegistrar: Map[Class[_], KryoRegistrationId[NodeSharedOrSharedRegistrationIdRange]] =
     Map[Class[_], NodeSharedKryoRegistrationId](
-      classOf[PingConsensusHealthStatus] -> 500,
-      classOf[PingHealthCheckKey] -> 501,
-      classOf[HealthCheckRoundId] -> 502,
-      classOf[PeerAvailable] -> 503,
-      classOf[PeerUnavailable] -> 504,
-      classOf[PeerUnknown] -> 505,
+      // classOf[PingConsensusHealthStatus] -> 500,
```
scasplte2, line 33: since these are registered historically, I assume keeping the comments here is helpful for future development? If that is the case, should we add a comment to not remove past registrations or can the indices be overwritten?

modules/dag-l0/src/main/resources/application.conf
marcinwadon: I think everything possible should be moved in here and this `application.conf` would be stored within the app resources, so whenever the end user would like to configure something, you would be able to provide your own application.conf and it will be merged together. I think that's the clue of this ticket to make everything configurable instead of keeping hardcoded things in the code. So answering your question:

> Is the plan that only non-consensus breaking settings will be included in the configuration?

No and I think that's intended

modules/node-shared/src/main/scala/org/tessellation/node/shared/package.scala
marcinwadon: I need to remember what we did previously as we've already had such a situation. Probably we keep it commented out, not to reuse the id. Anyway, hopefully we'll get rid of kryo.

kpudlik (approved):

TheMMaciek (approved):

PR #829, opened 2024-02-22
Clean up unused params to fix compilation warnings

TheMMaciek:


build.sbt
```diff
@@ -6,6 +6,7 @@ ThisBuild / organizationName := "tessellation"
 
 ThisBuild / evictionErrorLevel := Level.Warn
 ThisBuild / scalafixDependencies += Libraries.organizeImports
+ThisBuild / libraryDependencySchemes += "io.circe" %% "circe-core" % VersionScheme.Always
```
TheMMaciek, line 9: This alleviates the warnings for `circe-core` `0.13.0` being evicted over `0.14.3`. I think we probably shouldn't merge it in the end as it's known that for `circe-magnolia-derivation` the `0.13.0` is binary compatible with `0.14.3`, but at the same time it's not compatible in general as circe core uses Early semver and `0.13.0` can be binary incompatible with `0.14.3`. I just wanted to present it as one of the options to fix that warning, but ultimately it would be better to wait for the `circe-magnolia-derivation` release with bumped up `circe-core` and accept the warning for now.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/consensus/message.scala
```diff
@@ -1,36 +1,44 @@
...
 import org.tessellation.schema.peer.PeerId
 
-import derevo.cats.{eqv, show}
 import derevo.circe.magnolia.{decoder, encoder}
 import derevo.derive
 
 object message {
 
-  @derive(eqv, show, encoder, decoder)
+  @nowarn
```
TheMMaciek, line 14: This is one of the places in the code (among others) where we get a false positive regarding the `encoder` and `decoder` being not used. We could alleviate it with `@nowarn` flag but it will at the same time suppress any future lawful warnings. I think I will resign from using the `@nowarn` for the mentioned reason in the end. If you have some idea on how we could alleviate it with no drawbacks, let me know.

marcinwadon (approved):

PR #831, opened 2024-02-23
Bump scala and sbt versions

marcinwadon:


modules/shared/src/main/scala/org/tessellation/currency/dataApplication/storage/TraverseLocalFileSystemTempStorage.scala
```diff
@@ -57,8 +57,8 @@ object TraverseLocalFileSystemTempStorage {
   case class SnapshotAlreadyExistsInTempStorage(ordinal: SnapshotOrdinal) extends TempStorageError
   case class SnapshotNotFoundInTempStorage(ordinal: SnapshotOrdinal) extends TempStorageError
 
-  private def make[F[_]: Files]: Resource[F, Path] =
-    Files[F].tempDirectory
+  private def make[F[_]: Async]: Resource[F, Path] =
```
TheMMaciek, line 60: Why?

TheMMaciek (approved):

modules/shared/src/main/scala/org/tessellation/currency/dataApplication/storage/TraverseLocalFileSystemTempStorage.scala
marcinwadon: dunno why, but it didn't work at first glance. I didn't spend more time on this, but explicit usage with `forAsync` looks imo correct

PR #836, opened 2024-02-27
Allow signing using node keys via L0/L1 node context

scasplte2:
## Summary
Added method `Hash => F[SignatureProof]` to `L0NodeContext` and `L1NodeContext` to allow for signing data using the node key pair within the Data Application scope.

## Changes
- Updated node context traits, implementations, and calling locations to include `KeyPair` argument

## Testing
- unit tests pass
- integration tests pass

## Tickets
PROT-577

### DIFF
--- a/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
+++ b/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
@@ -64,7 +64,7 @@ abstract class CurrencyL0App(
       storages <- Storages.make[IO](sharedStorages, cfg.snapshot, method.globalL0Peer, dataApplicationService).asResource
       p2pClient = P2PClient.make[IO](sharedP2PClient, sharedResources.client, sharedServices.session)
       validators = Validators.make[IO](seedlist)
-      implicit0(nodeContext: L0NodeContext[IO]) = L0NodeContext.make[IO](storages.snapshot)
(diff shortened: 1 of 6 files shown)


marcinwadon:
Do we really need to allow signing an arbitrary data using the node's key? I see it as a simple vector of attack - just pass a transaction and you can withdraw funds from the node

marcinwadon (changes requested):
Please answer my concern so I would approve that. For now I see this as a big node's backdoor

IPadawans:
Do you just want to retrieve the SignatureProof using the node keypair? If so, besides Marcin concern, maybe we can choose a different name for the method

scasplte2:
I agree that node keys should be handled with care but I don't see how this could be a vector of attack since it is on the developer interface and not publicly facing. As such it seems like it is up to the developer to not compromise the security of their application by maintaining operational security of their API and infrastructure. 

The data application that I am working on requires the node to generate a keypair to sign data that is incoming ([as shown here](https://github.com/Constellation-Labs/flight-data/blob/4a70bce735925a01c2138bb7cbc34baa1f68fc01/source/project/flight-event-tracker/modules/data-l1/src/main/scala/io/constellationnetwork/data_l1/Main.scala#L45)). As such, without this functionality I will need to recreate the key storage utilities to load and unload keys from disk or a configuration in order to maintain a stable identity for the node.

scasplte2:
after speaking with @ryle-goehausen this can be closed since the node is expected to receive externally signed updates from the EVM contracts. This will alleviate the need for the data application to add signatures when synchronizing updates into the L1 data queue.

PR #837, opened 2024-02-27
Adding treatment when rejecting join due to node not being on the seedlist

IPadawans:
### Changes
+ Updating `ClusterRoutes.scala` to deal with rejection due to node not being on the seedlist


modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/ClusterRoutes.scala
```diff
@@ -48,13 +48,11 @@ final case class ClusterRoutes[F[_]: Async](
...
-            case NodeStateDoesNotAllowForJoining(nodeState) =>
-              Conflict(s"Node state=${nodeState} does not allow for joining the cluster.")
+            case NodeStateDoesNotAllowForJoining(nodeState) => Conflict(s"Node state=$nodeState does not allow for joining the cluster.")
             case PeerAlreadyJoinedWithNewerSession(id, _, _, _) => Conflict(s"Peer id=${id.show} already joined with newer session.")
-            case SessionAlreadyExists =>
-              Conflict(s"Session already exists.")
-            case _ =>
-              InternalServerError("Unknown error.")
+            case SessionAlreadyExists                           => Conflict(s"Session already exists.")
+            case PeerNotInSeedlist(id)                          => Conflict(s"Peer $id not in seedlist")
```
TheMMaciek, line 54: I think `PeerNotOnSeedlist` is the correct wording.

TheMMaciek (changes requested):
I would change the commit message to something like:
`Provide feedback when joining if node is not on the seedlist`
The current commit is kind of vague and bit too long - mine is a bit too long as well :D

modules/node-shared/src/main/scala/org/tessellation/node/shared/http/routes/ClusterRoutes.scala
IPadawans: 👍🏼

TheMMaciek (approved):

marcinwadon (approved):

kpudlik (approved):

PR #838, opened 2024-02-29
Reject transactions for address locked after a set ordinal

sdavidp:


modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/transaction/TransactionValidator.scala
```diff
@@ -17,7 +18,10 @@ import eu.timepit.refined.auto._
 
 trait TransactionValidator[F[_]] {
 
-  def validate(signedTransaction: Signed[Transaction]): F[TransactionValidationErrorOr[Signed[Transaction]]]
+  def validate(
+    signedTransaction: Signed[Transaction],
+    ordinal: Option[SnapshotOrdinal]
```
marcinwadon, line 23: I think you should always know the ordinal to validate against. What is the use case when you don't know the ordinal?

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/block/processing/BlockAcceptanceManager.scala
```diff
@@ -37,7 +37,8 @@ object BlockAcceptanceManager {
 
       def acceptBlocksIteratively(
         blocks: List[Signed[Block]],
-        context: BlockAcceptanceContext[F]
+        context: BlockAcceptanceContext[F],
+        maybeOrdinal: Option[SnapshotOrdinal]
```
marcinwadon, line 41: when you accept the block, you should already know the snapshot the block belongs to 🤔 unless it's used on L1 where you don't know the snapshot but you know the latest one

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -121,7 +121,7 @@ class StateChannel[
   private val runConsensus: Pipe[F, BlockConsensusInput, FinalBlock] =
     _.evalTap(input => logger.debug(s"Received block consensus input to process: ${input.show}"))
       .evalMap(
-        new BlockConsensusCell[F](_, blockConsensusContext)
+        new BlockConsensusCell[F](_, blockConsensusContext, storages.lastSnapshot.getOrdinal)
```
marcinwadon, line 124: if you don't know the last snapshot, the consensus is forbidden and should never happen (to make a consensus, you need to know tips, and tips are taken from the latest snapshot). correct me if I'm wrong @TheMMaciek

marcinwadon:
@sdavidp please remember about the PR label before releasing the hotfix

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/transaction/TransactionValidator.scala
kpudlik: I guess that the problem is that the only case when ordinal can be `Option` is when `setInitial` and `set` have not been called yet. Afaik transactions can be sent to the node that still hasn't finished the download. Maybe we should reject all the transactions if the snapshot is not set (or to be more specific if the node is in DownloadInProgress state)?

Another solution could be to use `SnapshotOrdinal.MaxValue` as default but it is a dirty one. On one hand it blocks transaction "just in case" but on the other hand it can make false negatives.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/transaction/TransactionValidator.scala
```diff
@@ -29,12 +33,17 @@ object TransactionValidator {
...
+  )
+
   def make[F[_]: Async](
     signedValidator: SignedValidator[F]
   ): TransactionValidator[F] =
     new TransactionValidator[F] {
       def validate(
-        signedTransaction: Signed[Transaction]
+        signedTransaction: Signed[Transaction],
+        maybeOrdinal: Option[SnapshotOrdinal] = None
```
kpudlik, line 46: If such validation needs some context (like ordinal from snapshot storage) then it should be in `ContextualTransactionValidator` probably 🤔

kpudlik (changes requested):

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/transaction/TransactionValidator.scala
sdavidp: @kpudlik There is already code in there to exclude the locked address. I thought these changes are for a different situation, that is, meant to be executed from the block validator.

ryle-goehausen: Contextual only blocks the L1, previous patch already implemented it there, so it'll reject transactions at L1 layer on submission, this change should only really affect block validation in that sense.

ryle-goehausen (approved):

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
```diff
@@ -47,6 +48,7 @@ object CurrencySnapshotAcceptanceManager {
       lastSnapshotContext: CurrencySnapshotContext,
       lastActiveTips: SortedSet[ActiveTip],
       lastDeprecatedTips: SortedSet[DeprecatedTip],
+      ordinal: SnapshotOrdinal,
```
marcinwadon, line 51: I see that on `develop` branch we already have a `snapshotOrdinal` property here. Would you mind doing the same here (name/order)? I think it would be easier for later merging with develop

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -120,9 +120,10 @@ class StateChannel[
 
   private val runConsensus: Pipe[F, BlockConsensusInput, FinalBlock] =
     _.evalTap(input => logger.debug(s"Received block consensus input to process: ${input.show}"))
-      .evalMap(
-        new BlockConsensusCell[F](_, blockConsensusContext)
-          .run()
+      .evalMap(blockConsensusInput =>
+        OptionT(storages.lastSnapshot.getOrdinal)
+          .getOrRaise(new Exception("last snapshot ordinal required"))
```
kpudlik, line 125: I see such exception in `BlockService` as well. I am wondering about the best approach to handle DRY.

Maybe introducing 2 methods: `getOrdinal: SnapshotOrdinal` which throws and `getOrdinalOption`?
Let me know what you think.

kpudlik (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
sdavidp: I could make this change to the trait, if that's agreeable.
```
def getOrdinal(implicit mt: MonadThrow[F]): F[SnapshotOrdinal] =
  OptionT(getOrdinalOption)
    .getOrRaise(new Exception("last snapshot ordinal required"))
```
cc: @marcinwadon

marcinwadon: What trait you mention @sdavidp ?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
```diff
@@ -120,9 +120,9 @@ class StateChannel[
 
   private val runConsensus: Pipe[F, BlockConsensusInput, FinalBlock] =
     _.evalTap(input => logger.debug(s"Received block consensus input to process: ${input.show}"))
-      .evalMap(
-        new BlockConsensusCell[F](_, blockConsensusContext)
-          .run()
+      .evalMap(blockConsensusInput =>
+        storages.lastSnapshot.getOrdinal
```
marcinwadon, line 124: is it ensured the process starts when the `storages.lastSnapshot` has been set?

sdavidp: I don't understand.
cc: @ryle-goehausen

sdavidp: I'm confused. Didn't you mention earlier that at this point we must have a snapshot or consensus can't happen?

sdavidp: I'm currently refactoring so that an `F[Option[Snapshot]]` is passed instead. The last hurdle is the previous issue with constructing `BlockAcceptanceManager` in `SharedServices`.

sdavidp: refactored to pass `F[Option[Snapshot]]` which isn't evaluated until we get into the block validator

marcinwadon: > I'm confused. Didn't you mention earlier that at this point we must have a snapshot or consensus can't happen?

well, that's my assumption that the consensus cannot run without a snapshot (we don't know tips), but I ask if you confirmed that (I did not). so it should be fine to get the ordinal here and throw if it's not set, but I'd like you to make sure we always have the ordinal in this place. (like - where the check for snapshot is, before this or somewhere after but before starting the round)

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
sdavidp: @marcinwadon I was referring to `LastSnapshotStorage` -- but the code has been refactored again and this part removed.

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
sdavidp: @marcinwadon I had not confirmed it but after looking at it again with @ryle-goehausen we noticed other issues with getting the ordinal there so instead getting the ordinal isn't evaluated until we're in the validator, if we're getting it from the storage. If it's already available then it's passed in a pure `F`.

marcinwadon: I think that passing down the lazy evaluated `F` brings race condition issues and I would suggest to re-check the previous implementation to pass the ordinal instead of evaluating the ordinal in the most bottom(children) component.

marcinwadon: cc: @TheMMaciek @kpudlik

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
marcinwadon: @sdavidp I would change "last snapshot ordinal required" to match other errors, but also I'd explicitly say that this state is unexpected/forbidden. what about: "Unexpected state - could not find the latest snapshot ordinal"?

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/block/processing/BlockNotAcceptedReason.scala
```diff
@@ -25,6 +25,9 @@ case class ValidationFailed(reasons: NonEmptyList[BlockValidationError]) extends
 @derive(eqv, show)
 case class ParentNotFound(parent: BlockReference) extends BlockRejectionReason
 
+@derive(eqv, show)
+case object SnapshotOrdinalUnavailable extends BlockRejectionReason
```
marcinwadon, line 29: for future, I think that blocks should not be rejected because of the fact the node is not initialized yet.
either we should return "not ready yet" on the endpoint not to receive data for consensus (in a similar way we do for snapshots endpoint) or we should queue txs/blocks and start processing them when the node is fully initialized. what do you think? can you create a ticket for that?
cc: @ryle-goehausen @kpudlik @TheMMaciek

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/block/processing/BlockValidator.scala
```diff
@@ -112,5 +116,22 @@ object BlockValidator {
             else
               none
         }
+
+      private val addressesLockedAtOrdinal: Map[Address, SnapshotOrdinal] = Map(
+        Address("DAG6LvxLSdWoC9uJZPgXtcmkcWBaGYypF6smaPyH") -> SnapshotOrdinal(1447110L) // NOTE: BitForex
+      )
```
marcinwadon, line 122: Please move this part to the `application.conf` (the resource-based configuration PR is already merged). Also, I think this configuration should be per environment (I assume you've seen my PR as it was open for a very long time, but if no - see how the last kryo ordinal or the last global snapshot ordinal are specified in the config). So, for mainnet, we should lock this address at ordinal X, but for testnet we should do that (or maybe not and we don't care) on a different ordinal.

marcinwadon (changes requested):
Looks good to me now. Please move the hardcoded configuration to the application.conf and we can merge it :)

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/block/processing/BlockValidator.scala
sdavidp: @marcinwadon That code is on _develop_ ... this is on a hotfix branch. Maybe moving it to `application.conf` can be done in a later PR?

marcinwadon (approved):

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/block/processing/BlockNotAcceptedReason.scala
sdavidp: Should it extend `BlockAwaitReason` instead, or are you suggesting something else altogether?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
sdavidp: @marcinwadon What about `IllegalStateException("could not find the latest snapshot ordinal")`

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/block/processing/BlockNotAcceptedReason.scala
marcinwadon: I'm not familiar with the errors within the block processing, would have to check precisely. I think that creating the ticket to offload this for future would be best

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala
marcinwadon: fine for me, however, start with the uppercase as we do for other messages :D

marcinwadon (approved):

kpudlik (approved):

PR #846, opened 2024-03-21
Add CurrencyMessage to schema

sdavidp:


modules/shared/src/main/scala/org/tessellation/currency/schema/currency.scala
```diff
@@ -110,6 +111,8 @@ object currency {
     stateProof: CurrencySnapshotStateProof,
     epochProgress: EpochProgress,
     dataApplication: Option[DataApplicationPart] = None,
+    adminFeeKeyMessage: Option[CurrencyMessage] = None,
+    stakingMessage: Option[CurrencyMessage] = None,
```
marcinwadon, line 115: is this field meant to be required or optional? if it's optional, I think we discussed on the meeting to use `Option[Option[CurrencyMessage]]` for now.

If we would like to add an `A` field to the snapshot schema, we need to wrap it into option not to influence the hashing. Hence, `Option[A]` is the type for all newly added fields to the existing schema. And as far as I remember, the message won't be there in all snapshots (but only if there is an actual change), so we need to add an optional field `A = Option[CurrencyMessage]`. That being said, the final type would be:
```
having A = Option[CurrencyMessage]
Option[A] = Option[Option[CurrencyMessage]]
```

Also, I think that having a simple type here, you can combine it into a single iterable field:
```
messages: Option[List[CurrencyMessage]]
```
which is what we were talking about some time ago that it would allow for extending the `CurrencyMessage` by additional messages' types without changing the schema further.

marcinwadon: @TheMMaciek please take a look as well

sdavidp: changed

marcinwadon (approved):
Looks correct to me. Please @TheMMaciek review as well

modules/shared/src/main/scala/org/tessellation/currency/schema/currency.scala
TheMMaciek: One last thing we could consider is using a Set instead of a List in my opinion. We don't expect any repeated events and to that matter we don't really expect two events of the same `messageType` (e.g. two consecutive staking address messages) to make it's way to the snapshot. Using Set doesn't solve that case but maybe we shouldn't try to solve it, the last message of a certain kind will always be used and the older ones need to stay there for the chain consistency even if they never take any effect.

marcinwadon: Yep, I support that /\

marcinwadon (dismissed):

modules/shared/src/main/scala/org/tessellation/currency/schema/currency.scala
TheMMaciek: Sorry I think that with json being the hashed representation we should make it a SortedSet. Not sure that json will always return a regular Set with the same order (I don't think we can expect that).

TheMMaciek (approved):

marcinwadon (approved):

PR #847, opened 2024-03-28
Update schema for currency swap

sdavidp:


modules/shared/src/main/scala/org/tessellation/schema/swap.scala
```diff
@@ -0,0 +1,94 @@
...
+  case class SpendTransaction(
+    fee: SpendTransactionFee,
+    validUntilOrdinal: SnapshotOrdinal,
+    allowSpendRef: Hash,
+    currencyId: CurrencyId,
+    amount: TransactionAmount
+  )
+
+  @derive(decoder, encoder, order, show)
+  case class AtomicAction(pending: AllowSpend, transaction: Transaction)
```
AlexBrandes, line 93: In the current case, this would accept two SpendTransaction instances. 

The ideal implementation would be to make this a wrapper (however implemented) that could accept different transaction types so that it can be extended in the future rather than being fixed to this specific case.

modules/shared/src/main/scala/org/tessellation/schema/swap.scala
```diff
@@ -0,0 +1,94 @@
...
+  ) extends DataUpdate
+
+  @derive(decoder, encoder, order, show)
+  @newtype
+  case class SpendTransactionFee(value: PosLong)
+
+  @derive(decoder, encoder, order, show)
+  case class SpendTransaction(
+    fee: SpendTransactionFee,
+    validUntilOrdinal: SnapshotOrdinal,
```
AlexBrandes, line 86: Let's make sure this field's name matches the corresponding field in AllowSpend, making both either `lastValidOrdinal` or `validUntilOrdinal`.

modules/shared/src/main/scala/org/tessellation/schema/swap.scala
```diff
@@ -0,0 +1,94 @@
...
+    reference: SwapReference,
+    allowReference: Hash,
+    minAmount: TransactionAmount,
+    maxAmount: TransactionAmount,
+    salt: TransactionSalt,
+    maxValidGsOrdinal: SnapshotOrdinal,
+    poolId: Option[PoolId],
+    priceRange: Option[PriceRange]
+//    limitParameters: List[LimitParameter]
+  ) extends DataUpdate
```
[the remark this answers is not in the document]
sdavidp: removed

modules/shared/src/main/scala/org/tessellation/schema/swap.scala
sdavidp: Adjusted class name, field types, and extended a trait.

modules/shared/src/main/scala/org/tessellation/schema/swap.scala
sdavidp: changed. also adjusted `currencyId` to match.

modules/shared/src/main/scala/org/tessellation/schema/swap.scala
```diff
@@ -0,0 +1,72 @@
...
+
+object swap {
+
+  @derive(decoder, encoder, order, show)
+  @newtype
+  case class AllowSpendFee(value: PosLong)
+
+  @derive(decoder, encoder, order, show)
+  @newtype
+  case class CurrencyId(value: Option[Address])
```
marcinwadon, line 21: How the `CurrencyId` may have an empty `value`? Isn't it non-existent `CurrencyId`?
IMO it should be an `Option[CurrencyId]` in the `AllowSpend` class if the currencyId is an optional type

modules/shared/src/main/scala/org/tessellation/schema/swap.scala
```diff
@@ -0,0 +1,72 @@
...
+  case class AllowSpend(
+    source: Address,
+    destination: Address,
+    currency: CurrencyId,
+    amount: TransactionAmount,
+    fee: AllowSpendFee,
+    parent: AllowSpendReference,
+    salt: TransactionSalt,
+    lastValidOrdinal: SnapshotOrdinal,
+    approverAddresses: List[Address]
```
marcinwadon, line 37: just `approvers`?

modules/shared/src/main/scala/org/tessellation/schema/swap.scala
```diff
@@ -0,0 +1,72 @@
...
+  @derive(decoder, encoder, order, show)
+  @newtype
+  case class AllowSpendReference(value: Hash) // exact type TBD (e.g. lastRef?)
+
+  @derive(decoder, encoder, order, show)
+  case class AllowSpend(
+    source: Address,
+    destination: Address,
+    currency: CurrencyId,
+    amount: TransactionAmount,
```
marcinwadon, line 32: IMO we shouldn't mix it with `Transaction` class, so `SwapAmount` and `SwapSalt` in here 🤔

modules/shared/src/main/scala/org/tessellation/schema/swap.scala
```diff
@@ -0,0 +1,72 @@
...
+  @newtype
+  case class SwapReference(value: Address) // exact type TBD
+
+  @derive(decoder, encoder, order, show)
+  @newtype
+  case class SpendTransactionFee(value: PosLong)
+
+  @derive(decoder, encoder, order, show)
+  case class SpendTransaction(
+    fee: SpendTransactionFee,
```
marcinwadon, line 61: IMO there should be no differentiation for transaction type. We already had `DAGTransaction` and `CurrencyTransaction` that we decided to merge under just `Transaction` and use composition over inheritance to extend the functionalities. The same should be applied to swaps

AlexBrandes: There are some pretty significant differences between SpendTransaction and Transaction. The most important one imo is that SpendTransaction is Signed by the approver(s) of an AllowSpend, not the address transferring the tokens. It also has a required reference to an allowSpend which is needed for validation of the `amount` field. Do you think those aspects would still be best supported as Transaction w/some metadata?

marcinwadon: > The most important one imo is that SpendTransaction is Signed by the approver(s) of an AllowSpend, not the address transferring the tokens

The `Transaction` itself is just a data class. We have a separate `Signed[A]` wrapper and the signing process can be defined differently for various purposes. What you wrote is just the validation function -> we would require this or that signature. Anyway, it's disjoint

> It also has a required reference to an allowSpend which is needed for validation of the amount field.

This can be a metadata part indeed.

---

If the SpendTransaction should be handled and managed differently than a normal Transaction, then it's okay to make it a totally separate thing. If the only difference is some additional validation based on the metadata, I see no point for making it separate.

AlexBrandes: Seems reasonable. I'll let others comment on it but I don't have an issue with that structure.

sdavidp: @ryle-goehausen ?

sdavidp: @marcinwadon @AlexBrandes @ryle-goehausen 

Are there any new/updated thoughts on this?

I see Marcin's concern about previously merging `DAGTransaction` and `CurrencyTransaction`. Isn't `SpendTransaction` a new/different thing? Isn't the flow going to be entirely different?

It feels to me that adding metadata to `Transaction` instead of creating a new class may introduce problems now or in the future with trying to assign meaning to existing `Transaction` fields that don't really fit or don't exist for `SpendTransaction`. E.g.  how do we set `Transaction.parent` ?

ryle-goehausen: As discussed in meeting, I think it doesn't make sense to re-use the old transaction class. We can discuss creating a more generic transaction in the future, but the previous schema is locked and we want to migrate away from that. For now, we should follow updates to this by handling it as a special case and then discuss a way to handle this more generically in the future.

marcinwadon:
please add a label

modules/shared/src/main/scala/org/tessellation/schema/swap.scala
```diff
@@ -0,0 +1,78 @@
...
+  @newtype
+  case class AllowSpendFee(value: PosLong)
+
+  @derive(decoder, encoder, order, show)
+  @newtype
+  case class CurrencyId(value: Address)
+
+  @derive(decoder, encoder, order, show)
+  @newtype
+  case class AllowSpendReference(value: Hash) // exact type TBD (e.g. lastRef?)
```
marcinwadon, line 31: is it a `TODO`? IMO you should create a ticket for that as it can easily get lost. otherwise, is the PR ready to be merged?

modules/shared/src/main/scala/org/tessellation/schema/swap.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  @derive(decoder, encoder, order, show)
+  case class AllowSpend(
+    source: Address,
+    destination: Address,
+    currency: CurrencyId,
+    amount: SwapAmount,
+    fee: AllowSpendFee,
+    parent: AllowSpendReference,
+    salt: SwapSalt,
```
marcinwadon, line 41: is the `salt` required?

modules/shared/src/main/scala/org/tessellation/schema/swap.scala
```diff
@@ -0,0 +1,78 @@
...
+  @derive(decoder, encoder, order, show)
+  @newtype
+  case class SwapAMMDataUpdateFee(value: PosLong)
+
+  @derive(decoder, encoder, order, show)
+  case class PriceRange(min: SwapAmount, max: SwapAmount)
+
+  @derive(decoder, encoder, order, show)
+  @newtype
+  case class PoolId(value: Address) // exact type TBD
```
marcinwadon, line 55: as above

modules/shared/src/main/scala/org/tessellation/schema/swap.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  @derive(decoder, encoder, order, show)
+  @newtype
+  case class SwapReference(value: Address) // exact type TBD
```
marcinwadon, line 59: as above

marcinwadon: btw. I comment my code using `TODO: @mwadon lorem ipsum` so I can easily find my todos and resolve them before merging the PR

modules/shared/src/main/scala/org/tessellation/schema/swap.scala
```diff
@@ -0,0 +1,78 @@
...
+  @derive(decoder, encoder, order, show)
+  @newtype
+  case class SpendTransactionFee(value: PosLong)
+
+  @derive(decoder, encoder, order, show)
+  case class SpendTransaction(
+    fee: SpendTransactionFee,
+    lastValidOrdinal: SnapshotOrdinal,
+    allowSpendRef: Hash,
+    currency: Option[CurrencyId],
```
marcinwadon, line 70: so if there is no currency defined, it is a DAG spend transaction, correct?

modules/shared/src/main/scala/org/tessellation/schema/swap.scala
ryle-goehausen: Empty should indicate DAG, either level of option representation should convey the same thing, either it's a Some(None) or None for DAG indicator

modules/shared/src/main/scala/org/tessellation/schema/swap.scala
ryle-goehausen: I think we should just use the same fields as transaction reference here.

modules/shared/src/main/scala/org/tessellation/schema/swap.scala
ryle-goehausen: I think this was just a convention following the original transaction schema, I don't think it's really required there either due to unique transaction references. Not sure the original reason behind that.

modules/shared/src/main/scala/org/tessellation/schema/swap.scala
ryle-goehausen: These values will only be used in the metagraph itself, so it shouldn't matter what they are here, they'll be removed from tessellation and put in project marcus is creating.

modules/shared/src/main/scala/org/tessellation/schema/swap.scala
ryle-goehausen: yes

modules/shared/src/main/scala/org/tessellation/schema/swap.scala
sdavidp: removed comment

modules/shared/src/main/scala/org/tessellation/schema/swap.scala
sdavidp: removed

modules/shared/src/main/scala/org/tessellation/schema/swap.scala
sdavidp: changed

marcinwadon (approved):
The schema looks okay for me

PR #848, opened 2024-04-02
Reactive retry for state channel binaries

kpudlik:


modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
```diff
@@ -0,0 +1,178 @@
...
+        val avgOrdinalsPassed = PosLong.unsafeFrom(Math.round(ordinalsPassed.sum.toDouble / ordinalsPassed.length))
+        expectedOrdinalsToConfirmR.set(avgOrdinalsPassed).as(avgOrdinalsPassed)
+      }
+
+      private def enqueue(binary: Hashed[StateChannelSnapshotBinary]): F[Unit] =
+        for {
+          currentOrdinal <- lastGlobalSnapshotStorage.getOrdinal.map(_.getOrElse(SnapshotOrdinal.MinValue))
+          _ <- pendingR.update { pending =>
+            if (pending.exists(_.binary.hash === binary.hash)) pending
+            else pending :+ TrackedBinary(binary, currentOrdinal)
```
sdavidp, line 138: Is `:+` (append) the best operation here? Or maybe change the data structure? Given we're appending on a `List` that means each append is _O(n)_ and it makes a copy of the list.

kpudlik: Fair point. We need to append elements and either treat it as LIFO or FIFO (with dequeueN) depending on `catchUpMode` so I think we can switch to `Queue` as it has a constant time for append.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
```diff
@@ -0,0 +1,195 @@
...
+
+      def processPending: F[Unit] = {
+        val retry = for {
+          pending <- pendingR.get
+          avgConfirmedPerSnapshot <- avgConfirmedCountPerSnapshotR.get
+          cap = Math
+            .round(avgConfirmedPerSnapshot.toDouble + (avgConfirmedPerSnapshot.toDouble / 2))
+            .toInt // Try sending 50% more than avg confirmed
+          toRetry = pending.take(cap)
+          _ <- toRetry.traverse(tracked => enqueue(tracked.binary))
```
TheMMaciek, line 107: Seems like you are enqueuing the binaries once more rather than resending them. The only moment you are using `post` function is in `process` and only if we are not in catchup mode.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
```diff
@@ -0,0 +1,195 @@
...
+            }
+            hasStalled || prev
+          }
+        )
+
+      private def enqueue(binary: Hashed[StateChannelSnapshotBinary]): F[Unit] =
+        for {
+          currentOrdinal <- lastGlobalSnapshotStorage.getOrdinal.map(_.getOrElse(SnapshotOrdinal.MinValue))
+          _ <- pendingR.update { pending =>
+            if (pending.exists(_.binary.hash === binary.hash)) pending
```
TheMMaciek, line 154: Is there a valid scenario where a situation like this would happen? Seems like we shouldn't need such check.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
```diff
@@ -0,0 +1,195 @@
...
+  ): F[StateChannelBinarySender[F]] = (
+    Ref.of[F, List[TrackedBinary]](List.empty),
+    Ref.of[F, PosLong](PosLong.MinValue),
+    Ref.of[F, NonNegLong](NonNegLong.MinValue),
+    Ref.of[F, Boolean](false)
+  ).mapN(make[F](_, _, _, _, identifierStorage, globalL0ClusterStorage, lastGlobalSnapshotStorage, stateChannelSnapshotClient))
+
+  def make[F[_]: Async: Hasher: Temporal](
+    pendingR: Ref[F, List[TrackedBinary]],
+    expectedOrdinalsToConfirmR: Ref[F, PosLong],
```
TheMMaciek, line 50: It's set in one place but never fetched -> effectively seems like not needed.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
kpudlik: Good catch. Previously both methods were merged so I somehow missed it.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
kpudlik: Yeah, now it can be just a pure function as we use it in the same place where we calculate it.

Actually, it is a question if `expectedOrdinalsToConfirm` should be calculated from the confirmation ordinal speed of the current global snapshot or the previous one.

If current -> that `Ref` is not needed and we can calculate it in place
If previous -> it is still needed and we need to `getAndSet`

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
```diff
@@ -0,0 +1,200 @@
...
+          currentOrdinal <- lastGlobalSnapshotStorage.getOrdinal.map(_.getOrElse(SnapshotOrdinal.MinValue))
+          _ <- pendingR.update { pending =>
+            pending.enqueue(TrackedBinary(binary, currentOrdinal))
+          }
+        } yield ()
+
+      private def post(binary: Hashed[StateChannelSnapshotBinary]): F[Unit] = {
+        val sendRetries = 5
+
+        val retryPolicy: RetryPolicy[F] = RetryPolicies.limitRetries(sendRetries)
```
sdavidp, line 167: This looks like it retries immediately.
Are the causes of failure such that we can do that?

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
```diff
@@ -0,0 +1,200 @@
...
+          _.isRight.pure[F]
+
+        def onFailure(binaryHashed: Hashed[StateChannelSnapshotBinary]) =
+          (_: Either[NonEmptyList[StateChannelValidationError], Unit], details: RetryDetails) =>
+            logger.info(s"Retrying sending ${binaryHashed.hash.show} to Global L0 after rejection. Retries so far ${details.retriesSoFar}")
+
+        def onError(binaryHashed: Hashed[StateChannelSnapshotBinary]) = (_: Throwable, details: RetryDetails) =>
+          logger.info(s"Retrying sending ${binaryHashed.hash.show} to Global L0 after error. Retries so far ${details.retriesSoFar}")
+
+        retryingOnFailuresAndAllErrors[Either[NonEmptyList[StateChannelValidationError], Unit]](
```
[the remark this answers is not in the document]
kpudlik: `retryingOnAllErrors` works with `MonadError` so compiler forces me to provide it implicitly for request wrapped by `F`
`retryingOnFailures` works with `wasSuccessful` predicate so we can decide if retry or not

`retryingOnFailuresAndAllErrors` is a combination of both so as long as `java.lang.Error` thrown in the request is wrapped by `MonadError` then it will be catched. All the operations inside retry policy are wrapped by `F` with `MonadError[F, E]` so we should be fine.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
kpudlik: There is no need to delay because we use `globalL0ClusterStorage.getRandomPeer` inside. So if it fails then we immediately retry to another peer selected randomly.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
```diff
@@ -0,0 +1,200 @@
...
+        val retryPolicy: RetryPolicy[F] = RetryPolicies.limitRetries(sendRetries)
+
+        def wasSuccessful: Either[NonEmptyList[StateChannelValidationError], Unit] => F[Boolean] =
+          _.isRight.pure[F]
+
+        def onFailure(binaryHashed: Hashed[StateChannelSnapshotBinary]) =
+          (_: Either[NonEmptyList[StateChannelValidationError], Unit], details: RetryDetails) =>
+            logger.info(s"Retrying sending ${binaryHashed.hash.show} to Global L0 after rejection. Retries so far ${details.retriesSoFar}")
+
+        def onError(binaryHashed: Hashed[StateChannelSnapshotBinary]) = (_: Throwable, details: RetryDetails) =>
```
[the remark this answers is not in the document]
kpudlik: This is the code that I have just moved from another file so it has been working like that already.
I see that we already have inside the retried action:

```scala
stateChannelSnapshotClient
    .send(identifier, binary.signed)(l0Peer)
    .onError(e => logger.warn(e)(s"Sending ${binary.hash.show} snapshot to Global L0 peer ${l0Peer.show} failed!"))
    .flatTap {
       case Right(_) => logger.info(s"Sent ${binary.hash.show} to Global L0 peer ${l0Peer.show}")
       case Left(errors) =>
            logger.error(s"Snapshot ${binary.hash.show} rejected by Global L0 peer ${l0Peer.show}. Reasons: ${errors.show}")
}
```

so throwables are indeed logged.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
```diff
@@ -0,0 +1,200 @@
...
+          )
+          .whenA(pending.nonEmpty)
+      } yield ()
+
+      def processPending: F[Unit] = {
+        val retry = for {
+          pending <- pendingR.get
+          avgConfirmedPerSnapshot <- avgConfirmedCountPerSnapshotR.get
+          cap = Math
+            .round(avgConfirmedPerSnapshot.toDouble + (avgConfirmedPerSnapshot.toDouble / 2))
```
TheMMaciek, line 105: This rounds to 0 for values smaller than 0.4. So for such a case we will stop retrying at all. The question is if that's unlikely to go down to such a small number or not (maybe it's ok to let it just get stuck eventually). I would actually add some kind of retry counter counting "number of retries without new confirmations". Once the average is small like 0.2 (and decreasing) and counter increase we should retry at least once every N retries if we multiply average by retry counter (increasing counter would need to have greater impact than decreasing average).

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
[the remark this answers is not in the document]
sdavidp: That's my concern, that we're catching `java.lang.Error` where my understanding is almost all cases should not be caught.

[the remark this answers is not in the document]
kpudlik: Domain-specific errors are exposed as `Either` and `StateChannelValidationError`  and handled by `retryingOnFailures`. The other errors are catched by `retryingOnAllErrors`.

If we only want to catch our errors then we could switch from `retryingOnFailuresAndAllErrors` to `retryingOnFailures` but for me, the current implementation is fine.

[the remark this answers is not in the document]
kpudlik: Also, keep in mind that we also use `retryingOnAllErrors` policy in consensus and healthcheck so if we want to switch then we should switch everywhere.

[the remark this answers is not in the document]
kpudlik: @marcinwadon @TheMMaciek wdyt? We use various retry policies in our code. Some are retrying only on domain-specific `Throwable`, some on `Left`, some on any `Throwable`s.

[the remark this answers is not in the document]
sdavidp: @kpudlik Ok, I think I've answered my own question and now I remember looking into this before.

I think we're ok because Cats has checks for `NonFatal` exceptions so it looks like those are the only ones that will be trapped.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
kpudlik: Yeah. Currently, when there are no confirmed snapshots, the average confirmations count gets divided by 2 which means that the value is going to go to zero. The quick and simple solution would be to use `ceil` so there is always at least `1` to retry. On the other hand, if Global L0 is stuck for some reason - we do not want to keep spamming with snapshots.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
kpudlik: I decided to use the current values and in some cases make a fallback to prev (hence the Ref is still needed)

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
kpudlik: >if we multiply average by retry counter

So what you propose is the following:
For avg = 10 and retriesWithoutConfirmation = 3 we should retry at least 1 snapshot after 30 ordinals?

kpudlik: How about this:

```scala
val ordinalsWithoutConfirmation = ...
val maxOrdinalsWithoutConfirmation = 5

val cap = if (ordinalsWithoutConfirmation >= maxOrdinalsWithoutConfirmation && average < 1) {
    if (ordinalsWithoutConfirmation % 5 == 0) 1 else 0  // If average < 1 try to send 1 every 5 ordinals
} else {
   average
}
```

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
```diff
@@ -0,0 +1,238 @@
...
+
+object State {
+  def empty: State = State(
+    pending = Queue.empty[TrackedBinary],
+    cap = 4L,
+    retryMode = false,
+    noConfirmationsSinceRetryCount = 0L,
+    backoffExponent = 0L
+  )
+}
```
sdavidp, line 47: Not sure if I mentioned this before:

1. These are not part of the interface but only part of the implementation in the companion object, so shouldn't their scope be confined there?
2. Is there any benefit to using refined types for an internal data structure?

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
```diff
@@ -0,0 +1,238 @@
...
+              val newState = state.copy(pending = updatedPending)
+              (newState, action)
+          }.flatMap(identity)
+        }
+
+      def confirm(globalSnapshot: GlobalIncrementalSnapshot): F[Unit] = for {
+        identifier <- identifierStorage.get
+        binaries = globalSnapshot.stateChannelSnapshots.get(identifier).toList.flatMap(_.toList)
+        confirmedHashesInGlobalSnapshot <- binaries.traverse(_.toHashed).map(_.map(_.hash))
+
```
sdavidp, line 99: That is a very long lambda function for updating `stateR`. Can it be simplified/decomposed?

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
marcinwadon: > Is there any benefit to using refined types for an internal data structure?

being constrained is basically less error-prone and will guide you step by step if you'll come back to the code after a while. imo it's the same reason why you like strictly typing "internal" variables to see the type matching what you expected before writing the implementation. so tl;dr; we should not differentiate whether it's internal or exposed part of the application, refined isn't there for exposing more constrained API, but mostly for our internal usage

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
marcinwadon: @sdavidp What you mean by decomposing it? The fn itself is decomposed into smaller parts and variables. Would moving the internal fn to a separate function work for you?

```scala
_ <- stateR.update(updateState(dep1, dep2))

[...]

def updateState(dep1, dep2)(state): NewState = {
 ...
}
```

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
```diff
@@ -0,0 +1,238 @@
...
+                    logger.info(s"Sent ${binary.hash.show} to Global L0 peer ${l0Peer.show}") >>
+                      stateR.update { state =>
+                        val tracked =
+                          state.pending.zipWithIndex.collectFirst { case found @ (tracked, _) if tracked.binary === binary => found }
+                        tracked match {
+                          case Some((tracked, i)) =>
+                            val updatedTracked = tracked.copy(sendsSoFar = NonNegLong.unsafeFrom(tracked.sendsSoFar + 1))
+                            state.copy(pending = state.pending.updated(i, updatedTracked))
+                          case None => state
+                        }
```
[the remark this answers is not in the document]
kpudlik: I agree, if we need to iterate over whole collection because we used `zipWithIndex` then we still have O(n) and we can use `.map` with `if` inside. However maybe we should use another structure like `Vector` because we want to have a constant append and efficient update but we are always updating at most `cap` elements so `O(cap)` in worst case.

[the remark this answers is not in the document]
kpudlik: >how how often is there going to be something sent that isn't "tracked" ?

We always send tracked elements. Elements are always enqueued before posting.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
```diff
@@ -0,0 +1,238 @@
...
+                  post(binary).flatTap(_ => logger.info(s"Snapshot binary of hash ${binary.hash} enqueued and sent to GL0"))
+              val newState = state.copy(pending = updatedPending)
+              (newState, action)
+          }.flatMap(identity)
+        }
+
+      def confirm(globalSnapshot: GlobalIncrementalSnapshot): F[Unit] = for {
+        identifier <- identifierStorage.get
+        binaries = globalSnapshot.stateChannelSnapshots.get(identifier).toList.flatMap(_.toList)
+        confirmedHashesInGlobalSnapshot <- binaries.traverse(_.toHashed).map(_.map(_.hash))
```
sdavidp, line 98: Since this is searched inside a loop over the pending queue it should be Set, unless the queue is always going to be very small

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
sdavidp: It looks like the lambda is performing calculations for each field of the state object so maybe if that logic was decoupled it would read better? I'll leave it to you two to decide.

sdavidp (approved):
As mentioned in Slack we should have unit tests in place, especially for `StateChannelBinarySender`

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
kpudlik: The `stateChannelSnapshots` for `identifier` are `NonEmptyList` but all these snapshots are unique for sure (I guess we didn't use `NonEmptySet` because it is `SortedSet` under the hood so it reorders binaries but we need to keep the order). I can make it a `Set` just to be sure here though.

kpudlik: And of course `contains` on `Set` is much faster if this is what you mean.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/StateChannelBinarySender.scala
kpudlik: We could either extract the pure part as @marcinwadon suggested or move each of the cases to a separate function but don't ask me because I may be biased as I wrote the statements and for me it is readable.

PR #850, opened 2024-04-04
Pinned schema for kryo compatibility

marcinwadon:


modules/shared/src/main/scala/org/tessellation/schema/GlobalSnapshotInfo.scala
```diff
@@ -58,6 +58,81 @@ object GlobalSnapshotStateProof {
...
+      balances,
+      lastCurrencySnapshots.map {
+        case (k, v) =>
+          (
+            k,
+            v.map {
+              case (Signed(inc, proofs), info) => (Signed(inc.toCurrencyIncrementalSnapshot, proofs), info)
+            }
+          )
+      },
```
sdavidp, line 84: What are your thoughts on this ?
```
      gs.lastCurrencySnapshots.view.mapValues(v =>
        v.map {
          case (Signed(inc, proofs), info) =>
            (Signed(CurrencyIncrementalSnapshotV1.fromCurrencyIncrementalSnapshot(inc), proofs), info)
        }
      ).to(gs.lastCurrencySnapshots.sortedMapFactory),
```

sdavidp: (I pasted that to the wrong place, .... but similar to code later on.

marcinwadon: @sdavidp what is `sortedMapFactory`?

marcinwadon: I see. I've already tried `view.mapValues` but didn't know the trick with sorted map factory. thanks!

kpudlik (approved):

PR #852, opened 2024-04-09
Viewable queue exposed in Data Application Routes

scasplte2:
## Summary
This PR introduces a `ViewableQueue` trait and implementation in order to expose pending data updates via a `GET` public HTTP route. This work is motivated to provide a "mempool" equivalent at the metagraph layer while maintaining atomicity, minimizing changes to existing usages of `currency.l1.modules.Queues`, and reusing the available API of Cats standard `Queue`.

Additionally, smaller changes were incorporated to facilitate encoding and decoding of `DataUpdates` and  `HttpSuite` was updated to `MutableIOSuite` as the override of `Res` and `def sharedResource` were causing compilation errors.

## Changes
- Updated `currencyL1` project definition to include test dependencies
- Updated usages of `dataUpdates: Queue -> ViewableQueue`
- Added `GET` route at `/data` to return the latest view of pending updates
- Added `MockDataApplicationService` and `...State` to facilitate testing
- Added `EntityEncoder[F, Signed[DataUpdate]]` in `DataApplication` package 
- Updated `HttpSuite` from `SimpleIOSuite -> MutableIOSuite`

## Testing
- Added `ViewbleQueueSuite` unit tests
- Added `DataApplicationRoutesSuite` integration tests cases handling
  - empty queue
  - single element
  - multiple elements

### DIFF
--- a/modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/DataApplicationRoutes.scala
+++ b/modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/DataApplicationRoutes.scala
@@ -10,6 +10,7 @@ import org.tessellation.currency.l1.domain.error.{InvalidDataUpdate, InvalidSign
 import org.tessellation.currency.schema.currency.{CurrencyIncrementalSnapshot, CurrencySnapshotInfo}
 import org.tessellation.ext.http4s.error._
 import org.tessellation.node.shared.domain.cluster.storage.L0ClusterStorage
+import org.tessellation.node.shared.domain.queue.ViewableQueue
 import org.tessellation.node.shared.domain.snapshot.storage.LastSnapshotStorage
 import org.tessellation.routes.internal._
 import org.tessellation.schema.{GlobalIncrementalSnapshot, GlobalSnapshotInfo}
@@ -17,20 +18,19 @@ import org.tessellation.security.signature.Signed
 import org.tessellation.security.{Hasher, SecurityProvider}
 
 import eu.timepit.refined.auto._
-import io.circe.Decoder
-import io.circe.shapes._
-import org.http4s.HttpRoutes
+import io.circe.syntax.EncoderOps
+import io.circe.{Decoder, Encoder}
 import org.http4s.circe.CirceEntityCodec.{circeEntityDecoder, circeEntityEncoder}
 import org.http4s.dsl.Http4sDsl
+import org.http4s.{EntityDecoder, HttpRoutes}
+import org.typelevel.log4cats.SelfAwareStructuredLogger
 import org.typelevel.log4cats.slf4j.Slf4jLogger
-import shapeless._
-import shapeless.syntax.singleton._
 
 final case class DataApplicationRoutes[F[_]: Async: Hasher: SecurityProvider: L1NodeContext](
   dataApplicationPeerConsensusInput: Queue[F, Signed[ConsensusInput.PeerConsensusInput]],
   l0ClusterStorage: L0ClusterStorage[F],
   dataApplication: BaseDataApplicationL1Service[F],
-  dataUpdatesQueue: Queue[F, Signed[DataUpdate]],
+  dataUpdatesQueue: ViewableQueue[F, Signed[DataUpdate]],
   lastGlobalSnapshotStorage: LastSnapshotStorage[F, GlobalIncrementalSnapshot, GlobalSnapshotInfo],
   lastCurrencySnapshotStorage: LastSnapshotStorage[F, CurrencyIncrementalSnapshot, CurrencySnapshotInfo]
 )(implicit S: Supervisor[F])
@@ -38,13 +38,13 @@ final case class DataApplicationRoutes[F[_]: Async: Hasher: SecurityProvider: L1
     with PublicRoutes[F]
     with P2PRoutes[F] {
 
-  def logger = Slf4jLogger.getLogger[F]
+  def logger: SelfAwareStructuredLogger[F] = Slf4jLogger.getLogger[F]
 
   protected val prefixPath: InternalUrlPrefix = "/"
 
   protected val public: HttpRoutes[F] = HttpRoutes.of[F] {
     case req @ POST -> Root / "data" =>
-      implicit val decoder = dataApplication.signedDataEntityDecoder
+      implicit val decoder: EntityDecoder[F, Signed[DataUpdate]] = dataApplication.signedDataEntityDecoder
 
       req
         .asR[Signed[DataUpdate]] {
@@ -57,7 +57,7 @@ final case class DataApplicationRoutes[F[_]: Async: Hasher: SecurityProvider: L1
                   case Invalid(e) => InternalServerError(InvalidDataUpdate(e.toString).toApplicationError)
                   case Valid(_) =>
                     dataUpdatesQueue.offer(hashed.signed) >>
-                      Ok(("hash" ->> hashed.hash.value) :: HNil)
+                      Ok(("hash" -> hashed.hash.value).asJson)
                 }
           }
         }
@@ -65,6 +65,11 @@ final case class DataApplicationRoutes[F[_]: Async: Hasher: SecurityProvider: L1
 
     case GET -> Root / "l0" / "peers" =>
       l0ClusterStorage.getPeers.flatMap(Ok(_))
+
+    case GET -> Root / "data" =>
+      implicit val signedEncoder: Encoder[Signed[DataUpdate]] = Signed.encoder(dataApplication.dataEncoder)
+
+      dataUpdatesQueue.view.flatMap(Ok(_))
   }
 
   protected val p2p: HttpRoutes[F] = HttpRoutes.of[F] {
--- a/modules/currency-l1/src/test/scala/org/tessellation/currency/l1/http/DataApplicationRoutesSuite.scala
+++ b/modules/currency-l1/src/test/scala/org/tessellation/currency/l1/http/DataApplicationRoutesSuite.scala
@@ -1,0 +1,142 @@
+package org.tessellation.currency.l1.http
+
+import cats.data.NonEmptySet
+import cats.effect.std.{Queue, Random, Supervisor}
+import cats.effect.{IO, Resource}
+import cats.syntax.all._
+
+import org.tessellation.currency.dataApplication.{ConsensusInput, DataUpdate, L1NodeContext}
+import org.tessellation.currency.l1.DummyDataApplicationL1Service
+import org.tessellation.currency.l1.DummyDataApplicationState.{DummyUpdate, dummyUpdateGen}
+import org.tessellation.currency.l1.node.L1NodeContext
+import org.tessellation.currency.schema.currency.{CurrencyIncrementalSnapshot, CurrencySnapshotInfo}
+import org.tessellation.ext.cats.effect.ResourceIO
+import org.tessellation.json.JsonSerializer
+import org.tessellation.kryo.KryoSerializer
+import org.tessellation.node.shared.domain.cluster.storage.L0ClusterStorage
+import org.tessellation.node.shared.domain.queue.ViewableQueue
+import org.tessellation.node.shared.domain.snapshot.storage.LastSnapshotStorage
+import org.tessellation.schema._
+import org.tessellation.schema.generators.signedOf
+import org.tessellation.schema.peer.L0Peer
+import org.tessellation.schema.snapshot.{Snapshot, SnapshotInfo}
+import org.tessellation.security._
+import org.tessellation.security.signature.Signed
+import org.tessellation.shared.sharedKryoRegistrar
+
+import org.http4s.Method.GET
+import org.http4s._
+import org.http4s.client.dsl.io._
+import org.http4s.implicits.http4sLiteralsSyntax
+import org.scalacheck.Gen
+import suite.HttpSuite
+
+object DataApplicationRoutesSuite extends HttpSuite {
+
+  type Res = (SecurityProvider[IO], Hasher[IO], Supervisor[IO], Random[IO])
+
+  def construct(updateQueue: ViewableQueue[F, Signed[DataUpdate]]): IO[HttpRoutes[IO]] =
+    sharedResource.use { res =>
+      implicit val (sp, h, sv, r) = res
+      for {
+        lastGlobalSnapshotStorage <- mockLastSnapshotStorage[GlobalIncrementalSnapshot, GlobalSnapshotInfo]
+        lastCurrencySnapshotStorage <- mockLastSnapshotStorage[CurrencyIncrementalSnapshot, CurrencySnapshotInfo]
+        implicit0(ctx: L1NodeContext[IO]) = L1NodeContext.make[IO](lastGlobalSnapshotStorage, lastCurrencySnapshotStorage)
+        consensusQueue <- Queue.unbounded[F, Signed[ConsensusInput.PeerConsensusInput]]
+        l0ClusterStorage <- mockL0ClusterStorage
+        l1Service <- DummyDataApplicationL1Service.make[IO]
+        dataApi = DataApplicationRoutes(
+          consensusQueue,
+          l0ClusterStorage,
+          l1Service,
+          updateQueue,
+          lastGlobalSnapshotStorage,
+          lastCurrencySnapshotStorage
+        )
+      } yield dataApi.publicRoutes
+    }
+
+  def sharedResource: Resource[IO, Res] = for {
+    implicit0(ks: KryoSerializer[IO]) <- KryoSerializer.forAsync[IO](sharedKryoRegistrar)
+    sp <- SecurityProvider.forAsync[IO]
+    implicit0(json2bin: JsonSerializer[IO]) <- JsonSerializer.forSync[IO].asResource
+    h: Hasher[IO] <- IO(Hasher.forSync[IO](_ => JsonHash)).asResource
+    sv <- Supervisor[IO]
+    r <- Random.scalaUtilRandom.asResource
+  } yield (sp, h, sv, r)
+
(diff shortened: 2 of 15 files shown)


modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/DataApplicationRoutes.scala
```diff
@@ -57,14 +53,19 @@ final case class DataApplicationRoutes[F[_]: Async: Hasher: SecurityProvider: L1
                   case Invalid(e) => InternalServerError(InvalidDataUpdate(e.toString).toApplicationError)
                   case Valid(_) =>
                     dataUpdatesQueue.offer(hashed.signed) >>
-                      Ok(("hash" ->> hashed.hash.value) :: HNil)
+                      Ok(("hash" -> hashed.hash.value).asJson)
```
scasplte2, line 56: I changed this to use Circe directly as the use of shapeless was causing an issue at some point. I don't recall exactly the issue but it was something like "unable to find Encoder[<complexShapelessThing>]". Using Circe reduces the overall number of deps in this file and replicates the previous behavior but if there is a non-obvious reason this should not be changed, I can revert back.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/queue/ViewableQueue.scala
```diff
@@ -0,0 +1,49 @@
...
+  def view: F[List[A]]
+}
+
+object ViewableQueue {
+
+  def unbounded[F[_]: Concurrent, A]: F[ViewableQueue[F, A]] =
+    Queue.unbounded[F, A].flatMap(q => make(q))
+
+  def make[F[_]: Concurrent, A](queue: Queue[F, A]): F[ViewableQueue[F, A]] =
+    Ref.of[F, List[A]](List.empty).map { state =>
```
scasplte2, line 25: I realized there may be an issue with synchronization between the `Ref` and the `Queue`. I'll investigate other implementations that wrap these two effects into one atomic update.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/queue/ViewableQueue.scala
```diff
@@ -0,0 +1,51 @@
...
+              queue.tryOfferN(items).flatMap { failedItems =>
+                val successfulItems = items.diff(failedItems)
+                state.update(_ ++ successfulItems).as(failedItems)
+              }
+            }
+
+          def tryTakeN(n: Option[Int]): F[List[A]] =
+            mutex.lock.use(_ => queue.tryTakeN(n).flatTap(taken => state.update(cur => cur.diff(taken))))
+
+          def view: F[List[A]] = state.get
```
scasplte2, line 48: Views may be out of date if queried while the mutex is locked but my intention is to minimize the impact to write operations.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/queue/ViewableQueue.scala
scasplte2: added a Mutex lock and wrapped write operations to ensure atomicity

modules/currency-l1/src/test/scala/org/tessellation/currency/l1/http/DataApplicationRoutesSuite.scala
```diff
@@ -0,0 +1,118 @@
...
+  def construct(peer: Peer)(updateQueue: ViewableQueue[F, Signed[DataUpdate]]): IO[HttpRoutes[IO]] =
+    sharedResource.use { res =>
+      implicit val (sp, h, sv, r) = res
+      for {
+        lastGlobalSnapshotStorage <- LastSnapshotStorage.make[F, GlobalIncrementalSnapshot, GlobalSnapshotInfo]
+        lastCurrencySnapshotStorage <- LastSnapshotStorage.make[F, CurrencyIncrementalSnapshot, CurrencySnapshotInfo]
+        implicit0(ctx: L1NodeContext[IO]) = L1NodeContext.make(lastGlobalSnapshotStorage, lastCurrencySnapshotStorage)
+        consensusQueue <- Queue.unbounded[F, Signed[ConsensusInput.PeerConsensusInput]]
+        l0ClusterStorage <- L0ClusterStorage.make(L0Peer.fromPeer(peer))
+        l1Service <- MockDataApplicationL1Service.make
```
marcinwadon, line 54: Do you really need all of these to be initialized properly? You may provide some stubs instead

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/queue/ViewableQueue.scala
```diff
@@ -0,0 +1,51 @@
...
+  def unbounded[F[_]: Concurrent, A]: F[ViewableQueue[F, A]] =
+    Queue.unbounded[F, A].flatMap(q => make(q))
+
+  def make[F[_]: Concurrent, A](queue: Queue[F, A]): F[ViewableQueue[F, A]] =
+    (
+      Ref[F].of(List.empty[A]),
+      Mutex[F]
+    ).tupled.map {
+      case (state, mutex) =>
+        new ViewableQueue[F, A] {
```
marcinwadon, line 30: We should avoid locking if possible and write lock-free code or without semantically blocking.
`tryOfferN` and `tryTakeN` implementations on Queue take this approach, hence their name `try...`. By wrapping it basically by mutex, it's no longer `try` and it's blocking now. Just thinking out loud (didn't validate if it works) maybe you can keep the queue inside the Ref, so you use the Ref's internal mechanism for retrying in case of concurrent update. Or maybe we should not use a `Queue` anymore, if we need to see all the data anyway.

marcinwadon (changes requested):

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/DataApplicationRoutes.scala
marcinwadon: That's okay :+1:

modules/currency-l1/src/test/scala/org/tessellation/currency/l1/http/DataApplicationRoutesSuite.scala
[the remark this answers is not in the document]
scasplte2: I will look into stubbing some of these.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/queue/ViewableQueue.scala
scasplte2: @marcinwadon I attempted to write a test to explore this concern by comparing a blocking and non-blocking version of `ViewableQueue`. Indeed, I am unable to get the `Ref` and `Queue` out of sync (although they do not share the same order), so it would seem that the non-blocking version would be preferred. However, I was curious to explore timing differences between the two implementations and found that the blocking implementation may be more performant (at least in this particular scenario). That was unexpected so perhaps the test is not well suited but in any case here are the two tests with results. 

- With 20k Int, non-blocking, I get an elapsed time of "Test took 8843498 microseconds."
- With 20k Int, blocking, I get an elapsed time of "Test took 2255935 microseconds."

Note that I ran the tests independently to get a more accurate estimate of timing. If you run them at once their timing will be nearly identical.

## Test
```Scala
object ViewableQueueSuite extends SimpleIOSuite with Checkers {

//  test("Insert 20k Int to blocking Queue") {
//    for {
//      start <- Clock[IO].realTime
//      queue <- Queue.unbounded[IO, Int]
//      viewQueue <- ViewableQueue.makeBlocking[IO, Int](queue)
//      _ <- List.range(0, 20000).map(i => viewQueue.offer(i)).parSequence
//      result <- (viewQueue.view, viewQueue.tryTakeN(Some(20000))).parTupled
//      (viewResult, takeResult) = result
//      end <- Clock[IO].realTime
//      _ <- IO.println(s"Test took ${end - start}.")
//    } yield expect(viewResult.size == takeResult.size)
//  }

  test("Insert 20k Int to NON-blocking Queue") {
    for {
      start <- Clock[IO].realTime
      queue <- Queue.unbounded[IO, Int]
      viewQueue <- ViewableQueue.makeNonblocking[IO, Int](queue)
      _ <- List.range(0, 20000).map(i => viewQueue.offer(i)).parSequence
      result <- (viewQueue.view, viewQueue.tryTakeN(Some(20000))).parTupled
      (viewResult, takeResult) = result
      end <- Clock[IO].realTime
      _ <- IO.println(s"Test took ${end - start}.")
    } yield expect(viewResult.size == takeResult.size)
  }
}
```

marcinwadon: I think you should measure the time of the particular method only, not the initialization and numbers generation. Anyway, the test doesn't check the concurrent access which would happen in real life

scasplte2: Do you have any suggestions on ways I can stress test this? I have made an attempt with the test below and find that the blocking version is approx 4x faster than the non-blocking version for 20k entries inserted via 100 Fibers each inserting 200 elements with small randomized delay between each element.

**Result**
Blocking -------- Total insertion time for all fibers: 2603593 microseconds
Non-blocking - Total insertion time for all fibers: 7764411 microseconds

**Test**
```Scala
  test("Insert 20k Int to NON-blocking Queue via multiple fibers") {
    for {
      queue <- Queue.unbounded[IO, Int]
      viewQueue <- ViewableQueue.makeNonblocking[IO, Int](queue)
      actions = List
        .range(0, 100)
        .map(i => List.range(i * 200, (i + 1) * 200))
        .flatMap(list => list.map(item => IO.sleep((math.random() * 2).millisecond) *> viewQueue.offer(item)))
      start <- Clock[IO].realTime
      fibers <- actions.parTraverseN(10)(_.start)
      _ <- fibers.traverse(_.join)
      end <- Clock[IO].realTime
      _ <- IO.println(s"Total insertion time for all fibers: ${end - start}")
      result <- (viewQueue.view, viewQueue.tryTakeN(Some(20000))).tupled
      (viewResult, takeResult) = result
    } yield expect(viewResult.size == takeResult.size)
  }
````

Comment (marcinwadon) [reply modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/queue/ViewableQueue.scala:30]: @scasplte2 Can you try this instead?

```scala
object ViewableQueue {

  def make[F[_]: Concurrent, A]: F[ViewableQueue[F, A]] =
    Ref[F].of(List.empty[A]).map { state =>
      new ViewableQueue[F, A] {
        def offer(item: A): F[Unit] =
          state.update(_ :+ item)

        def take: F[A] =
          state.modify {
            case head :: tail => (tail, head)
            case Nil => throw new NoSuchElementException("take from an empty queue")
          }

        def tryOfferN(items: List[A]): F[List[A]] =
          state.modify { current =>
            val updated = current ++ items
            (updated, List.empty[A])
          }

        def tryTakeN(n: Option[Int]): F[List[A]] =
          state.modify { current =>
            val (taken, remaining) = current.splitAt(n.getOrElse(0))
            (remaining, taken)
          }

        def view: F[List[A]] = state.get
      }
    }
}
```

or, to optimize the `offer` method to be O(1), we can use sth like [this](https://gist.github.com/ASRagab/82411aa5b78a9b06589d79238b15520e), so the `offer` would have O(1), and `take` would have amortized O(1) (O(n) if necessary but amortized over n `take` operations).

```scala
object ViewableQueue {

  case class QueueState[A](enqueueList: List[A], dequeueList: List[A])

  def emptyState[A]: QueueState[A] = QueueState(Nil, Nil)

  def make[F[_]: Concurrent, A]: F[ViewableQueue[F, A]] =
    Ref[F].of(emptyState[A]).map { state =>
      new ViewableQueue[F, A] {
        def offer(item: A): F[Unit] =
          state.update {
            case QueueState(enqueue, dequeue) => QueueState(item :: enqueue, dequeue)
          }

        private def ensureNonEmpty: F[Unit] =
          state.modify {
            case QueueState(enqueue, Nil) => (QueueState(Nil, enqueue.reverse), ())
            case s => (s, ())
          }

        def take: F[A] =
          for {
            _ <- ensureNonEmpty
            taken <- state.modify {
              case QueueState(enqueue, head :: tail) => (QueueState(enqueue, tail), head)
              case _ => throw new NoSuchElementException("take from an empty queue") // better handling needed, consider handling this later in effect
            }
          } yield taken

        def tryOfferN(items: List[A]): F[List[A]] =
          state.update {
            case QueueState(enqueue, dequeue) => QueueState(items.reverse ++ enqueue, dequeue)
          }.as(List.empty[A])

        def tryTakeN(n: Option[Int]): F[List[A]] =
          for {
            _ <- ensureDequeueListNonEmpty
            taken <- state.modify {
              case QueueState(enqueue, dequeue) =>
                val (taken, remaining) = dequeue.splitAt(n.getOrElse(0))
                (QueueState(enqueue, remaining), taken)
            }
          } yield taken

        def view: F[List[A]] =
          state.get.map {
            case QueueState(enqueue, dequeue) => dequeue ++ enqueue.reverse
          }
      }
    }
}
```

I think we don't need to use a Queue if we're going to keep all elements in the `List` anyway.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/queue/ViewableQueue.scala
```diff
@@ -0,0 +1,51 @@
...
+          def tryOfferN(items: List[A]): F[List[A]] =
+            mutex.lock.use { _ =>
+              queue.tryOfferN(items).flatMap { failedItems =>
+                val successfulItems = items.diff(failedItems)
+                state.update(_ ++ successfulItems).as(failedItems)
+              }
+            }
+
+          def tryTakeN(n: Option[Int]): F[List[A]] =
+            mutex.lock.use(_ => queue.tryTakeN(n).flatTap(taken => state.update(cur => cur.diff(taken))))
```
marcinwadon, line 46: btw `.use { _ => fn }` is equivalent to `.surround { fn }`

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/queue/ViewableQueue.scala
scasplte2: Just to share, the test results for the second version you shared using `QueueState`.

**Test**
- Single Fiber
  - Insertion time: 519862 microseconds.
  - Retrieval time 50224 microseconds.
- Multiple fibers
  - Total insertion time for all fibers: 739888 microseconds
  - Retrieval time 26752 microseconds.

marcinwadon: Does it mean that the insertion is slightly slower but the retrieval is much faster? 🤔 Anyway, both ways would work fine IMO, so your call

modules/currency-l1/src/test/scala/org/tessellation/currency/l1/http/DataApplicationRoutesSuite.scala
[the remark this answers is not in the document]
scasplte2: @marcinwadon I've gone ahead and mocked the storages and the L1 service in the latest commit. I believe this is sufficient but let me know if there are other improvements that should be made.

marcinwadon:
@scasplte2 please remember to squash your commits into a single atomic change.

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/DataApplicationRoutes.scala
marcinwadon: @scasplte2 oh, I remember now. Are you able to use the new syntax with multiple fields (multiple types)? I think that was why we used shapeless.

modules/currency-l1/src/test/scala/org/tessellation/currency/l1/MockDataApplicationL1Service.scala
```diff
@@ -0,0 +1,75 @@
...
+import org.tessellation.schema.SnapshotOrdinal
+import org.tessellation.security.hash.Hash
+import org.tessellation.security.signature.Signed
+
+import io.circe.{Decoder, Encoder}
+import org.http4s.circe.CirceEntityCodec.circeEntityDecoder
+import org.http4s.circe.jsonEncoderOf
+import org.http4s.{EntityDecoder, EntityEncoder, HttpRoutes}
+
+object MockDataApplicationL1Service {
```
marcinwadon, line 19: It's actually a stub :)

marcinwadon: Maybe you can name it Dummy...Service, DummyUpdate, and so on 🤔

marcinwadon (dismissed):

marcinwadon (approved):

modules/currency-l1/src/test/scala/org/tessellation/currency/l1/MockDataApplicationL1Service.scala
scasplte2: renamed to dummy prefix

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/DataApplicationRoutes.scala
```diff
@@ -92,4 +93,6 @@ final case class DataApplicationRoutes[F[_]: Async: Hasher: SecurityProvider: L1
           logger.error(err)(s"An error occured") >> InternalServerError()
         }
   }
+
+  def logger = Slf4jLogger.getLogger[F]
```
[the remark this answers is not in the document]
scasplte2: I will watch for this, I commonly use Ctrl+Alt+L to format my code and IntelliJ must have a different code arrangement configured in the settings.

PR #855, opened 2024-04-15
Calculate and validate minimal metagraph fee

TheMMaciek:
I plan to add couple of simple unit tests for the fee calculator but other than that this PR is ready for review.
UPDATE: tests added

FYI we will need to set the configurations for every env before deploying.

### DIFF
--- a/modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/statechannel/FeeCalculator.scala
+++ b/modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/statechannel/FeeCalculator.scala
@@ -1,0 +1,92 @@
+package org.tessellation.node.shared.domain.statechannel
+
+import java.math.{MathContext, RoundingMode}
+
+import cats.MonadThrow
+import cats.data.{NonEmptyList, NonEmptySet}
+import cats.syntax.all._
+
+import scala.collection.immutable.SortedMap
+
+import org.tessellation.currency.schema.currency.SnapshotFee
+import org.tessellation.schema.SnapshotOrdinal
+import org.tessellation.schema.balance.Balance
+
+import eu.timepit.refined.api.Refined
+import eu.timepit.refined.auto._
+import eu.timepit.refined.numeric.Interval
+import eu.timepit.refined.types.numeric.{NonNegBigDecimal, NonNegInt, NonNegLong}
+
+case class FeeCalculatorConfig(
+  baseFee: NonNegLong,
+  stakingWeight: NonNegBigDecimal,
+  computationalCost: NonNegLong,
+  proWeight: NonNegBigDecimal
+)
+
+object FeeCalculatorConfig {
+  val noFee: FeeCalculatorConfig =
+    FeeCalculatorConfig(baseFee = 0L, stakingWeight = BigDecimal(0.0), computationalCost = 1L, proWeight = BigDecimal(0.0))
+
+  def getByOrdinal(configs: SortedMap[SnapshotOrdinal, FeeCalculatorConfig])(ordinal: SnapshotOrdinal): FeeCalculatorConfig =
+    configs
+      .rangeTo(ordinal)
+      .lastOption
+      .map { case (_, config) => config }
+      .getOrElse(FeeCalculatorConfig.noFee)
+}
+
+sealed trait FeeCalculator[F[_]] {
+  def calculateRecommendedFee(
+    maybeOrdinal: Option[SnapshotOrdinal],
+    delay: NonNegLong = 0L
+  )(staked: Balance, sizeKb: NonNegInt, tip: NonNegLong = 0L, proScore: FeeCalculator.ProScore = 0.0d): F[SnapshotFee]
+}
+
+object FeeCalculator {
+  def make[F[_]: MonadThrow](configs: SortedMap[SnapshotOrdinal, FeeCalculatorConfig]): FeeCalculator[F] =
+    new FeeCalculator[F] {
+      private val allConfigs: NonEmptyList[FeeCalculatorConfig] =
+        NonEmptyList
+          .fromList(configs.values.toList)
+          .getOrElse(NonEmptyList.one(FeeCalculatorConfig.noFee))
+
+      private def getConfig = FeeCalculatorConfig.getByOrdinal(configs)(_)
+
+      def calculateRecommendedFee(
+        maybeOrdinal: Option[SnapshotOrdinal],
+        delay: NonNegLong
+      )(staked: Balance, sizeKb: NonNegInt, tip: NonNegLong, proScore: ProScore): F[SnapshotFee] =
+        maybeOrdinal
+          .fold(allConfigs) { ordinal =>
+            NonEmptySet
+              .of(ordinal, ordinal.plus(delay))
+              .toNonEmptyList
+              .map(getConfig)
+          }
+          .traverse(calculate(_)(staked, sizeKb, tip, proScore))
+          .map(_.maximum)
+    }
+
+  private def calculate[F[_]: MonadThrow](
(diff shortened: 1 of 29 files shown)


modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/statechannel/FeeCalculator.scala
```diff
@@ -0,0 +1,89 @@
...
+          .rangeTo(ordinal)
+          .lastOption
+          .map { case (_, config) => config }
+          .getOrElse(FeeCalculatorConfig.noFee)
+
+      /** Calculates the recommended fee for given stake, size and pro score and adds optional tip on top. Calculates the worst case
+        * scenario (highest) fee based on the provided configs range picked as follows:
+        *   - all configs when ordinal is None
+        *   - ordinal to ordinal + delay when ordinal is non empty.
+        */
```
TheMMaciek, line 52: Do we want such scaladoc explaining a bit, or do we consider a clean code enough?

modules/shared/src/main/scala/org/tessellation/schema/currencyMessage.scala
```diff
@@ -60,4 +62,21 @@ object currencyMessage {
   case class CurrencyMessage(messageType: MessageType, address: Address, parentOrdinal: MessageOrdinal) {
     def ordinal: MessageOrdinal = parentOrdinal.next
   }
+
+  def fetchStakingAddress(state: CurrencySnapshotInfo): Option[Address] =
+    state.lastMessages
+      .flatMap(_.view.filterKeys(_ == MessageType.Staking).values.maxByOption(_.ordinal))
+      .map(_.address)
```
sdavidp, line 69: Why is it implemented as above instead of this?
```
state.lastMessages
  .flatMap(_.get(MessageType.Staking))
  .map(_.address)
```
Is there a plan to change the implementation so that the values are a list?

TheMMaciek: I guess my mind was fixed on the idea that there may be multiple messages of the same type in the snapshot (but with increasing ordinal) and hence I reflected that here as well - which (you are right) doesn't make sense for the Map having `MessageType` as a key. Will change the code to your implementation. Thanks

sdavidp: Thank you @TheMMaciek . I'm doing something similar in my PR so I'll copy your code there.

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/statechannel/FeeCalculator.scala
```diff
@@ -0,0 +1,95 @@
...
+        NonEmptyList
+          .fromList(configByOrdinal.values.toList)
+          .getOrElse(NonEmptyList.one(FeeCalculatorConfig.noFee))
+
+      private def getConfigByOrdinal(ordinal: SnapshotOrdinal): FeeCalculatorConfig =
+        configByOrdinal
+          .rangeTo(ordinal)
+          .lastOption
+          .map { case (_, config) => config }
+          .getOrElse(FeeCalculatorConfig.noFee)
```
sdavidp, line 52: Can we move this into the companion object `FeeCalculatorConfig` ? It's something I'll need to use too.

sdavidp:
LGTM

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/statechannel/FeeCalculator.scala
```diff
@@ -0,0 +1,97 @@
...
+      BigDecimal(1) / (BigDecimal(1) + BigDecimal(staked.value) * stakingWeight + BigDecimal(proScore) * proWeight)
+
+    val roundUp = new MathContext(0, RoundingMode.UP)
+    val fee = (BigDecimal(baseFee) * workAmount * workMultiplier + BigDecimal(tip)).round(roundUp).toBigInt
+
+    Either
+      .cond(fee.isValidLong, fee.longValue, "calculated fee exceeded long max value")
+      .flatMap(NonNegLong.from)
+      .map(SnapshotFee(_))
+      .leftMap(msg => new Error(s"Critical error during snapshot fee calculation: $msg!"))
```
sdavidp, line 92:
```suggestion
      .leftMap(msg => new IllegalStateException(s"Critical error during snapshot fee calculation: $msg!"))
```

I don't think we should be creating `java.lang.Error`

TheMMaciek: I can change it, but can you elaborate why we shouldn't?

TheMMaciek: I used `Error` as it is an unhandleable situation - we can't recover from it.

sdavidp: > I can change it, but can you elaborate why we shouldn't?

Same reason why I don't feel we should new `Throwable`.
My feeling has been `java.lang.Error` is reserved for the JVM whereas `Exception` is for us mere mortals.
I'm open to a more modern explanation/approach though...

TheMMaciek: Ok I did some more reading and I agree it seems `Error` is for more severe and mainly architectural stuff, like `StackOverflow` and such. Hence using `IllegalStateException` seems more appropriate given it's still an unchecked exception.

kpudlik (dismissed):

sdavidp (dismissed):

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/statechannel/FeeCalculator.scala
marcinwadon: If you aimed to write a scaladoc, it should be sth like that:
```scala
/** Calculates the recommended fee for a given stake, size, pro score, and optional tip.
  *
  * This method calculates the worst-case scenario (highest) fee based on the provided configs range.
  * The range is determined as follows:
  *   - If `ordinal` is `None`, all configs are considered.
  *   - If `ordinal` is non-empty, the range is from `ordinal` to `ordinal + delay`.
  *
  * @param maybeOrdinal an optional [[SnapshotOrdinal]] representing the starting ordinal for the config range
  * @param delay a [[NonNegLong]] representing the number of ordinals to include in the config range
  * @param staked the [[Balance]] of the staked amount
  * @param sizeKb the size of the transaction in kilobytes as a [[NonNegInt]]
  * @param tip an optional [[NonNegLong]] representing the additional tip amount to include in the fee calculation
  * @param proScore the [[ProScore]] of the transaction sender
  * @return an [[F]] containing the calculated [[SnapshotFee]] representing the recommended fee
  */
```

TheMMaciek: I didn't aim for a fully fledged scaladoc. I wanted to give enough insight, so that it's easier to understand the reasoning. My question was rather about whether we want such minimalistic description or do we consider code to be enough (self explaining I guess). We don't add much descriptions like that in general.

marcinwadon: If you want to add some side note, go for it. but IMO it doesn't add much value since the code is self explanatory. Anyway, take a look at my rewrite of the comment which in my opinion is more straightforward to understand.

kpudlik (dismissed):

PR #856, opened 2024-04-15
Hashing logic

marcinwadon:


kpudlik (changes requested):

modules/shared/src/main/scala/org/tessellation/security/Hasher.scala
```diff
@@ -28,25 +24,50 @@ trait Hasher[F[_]] {
...
+  def getForOrdinal(ordinal: SnapshotOrdinal): Hasher[F]
+
+  def withCurrent[A](fn: Hasher[F] => A): A = fn(getCurrent)
+  def getCurrent: Hasher[F]
+}
+
+object HasherSelector {
+  def apply[F[_]: HasherSelector]: HasherSelector[F] = implicitly
+
+  def alwaysCurrent[F[_]: HasherSelector]: HasherSelector[F] = forSyncAlwaysCurrent(HasherSelector[F].getCurrent)
```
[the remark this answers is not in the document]
marcinwadon: Yes, in places where we share the same code, based on the passed Snapshot type, we want or we don't want the switching logic. The easiest is to pass hasher selector with fixed hasher.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
```diff
@@ -182,7 +187,7 @@ abstract class CurrencyL0App(
                   NodeState.Initial,
                   NodeState.LoadingGenesis,
                   NodeState.GenesisReady
-                )(programs.genesis.accept(dataApplicationService)(m.genesisPath)) >>
+                )(hasherSelector.withCurrent(implicit hasher => programs.genesis.accept(dataApplicationService)(m.genesisPath))) >>
```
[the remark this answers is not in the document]
marcinwadon: You're right :+1: but I'm not sure about the genesis using current. If we do not validate the genesis, then we can do that. But technically, on mainnet/testnet/integrationnet genesis was created using kryo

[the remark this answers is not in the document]
kpudlik: Fair point 👍🏻

kpudlik (dismissed):

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/SentStateChannelBinaryTrackingService.scala
```diff
@@ -0,0 +1,88 @@
...
+    identifierStorage: IdentifierStorage[F]
+  ): SentStateChannelBinaryTrackingService[F] =
+    new SentStateChannelBinaryTrackingService[F] {
+
+      private val logger = Slf4jLogger.getLogger
+
+      private def updateByGlobalSnapshotBinaries(
+        binaries: List[Signed[StateChannelSnapshotBinary]]
+      )(implicit hasher: Hasher[F]): F[Unit] = {
+        val confirmedHashes = binaries.traverse(_.toHashed).map(_.flatMap(b => List(b.hash, b.lastSnapshotHash)))
```
sdavidp, line 45: Would it be more efficient if this were a `Set` since we're later doing `contains` in a loop?
Also, is there a chance that some hashes will be duplicated? E.g. `b.hash` in one binary is the `b.lastSnapshotHash` in another?

kpudlik: I guess that PR isn't rebased on `develop` @marcinwadon because the service you are commenting @sdavidp doesn't exist anymore on `develop` and it already uses `Set`:

`confirmedHashesInGlobalSnapshot <- binaries.traverse(_.toHashed).map(_.map(_.hash)).map(_.toSet)`.

>Also, is there a chance that some hashes will be duplicated? E.g. b.hash in one binary is the b.lastSnapshotHash in another?

Yes, for example:
```
List(
   Binary(hash: "A", lastSnapshotHash: "000"),
   Binary(hash: "B", lastSnapshotHash: "A")
)
```

gives `List("A", "000" "B", "A")` which makes it even worse without `Set`.

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/services/SentStateChannelBinaryTrackingService.scala
```diff
@@ -0,0 +1,88 @@
...
+
+        confirmedHashes.flatMap { confirmed =>
+          pendingR.modify { currPending =>
+            val newPending = currPending.zipWithIndex.collect {
+              case ((hashed, _), cuttingPoint) if confirmed.contains(hashed.hash) => cuttingPoint + 1
+            }.lastOption
+              .map(currPending.splitAt)
+              .map { case (_, stillPending) => stillPending }
+              .getOrElse(currPending)
```
sdavidp, line 54: Is this equivalent to getting the `stillPending` list?
```
val stillPending = currPending.drop(1 + currPending.lastIndexWhere {
  case (hashed, _) => confirmed.contains(hashed.hash)
})
```

modules/shared/src/main/scala/org/tessellation/security/Hasher.scala
```diff
@@ -57,36 +78,79 @@ object Hasher {
...
+          JsonSerializer[F].serialize(d.toEncode)(d.jsonEncoder)
+        case _ =>
+          JsonSerializer[F].serialize[A](data)
+      }).map(Hash.fromBytes)
 
+    def hash[A: Encoder](data: A): F[Hash] =
+      hashJson(data)
   }
+
+  // def forSync[F[_]: Sync: KryoSerializer: JsonSerializer](hashSelect: HashSelect): Hasher[F] = new Hasher[F] {
```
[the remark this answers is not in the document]
marcinwadon: I'll remove it at the very end

kpudlik (approved):

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/Download.scala
```diff
@@ -120,24 +119,29 @@ object Download {
...
-            globalSnapshotContextFns
-              .createContext(lastContext, lastSnapshot, snapshot)
-              .handleErrorWith(_ => InvalidChain.raiseError[F, GlobalSnapshotContext])
-              .flatTap { _ =>
-                snapshotStorage.writePersisted(snapshot)
-              }
-              .map((snapshot, _))
+        hasherSelector.forOrdinal(lastSnapshot.ordinal.next) { implicit hasher =>
+          fetchSnapshot(none, lastSnapshot.ordinal.next).flatMap { snapshot =>
+            lastSnapshot.toHashed[F].flatMap { hashed =>
```
sdavidp, line 124: This may be correct but it's confusing to know when we hit the boundary condition.
Why is `lastSnapshot` hashed with the hasher retrieved with `lastSnapshot.ordinal.next` instead of `lastSnapshot.ordinal` ?

marcinwadon: Let me take a closer look :+1: Thanks for spot

sdavidp: The thing that occurred to me is when we hit the boundary we'll need 2 hashers, won't we?

sdavidp:
LGTM

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/Download.scala
marcinwadon: @sdavidp fixed

modules/shared/src/main/scala/org/tessellation/security/Hasher.scala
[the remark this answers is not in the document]
marcinwadon: removed

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/Download.scala
```diff
@@ -121,23 +120,30 @@ object Download {
...
+            .flatMap { hashed =>
+              Applicative[F].unlessA {
+                Validator.isNextSnapshot(hashed, snapshot.value)
+              }(InvalidChain.raiseError[F, Unit])
+            } >>
             globalSnapshotContextFns
               .createContext(lastContext, lastSnapshot, snapshot)
               .handleErrorWith(_ => InvalidChain.raiseError[F, GlobalSnapshotContext])
               .flatTap { _ =>
                 snapshotStorage.writePersisted(snapshot)
```
sdavidp, line 136: It's not clear to me which hasher is used here when the hash filename is generated.

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/snapshot/programs/Download.scala
```diff
@@ -121,23 +120,30 @@ object Download {
         val (lastSnapshot, lastContext) = result
 
         fetchSnapshot(none, lastSnapshot.ordinal.next).flatMap { snapshot =>
-          lastSnapshot.toHashed[F].flatMap { hashed =>
-            Applicative[F].unlessA {
-              Validator.isNextSnapshot(hashed, snapshot.value)
-            }(InvalidChain.raiseError[F, Unit])
-          } >>
+          hasherSelector
+            .forOrdinal(lastSnapshot.ordinal) { implicit hasher =>
```
sdavidp, line 124: Makes sense, and seems to match other places where we get hasher for `lastSnapshot`

sdavidp (approved):

kpudlik (approved):

sdavidp (dismissed):

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotValidator.scala
```diff
@@ -141,6 +144,11 @@ object CurrencySnapshotValidator {
                   creationResult.focus(_.artifact.dataApplication).replace(expected.dataApplication)
               }
             }
+            .map { creationResult =>
+              if (creationResult.artifact.messages.isEmpty || creationResult.artifact.messages.exists(_.isEmpty))
```
sdavidp, line 148:
```suggestion
              if (creationResult.artifact.messages.forall(_.isEmpty))
```
scala> Option.empty[List[Int]].forall(_.isEmpty)
val res0: Boolean = true

scala> Option[List[Int]](List.empty).forall(_.isEmpty)
val res1: Boolean = true

scala> Option[List[Int]](List(1)).forall(_.isEmpty)
val res2: Boolean = false

sdavidp (dismissed):

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotValidator.scala
marcinwadon: good tip, thanks :+1:

sdavidp (approved):

PR #857, opened 2024-04-19
Add fee transaction and update currency snapshot schema

TheMMaciek:


modules/shared/src/main/scala/org/tessellation/currency/schema/FeeTransaction.scala
```diff
@@ -0,0 +1,22 @@
...
+import derevo.circe.magnolia.{decoder, encoder}
+import derevo.derive
+
+@derive(decoder, encoder, order, ordering, show)
+case class FeeTransaction(
+  source: Address,
+  destination: Address,
+  amount: Amount,
+  parent: TransactionReference,
+  salt: TransactionSalt
```
TheMMaciek, line 18: I wonder if we need it for `FeeTransaction`. I actually don't remember if it was necessary for regular transaction in the end 🤔

modules/shared/src/main/scala/org/tessellation/currency/schema/FeeTransaction.scala
```diff
@@ -0,0 +1,22 @@
...
+import org.tessellation.schema.balance.Amount
+import org.tessellation.schema.transaction.{TransactionOrdinal, TransactionReference, TransactionSalt}
+
+import derevo.cats.{order, show}
+import derevo.circe.magnolia.{decoder, encoder}
+import derevo.derive
+
+@derive(decoder, encoder, order, ordering, show)
+case class FeeTransaction(
+  source: Address,
```
TheMMaciek, line 14: @ryle-goehausen will this transaction always be signed by the owner? I wonder if we can maybe ditch the source field and just derive it from the signer id.

modules/shared/src/main/scala/org/tessellation/currency/schema/FeeTransaction.scala
```diff
@@ -0,0 +1,22 @@
...
+  destination: Address,
+  amount: Amount,
+  parent: TransactionReference,
```
kpudlik, line 17: Hmm, according to ticket "It must have a unique FeeTransactionReference distinct from other transaction references". Are we staying with `TransactionReference` anyway?

TheMMaciek: Yeah we talked about it and it seems it's not strictly necessary to not reuse `TransactionReference` schema itself as long as we set the values correctly in the code - e.i. we don't use regular tx refs as fee tx refs by mistake. We could use a new `FeeTransactionReference` schema if we decide that this will help us in not making some programmatic error.

marcinwadon: I support that. We have many places in the code where we operate on references, so it would be easier to see a different type

TheMMaciek: Cool let's do that then.

modules/shared/src/main/scala/org/tessellation/currency/schema/FeeTransaction.scala
ryle-goehausen: I think it would be better to leave this in here as it was in original docs.

sdavidp (approved):

PR #862, opened 2024-04-30
Add fee transactions to CurrencyIncrementalSnapshot

sdavidp:


modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/DataApplicationSnapshotAcceptanceManager.scala
```diff
@@ -174,10 +176,14 @@ object DataApplicationSnapshotAcceptanceManager {
           service.hashCalculatedState(newDataState.calculated)
         )
 
+        feeTransactions <- OptionT.liftF(
+          service.calculateFees(validatedUpdates)
```
TheMMaciek, line 180: Not sure if `calculateFees` is a correct name as if I understand correctly the fees can only be retrieved, but they can't be created on the spot.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotCreator.scala
```diff
@@ -170,6 +170,10 @@ object CurrencySnapshotCreator {
 
           (height, subHeight) <- getHeightAndSubHeight(lastArtifact, deprecated, remainedActive, accepted)
 
+          feeTransactions = dataApplicationAcceptanceResult
+            .map(result => SortedSet.from(result.feeTransactions))
+            .filter(_.nonEmpty)
```
TheMMaciek, line 175: You need to consider whether in the empty case `feeTransactions` field should be a `None` or `Some(SortedSet.empty)`. The none case means that this field is not yet used and some with empty set means that it's used but there are no fees to be paid. Not saying that we can't put None in (maybe it's interchangeable), but the semantic difference is there and it will be reflected in the hash.

sdavidp: @TheMMaciek That's what I had yesterday but `Some(SortedSet.empty)` causes the github tests to fail. If that's the correct implementation then we need to adjust the tests?

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/DataApplicationSnapshotAcceptanceManager.scala
sdavidp: @ryle-goehausen ☝️  thoughts?

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotCreator.scala
TheMMaciek: I guess, `Some(SortedSet.empty)` should work in my opinion as it's a valid scenario in my opinion.

sdavidp: @TheMMaciek I resolved it the same way it's handled for the `messages` field.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/DataApplicationSnapshotAcceptanceManager.scala
ryle-goehausen: extractFees is more indicative I think, since they exist already and we're just pulling them from nested data

sdavidp: Changed

kpudlik (dismissed):

kpudlik (approved):

PR #863, opened 2024-05-07
Fix currency hashing

marcinwadon:


modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotValidator.scala
```diff
@@ -132,29 +133,60 @@ object CurrencySnapshotValidator {
...
         mkEvents.flatMap { events =>
-          currencySnapshotCreator
-            .createProposalArtifact(lastArtifact.ordinal, lastArtifact, lastContext, trigger, events, rewards, facilitators)
+          def usingKryo =
+            currencySnapshotCreator
+              .createProposalArtifact(
+                lastArtifact.ordinal,
+                lastArtifact,
+                lastContext,
+                Hasher.forJson,
```
IPadawans, line 144: Shouldn't be Hasher.forKryo?

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotValidator.scala
```diff
@@ -132,29 +133,60 @@ object CurrencySnapshotValidator {
...
+                facilitators
+              )
+
+          def usingJson =
+            currencySnapshotCreator
+              .createProposalArtifact(
+                lastArtifact.ordinal,
+                lastArtifact,
+                lastContext,
+                Hasher.forKryo,
```
IPadawans, line 157: Shouldn't be Hasher.forJson?

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotValidator.scala
```diff
@@ -132,29 +133,60 @@ object CurrencySnapshotValidator {
         }
       })
 
-      val recreateFn = (trigger: ConsensusTrigger) =>
+      def recreateFn(trigger: ConsensusTrigger) =
         mkEvents.flatMap { events =>
-          currencySnapshotCreator
-            .createProposalArtifact(lastArtifact.ordinal, lastArtifact, lastContext, trigger, events, rewards, facilitators)
+          def usingKryo =
```
kpudlik, line 138: I would consider

```scala
          def usingHasher = (hasher: Hasher[F]) => currencySnapshotCreator
            .createProposalArtifact(
              lastArtifact.ordinal,
              lastArtifact,
              lastContext,
              hasher,
              trigger,
              events,
              rewards,
              facilitators
            )
          
          def usingKryo = usingHasher(Hasher.forKryo)

          def usingJson = usingHasher(Hasher.forJson)
```

because the only difference between these 2 is the hasher provided but it is still not very DRY

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotValidator.scala
kpudlik: Yeah, it will work with both ways: Json with Kryo fallback and Kryo with Json fallback but semantically it is a mistake and should be reversed.

kpudlik (approved):

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotValidator.scala
marcinwadon: yep, you're right @IPadawans

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotValidator.scala
marcinwadon: :+1:

PR #867, opened 2024-05-14
Apply FeeTransactions on proposal's balances

kpudlik:


marcinwadon (approved):

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
```diff
@@ -166,6 +178,31 @@ object CurrencySnapshotAcceptanceManager {
           .getOrElse(acc)
       }
 
+    private def acceptFeeTxs(
+      balances: SortedMap[Address, Balance],
+      maybeTxs: Option[SortedSet[Signed[FeeTransaction]]]
+    ): (SortedMap[Address, Balance], Option[SortedSet[Signed[FeeTransaction]]]) =
```
TheMMaciek, line 184: Shouldn't this function return an Either or even throw if the fee transaction cannot be applied?

kpudlik: Yeah. I have changed it to throw unexpected state when someone implements fee transactions in a wrong way so balance is negative for example.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
```diff
@@ -166,6 +180,31 @@ object CurrencySnapshotAcceptanceManager {
...
+                val updatedBalances = balances
+                  .updated(tx.source, src)
+                  .updated(tx.destination, dst)
+                val acceptedTxs = txs.map(_ + tx)
+                (updatedBalances, acceptedTxs)
+            }.fold(e => new Throwable(s"Unexpected state when applying fee transactions: ${e.show}").raiseError, _.pure[F])
+          }
+        case None =>
+          (balances, maybeTxs).pure[F]
+      }
```
sdavidp, line 206: 1. I think this will abort on the first fee transaction that fails. Is that ok?
2. What do you think about this refactor?

```suggestion
    private def acceptFeeTxs(
      balances: SortedMap[Address, Balance],
      maybeTxs: Option[SortedSet[Signed[FeeTransaction]]]
    ): F[(SortedMap[Address, Balance], Option[SortedSet[Signed[FeeTransaction]]])] =
      maybeTxs match {
        case None => (balances, maybeTxs).pure[F]
        case Some(txs) =>
          txs
            .foldLeftM((balances, SortedSet.empty[Signed[FeeTransaction]])) {
              case ((balances, txs), tx) =>
                (
                  balances.getOrElse(tx.source, Balance.empty).minus(tx.amount).liftTo[F],
                  balances.getOrElse(tx.destination, Balance.empty).plus(tx.amount).liftTo[F]
                ).mapN {
                  case (src, dst) =>
                    val updatedBalances = balances
                      .updated(tx.source, src)
                      .updated(tx.destination, dst)
                    (updatedBalances, txs + tx)
                }
                  .adaptErr(e => new ArithmeticException(s"Unexpected state when applying fee transaction ${tx.show}: ${e.getMessage}"))
            }
            .map { case (m, s) => (m, s.some) }
      }
```

kpudlik: >I think this will abort on the first fee transaction that fails. Is that ok?

If fee transactions are being accepted in order then it is the correct way to handle it since it is an unexpected state caused by bad implementation of data application. Alternatively we could just keep applying balances (treated as just non-refined `Long`) and just verify at the very end if any of these `Long`s is negative (doesn't fit `NonNegLong`). Maybe the alternative way is better 🤔

kpudlik: >What do you think about this refactor?
I like the `ArithmeticException` one :+1:

sdavidp (dismissed):
@TheMMaciek @marcinwadon Do you want to review as well?

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
kpudlik: I have implemented the alternative version mentioned above^

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
```diff
@@ -166,6 +178,35 @@ object CurrencySnapshotAcceptanceManager {
...
+      maybeTxs: Option[SortedSet[Signed[FeeTransaction]]]
+    ): F[(SortedMap[Address, Balance], Option[SortedSet[Signed[FeeTransaction]]])] =
+      maybeTxs match {
+        case None => (balances, maybeTxs).pure[F]
+        case Some(txs) =>
+          txs
+            .foldLeft(balances.view.mapValues(_.value.value).toMap) {
+              case (balances, tx) =>
+                val src = balances.getOrElse(tx.source, Balance.empty.value.value) - tx.amount.value
+                val dst = balances.getOrElse(tx.destination, Balance.empty.value.value) + tx.amount.value
```
TheMMaciek, line 192: We need to remember about adding FeeTransaction validation checking that `src` and `dst` are different. In other case it would be possible to generate balances for oneself here. We could also think about an implementation for which it is ambivalent which address is sending to which other address (I guess a list of balance updates - pairs of positive and negative)

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
```diff
@@ -166,6 +178,35 @@ object CurrencySnapshotAcceptanceManager {
...
+
+                balances
+                  .updated(tx.source, src)
+                  .updated(tx.destination, dst)
+            }
+            .toList
+            .foldLeftM(SortedMap.empty[Address, Balance]) {
```
TheMMaciek, line 199: I wonder if we could come up with an implementation that only transforms addresses influenced by fee transactions here to balance, rather than all the balances (I guess most of them unchanged). Maybe we could start with finding all the src and dst addresses, then fetch balances for those addresses and then overwrite their balances in the balances map rather than taking the whole balances map in line 189.

sdavidp (approved):

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
```diff
@@ -166,6 +178,41 @@ object CurrencySnapshotAcceptanceManager {
...
+              case (acc, (address, balance)) =>
+                NonNegLong
+                  .from(balance)
+                  .map(Balance(_))
+                  .map(acc.updated(address, _))
+                  .leftMap(e => new ArithmeticException(s"Unexpected state when applying fee transactions: $e"))
+                  .liftTo[F]
+            }
+            .map { updates =>
+              (balances ++ updates, txs.some)
```
TheMMaciek, line 212: I think you are only accepting all or none, so possibly you don't need to pass fee txs to acceptance manager. You will either way throw an error if you can't accept a single fee tx.

TheMMaciek (approved):
LGTM

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
kpudlik: I left it just for consistency as all the `acceptXYZTransactions` return both balances and txs.

PR #870, opened 2024-05-17
Add CLI command to merge currency message files

sdavidp:
This has been tested with the following test cases:

1. One valid staking message file
2. Several identical staking messages files
3. Several valid staking message files and one with bad json (produces error)
4. Several valid staking message files and one with bad proof (produces error)
5. Several valid staking message files and one with different ordinal (produces error)
6. Several valid staking messages files with different proofs, including one which contains all proofs: result is identical to the file with all the proofs.

### DIFF
--- a/modules/wallet/src/main/scala/org/tessellation/wallet/Main.scala
+++ b/modules/wallet/src/main/scala/org/tessellation/wallet/Main.scala
@@ -3,6 +3,8 @@ package org.tessellation.wallet
 import java.security.KeyPair
 
 import cats.MonadThrow
+import cats.data.Validated.{Invalid, Valid}
+import cats.data._
 import cats.effect.std.Console
 import cats.effect.{Async, ExitCode, IO}
 import cats.syntax.all._
@@ -17,7 +19,7 @@ import org.tessellation.schema.currencyMessage.{CurrencyMessage, MessageOrdinal,
 import org.tessellation.schema.transaction.{Transaction, TransactionAmount, TransactionFee}
 import org.tessellation.security._
 import org.tessellation.security.key.ops._
-import org.tessellation.security.signature.Signed
+import org.tessellation.security.signature.{Signed, SignedValidator}
 import org.tessellation.shared.sharedKryoRegistrar
 import org.tessellation.wallet.cli.env.EnvConfig
 import org.tessellation.wallet.cli.method._
@@ -27,6 +29,7 @@ import org.tessellation.wallet.transaction.createTransaction
 import com.monovore.decline._
 import com.monovore.decline.effect._
 import fs2.io.file.Path
+import io.circe.Encoder
 import io.circe.syntax._
 import io.estatico.newtype.ops._
 import org.typelevel.log4cats.SelfAwareStructuredLogger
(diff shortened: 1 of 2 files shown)


ryle-goehausen (approved):

modules/wallet/src/main/scala/org/tessellation/wallet/Main.scala
```diff
@@ -75,6 +78,12 @@ object Main
                     createCurrencyMessage[IO](keyPair, MessageType.Staking, address, parentOrdinal, outputPath)
                       .handleErrorWith(err => logger.error(err)(s"Error while creating staking signing message."))
                       .as(ExitCode.Success)
+                  case MergeSigningMessages(files, outputPath: Option[Path]) =>
+                    implicit val hasher = Hasher.forJson[IO]
+                    mergeMessages[IO](files)
+                      .flatMap(writeJson[IO, Signed[CurrencyMessage]](outputPath))
+                      .handleErrorWith(err => logger.error(err)(s"Error merging currency messages."))
+                      .as(ExitCode.Success)
```
marcinwadon, line 86: so in case of error, we end up with success exit code?

modules/wallet/src/main/scala/org/tessellation/wallet/Main.scala
```diff
@@ -134,4 +143,37 @@ object Main
...
+      Validated.condNec(
+        inputs.toIterable.map(_.value).toSet.size == 1,
+        inputs,
+        "Messages are not identical"
+      )
+
+    def validateProofs(inputs: SignedMessages): F[SignedMessagesErrorOr[SignedMessages]] =
+      inputs
+        .traverse(input => signedValidator.validateSignatures(input))
+        .map(_.reduceLeft(_ *> _))
```
marcinwadon, line 166: `*>` -> `>>`

modules/wallet/src/main/scala/org/tessellation/wallet/Main.scala
sdavidp: @marcinwadon It's the same pattern used on all the commands. `ExitCode.Error` makes more sense to me for command-line tools though, which is the Unix convention.

sdavidp: @marcinwadon @ryle-goehausen 
I could make this change to all the commands:
Do the `.as` so on success we exit with zero, but on error with one. WDYT?
```
.as(ExitCode.Success)
.handleErrorWith(err => logger.error(err)(s"Error merging currency messages.").as(ExitCode.Error))
```

marcinwadon: :+1:

modules/wallet/src/main/scala/org/tessellation/wallet/Main.scala
```diff
@@ -134,4 +148,37 @@ object Main
...
+        .map(_.leftMap(_.map(_.show)).as(inputs))
+
+    files.traverse { p =>
+      OptionT(readFromJsonFile[F, Signed[CurrencyMessage]](p))
+        .getOrRaise(new Exception(s"Unable to load JSON from $p"))
+    }
+      .flatMap(inputs => validateProofs(inputs).map(_ *> validateMessageValues(inputs)))
+      .flatMap {
+        case Valid(inputs)   => Signed(inputs.head.value, inputs.map(_.proofs).reduceLeft(_ ++ _)).pure[F]
```
TheMMaciek, line 180: I would just think whether sorting signatures would make sense so that by executing in different order of files or on different machines we do get the same output with the same inputs.

TheMMaciek (approved):
In general LGTM. I'm ok with it throwing an error and not merging any signatures if there is a single bad apple as long as the user gets a clear feedback on which object/signature was faulty so that discarding (or modifying) file with wrong input is possible (no trial and error needed). We could also try to merge the majority of objects and signatures so the user doesn't have to clean up but we would need to clearly indicate that not all messages were uniform/signed correctly or both. Given that we want to push it forward I think current approach pushing the responsibility to the user is ok.

So I would only verify whether from the error returned a user is able to adjust the command and exclude failing files and I would also consider whether the signatures shouldn't be ordered.

sdavidp:
> In general LGTM. I'm ok with it throwing an error and not merging any signatures if there is a single bad apple as long as the user gets a clear feedback on which object/signature was faulty so that discarding (or modifying) file with wrong input is possible (no trial and error needed). We could also try to merge the majority of objects and signatures so the user doesn't have to clean up but we would need to clearly indicate that not all messages were uniform/signed correctly or both. Given that we want to push it forward I think current approach pushing the responsibility to the user is ok.
> 
> So I would only verify whether from the error returned a user is able to adjust the command and exclude failing files and I would also consider whether the signatures shouldn't be ordered.

@ryle-goehausen What are your thoughts?

sdavidp:
> In general LGTM. I'm ok with it throwing an error and not merging any signatures if there is a single bad apple as long as the user gets a clear feedback on which object/signature was faulty so that discarding (or modifying) file with wrong input is possible (no trial and error needed). We could also try to merge the majority of objects and signatures so the user doesn't have to clean up but we would need to clearly indicate that not all messages were uniform/signed correctly or both. Given that we want to push it forward I think current approach pushing the responsibility to the user is ok.
> So I would only verify whether from the error returned a user is able to adjust the command and exclude failing files and I would also consider whether the signatures shouldn't be ordered.

@TheMMaciek I thought of that too and started keeping the files associated with the inputs but decided to first keep it simple in case that wasn't necessary. I'm not sure how we'd report which file is the problem when we fail on mismatched message values, but for invalid signatures it is easier if we mention which file is the culprit. Even without the file I think the user can still figure it out because the `id` and `signature` is displayed.
```
InvalidSignatures{invalidSignatures=NonEmptySortedSet(SignatureProof{id=ba43050d,signature=30460221})}
```
My examples are contrived so there are multiple files with the same `id` and `signature` but I suspect in practice grepping for the offending id/sig prefix will find the bad file.

WDYT?

modules/wallet/src/main/scala/org/tessellation/wallet/Main.scala
TheMMaciek: Great, I saw an `NonEmptyList` somewhere and thought it was for the sigs as well.

TheMMaciek:
> @TheMMaciek I thought of that too and started keeping the files associated with the inputs but decided to first keep it simple in case that wasn't necessary. I'm not sure how we'd report which file is the problem when we fail on mismatched message values, but for invalid signatures it is easier if we mention which file is the culprit. Even without the file I think the user can still figure it out because the id and signature is displayed.

> InvalidSignatures{invalidSignatures=NonEmptySortedSet(SignatureProof{id=ba43050d,signature=30460221})}

> My examples are contrived so there are multiple files with the same id and signature but I suspect in practice grepping for the offending id/sig prefix will find the bad file.

`WDYT?`

I agree that determining which file is the "correct" one in terms of the mismatched message is hard - we would need to determine majority on the messages which already is a part of the functionality we want to avoid. So maybe just indicating that the message is not uniform across all the files is enough - the user can compare that part with any comparing tool. We could also assume that the first file is the reference file and compare messages against this one - then we could indicate exactly which files are offending. For the signatures I think we could explicitly say which signature for which id and file is wrong (as this is after determining message uniformity).

sdavidp:
> > @TheMMaciek I thought of that too and started keeping the files associated with the inputs but decided to first keep it simple in case that wasn't necessary. I'm not sure how we'd report which file is the problem when we fail on mismatched message values, but for invalid signatures it is easier if we mention which file is the culprit. Even without the file I think the user can still figure it out because the id and signature is displayed.
> 
> > InvalidSignatures{invalidSignatures=NonEmptySortedSet(SignatureProof{id=ba43050d,signature=30460221})}
> 
> > My examples are contrived so there are multiple files with the same id and signature but I suspect in practice grepping for the offending id/sig prefix will find the bad file.
> 
> > `WDYT?`
> 
> I agree that determining which file is the "correct" one in terms of the mismatched message is hard - we would need to determine majority on the messages which already is a part of the functionality we want to avoid. So maybe just indicating that the message is not uniform across all the files is enough - the user can compare that part with any comparing tool. We could also assume that the first file is the reference file and compare messages against this one - then we could indicate exactly which files are offending. For the signatures I think we could explicitly say which signature for which id and file is wrong (as this is after determining message uniformity).

Well the message is fairly simple right now -- just 3 fields -- so it should be easy to see which doesn't match.
```
case class CurrencyMessage(messageType: MessageType, address: Address, parentOrdinal: MessageOrdinal)
```

PR #871, opened 2024-05-24
Ensure Currency fee transactions not used yet

sdavidp:


modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotValidator.scala
```diff
@@ -160,10 +160,6 @@ object CurrencySnapshotValidator {
               if (creationResult.artifact.messages.forall(_.isEmpty))
                 creationResult.focus(_.artifact.messages).replace(expected.messages)
               else creationResult
-            }.map { creationResult =>
-              if (creationResult.artifact.feeTransactions.forall(_.isEmpty))
-                creationResult.focus(_.artifact.feeTransactions).replace(expected.feeTransactions)
-              else creationResult
```
TheMMaciek, line 166: What is the logic behind this code? 🤔 I'm wondering both for fee transactions and messages, whether this code makes sense. It doesn't seem like these types are outside of the core protocol reach, so I wonder why would we overwrite the values in any case 🤔.

TheMMaciek: I guess it may make sense for fee transactions to be overwritten given that they are derived from data updates we don't have control over, but I don't see a reason for messages to be overwritten 🤔. For fee txs shouldn't we only overwrite if the data application implementation is not provided and not if the outcome is empty?

sdavidp: @marcinwadon  ☝️

marcinwadon: @TheMMaciek @sdavidp When the specific metagraph implementation is not provided (like on the Global L0 - we don't know the specific implementation of each metagraph), we cannot validate those fields. We assume it was validated on Currency level and signed by validators. So on the Global L0 when we want to validate the currency snapshot, we can validate known-fields only and we need to keep "as is" other fields (here - we try to recreate the snapshot using the same input but we cannot recreate all fields correctly, so we need to rewrite it as it is).

TheMMaciek: Yeah exactly so currency messages aren't something that needs to be rewritten. Fee transactions need to be given they are derived from data application implementation. But currency messages are 100% validatable. Also isn't it that we should rewrite all these fields the moment the data application implemantation is not provided rather than when the field is empty after recreation?

marcinwadon: I'm already on vacations so cannot help much but the best would be to ask @kpudlik anyway as he was the original author and he knows the idea behind it. tl;dr; you'd have to check the history of this code in the git

TheMMaciek: Yeah I think we should overwrite the fields whenever the data application implementation is not defined, so in this match:
```
maybeDataApplication match {
  case Some(_) => creationResult
  case None =>
}
```

The currency messages aren't something we can't validate so we shouldn't overwrite those I think.

TheMMaciek (approved):
When we re-add fee transactions validation later I think they should be overwritten along with `dataApplication` in case the implementation is not provided as they result from data application implementation.

PR #873, opened 2024-05-29
Snapshot fees functionality fixes

TheMMaciek:
I'm still going through some test cases so I'm creating this as a draft PR. Although I think it's mostly ready so please review.

This PR:
- improves currency message validation
- forces a synchronous processing of messages which is good enough for their intended functionality (either first or next message, no skipping or sending in advance)
- sets the last messages state in currency snapshot info correctly
- improves message selection based on deterministic order and validation during CL0 consensus


modules/shared/src/main/scala/org/tessellation/security/signature/SignedValidator.scala
```diff
@@ -79,6 +84,17 @@ object SignedValidator {
...
+    def isSignedBy[A: Encoder](
+      signed: Signed[A],
+      signerAddress: Address
+    ): F[SignedValidationErrorOr[Signed[A]]] =
+      signed.proofs.existsM { proof =>
+        proof.id.toAddress.map(_ === signerAddress)
+      }.ifM(
+        signed.validNec[SignedValidationError].pure[F],
+        Validated.invalidNec[SignedValidationError, Signed[A]](NotSignedByAddressOwner).pure[F]
+      )
```
sdavidp, line 96: If we use `.ifF` instead then we don't need to lift the results with `pure`.
Or we can simplify it a bit further by doing this?
```suggestion
      signed.proofs
        .existsM(_.id.toAddress.map(_ === signerAddress))
        .map(Validated.condNec(_, signed, NotSignedByAddressOwner))
```

TheMMaciek: Yeah the less code the better. It makes sense to use `ifF` here.

TheMMaciek: Yeah it's simpler I will use that, thanks.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencyMessageValidator.scala
```diff
@@ -0,0 +1,113 @@
...
+            .map(_.flatMap { ownerId =>
+              NonEmptySet
+                .fromSet(message.proofs.filter(_.id =!= ownerId))
+                .map(sigs => Signed(message.value, sigs))
+            })
+            .map {
+              case Some(noOwnerSigMsg) =>
+                validator
+                  .validateSignaturesWithSeedlist(seedlistPeers, noOwnerSigMsg)
+                  .map(_ => message)
```
sdavidp, line 66:
```suggestion
                  .as(message)
```

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencyMessageValidator.scala
```diff
@@ -0,0 +1,113 @@
...
+
+        lastMessages.get(message.messageType) match {
+          case Some(lastMessage) if message.parentOrdinal =!= lastMessage.ordinal =>
+            Async[F].pure(NotANextMessage.invalidNec)
+          case None if message.parentOrdinal =!= MessageOrdinal.MinValue =>
+            Async[F].pure(FirstMessageWithWrongOrdinal.invalidNec)
+          case _ if !isMetagraphIdValid =>
+            Async[F].pure(WrongMetagraphId.invalidNec)
+          case _ =>
+            validateSignatures(message).map(mapError)
```
sdavidp, line 92:
```suggestion
            validateSignatures(message)
```
... if we put the `leftMap` up above

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencyMessageValidator.scala
```diff
@@ -0,0 +1,113 @@
...
+            hasNoDuplicates = validator.validateUniqueSigners(message)
+            isSignedCorrectly <- validator.validateSignatures(message)
+            isSignedByMajority = validator.validateSignedBySeedlistMajority(allowancePeers, message)
+            isSignedBySeedlistPeers <- validateIfSignedBySeedlistPeersExcludingOwner(message)
+          } yield
+            isSignedByOwner
+              .productR(hasNoDuplicates)
+              .productR(isSignedCorrectly)
+              .productR(isSignedByMajority)
+              .productR(isSignedBySeedlistPeers)
```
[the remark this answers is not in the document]
TheMMaciek: I will check it

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/http/routes/CurrencyMessageRoutes.scala
sdavidp, line 1: I feel like this class deserves a test suite.
Also, am I understanding correctly that if there are no messages in snapshot storage then we return `ServiceUnavailable` ?

If so, who/what stores the first message?

TheMMaciek: There is no storage of messages. The messages are stored in the currency snapshot info. If there is no message in the last known snapshot info then we require the sent message to be a first one. The parallelity is reduced in that scenario where we require to send the exactly next one. But for the intended functionality of currency messages there is no need in my opinion to accept multiple chained messages designated to be put in the same snapshot.

TheMMaciek: We return `ServiceUnavailable` when there is not first snapshot in the storage yet. I.e. the node process is obviously not ready yet to accept messages.

TheMMaciek: I think unit tests are needed more in the case of now evolved validator, rather than this routes class. The minute I finish integration tests I think I will implement unit tests for validator as well (to safeguard the implementation for the future).

sdavidp (approved):

kpudlik (approved):

PR #875, opened 2024-06-06
Subtract metagraph fee from GL0 balances

TheMMaciek:


modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
```diff
@@ -73,14 +74,16 @@ object GlobalSnapshotAcceptanceManager {
...
+          stateChannelEventsProcessor
+            .process(
+              ordinal,
+              lastSnapshotContext.copy(balances = lastSnapshotContext.balances ++ acceptanceResult.contextUpdate.balances),
+              scEvents,
+              validationType
+            )
         sCSnapshotHashes <- scSnapshots.toList.traverse {
-          case (address, nel) => nel.head.toHashed.map(address -> _.hash)
+          case (address, nel) => nel.last.toHashed.map(address -> _.hash)
```
TheMMaciek, line 86: I know it's generally non-performant to get the last element of a list like that. I tried changing the collection to be NonEmptyChain or Chain, but in the end decided that it makes no sense at this moment. We start the processing with a list and we need to end with a list (because of schema), if we transform to NEC and back from NEC to have the correct type we may add more boilerplate and end up with the same complexity - we can always change it in the future if decide it makes sense. For the most cases it won't matter much as we won't work with long lists here - 1, 2 elements is probably what we have in those.

sdavidp: @TheMMaciek I think this is a good case for a comment so that future readers won't question it.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -89,16 +87,33 @@ object GlobalSnapshotStateChannelEventsProcessor {
           .flatMap { case (_, validatedEvents) => processStateChannelEvents(snapshotOrdinal, lastGlobalSnapshotInfo, validatedEvents) }
           .flatMap {
             case (scSnapshots, returnedSCEvents) =>
-              processCurrencySnapshots(lastGlobalSnapshotInfo, scSnapshots)
-                .map(calculateLastCurrencySnapshots(_, lastGlobalSnapshotInfo))
-                .map((scSnapshots, _, returnedSCEvents))
+              processCurrencySnapshots(snapshotOrdinal, lastGlobalSnapshotInfo, scSnapshots).map { accepted =>
```
TheMMaciek, line 90: `process` and `processCurrencySnapshots` functions are were the core changes happen - please review thoroughly

TheMMaciek: We know do final acceptance of binary and currency snapshot in the same function whereas previously we were accepting binaries separately. Accepting binaries separately wasn't correct in the first place - if a binary is a currency snapshot that can't be accepted why would we accept it as a binary - but currently we can also only accept custom binaries (non currency) if the fee is not required as we don't have a way to fetch a fee deduction address from custom binary.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -89,16 +87,33 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
+          (NonEmptyList[(Signed[StateChannelSnapshotBinary], Option[CurrencySnapshotWithState])], Map[Address, Balance])
+        ],
         lastGlobalSnapshotInfo: GlobalSnapshotInfo
       ): SortedMap[Address, CurrencySnapshotWithState] = {
-        val lastCurrencySnapshots = processedCurrencySnapshots.map { case (k, v) => k -> v.last }
+        val lastCurrencySnapshots = processedCurrencySnapshots.map {
+          case (k, (v, _)) =>
+            val lastState = v.collect { case (_, Some(state)) => state }.lastOption
+            k -> lastState
+        }.collect { case (key, Some(state)) => key -> state }
```
sdavidp, line 116: I think we can simplify the first `.map` ?
```suggestion
        val lastCurrencySnapshots = processedCurrencySnapshots
          .map { case (k, (v, _)) => k -> v.toList.flatMap(_._2).lastOption }
          .collect { case (key, Some(state)) => key -> state }
```

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -89,16 +87,33 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
           .flatMap {
             case (scSnapshots, returnedSCEvents) =>
-              processCurrencySnapshots(lastGlobalSnapshotInfo, scSnapshots)
-                .map(calculateLastCurrencySnapshots(_, lastGlobalSnapshotInfo))
-                .map((scSnapshots, _, returnedSCEvents))
+              processCurrencySnapshots(snapshotOrdinal, lastGlobalSnapshotInfo, scSnapshots).map { accepted =>
+                val lastCurrencyStates = calculateLastCurrencySnapshots(accepted, lastGlobalSnapshotInfo)
+                val finalScSnapshots = accepted.map { case (k, (v, _)) => k -> v.map(_._1) }
+                // TODO: ASSUMING that owner addresses are restricted from being shared at this point
+                val balanceUpdates = accepted.values.map(_._2).foldLeft(Map.empty[Address, Balance]) { case (agg, update) => agg ++ update }
```
sdavidp, line 94:
```suggestion
                val balanceUpdates = accepted.values.map(_._2).foldLeft(Map.empty[Address, Balance])(_ ++ _)
```

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -17,45 +17,49 @@ import scala.collection.immutable.SortedMap
...
+  )(implicit hasher: Hasher[F]): F[StateChannelAcceptanceResult]
 
   def processCurrencySnapshots(
+    snapshotOrdinal: SnapshotOrdinal,
     lastGlobalSnapshotInfo: GlobalSnapshotInfo,
     events: SortedMap[Address, NonEmptyList[Signed[StateChannelSnapshotBinary]]]
-  )(implicit hasher: Hasher[F]): F[SortedMap[Address, NonEmptyList[CurrencySnapshotWithState]]]
+  )(
+    implicit hasher: Hasher[F]
+  ): F[(SortedMap[Address, (NonEmptyList[(Signed[StateChannelSnapshotBinary], Option[CurrencySnapshotWithState])], Map[Address, Balance])])]
```
sdavidp, line 53: (unnecessary outer parentheses)
The value of the map is a complicated tuple. I feel like it deserves to be defined as a case class. It will help make the code a lot easier to follow.

sdavidp: (I forgot to mention that it occurs 5 times in the file)

TheMMaciek: Introduced type descriptions. Didn't add case classes to not introduce boilerplate.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
TheMMaciek: Yeah that workks, thanks.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
TheMMaciek: It is used in multiple places, not sure if explaining it in one makes sense. Wouldn't like to pollute code with comments as well if they are not strictly neccessary.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -114,74 +128,125 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
-                      .map {
-                        _.toOption.map { snapshot =>
-                          (NonEmptyList.one(snapshot.asLeft).some, tail)
-                            .asLeft[Result]
-                        }
+      )(implicit hasher: Hasher[F]): F[SortedMap[Address, MetagraphAcceptanceResult]] = {
+        val isFeeRequired = feeCalculator.isFeeRequired(snapshotOrdinal)
+
+        def deserialize[A: Decoder](binary: Signed[StateChannelSnapshotBinary]): F[Option[A]] =
+          jsonBrotliBinarySerializer.deserialize[A](binary.value.content).map(_.toOption)
```
sdavidp, line 138: I see doing `toOption` is the same as before but do you know why we are throwing away the error?

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -114,74 +128,125 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
+                      else ((NonEmptyList.one((head, none)), emptyBalanceUpdate).some, tail).asLeft
+                  }
+
+                case (current @ Some((nel, balanceUpdate)), head :: tail) =>
+                  nel.head match {
+                    case (_, None) =>
+                      deserialize[Signed[CurrencySnapshot]](head).map {
+                        case Some(snapshot) => // full snapshot - we don't subtract fee
+                          (
+                            (nel.prepend((head, snapshot.asLeft.some)), balanceUpdate).some,
```
sdavidp, line 180: Is there a reason for using `prepend` in this class instead of `::` ?

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -114,74 +128,125 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
+                            tail
+                          ).asLeft
+                        case None => // no full snapshot yet - we only accept the binary if fee is not required
+                          if (isFeeRequired) current.asRight
+                          else ((nel.prepend((head, none)), balanceUpdate).some, tail).asLeft
```
sdavidp, line 185:
```suggestion
                          Either.cond(
                            isFeeRequired,
                            current,
                            (((head, none) :: nel, balanceUpdate).some, tail)
```

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -114,74 +128,125 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
+                          (
+                            (
+                              nel.prepend((head, (snapshot, fullSnapshot.value.info.toCurrencySnapshotInfo).asRight.some)),
+                              balanceUpdate
+                            ).some,
+                            tail
+                          ).asLeft
+                        case None => // no first incremental yet - we only accept the binary if fee is not required
+                          if (isFeeRequired) current.asRight
+                          else ((nel.prepend((head, lastCurrState)), balanceUpdate).some, tail).asLeft
```
sdavidp, line 200:
```suggestion
                          Either.cond(isFeeRequired, current, (((head, lastCurrState) :: nel, balanceUpdate).some, tail))
```

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
TheMMaciek: Because we don't know if the deserialization is going to work - it may or may not be a `CurrencySnapshot`/`CurrencyIncrementalSnapshot`

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
TheMMaciek: There is no difference between `prepend` and `::` right? Here anyway I think it's more readable to use `.prepend()` given that added object is more complex.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
TheMMaciek: I think I like the current form. And it's also correctly type inferring right now, for different syntaxes the type inference sometimes doesn't work.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
TheMMaciek: So deserializing error is not unexpected in any way.

PR #876, opened 2024-06-07
Fix currency message acceptance for historical flow

TheMMaciek:


modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
```diff
@@ -126,7 +126,11 @@ object CurrencySnapshotAcceptanceManager {
         messagesToReject
       )
 
-      csi = CurrencySnapshotInfo(transactionsRefs, updatedBalancesByRewards, messagesForContextUpdate.some)
+      csi = CurrencySnapshotInfo(
+        transactionsRefs,
+        updatedBalancesByRewards,
+        Option.when(messagesForContextUpdate.nonEmpty)(messagesForContextUpdate)
```
TheMMaciek, line 132: The state of currency messages will appear for the first time when actual messages are processed - SortedMap non empty. From that point on it will always be Some.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotCreator.scala
```diff
@@ -182,7 +182,7 @@ object CurrencySnapshotCreator {
             stateProof,
             currentEpochProgress,
             dataApplicationAcceptanceResult.map(_.dataApplicationPart),
-            messagesAcceptanceResult.accepted.toSortedSet.some,
+            Option.when(snapshotInfo.lastMessages.nonEmpty)(messagesAcceptanceResult.accepted.toSortedSet),
```
TheMMaciek, line 185: Start adding messages to the CurrencyIncrementalSnapshot the moment currencyMessages state is non empty. It will also set the list as empty after first message appears.

sdavidp (approved):

PR #877, opened 2024-06-11
Extract kilobytes size calculation

kpudlik:


TheMMaciek (approved):

IPadawans (approved):

modules/shared/src/main/scala/org/tessellation/json/SizeCalculator.scala
```diff
@@ -0,0 +1,24 @@
...
+
+object SizeCalculator {
+  def bytes[F[_]: Functor: JsonSerializer, A: Encoder](a: A): F[NonNegInt] =
+    JsonSerializer[F].serialize(a).map { binary =>
+      NonNegInt.unsafeFrom(binary.size)
+    }
+
+  def kilobytes[F[_]: Functor: JsonSerializer, A: Encoder](a: A): F[NonNegInt] = bytes(a).map(_.toLong).map(toKilobytes)
+
+  def toKilobytes(bytesSize: Long): NonNegInt =
```
sdavidp, line 18: I don't understand why we're going from an `Int` to a `Long` and then back down to an `Int`

kpudlik: Previously the code was duplicated:

Place 1 where `bytesSize` is `NonNegInt`:
```scala
    NonNegInt.unsafeFrom(
      (BigDecimal(bytesSize) / BigDecimal(1024))
        .setScale(0, BigDecimal.RoundingMode.UP)
        .toInt
    )
```

Place 2 where `noSigsBytesSize + signatureCount * snapshotSizeConfig.singleSignatureSizeInBytes` is `Long`:
```scala
    NonNegInt.unsafeFrom(
      (BigDecimal(noSigsBytesSize + signatureCount * snapshotSizeConfig.singleSignatureSizeInBytes) / BigDecimal(1024))
        .setScale(0, BigDecimal.RoundingMode.UP)
        .toInt
    )
```

I unified it to a shared function but to make it happy I allowed it to pass `Long` as well because this is how it worked before `Long -> BigDecimal -> NonNegLong`.

We can actually get rid of `NonNegInt` and use `NonNegLong` everywhere. WDYT @sdavidp @TheMMaciek?

sdavidp: Max positive Int is about 2GB -- are we expecting anything bigger than that in a single event/snapshot? If so then sure, we can go to Long, otherwise Int would be sufficient.

sdavidp: Or another way to think of this is if we think this size calculator will be used in other places where the thing might be bigger than 2GB then sure, going to Long would be safer.

PR #880, opened 2024-06-18
Setting global snapshot after running currency rollback

IPadawans:
### Problem
When we rollback an existing metagraph, and then send the snapshot fee `owner` message, the first global snapshot is set to 0 because wasn't being set before the first consensus round.
You can check the image below with some custom logs I've added to debug
Captura de Tela 2024-06-18 às 13 13 36

### Solution
I'm filling the lastGlobalSnapshot when running as rollback, I've already tested and it's working as expected

### NOTE
This PR was already created for the hotfix branch so we can deploy it as soon as we fix it. The DOR Metagraph in Testnet are already fixed because I've generated a custom version locally to unlock

### DIFF
--- a/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
+++ b/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/CurrencyL0App.scala
@@ -165,6 +165,8 @@ abstract class CurrencyL0App(
 
         case other =>
           for {
+            _ <- StateChannel.performGlobalL0PeerDiscovery[IO](storages, programs)
+
             innerProgram <- other match {
               case rv: RunValidator =>
                 storages.identifier.setInitial(rv.identifier) >>
--- a/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/StateChannel.scala
+++ b/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/StateChannel.scala
@@ -70,12 +70,7 @@ object StateChannel {
       Stream
         .awakeEvery[F](awakePeriod)
         .evalMap { _ =>
-          storages.lastGlobalSnapshot.get.flatMap {
-            case None =>
-              storages.globalL0Cluster.getRandomPeer.flatMap(p => programs.globalL0PeerDiscovery.discoverFrom(p))
-            case Some(latestSnapshot) =>
-              programs.globalL0PeerDiscovery.discover(latestSnapshot.signed.proofs.map(_.id).map(PeerId._Id.reverseGet))
-          }
+          performGlobalL0PeerDiscovery(storages, programs)
         }
         .handleErrorWith { error =>
           Stream.eval(logger.error(error)("Error during global L0 peer discovery"))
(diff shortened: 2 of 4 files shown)


marcinwadon:
I think setting an arbitrary global snapshot as initial may be misleading and error-prone sometimes. IMO we should wait until the node is fully Ready (afaik it goes into Observing state and then to Ready after the first consensus round, correct me if I'm wrong) - so it should not send the snapshot fee `owner` message before the first consensus round. We need to wait to fully initialize the node. By setting the global snapshot based on consensus, we make sure we have the correct one set in the storage. If there would be a gap (let's say you fetched ordinal=3 but in a meantime, the consensus for ordinal=4 is finishing and node would join consensus at ordinal=5 -> then, it will fail setting the snapshot with ordinal=5 as there would be a gap in the storage).

IPadawans:
> I think setting an arbitrary global snapshot as initial may be misleading and error-prone sometimes. IMO we should wait until the node is fully Ready (afaik it goes into Observing state and then to Ready after the first consensus round, correct me if I'm wrong) - so it should not send the snapshot fee `owner` message before the first consensus round. We need to wait to fully initialize the node. By setting the global snapshot based on consensus, we make sure we have the correct one set in the storage. If there would be a gap (let's say you fetched ordinal=3 but in a meantime, the consensus for ordinal=4 is finishing and node would join consensus at ordinal=5 -> then, it will fail setting the snapshot with ordinal=5 as there would be a gap in the storage).

@marcinwadon AFAIK the message should be sent before the first consensus round, otherwise we will create an invalid snapshot. I've made this test around this branch in production with DOR Metagraph in testnet and it worked. I agree with you've proposed, however, I think it will be a more complex solution. Don't you think we can go with this one? I

IPadawans:
> I think setting an arbitrary global snapshot as initial may be misleading and error-prone sometimes. IMO we should wait until the node is fully Ready (afaik it goes into Observing state and then to Ready after the first consensus round, correct me if I'm wrong) - so it should not send the snapshot fee `owner` message before the first consensus round. We need to wait to fully initialize the node. By setting the global snapshot based on consensus, we make sure we have the correct one set in the storage. If there would be a gap (let's say you fetched ordinal=3 but in a meantime, the consensus for ordinal=4 is finishing and node would join consensus at ordinal=5 -> then, it will fail setting the snapshot with ordinal=5 as there would be a gap in the storage).

marcinwadon:
@IPadawans 
>  I think this solution is not setting an arbitrary snapshot.
It is an arbitrary snapshot since you fetch "some" latest snapshot from "some" random peer.

sdavidp:
@IPadawans @marcinwadon @ryle-goehausen 

I need to dig a lot more but right now I'm not convinced this proposed solution is safe. Intuitively it feels like we're bypassing the normal flow.

IPadawans:
> @IPadawans
> 
> > I think this solution is not setting an arbitrary snapshot.
> 
> It is an arbitrary snapshot since you fetch "some" latest snapshot from "some" random peer.

yeah @marcinwadon you're right sorry. The point here is that when the first consensus round starts, we will already produce the snapshot, right? If we produce the snapshot without the owner message I think it will be rejected and the chain will be break. I can be wrong, so please correct me if I'm wrong. @sdavidp do you know if the behavior I've told above is right?

sdavidp:
> If we produce the snapshot without the owner message I think it will be rejected and the chain will be break. I can be wrong, so please correct me if I'm wrong. @sdavidp do you know if the behavior I've told above is right?

Why do you say the chain will break without owner message?

IPadawans:
> > If we produce the snapshot without the owner message I think it will be rejected and the chain will be break. I can be wrong, so please correct me if I'm wrong. @sdavidp do you know if the behavior I've told above is right?
> 
> Why do you say the chain will break without owner message?

Because the global will reject the snapshot without the owner

TheMMaciek:
There are two cases here:
1) When the global ordinal is not set correctly on CL0 and defaults to 0 the wrong snapshot fee config will be loaded. That will probably be no fee config which will result in not setting any fee which will result in rejection on GL0.
2) If the owner message is not sent along with the first currency snapshot after rollback when fees are already active on GL0, then even if the fee is calculated correctly there will be no address to subtract that fee from so again the snapshot will be rejected on GL0.

marcinwadon (approved):

PR #884, opened 2024-06-26
Fixing collateral check in currency modules

IPadawans:
### Changes
+ Fixing the collateral check in currency modules to be independent of the global checker.
+ Currently, in the global layer mainnet, it's required to have 250K Tokens (DAG for the global layer and L0 Token for the currency layer). So, I've fixed this to use the CL_COLLATERAL parameter or be 0 if not provided (ONLY IN CURRENCY LAYER, THE GLOBAL LAYER KEEP WITH THE SAME BEHAVIOR)

### DIFF
--- a/modules/node-shared/src/main/scala/org/tessellation/node/shared/app/TessellationIOApp.scala
+++ b/modules/node-shared/src/main/scala/org/tessellation/node/shared/app/TessellationIOApp.scala
@@ -106,6 +106,7 @@ abstract class TessellationIOApp[A <: CliMethod](
                 LoggerConfigurator.configureLogger[IO](cfg.environment) >>
                 logger.info(s"App environment: ${cfg.environment}") >>
                 logger.info(s"App version: ${version.show}") >>
+                logger.info(s"App collateral: ${cfg.collateral.amount.show}") >>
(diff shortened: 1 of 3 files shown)


modules/node-shared/src/main/scala/org/tessellation/node/shared/app/TessellationIOApp.scala
```diff
@@ -106,6 +106,7 @@ abstract class TessellationIOApp[A <: CliMethod](
                 LoggerConfigurator.configureLogger[IO](cfg.environment) >>
                 logger.info(s"App environment: ${cfg.environment}") >>
                 logger.info(s"App version: ${version.show}") >>
+                logger.info(s"App collateral: ${cfg.collateral.amount.value.value}") >>
```
marcinwadon, line 109: `cfg.collateral.amount.show`

IPadawans: Updated

marcinwadon (approved):

kpudlik (approved):

PR #890, opened 2024-07-08
Adding validations to reject repeated owner or staking address

IPadawans:
### Changes
+ Added new storage to store all metagraphs and fee addresses in both layers: Global and Metagraph.
+ Implemented validation in StateChannel validations to reject snapshots with existing addresses.
+ Updated the validation function to accept not only staking balances but also addresses.
+ Added new validation in StateChannel validation to reject duplicate entries.
+ Updated the current tests to ensure compatibility with the updates.
+ Added validation on the endpoint to reject existing wallets. This provides an additional layer of validation, rejecting messages of existing wallets directly at the endpoint, resulting in two layers of validation.
+ Added code to populate existing metagraphs and addresses based on the lastGlobalSnapshot, and to fill values as snapshots are accepted.
+ We decided to not change the schema to add the addresses OWNER and STAKING, so we are deserializing the binary to read those information

### Tests
I've made tests in both scenarios, genesis and rollback in global layer

### DIFF
--- a/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/http/routes/CurrencyMessageRoutes.scala
+++ b/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/http/routes/CurrencyMessageRoutes.scala
@@ -9,12 +9,15 @@ import scala.collection.immutable.SortedMap
 import org.tessellation.currency.l0.node.IdentifierStorage
 import org.tessellation.currency.schema.currency.{CurrencyIncrementalSnapshot, CurrencySnapshotInfo}
 import org.tessellation.kernel._
-import org.tessellation.node.shared.domain.snapshot.storage.SnapshotStorage
+import org.tessellation.node.shared.domain.snapshot.storage.{LastSnapshotStorage, SnapshotStorage}
+import org.tessellation.node.shared.domain.statechannel.StateChannelValidator.getFeeAddresses
 import org.tessellation.node.shared.infrastructure.snapshot.CurrencyMessageValidator
 import org.tessellation.node.shared.snapshot.currency.{CurrencyMessageEvent, CurrencySnapshotEvent}
 import org.tessellation.routes.internal._
+import org.tessellation.schema.address.Address
 import org.tessellation.schema.currencyMessage.{CurrencyMessage, MessageType}
 import org.tessellation.schema.http.{ErrorCause, ErrorResponse}
+import org.tessellation.schema.{GlobalIncrementalSnapshot, GlobalSnapshotInfo}
 import org.tessellation.security.Hasher
 import org.tessellation.security.signature.Signed
 
@@ -31,7 +34,8 @@ class CurrencyMessageRoutes[F[_]: Async: Hasher](
   mkCell: CurrencySnapshotEvent => Cell[F, StackF, _, Either[CellError, Ω], _],
   validator: CurrencyMessageValidator[F],
   snapshotStorage: SnapshotStorage[F, CurrencyIncrementalSnapshot, CurrencySnapshotInfo],
-  identifierStorage: IdentifierStorage[F]
+  identifierStorage: IdentifierStorage[F],
+  lastGlobalSnapshotStorage: LastSnapshotStorage[F, GlobalIncrementalSnapshot, GlobalSnapshotInfo]
 ) extends Http4sDsl[F]
     with PublicRoutes[F] {
 
@@ -47,7 +51,13 @@ class CurrencyMessageRoutes[F[_]: Async: Hasher](
           _.map(_._2).map(_.lastMessages.getOrElse(SortedMap.empty[MessageType, Signed[CurrencyMessage]]))
         }
         metagraphId <- identifierStorage.get
-        maybeResult <- maybeLastMsgs.traverse(validator.validate(msg, _, metagraphId))
+        allFeesAddresses <- lastGlobalSnapshotStorage.getCombined.map {
+          case Some((_, info)) =>
+            getFeeAddresses(info)
+          case None => SortedMap.empty[Address, Set[Address]]
+        }
+
+        maybeResult <- maybeLastMsgs.traverse(validator.validate(msg, _, metagraphId, allFeesAddresses))
         response <- maybeResult match {
           case Some(Invalid(errors)) =>
             logger
--- a/modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/statechannel/StateChannelService.scala
+++ b/modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/statechannel/StateChannelService.scala
@@ -6,9 +6,10 @@ import cats.effect.kernel.Async
 import cats.syntax.all._
 
 import org.tessellation.dag.l0.domain.cell.{L0Cell, L0CellInput}
-import org.tessellation.node.shared.domain.statechannel.StateChannelValidator
-import org.tessellation.node.shared.domain.statechannel.StateChannelValidator.StateChannelValidationError
-import org.tessellation.schema.currencyMessage.fetchStakingBalance
+import org.tessellation.node.shared.domain.statechannel.StateChannelValidator.{StateChannelValidationError, getFeeAddresses}
+import org.tessellation.node.shared.domain.statechannel.{SnapshotFeesInfo, StateChannelValidator}
+import org.tessellation.schema.address.Address
+import org.tessellation.schema.currencyMessage.{fetchMetagraphFeesAddresses, fetchStakingBalance}
 import org.tessellation.schema.{GlobalIncrementalSnapshot, GlobalSnapshotInfo}
 import org.tessellation.security.Hasher
 import org.tessellation.security.signature.Signed
@@ -22,7 +23,10 @@ trait StateChannelService[F[_]] {
 
 object StateChannelService {
 
-  def make[F[_]: Async](mkDagCell: L0Cell.Mk[F], stateChannelValidator: StateChannelValidator[F]): StateChannelService[F] =
+  def make[F[_]: Async](
+    mkDagCell: L0Cell.Mk[F],
+    stateChannelValidator: StateChannelValidator[F]
+  ): StateChannelService[F] =
     new StateChannelService[F] {
 
       def process(
@@ -31,15 +35,20 @@ object StateChannelService {
       )(implicit hasher: Hasher[F]): F[Either[NonEmptyList[StateChannelValidationError], Unit]] = {
         val (snapshot, state) = globalSnapshotAndState
         val staked = fetchStakingBalance(stateChannelOutput.address, state)
+        val (ownerAddress, stakingAddress) = fetchMetagraphFeesAddresses(stateChannelOutput.address, state)
+        val allFeesAddresses: Map[Address, Set[Address]] = getFeeAddresses(state)
+        val snapshotFeesInfo = SnapshotFeesInfo(allFeesAddresses, staked, ownerAddress, stakingAddress)
 
-        stateChannelValidator.validate(stateChannelOutput, snapshot.ordinal, staked).flatMap {
-          case Valid(_) =>
-            mkDagCell(L0CellInput.HandleStateChannelSnapshot(stateChannelOutput))
-              .run()
-              .as(().asRight[NonEmptyList[StateChannelValidationError]])
-          case Invalid(errors) => errors.toNonEmptyList.asLeft[Unit].pure[F]
-        }
+        for {
+          validations <- stateChannelValidator.validate(stateChannelOutput, snapshot.ordinal, snapshotFeesInfo)
+          result <- validations match {
+            case Valid(_) =>
+              mkDagCell(L0CellInput.HandleStateChannelSnapshot(stateChannelOutput))
+                .run()
+                .as(().asRight[NonEmptyList[StateChannelValidationError]])
+            case Invalid(errors) => errors.toNonEmptyList.asLeft[Unit].pure[F]
+          }
+        } yield result
       }
-
     }
 }
--- a/modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/statechannel/FeeCalculator.scala
+++ b/modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/statechannel/FeeCalculator.scala
@@ -10,6 +10,7 @@ import scala.collection.immutable.SortedMap
 
 import org.tessellation.currency.schema.currency.SnapshotFee
 import org.tessellation.schema.SnapshotOrdinal
+import org.tessellation.schema.address.Address
 import org.tessellation.schema.balance.Balance
 
 import derevo.cats.eqv
@@ -27,6 +28,17 @@ case class FeeCalculatorConfig(
   computationalCost: NonNegLong,
   proWeight: NonNegBigDecimal
 )
+
+case class SnapshotFeesInfo(
+  allFeesAddresses: Map[Address, Set[Address]],
+  stakingBalance: Balance,
+  ownerAddress: Option[Address],
+  stakingAddress: Option[Address]
+)
+
+object SnapshotFeesInfo {
+  def empty: SnapshotFeesInfo = SnapshotFeesInfo(Map.empty, Balance.empty, none, none)
+}
 
 object FeeCalculatorConfig {
   val noFee: FeeCalculatorConfig =
--- a/modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/statechannel/StateChannelValidator.scala
+++ b/modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/statechannel/StateChannelValidator.scala
@@ -13,10 +13,10 @@ import org.tessellation.ext.cats.syntax.validated._
 import org.tessellation.json.{JsonSerializer, SizeCalculator}
 import org.tessellation.node.shared.domain.seedlist.SeedlistEntry
 import org.tessellation.node.shared.domain.statechannel.StateChannelValidator.StateChannelValidationErrorOr
-import org.tessellation.schema.SnapshotOrdinal
 import org.tessellation.schema.address.Address
 import org.tessellation.schema.balance.Balance
 import org.tessellation.schema.peer.PeerId
+import org.tessellation.schema.{GlobalSnapshotInfo, SnapshotOrdinal}
 import org.tessellation.security.Hasher
 import org.tessellation.security.hash.Hash
 import org.tessellation.security.signature.SignedValidator.SignedValidationError
@@ -34,19 +34,43 @@ trait StateChannelValidator[F[_]] {
   def validate(
     stateChannelOutput: StateChannelOutput,
     snapshotOrdinal: SnapshotOrdinal,
-    staked: Balance
+    snapshotFeesInfo: SnapshotFeesInfo
   )(implicit hasher: Hasher[F]): F[StateChannelValidationErrorOr[StateChannelOutput]]
+
   def validateHistorical(
     stateChannelOutput: StateChannelOutput,
     snapshotOrdinal: SnapshotOrdinal,
-    staked: Balance
+    snapshotFeesInfo: SnapshotFeesInfo
   )(
     implicit hasher: Hasher[F]
   ): F[StateChannelValidationErrorOr[StateChannelOutput]]
-
 }
 
 object StateChannelValidator {
+  def getFeeAddresses(info: GlobalSnapshotInfo): Map[Address, Set[Address]] =
+    info.lastCurrencySnapshots.collect {
+      case (address, Right((_, info))) => address -> info.lastMessages.toList.flatMap(_.values.map(_.address).toList).toSet
+    }
+
+  def validateIfAddressAlreadyUsed(
+    metagraphId: Address,
+    existingFeesAddresses: Map[Address, Set[Address]],
+    maybeAddress: Option[Address]
+  ): StateChannelValidationErrorOr[Unit] = maybeAddress match {
+    case None => ().validNec
+    case Some(address) =>
+      val existsInMetagraphAddresses = existingFeesAddresses.get(metagraphId).exists(_.contains(address))
+      if (existsInMetagraphAddresses) {
+        ().validNec
+      } else {
+        val existsInAllAddresses = existingFeesAddresses.exists { case (_, addresses) => addresses.contains(address) }
+        if (existsInAllAddresses) {
+          AddressAlreadyInUse.invalidNec
+        } else {
+          ().validNec
+        }
+      }
+  }
 
   def make[F[_]: Async: JsonSerializer](
     signedValidator: SignedValidator[F],
@@ -59,29 +83,45 @@ object StateChannelValidator {
     def validate(
       stateChannelOutput: StateChannelOutput,
       globalOrdinal: SnapshotOrdinal,
-      staked: Balance
+      snapshotFeesInfo: SnapshotFeesInfo
     )(implicit hasher: Hasher[F]): F[StateChannelValidationErrorOr[StateChannelOutput]] =
-      validateHistorical(stateChannelOutput, globalOrdinal, staked).map(
-        _.product(validateAllowedSignatures(stateChannelOutput)).as(stateChannelOutput)
-      )
+      for {
+        historical <- validateHistorical(stateChannelOutput, globalOrdinal, snapshotFeesInfo)
+        allowedSignatures = validateAllowedSignatures(stateChannelOutput)
+      } yield
+        historical
+          .product(allowedSignatures)
+          .as(stateChannelOutput)
 
     def validateHistorical(
       stateChannelOutput: StateChannelOutput,
       globalOrdinal: SnapshotOrdinal,
-      staked: Balance
+      snapshotFeesInfo: SnapshotFeesInfo
     )(implicit hasher: Hasher[F]): F[StateChannelValidationErrorOr[StateChannelOutput]] =
       for {
         signaturesV <- signedValidator
           .validateSignatures(stateChannelOutput.snapshotBinary)
           .map(_.errorMap[StateChannelValidationError](InvalidSigned))
         snapshotSizeV <- validateSnapshotSize(stateChannelOutput.snapshotBinary)
-        snapshotFeeV <- validateSnapshotFee(stateChannelOutput.snapshotBinary, globalOrdinal, staked)
+        snapshotFeeV <- validateSnapshotFee(stateChannelOutput.snapshotBinary, globalOrdinal, snapshotFeesInfo.stakingBalance)
         genesisAddressV = validateStateChannelGenesisAddress(stateChannelOutput.address, stateChannelOutput.snapshotBinary)
+        ownerAddressValidationV = validateIfAddressAlreadyUsed(
+          stateChannelOutput.address,
+          snapshotFeesInfo.allFeesAddresses,
+          snapshotFeesInfo.ownerAddress
+        )
+        stakingAddressValidationV = validateIfAddressAlreadyUsed(
+          stateChannelOutput.address,
+          snapshotFeesInfo.allFeesAddresses,
+          snapshotFeesInfo.stakingAddress
+        )
       } yield
         signaturesV
           .product(snapshotSizeV)
           .product(snapshotFeeV)
           .product(genesisAddressV)
+          .product(ownerAddressValidationV)
+          .product(stakingAddressValidationV)
           .as(stateChannelOutput)
 
     private def validateSnapshotSize(
@@ -157,7 +197,6 @@ object StateChannelValidator {
         StateChannellGenesisAddressNotGeneratedFromData(address).invalidNec
       else
         signedSC.validNec
-
   }
 
   @derive(eqv, show, decoder, encoder)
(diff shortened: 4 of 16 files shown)


modules/currency-l0/src/main/scala/org/tessellation/currency/l0/http/routes/CurrencyMessageRoutes.scala
```diff
@@ -31,7 +31,8 @@ class CurrencyMessageRoutes[F[_]: Async: Hasher](
   mkCell: CurrencySnapshotEvent => Cell[F, StackF, _, Either[CellError, Ω], _],
   validator: CurrencyMessageValidator[F],
   snapshotStorage: SnapshotStorage[F, CurrencyIncrementalSnapshot, CurrencySnapshotInfo],
-  identifierStorage: IdentifierStorage[F]
+  identifierStorage: IdentifierStorage[F],
+  feesAddressesStorage: FeesAddressesStorage[F]
```
marcinwadon, line 35: I'd be more generic for the storage name. `AddressStorage` or `FeeStorage` 🤔 
`FeesAddresses` is very very specific

modules/node-shared/src/main/resources/application.conf
```diff
@@ -75,7 +75,7 @@ fee-configs {
     }
   }
   testnet: {
-    1943000: {
+    10: {
```
marcinwadon, line 78: This should not be modified

modules/node-shared/src/main/resources/application.conf
```diff
@@ -91,9 +91,9 @@ fee-configs {
     }
   }
   dev: {
-    0: {
-      base-fee: 0,
-      staking-weight: 0.0,
+    5: {
```
marcinwadon, line 94: Shouldn't it stay at `0`? If no, what's the reason?

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/snapshot/storage/FeesAddressesStorage.scala
```diff
@@ -0,0 +1,10 @@
+package org.tessellation.node.shared.domain.snapshot.storage
+
+import org.tessellation.schema.GlobalSnapshotInfo
+import org.tessellation.schema.address.Address
+
+trait FeesAddressesStorage[F[_]] {
+  def setAllFeesAddresses(globalSnapshotInfo: GlobalSnapshotInfo): F[Unit]
+  def addMetagraphFeeAddresses(metagraphId: Address, addresses: List[Address]): F[Unit]
+  def getAllFeesAddresses: F[Map[Address, List[Address]]]
+}
```
marcinwadon, line 10: I think it can be reduced to:
```scala
trait FeeStorage[F[_]] {
  def set(globalSnapshotInfo: GlobalSnapshotInfo): F[Unit]
  def addMetagraph(metagraphId: Address, addresses: List[Address]): F[Unit]
  def getAll: F[Map[Address, List[Address]]]
}
```

As I see, we use `Address` for metagraph identifier. It may be worth creating an alias or newtype like:
```scala
@newtype
case class MetagraphId(address: Address)
```

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/statechannel/FeeCalculator.scala
```diff
@@ -28,6 +29,17 @@ case class FeeCalculatorConfig(
   proWeight: NonNegBigDecimal
 )
 
+case class SnapshotFeesInfo(
+  allFeesAddresses: Map[Address, List[Address]],
+  stakingBalance: Balance,
+  maybeOwnerAddress: Option[Address],
+  maybeStakingAddress: Option[Address]
```
marcinwadon, line 36: `maybe` in the name is redundant since the type is self explanatory

marcinwadon (changes requested):

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencyMessageValidator.scala
```diff
@@ -66,6 +72,25 @@ object CurrencyMessageValidator {
...
+          } else {
+            val allAddresses = existingFeesAddresses.values.flatten.toList
+            if (allAddresses.contains(messageAddress)) {
+              AddressAlreadyInUse.invalidNec
+            } else {
+              message.validNec
+            }
+          }
+        }
+
```
kpudlik, line 93: I'd like to follow up on my previous comment about storage and synchronization issues. You already have access to `GlobalSnapshotInfo` (via `LastSnapshotStorage[F, GlobalIncrementalSnapshot, GlobalSnapshotInfo]`), so instead of having a separate storage for message addresses, you could just retrieve all the information from `GlobalSnapshotInfo` and have a single source of truth. Currency L0 and Global L0 both have access to such a storage.

```scala
               def validateIfAddressAlreadyUsed(
          metagraphId: Address,
          message: Signed[CurrencyMessage]
        ) =
          lastGlobalSnapshotStorage.getCombined.map {
            case Some((_, info)) =>
              info.lastCurrencySnapshots.collect {
                case (address, Right((_, info))) => address -> info.lastMessages.toList.flatMap(_.values.map(_.address).toList).toSet
              }
            case None => SortedMap.empty[Address, Set[Address]]
          }.map { metagraphAddresses =>
            val messageAddress = message.address

            if (metagraphAddresses.get(metagraphId).exists(_.contains(messageAddress))) {
              message.validNec
            } else {
              if (metagraphAddresses.exists { case (_, addresses) => addresses.contains(messageAddress) }) {
                AddressAlreadyInUse.invalidNec
              } else {
                message.validNec
              }
            }
          }
```

kpudlik (changes requested):

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/http/routes/CurrencyMessageRoutes.scala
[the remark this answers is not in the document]
IPadawans: Got it thanks, I've removed the storage like we discussed

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/http/routes/CurrencyMessageRoutes.scala
```diff
@@ -47,7 +48,8 @@ class CurrencyMessageRoutes[F[_]: Async: Hasher](
           _.map(_._2).map(_.lastMessages.getOrElse(SortedMap.empty[MessageType, Signed[CurrencyMessage]]))
         }
         metagraphId <- identifierStorage.get
-        maybeResult <- maybeLastMsgs.traverse(validator.validate(msg, _, metagraphId))
+        allFeesAddresses <- feesAddressesStorage.getAllFeesAddresses
```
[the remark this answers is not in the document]
IPadawans: 👍🏼

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/domain/statechannel/StateChannelService.scala
```diff
@@ -31,15 +36,19 @@ object StateChannelService {
...
-          case Valid(_) =>
-            mkDagCell(L0CellInput.HandleStateChannelSnapshot(stateChannelOutput))
-              .run()
-              .as(().asRight[NonEmptyList[StateChannelValidationError]])
-          case Invalid(errors) => errors.toNonEmptyList.asLeft[Unit].pure[F]
-        }
+        val ownerAndStakingAddresses = fetchMetagraphFeesAddresses(stateChannelOutput.address, state)
+        for {
+          allFeesAddresses <- feesAddressesStorage.getAllFeesAddresses
+          snapshotFeesInfo = SnapshotFeesInfo(allFeesAddresses, staked, ownerAndStakingAddresses._1, ownerAndStakingAddresses._2)
```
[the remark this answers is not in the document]
IPadawans: 👍🏼

modules/node-shared/src/main/resources/application.conf
IPadawans: 👍🏼

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/statechannel/FeeCalculator.scala
IPadawans: 👍🏼

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/http/routes/CurrencyMessageRoutes.scala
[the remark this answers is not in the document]
IPadawans: 👍🏼

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencyMessageValidator.scala
IPadawans: 👍🏼

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/statechannel/StateChannelValidator.scala
```diff
@@ -158,6 +174,27 @@ object StateChannelValidator {
       else
         signedSC.validNec
 
+    private def validateIfAddressAlreadyUsed(
```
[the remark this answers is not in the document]
IPadawans: Updated

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/snapshot/storage/FeesAddressesStorage.scala
kpudlik: @marcinwadon about the alias - would be nice to see:
```scala
Map[MetagraphId, List[Address]]
```

But I am wondering how much refactoring it needs to change it everywhere :D

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -67,21 +66,65 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
         snapshotOrdinal: SnapshotOrdinal,
         lastGlobalSnapshotInfo: GlobalSnapshotInfo,
         events: List[StateChannelOutput],
         validationType: StateChannelValidationType
-      )(implicit hasher: Hasher[F]): F[StateChannelAcceptanceResult] =
+      )(implicit hasher: Hasher[F]): F[StateChannelAcceptanceResult] = {
+        val allFeesAddresses: Map[Address, Set[Address]] =
+          lastGlobalSnapshotInfo.lastCurrencySnapshots.collect {
+            case (address, Right((_, info))) => Map(address -> info.lastMessages.toList.flatMap(_.values.map(_.address).toList).toSet)
+          }.flatten.toMap
```
kpudlik, line 101: I have already seen that code 🤔 
Can we make it DRY somehow?

Maybe we can make `CurrencyMessageValidator` dependent on `LastSnapshotStorage`?
Or at least make a method somewhere like `getFeeAddresses(info: GlobalSnapshtoInfo)`

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/statechannel/StateChannelValidator.scala
```diff
@@ -34,20 +34,40 @@ trait StateChannelValidator[F[_]] {
...
 object StateChannelValidator {
 
+  def validateIfAddressAlreadyUsed(
+    metagraphId: Address,
+    existingFeesAddresses: Map[Address, Set[Address]],
+    maybeAddress: Option[Address]
+  ): StateChannelValidationErrorOr[Unit] = maybeAddress match {
+    case None => ().validNec
+    case Some(address) =>
+      val metagraphAddresses = existingFeesAddresses.getOrElse(metagraphId, List.empty)
```
kpudlik, line 58: You can omit constructing empty list as a fallback by directly using `exists` on `Option`. Let's save some memory ;)

```scala
val existsInMetagraphAddresses = existingFeesAddresses.get(metagraphId).exists(_.contains(address))
        
if (existsInMetagraphAddresses) {
  stateChannelOutput.snapshotBinary.validNec
} else {
  ...
}
```

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/statechannel/StateChannelValidator.scala
```diff
@@ -34,20 +34,40 @@ trait StateChannelValidator[F[_]] {
...
+      if (metagraphAddresses.toList.contains(address)) {
+        ().validNec
+      } else {
+        val allAddresses = existingFeesAddresses.values.flatten.toList
```
kpudlik, line 62: The problem with such an approach is that you always iterate over the whole `Map` to construct the flattened `List` and then iterate over that flattened `List` again. We will have a lot of addresses there in the future so I think you can just check the `Map[Address, Set[Address]]` directly without making it a flattened list. With such an approach you stop iterating over `Map` when you find the address so it is optimized:

```scala
val existsInAllAddresses = existingFeesAddresses.exists { case (_, addresses) => addresses.contains(address)}
          
if (existsInAllAddresses) {
  AddressAlreadyInUse.invalidNec
} else {
  stateChannelOutput.snapshotBinary.validNec
}
```

kpudlik (changes requested):

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/statechannel/StateChannelValidator.scala
kpudlik: Also one addition:

We have `Set[Address]` but you made a fallback to `List[Address]`.
These 2 structures are different:

- List has it's specific order and can have duplicates and `contains` is `O(n)`
- Set doesn't have any particular order and `contains` is `O(1)` or `O(log n)`

In general in Scala we should be careful with using proper structures :+1:

Also some interesting reference if you want to read more: https://docs.scala-lang.org/overviews/collections-2.13/performance-characteristics.html

kpudlik (changes requested):

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
IPadawans: Sure, I've update to have the function
`getFeeAddresses`

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/statechannel/StateChannelValidator.scala
IPadawans: Cool, thanks =]

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/statechannel/StateChannelValidator.scala
IPadawans: Cool, thanks =]

modules/node-shared/src/main/scala/org/tessellation/node/shared/domain/statechannel/StateChannelValidator.scala
```diff
@@ -34,19 +34,43 @@ trait StateChannelValidator[F[_]] {
...
     implicit hasher: Hasher[F]
   ): F[StateChannelValidationErrorOr[StateChannelOutput]]
-
 }
 
 object StateChannelValidator {
+  def getFeeAddresses(info: GlobalSnapshotInfo): Map[Address, Set[Address]] =
+    info.lastCurrencySnapshots.collect {
+      case (address, Right((_, info))) => Map(address -> info.lastMessages.toList.flatMap(_.values.map(_.address).toList).toSet)
+    }.flatten.toMap
```
kpudlik, line 53: I think that creating an immediate `Map` and flattening is redundant. It should work as well:

```scala
info.lastCurrencySnapshots.collect {
  case (address, Right((_, info))) => address -> info.lastMessages.toList.flatMap(_.values.map(_.address).toList).toSet
}
```

Basically if you return a tuple from `collect` called on `Map` then it stays as `Map`:

kpudlik (changes requested):

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -67,21 +67,62 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
-          }).map(_.errorMap(error => (event.address, error)))
+            validationV.flatMap { validation =>
+              if (validation.isValid) {
+                Async[F]
+                  .delay(
+                    allFeesAddresses.updated(
+                      event.address,
+                      Set(snapshotFeesInfo.ownerAddress, snapshotFeesInfo.stakingAddress).flatten
+                    )
+                  )
```
kpudlik, line 118: Isn't it redundant? As we update a local `val` and we newer pass it forward.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -67,21 +67,62 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
-            case StateChannelValidationType.Full =>
-              stateChannelValidator.validate(event, snapshotOrdinal, stakingBalance)
-            case StateChannelValidationType.Historical =>
-              stateChannelValidator.validateHistorical(event, snapshotOrdinal, stakingBalance)
-          }).map(_.errorMap(error => (event.address, error)))
+            validationV.flatMap { validation =>
+              if (validation.isValid) {
+                Async[F]
+                  .delay(
+                    allFeesAddresses.updated(
```
kpudlik, line 114: `updated` returns a new `Map` but it never gets assigned.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -67,26 +66,63 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
+                  case StateChannelValidationType.Historical =>
+                    stateChannelValidator.validateHistorical(event, snapshotOrdinal, snapshotFeesInfo)
+                }
+
+                validationV.map {
+                  case valid @ Validated.Valid(event) =>
+                    val updatedAllFeesAddresses = prevAllFeeAddresses.updated(
+                      event.address,
+                      Set(snapshotFeesInfo.ownerAddress, snapshotFeesInfo.stakingAddress).flatten
+                    )
```
kpudlik, line 117: I am wondering if we shouldn't concat it with the previous `Set` instead of overriding it in every iteration.

```scala
val updatedAllFeesAddresses = prevAllFeeAddresses.updatedWith(event.address) { existing => 
   val added = Set(snapshotFeesInfo.ownerAddress, snapshotFeesInfo.stakingAddress).flatten
   existing.map(_ ++ added).orElse(added.some)
}
```

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -67,26 +66,63 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
+        event.snapshotBinary.value.lastSnapshotHash match {
+          case hash if hash == Hash.empty => SnapshotFeesInfo.empty.pure // genesis
+          case _ =>
+            deserialize[Signed[CurrencyIncrementalSnapshot]](event.snapshotBinary).flatMap {
+              case None => new Exception("Could not get snapshot after deserializing").raiseError[F, SnapshotFeesInfo]
+              case Some(snapshot) =>
+                Async[F].delay {
+                  val stakingBalance = fetchStakingBalance(event.address, lastGlobalSnapshotInfo)
+                  val maybeOwnerAddress = snapshot.value.messages.flatMap(_.find(_.messageType == MessageType.Owner)).map(_.address)
+                  val maybeStakingAddress = snapshot.value.messages.flatMap(_.find(_.messageType == MessageType.Staking)).map(_.address)
```
kpudlik, line 86: `snapshot.value.messages` are `SortedSet` which indicates that we can have `n` messages there (not only 2 messages). If only 1 staking and 1 owner are allowed then I am wondering why `messages` in `CurrencyIncrementalSnapshot` are not something like: `case class Messages(Option[Owner], Option[Staking])`.

But if `SortedSet` was used on purpose (so we allow `n` messages) then I would be careful with `find` as if finds the very first message of a specific type.

If `SortedSet` uses `Ordering` based on ordinal ASC then it will return the one with lowest ordinal (reverse needed to `findLast`)
If `SortedSet` uses `Ordering` based on ordinal DESC then it will return the one with highest ordinal
If `SortedSet` uses `Ordering` which is not specified then it is kinda unpredictable and I would `maxBy(_.ordinal)` or sth.

kpudlik (changes requested):

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
kpudlik: Of course, the ultimate solution would be to provide proper `Ordering` to `SortedSet` globally based on `ordinal` because sorting `SortedSet` manually makes `SortedSet` useless.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
kpudlik: I mean, we probably should accumulate addresses every iteration. Because previously we were only keeping 2 last addresses

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -67,26 +66,64 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
+      ): F[SnapshotFeesInfo] =
+        event.snapshotBinary.value.lastSnapshotHash match {
+          case hash if hash == Hash.empty => SnapshotFeesInfo.empty.pure // genesis
+          case _ =>
+            deserialize[Signed[CurrencyIncrementalSnapshot]](event.snapshotBinary).flatMap {
+              case None => new Exception("Could not get snapshot after deserializing").raiseError[F, SnapshotFeesInfo]
+              case Some(snapshot) =>
+                Async[F].delay {
+                  val stakingBalance = fetchStakingBalance(event.address, lastGlobalSnapshotInfo)
+                  val sortedMessagesDesc = snapshot.value.messages.map(_.toList.sortBy(_.ordinal.value.value)(Ordering[Long].reverse))
```
kpudlik, line 85: As a follow up to https://github.com/Constellation-Labs/tessellation/pull/890#discussion_r1670923990

```scala
@derive(eqv, show, encoder, decoder, order, ordering)
case class CurrencyMessage(messageType: MessageType, address: Address, metagraphId: Address, parentOrdinal: MessageOrdinal) {
    def ordinal: MessageOrdinal = parentOrdinal.next
}
```

In such code the `Ordering[CurrencyMessage]` sorts by all the fields... instead of just by `ordinal`. That being said, the `SortedSet[Signed[CurrencyMessage]]` is wrongly sorted in snapshots because the wrong `Ordering` was provided in `CurrencyMessage`. Of course `SortedSet` makes sure that all the nodes have it sorted in the same way but in all the places where you want to either get first or last ordinal you can't simply rely on `head` or `last`. It is a little bit error-prone.

The worst part is that even if you provide the correct one now globally:

```scala
  @derive(eqv, show, encoder, decoder)
  case class CurrencyMessage(messageType: MessageType, address: Address, metagraphId: Address, parentOrdinal: MessageOrdinal) {
    def ordinal: MessageOrdinal = parentOrdinal.next
    implicit val order: Order[CurrencyMessage] = Order[MessageOrdinal].contramap(_.ordinal)
    implicit val ordering: Ordering[CurrencyMessage] = Order[CurrencyMessage].toOrdering
  }
```

Then new snapshots will be fine but old ones will be incorrect because these were sorted in an old way so hash is calculated in a different way...

So I guess we can't fix it without wiping the testnet/integrationnnet chain and we need to sort it locally.

So TL:DR I am fine with sorting it there unless we verify if we can fix it globally.

cc @sdavidp

kpudlik, line 85: Alternatively you can just use `-` before the value:

```scala
val sortedMessagesDesc = snapshot.value.messages.map(_.toList.sortBy(-_.ordinal.value.value))
```

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -67,26 +66,64 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
+                  val maybeOwnerAddress = sortedMessagesDesc.flatMap(_.find(_.messageType == MessageType.Owner)).map(_.address)
```
kpudlik, line 86: Use `===` instead of `==` (you need to change all the `import cats.syntax.XYZ` into single `import cats.syntax.all._`)

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -67,26 +66,64 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
+                  val maybeStakingAddress = sortedMessagesDesc.flatMap(_.find(_.messageType == MessageType.Staking)).map(_.address)
```
kpudlik, line 87: Use `===` instead of `==` (you need to change all the `import cats.syntax.XYZ` into single `import cats.syntax.all._`)

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencyMessageValidator.scala
```diff
@@ -89,7 +96,14 @@ object CurrencyMessageValidator {
           case _ if !isMetagraphIdValid =>
             Async[F].pure(WrongMetagraphId.invalidNec)
           case _ =>
-            validateSignatures(message)
+            for {
+              validSignature <- validateSignatures(message)
+              addressAlreadyUsed =
```
kpudlik, line 101: It should be rather `addressNotUsed` because we have `validSignature` instead of `invalidSignature`

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
IPadawans: Got it, so let's keep locally for now

IPadawans: 👍🏼

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
IPadawans: 👍🏼

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
IPadawans: 👍🏼

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencyMessageValidator.scala
IPadawans: 👍🏼

PR #891, opened 2024-07-09
Fix GlobalSnapshotInfo storage

marcinwadon:
Use fallback class for Kryo deserialization of GlobalSnapshotInfo

### DIFF
--- a/modules/shared/src/main/scala/org/tessellation/security/signature/Signed.scala
+++ b/modules/shared/src/main/scala/org/tessellation/security/signature/Signed.scala
@@ -2,6 +2,7 @@ package org.tessellation.security.signature
 
 import java.security.KeyPair
 
+import cats._
 import cats.data.NonEmptySet
 import cats.effect.Async
 import cats.syntax.applicative._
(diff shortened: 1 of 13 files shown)


kpudlik (approved):

IPadawans (dismissed):

scasplte2 (approved):
LGTM

modules/shared/src/main/scala/org/tessellation/security/signature/Signed.scala
```diff
@@ -33,6 +33,10 @@ import org.scalacheck.Arbitrary.arbitrary
 case class Signed[+A](value: A, proofs: NonEmptySet[SignatureProof])
 
 object Signed {
+  implicit val functor: Functor[Signed] = new Functor[Signed] {
+    def map[A, B](fa: Signed[A])(f: A => B): Signed[B] = Signed(f(fa.value), fa.proofs)
```
scasplte2, line 37: Shouldn't applying a function to any value invalidate the associated proofs?

marcinwadon: Not always. CurrencyIncrementalSnapshotV1 should be equivalent to CurrencyIncrementalSnapshot with that difference that newer fields are None (schema upgrade). Hence, the calculated hash should be equally the same so the proofs should match

IPadawans (approved):

PR #893, opened 2024-07-10
Fixing data-l1 consensus when lastGlobalSnapshot isn't available yet

IPadawans:
### Changes
+ We had a bug that the data-l1 consensus was being stopped when the lastGlobalSnapshot wasn't present. This error was halting the streaming and blocking the node from starting the own consensus
+ I've followed the pattern of the currency layer and added a validation to the data consensus before trying to start the own consensus

### Tests
It was working as expected, as the image below illustrates 
Captura de Tela 2024-07-10 às 11 04 43

### DIFF
--- a/modules/currency-l1/src/main/scala/org/tessellation/currency/l1/DataApplication.scala
+++ b/modules/currency-l1/src/main/scala/org/tessellation/currency/l1/DataApplication.scala
@@ -11,27 +11,33 @@ import scala.concurrent.duration._
 
 import org.tessellation.currency.dataApplication.ConsensusInput.OwnerConsensusInput
 import org.tessellation.currency.dataApplication._
+import org.tessellation.currency.l1.domain.dataApplication.consensus.Validator.{
+  canStartOwnDataConsensus,
+  isLastGlobalSnapshotPresent,
+  isPeerInputValid
+}
 import org.tessellation.currency.l1.domain.dataApplication.consensus.{ConsensusClient, ConsensusState, Engine}
 import org.tessellation.currency.l1.modules.{Queues, Services}
 import org.tessellation.currency.schema.currency._
+import org.tessellation.dag.l1.domain.consensus.block.config.DataConsensusConfig
 import org.tessellation.dag.l1.http.p2p.L0BlockOutputClient
 import org.tessellation.node.shared.domain.cluster.storage.{ClusterStorage, L0ClusterStorage}
+import org.tessellation.node.shared.domain.node.NodeStorage
 import org.tessellation.node.shared.domain.snapshot.storage.LastSnapshotStorage
 import org.tessellation.schema.peer.PeerId
 import org.tessellation.schema.{GlobalIncrementalSnapshot, GlobalSnapshotInfo}
-import org.tessellation.security.signature.Signed
 import org.tessellation.security.{Hasher, SecurityProvider}
 
 import fs2.{Pipe, Stream}
-import io.circe.Encoder
-import io.circe.generic.semiauto.deriveEncoder
 import org.typelevel.log4cats.slf4j.Slf4jLogger
 
 object DataApplication {
   def run[F[_]: Async: Random: Hasher: SecurityProvider: L1NodeContext](
+    dataConsensusCfg: DataConsensusConfig,
     clusterStorage: ClusterStorage[F],
     l0ClusterStorage: L0ClusterStorage[F],
     lastGlobalSnapshot: LastSnapshotStorage[F, GlobalIncrementalSnapshot, GlobalSnapshotInfo],
+    nodeStorage: NodeStorage[F],
     blockOutputClient: L0BlockOutputClient[F],
     consensusClient: ConsensusClient[F],
     services: Services[
@@ -49,39 +55,34 @@ object DataApplication {
     def logger = Slf4jLogger.getLogger[F]
 
     def inspectionTrigger: Stream[F, OwnerConsensusInput] =
-      Stream.awakeEvery(5.seconds).as(ConsensusInput.InspectionTrigger)
+      Stream
+        .awakeEvery(5.seconds)
+        .evalFilter(_ => isLastGlobalSnapshotPresent(lastGlobalSnapshot))
+        .as(ConsensusInput.InspectionTrigger)
 
     def ownRoundTrigger: Stream[F, OwnerConsensusInput] =
       Stream
         .awakeEvery(5.seconds)
+        .evalFilter { _ =>
+          canStartOwnDataConsensus(
+            nodeStorage,
+            clusterStorage,
+            lastGlobalSnapshot,
+            dataConsensusCfg.peersCount
+          ).handleErrorWith { e =>
+            logger.warn(e)("Failure checking if own consensus can be kicked off!").as(false)
+          }
+        }
         .as(ConsensusInput.OwnRoundTrigger)
 
     def ownerBlockConsensusInputs =
       inspectionTrigger.merge(ownRoundTrigger)
 
-    def isPeerInputValid(input: Signed[ConsensusInput.PeerConsensusInput]): F[Boolean] =
-      for {
-        validSignature <- input.value match {
-          case proposal: ConsensusInput.Proposal =>
-            implicit val e = ConsensusInput.Proposal.encoder(dataApplicationService.dataEncoder)
-            Signed(proposal, input.proofs).hasValidSignature
-
-          case signatureProposal: ConsensusInput.SignatureProposal =>
-            implicit val e: Encoder[ConsensusInput.SignatureProposal] = deriveEncoder
-            Signed(signatureProposal, input.proofs).hasValidSignature
-
-          case cancellation: ConsensusInput.CancelledCreationRound =>
-            implicit val e: Encoder[ConsensusInput.CancelledCreationRound] = deriveEncoder
-            Signed(cancellation, input.proofs).hasValidSignature
-        }
-
-        signedExclusively = input.isSignedExclusivelyBy(PeerId._Id.get(input.value.senderId))
-      } yield validSignature && signedExclusively
-
     def peerBlockConsensusInputs =
       Stream
         .fromQueueUnterminated(queues.dataApplicationPeerConsensusInput)
-        .evalFilter(isPeerInputValid(_))
+        .evalFilter(isPeerInputValid(_, dataApplicationService))
+        .evalFilter(_ => isLastGlobalSnapshotPresent(lastGlobalSnapshot))
         .map(_.value)
 
     def blockConsensusInputs =
@@ -91,6 +92,7 @@ object DataApplication {
       _.evalMapAccumulate(ConsensusState.Empty) {
         Engine
           .fsm(
+            dataConsensusCfg,
             dataApplicationService,
             clusterStorage,
             lastGlobalSnapshot,
@@ -108,7 +110,7 @@ object DataApplication {
             .eval(logger.debug(s"Data application block created! Hash=${hashedBlock.hash}, ProofsHash=${hashedBlock.proofsHash}"))
             .as(fb)
         case (_, ConsensusOutput.CleanedConsensuses(ids)) =>
-          Stream.eval(logger.debug(s"Cleaned consensuses ids=${ids}")) >> Stream.empty
+          Stream.eval(logger.debug(s"Cleaned consensuses ids=$ids")) >> Stream.empty
         case (_, ConsensusOutput.Noop) => Stream.empty
       }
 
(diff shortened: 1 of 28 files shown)


modules/currency-l1/src/main/scala/org/tessellation/currency/l1/DataApplication.scala
```diff
@@ -54,6 +63,20 @@ object DataApplication {
     def ownRoundTrigger: Stream[F, OwnerConsensusInput] =
       Stream
         .awakeEvery(5.seconds)
+        .evalMap { _ =>
```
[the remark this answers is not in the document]
IPadawans: Not actually. I don't know much about the semaphore, the previous version didn't use the semaphore in any places so I thought it might not be used. Do you think we should add? @marcinwadon

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/Validator.scala
```diff
@@ -73,7 +74,38 @@ object Validator {
...
+    lastGlobalSnapshot: LastSnapshotStorage[F, GlobalIncrementalSnapshot, GlobalSnapshotInfo],
+    peersCount: PosInt,
+    tipsCount: PosInt
+  ): F[Boolean] =
+    for {
+      noOwnRoundInProgress <- consensusStorage.ownConsensus.get.map(_.isEmpty)
+      stateReadyForConsensus <- nodeStorage.getNodeState.map(isReadyForBlockConsensus)
+      enoughPeers <- enoughPeersForConsensus(clusterStorage, peersCount)
+      enoughTips <- enoughTipsForConsensus(blockStorage, tipsCount)
+      lastGlobalSnapshot <- lastGlobalSnapshot.get
```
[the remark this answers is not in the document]
IPadawans: 👍🏼

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/DataApplication.scala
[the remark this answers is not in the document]
scasplte2: In the `StateChannel` [class](https://github.com/Constellation-Labs/tessellation/blob/49bb9209ee289652d24de340efae46c04bb5f3e3/modules/dag-l1/src/main/scala/org/tessellation/dag/l1/StateChannel.scala) there are `Semaphore(1)` instances created in the `make` method that are then used in the outer function calling `canStartOwnConsensus`. As I understand, Semaphore(1) is equivalent to a Mutex, which is just an explicit lock on an IO resource. I'm also still learning the nuance but I believe the lock is needed here since we have multiple processes which need to be in an "open" status (therefore stable and not in the middle of a modification) before going about the consensus operation.

I think you may also want to import `org.tessellation.ext` to access the `.evalMapLocked` helper method to recreate the process from StateChannel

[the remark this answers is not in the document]
marcinwadon: I think it's because the currency accepts and stores transactions but we do not store data updates, we just send them to L0 for consensus acceptance. So L1 here is kind a proxy. I'm not sure about the function you've just added, hence the question. The currency consensus uses `evalMapLocked` with block acceptance semaphore (I bet it's a mutex).

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/Validator.scala
```diff
@@ -73,7 +74,37 @@ object Validator {
             if (!enoughTips) "Not enough tips" else "",
             if (!enoughTxs) "No transactions" else ""
           ).filter(_.nonEmpty).mkString(", ")
-          logger.debug(s"Cannot start own consensus: ${reason}")
+          logger.debug(s"Cannot start currency own consensus: $reason")
+        }
+    } yield res
+
+  def canStartOwnDataConsensus[F[_]: Async](
```
marcinwadon, line 81: I think it's defined in wrong place. `consensus/block` is for the currency consensus and it's defined in `dag-l1`.
image

DAG L1 knows nothing about data updates. The fn you've just created should be defined in the currency L1. There is a namespace `org.tessellation.currency.l1.domain.dataApplication.consensus`

marcinwadon (changes requested):

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/Validator.scala
kpudlik: Also name "ownCurrencyConsensus" sounds strange in dag-l1. DAG is technically a currency but we rather use currency for non-dag currencies. Wdyt?

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/Validator.scala
```diff
@@ -73,7 +74,37 @@ object Validator {
...
+    for {
+      noOwnRoundInProgress <- consensusStorage.ownConsensus.get.map(_.isEmpty)
+      stateReadyForConsensus <- nodeStorage.getNodeState.map(isReadyForBlockConsensus)
+      enoughPeers <- enoughPeersForConsensus(clusterStorage, peersCount)
+      enoughTips <- enoughTipsForConsensus(blockStorage, tipsCount)
+      lastGlobalSnapshotPresent <- lastGlobalSnapshot.get.map(_.isDefined)
+
+      res = noOwnRoundInProgress && stateReadyForConsensus && enoughPeers && enoughTips && lastGlobalSnapshotPresent
+      _ <-
+        Applicative[F].whenA(!res) {
```
kpudlik, line 99: Eventually `Applicative[F].unlessA(res)`. Whichever you prefer.

IPadawans: Let's keep this to be in the same pattern than currency validation

modules/dag-l1/src/main/scala/org/tessellation/dag/l1/domain/consensus/block/Validator.scala
IPadawans: I agree

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/DataApplication.scala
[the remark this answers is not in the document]
IPadawans: I think that too, Marcin

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/domain/dataApplication/consensus/Validator.scala
```diff
@@ -0,0 +1,56 @@
...
+import org.tessellation.schema.{GlobalIncrementalSnapshot, GlobalSnapshotInfo}
+
+import eu.timepit.refined.types.numeric.PosInt
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+
+object Validator {
+
+  private def logger[F[_]: Async] = Slf4jLogger.getLogger
+
+  def isReadyForBlockConsensus(state: NodeState): Boolean = state == Ready
```
marcinwadon, line 25: It's not a block consensus but a data application consensus

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/domain/dataApplication/consensus/Validator.scala
```diff
@@ -0,0 +1,55 @@
...
+    blockStorage: BlockStorage[F],
+    lastGlobalSnapshot: LastSnapshotStorage[F, GlobalIncrementalSnapshot, GlobalSnapshotInfo],
+    peersCount: PosInt,
+    tipsCount: PosInt
+  ): F[Boolean] =
+    for {
+      noOwnRoundInProgress <- consensusStorage.ownConsensus.get.map(_.isEmpty)
+      stateReadyForConsensus <- nodeStorage.getNodeState.map(isReadyForDataApplicationConsensus)
+      enoughPeers <- enoughPeersForConsensus(clusterStorage, peersCount)
+      enoughTips <- enoughTipsForConsensus(blockStorage, tipsCount)
```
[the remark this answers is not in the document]
IPadawans: We don't need, removed

IPadawans:
@marcinwadon we had a second spot that is offering data to the queue
```scala
protected val p2p: HttpRoutes[F] = HttpRoutes.of[F] {
    case req @ POST -> Root / "consensus" / "data-application" =>
      implicit val decoder: Decoder[ConsensusInput.Proposal] = ConsensusInput.Proposal.decoder(dataApplication.dataDecoder)

      req
        .as[Signed[ConsensusInput.Proposal]]
        .map(_.asInstanceOf[Signed[ConsensusInput.PeerConsensusInput]])
        .handleErrorWith { _ =>
          req
            .as[Signed[ConsensusInput.SignatureProposal]]
            .map(_.asInstanceOf[Signed[ConsensusInput.PeerConsensusInput]])
            .handleErrorWith { _ =>
              req
                .as[Signed[ConsensusInput.CancelledCreationRound]]
                .map(_.asInstanceOf[Signed[ConsensusInput.PeerConsensusInput]])
            }
        }
        .flatMap { consensusInput =>
          S.supervise(dataApplicationPeerConsensusInput.offer(consensusInput))
        }
        .flatMap(_ => Ok())
        .handleErrorWith { err =>
          logger.error(err)(s"An error occured") >> InternalServerError()
        }
```
So I've added additional validations in another places to avoid the error

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/DataApplication.scala
```diff
@@ -48,43 +56,39 @@ object DataApplication {
...
+    val peerBlockConsensusInputs =
       Stream
         .fromQueueUnterminated(queues.dataApplicationPeerConsensusInput)
-        .evalFilter(isPeerInputValid(_))
+        .evalFilter(input => isPeerInputValid(input, dataApplicationService))
+        .evalFilter(_ => isLastGlobalSnapshotPresent(lastGlobalSnapshot))
         .map(_.value)
 
-    def blockConsensusInputs =
+    val blockConsensusInputs =
```
marcinwadon, line 91: why?

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/DataApplication.scala
```diff
@@ -48,43 +56,39 @@ object DataApplication {
...
-
-        signedExclusively = input.isSignedExclusivelyBy(PeerId._Id.get(input.value.senderId))
-      } yield validSignature && signedExclusively
-
-    def peerBlockConsensusInputs =
+    val peerBlockConsensusInputs =
       Stream
         .fromQueueUnterminated(queues.dataApplicationPeerConsensusInput)
-        .evalFilter(isPeerInputValid(_))
+        .evalFilter(input => isPeerInputValid(input, dataApplicationService))
```
marcinwadon, line 87: `.evalFilter(isPeerInputValid(_, dataApplicationService))`

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/DataApplication.scala
```diff
@@ -49,39 +55,34 @@ object DataApplication {
...
       Stream
         .awakeEvery(5.seconds)
+        .evalFilter { _ =>
+          canStartOwnDataConsensus(
+            nodeStorage,
+            clusterStorage,
+            lastGlobalSnapshot,
+            appConfig.dataConsensus.peersCount
+          ).handleErrorWith { e =>
+            logger.warn(e)("Failure checking if own consensus can be kicked off!").map(_ => false)
```
marcinwadon, line 73: FYI `.map(_ => a)` is equivalent to `.as(a)`

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/domain/dataApplication/consensus/Engine.scala
```diff
@@ -95,9 +97,9 @@ object Engine {
 
     implicit val dataEncoder = dataApplication.dataEncoder
 
-    def peersCount = 2
-    def timeout = 10.seconds
-    def maxDataUpdatesToDequeue = 50
+    def peersCount = appConfig.dataConsensus.peersCount.value
+    def timeout = appConfig.dataConsensus.timeout
+    def maxDataUpdatesToDequeue = appConfig.dataConsensus.maxDataUpdatesToDequeue.value
```
marcinwadon, line 102: It's okay to pass whole `appConfig` but it would be better to pass just `dataConsensus`, so here you'd have just `dataConsensusCfg.peersCount.value` and so on

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/domain/dataApplication/consensus/Validator.scala
```diff
@@ -0,0 +1,86 @@
...
+import eu.timepit.refined.types.numeric.PosInt
+import io.circe.Encoder
+import io.circe.generic.semiauto.deriveEncoder
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+
+object Validator {
+
+  private def logger[F[_]: Async] = Slf4jLogger.getLogger
+
+  private def isReadyForDataApplicationConsensus(state: NodeState): Boolean = state == Ready
```
marcinwadon, line 28: if you don't share the code, it can be just `isReadyForConsensus` :) you don't need to repeat data application since it's in the dataApplication/consensus package

marcinwadon (dismissed):

PR #894, opened 2024-07-10
Replace endpoint /data/validate with /data/estimate-fee

sdavidp:


modules/shared/src/main/scala/org/tessellation/currency/dataApplication/package.scala
```diff
@@ -202,6 +213,12 @@ object BaseDataApplicationContextualOps {
       ): F[DataApplicationValidationErrorOr[Unit]] =
         service.validateFee(gsOrdinal)(update.asInstanceOf[Signed[D]])
 
+      override def estimateFee(gsOrdinal: SnapshotOrdinal)(update: DataUpdate)(
+        implicit context: Context,
+        A: Applicative[F]
+      ): F[EstimatedFee] =
+        service.estimateFee(gsOrdinal)(update.asInstanceOf[Signed[D]])
```
scasplte2, line 220: why cast this to a `Signed`?

modules/shared/src/main/scala/org/tessellation/currency/schema/EstimatedFee.scala
```diff
@@ -0,0 +1,15 @@
...
+
+import org.tessellation.schema.address.Address
+import org.tessellation.schema.balance.Amount
+
+import derevo.cats.{order, show}
+import derevo.circe.magnolia.encoder
+import derevo.derive
+
+@derive(encoder, order, show)
+case class EstimatedFee(fee: Amount, address: Option[Address])
```
scasplte2, line 11: I don't understand the purpose of the `Option[Address]` in the fee estimation. For the Fee itself I can see needing an `Address` and `Amount` but the use here is unclear to me.

modules/shared/src/main/scala/org/tessellation/currency/dataApplication/package.scala
sdavidp: The class is defined in terms of the type variable `D <: DataUpdate`
`update` is of type `Signed[DataUpdate]` and the method requires a `Signed[D]`

modules/shared/src/main/scala/org/tessellation/currency/schema/EstimatedFee.scala
marcinwadon: @sdavidp If you say so, it should be either no fee or fee + address. Right now you allow an impossible state to be defined: fee + no address.

So it should be defined as below:
```
sealed trait EstimatedFee

case object Zero extends EstimatedFee
case class Estimated(amount: Amount, address: Address) extends EstimatedFee

```

sdavidp: @marcinwadon Ok, I will change the implementation. However, the ticket wants the encoded JSON for _zero_ to be `{ "fee" : 0 }`

modules/shared/src/main/scala/org/tessellation/currency/dataApplication/package.scala
sdavidp: As discussed in our meeting, the method does NOT want a `Signed[D]`. I'll fix it.

modules/shared/src/main/scala/org/tessellation/currency/schema/EstimatedFee.scala
sdavidp: @marcinwadon Updated.

marcinwadon:
@sdavidp Hey, there is a label missing

scasplte2 (approved):
everything seems in order here now once the base is updated

marcinwadon (approved):

PR #899, opened 2024-07-18
Use given fee transactions if no data application available

sdavidp:


modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotValidator.scala
```diff
@@ -156,6 +158,10 @@ object CurrencySnapshotValidator {
                 case None =>
                   creationResult.focus(_.artifact.dataApplication).replace(expected.dataApplication)
               }
+            }.map { creationResult =>
+              if (creationResult.artifact.messages.forall(_.isEmpty))
+                creationResult.focus(_.artifact.messages).replace(expected.messages)
+              else creationResult
```
sdavidp, line 164: Do we still need this? This was brought over when the branch was first created and it's not in develop now.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotCreator.scala
```diff
@@ -54,7 +55,8 @@ trait CurrencySnapshotCreator[F[_]] {
     trigger: ConsensusTrigger,
     events: Set[CurrencySnapshotEvent],
     rewards: Option[Rewards[F, CurrencySnapshotStateProof, CurrencyIncrementalSnapshot, CurrencySnapshotEvent]],
-    facilitators: Set[PeerId]
+    facilitators: Set[PeerId],
+    maybeFeeTransactions: Option[SortedSet[Signed[FeeTransaction]]] = None
```
kpudlik, line 59: So we introduced that field just to use it for the single validator on Global L0? 🤔 Sounds a little bit hacky to me like adding parameters that are used only in tests or sth.

What do you think about implementing it like `rewards`? So we make it a service/function instead of collection.

For example we could introduce a method in `DataApplicationSnapshotAcceptanceManager`:

```scala
trait DataApplicationSnapshotAcceptanceManager {
  //... rest of methods
  def calculateFeeTransactions(validatedUpdates: List[Signed[DataUpdate]]): List[FeeTransaction]
}
```

and metagraphs could just:
```scala
def calculateFeeTransactions(validatedUpdates: List[Signed[DataUpdate]]): List[FeeTransaction] = 
  service.extractFees(validatedUpdates)
```
and global could:

```scala
def calculateFeeTransactions(validatedUpdates: List[Signed[DataUpdate]]): List[FeeTransaction] = 
  expected.feeTransactions
```

WDYT?

sdavidp: @kpudlik I agree it is hacky, and I didn't like it but it got us working.

I like your idea about passing a function instead, and I think @marcinwadon may have suggested that too but I didn't get what he meant. Let me put something together, as well as the default error type @AlexBrandes is asking for and I'll push that here.

cc: @ryle-goehausen

sdavidp: @kpudlik Maybe I'm missing something but it seems like any other solution is going to be equally contrived, if not worse due to how far up the changes ripple, and in the end all it will be is a function that ignores its input and returns `expected.feeTransactions`.

I'll try a bit more but not sure I'm seeing much benefit.

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotCreator.scala
```diff
@@ -82,7 +84,8 @@ object CurrencySnapshotCreator {
       trigger: ConsensusTrigger,
       events: Set[CurrencySnapshotEvent],
       rewards: Option[Rewards[F, CurrencySnapshotStateProof, CurrencyIncrementalSnapshot, CurrencySnapshotEvent]],
-      facilitators: Set[PeerId]
+      facilitators: Set[PeerId],
+      maybeFeeTransactions: Option[SortedSet[Signed[FeeTransaction]]] = None
```
marcinwadon, line 88: either `feeTransactions` or change `rewards` to `maybeRewards`. Please don't mix it (same above).

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotCreator.scala
```diff
@@ -132,6 +135,7 @@ object CurrencySnapshotCreator {
 
           feeTransactions = dataApplicationAcceptanceResult
             .map(result => SortedSet.from(result.feeTransactions))
+            .orElse(maybeFeeTransactions)
```
marcinwadon, line 138: I think the fallback should be there only if there is no data application implementation provided

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotCreator.scala
marcinwadon: @sdavidp If there is a possibility, I'd highly recommend stubbing the data application service, so this part on screenshot would run on a stubbed service. That being said, we won't modify or add "a special case" to the acceptance. We'd just predefine the data application.
image

sdavidp: @marcinwadon Isn't that risky? If we stub the service then all functions in the service will be called and that might result in different behavior than if it is left to `None`

marcinwadon: you mean the service left to `None`? I don't understand what is the risk here, but anyway, we cannot run the data application since we don't know its implementation. So all we can do is to stub the fees to the expected ones. Yet, if you are able to stub it somehow instead of explicitly passing the expected fees, that would be better imo.

sdavidp: The risk is that there places where a different code path is followed when the data application is `None` or defined -- if defined then functions in the data app will be called.

Stubbing it out means defining functions that are required to be defined or leaving it as `???`, and either overriding or keeping those functions that have a default implementation.

Is this what you'd like me to do?
cc: @ryle-goehausen

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotCreator.scala
sdavidp: @marcinwadon Isn't that what the above is doing?

- `dataApplicationAcceptanceResult` is nonEmpty if `dataApplicationSnapshotAcceptanceManager` is nonEmpty
- `dataApplicationSnapshotAcceptanceManager` is nonEmpty if the data application (and calculated state storage) are nonEmpty

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotCreator.scala
sdavidp: @marcinwadon I was thinking of renaming it to `fallbackFeeTransactions`. WDYT?

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotCreator.scala
ryle-goehausen: in my opinion both of these solutions are somewhat close to each other. I don't think we need to worry about undefined portions of the data application, since that will throw errors on the GL0 as we encountered already in testing. And if we're adding a parameter here, adding it instead through stubbing it into the data application is relatively similar in terms of functionality, where if it's not present it should just throw an error. I think it's fine to stub instead of adding a parameter, and assume any other invocations just throw errors.

sdavidp: I'm not clear on how this improves things so maybe I'm missing something. Do we want this?

Create stub BaseDataApplicationL0Service and pass into node-shared:CurrencySnapshotValidator.make ?
If so, there are the following concerns.
  - which calls BaseDataApplicationL0Service.deserialize
  - changes the code path taken in `check` for setting the artifact.dataApplication
  - and then
    - passes it into CurrencySnapshotCreator.createProposalArtifact ? (this requires code changes there)
    - passes it into a stub of the DataApplicationManager that's passed into CurrencySnapshotCreator.make? (this will cause changes to ripple up)
  - . . . and the rabbit hole continues as we

sdavidp: I didn't get a chance to finish before accidentally submitting the above so will edit it now.

marcinwadon: hey hey hey, I've just said in my latest comment that I've realized we **cannot** stub the data application. Simply we cannot, as on the GL0 we do now know data application implementation details. Hence, I suggest to stub it in some other place rather than passing the expected fees directly to `createArtifact` function (if there is such possibility).

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotCreator.scala
marcinwadon: Ok. I didn't check the inside implementation of `dataApplicationAcceptanceResult`. If it's non empty if and only if data application is implemented, then it's fine :+1: Looking for this part only, it looks for me that the result may be empty if there is any internal error of the data application acceptance (and we should throw then, not fallback). but if that's not the case then it's okay :)

modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotCreator.scala
sdavidp: So change this,
```
trait CurrencySnapshotCreator[F[_]] {
  def createProposalArtifact(
    lastKey: SnapshotOrdinal,
    lastArtifact: Signed[CurrencySnapshotArtifact],
    lastContext: CurrencySnapshotContext,
    lastArtifactHasher: Hasher[F],
    trigger: ConsensusTrigger,
    events: Set[CurrencySnapshotEvent],
    rewards: Option[Rewards[F, CurrencySnapshotStateProof, CurrencyIncrementalSnapshot, CurrencySnapshotEvent]],
    facilitators: Set[PeerId],
    maybeFeeTransactions: Option[SortedSet[Signed[FeeTransaction]]] = None
  )(implicit hasher: Hasher[F]): F[CurrencySnapshotCreationResult[CurrencySnapshotEvent]]
}
```
to this?
```
trait CurrencySnapshotCreator[F[_]] {
  def createProposalArtifact(
    lastKey: SnapshotOrdinal,
    lastArtifact: Signed[CurrencySnapshotArtifact],
    lastContext: CurrencySnapshotContext,
    lastArtifactHasher: Hasher[F],
    trigger: ConsensusTrigger,
    events: Set[CurrencySnapshotEvent],
    rewards: Option[Rewards[F, CurrencySnapshotStateProof, CurrencyIncrementalSnapshot, CurrencySnapshotEvent]],
    facilitators: Set[PeerId],
    service: Option[BaseDataApplicationService] = None
  )(implicit hasher: Hasher[F]): F[CurrencySnapshotCreationResult[CurrencySnapshotEvent]]
}
```
and then we only use that passed in service to extract fees if the data manager wasn't defined?

sdavidp: If we do the above I'm going to need an `L0NodeContext` from somewhere, and that's currently not available in the `CurrencySnapshotCreator`.

marcinwadon: >  if the data manager wasn't defined

what is data manager?

sdavidp: Does this answer your question?
```
object CurrencySnapshotCreator {

  def make[F[_]: Async: JsonSerializer](
    currencySnapshotAcceptanceManager: CurrencySnapshotAcceptanceManager[F],
    dataApplicationSnapshotAcceptanceManager: Option[DataApplicationSnapshotAcceptanceManager[F]],
    snapshotSizeConfig: SnapshotSizeConfig,
    currencyEventsCutter: CurrencyEventsCutter[F]
  ): CurrencySnapshotCreator[F] = new CurrencySnapshotCreator[F] {
```

sdavidp: Screenshot 2024-07-25 at 7 14 57 AM

sdavidp: and in case you're wondering, `maybeLastDataApplication` is defined earlier in this function; it should probably be renamed.

Screenshot 2024-07-25 at 7 17 27 AM

sdavidp: > I don't like the idea of passing fee transactions as a parameter here (as written in the comment above). It's not clear when we should fill this field and when not, so it's error prone.

I don't like passing in the set of fee transactions either but at least it's more explicit and doesn't pretend to be something it's not.

My feeling is that passing in a stub or re-implementation of the data application service or the data snapshot acceptance manager is more error-prone because the functions receiving the stub have no idea (based on the function signature) that it's not a true implementation but we still need to use it specially and only in certain places.

If you really don't want the set passed in directly we can pass a function,
`feeTransactionFn: Option[() => SortedSet[Signed[FeeTransaction]]] = None`

marcinwadon: Ok, thanks :) I will let you decide on the final solution

kpudlik (approved):

marcinwadon (approved):

PR #903, opened 2024-07-25
Add TransactionFeeEstimator overridable with /transactions/estimate-fee route

kpudlik:


modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/TransactionRoutes.scala
```diff
@@ -0,0 +1,44 @@
...
+      implicit val hasher = txHasher
+
+      for {
+        transaction <- req.as[Signed[Transaction]]
+        hashedTransaction <- transaction.toHashed[F]
+        fee <- transactionFeeEstimator match {
+          case Some(estimator) => estimator.estimate(hashedTransaction)
+          case None            => TransactionFee.zero.pure[F]
+        }
+        response <- Ok(("fee" ->> fee.value.value) :: HNil)
```
scasplte2, line 41: would you be opposed to using Circe here instead of shapeless for the sake of consistency with the `DataApplicationRoutes`? Similar to
```
Ok(("hash" -> hashed.hash.value).asJson)
```

I don't recall the exact issue but I believe it was something like a compiler inference issue with deriving the EntityEncoder

scasplte2 (approved):

marcinwadon (dismissed):

modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/TransactionRoutes.scala
kpudlik: Cool, I switched to Circe :+1:

marcinwadon (approved):

scasplte2 (approved):

PR #906, opened 2024-07-29
fix: changed /data response to match /transactions

sdavidp:
The current implementation of the `/data` response produces a JSON array of two elements:
```
["hash","047e9423a1e6f9ee53596abf787884d007c82e1ec898a599f2048961ce5872e2"]
```
but I think the intention was to produce an object similar to the `/transactions` endpoint:
```
{"hash":"4ee922ad6b244df6ec5b84a54cb9024d92607cba3cbc2c7b283e0b2fa053cf5e"}
```

### DIFF
--- a/modules/currency-l1/src/test/scala/org/tessellation/currency/l1/http/DataApplicationRoutesSuite.scala
+++ b/modules/currency-l1/src/test/scala/org/tessellation/currency/l1/http/DataApplicationRoutesSuite.scala
@@ -1,5 +1,8 @@
 package org.tessellation.currency.l1.http
 
+import java.nio.charset.StandardCharsets
+import java.util.Base64
+
 import cats.Applicative
 import cats.data.NonEmptySet
 import cats.data.Validated.invalidNec
@@ -29,6 +32,7 @@ import org.tessellation.security.signature.Signed
 import org.tessellation.shared.sharedKryoRegistrar
 
 import io.circe.syntax.EncoderOps
+import io.circe.{Encoder, Json}
 import org.http4s.Method.{GET, POST}
 import org.http4s._
 import org.http4s.client.dsl.io._
@@ -45,7 +49,7 @@ object DataApplicationRoutesSuite extends HttpSuite {
   val defaultCurrencySnapshotStorage = mockLastSnapshotStorage[CurrencyIncrementalSnapshot, CurrencySnapshotInfo]()
 
   def construct(
-    updateQueue: ViewableQueue[F, Signed[DataUpdate]],
+    updateQueue: ViewableQueue[IO, Signed[DataUpdate]],
     l1Service: BaseDataApplicationL1Service[IO] = defaultL1Service,
     lastGlobalSnapshotStorage: LastSnapshotStorage[IO, GlobalIncrementalSnapshot, GlobalSnapshotInfo] = defaultGlobalSnapshotStorage
   ): IO[HttpRoutes[IO]] =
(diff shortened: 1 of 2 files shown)


marcinwadon (approved):

marcinwadon:
@sdavidp label is missing

scasplte2:
We can go back to shapeless if that is preferred but I was running into compiler inference issues at some point with trying to derive the `EntityEncoder` from the implicit scope. Invoking Shapeless `HNil` seemed like overkill since the expected response was JSON so my aim was to simplify from using the `HNil` to format a response. I had not noticed that my first approach produced an array instead of an object though so perhaps Shapeless is the easiest route to get the proper response.

scasplte2:
the correct approach to avoiding the shapeless import would be to return
`Ok(Json.obj("hash" -> hashed.hash.value.asJson))`

sdavidp:
> We can go back to shapeless if that is preferred but I was running into compiler inference issues at some point with trying to derive the `EntityEncoder` from the implicit scope. Invoking Shapeless `HNil` seemed like overkill since the expected response was JSON so my aim was to simplify from using the `HNil` and format the expected response directly. I had not noticed that my approach produced an array instead of an object though so perhaps Shapeless is the easiest route to get the proper response.

I suspect the reason an array was produced instead of an object is that we were constructing json from a tuple. If we wanted to generalize producing a json object for our `newtype` objects we could write this:
```
def toJsonObject[A: Encoder](name: String, a: A): Json = Json.obj(name -> a.asJson)
```

scasplte2:
This is the error that I encounter through the IDE and when running `sbt compile` via CLI when the return type is `Ok(("hash" ->> hashed.hash.value) :: HNil)`

```
[error] /home/scas/git/tessellation/modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/DataApplicationRoutes.scala:87:60: Cannot convert from String with shapeless.labelled.KeyTag[String("hash"),String] :: shapeless.HNil to an Entity, because no EntityEncoder[F, String with shapeless.labelled.KeyTag[String("hash"),String] :: shapeless.HNil] instance could be found.
[error]                 dataUpdatesQueue.offer(hashed.signed) >> Ok(("hash" ->> hashed.hash.value) :: HNil)
[error]                                                            ^
```

I am unsure what about my environment could cause the failure though as this doesn't seem to be a problem for you all.

sdavidp:
> This is the error that I encounter through the IDE and when running `sbt compile` via CLI when the return type is `Ok(("hash" ->> hashed.hash.value) :: HNil)`
> 
> ```
> [error] /home/scas/git/tessellation/modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/DataApplicationRoutes.scala:87:60: Cannot convert from String with shapeless.labelled.KeyTag[String("hash"),String] :: shapeless.HNil to an Entity, because no EntityEncoder[F, String with shapeless.labelled.KeyTag[String("hash"),String] :: shapeless.HNil] instance could be found.
> [error]                 dataUpdatesQueue.offer(hashed.signed) >> Ok(("hash" ->> hashed.hash.value) :: HNil)
> [error]                                                            ^
> ```
> 
> I am unsure what about my environment could cause the failure though as this doesn't seem to be a problem for you all.

I think you resolve that with an import (which IntellliJ doesn't realize is needed):
```
import io.circe.shapes._
```

scasplte2:
> > This is the error that I encounter through the IDE and when running `sbt compile` via CLI when the return type is `Ok(("hash" ->> hashed.hash.value) :: HNil)`
> > ```
> > [error] /home/scas/git/tessellation/modules/currency-l1/src/main/scala/org/tessellation/currency/l1/http/DataApplicationRoutes.scala:87:60: Cannot convert from String with shapeless.labelled.KeyTag[String("hash"),String] :: shapeless.HNil to an Entity, because no EntityEncoder[F, String with shapeless.labelled.KeyTag[String("hash"),String] :: shapeless.HNil] instance could be found.
> > [error]                 dataUpdatesQueue.offer(hashed.signed) >> Ok(("hash" ->> hashed.hash.value) :: HNil)
> > [error]                                                            ^
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I am unsure what about my environment could cause the failure though as this doesn't seem to be a problem for you all.
> 
> I think you resolve that with an import (which IntellliJ doesn't realize is needed):
> 
> ```
> import io.circe.shapes._
> ```

That worked, I missed the import when testing again and I suspect that I accidentally removed it when using the IntelliJ formatting tool. Thank you for the clarification.

AlexBrandes (changes requested):
It looks like one of the reasons this bug was possible was that we don't have any tests on the POST /data endpoint. 

Would you add a simple test that verifies the response schema on a successful request?

sdavidp:
> It looks like one of the reasons this bug was possible was that we don't have any tests on the POST /data endpoint.
> 
> Would you add a simple test that verifies the response schema on a successful request?

@AlexBrandes  Added.

AlexBrandes (approved):
LGTM

modules/currency-l1/src/test/scala/org/tessellation/currency/l1/http/DataApplicationRoutesSuite.scala
```diff
@@ -177,6 +194,31 @@ object DataApplicationRoutesSuite extends HttpSuite {
     }
   }
 
+  test("POST /data returns OK . . .") {
```
marcinwadon, line 197: what does `. . .` mean?

marcinwadon (dismissed):

modules/currency-l1/src/test/scala/org/tessellation/currency/l1/http/DataApplicationRoutesSuite.scala
sdavidp: @marcinwadon How about this?
`"POST /data returns OK and response as JSON object for valid updates"`

marcinwadon: do you check whether it's a valid JSON object or you check for the specific schema?
maybe: `POST /data returns 200 Ok and update's hash as a json object`

marcinwadon (approved):

marcinwadon (approved):

PR #920, opened 2024-08-07
ci: automated release action

marcinwadon:


.github/workflows/release-please.yml
```diff
@@ -0,0 +1,123 @@
...
+          {
+            echo "CHANGELOG<<EOF"
+            npx commit-and-tag-version --dry-run | sed -n '/---/,/---/p' | sed '1d;$d'
+            echo EOF
+          } >> "$GITHUB_OUTPUT"
+
+      - name: Set Git config
+        run: |
+          git config user.name github-actions[bot]
+          git config user.email 41898282+github-actions[bot]@users.noreply.github.com
```
kpudlik, line 73: Shouldn't we keep it in envs for actions?

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,123 @@
...
+
+      - name: Bump version.sbt and create release tag
+        if: steps.check_commit_message.outputs.skip == 'false'
+        env:
+          VERSION: ${{ steps.get_version.outputs.VERSION }}
+        run: |
+          echo "Release-As: $VERSION" >> $GITHUB_STEP_SUMMARY
+
+          echo "ThisBuild / version := \"$VERSION\"" > version.sbt
```
kpudlik, line 82: Eventually we could use `^([0-9]+)\.([0-9]+)\.([0-9]+)(?:-([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?(?:\+[0-9A-Za-z-]+)?$ ` to just replace the version in file. It gives more flexibility but I am also fine with existing solution of replacing whole file.

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,123 @@
+on:
+  push:
+    branches:
+      - release/testnet
+      - release/integrationnet
+      - release/mainnet
+
+name: Release please
```
kpudlik, line 8: Maybe just `Release`? Since we do not use release-please anymore. Same with file name.

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,123 @@
...
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  build:
+    name: Release please
```
kpudlik, line 16: aa

kpudlik (changes requested):

.github/workflows/release-please.yml
marcinwadon: I don't think so. There is only one configuration for github-actions bot, it won't be changed ever.

kpudlik (approved):

scasplte2 (approved):

PR #924, opened 2024-08-10
fix: adding retry mechanism to fetch global snapshots from another peer if it fails during rollback

IPadawans:
### Changes
+ Adding a retry mechanism of 10 times when not find the snapshot in a peer, during rollback.
+ This error was finishing the metagraph restart when we fetch from a peer that might not contain the global snapshot per hash
+ I've also added more logs to follow during the traverse.

### Tests
It's working as expected
Captura de Tela 2024-08-10 às 09 25 57
Captura de Tela 2024-08-10 às 09 39 29

### DIFF
--- a/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/programs/Rollback.scala
+++ b/modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/programs/Rollback.scala
@@ -10,6 +10,7 @@ import org.tessellation.currency.dataApplication.storage.CalculatedStateLocalFil
 import org.tessellation.currency.dataApplication.{BaseDataApplicationL0Service, DataApplicationTraverse, L0NodeContext}
 import org.tessellation.currency.l0.node.IdentifierStorage
 import org.tessellation.currency.l0.snapshot.CurrencyConsensusManager
+import org.tessellation.currency.l0.snapshot.programs.Download.CannotFetchSnapshot
 import org.tessellation.currency.l0.snapshot.schema.{CurrencyConsensusOutcome, Finished}
 import org.tessellation.currency.schema.currency.{CurrencyIncrementalSnapshot, CurrencySnapshotContext, CurrencySnapshotInfo}
 import org.tessellation.json.JsonSerializer
@@ -70,7 +71,20 @@ object Rollback {
 
       _ <- dataApplication.map {
         case ((da, cs)) =>
-          val fetchSnapshot: Hash => F[Option[Hashed[GlobalIncrementalSnapshot]]] = (hash: Hash) => globalL0Service.pullGlobalSnapshot(hash)
+          val attempts_to_fetch_global_snapshot = 10
+          val fetchSnapshot: Hash => F[Option[Hashed[GlobalIncrementalSnapshot]]] = (hash: Hash) => {
+            def attemptFetch(remainingAttempts: Int): F[Option[Hashed[GlobalIncrementalSnapshot]]] =
+              globalL0Service.pullGlobalSnapshot(hash).flatMap {
+                case Some(snapshot) => snapshot.some.pure
+                case None if remainingAttempts > 1 =>
+                  logger.warn("Could not get global snapshot, trying again with another peer") >> attemptFetch(remainingAttempts - 1)
+                case None =>
+                  logger.error("Could not get global snapshot after 10 tries") >>
+                    CannotFetchSnapshot.raiseError[F, Option[Hashed[GlobalIncrementalSnapshot]]]
+              }
+
+            attemptFetch(attempts_to_fetch_global_snapshot)
+          }
 
           DataApplicationTraverse.make[F](globalSnapshot, fetchSnapshot, da, cs, identifier).flatMap { dat =>
             dat.loadChain().flatMap {
--- a/modules/shared/src/main/scala/org/tessellation/currency/dataApplication/DataApplicationTraverse.scala
+++ b/modules/shared/src/main/scala/org/tessellation/currency/dataApplication/DataApplicationTraverse.scala
@@ -150,7 +150,8 @@ object DataApplicationTraverse {
                         }
 
                       case _ =>
-                        updateCache(snapshot).as(tail.asLeft[Output])
+                        logger.info(s"Could not get calculated state of ordinal: ${snapshot.ordinal}, updating cache") >>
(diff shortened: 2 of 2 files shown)


modules/shared/src/main/scala/org/tessellation/currency/dataApplication/DataApplicationTraverse.scala
```diff
@@ -169,7 +170,8 @@ object DataApplicationTraverse {
                   )).raiseError[F, Either[Acc, Output]]
                 case Validated.Valid(snapshots) =>
                   logger.info(
-                    s"Found ${snapshots.size.show} snapshots at global snapshot ordinal=${globalSnapshot.ordinal.show}, performing nested recursion."
+                    s"Found ${snapshots.size.show} snapshots at global snapshot ordinal=${globalSnapshot.ordinal.show}, Ordinals: ${snapshots
+                        .map(_.ordinal)} performing nested recursion."
```
marcinwadon, line 174: `_.ordinal.show`

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/programs/Rollback.scala
```diff
@@ -70,7 +71,20 @@ object Rollback {
 
       _ <- dataApplication.map {
         case ((da, cs)) =>
-          val fetchSnapshot: Hash => F[Option[Hashed[GlobalIncrementalSnapshot]]] = (hash: Hash) => globalL0Service.pullGlobalSnapshot(hash)
+          val attempts_to_fetch_global_snapshot = 10
```
marcinwadon, line 74: please don't use snake case in scala

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/programs/Rollback.scala
```diff
@@ -70,7 +71,20 @@ object Rollback {
...
+          val fetchSnapshot: Hash => F[Option[Hashed[GlobalIncrementalSnapshot]]] = (hash: Hash) => {
+            def attemptFetch(remainingAttempts: Int): F[Option[Hashed[GlobalIncrementalSnapshot]]] =
+              globalL0Service.pullGlobalSnapshot(hash).flatMap {
+                case Some(snapshot) => snapshot.some.pure
+                case None if remainingAttempts > 1 =>
+                  logger.warn("Could not get global snapshot, trying again with another peer") >> attemptFetch(remainingAttempts - 1)
+                case None =>
+                  logger.error("Could not get global snapshot after 10 tries") >>
```
marcinwadon, line 82: message is not correlated with the variable, so it's easy to make a mistake (changing one place but not the other)

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/programs/Rollback.scala
```diff
@@ -70,7 +71,20 @@ object Rollback {
...
       _ <- dataApplication.map {
         case ((da, cs)) =>
-          val fetchSnapshot: Hash => F[Option[Hashed[GlobalIncrementalSnapshot]]] = (hash: Hash) => globalL0Service.pullGlobalSnapshot(hash)
+          val attempts_to_fetch_global_snapshot = 10
+          val fetchSnapshot: Hash => F[Option[Hashed[GlobalIncrementalSnapshot]]] = (hash: Hash) => {
+            def attemptFetch(remainingAttempts: Int): F[Option[Hashed[GlobalIncrementalSnapshot]]] =
+              globalL0Service.pullGlobalSnapshot(hash).flatMap {
+                case Some(snapshot) => snapshot.some.pure
+                case None if remainingAttempts > 1 =>
+                  logger.warn("Could not get global snapshot, trying again with another peer") >> attemptFetch(remainingAttempts - 1)
```
marcinwadon, line 80: does it try with another peer indeed? I don't see any filtering of peers

marcinwadon (changes requested):
as in comments

marcinwadon:
As I think about it - maybe the issue is not with the fetch snapshot process and your fix is just a shortcut, and the issue may be with the node status itself. There should be no possibility that node is Ready and does not answer correctly with snapshots. Or maybe we fetch from non-ready peers? Have you checked that first?

modules/currency-l0/src/main/scala/org/tessellation/currency/l0/snapshot/programs/Rollback.scala
```diff
@@ -70,7 +71,20 @@ object Rollback {
 
       _ <- dataApplication.map {
         case ((da, cs)) =>
-          val fetchSnapshot: Hash => F[Option[Hashed[GlobalIncrementalSnapshot]]] = (hash: Hash) => globalL0Service.pullGlobalSnapshot(hash)
+          val attempts_to_fetch_global_snapshot = 10
+          val fetchSnapshot: Hash => F[Option[Hashed[GlobalIncrementalSnapshot]]] = (hash: Hash) => {
+            def attemptFetch(remainingAttempts: Int): F[Option[Hashed[GlobalIncrementalSnapshot]]] =
+              globalL0Service.pullGlobalSnapshot(hash).flatMap {
+                case Some(snapshot) => snapshot.some.pure
+                case None if remainingAttempts > 1 =>
```
marcinwadon, line 79: is `None` returned only when peer explicitly says there is no such snapshot, or is `None` returned only when there was an error (like the node was unresponsive or there were some networking issues)?

AlexBrandes:
> As I think about it - maybe the issue is not with the fetch snapshot process and your fix is just a shortcut, and the issue may be with the node status itself. There should be no possibility that node is Ready and does not answer correctly with snapshots. Or maybe we fetch from non-ready or unresponsive peers? Have you checked that first?

In the cases that Marcus and I looked at together the nodes were in Ready state but stuck at an old ordinal, likely out of consensus but not removed from the network. The requests for snapshots were 404ing. Requests for /global-snapshots/latest (as a debug step) were succeeding but ordinal was not increasing.

I would defer to your judgement overall here but I would suggest we still include fallback behavior like this even if it should be impossible for a node to be in Ready state and not respond correctly. A retry with the issue logged would let us recover gracefully in the case of some future unknown bug even if we fix the root cause of this one.

marcinwadon:
I have opened #925 that uses cats-retry, so I'm closing this PR.

PR #928, opened 2024-08-22
feat: path generator for hash and ordinal nesting

marcinwadon:


modules/shared/src/main/scala/org/tessellation/path/PathGenerator.scala
```diff
@@ -0,0 +1,44 @@
+package org.tessellation.path
```
kpudlik, line 1: How about `org.tessellation.storage`?

kpudlik (approved):

modules/shared/src/main/scala/org/tessellation/path/PathGenerator.scala
marcinwadon: :+1:

kpudlik (approved):

scasplte2 (approved):

PR #936, opened 2024-09-05
feat!: optimized incremental snapshot storage

marcinwadon:
BREAKING CHANGE: Incremental snapshots would be stored by hash and ordinal in a more optimized way.
See https://github.com/Constellation-Labs/snapshot-migration for migration.

### DIFF
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
@@ -16,9 +16,11 @@ import io.constellationnetwork.schema.{GlobalIncrementalSnapshot, GlobalSnapshot
 import io.constellationnetwork.security.Hasher
 import io.constellationnetwork.security.hash.Hash
 import io.constellationnetwork.security.signature.Signed
-import io.constellationnetwork.storage.SerializableLocalFileSystemStorage
+import io.constellationnetwork.storage.PathGenerator._
+import io.constellationnetwork.storage.{PathGenerator, SerializableLocalFileSystemStorage}
 
 import better.files.File
+import fs2.Stream
 import fs2.io.file.Path
 import io.circe.{Decoder, Encoder}
 import io.estatico.newtype.ops._
@@ -30,6 +32,10 @@ abstract class SnapshotLocalFileSystemStorage[
 ](
   path: Path
 ) extends SerializableLocalFileSystemStorage[F, Signed[S]](path) {
+
+  val ordinalChunkSize = ChunkSize(20000)
+  val hashPathGenerator = PathGenerator.forHash(Depth(2), PrefixSize(3))
+  val ordinalPathGenerator = PathGenerator.forOrdinal(ordinalChunkSize)
 
   private val logger = Slf4jLogger.getLogger[F]
 
@@ -62,19 +68,19 @@ abstract class SnapshotLocalFileSystemStorage[
     read(toOrdinalName(ordinal))
 
   def read(hash: Hash): F[Option[Signed[S]]] =
-    read(hash.coerce[String])
+    read(toHashName(hash))
 
   def exists(ordinal: SnapshotOrdinal): F[Boolean] =
     exists(toOrdinalName(ordinal))
 
   def exists(hash: Hash): F[Boolean] =
-    exists(hash.coerce[String])
+    exists(toHashName(hash))
 
   def delete(ordinal: SnapshotOrdinal): F[Unit] =
     delete(toOrdinalName(ordinal))
 
   def getPath(hash: Hash): F[File] =
-    getPath(hash.coerce[String])
+    getPath(toHashName(hash))
 
   def getPath(snapshot: Signed[S])(implicit hasher: Hasher[F]): F[File] =
     toHashName(snapshot.value).flatMap { hashName =>
@@ -82,7 +88,7 @@ abstract class SnapshotLocalFileSystemStorage[
     }
 
   def move(hash: Hash, to: File): F[Unit] =
-    move(hash.coerce[String], to)
+    move(toHashName(hash), to)
 
   def move(snapshot: Signed[S], to: File)(implicit hasher: Hasher[F]): F[Unit] =
     toHashName(snapshot.value).flatMap { hashName =>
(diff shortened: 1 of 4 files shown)


scasplte2 (approved):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
```diff
@@ -31,6 +33,10 @@ abstract class SnapshotLocalFileSystemStorage[
   path: Path
 ) extends SerializableLocalFileSystemStorage[F, Signed[S]](path) {
 
+  val ordinalChunkSize = ChunkSize(20000)
```
IPadawans, line 36: Can we move this number to be in configuration?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
```diff
@@ -31,6 +33,10 @@ abstract class SnapshotLocalFileSystemStorage[
...
+  val hashPathGenerator = PathGenerator.forHash(Depth(2), PrefixSize(3))
```
IPadawans, line 37: Can we move these numbers to be in the configuration?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
```diff
@@ -97,11 +103,50 @@ abstract class SnapshotLocalFileSystemStorage[
...
+          }
+        )
+
+    go(baseDirectory)
+  }
+
   private def toOrdinalName(snapshot: S): String = toOrdinalName(snapshot.ordinal)
-  private def toOrdinalName(ordinal: SnapshotOrdinal): String = ordinal.value.value.toString
+  private def toOrdinalName(ordinal: SnapshotOrdinal): String =
+    "ordinal/" + ordinalPathGenerator.get(ordinal.value.value.toString)
```
IPadawans, line 143: Instead 
ordinal.value.value.toString
maybe
ordinal.show?

modules/node-shared/src/test/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorageSuite.scala
```diff
@@ -24,6 +26,9 @@ import SnapshotLocalFileSystemStorage.UnableToPersistSnapshot
 
 object SnapshotLocalFileSystemStorageSuite extends MutableIOSuite with Checkers {
 
+  val hashPathGenerator = PathGenerator.forHash(Depth(2), PrefixSize(3))
```
IPadawans, line 29: Can we move this number to be in the configuration?

modules/node-shared/src/test/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorageSuite.scala
```diff
@@ -24,6 +26,9 @@ import SnapshotLocalFileSystemStorage.UnableToPersistSnapshot
...
+  val ordinalPathGenerator = PathGenerator.forOrdinal(ChunkSize(20000))
```
IPadawans, line 30: Can we move these numbers to be in the configuration?

modules/node-shared/src/test/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorageSuite.scala
```diff
@@ -55,19 +60,19 @@ object SnapshotLocalFileSystemStorageSuite extends MutableIOSuite with Checkers
   ): IO[UnableToPersistSnapshot] =
     snapshot.toHashed.map(hashed =>
       UnableToPersistSnapshot(
-        hashed.ordinal.value.toString,
-        hashed.hash.value,
+        "ordinal/" + ordinalPathGenerator.get(hashed.ordinal.value.toString),
```
IPadawans, line 63: Instead
ordinal.value.value.toString
maybe
ordinal.show?

modules/node-shared/src/test/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorageSuite.scala
```diff
@@ -55,19 +60,19 @@ object SnapshotLocalFileSystemStorageSuite extends MutableIOSuite with Checkers
...
 
   private def mkHashFile(tmpDir: File, snapshot: Signed[GlobalIncrementalSnapshot])(
     implicit H: Hasher[IO]
   ): IO[File] =
-    snapshot.toHashed.map(hashed => tmpDir / hashed.hash.value)
+    snapshot.toHashed.map(hashed => tmpDir / "hash" / hashPathGenerator.get(hashed.hash.value))
 
   private def mkOrdinalFile(tmpDir: File, snapshot: Signed[GlobalIncrementalSnapshot]): File =
-    tmpDir / snapshot.ordinal.value.value.toString
+    tmpDir / "ordinal" / ordinalPathGenerator.get(snapshot.ordinal.value.value.toString)
```
IPadawans, line 75: Instead
ordinal.value.value.toString
maybe
ordinal.show?

IPadawans (changes requested):
Just a couple points, otherwise looks good 👍🏼

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
marcinwadon: I thought about that and I can do it, but I decided not to. This way, it will be compiled into the code and it won't be possible to change it from configuration (per node) which would help managing the same structure of files across nodes (and for nodectl/archiver too)

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
marcinwadon: No, `show` is for showing/logging purpose. It adds additional formatting to the string that we do not want here

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
IPadawans: Hmm, got it, no problem then, I was just curious

modules/node-shared/src/test/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorageSuite.scala
marcinwadon: as above, however, by explicitly specifying it here too, the test would detect if we would change the configuration by mistake in the snapshot storage class

IPadawans: 👍🏼

IPadawans (dismissed):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
kpudlik: You can still have it in the configuration classes but have it hardcoded there 🤔  I think we used that pattern before.

kpudlik (dismissed):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
marcinwadon: @kpudlik fair point, I'll take a look

scasplte2 (approved):

kpudlik (approved):

PR #938, opened 2024-09-11
feat: allow spend schema

marcinwadon:


modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala
```diff
@@ -36,8 +94,51 @@ object swap {
...
+  trait AllowSpendStatusCodecs {
+    implicit val encode: Encoder[AllowSpendStatus] = Encoder.encodeString.contramap[AllowSpendStatus](_.entryName)
+    implicit val decode: Decoder[AllowSpendStatus] =
+      Decoder.decodeString.emapTry(s => Try(AllowSpendStatus.withName(s)))
+  }
+
+  @derive(decoder, encoder, show)
+  case class SwapBlock(
+    roundId: RoundId,
+    transactions: NonEmptySet[Signed[SwapTransaction]]
```
scasplte2, line 130: Perhaps I am missing something but who would generate the proofs associated with the `SwapTransaction`? My understanding is that these transaction should be automatically generated by the metagraph when there are matching `AllowSpend` transactions within the order book maintained by the AMM metagraph.

marcinwadon: Good question, but I don't know the answer. I was following our old meetings' notes. Maybe
@ryle-goehausen @AlexBrandes?

ryle-goehausen: Hey, I would reference the Currency Swapping notion doc for latest information. As far as I can tell this is just a naming issue with the trait SwapTransaction. We've eliminated MetagraphAllowSpend in favor of SpendTransaction (generated via metagraph snapshot embedded). So I think the user defined AllowSpend is fine to use with Signed[] (since it's wallet produced.) If we have the final SpendTransaction it won't conform to this trait though since the signatures are associated with the metagraph snapshot.

SwapTransaction might be more appropriate name internal to the AMM logic, which I think is what James is referring to just to clarify.

marcinwadon: @ryle-goehausen do I understand correctly that AllowSpend Transaction should be Signed by individuals and SpendTransaction won't be ever signed?

scasplte2: yes, that was my understanding. The `SwapTransaction` (or whichever transaction actually changes balances) is meant to be verified via an inclusion proof within a `CurrencySnapshot`, while an `AllowSpend` is a user generated "authorization" that can be verified by a signature proof.

marcinwadon: Hmm, it complicates things a little bit as we talked about making those transactions generic, meaning they need to follow the same format/flow. Looks like there is no point to merge allow spends and spend txs under the same trait. 

The other solution would be to sign it by all the peers under the consensus but it doesn't make much sense. It may work like reward transactions (so no signature at all)

kpudlik (approved):

PR #939, opened 2024-09-12
feat: rejoining to correct fork

kpudlik:


modules/currency-l1/src/main/scala/io/constellationnetwork/currency/l1/cli/method.scala
```diff
@@ -89,6 +91,22 @@ object method {
     }
   }
 
+  case class RunRejoinToCorrectFork(
```
marcinwadon, line 94: I'd rename it to `RunValidatorWithJoin` or `RunValidatorWithJoinAttempt`

marcinwadon (changes requested):
One comment, otherwise looks legit

marcinwadon (approved):

marcinwadon (approved):

PR #943, opened 2024-09-16
feat: adding hash to fee transaction schema

IPadawans:


IPadawans:
I've preferred to call dataUpdateHash to be more descriptive. I feel that only hash is not detailed enough so users can understand that it refers to the dataUpdateHash

scasplte2 (approved):

modules/shared/src/main/scala/org/tessellation/currency/schema/feeTransaction.scala
```diff
@@ -21,7 +21,8 @@ object feeTransaction {
     destination: Address,
     amount: Amount,
     parent: FeeTransactionReference,
-    salt: TransactionSalt
+    salt: TransactionSalt,
+    dataUpdateHash: Hash
```
marcinwadon, line 25: I think that `dataUpdate` would be descriptive enough. You don't need to put `hash` in the name, the same way you don't add `sourceAddress` or `destinationAddress`

IPadawans: 👍🏼

scasplte2 (approved):

PR #944, opened 2024-09-19
feat: allow spend L1 consensus

marcinwadon:


modules/shared/src/main/scala/io/constellationnetwork/currency/swap/ConsensusInput.scala
```diff
@@ -0,0 +1,60 @@
...
+    val owner: PeerId
+  }
+
+  @derive(encoder, decoder)
+  case class Proposal(
+    roundId: RoundId,
+    senderId: PeerId,
+    owner: PeerId,
+    facilitators: Set[PeerId],
+    transactions: Set[Signed[SwapTransaction]]
```
scasplte2, line 35: is the swap transaction signed by the proposing peer?

marcinwadon: I don't know who should sign it. @ryle-goehausen? Here, I just need a proof that the transaction is correctly signed.

scasplte2: I think this is alright given that the only swap transaction is AllowSpend, which is user generated. It is the balance update request that the AMM generates that has to be treated as rewards.

scasplte2 (approved):

PR #945, opened 2024-09-23
feat: allow spend L1 consensus trigger

marcinwadon:


modules/currency-l1/src/main/scala/io/constellationnetwork/currency/l1/http/SwapRoutes.scala
```diff
@@ -0,0 +1,91 @@
...
+    // TODO: @mwadon - validation
+    onValid
+
+  protected val public: HttpRoutes[F] = HttpRoutes.of[F] {
+    case GET -> Root / "swap" =>
+      swapTransactionsQueue.view.flatMap(Ok(_))
+
+    case req @ POST -> Root / "swap" =>
+      req
+        .asR[Signed[SwapTransaction]] {
```
scasplte2, line 55: Perhaps I am still misunderstanding but I thought the SwapTransaction was to be treated like Rewards and not generated by the end-user? Doesn't this structure require the user to sign and generate the swap transaction?

scasplte2: Perhaps the `SwapTransaction` is what was previously called AllowSpend? In that case, I understand. I was thinking this is the balance updating transaction and not the "user authorization" portion of the process.

marcinwadon: Yes, it's a trait above the AllowSpend. @AlexBrandes wanted to make it possible to put new types of transactions in the future. Maybe SwapTransaction is a bit unfortunate name, we can think of something better.

marcinwadon: I think we should definitely refine and write down naming conventions for AMM 🤔

scasplte2: yes, I was able to find it [here](https://github.com/Constellation-Labs/tessellation/blob/8dde3dd93cae65bb17d77fd7ca47e797962c8cac/modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala#L88). Thanks for clarifying as well.

scasplte2 (approved):

IPadawans (approved):

scasplte2 (approved):

PR #947, opened 2024-10-04
feat: adding spend transactions to currency incremental snapshot schema

IPadawans:
### Changes
+ Adding spendTransactions to the CurrencySnapshot schema
+ Updated field names to match notion doc

### DIFF
--- a/modules/shared/src/main/scala/io/constellationnetwork/currency/schema/currency.scala
+++ b/modules/shared/src/main/scala/io/constellationnetwork/currency/schema/currency.scala
@@ -16,7 +16,7 @@ import io.constellationnetwork.schema.epoch.EpochProgress
 import io.constellationnetwork.schema.height.{Height, SubHeight}
 import io.constellationnetwork.schema.semver.SnapshotVersion
 import io.constellationnetwork.schema.snapshot._
-import io.constellationnetwork.schema.swap.{AllowSpend, AllowSpendReference}
+import io.constellationnetwork.schema.swap._
 import io.constellationnetwork.schema.transaction._
 import io.constellationnetwork.security._
 import io.constellationnetwork.security.hash.{Hash, ProofsHash}
@@ -162,6 +162,7 @@ object currency {
     dataApplication: Option[DataApplicationPart] = None,
     messages: Option[SortedSet[Signed[CurrencyMessage]]] = None,
     feeTransactions: Option[SortedSet[Signed[FeeTransaction]]] = None,
+    spendTransactions: Option[SortedSet[SpendTransaction]] = None,
     version: SnapshotVersion = SnapshotVersion("0.0.1")
   ) extends IncrementalSnapshot[CurrencySnapshotStateProof]
 
@@ -179,6 +180,7 @@ object currency {
           stateProof.toCurrencySnapshotStateProof,
           snapshot.epochProgress,
           snapshot.dataApplication,
+          None,
           None,
           None,
           snapshot.version
@@ -212,6 +214,7 @@ object currency {
         stateProof.toCurrencySnapshotStateProof,
         epochProgress,
         dataApplication,
+        None,
         None,
         None,
         version
@@ -265,6 +268,7 @@ object currency {
           stateProof.toCurrencySnapshotStateProof,
           genesis.epochProgress,
           genesis.dataApplication,
+          None,
           None,
           None,
           genesis.version
--- a/modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala
+++ b/modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala
@@ -15,6 +15,7 @@ import io.constellationnetwork.ext.crypto._
 import io.constellationnetwork.ext.derevo.ordering
 import io.constellationnetwork.schema.address.Address
 import io.constellationnetwork.schema.balance.Amount
+import io.constellationnetwork.schema.epoch.EpochProgress
 import io.constellationnetwork.schema.round.RoundId
 import io.constellationnetwork.security.hash.Hash
 import io.constellationnetwork.security.signature.Signed
@@ -24,7 +25,7 @@ import derevo.cats.{eqv, order, show}
 import derevo.circe.magnolia.{decoder, encoder}
 import derevo.derive
 import enumeratum._
-import eu.timepit.refined.auto.{autoRefineV, autoUnwrap, _}
+import eu.timepit.refined.auto.{autoRefineV, _}
 import eu.timepit.refined.cats._
 import eu.timepit.refined.types.numeric.{NonNegLong, PosLong}
 import io.circe.{Decoder, Encoder}
@@ -158,14 +159,30 @@ object swap {
   @newtype
   case class SpendTransactionFee(value: PosLong)
 
+  @derive(decoder, encoder, order, ordering, show)
+  sealed trait SpendTransaction
+
   @derive(decoder, encoder, order, show)
-  case class SpendTransaction(
+  @newtype
+  case class SpendTransactionReference(hash: Hash)
+
+  object SpendTransactionReference {
+    def of[F[_]: Async](spendTransaction: SpendTransaction)(implicit hasher: Hasher[F]): F[SpendTransactionReference] =
+      hasher.hash(spendTransaction).map(SpendTransactionReference(_))
+  }
+  @derive(decoder, encoder, order, show)
+  case class PendingSpendTransaction(
     fee: SpendTransactionFee,
-    lastValidOrdinal: SnapshotOrdinal,
+    lastValidEpochProgress: EpochProgress,
     allowSpendRef: Hash,
     currency: Option[CurrencyId],
     amount: SwapAmount
-  )
+  ) extends SpendTransaction
+
+  @derive(decoder, encoder, order, show)
+  case class ConcludedSpendTransaction(
+    spendTransactionRef: SpendTransactionReference
+  ) extends SpendTransaction
 
   sealed trait SwapAction
 


modules/shared/src/main/scala/io/constellationnetwork/currency/schema/currency.scala
```diff
@@ -162,6 +162,7 @@ object currency {
     dataApplication: Option[DataApplicationPart] = None,
     messages: Option[SortedSet[Signed[CurrencyMessage]]] = None,
     feeTransactions: Option[SortedSet[Signed[FeeTransaction]]] = None,
+    spendTransactions: Option[SortedSet[SpendTransaction]] = None,
```
marcinwadon, line 165: Does the SpendTransaction should be a part of Currency snapshot?

modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala
```diff
@@ -161,12 +161,17 @@ object swap {
...
-    lastValidOrdinal: SnapshotOrdinal,
+    validUntilOrdinal: SnapshotOrdinal,
     allowSpendRef: Hash,
-    currency: Option[CurrencyId],
+    currencyId: Option[CurrencyId],
     amount: SwapAmount
   )
 
+  object SpendTransaction {
+    implicit val ordering: Ordering[SpendTransaction] =
```
marcinwadon, line 171: It shouldn't be an implicit ordering. Because of that line, two different allow spends with the same validUntilOrdinal would be considered equal. Ordering on partial fields should be defined per use case

modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala
```diff
@@ -161,12 +161,17 @@ object swap {
   @derive(decoder, encoder, order, show)
   case class SpendTransaction(
     fee: SpendTransactionFee,
-    lastValidOrdinal: SnapshotOrdinal,
+    validUntilOrdinal: SnapshotOrdinal,
     allowSpendRef: Hash,
-    currency: Option[CurrencyId],
+    currencyId: Option[CurrencyId],
```
marcinwadon, line 166: IMO currency was a better name. We see it's an identifier by its type.

modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala
```diff
@@ -161,12 +161,17 @@ object swap {
   @derive(decoder, encoder, order, show)
   case class SpendTransaction(
     fee: SpendTransactionFee,
-    lastValidOrdinal: SnapshotOrdinal,
+    validUntilOrdinal: SnapshotOrdinal,
```
marcinwadon, line 164: "Until" means inclusive or exclusive? "lastValidOrdinal" wasn't ambiguous

marcinwadon (changes requested):

modules/shared/src/main/scala/io/constellationnetwork/currency/schema/currency.scala
IPadawans: Yes
https://constellationnetwork.atlassian.net/browse/PROT-843

modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala
IPadawans: I was following the names described in the notion docs
https://www.notion.so/Currency-Swapping-cf462ad9c01d4e93902429282e7267e8#e1668bee754d49bb9f8cfb2001fde0ea

modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala
IPadawans: 👍🏼 
However, I need the 
`@derive(..., ordering, ...)`
otherwise it will fail the SortedSet

modules/shared/src/main/scala/io/constellationnetwork/currency/schema/currency.scala
IPadawans: @marcinwadon we need to align how we will populate this. We need to somehow extract this in the output of `combine` function.

marcinwadon: I know the ticket but it may be outdated. I think the most recent one is the sequence diagram we've discussed together on the meeting. We're still discussing how metagraph would scan the network, so it may be that it would have to be aggregated on GL0 or maybe we would have more than one tote for marking the swap. I'd wait with it until we have the process well described

modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala
marcinwadon: Notion doc isn't up to date for sure

modules/shared/src/main/scala/io/constellationnetwork/currency/schema/currency.scala
marcinwadon: I mean, I'd like to make schema update once, so let's make sure it's correct and covers our flow well

marcinwadon: Re: how to populate - maybe it needs to go a different path than the combine fn -> enqueued directly as an event to currency consensus, as allow spends on GL0.

IPadawans: Got it @marcinwadon . @ryle-goehausen has described that we will have the SpendTransaction in GlobalSnapshots and CurrencySnapshots. So, I think we can wait, what do you think @ryle-goehausen ?

modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala
IPadawans: 👍🏼

modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala
IPadawans: 👍🏼

ryle-goehausen: Fine with updating notion doc, really either is fine here.

modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala
marcinwadon: no, I think you don't. You need it indeed in a place you create a sorted set, but not globally

marcinwadon: and, it will be used to deduplicate entries. so keep in mind that "two different allow spends with the same validUntilOrdinal would be considered equal"

modules/shared/src/main/scala/io/constellationnetwork/currency/schema/currency.scala
```diff
@@ -162,6 +165,7 @@ object currency {
...
```
marcinwadon, line 168: I think we need to keep spend transactions as Signed (signatures should be combined from all nodes taking part in the consensus). We would have to validate if the proper AMM was an issuer of the SpendTransaction. Especially that we want to keep it outside of the currency snapshot it belongs to (in GlobalSnapshotInfo).

modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala
```diff
@@ -158,14 +159,35 @@ object swap {
   @newtype
   case class SpendTransactionFee(value: PosLong)
 
+  @derive(decoder, encoder, order, ordering, show)
+  sealed trait SpendTransaction extends SwapTransaction
```
marcinwadon, line 163: I think it shouldn't extend the SwapTransaction

modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala
```diff
@@ -158,14 +159,35 @@ object swap {
...
+  @newtype
+  case class SpendTransactionReference(hash: Hash)
+
+  object SpendTransactionReference {
+    def of[F[_]: Async](spendTransaction: SpendTransaction)(implicit hasher: Hasher[F]): F[SpendTransactionReference] =
+      hasher.hash(spendTransaction).map(SpendTransactionReference(_))
+  }
   @derive(decoder, encoder, order, show)
-  case class SpendTransaction(
+  case class CreatedSpendTransaction(
```
marcinwadon, line 174: Maybe PendingSpendTransaction? (or Waiting)

modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala
```diff
@@ -158,14 +159,35 @@ object swap {
...
-    lastValidOrdinal: SnapshotOrdinal,
+    lastValidEpochProgress: EpochProgress,
     allowSpendRef: Hash,
     currency: Option[CurrencyId],
     amount: SwapAmount
-  )
+  ) extends SpendTransaction
+
+  @derive(decoder, encoder, order, show)
+  case class PendingSpendTransaction(
```
marcinwadon, line 183: What is a difference between Created and Pending? I thought we just need Pending and Completed

marcinwadon (changes requested):

modules/shared/src/main/scala/io/constellationnetwork/currency/schema/currency.scala
IPadawans: 👍🏼

modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala
IPadawans: 👍🏼

modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala
IPadawans: 👍🏼

modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala
IPadawans: 👍🏼

modules/shared/src/main/scala/io/constellationnetwork/currency/schema/currency.scala
```diff
@@ -79,7 +80,8 @@ object currency {
     lastMessages: Option[SortedMap[MessageType, Signed[CurrencyMessage]]],
     lastFeeTxRefs: Option[SortedMap[Address, TransactionReference]],
     lastAllowSpendRefsProof: Option[SortedMap[Address, AllowSpendReference]],
-    activeAllowSpends: Option[SortedMap[Address, List[Signed[AllowSpend]]]]
+    activeAllowSpends: Option[SortedMap[Address, List[Signed[AllowSpend]]]],
+    activeSpendTransactions: Option[SortedMap[Address, List[SpendTransaction]]]
```
marcinwadon, line 84: I think we need to keep them per amm metagraph, so you should wrap it by another map of addresses. Also, we would need probably the observed ordinal as for allow spends (@kpudlik?)

marcinwadon (commented):
All good, except the aggregation schema

modules/shared/src/main/scala/io/constellationnetwork/currency/schema/currency.scala
marcinwadon: Ok, I see it's a schema of currency, so this one is correct. Schema in GlobalSnapshotInfo would be different

marcinwadon: but, `SpendTransaction` does not have a `source` address or anything like that. What is this `Address` type in  `Option[SortedMap[Address, List[SpendTransaction]]]`?

marcinwadon (approved):

PR #948, opened 2024-10-08
fix: catch exceptions during DataApp onSnapshotConsensusResult

scasplte2:
## Summary
Invocation of `onSnapshotConsensusResult` may throw an exception due to improper error handling by the metagraph developer. If an exception is thrown the underlying consensus process enters a stalled state and the ML0 application must be restarted to recover.

## Changes
- Add error handling to log the exception and return an effect for the state advancement

## Testing
- Tested against a local metagraph with exceptions during `onSnapshotConsensusResult` and consensus process is no longer stalled

### DIFF
--- a/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensusStateAdvancer.scala
+++ b/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensusStateAdvancer.scala
@@ -294,7 +294,7 @@ object CurrencySnapshotConsensusStateAdvancer {
                         gossipForkInfo(gossip, signedMajorityArtifact) >>
                         maybeDataApplication.traverse_ { da =>
                           signedMajorityArtifact.toHashed >>= da.onSnapshotConsensusResult
-                        }
+                        }.handleErrorWith(logger.error(_)("Unhandled exception during onSnapshotConsensusResult"))
 
                       newState = state.copy(status =
                         identity[CurrencySnapshotStatus](


modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensusStateAdvancer.scala
```diff
@@ -294,7 +294,7 @@ object CurrencySnapshotConsensusStateAdvancer {
                         gossipForkInfo(gossip, signedMajorityArtifact) >>
                         maybeDataApplication.traverse_ { da =>
                           signedMajorityArtifact.toHashed >>= da.onSnapshotConsensusResult
-                        }
+                        }.handleErrorWith(logger.error(_)("Unhandled exception during onSnapshotConsensusResult"))
```
marcinwadon, line 297: What happens with the consensus then? I believe the consensus shouldn't progress if there is non-recoverable issue in data application. Right now the error would be "swallowed" and what then?

marcinwadon:
FYI - there is no need to label PRs since the PR title contains the scope and labels were used for changelog generation (no longer used)

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensusStateAdvancer.scala
scasplte2: The `onSnapshotConsensusResult` is just a side effecting operation and does not return a value for consensus so I thought it would be appropriate to handle in this way. The issue that caused this was that I was trying to push data to an external API in this method but occasionally the library invoking the API would throw an exception. It would seem more fail-proof to have the metagraph developer add in additional exception handling in their implementation if this method is critical to the functioning of the application, rather than the case now where a poorly designed method can completely crash the application.

scasplte2: well not "crash" but it completely stalls the consensus and the only log messages that begin to appear are something like "Facilitating consensus..." one after another. This makes it unclear even what happened so that the developer could identify what portion of the application to debug.

marcinwadon: ok, got it :+1: if that's fire and forget, then it seems fine

marcinwadon (approved):

PR #949, opened 2024-10-09
feat:  move locked addresses to config

gclaramunt:
Changes: move locked addresses from TransactionValidator to application.,conf

### DIFF
--- a/modules/node-shared/src/main/resources/application.conf
+++ b/modules/node-shared/src/main/resources/application.conf
@@ -99,4 +99,10 @@ fee-configs {
     }
   }
 }
-
+addresses {
+  locked: [
+    DAGSTARDUSTCOLLECTIVEHZOIPHXZUBFGNXWJETZVSPAPAHMLXS, //stardust address
+    DAG0qgcEbMk8vQL6VrnbhMreNeEFXk12v1BvERCb,
+    DAG2KQrN97LpA5gRerJAQ5mDuy6kjC2dDtMr58fe,
+  ]
+}
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/transaction/TransactionValidator.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/transaction/TransactionValidator.scala
@@ -5,6 +5,7 @@ import cats.effect.Async
 import cats.syntax.all._
 
 import io.constellationnetwork.ext.cats.syntax.validated._
+import io.constellationnetwork.node.shared.config.types.AddressesConfig
 import io.constellationnetwork.node.shared.domain.transaction.TransactionValidator.TransactionValidationErrorOr
 import io.constellationnetwork.schema.address.Address
 import io.constellationnetwork.schema.transaction.Transaction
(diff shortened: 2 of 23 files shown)


modules/node-shared/src/main/resources/application.conf
```diff
@@ -100,3 +100,10 @@ fee-configs {
   }
 }
 
+
+addresses {
+    locked: [
+        DAG0qgcEbMk8vQL6VrnbhMreNeEFXk12v1BvERCb,
+        DAG2KQrN97LpA5gRerJAQ5mDuy6kjC2dDtMr58fe,
+        ]
+}
```
marcinwadon, line 109: looks like bad indent

modules/node-shared/src/main/resources/application.conf
```diff
@@ -100,3 +100,10 @@ fee-configs {
   }
 }
 
+
+addresses {
```
kpudlik, line 104: Do we guarantee the consistency of such config entries in the cluster? I can imagine one node not setting these locked addresses in config so it will accept transactions from these addresses.

Afaik we do not validate locked addresses during consensus - we only do that on the endpoint when transactions are persisted in `TransactionStorage`. I think we should double-check if we reject transactions/blocks with addresses that are locked during consensus.

kpudlik (changes requested):

modules/node-shared/src/main/resources/application.conf
kpudlik: There is a `private val addressesLockedAtOrdinal: Map[Address, SnapshotOrdinal] = Map.empty` in `BlockValidator` and I have a feeling that we should use addresses from config there as well instead of empty map 🤔

marcinwadon (dismissed):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/transaction/TransactionValidator.scala
```diff
@@ -75,6 +69,8 @@ object TransactionValidator {
         else
           signedTx.validNec[TransactionValidationError]
 
+      private val lockedAddresses = cfg.locked + stardustPrimary
```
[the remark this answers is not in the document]
gclaramunt: stardustPrimary is used in other places too (e.g. mainnetProgramsDistributionConfig)  Eventually I think moving all the hardcoded addresses to the config file is a good idea (I've already created an addresses section in config) , but for now the idea is to start with a shorter scope. See [PROT-622](https://constellationnetwork.atlassian.net/browse/PROT-622?atlOrigin=eyJpIjoiNDNkNzEyMGUwOThhNGQ3MWI5MmZkOTcwMjFkYzY1OTQiLCJwIjoiaiJ9 )

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/services/AddressService.scala
```diff
@@ -15,6 +15,7 @@ import io.estatico.newtype.ops._
 
 object AddressService {
   def make[F[_]: Applicative, S <: Snapshot, C <: SnapshotInfo[_]](
+    cfg: SharedConfig,
```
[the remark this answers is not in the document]
gclaramunt: yeah, I was on the verge on that.

modules/node-shared/src/main/resources/application.conf
gclaramunt: good question, I see transactions are validated also in processProposal of BlockConsensusCell  and in BlockValidator (among other places) but I still need to understand the whole picture :)

kpudlik: Yeah, I see that such a node's proposal will be considered as "Incorrect block" so I guess we should be fine. But worth testing on the cluster anyway :+1:

marcinwadon:
Please rebase your branch and squash all commits into a single change. Basically we aim for 1 commit per 1 PR, unless you want to push more changes related to each other under one PR. Changes to your original change should be amended or squashed (amending is easier/faster) and branch should be rebased (instead of merging).

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/transaction/TransactionValidator.scala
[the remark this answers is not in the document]
marcinwadon: Ok, but it's misleading having some configurable and some hardcoded. I see no issue moving locked all to config and still use stardust primary for other places for now

marcinwadon (commented):
I'm not going to block the PR but still, I'll just add the stardust address as a third element on the list and remove this `locked + stardust` from code. This ticket was aimed to move locked addresses to config afaik, so I agree moving distribution addresses is too wide  (and I didn't propose that). Anyway, you don't need to move stardust distribution address just to put it on lock in the config.

gclaramunt:
>  remove this locked + stardust 
My initial reasoning was to avoid having the stardust address defined twice, but removing it makes for a simpler code, so I agree.

marcinwadon (approved):

kpudlik (approved):

PR #956, opened 2024-10-17
feat: include jar hash in peer metadata

gclaramunt:
calculate fat jar hash during startup and report it in cluster data

### DIFF
--- a/modules/shared/src/main/scala/io/constellationnetwork/env/JarSignature.scala
+++ b/modules/shared/src/main/scala/io/constellationnetwork/env/JarSignature.scala
@@ -1,0 +1,43 @@
+package io.constellationnetwork.env
+
+import java.security.MessageDigest
+
+import cats.effect.Async
+import cats.implicits.{catsSyntaxFlatMapOps, toFunctorOps}
+
+import io.constellationnetwork.security.hash.Hash
+import io.constellationnetwork.security.hex.Hex
+
+import fs2.io.file.{Files, Path}
+import fs2.{Chunk, Stream}
+
+object JarSignature {
+
+  def jarHash[F[_]: Async: Files]: F[Hash] = {
+    val jarPath = JarSignature.getClass.getProtectionDomain.getCodeSource.getLocation.toURI
+    digestOf(Files[F].readAll(Path(jarPath.getPath)))
+  }
+
+  def digestOf[F[_]: Async](bytes: Stream[F, Byte]): F[Hash] = {
+
+    val streamChunkSizeBytes = 524288
+
+    digestInstance >>= { digest =>
+      bytes
+        .chunkN(streamChunkSizeBytes)
+        .evalMap(updateDigest(digest, _))
+        .compile
+        .drain >> toHash(digest)
+    }
+  }
+
+  private def digestInstance[F[_]: Async] =
+    Async[F].delay(MessageDigest.getInstance("SHA-256"))
+
+  private def updateDigest[F[_]: Async](digest: MessageDigest, chunk: Chunk[Byte]) =
+    Async[F].delay(digest.update(chunk.toArray))
+
+  private def toHash[F[_]: Async](digest: MessageDigest) =
+    Async[F].delay(digest.digest()).map(Hex.fromBytes(_)).map(_.value).map(Hash.apply)
+
+}
--- a/modules/shared/src/test/scala/io/constellationnetwork/env/JarSignatureSuite.scala
+++ b/modules/shared/src/test/scala/io/constellationnetwork/env/JarSignatureSuite.scala
@@ -1,0 +1,30 @@
+package io.constellationnetwork.env
+
+import cats.effect._
+
+import io.constellationnetwork.security.hash.Hash
+
+import fs2.Stream
+import weaver._
+import weaver.scalacheck.Checkers
+
+object JarSignatureSuite extends SimpleIOSuite with Checkers {
+
+  test("digest of empty") {
+
+    JarSignature.digestOf[IO](Stream.empty).map { hash =>
+      expect.eql(Hash("e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"), hash)
+    }
+
+  }
+
+  test("digest of data") {
+
+    val data = fs2.Stream[IO, Byte](0x01, 0x02, 0x0f, 0x14, 0x09)
+    JarSignature.digestOf[IO](data).map { hash =>
+      expect.eql(Hash("ea2c1f710fa03eb873135021041c8ebf3038b993e5c3f1fc9b9adb0e7da211ae"), hash)
+    }
+
+  }
+
+}
--- a/build.sbt
+++ b/build.sbt
@@ -44,6 +44,7 @@ lazy val commonTestSettings = Seq(
 ThisBuild / assemblyMergeStrategy := {
   case "logback.xml"                                       => MergeStrategy.first
   case x if x.contains("io.netty.versions.properties")     => MergeStrategy.discard
+  case x if x.contains("scala.semanticdb")                 => MergeStrategy.discard
   case x if x.contains("rally-version.properties")         => MergeStrategy.concat
   case PathList(xs @ _*) if xs.last == "module-info.class" => MergeStrategy.first
   case x =>
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/app/TessellationIOApp.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/app/TessellationIOApp.scala
@@ -6,6 +6,7 @@ import cats.effect._
 import cats.effect.std.{Random, Supervisor}
 import cats.syntax.all._
 
+import io.constellationnetwork.env.JarSignature
 import io.constellationnetwork.env.env._
 import io.constellationnetwork.ext.cats.effect._
 import io.constellationnetwork.ext.kryo._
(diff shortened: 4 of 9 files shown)


modules/shared/src/main/scala/io/constellationnetwork/env/JarSignature.scala
```diff
@@ -0,0 +1,21 @@
...
+import java.nio.file.Files
+
+import io.constellationnetwork.security.hash.Hash
+
+object JarSignature {
+
+  lazy val hash: Hash = jarHash
+
+  private def jarHash = {
+    // assume fat jar
```
IPadawans, line 13: We usually do not use comments over the codebase, this is necessary?

IPadawans (approved):
Just a comment, but overall LGTM

modules/shared/src/main/scala/io/constellationnetwork/env/JarSignature.scala
gclaramunt: not really, I'll  remove it.

modules/shared/src/main/scala/io/constellationnetwork/env/JarSignature.scala
```diff
@@ -0,0 +1,20 @@
...
+import io.constellationnetwork.security.hash.Hash
+
+object JarSignature {
+
+  lazy val hash: Hash = jarHash
+
+  private def jarHash = {
+    val jarFile = new File(JarSignature.getClass.getProtectionDomain.getCodeSource.getLocation.toURI)
+    if (jarFile.isFile) {
+      val buffer = Files.readAllBytes(jarFile.toPath)
```
scasplte2, line 15: The fat jar is usually around 100 MB, so this may be a bit resource intensive but our Hash assumes small preimage inputs so that would probably need to be refactored if we wanted to use it here. The other option may just be to reproduce some code and just use the bouncycastle `sha256` directly.

Another concern I had was about the reproducibility of this hash. I don't recall specifics but there may be an issue with far jars being slightly different builds between machines / environments. I believe the solution involved the `sbt reproducible builds` [plugin ](https://github.com/raboof/sbt-reproducible-builds) but the use case may have been slightly different.

marcinwadon: I agree with @scasplte2, it's a resource intensive operation, so it should be wrapped into an effect type that can efficiently handle cpu-intensive operations. @gclaramunt we do prefer functional style in the codebase (cats, cats-effect and fs2). Please take a look how it works and how it handles file i/o. reading bytes from file should be implemented using FS2, so in a streaming fashion, instead of single blocking operation.

marcinwadon (changes requested):

modules/shared/src/main/scala/io/constellationnetwork/env/JarSignature.scala
gclaramunt: true, it can create a memory spike (specially if the jar keeps growing), and yeah, I found some scenarios where the hash is not consistent, I'm changing my approach.

modules/shared/src/main/scala/io/constellationnetwork/env/JarSignature.scala
```diff
@@ -0,0 +1,31 @@
...
+import cats.effect.IO
+
+import io.constellationnetwork.security.hash.Hash
+import io.constellationnetwork.security.hex.Hex
+
+import fs2.io.file.{Files, Path}
+
+object JarSignature {
+
+  val jarHash: IO[Hash] = {
```
marcinwadon, line 14: please use tagless algebra that we use in the whole codebase. define the least type required by the function and do not use the `IO` type here. `IO` is the type provided when you run the code, so in the `TessellationIOApp` or tests

marcinwadon, line 14: please use tagless algebra that we use in the whole codebase. define the least type required by the function and do not use the `IO` type here. `IO` is the type provided when you run the code, so in the `TessellationIOApp` or tests

modules/shared/src/main/scala/io/constellationnetwork/env/JarSignature.scala
```diff
@@ -0,0 +1,31 @@
...
+
+    IO(MessageDigest.getInstance("SHA-256")).flatMap { digest =>
+      Files[IO]
+        .readAll(Path(jarPath))
+        .chunkN(chunkSize) // Process file in chunks
+        .evalMap { chunk =>
+          IO(digest.update(chunk.toArray))
+        }
+        .compile
+        .drain >> IO(Hash(Hex.fromBytes(digest.digest()).value))
```
marcinwadon, line 27: please compose the smallest possible "programs" instead of doing everything at once. it should look like:
```scala
Async[F].delay(digest.digest())
   .map(_.value)
   .map(Hex.fromBytes)
   .map(Hash(_))
```

use `.delay` only for the part required to be delayed. it helps with thread shifting and increases performance of the application.

https://typelevel.org/cats-effect/docs/2.x/datatypes/io#keep-granularity

marcinwadon (dismissed):

modules/shared/src/main/scala/io/constellationnetwork/env/JarSignature.scala
gclaramunt: Hmm... the link talks about effectful functions. I agree that digest should be delayed because is an effect. This particular code executes only once, so I don't mind doing the changes you suggest, but I think it deserves a longer and more nuanced discussion for more performance critical code.

If you mean something like
```scala
Async[F].delay(digest.digest()).map {
  d => Hash(Hex.fromBytes(d).value)
}
```
I totally agree, doing everything in the `delay` probably has a slight performance advantage, but there's value in keeping the pure part separated.

modules/shared/src/main/scala/io/constellationnetwork/env/JarSignature.scala
```diff
@@ -0,0 +1,42 @@
...
+    digestOf(Files[F].readAll(Path(jarPath.getPath)))
+  }
+
+  def digestOf[F[_]: Sync: Files](bytes: fs2.Stream[F, Byte]): F[Hash] = {
+    val chunkSize = 524288
+
+    digestInstance >>= { digest =>
+      bytes
+        .chunkN(chunkSize)
+        .evalMap(updateDigest(digest, _))
```
marcinwadon, line 27: maybe `map2` or `mapN`?

modules/shared/src/main/scala/io/constellationnetwork/env/JarSignature.scala
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  private def digestInstance[F[_]: Sync] =
+    Sync[F].delay(MessageDigest.getInstance("SHA-256"))
+
+  private def updateDigest[F[_]: Sync](digest: MessageDigest, chunk: Chunk[Byte]) =
+    Sync[F].delay(digest.update(chunk.toArray))
+
+  private def toHash[F[_]: Sync](digest: MessageDigest) =
+    Sync[F].delay(digest.digest()).map(d => Hash(Hex.fromBytes(d).value))
```
marcinwadon, line 40: `.map(Hex.fromBytes).map(_value).map(Hash)`

modules/shared/src/test/scala/io/constellationnetwork/env/JarSignatureSuite.scala
```diff
@@ -0,0 +1,29 @@
...
+import io.constellationnetwork.security.hash.Hash
+
+import weaver._
+import weaver.scalacheck.Checkers
+
+object JarSignatureSuite extends SimpleIOSuite with Checkers {
+
+  test("digest of empty") {
+
+    JarSignature.digestOf[IO](fs2.Stream.empty).map { hash =>
```
marcinwadon, line 14: you can import `fs2.Stream`

modules/shared/src/main/scala/io/constellationnetwork/env/JarSignature.scala
```diff
@@ -0,0 +1,42 @@
...
+
+object JarSignature {
+
+  def jarHash[F[_]: Sync: Files]: F[Hash] = {
+    val jarPath = JarSignature.getClass.getProtectionDomain.getCodeSource.getLocation.toURI
+    digestOf(Files[F].readAll(Path(jarPath.getPath)))
+  }
+
+  def digestOf[F[_]: Sync: Files](bytes: fs2.Stream[F, Byte]): F[Hash] = {
+    val chunkSize = 524288
```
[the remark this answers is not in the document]
gclaramunt: I'm open to suggestions. Is how many bytes we take on each chunk. I've tuned the actual value by doing different runs in a relatively under-powered machine: is only 25% slower than the imperative code, with an upper bound of  512Kb memory usage, which I think is a good tradeoff.

modules/shared/src/main/scala/io/constellationnetwork/env/JarSignature.scala
gclaramunt: I like pointfree and in this case doesn't make a difference, so I'll change it.
That said,  Hash and Hex are newtypes, so Hash(_) and  hex.value are just casts (essentially no-ops ) and for performance critical code, I think we should avoid sending empty IO.Map objects to the main effect loop.

modules/shared/src/main/scala/io/constellationnetwork/env/JarSignature.scala
gclaramunt: sounds interesting, where do you suggest it?

modules/shared/src/main/scala/io/constellationnetwork/env/JarSignature.scala
```diff
@@ -0,0 +1,45 @@
...
+  def jarHash[F[_]: Async: Files]: F[Hash] = {
+    val jarPath = JarSignature.getClass.getProtectionDomain.getCodeSource.getLocation.toURI
+    digestOf(Files[F].readAll(Path(jarPath.getPath)))
+  }
+
+  def digestOf[F[_]: Async](bytes: Stream[F, Byte]): F[Hash] = {
+
+    val parCount = Math.max(2, Runtime.getRuntime.availableProcessors() - 1)
+
+    val streamChunkSizeBytes = 524288 / parCount
```
marcinwadon, line 25: for a single available processor, it will be a division by 0

kpudlik (approved):

PR #957, opened 2024-10-21
feat: keep binaries confirmed in global snapshot with confirmation proof

kpudlik:
Previously we were removing confirmed binaries from the tracking queue. Now we need to keep them together with confirmation proof (the reference to Global Snapshot in which binary was confirmed)

### DIFF
--- a/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/services/StateChannelBinarySender.scala
+++ b/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/services/StateChannelBinarySender.scala
@@ -12,24 +12,44 @@ import io.constellationnetwork.node.shared.domain.cluster.storage.L0ClusterStora
 import io.constellationnetwork.node.shared.domain.snapshot.storage.LastSnapshotStorage
 import io.constellationnetwork.node.shared.domain.statechannel.StateChannelValidator.StateChannelValidationError
 import io.constellationnetwork.node.shared.http.p2p.clients.StateChannelSnapshotClient
+import io.constellationnetwork.schema.epoch.EpochProgress
 import io.constellationnetwork.schema.{GlobalIncrementalSnapshot, GlobalSnapshotInfo, SnapshotOrdinal}
+import io.constellationnetwork.security.hash.Hash
 import io.constellationnetwork.security.{Hashed, Hasher}
 import io.constellationnetwork.statechannel.StateChannelSnapshotBinary
 
+import derevo.cats.eqv
+import derevo.derive
 import eu.timepit.refined.auto._
+import eu.timepit.refined.cats._
 import eu.timepit.refined.types.all.NonNegLong
 import eu.timepit.refined.types.numeric.PosLong
 import org.typelevel.log4cats.slf4j.Slf4jLogger
 import retry._
 
-case class TrackedBinary(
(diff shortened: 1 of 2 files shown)


modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/services/StateChannelBinarySender.scala
```diff
@@ -12,24 +12,48 @@ import io.constellationnetwork.node.shared.domain.cluster.storage.L0ClusterStora
...
+}
+
+case class PendingBinary(
   binary: Hashed[StateChannelSnapshotBinary],
   enqueuedAtOrdinal: SnapshotOrdinal,
   sendsSoFar: NonNegLong
-)
+) extends TrackedBinary
+
+case class ConfirmedBinary(
```
IPadawans, line 48: I'm not sure if it's possible, maybe we can have only a reference to the PendingBinary and then extract the fields there? Is it possible to avoid duplicating fields?

IPadawans (commented):
Just a small comment, I don't know if its possible, otherwise LGTM

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/services/StateChannelBinarySender.scala
kpudlik: Good idea. I will change that.

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/services/StateChannelBinarySender.scala
```diff
@@ -82,59 +103,75 @@ object StateChannelBinarySender {
...
 
               (cap, backoffExponent, noConfirmationsSinceRetryCount)
             } else if (updatedRetryMode) {
-              val confirmedCount = maybeConfirmationIndex.map(_ + 1L).getOrElse(0L)
+              val confirmedCount = maybeHighestConfirmationIndex.map(_ + 1L).getOrElse(0L)
 
               if (confirmedCount > 0) {
                 val cap = {
                   val maxCap = confirmedCount * confirmedCountMultiplier
                   val log2 = (x: Double) => Math.log10(x) / Math.log10(2.0)
```
scasplte2, line 172: Alternative log2 operation if you want to avoid the floating point operation, it is recursive though so maybe the floating point is more performant.

        @tailrec
        def log2(n: Int, acc: Int = 0): Int =
          if (n <= 1) acc
          else log2(n >> 1, acc + 1)

scasplte2 (dismissed):

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/services/StateChannelBinarySender.scala
kpudlik: In floating point based solution `log2(9)` is 3.169925001442312 so `Math.ceil` makes it 4. In your solution it will be 3. Since both functions produce different results I would leave it as it is for now to not change the maths behind the retry mode. I do not remember if `Math.ceil` was used on purpose so I would rather keep it as it is for now.

marcinwadon (approved):

PR #958, opened 2024-10-22
Add an export notice

effects-ai:
This PR add an export notice.


marcinwadon (changes requested):
I agree with @kpudlik. We do have `docs:` prefix and this PR looks like this.

marcinwadon (approved):
FYI: PR labels are not required right now, and actually unused.

kpudlik (approved):

PR #960, opened 2024-10-22
fix: remove guava

effects-ai:
This PR removes `guava` from the list of dependencies.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,1 +1,32 @@
 Copyright 2020-2023, Constellation Network Inc.
+
+This product includes software developed at
+Constellation Network Inc. (https://constellationnetwork.io/).
+
+The below notice derives from: https://github.com/apache/spark/blob/master/NOTICE
+and originates from The Apache Software Foundation (http://www.apache.org/).
+It has been modified to include descriptions of libraries providing crypto
+functionality within this project.
+
+Export Control Notice
+---------------------
+
+This distribution includes cryptographic software. The country in which you currently reside may have
+restrictions on the import, possession, use, and/or re-export to another country, of encryption software.
+BEFORE using any encryption software, please check your country's laws, regulations and policies concerning
+the import, possession, or use, and re-export of encryption software, to see if this is permitted. See
+<http://www.wassenaar.org/> for more information.
+
+The U.S. Government Department of Commerce, Bureau of Industry and Security (BIS), has classified this
+software as Export Commodity Control Number (ECCN) 5D002.C.1, which includes information security software
+using or performing cryptographic functions with asymmetric algorithms. The form and manner of this Apache
+Software Foundation distribution makes it eligible for export under the License Exception ENC Technology
+Software Unrestricted (TSU) exception (see the BIS Export Administration Regulations, Section 740.13) for
+both object code and source code.
+
+The following provides more details on the included cryptographic software:
+
+This software uses Bouncy Castle (https://www.bouncycastle.org/) for digital signatures and validation,
+certificates and certificate validation, key generation and management, and secure random number generation;
+and Java Security API (https://docs.oracle.com/en/java/javase/11/security/java-security-overview1.html)
+for message digests.
--- a/build.sbt
+++ b/build.sbt
@@ -20,7 +20,7 @@ val ghTokenSource = TokenSource.GitConfig("github.token") || TokenSource.Environ
 githubTokenSource := ghTokenSource
 
 lazy val commonSettings = Seq(
-  scalacOptions ++= List("-Ymacro-annotations", "-Yrangepos", "-Wconf:cat=unused:info", "-language:reflectiveCalls"),
+  scalacOptions ++= List("-Ymacro-annotations", "-Yrangepos", "-Wconf:cat=unused:info", "-language:reflectiveCalls", "-Wconf:msg=While parsing annotations in:silent"),
   scalafmtOnCompile := true,
   scalafixOnCompile := true,
   resolvers ++= List(
@@ -214,7 +214,6 @@ lazy val shared = (project in file("modules/shared"))
       Libraries.fs2Core,
       Libraries.fs2DataCsv,
       Libraries.fs2DataCsvGeneric,
-      Libraries.guava,
       Libraries.log4cats,
       Libraries.logback % Runtime,
       Libraries.logstashLogbackEncoder % Runtime,
--- a/modules/shared/src/main/scala/io/constellationnetwork/schema/address.scala
+++ b/modules/shared/src/main/scala/io/constellationnetwork/schema/address.scala
@@ -24,8 +24,8 @@ object address {
   object Address {
 
     def fromBytes(bytes: Array[Byte]): Address = {
-      val hashCode = Hash.hashCodeFromBytes(bytes)
-      val encoded = Base58.encode(hashCode.asBytes().toIndexedSeq)
+      val hashCode = Hash.sha256FromBytes(bytes)
+      val encoded = Base58.encode(hashCode.toIndexedSeq)
       val end = encoded.slice(encoded.length - 36, encoded.length)
       val validInt = end.filter(Character.isDigit)
       val ints = validInt.map(_.toString.toInt)
--- a/modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
+++ b/modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
@@ -1,16 +1,15 @@
 package io.constellationnetwork.security
 
 import java.nio.charset.StandardCharsets
+import java.security.MessageDigest
 
 import cats.Show
 
 import io.constellationnetwork.ext.derevo.ordering
 
-import com.google.common.hash.{HashCode, Hashing}
 import derevo.cats.{order, show}
 import derevo.circe.magnolia.{decoder, encoder}
 import derevo.derive
-import eu.timepit.refined.auto._
 import io.estatico.newtype.macros.newtype
 import io.estatico.newtype.ops._
 import org.scalacheck.{Arbitrary, Gen}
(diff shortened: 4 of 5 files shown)


NOTICE
```diff
@@ -1 +1,32 @@
...
+Software Foundation distribution makes it eligible for export under the License Exception ENC Technology
+Software Unrestricted (TSU) exception (see the BIS Export Administration Regulations, Section 740.13) for
+both object code and source code.
+
+The following provides more details on the included cryptographic software:
+
+This software uses Bouncy Castle (https://www.bouncycastle.org/) for digital signatures and validation,
+certificates and certificate validation, key generation and management, and secure random number generation;
+and Java Security API (https://docs.oracle.com/en/java/javase/11/security/java-security-overview1.html)
+for message digests.
```
kpudlik, line 32: I think it should be a part of a separate commit/PR. Since the title is "fix: remove guava" it should only remove `guava` dependency.

I would put it to separate commit/PR called `docs: add NOTICE` or sth.

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
```diff
@@ -25,11 +24,19 @@ object hash {
...
+      md.digest()
+    }
+
+    def fromBytes(bytes: Array[Byte]): Hash = {
+      val sha256Bytes = sha256FromBytes(bytes)
+      val sha256String = sha256Bytes
+        .foldLeft(new StringBuilder(64))((acc, x) => acc.append(s"%02x".format(x)))
+        .toString
+      Hash(sha256String)
+    }
```
kpudlik, line 39: Have you verified if new methods produce same hashes as `Hashing` one?

kpudlik (changes requested):

build.sbt
```diff
@@ -20,7 +20,7 @@ val ghTokenSource = TokenSource.GitConfig("github.token") || TokenSource.Environ
 githubTokenSource := ghTokenSource
 
 lazy val commonSettings = Seq(
-  scalacOptions ++= List("-Ymacro-annotations", "-Yrangepos", "-Wconf:cat=unused:info", "-language:reflectiveCalls"),
+  scalacOptions ++= List("-Ymacro-annotations", "-Yrangepos", "-Wconf:cat=unused:info", "-language:reflectiveCalls", "-Wconf:msg=While parsing annotations in:silent"),
```
marcinwadon, line 23: Why do we need this? Does removing `guava` enforce adding this `Wconf` parameter?

modules/shared/src/main/scala/io/constellationnetwork/schema/address.scala
```diff
@@ -24,8 +24,8 @@ object address {
   object Address {
 
     def fromBytes(bytes: Array[Byte]): Address = {
-      val hashCode = Hash.hashCodeFromBytes(bytes)
-      val encoded = Base58.encode(hashCode.asBytes().toIndexedSeq)
+      val hashCode = Hash.sha256FromBytes(bytes)
```
marcinwadon, line 27: Is it still a `hashCode`? You've changed the function name but not a variable

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
marcinwadon: I would be super super careful doing this. Hashing is the base of the whole codebase. We should test it again starting from genesis, revalidating the chain (partially at least) to make sure we won't make a breaking change and the hashing remains consistent.

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
```diff
@@ -25,11 +24,19 @@ object hash {
...
-    def hashCodeFromBytes(bytes: Array[Byte]): HashCode =
-      Hashing.sha256().hashBytes(bytes)
-
-    def fromBytes(bytes: Array[Byte]): Hash =
-      Hash(hashCodeFromBytes(bytes).toString)
+    def sha256FromBytes(bytes: Array[Byte]): Array[Byte] = {
+      val md = MessageDigest.getInstance("SHA-256")
+      md.update(bytes)
+      md.digest()
+    }
```
marcinwadon, line 31: I think you should just change the implementation of the `hashCodeFromBytes`, not the fn name. We don't know who/how depends on that.

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
```diff
@@ -25,11 +24,19 @@ object hash {
...
   object Hash {
 
-    def hashCodeFromBytes(bytes: Array[Byte]): HashCode =
-      Hashing.sha256().hashBytes(bytes)
-
-    def fromBytes(bytes: Array[Byte]): Hash =
-      Hash(hashCodeFromBytes(bytes).toString)
+    def sha256FromBytes(bytes: Array[Byte]): Array[Byte] = {
+      val md = MessageDigest.getInstance("SHA-256")
+      md.update(bytes)
```
marcinwadon, line 29: Is it performant to create a new MessageDigest for each run? afaik google uses clone before each use, to use a single instance but keep it immutable/"clean" for external use.

can you run some benchmark to compare the old and new code? as I wrote in a separate comment, hashing is the most base level of the codebase and node runs it very very often.

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
effects-ai: See https://github.com/effects-ai/tessellation/blob/sha256_tests/modules/shared/src/test/scala/io/constellationnetwork/security/GuavaJSATest.scala

modules/shared/src/main/scala/io/constellationnetwork/schema/address.scala
effects-ai: It's not an instance of `HashCode`, it's a byte array that contains a SHA-256 digest.

build.sbt
effects-ai: It's needed to avoid this error:
```
[error] While parsing annotations in /Users/effects-ai/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/io/micrometer/micrometer-commons/1.10.2/micrometer-commons-1.10.2.jar(io/micrometer/common/lang/Nullable.class), could not find MAYBE in enum <none>.
[error] This is likely due to an implementation restriction: an annotation argument cannot refer to a member of the annotated class (scala/bug#7014).
[error] No warnings can be incurred under -Werror.
[error] two errors found
```

modules/shared/src/main/scala/io/constellationnetwork/schema/address.scala
marcinwadon: Ok, so you should change the `hashCode` variable name too

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
marcinwadon: Ok, so what do we gain removing guava?

build.sbt
marcinwadon: Ok, have you seen this error before removing guava? If so, it should be a separate change, not together with removing guava. Otherwise, it would be impossible to revert if needed. Also, it would mean it's your specific setup improperly configured, as everything is working correctly for everybody else over a long time.

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
marcinwadon: What's the output?

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
marcinwadon: I agree, but you need to remember about compatibility too. We do have external dependencies on our code, so we cannot refactor everything as we like to.

build.sbt
marcinwadon: btw:
```
[error] two errors found
```

What are those?

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
effects-ai: All green, since all tests use the same java.security.MessageDigest inside, plus

```
MessageDigest.clone() 32389 ms
MessageDigest.getInstance() 33180 ms
Guava 32336 ms
```

YMMV, though.

build.sbt
effects-ai: > ```
> [error] While parsing annotations in /Users/effects-ai/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/io/micrometer/micrometer-commons/1.10.2/micrometer-commons-1.10.2.jar(io/micrometer/common/lang/Nullable.class), could not find MAYBE in enum <none>.
> [error] This is likely due to an implementation restriction: an annotation argument cannot refer to a member of the annotated class (scala/bug#7014).
> [error] No warnings can be incurred under -Werror.
> ```

effects-ai: > Ok, have you seen this error before removing guava? If so, it should be a separate change, not together with removing guava. Otherwise, it would be impossible to revert if needed. Also, it would mean it's your specific setup improperly configured, as everything is working correctly for everybody else over a long time.

I'll try to setup a completely different setup to verify.

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
effects-ai: Such changes can be added to the release notes.

build.sbt
effects-ai: It reproducible on a clean install of Ubuntu 22.0 as well.

This is another reason to drop guava - it hides issues with other dependencies. This issue affects all versions of `micrometer` up to 1.12, micrometer 1.13 seems to be fixed, but it requires code changes in `Metrics.scala`.

AlexBrandes (changes requested):
Let's discuss at the meeting tomorrow but on the surface this change appears much too risky with a very small potential benefit. It would need extensive testing by the team to prove that it's valid and doesn't cause issues downstream. I don't think that it makes sense at this early stage in onboarding to the team.

effects-ai:
Closing this PR in favor of https://github.com/Constellation-Labs/tessellation/pull/965

PR #962, opened 2024-10-25
feat: getLastSynchronizedGlobalSnapshot support for L0NodeContext

kpudlik:
Based on the "fixed-window" solution the following algorithm is implemented:

1. Confirm pending sent binaries (just as it was working before).
2. Take the highest confirmed binary index and subtract the fixed window size from it to get the highest Currency Snapshot (binary) that is still within the window. Remove all confirmed binaries that are older, as these are out of the fixed window bounds.
3. Take the global snapshot in which the fixed window's Currency Snapshot (binary) was confirmed and provide it to the Data Application. Remove all stored global snapshots that are older, as these will no longer be returned.

It needs to be tested on the cluster and may have some edge cases, so I would treat it as a proof of concept (PoC).

### DIFF
--- a/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/CurrencyL0App.scala
+++ b/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/CurrencyL0App.scala
@@ -77,8 +77,6 @@ abstract class CurrencyL0App(
       p2pClient = P2PClient.make[IO](sharedP2PClient, sharedResources.client, sharedServices.session)
       maybeAllowanceList = StateChannelAllowanceLists.get(cfg.environment)
       validators = Validators.make[IO](cfg.shared, seedlist, maybeAllowanceList, Hasher.forKryo[IO])
-      implicit0(nodeContext: L0NodeContext[IO]) = L0NodeContext
-        .make[IO](storages.snapshot, hasherSelectorAlwaysCurrent, storages.identifier)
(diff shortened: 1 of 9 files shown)


IPadawans (approved):
LGTM

marcinwadon (dismissed):
Looks good but I'm not able to verify the logic if it's correct. @ryle-goehausen please review the PR before merging

marcinwadon:
@ryle-goehausen @AlexBrandes this implementation looks like non deterministic, because it depends on the local view of the latest global snapshot. it will be different for each node, so each node would choose a different cutoff ordinal.

kpudlik:
> @ryle-goehausen @AlexBrandes this implementation looks like non deterministic, because it depends on the local view of the latest global snapshot. it will be different for each node, so each node would choose a different cutoff ordinal.

Yep. If N is substracted from the latest CIS then it will get desynchronized if latest CIS is also desynchronized. Of course there are cases when multiple CIS are in single GIS but very likely fixed-window will cause forks anyway. That's why I think we need to switch to consensus-based solution.

According to diagram:
Fixed Window (C+3)- 5 = (C-2)  -> Find(GL0 = G-2)
But some nodes are C+3 but some may be already C+4 which may result in C-3 that is in GLO = G-1.

kpudlik:
I am closing the PR since we changed the solution to consensus-based one. I will branch out from that one to reimplement the solution.

PR #964, opened 2024-10-30
feat: adding additional features that will be used by amm metagraph

marcus-girolneto:
### Changes
+ Additional parts that should be used by amm metagraph, such as:
-> getMetagraphId: We will use this to get the AllowSpends aggregated in GlobalSnapshotInfo to the current metagraph
-> allowSpend - currency: Changed to optional since we need it as none to represent DAG
-> ordering: adding ordering to the missing parts that we need in amm metagraph

### Note
+ This PR references the branch that adds lastGlobalSnapshotSync, and not directly the allow-spend branch

### DIFF
--- a/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/CurrencyL0App.scala
+++ b/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/CurrencyL0App.scala
@@ -102,7 +102,13 @@ abstract class CurrencyL0App(
         )
         .asResource
       implicit0(nodeContext: L0NodeContext[IO]) = L0NodeContext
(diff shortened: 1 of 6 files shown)


marcinwadon:
@IPadawans Krystian's change most likely won't be merged, as we decided on Slack to implement consensus-based solution.

ryle-goehausen (approved):

IPadawans:
@marcinwadon but even using the consensus-based, the changes I've proposed should be included. As soon as we have something in this new PR I'll update this one

marcinwadon:
Yes, I agree. But these changes should not be based on Krystian's change because his branch won't be merged. You should rebase your code against the feature branch itself. Otherwise, you'd merge changes to the dead branch and you'll have to redo your work later anyway.

PR #965, opened 2024-10-30
refactor: deprecate Guava

effects-ai:
This PR deprecates the Guava version of `Hash.fromBytes()` method. 

It doesn't remove Guava from the dependency list, and provides compatibility tests.

### DIFF
--- a/modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
+++ b/modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
@@ -1,8 +1,9 @@
 package io.constellationnetwork.security
 
 import java.nio.charset.StandardCharsets
+import java.security.MessageDigest
 
-import cats.Show
+import cats.{Eq, Show}
 
 import io.constellationnetwork.ext.derevo.ordering
 
@@ -10,7 +11,6 @@ import com.google.common.hash.{HashCode, Hashing}
 import derevo.cats.{order, show}
 import derevo.circe.magnolia.{decoder, encoder}
 import derevo.derive
-import eu.timepit.refined.auto._
 import io.estatico.newtype.macros.newtype
 import io.estatico.newtype.ops._
 import org.scalacheck.{Arbitrary, Gen}
@@ -24,12 +24,39 @@ object hash {
   }
 
   object Hash {
+    private val hexDigits = "0123456789abcdef".toCharArray
+    private val sha256 = MessageDigest.getInstance("SHA-256")
 
+    case class Sha256Digest(private val bytes: Array[Byte]) {
+      def toByteArray: Array[Byte] = Array.copyOf(bytes, bytes.length)
+
+      def toHexString: String =
+        bytes
+          .foldLeft(new StringBuilder(bytes.length * 2)) { (sb, b) =>
+            sb.append(hexDigits((b >> 4) & 0xf)).append(hexDigits(b & 0xf))
+          }
+          .toString
+
+      def toIndexedSeq: IndexedSeq[Byte] = bytes.toIndexedSeq
+    }
+
+    object Sha256Digest {
+      implicit val show: Show[Sha256Digest] = Show.show(s => s"Sha256Digest(${s.toHexString})")
+      implicit val eq: Eq[Sha256Digest] = Eq.instance((a, b) => a.bytes.sameElements(b.bytes))
+    }
+
+    @deprecated("Use sha256DigestFromBytes() instead", since = "3.0")
     def hashCodeFromBytes(bytes: Array[Byte]): HashCode =
       Hashing.sha256().hashBytes(bytes)
 
+    def sha256DigestFromBytes(bytes: Array[Byte]): Sha256Digest = {
+      val md = sha256.clone().asInstanceOf[MessageDigest]
+      md.update(bytes)
+      Sha256Digest(md.digest())
+    }
+
     def fromBytes(bytes: Array[Byte]): Hash =
-      Hash(hashCodeFromBytes(bytes).toString)
+      Hash(sha256DigestFromBytes(bytes).toHexString)
 
     def empty: Hash = Hash(s"%064d".format(0))
 
--- a/modules/shared/src/test/scala/io/constellationnetwork/security/HashSuite.scala
+++ b/modules/shared/src/test/scala/io/constellationnetwork/security/HashSuite.scala
@@ -1,7 +1,10 @@
 package io.constellationnetwork.security
 
+import cats.Show
 import cats.effect.{IO, Resource}
 import cats.syntax.all._
+
+import scala.annotation.nowarn
 
 import io.constellationnetwork.ext.cats.effect.ResourceIO
 import io.constellationnetwork.json.JsonSerializer
@@ -9,12 +12,14 @@ import io.constellationnetwork.kryo.KryoSerializer
 import io.constellationnetwork.schema.epoch.EpochProgress
 import io.constellationnetwork.schema.{GlobalSnapshot, SnapshotOrdinal}
 import io.constellationnetwork.security.hash.Hash
+import io.constellationnetwork.security.hex.Hex
 import io.constellationnetwork.shared.{SharedKryoRegistrationId, sharedKryoRegistrar}
 
 import derevo.circe.magnolia.encoder
 import derevo.derive
 import eu.timepit.refined.auto._
 import io.circe.Encoder
+import org.scalacheck.{Arbitrary, Gen}
 import weaver.MutableIOSuite
 import weaver.scalacheck.Checkers
 
@@ -56,6 +61,20 @@ object HashSuite extends MutableIOSuite with Checkers {
     val testUpdated = TestUpdated(2, None)
 
     res.withCurrent(implicit hasher => (hasher.hash(test), hasher.hash(testUpdated)).mapN(expect.eql(_, _)))
+  }
+
+  test("ensure guava/JSA compatibility") {
+    implicit val byteArrayShow: Show[Array[Byte]] = Show.show(a => Hex.fromBytes(a).toString)
+
+    val byte = Arbitrary.arbitrary[Byte]
+    val bytes = Gen.listOfN(1024, byte).map(_.toArray)
+
+    forall(bytes) { data =>
+      @nowarn
+      val hashCode = Hash.hashCodeFromBytes(data)
+      val sha256Digest = Hash.sha256DigestFromBytes(data)
+      expect.eql(hashCode.toString, sha256Digest.toHexString)
+    }
   }
 
   @derive(encoder)
--- a/modules/shared/src/main/scala/io/constellationnetwork/schema/address.scala
+++ b/modules/shared/src/main/scala/io/constellationnetwork/schema/address.scala
@@ -24,8 +24,8 @@ object address {
   object Address {
 
     def fromBytes(bytes: Array[Byte]): Address = {
-      val hashCode = Hash.hashCodeFromBytes(bytes)
-      val encoded = Base58.encode(hashCode.asBytes().toIndexedSeq)
+      val sha256Digest = Hash.sha256DigestFromBytes(bytes)
+      val encoded = Base58.encode(sha256Digest.toIndexedSeq)
       val end = encoded.slice(encoded.length - 36, encoded.length)
       val validInt = end.filter(Character.isDigit)
       val ints = validInt.map(_.toString.toInt)


modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
```diff
@@ -24,12 +25,38 @@ object hash {
...
+      def toIndexedSeq: IndexedSeq[Byte] = bytes.toIndexedSeq
+
+      def toHexString: String =
+        bytes
+          .foldLeft(new StringBuilder(bytes.length * 2)) { (sb, b) =>
+            sb.append(hexDigits((b >> 4) & 0xf)).append(hexDigits(b & 0xf))
+          }
+          .toString
+
+      override def toString: String = s"Sha256Digest($toHexString)"
```
marcinwadon, line 45: Please implement a Show type class instead of this toString

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
```diff
@@ -24,12 +25,38 @@ object hash {
...
   object Hash {
+    private val hexDigits = "0123456789abcdef".toCharArray
+    private val sha256 = MessageDigest.getInstance("SHA-256")
 
+    case class Sha256Digest(private val bytes: Array[Byte]) {
+      require(bytes.length == 32, "invalid SHA-256 digest length")
+
+      def toByteArray: Array[Byte] = Array.copyOf(bytes, bytes.length)
+
+      def toIndexedSeq: IndexedSeq[Byte] = bytes.toIndexedSeq
```
marcinwadon, line 36: Do we need to expose that publicly?

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
```diff
@@ -24,12 +25,38 @@ object hash {
...
+        bytes
+          .foldLeft(new StringBuilder(bytes.length * 2)) { (sb, b) =>
+            sb.append(hexDigits((b >> 4) & 0xf)).append(hexDigits(b & 0xf))
+          }
+          .toString
+
+      override def toString: String = s"Sha256Digest($toHexString)"
+    }
+
+    @deprecated("Use sha256DigestFromBytes() instead", since = "2.12.1-20")
```
marcinwadon, line 48: This is a testnet version that we don't release publicly for integrations. Instead, we should either keep there a mainnet (so the version we would actually release) or bump it just before the release.

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
```diff
@@ -24,12 +25,38 @@ object hash {
...
     def hashCodeFromBytes(bytes: Array[Byte]): HashCode =
       Hashing.sha256().hashBytes(bytes)
 
+    def sha256DigestFromBytes(bytes: Array[Byte]): Sha256Digest = {
+      val md = sha256.clone().asInstanceOf[MessageDigest]
```
marcinwadon, line 53: Is asInstanceOf required?

modules/shared/src/test/scala/io/constellationnetwork/security/HashSuite.scala
```diff
@@ -58,6 +60,21 @@ object HashSuite extends MutableIOSuite with Checkers {
...
+    val (guavaHashes, jsaHashes) = 0.to(1_000_000).foldLeft((Vector.empty[String], Vector.empty[String])) { (acc, _) =>
+      val (guava, jsa) = acc
+      scala.util.Random.nextBytes(data)
+      @nowarn
+      val hashCode = Hash.hashCodeFromBytes(data)
+      val sha256Digest = Hash.sha256DigestFromBytes(data)
+      (guava :+ hashCode.toString, jsa :+ sha256Digest.toHexString)
+    }
+
+    IO(expect(guavaHashes == jsaHashes))
```
marcinwadon, line 75: If you don't use the IO, don't use the effectful test. Just use pureTest if the test is meant to be pure

modules/shared/src/test/scala/io/constellationnetwork/security/HashSuite.scala
```diff
@@ -58,6 +60,21 @@ object HashSuite extends MutableIOSuite with Checkers {
     res.withCurrent(implicit hasher => (hasher.hash(test), hasher.hash(testUpdated)).mapN(expect.eql(_, _)))
   }
 
+  test("ensure guava/JSA compatibility") {
+    val data = Array.ofDim[Byte](1024)
```
marcinwadon, line 64: Can you try to use a generator for this?

modules/shared/src/test/scala/io/constellationnetwork/security/HashSuite.scala
```diff
@@ -58,6 +60,21 @@ object HashSuite extends MutableIOSuite with Checkers {
...
+
+    val (guavaHashes, jsaHashes) = 0.to(1_000_000).foldLeft((Vector.empty[String], Vector.empty[String])) { (acc, _) =>
+      val (guava, jsa) = acc
+      scala.util.Random.nextBytes(data)
```
marcinwadon, line 68: You should use Random from cats

modules/shared/src/test/scala/io/constellationnetwork/security/HashSuite.scala
marcinwadon, line 75: Expect should use Eq (===) from cats, instead of the universal equality. Also, there is also expect same or expect equal functions that would work in your case

marcinwadon (changes requested):

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
effects-ai: Yes

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
effects-ai: Done

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
effects-ai: Done

modules/shared/src/test/scala/io/constellationnetwork/security/HashSuite.scala
effects-ai: Not applicable anymore

effects-ai: Done

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
marcinwadon: there is still the overriding `toString`

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
```diff
@@ -24,12 +24,42 @@ object hash {
   }
 
   object Hash {
+    private val hexDigits = "0123456789abcdef".toCharArray
+    private val sha256 = MessageDigest.getInstance("SHA-256")
 
+    case class Sha256Digest(private val bytes: Array[Byte]) {
+      require(bytes.length == 32, "invalid SHA-256 digest length")
+
```
marcinwadon, line 32: instead of throwing errors, you should make the case class constructor private and create an apply function for `Sha256Digest` that will explicitly return an Either. in FP, you want to avoid side effects.  another way would be to use refined types if possible

marcinwadon (changes requested):

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
effects-ai: Done

modules/shared/src/test/scala/io/constellationnetwork/security/HashSuite.scala
```diff
@@ -58,6 +63,20 @@ object HashSuite extends MutableIOSuite with Checkers {
...
   }
 
+  test("ensure guava/JSA compatibility") {
+    implicit val byteArrayShow: Show[Array[Byte]] = Show.show(a => Hex.fromBytes(a).toString)
+
+    val byte = Arbitrary.arbitrary[Byte]
+    val bytes = Gen.listOfN(1024, byte).map(_.toArray)
+
+    forall(bytes) { data =>
+      @nowarn
```
marcinwadon, line 73: what was the warn?

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
```diff
@@ -24,12 +24,53 @@ object hash {
...
+    }
+
+    @deprecated("Use sha256DigestFromBytes() instead", since = "3.0")
     def hashCodeFromBytes(bytes: Array[Byte]): HashCode =
       Hashing.sha256().hashBytes(bytes)
 
+    def sha256DigestFromBytes(bytes: Array[Byte]): Sha256Digest = {
+      val md = sha256.clone().asInstanceOf[MessageDigest]
+      md.update(bytes)
+      Sha256Digest.Sha256DigestImpl(md.digest())
```
marcinwadon, line 69: hmm, you have created safe a apply function, but you don't use it at all 🤔

marcinwadon (changes requested):

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
effects-ai: I don't need it here. I know that `MessageDigest.digest()` will always return exactly 32 bytes, and if it ever doesn't, that would indicate a catastrophic failure. In that case, the number of bytes returned by `MessageDigest.digest()` would be the least of my concerns.

modules/shared/src/test/scala/io/constellationnetwork/security/HashSuite.scala
effects-ai: Deprecation obviously.

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
```diff
@@ -24,12 +24,53 @@ object hash {
   }
 
   object Hash {
+    private val hexDigits = "0123456789abcdef".toCharArray
+    private val sha256 = MessageDigest.getInstance("SHA-256")
 
+    trait Sha256Digest {
```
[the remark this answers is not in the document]
effects-ai: This is how people create a private constructor for a case class with a public apply method: https://stackoverflow.com/questions/20030826/scala-case-class-private-constructor-but-public-apply-method

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
```diff
@@ -24,12 +24,53 @@ object hash {
...
+
+      def apply(bytes: Array[Byte]): Either[String, Sha256Digest] =
+        if (bytes.length == 32) {
+          Right(Sha256DigestImpl(bytes))
+        } else {
+          Left("invalid SHA-256 digest length")
+        }
+
+      private[constellationnetwork] case class Sha256DigestImpl(private val bytes: Array[Byte]) extends Sha256Digest {
+        override def toByteArray: Array[Byte] = Array.copyOf(bytes, bytes.length)
```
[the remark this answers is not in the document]
effects-ai: It does override a trait method.

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
```diff
@@ -24,12 +24,53 @@ object hash {
...
+
+        override def toHexString: String =
```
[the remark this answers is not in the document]
effects-ai: same

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
```diff
@@ -24,12 +24,53 @@ object hash {
...
+          bytes
+            .foldLeft(new StringBuilder(bytes.length * 2)) { (sb, b) =>
+              sb.append(hexDigits((b >> 4) & 0xf)).append(hexDigits(b & 0xf))
+            }
+            .toString
+
+        override private[constellationnetwork] def toIndexedSeq: IndexedSeq[Byte] = bytes.toIndexedSeq
```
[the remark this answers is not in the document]
effects-ai: same

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
```diff
@@ -24,12 +24,53 @@ object hash {
...
+      implicit val show: Show[Sha256Digest] = Show.show(s => s"Sha256Digest(${s.toHexString})")
+
+      def apply(bytes: Array[Byte]): Either[String, Sha256Digest] =
+        if (bytes.length == 32) {
+          Right(Sha256DigestImpl(bytes))
+        } else {
+          Left("invalid SHA-256 digest length")
+        }
+
+      private[constellationnetwork] case class Sha256DigestImpl(private val bytes: Array[Byte]) extends Sha256Digest {
```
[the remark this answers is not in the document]
effects-ai: Not applicable anymore

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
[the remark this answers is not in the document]
effects-ai: Not applicable anymore

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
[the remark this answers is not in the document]
effects-ai: Not applicable anymore

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
```diff
@@ -24,12 +24,53 @@ object hash {
...
   object Hash {
+    private val hexDigits = "0123456789abcdef".toCharArray
+    private val sha256 = MessageDigest.getInstance("SHA-256")
 
+    trait Sha256Digest {
+      def toByteArray: Array[Byte]
+
+      def toHexString: String
+
+      private[constellationnetwork] def toIndexedSeq: IndexedSeq[Byte]
```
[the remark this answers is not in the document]
effects-ai: See https://github.com/Constellation-Labs/tessellation/pull/965#discussion_r1828190453

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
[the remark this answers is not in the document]
effects-ai: Not applicable anymore

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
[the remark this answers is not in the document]
effects-ai: Not applicable anymore

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
```diff
@@ -24,12 +24,53 @@ object hash {
...
+    }
+
+    object Sha256Digest {
+      implicit val show: Show[Sha256Digest] = Show.show(s => s"Sha256Digest(${s.toHexString})")
+
+      def apply(bytes: Array[Byte]): Either[String, Sha256Digest] =
+        if (bytes.length == 32) {
+          Right(Sha256DigestImpl(bytes))
+        } else {
+          Left("invalid SHA-256 digest length")
```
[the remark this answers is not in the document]
effects-ai: Not applicable anymore

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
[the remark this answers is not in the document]
marcinwadon: It implements the trait method but it does not override it

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
```diff
@@ -24,12 +24,39 @@ object hash {
...
+      def toByteArray: Array[Byte] = Array.copyOf(bytes, bytes.length)
+
+      def toHexString: String =
+        bytes
+          .foldLeft(new StringBuilder(bytes.length * 2)) { (sb, b) =>
+            sb.append(hexDigits((b >> 4) & 0xf)).append(hexDigits(b & 0xf))
+          }
+          .toString
+
+      private[constellationnetwork] def toIndexedSeq: IndexedSeq[Byte] = bytes.toIndexedSeq
```
marcinwadon, line 40: Again - why it's private if you open it for the whole package?

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
```diff
@@ -24,12 +24,39 @@ object hash {
...
+    }
+
+    object Sha256Digest {
+      implicit val show: Show[Sha256Digest] = Show.show(s => s"Sha256Digest(${s.toHexString})")
+      implicit val eq: Eq[Sha256Digest] = Eq.instance((a, b) => a.bytes.sameElements(b.bytes))
```
marcinwadon, line 45: Instead of specifying custom Eq, you can derive existing one

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
```diff
@@ -24,12 +24,39 @@ object hash {
   }
 
   object Hash {
+    private val hexDigits = "0123456789abcdef".toCharArray
+    private val sha256 = MessageDigest.getInstance("SHA-256")
 
+    case class Sha256Digest(private val bytes: Array[Byte]) {
+      def toByteArray: Array[Byte] = Array.copyOf(bytes, bytes.length)
```
marcinwadon, line 31: Why not just "bytes"?

effects-ai: Because `Array[Byte]` is mutable.

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
effects-ai: Using `@derive`? It doesn't compile. 

```
magnolia: could not find Eq.Typeclass for type Array[Byte]
```

But my implementation both compiles and works.

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
effects-ai: I made it public again. First, it doesn't cause any issues, and second, it helps save memory for the client if they choose to convert hashes into `Vector`s of bytes.

effects-ai:
> any update on hash calculation?

This PR doesn't modify the hash calculation, it merely deprecates the original method of invoking the calculation.

modules/shared/src/main/scala/io/constellationnetwork/security/Hash.scala
marcinwadon: It does not compile because you need to provide an implicit instance of an `Eq[Array[Byte]]` which is:
```scala
implicit val byteArrayEq: Eq[Array[Byte]] = Eq.instance(Arrays.equals(_, _))
```

and `Eq` instance for `Sha256Digest` you can simply derive. Many implementations would compile and work :)

marcinwadon: so you need an `Eq` instance for just `Array[Byte]` instead of an `Sha256Digest`. Many implementations would compile and work :)

marcinwadon (approved):

PR #966, opened 2024-11-04
feat: adding features required by amm metagraph

IPadawans:
### Changes
+ Additional parts that should be used by amm metagraph, such as:
-> getMetagraphId: We will use this to get the AllowSpends aggregated in GlobalSnapshotInfo to the current metagraph
-> allowSpend - currency: Changed to optional since we need it as none to represent DAG
-> ordering: adding ordering to the missing parts that we need in amm metagraph

### DIFF
--- a/modules/shared/src/main/scala/io/constellationnetwork/currency/dataApplication/package.scala
+++ b/modules/shared/src/main/scala/io/constellationnetwork/currency/dataApplication/package.scala
@@ -15,6 +15,7 @@ import io.constellationnetwork.currency.schema.feeTransaction.FeeTransaction
 import io.constellationnetwork.routes.internal.ExternalUrlPrefix
 import io.constellationnetwork.schema.artifact.SharedArtifact
 import io.constellationnetwork.schema.round.RoundId
+import io.constellationnetwork.schema.swap.CurrencyId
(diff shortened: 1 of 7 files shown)


modules/shared/src/main/scala/io/constellationnetwork/currency/dataApplication/package.scala
```diff
@@ -543,4 +544,5 @@ trait L0NodeContext[F[_]] {
   def getCurrencySnapshot(ordinal: SnapshotOrdinal): F[Option[Hashed[CurrencyIncrementalSnapshot]]]
   def getLastCurrencySnapshotCombined: F[Option[(Hashed[CurrencyIncrementalSnapshot], CurrencySnapshotInfo)]]
   def securityProvider: SecurityProvider[F]
+  def getCurrencyId: F[Address]
```
scasplte2, line 547: should this return an `Address` or a `CurrencyId`? It seems confusingly named if it is for the former

IPadawans: Ooh you're right, I forgot we have the currencyId type now, let me update

scasplte2 (approved):

PR #968, opened 2024-11-12
fix: deduplicate snapshot rumors

effects-ai:


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/daemon/SnapshotEventsPublisherDaemon.scala
```diff
@@ -20,11 +21,16 @@ trait SnapshotEventsPublisherDaemon[F[_]] {
...
+    consensusEvents: Stream[F, E],
+    consensusStorage: ConsensusStorage[F, E, _, _, _, _, _, _]
   ): SnapshotEventsPublisherDaemon[F] =
     new SnapshotEventsPublisherDaemon[F] {
       def spawn: Daemon[F] = Daemon.spawn {
-        consensusEvents
+        consensusEvents.map { e =>
+          println(e)
+          e
+        }
```
IPadawans, line 32: It think you don't need to println here, even if you need probably you need to change to logger

IPadawans (changes requested):
I left one comment that probably you forgot to remove. Another topic, remember to run locally:
`sbt 'scalafixAll --check --rules OrganizeImports;scalafmtCheckAll'`
to check if your code is according to the patterns, it seems that you have a failed github task due to this.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/daemon/SnapshotEventsPublisherDaemon.scala
marcinwadon: i believe it's a leftover and should be removed. for tests, you should match and expect instead of visually check what's printed in console :)

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/metrics/Metrics.scala
```diff
@@ -234,4 +234,61 @@ object Metrics {
       }
   }
 
+  def noopIO: Metrics[IO] =
```
marcinwadon, line 237: if it's something used by tests only, it should go to the test runtime, not to the compile

marcinwadon (changes requested):
@effects-ai please squash/ammend the commit as you want to push a single change into the main branch. each PR is meant to be a single change (hence single commit).
commit named "address PR comments" is not a meaningful change and says nothing about the changes inside.
also, the fix is to your code that is not merged yet, so you fix the fix basically. hence, it should be ammended or squashed into your original merge request (as you fix your original code to be merged)

modules/currency-l0/src/test/scala/io/constellationnetwork/currency/l0/snapshot/SnapshotEventsPublisherDaemonSuite.scala
```diff
@@ -0,0 +1,122 @@
...
+import eu.timepit.refined.types.numeric.{NonNegLong, PosInt, PosLong}
+import fs2.Stream
+import io.circe.{Encoder, Json}
+import weaver.MutableIOSuite
+import weaver.scalacheck.Checkers
+
+object SnapshotEventsPublisherDaemonSuite extends MutableIOSuite with Checkers {
+
+  override type Res =
+    (Supervisor[IO], Hasher[IO], SecurityProvider[IO], Metrics[IO], Encoder[CurrencySnapshotEvent], KeyPair)
```
marcinwadon, line 37: does `Encoder` for `CurrencySnapshotEvent` need to be a resource? it's a simple value, not a resource

modules/currency-l0/src/test/scala/io/constellationnetwork/currency/l0/snapshot/SnapshotEventsPublisherDaemonSuite.scala
```diff
@@ -0,0 +1,122 @@
...
+      _ <- SnapshotEventsPublisherDaemon.make(gossip, events, consensusStorage).spawn.start
+      _ <- waitUntilQueueIsEmpty(l1OutputQueue)
+      rumorQueueSize <- rumorQueue.size
+      maybeRumor <- rumorQueue.tryTake
+    } yield {
+      val maybeEitherBlockEvent = maybeRumor.map { rumor =>
+        val cur = rumor.signed.value.content.hcursor
+        cur.downField("value").get[BlockEvent]("BlockEvent")
+      }
+      assert.eql(1, rumorQueueSize).and(expect(maybeEitherBlockEvent.contains(Right(blockEvent2))))
```
marcinwadon, line 119: `expect.eql`

modules/currency-l0/src/test/scala/io/constellationnetwork/currency/l0/snapshot/SnapshotEventsPublisherDaemonSuite.scala
```diff
@@ -0,0 +1,122 @@
...
+      gossip <- Gossip.make[IO](rumorQueue, selfId, Generation(PosLong(1)), keyPair)
+      _ <- SnapshotEventsPublisherDaemon.make(gossip, events, consensusStorage).spawn.start
+      _ <- waitUntilQueueIsEmpty(l1OutputQueue)
+      rumorQueueSize <- rumorQueue.size
+      maybeRumor <- rumorQueue.tryTake
+    } yield {
+      val maybeEitherBlockEvent = maybeRumor.map { rumor =>
+        val cur = rumor.signed.value.content.hcursor
+        cur.downField("value").get[BlockEvent]("BlockEvent")
+      }
```
marcinwadon, line 118: this can be inside the for comp. let's keep the `yield` as simple as possible

marcinwadon: also, I think you don't need to use a plain json to obtain a block event

PR #969, opened 2024-11-12
fix: fix a typo in a directory name

effects-ai:


IPadawans (changes requested):
Remember to run locally:
`sbt 'scalafixAll --check --rules OrganizeImports;scalafmtCheckAll'`
to check if your code is according to the patterns, it seems that you have a failed github task due to this.

marcinwadon:
> Remember to run locally: `sbt 'scalafixAll --check --rules OrganizeImports;scalafmtCheckAll'` to check if your code is according to the patterns, it seems that you have a failed github task due to this.

I don't think so. Looks like one action failed randomly because of the GitHub and the second action failed because of non deterministic test. I'd just retry them.

IPadawans (approved):

PR #972, opened 2024-11-12
feat: removed unused code from swap schema

IPadawans:


modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala
```diff
@@ -103,13 +103,6 @@ object swap {
     implicit object OrderingInstance extends OrderBasedOrdering[AllowSpend]
   }
 
-  @derive(encoder)
-  case class AllowSpendView(
```
marcinwadon, line 107: It is not unused (or won't be). It's supposed to be returned in the same way we return transaction view

marcinwadon (changes requested):

modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala
IPadawans: You're right, I've forgotten to return this code

marcinwadon (approved):

PR #975, opened 2024-11-20
feat: global snapshot sync consensus on L0

kpudlik:


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotCreator.scala
```diff
@@ -179,25 +194,31 @@ object CurrencySnapshotCreator {
...
             feeTransactions = none,
-            acceptedSharedArtifacts.some
+            currencySnapshotAcceptanceResult.sharedArtifacts.some
           )
 
-          context = CurrencySnapshotContext(lastContext.address, snapshotInfo)
+          context = CurrencySnapshotContext(lastContext.address, currencySnapshotAcceptanceResult.info)
           newAwaitingEvents = awaitingBlocks.map(BlockEvent(_)).toSet[CurrencySnapshotEvent] ++ awaitedEvents
-          newRejectedEvents = rejectedBlocks.map(BlockEvent(_)).toSet[CurrencySnapshotEvent] ++ messagesAcceptanceResult.notAccepted
+          // TODO: Add sync events that were rejected
```
marcinwadon, line 215: leftover?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotSyncValidator.scala
```diff
@@ -0,0 +1,64 @@
...
+        implicit hasher: Hasher[F]
+      ): F[GlobalSnapshotSyncOrError] = {
+
+        val seedlistPeers = seedlist.map(_.map(_.peerId))
+
+        for {
+          hasOnlyOneSignature <- validator.validateMaxSignatureCount(globalSnapshotSync, 1).pure[F]
+          isSignedCorrectly <- validator.validateSignatures(globalSnapshotSync)
+          isSignedBySeedlistPeer = validator.validateSignaturesWithSeedlist(seedlistPeers, globalSnapshotSync)
+          isSignedByFacilitator = validateIfSignedByFacilitator(globalSnapshotSync, facilitators)
```
marcinwadon, line 43: I'm not sure that's correct 🤔 I think that the joining peer also should be validated correctly 🤔

modules/shared/src/main/scala/io/constellationnetwork/currency/schema/currency.scala
```diff
@@ -162,6 +167,7 @@ object currency {
     epochProgress: EpochProgress,
     dataApplication: Option[DataApplicationPart] = None,
     messages: Option[SortedSet[Signed[CurrencyMessage]]] = None,
+    globalSnapshotSyncs: Option[SortedSet[Signed[GlobalSnapshotSync]]] = None,
```
marcinwadon, line 170: I think that the default value should be an empty set of events 🤔 this is what you'd set as a default value with no `Option`. an `Option` here is just to ensure backward compatibility

marcinwadon (changes requested):

modules/shared/src/main/scala/io/constellationnetwork/currency/schema/currency.scala
kpudlik: Don't we need backward compatibility there anyway? 🤔

marcinwadon: I mean, yes. The type should still remain the same. but the default value should be an empty sorted set rather than None. Consider it like:

```scala
sealed trait Wrapper[A]
case class ImplementedField[A](a: A) extends Wrapper[A]
case class NotImplementedField extends Wrapper[Nothing]
```

then, your type is:
```scala
globalSnapshotSyncs: Wrapper[SortedSet[Signed[GlobalSnapshotSync]] = ImplementedField(SortedSet.empty)
```

where for the previous global snapshots (backward compatibility) it would be:
```
.toGlobalSnapshotInfo(..., NotImplementedField)
```

Instead of this custom solution, we do use `Option` that is isomorphic to this

kpudlik: Ah, you are right!

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotSyncValidator.scala
kpudlik: The joining peer list is local to the node so different nodes will give then different results of validation. It will cause forking.

marcinwadon (dismissed):

marcinwadon (approved):

PR #976, opened 2024-11-21
feat: add cluster session to PeerInfo

gclaramunt:
Returns clusterSession on each node reported in "cluster/info". Required by https://github.com/Constellation-Labs/cllb/pull/18

### DIFF
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/cluster/services/Cluster.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/cluster/services/Cluster.scala
@@ -92,6 +92,7 @@ object Cluster {
             req.ip,
             req.publicPort,
             req.p2pPort,
+            req.clusterSession.value.toString,
             req.session.value.toString,
             req.state,
             req.jar
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/cluster/programs/Joining.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/cluster/programs/Joining.scala
@@ -197,6 +197,7 @@ class Joining[
         registrationRequest.ip,
         registrationRequest.publicPort,
         registrationRequest.p2pPort,
+        registrationRequest.clusterSession,
         registrationRequest.session,
         registrationRequest.state,
         Responsive,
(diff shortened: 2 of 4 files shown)


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/cluster/services/Cluster.scala
```diff
@@ -85,16 +85,18 @@ object Cluster {
...
             req.ip,
             req.publicPort,
             req.p2pPort,
             req.session.value.toString,
+            req.clusterSession.value.toString,
             req.state
           )
 
-          clusterStorage.getResponsivePeers.map(_.map(PeerInfo.fromPeer) + self)
+          clusterStorage.getResponsivePeers.map(_.map(PeerInfo.fromPeer(req.clusterSession)) + self)
```
marcinwadon, line 99: You put node (self) cluster session as some other peer's cluster session. Instead, you should add clusterSession to the `Peer` case class.

marcinwadon (changes requested):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/cluster/services/Cluster.scala
gclaramunt: the initial idea was use just the self session, but your suggestion makes sense

marcinwadon: IMO it does not make sense if you'd use the self session to present as some other node's session

marcinwadon (approved):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/cluster/services/Cluster.scala
gclaramunt: Shouldn't be all in the same session? Anyway, that code is outdated, and it takes the session from whatever is in the Peer object

marcinwadon: Well, they "should" :) The only check is made during Joining process. If we would change that in the future, your code would support it now, instead of blind assumptions :)

gclaramunt: yeah, the code now reports the cluster session id that the peer had at the time of joining.

PR #985, opened 2024-12-10
fix: skip jar hash if running directly from classes

gclaramunt:
When running inside intelliJ, there's no jar to hash and the calculation fails. This PR gives a default value if no jar and running from dev environment, if there's no jar and the env is not dev, it fails.

### DIFF
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/app/TessellationIOApp.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/app/TessellationIOApp.scala
@@ -6,8 +6,9 @@ import cats.effect._
 import cats.effect.std.{Random, Supervisor}
 import cats.syntax.all._
 
-import io.constellationnetwork.env.JarSignature
+import io.constellationnetwork.env.AppEnvironment.Dev
 import io.constellationnetwork.env.env._
+import io.constellationnetwork.env.{AppEnvironment, JarSignature}
 import io.constellationnetwork.ext.cats.effect._
 import io.constellationnetwork.ext.kryo._
 import io.constellationnetwork.json.JsonSerializer
@@ -32,6 +33,7 @@ import io.constellationnetwork.schema.generation.Generation
 import io.constellationnetwork.schema.peer.PeerId
 import io.constellationnetwork.schema.semver.TessellationVersion
 import io.constellationnetwork.security._
+import io.constellationnetwork.security.hash.Hash
 
 import com.monovore.decline.Opts
 import com.monovore.decline.effect.CommandIOApp
@@ -111,7 +113,7 @@ abstract class TessellationIOApp[A <: CliMethod](
                 logger.info(s"App environment: ${cfg.environment}") >>
                 logger.info(s"App version: ${version.show}") >>
                 logger.info(s"App collateral: ${cfg.collateral.amount.show}") >>
-                JarSignature.jarHash[IO].flatMap { jarHash =>
+                jarHash(cfg.environment).flatMap { jarHash =>
                   logger.info(s"Jar hash: ${jarHash.value}") >>
                     KryoSerializer.forAsync[IO](registrar).use { implicit _kryoPool =>
                       JsonSerializer.forSync[IO].asResource.use { implicit _jsonSerializer =>
@@ -252,6 +254,15 @@ abstract class TessellationIOApp[A <: CliMethod](
         }
       }
     }
+
+  private def jarHash(currentEnv: AppEnvironment): IO[Hash] =
+    JarSignature
+      .jarHash[IO]
+      .onError(logger.error(_)(s"Error calculating jar's hash."))
+      .recoverWith {
+        case _ if currentEnv === Dev =>
+          logger.warn(s"Using mock value for dev environment.") >> Hash.empty.pure[IO]
+      }
 
   private def loadKeyPair[F[_]: Async: SecurityProvider](
     keyStore: StorePath,
--- a/modules/shared/src/main/scala/io/constellationnetwork/env/JarSignature.scala
+++ b/modules/shared/src/main/scala/io/constellationnetwork/env/JarSignature.scala
@@ -3,7 +3,7 @@ package io.constellationnetwork.env
 import java.security.MessageDigest
 
 import cats.effect.Async
-import cats.implicits.{catsSyntaxFlatMapOps, toFunctorOps}
+import cats.syntax.all._
 
 import io.constellationnetwork.security.hash.Hash
 import io.constellationnetwork.security.hex.Hex
@@ -14,8 +14,8 @@ import fs2.{Chunk, Stream}
 object JarSignature {
 
   def jarHash[F[_]: Async: Files]: F[Hash] = {
-    val jarPath = JarSignature.getClass.getProtectionDomain.getCodeSource.getLocation.toURI
-    digestOf(Files[F].readAll(Path(jarPath.getPath)))
+    val jarPath = Path(JarSignature.getClass.getProtectionDomain.getCodeSource.getLocation.toURI.getPath)
+    digestOf(Files[F].readAll(jarPath))
   }
 
   def digestOf[F[_]: Async](bytes: Stream[F, Byte]): F[Hash] = {


modules/shared/src/main/scala/io/constellationnetwork/env/JarSignature.scala
```diff
@@ -14,8 +14,11 @@ import fs2.{Chunk, Stream}
 object JarSignature {
 
   def jarHash[F[_]: Async: Files]: F[Hash] = {
-    val jarPath = JarSignature.getClass.getProtectionDomain.getCodeSource.getLocation.toURI
-    digestOf(Files[F].readAll(Path(jarPath.getPath)))
+    val jarPath = Path(JarSignature.getClass.getProtectionDomain.getCodeSource.getLocation.toURI.getPath)
+    if (jarPath.extName === ".jar")
+      digestOf(Files[F].readAll(jarPath))
+    else
+      Hash("No jar found").pure
```
marcinwadon, line 21: Hmm, hash shouldn't be constructed like that. Either there should be a fallback to all zeros or sth like that, or imo preferable way would be to return option/either from this function, if we explicitly know it may fail

marcinwadon (changes requested):

modules/shared/src/main/scala/io/constellationnetwork/env/JarSignature.scala
gclaramunt: I didn't want to complicate the code for a pathological case that only happens when running locally, but I'll change it to Option.

IPadawans (dismissed):
LGTM

marcinwadon (approved):
@gclaramunt I'd strongly suggest returning an `F[Option[Hash]]` from this function, instead of implicitly returning an empty Hash

marcinwadon:
as `Hash.empty` gives us no value. if we cannot calculate the hash, simply return `None`

gclaramunt:
I thought about that, but since we need it for the registration request, making it optional will defeat the purpose of using it to make sure all nodes are using the same build. That forces us to either stop the service or provide a default value (either at `jarHash` method or `TessellationIOApp.main` method). I think the most correct behavior is to fail with an exception (channeled through IO) albeit a better one that "java.io.IOException: Is a directory" ... maybe "not running from Jar". But that makes difficult to run it directly from the classes in a local dev environment, so giving a 0 value seems a reasonable tradeoff. 

(maybe it can be solved with tweaking with intelliJ config too  :P )

marcinwadon:
> I thought about that, but since we need it for the registration request, making it optional will defeat the purpose of using it to make sure all nodes are using the same build

Hmm, why? It will be required to have `None` for each node, as the hash cannot be determined. It would work exactly the same as `Hash.empty` but we would explicitly specify, that the hash cannot be calculated, instead of putting a dummy hash there. I strongly suggest that option, because we do have a ticket in backlog to make hash validation on type level, and that may break this feature. but having an optional field in the registration request to explicitly specify we do have a hash or we cannot calculate it seems to be both simple and meaningful.

gclaramunt:
> Hmm, why? It will be required to have None for each node, as the hash cannot be determined.
My problem with hash = None is that is an invalid state of the system and I'm reluctant to add it to the external API.  
I agree that returning the empty hash might not be the best, so I've remove it and left to caller (TessellationIOApp) to handle the error. Let me know what you think, if you still don't like it, I'll change to Option.

marcinwadon:
Well, it still looks the same, as EOD you use an empty hash but the truth is that you cannot obtain the hash. An Option is not an invalid state of the system 🤔

gclaramunt:
I've added code to calculate directly from classfiles is there's no jar.  The calculation itself never fails, any failure is an IO failure and already signaled through IO, no need to add an extra option/either wrapper, (in the same way we don't wrap things like config in an option even if it can fail to read the file).

This isn't supposed to be a big refactoring, just a small change trying to make @kpudlik life a little better :)

modules/shared/src/main/scala/io/constellationnetwork/env/JarSignature.scala
```diff
@@ -31,6 +38,9 @@ object JarSignature {
     }
   }
 
+  def readAllBytesFromDir[F[_]: Files](dir: Path): Stream[F, Byte] =
+    Files[F].walk(dir).evalFilter(Files[F].isRegularFile).filter(p => p.endsWith("semanticdb")).flatMap(Files[F].readAll)
```
marcinwadon, line 42: you want to calculate based on semanticdb files only, or you wanted to exclude it?

gclaramunt: :facepalm: that always trip me :P

modules/shared/src/main/scala/io/constellationnetwork/env/JarSignature.scala
```diff
@@ -14,8 +14,15 @@ import fs2.{Chunk, Stream}
...
   def jarHash[F[_]: Async: Files]: F[Hash] = {
-    val jarPath = JarSignature.getClass.getProtectionDomain.getCodeSource.getLocation.toURI
-    digestOf(Files[F].readAll(Path(jarPath.getPath)))
+    val jarPath = Path(JarSignature.getClass.getProtectionDomain.getCodeSource.getLocation.toURI.getPath)
+
+    val bytes =
+      if (jarPath.extName === ".jar")
+        Files[F].readAll(jarPath)
+      else
+        readAllBytesFromDir(jarPath)
```
marcinwadon, line 23: Hmm, as I think about it, this will be non deterministic, as you may have a different content of a directory containing jar file. Hence, it's not a good source for hash calculation, that should be deterministic across builds. Also checking the file type by just its extension does not look good too and may be a source of issues. I'd rather stay with the previous version where you assumed the hash can be calculated. If that fails, return an error.

gclaramunt: Yes, fail if there's an IO problem is the right approach. Now, what to do in the rare situation you're running directly from class files? I don't think we need to support it for production nodes, but I want to make it a little easier if you need to debug from IDE.
There are a few workarounds: locally change the code to return something and revert it, which is cumbersome, or run the code externally from the jar and attach the IDE as remote debug. Those are not particularly appealing.
The other option in case of failure, is to return some arbitrary value: 0, or a hash of the classes, it won't match the eventual hash of the jar but since is a local run and not a production use, it won't matter much.

Now, if we want to run without the jar AND control the hash value (maybe to join another cluster) we can read the value from a system property in the failure recovery.

marcinwadon: > I don't think we need to support it for production nodes, but I want to make it a little easier if you need to debug from IDE.

Well, so either you make sure that the code works on dev environment only and production stays the same, or you implement it in a way that works for production nodes too.

> The other option in case of failure, is to return some arbitrary value: 0, or a hash of the classes, it won't match the eventual hash of the jar but since is a local run and not a production use, it won't matter much.

This fix actually will be on production nodes so it may happen on production nodes. So again, either you need to make sure it works on Dev (and maybe Testnet) environments only (check for example localhost usage for Dev environment when joining nodes together. it's forbidden in the production code but it works locally), or you need to make sure it won't break production (so returning 0s isn't a great option there).

gclaramunt: Checking for env is what I actually need, thanks!

modules/shared/src/main/scala/io/constellationnetwork/env/JarSignature.scala
```diff
@@ -31,6 +31,9 @@ object JarSignature {
     }
   }
 
+  def readAllBytesFromDir[F[_]: Files](dir: Path): Stream[F, Byte] =
```
marcinwadon, line 34: So this fn is not needed now, correct?

gclaramunt: removed

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/app/TessellationIOApp.scala
```diff
@@ -253,6 +255,15 @@ abstract class TessellationIOApp[A <: CliMethod](
       }
     }
 
+  private def jarHash(currentEnv: AppEnvironment): IO[Hash] =
+    JarSignature
+      .jarHash[IO]
+      .onError(e => logger.error(s"Error calculating jar's hash: ${e.getMessage}."))
```
marcinwadon, line 261: It's better to pass error explicitly, as it records the stack trace too:
`logger.error(e)("Error calculating jar's hash")`

marcinwadon (dismissed):
Approved, but please update the logging message according to the comment :+1:

PR #987, opened 2024-12-11
feat: add cl0 transaction tracking

effects-ai:


modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/ValidationErrorRoutes.scala
```diff
@@ -0,0 +1,47 @@
...
+  currencySnapshotValidationErrorStorage: ValidationErrorStorage[F, CurrencySnapshotEvent, Option[BlockRejectionReason]]
+) extends Http4sDsl[F]
+    with PublicRoutes[F] {
+
+  implicit val encoder: Encoder[(CurrencySnapshotEvent, Option[BlockRejectionReason], Instant)] =
+    new Encoder[(CurrencySnapshotEvent, Option[BlockRejectionReason], Instant)] {
+      override def apply(a: (CurrencySnapshotEvent, Option[BlockRejectionReason], Instant)): Json = Json.obj(
+        (
+          "block",
+          a._1 match {
```
marcinwadon, line 31: Please use explicit destructuring, as we do in the rest of the codebase. Using `._1` and `._2` obfuscate the code

modules/dag-l0/src/test/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessorSuite.scala
```diff
@@ -75,8 +78,15 @@ object GlobalSnapshotStateChannelEventsProcessorSuite extends MutableIOSuite {
         validators.currencyMessageValidator
       )
       currencyEventsCutter = CurrencyEventsCutter.make[IO](None)
+      validationErrorStorage = MockValidationErrorStorage.make[CurrencySnapshotEvent, Option[BlockRejectionReason]]()
```
marcinwadon, line 81: why we need a mock?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  def make[F[_]: Async, Artifact, ValidationError](
+    maxSize: Int
+  )(implicit clock: Clock[F]): F[ValidationErrorStorage[F, Artifact, ValidationError]] =
+    for {
+      updateSemaphore <- Semaphore[F](1)
+      validationErrorsR <- Ref[F].of(Vector.empty[(Artifact, ValidationError, Instant)])
+    } yield
+      new ValidationErrorStorage[F, Artifact, ValidationError] {
+        override def addValidationError(artifact: Artifact, errors: ValidationError): F[Unit] =
```
marcinwadon, line 27: `override` modifier is wrongly used. it overrides nothing as this code implements the trait, not overrides it or any parent class.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
```diff
@@ -0,0 +1,54 @@
...
+}
+
+object ValidationErrorStorage {
+
+  def make[F[_]: Async, Artifact, ValidationError](
+    maxSize: Int
+  )(implicit clock: Clock[F]): F[ValidationErrorStorage[F, Artifact, ValidationError]] =
+    for {
+      updateSemaphore <- Semaphore[F](1)
+      validationErrorsR <- Ref[F].of(Vector.empty[(Artifact, ValidationError, Instant)])
```
marcinwadon, line 24: looks like this tuple type could/should be a case class

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
```diff
@@ -0,0 +1,54 @@
...
+  def addValidationError(artifact: Artifact, errors: ValidationError): F[Unit]
+  def addValidationErrors(errors: NonEmptyList[(Artifact, ValidationError)]): F[Unit]
+  def allValidationErrors: F[Vector[(Artifact, ValidationError, Instant)]]
+}
+
+object ValidationErrorStorage {
+
+  def make[F[_]: Async, Artifact, ValidationError](
+    maxSize: Int
+  )(implicit clock: Clock[F]): F[ValidationErrorStorage[F, Artifact, ValidationError]] =
```
marcinwadon, line 21: why clock is passed here like this? it should be above along all implicit types for F

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
```diff
@@ -0,0 +1,54 @@
...
+          }
+
+        override def allValidationErrors: F[Vector[(Artifact, ValidationError, Instant)]] = validationErrorsR.get
+
+        override def addValidationErrors(errors: NonEmptyList[(Artifact, ValidationError)]): F[Unit] =
+          updateSemaphore.permit.use { _ =>
+            errors.foldLeft(addError(errors.head._1, errors.head._2)) { (acc, x) =>
+              for {
+                _ <- acc
+                res <- addError(x._1, x._2)
```
marcinwadon, line 39: as above, please don't use `._1` pattern as it obfuscates the code

marcinwadon, line 39: btw. you don't need a for comprehension for this. 

instead of
```scala
for {
  _ <- acc
  res <- addError(x._1, x._2)
} yield res
```

you could write:
```scala
acc >> addError(x._1, x._2)
```

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/modules/SharedServices.scala
```diff
@@ -84,12 +87,17 @@ object SharedServices {
 
       currencyEventsCutter = CurrencyEventsCutter.make[F](None)
 
+      currencySnapshotValidationErrorStorage <- ValidationErrorStorage.make[F, CurrencySnapshotEvent, Option[BlockRejectionReason]](
+        10000
```
marcinwadon, line 91: magic number, please use config

modules/node-shared/src/test/scala/io/constellationnetwork/node/shared/infrastructure/consensus/MockValidationErrorStorage.scala
```diff
@@ -0,0 +1,19 @@
...
+  def make[Artifact, ValidationError](): ValidationErrorStorage[IO, Artifact, ValidationError] =
+    new ValidationErrorStorage[IO, Artifact, ValidationError] {
+      override def addValidationError(artifact: Artifact, errors: ValidationError): IO[Unit] = IO.unit
+
+      override def addValidationErrors(errors: NonEmptyList[(Artifact, ValidationError)]): IO[Unit] = IO.unit
+
+      override def allValidationErrors: IO[Vector[(Artifact, ValidationError, Instant)]] =
+        IO.pure(Vector.empty)
+    }
+}
```
marcinwadon, line 19: instead of a mock, you could use a normal storage for tests 🤔

modules/node-shared/src/test/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorageSuite.scala
```diff
@@ -0,0 +1,33 @@
+package io.constellationnetwork.node.shared.infrastructure.consensus
+
+import cats.data.NonEmptyList
+import cats.effect.IO
+import cats.kernel.Eq
+
+import weaver.SimpleIOSuite
+
+object ValidationErrorStorageSuite extends SimpleIOSuite {
+  implicit val eq: Eq[IndexedSeq[(Int, String)]] = Eq.instance((a, b) => a == b)
```
marcinwadon, line 10: how it differs from `expect.same`?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
```diff
@@ -0,0 +1,54 @@
...
+  def allValidationErrors: F[Vector[(Artifact, ValidationError, Instant)]]
+}
+
+object ValidationErrorStorage {
+
+  def make[F[_]: Async, Artifact, ValidationError](
+    maxSize: Int
+  )(implicit clock: Clock[F]): F[ValidationErrorStorage[F, Artifact, ValidationError]] =
+    for {
+      updateSemaphore <- Semaphore[F](1)
```
marcinwadon, line 23: Why do you need a semaphore, what was your motivation to use it together with `Ref`? `Ref` is thread-safe and supports concurrent read/write (we use it a lot in the codebase). Is there anything special that it needs to be additionally locked?

marcinwadon (changes requested):

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/ValidationErrorRoutes.scala
```diff
@@ -0,0 +1,47 @@
...
+          }
+        ),
+        ("timestamp", a._3.asJson)
+      )
+    }
+
+  override protected val prefixPath: InternalUrlPrefix = "/currency"
+
+  override protected val public: HttpRoutes[F] = HttpRoutes.of[F] {
+    case GET -> Root / "validation-errors" =>
```
kpudlik, line 44: I am wondering if it shouldn't be a sub-resource of /snapshot/events/validation-errors but if it is currency-framework specific then maybe currency/snapshot-event-validation-errors or something more specific 🤔

WDYT?

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/modules/HttpApi.scala
```diff
@@ -43,7 +45,8 @@ object HttpApi {
     nodeVersion: TessellationVersion,
     httpCfg: HttpConfig,
     maybeDataApplication: Option[BaseDataApplicationL0Service[F]],
-    maybeMetagraphVersion: Option[MetagraphVersion]
+    maybeMetagraphVersion: Option[MetagraphVersion],
+    currencySnapshotValidationErrorStorage: ValidationErrorStorage[F, CurrencySnapshotEvent, Option[BlockRejectionReason]]
```
kpudlik, line 49: Technically it doesn't store snapshot validation errors. These are errors of snapshot events so inputs for creating the currency snapshot.

kpudlik (changes requested):

build.sbt
```diff
@@ -356,7 +356,7 @@ lazy val rosetta = (project in file("modules/rosetta"))
 lazy val dagL1 = (project in file("modules/dag-l1"))
   .enablePlugins(AshScriptPlugin)
   .enablePlugins(JavaAppPackaging)
-  .dependsOn(kernel, shared % "compile->compile;test->test", nodeShared, testShared % Test)
+  .dependsOn(kernel, shared % "compile->compile;test->test", nodeShared % "compile->compile;test->test", testShared % Test)
```
marcinwadon, line 359: do we still need it? we already have a `testShared` module so there should be no point to share `test` phase of `nodeShared` 🤔

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/ValidationErrorRoutes.scala
```diff
@@ -0,0 +1,45 @@
...
+import org.http4s.dsl.Http4sDsl
+
+final case class ValidationErrorRoutes[F[_]: Async](
+  currencySnapshotValidationErrorStorage: ValidationErrorStorage[F, CurrencySnapshotEvent, Option[BlockRejectionReason]]
+) extends Http4sDsl[F]
+    with PublicRoutes[F] {
+
+  implicit val encoder: Encoder[ValidationErrorStorageEntry[CurrencySnapshotEvent, Option[BlockRejectionReason]]] =
+    new Encoder[ValidationErrorStorageEntry[CurrencySnapshotEvent, Option[BlockRejectionReason]]] {
+      override def apply(a: ValidationErrorStorageEntry[CurrencySnapshotEvent, Option[BlockRejectionReason]]): Json = Json.obj(
```
marcinwadon, line 26: image

`Encoder` does not implement a default implementation of `apply`, so you don't override it, but just implement it. `override` modifier is wrongly used here

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/ValidationErrorRoutes.scala
```diff
@@ -0,0 +1,45 @@
...
+        (
+          "block",
+          a.artifact match {
+            case currency.BlockEvent(value)            => value.asJson
+            case currency.DataApplicationBlockEvent(_) => Json.Null
+            case currency.CurrencyMessageEvent(_)      => Json.Null
```
marcinwadon, line 32: why other events are ignored?

modules/dag-l0/src/test/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotTraverseSuite.scala
```diff
@@ -246,13 +248,20 @@ object GlobalSnapshotTraverseSuite extends MutableIOSuite with Checkers {
...
     implicit val hs = HasherSelector.forSyncAlwaysCurrent(H)
 
-    val currencySnapshotCreator =
-      CurrencySnapshotCreator
-        .make[IO](currencySnapshotAcceptanceManager, None, SnapshotSizeConfig(Long.MaxValue, Long.MaxValue), currencyEventsCutter)
-    val currencySnapshotValidator = CurrencySnapshotValidator.make[IO](currencySnapshotCreator, validators.signedValidator, None, None)
-
-    val currencySnapshotContextFns = CurrencySnapshotContextFunctions.make(currencySnapshotValidator)
     for {
+      validationErrorStorage <- ValidationErrorStorage.make[IO, CurrencySnapshotEvent, Option[BlockRejectionReason]](16)
```
marcinwadon, line 252: what is `16`? looks like magic number 🤔

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
```diff
@@ -0,0 +1,50 @@
...
+import cats.syntax.functor._
+
+import eu.timepit.refined.auto.autoUnwrap
+import eu.timepit.refined.types.all.PosInt
+
+trait ValidationErrorStorage[F[_], Artifact, ValidationError] {
+  def addValidationError(artifact: Artifact, errors: ValidationError): F[Unit]
+  def addValidationErrors(errors: NonEmptyList[(Artifact, ValidationError)]): F[Unit]
+  def allValidationErrors: F[Vector[ValidationErrorStorageEntry[Artifact, ValidationError]]]
+}
```
marcinwadon, line 17: instead of repeating the same name over and over, it may be:
```scala
trait ValidationErrorStorage ... {
  def add(artifact: Artifact, error: ValidationError): ...
  def add(artifact: Artifact, errors: NonEmptyList[ValidationError]): ...
  def get: F[Vector[...]]
}
```

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
```diff
@@ -0,0 +1,50 @@
...
+import cats.data.NonEmptyList
+import cats.effect.kernel.{Async, Clock, Ref}
+import cats.syntax.flatMap._
+import cats.syntax.functor._
+
+import eu.timepit.refined.auto.autoUnwrap
+import eu.timepit.refined.types.all.PosInt
+
+trait ValidationErrorStorage[F[_], Artifact, ValidationError] {
+  def addValidationError(artifact: Artifact, errors: ValidationError): F[Unit]
```
marcinwadon, line 14: is `ValidationError` a single error or multiple errors? Type says one thing but property name the other

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
```diff
@@ -0,0 +1,50 @@
...
+  def addValidationErrors(errors: NonEmptyList[(Artifact, ValidationError)]): F[Unit]
```
marcinwadon, line 15: do you add many errors for the same artifact? maybe it would be better not to repeat artifact in the tuple but take it out as a separate parameter

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
```diff
@@ -0,0 +1,50 @@
...
+case class ValidationErrorStorageEntry[Artifact, ValidationError](artifact: Artifact, validationError: ValidationError, timestamp: Instant)
+
+object ValidationErrorStorage {
+
+  def make[F[_]: Async, Artifact, ValidationError](
+    maxSize: PosInt
+  ): F[ValidationErrorStorage[F, Artifact, ValidationError]] =
+    for {
+      validationErrorsR <- Ref[F].of(Vector.empty[ValidationErrorStorageEntry[Artifact, ValidationError]])
+    } yield
```
marcinwadon, line 28: you don't need a for comprehension for such a simple statement.

```scala
Ref[F].of(Vector.empty[ValidationErrorStorageEntry[Artifact, ValidationError]]).map { validationErrorsR =>
  new ValidationErrorStorage[F, Artifact, ValidationError] {
    ...
  }
}
```

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
```diff
@@ -0,0 +1,50 @@
...
+      new ValidationErrorStorage[F, Artifact, ValidationError] {
+        override def addValidationError(artifact: Artifact, error: ValidationError): F[Unit] =
```
marcinwadon, line 30: again, you don't override a default implementation but just providing the first implementation, so it `override`s nothing

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
```diff
@@ -0,0 +1,50 @@
...
+          for {
+            timestamp <- Clock[F].realTimeInstant
```
marcinwadon, line 32: do you ever match errors for equality? how the timestamp is gonna by used? those timestamps are local timestamps and won't match any other timestamp in the network

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
```diff
@@ -0,0 +1,50 @@
...
+                ValidationErrorStorageEntry(artifact, error, timestamp) +: bounded
+              }
+          } yield res
+
+        override def addValidationErrors(errors: NonEmptyList[(Artifact, ValidationError)]): F[Unit] = {
+          val (headArtifact, headError) = errors.head
+          errors.foldLeft(addValidationError(headArtifact, headError)) {
+            case (acc, (artifact, error)) =>
+              acc >> addValidationError(artifact, error)
+          }
```
marcinwadon, line 45: instead of extracting the head and violating DRY rule, you can put an empty unit in the fold as a zero value and fold over all elements then. you can also use `traverse` probably, that would be easier than reducing values like that, since `F` is a monad and you can chain computations

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotCreator.scala
```diff
@@ -154,6 +157,16 @@ object CurrencySnapshotCreator {
...
 
+          rejectedBlockEvents = NonEmptyList.fromListUnsafe(acceptanceResult.notAccepted.collect {
+            case (b, r: BlockRejectionReason) => (BlockEvent(b), Some(r))
+          })
+          _ <- currencySnapshotValidationErrorStorage.addValidationErrors(rejectedBlockEvents)
+
+          rejectedMessageEvents = NonEmptyList.fromListUnsafe(
+            messagesAcceptanceResult.notAccepted.map(m => (CurrencyMessageEvent(m), None))
+          )
+          _ <- currencySnapshotValidationErrorStorage.addValidationErrors(rejectedMessageEvents)
```
marcinwadon, line 168: you ignore messages in the encoder, so why you add them here? is the Encoder invalid then?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/modules/SharedServices.scala
```diff
@@ -84,12 +87,17 @@ object SharedServices {
 
       currencyEventsCutter = CurrencyEventsCutter.make[F](None)
 
+      currencySnapshotValidationErrorStorage <- ValidationErrorStorage.make[F, CurrencySnapshotEvent, Option[BlockRejectionReason]](
+        cfg.validationErrorStorage.size
```
marcinwadon, line 91: does it compile? you've never provided this configuration variable 🤔

marcinwadon (changes requested):

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/ValidationErrorRoutes.scala
effects-ai: Done

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
effects-ai: Done

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
effects-ai: Done

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
effects-ai: Done

effects-ai: Done

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/modules/SharedServices.scala
effects-ai: Done

modules/node-shared/src/test/scala/io/constellationnetwork/node/shared/infrastructure/consensus/MockValidationErrorStorage.scala
effects-ai: Done

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/ValidationErrorRoutes.scala
effects-ai: They are place holders, will be removed with actual error codes.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
effects-ai: No, I add one error per one artifact.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
effects-ai: `ValidationError` can be whatever the client chooses.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/modules/SharedServices.scala
effects-ai: I does, the configuration variable does not affect the compilation process.

modules/dag-l0/src/test/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotTraverseSuite.scala
marcinwadon: This is great, I like numbers too. However, in the codebase, we put magic numbers in the configuration, so they are properly named and easily to find out and change.
https://medium.com/@codechuckle/code-smells-the-impact-of-magic-numbers-37a90b04cfbe

marcinwadon (changes requested):
Please respond to all comments, so we can review the code again.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
effects-ai: The timestamps will be used for time-based eviction. For the first iteration only size-base eviction is implemented.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
effects-ai: Not applicable anymore.

build.sbt
effects-ai: No we don't.

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/modules/HttpApi.scala
effects-ai: Does `currencySnapshotEventValidationErrorStorage` sound better?

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/ValidationErrorRoutes.scala
effects-ai: You're right. The API serves errors for all kinds of events, not just currency ones. I changed it to `/snapshots/events/errors`, and added `/snapshots/transactions/:txhash/errors`.

modules/dag-l0/src/test/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessorSuite.scala
effects-ai: I removed the mock.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
effects-ai: Done

effects-ai: In fact `override` is used perfectly correctly here. But I removed it in the sake of consistency.

modules/node-shared/src/test/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorageSuite.scala
effects-ai: Done

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
effects-ai: Reverted it back, becase it's needed to synchronized updates of block and transaction indexes.

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/ValidationErrorRoutes.scala
effects-ai: Done

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
effects-ai: Done

modules/dag-l0/src/test/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotTraverseSuite.scala
effects-ai: Done

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
effects-ai: Done.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotCreator.scala
effects-ai: Messages are not ignored in the encoder anymore.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/config/types.scala
```diff
@@ -147,4 +149,13 @@ object types {
 
   case class AddressesConfig(locked: Set[Address])
 
+  case class ValidationErrorStorageConfig(size: PosInt)
```
marcinwadon, line 152: Please take a look how other configurable variables are handled in the codebase (see: app conf). There is no such pattern like you did.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
marcinwadon: If so, do you share the semaphore between those two processes? By your description, looks like you need a mutually exclusive semaphore (mutex). Can you specify what the current semaphore actually does?

Anyway, please try to write lock-free code if possible. Here, you can calculate things upfront before making an update, so I don't think that the semaphore is required. Can you point me to an exact place that requires locking? Also, the semaphore should lock exactly that place instead of the whole function

marcinwadon (changes requested):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
marcinwadon: what does it override? I don't see any implementation provided that it overrides

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/ValidationErrorRoutes.scala
marcinwadon: afaik we spoke about `/transactions/:txhash/errors` with no `/snapshots` prefix 🤔 
Also, what are events errors?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
marcinwadon: Ok, but it's still a single error, not multiple errors like the parameter name suggests

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
marcinwadon: Ok. Is it okay that eviction will be different across different nodes?

marcinwadon: If you need timestamp purely for the eviction and not as a presentation data, you can keep the timestamp internal when storing an error.

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/ValidationErrorRoutes.scala
```diff
@@ -0,0 +1,86 @@
...
+          a.artifact match {
+            case currency.BlockEvent(block)                    => block.asJson
+            case currency.DataApplicationBlockEvent(dataBlock) => dataBlock.asJson
+            case currency.CurrencyMessageEvent(message)        => message.asJson
+          }
+        ),
+        ("reason", blockRejectionReasonToValidationError(a.validationError).asJson),
+        ("timestamp", a.timestamp.asJson)
+      )
+    }
```
marcinwadon, line 73: instead of manually crafting JSONs, please create a case class representing the returned type and automatic derivation will create an encoder or decoder.

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/ValidationErrorRoutes.scala
```diff
@@ -0,0 +1,86 @@
...
+
+  protected val prefixPath: InternalUrlPrefix = "/snapshots"
+
+  protected val public: HttpRoutes[F] = HttpRoutes.of[F] {
+    case GET -> Root / "events" / "errors" =>
```
marcinwadon, line 78: I think that "events" is incorrectly used there. Block is used as "an event for consensus", but it is not "an event". Saying that those errors are events' errors is imo wrong.

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/ValidationErrorRoutes.scala
```diff
@@ -0,0 +1,86 @@
...
+            case currency.DataApplicationBlockEvent(dataBlock) => dataBlock.asJson
+            case currency.CurrencyMessageEvent(message)        => message.asJson
+          }
+        ),
+        ("reason", blockRejectionReasonToValidationError(a.validationError).asJson),
+        ("timestamp", a.timestamp.asJson)
+      )
+    }
+
+  protected val prefixPath: InternalUrlPrefix = "/snapshots"
```
marcinwadon, line 75: afaik we agreed on the following endpoint format: `/transactions/<hash>/errors`.
`/snapshots/transactions/<hash>/errors` does not match current transactions endpoints.

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/modules/HttpApi.scala
```diff
@@ -73,7 +77,8 @@ sealed abstract class HttpApi[F[_]: Async: SecurityProvider: HasherSelector: Met
   nodeVersion: TessellationVersion,
   httpCfg: HttpConfig,
   maybeDataApplication: Option[BaseDataApplicationL0Service[F]],
-  maybeMetagraphVersion: Option[MetagraphVersion]
+  maybeMetagraphVersion: Option[MetagraphVersion],
+  currencySnapshotEventValidationErrorStorage: ValidationErrorStorage[F, CurrencySnapshotEvent, BlockRejectionReason]
```
marcinwadon, line 81: there is already `storages` parameter passed to `HttpApi`, so there is no need to pass the storage individually

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
```diff
@@ -0,0 +1,76 @@
+package io.constellationnetwork.node.shared.infrastructure.consensus
+
+import java.time.Instant
+
+import cats.effect.kernel.{Async, Clock, Ref}
+import cats.effect.std.Semaphore
+import cats.implicits.toTraverseOps
+import cats.syntax.flatMap._
+import cats.syntax.functor._
```
marcinwadon, line 9: please don't mix both types of importing implicits. we use `cats.syntax` in the codebase

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
```diff
@@ -0,0 +1,76 @@
...
+import eu.timepit.refined.types.all.PosInt
+
+trait ValidationErrorStorage[F[_], Artifact, ValidationError] {
+  def add(artifact: Artifact, error: ValidationError): F[Unit]
+
+  def add(errors: List[(Artifact, ValidationError)]): F[Unit]
+
+  def getAll: F[Vector[ValidationErrorStorageEntry[Artifact, ValidationError]]]
+
+  def get(transactionHash: Hash): F[Option[ValidationError]]
```
marcinwadon, line 23: can it be `get(hash: Hash): ...`? there cannot be two same hashes, and you can return validation errors for all kinds of stored data

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/modules/SharedServices.scala
```diff
@@ -84,12 +88,29 @@ object SharedServices {
 
       currencyEventsCutter = CurrencyEventsCutter.make[F](None)
 
+      currencySnapshotEventValidationErrorStorage <- HasherSelector[F].withCurrent(implicit hasher =>
+        ValidationErrorStorage.make[F, CurrencySnapshotEvent, BlockRejectionReason](
+          cfg.validationErrorStorage.size,
+          { event: CurrencySnapshotEvent =>
+            import cats.syntax.all._
```
marcinwadon, line 95: please move it to the top of the file

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/modules/SharedServices.scala
```diff
@@ -84,12 +88,29 @@ object SharedServices {
...
+
+            event match {
+              case BlockEvent(value) =>
+                value.value.transactions.toList.traverse(x => x.value.hash)
```
marcinwadon, line 99: `traverse(x => x.value.hash` -> `traverse(_.value.hash)`

modules/test-shared/src/main/scala/io/constellationnetwork/testconstants.scala
```diff
@@ -0,0 +1,9 @@
+package io.constellationnetwork
+
+import eu.timepit.refined.types.numeric.PosInt
+
+object testconstants {
```
marcinwadon, line 5: please don't create such "generic" containers. you can move it to the specific spec file as it is not reusable and bounded to the specific scenario.

modules/node-shared/src/test/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorageSuite.scala
```diff
@@ -0,0 +1,126 @@
...
+
+import cats.effect.IO
+
+import io.constellationnetwork.security.hash.Hash
+
+import eu.timepit.refined.types.numeric.PosInt
+import weaver.SimpleIOSuite
+
+object ValidationErrorStorageSuite extends SimpleIOSuite {
+  test("size based eviction") {
```
marcinwadon, line 11: @scasplte2 asked us to make a review of the metakit repo. There, I've already discussed with him that way of creating tests and why I consider them hard to read and maintain. If you didn't review the PR yet, please do so.

https://github.com/Constellation-Labs/metakit/pull/1#discussion_r1851120794

marcinwadon (changes requested):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
effects-ai: The value of `storedDataR` can be changed between lines 50 and 56, so the code path should be synchronized.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
effects-ai: I think so. It's a known limitation.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/config/types.scala
effects-ai: Done

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
marcinwadon: I understand that, but you can write it in a way that within a single operation you `get` and `update`, instead of separate operations, so there is no need for synchronization or using semaphores, and you write lock-free code then.

effects-ai: The issue is that we need to store validation errors not only for transactions but also for data updates, and potentially for other data types as well. The question, then, is how to create a generic data structure that can accommodate all possible data types. Storing just a `Hash` seems like a simple solution, as the REST API returns errors identified by hash.

I'm considering eliminating artifact storage entirely and storing only hashes along with their corresponding validation errors. 

In this case, the caller could extract the transaction (and other types of) hashes, so we wouldn't need to handle hashing within the storage. Additionally, this approach would eliminate the need for the semaphore, as we would only need to store a single `Map[Hash, ValidationError]`.

WDYT?

modules/node-shared/src/test/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorageSuite.scala
effects-ai: Fixed

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
effects-ai: Done

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/modules/SharedServices.scala
effects-ai: Done

modules/test-shared/src/main/scala/io/constellationnetwork/testconstants.scala
effects-ai: Done

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
effects-ai: Done

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/modules/SharedServices.scala
effects-ai: Done

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/modules/HttpApi.scala
effects-ai: The issue is that a `ValidationErrorStorage` is required to make a `CurrencySnapshotValidator` in the `SharedServices`, but the `Storages` has not been initialized before `SharedServices.make` is called. Would you like me to make `ValidationErrorStorage` a part of `SharedStorages` so I can pass it to `SharedServices.make`, rather than creating it there?

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/ValidationErrorRoutes.scala
effects-ai: Done

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/ValidationErrorRoutes.scala
effects-ai: Done

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/ValidationErrorRoutes.scala
effects-ai: Done

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/modules/HttpApi.scala
effects-ai: Done

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
```diff
@@ -0,0 +1,76 @@
...
+
+          def add(errors: List[(Artifact, ValidationError)]): F[Unit] =
+            errors.traverse {
+              case (artifact, error) =>
+                add(artifact, error)
+            }.void
+
+          def getAll: F[Vector[ValidationErrorStorageEntry[Artifact, ValidationError]]] = storedDataR.get.map(_.artifacts)
+
+          def get(hash: Hash): F[Option[ValidationError]] = storedDataR.get.map(_.transactions.get(hash))
```
marcinwadon, line 73: what about other types? shouldn't `get` return data update hashes or other events?

modules/node-shared/src/test/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorageSuite.scala
```diff
@@ -0,0 +1,126 @@
...
+
+import cats.effect.IO
+
+import io.constellationnetwork.security.hash.Hash
+
+import eu.timepit.refined.types.numeric.PosInt
+import weaver.SimpleIOSuite
+
+object ValidationErrorStorageSuite extends SimpleIOSuite {
+  test("validation error storage evicts older entries based on size when adding entries one by one") {
```
marcinwadon, line 11: you don't need to repeat "validation error storage" in every test, because it's natural that you are in the scope of ValidationErrorStorage suite. it's enough to write `"evicts older entries based on size when adding entries one by one"`

modules/node-shared/src/test/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorageSuite.scala
```diff
@@ -0,0 +1,126 @@
...
+        Hash("51")
+      )
+    } yield
+      expect
+        .same(IndexedSeq(), rejectedArtifacts)
+        .and(expect.same(List(None, None, None, None, None, None, None, None, None), rejectedTransactions))
+  }
+
+  private def getRejectedTransactions(storage: ValidationErrorStorage[IO, Int, String], txns: Hash*): IO[List[Option[String]]] = {
+    import cats.syntax.all._
```
marcinwadon, line 94: please move the import to the top

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  def add(errors: List[(Artifact, ValidationError)]): F[Unit]
+
+  def getAll: F[Vector[ValidationErrorStorageEntry[Artifact, ValidationError]]]
+
+  def get(hash: Hash): F[Option[ValidationError]]
+}
+
+case class ValidationErrorStorageEntry[Artifact, ValidationError](
+  artifact: Artifact,
```
marcinwadon, line 24: I've just realized that `artifact` name will be misleading here. You pass `CurrencySnapshotEvent` as an `Artifact` but in every other place in the codebase we treat snapshots as `Artifacts`. As the `Artifact` is **an output** of the consensus process. could you please rename it, so we would keep it consistent and not ambiguous? in fact, you use input of the consensus here (events). so maybe just `event: ConsensusEvent` or equivalent?

marcinwadon (dismissed):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
marcinwadon: You still return validation errors from narrowed scope (transactions only)

marcinwadon: If you need transactions only, the function should be named `getTransaction(hash: Hash)`. `get` suggests that you query for entries, so blocks

AlexBrandes (changes requested):
Since this PR adds new API endpoints, it needs to have API docs for those endpoints added before it can be merged. -> /docs/api

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
effects-ai: Fixed

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/ValidationErrorRoutes.scala
```diff
@@ -0,0 +1,99 @@
...
+) extends Http4sDsl[F]
+    with PublicRoutes[F] {
+  import models._
+
+  protected val prefixPath: InternalUrlPrefix = "/snapshots"
+
+  protected val public: HttpRoutes[F] = HttpRoutes.of[F] {
+    case GET -> Root / "blocks" / "errors" =>
+      currencySnapshotValidationErrorStorage.getAll
+        .map(_.map(RejectedBlock(_)))
```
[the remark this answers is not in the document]
effects-ai: Yeah, bad naming. It's not a block per se, but an error associated with a rejected block. Fixed.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
effects-ai: Fixed.

modules/node-shared/src/test/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorageSuite.scala
effects-ai: Fixed.

modules/node-shared/src/test/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorageSuite.scala
effects-ai: Done.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
effects-ai: Fixed.

marcinwadon:
I think there is something wrong in the PR as it shows all api docs files as new, instead of merging your changes into existing ones.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
```diff
@@ -0,0 +1,81 @@
...
+        )
+      )
+      .map { storedDataR =>
+        new ValidationErrorStorage[F, ConsensusEvent, ValidationError] {
+          def add(ConsensusEvent: ConsensusEvent, error: ValidationError): F[Unit] =
+            for {
+              timestamp <- Clock[F].realTimeInstant
+              transactions <- extractTransactionHashes(ConsensusEvent)
+              res <- storedDataR.update { storedData =>
+                val ConsensusEventsToDrop =
```
marcinwadon, line 55: we don't use pascal case for variables (except object variables like MinValue or Default)

effects-ai:
> I think there is something wrong in the PR as it shows all api docs files as new, instead of merging your changes into existing ones.

The feature branch needs to be rebased. I'm not allowed to to it so. I rebased only the PR's branch.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ValidationErrorStorage.scala
effects-ai: Fixed

ryle-goehausen (approved):

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/TransactionValidationErrorRoutes.scala
```diff
@@ -0,0 +1,33 @@
...
+    with PublicRoutes[F] {
+
+  import models._
+
+  protected val prefixPath: InternalUrlPrefix = "/transactions"
+
+  protected val public: HttpRoutes[F] = HttpRoutes.of[F] {
+    case GET -> Root / txHash / "errors" =>
+      currencySnapshotValidationErrorStorage
+        .getTransaction(Hash(txHash))
```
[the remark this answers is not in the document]
kpudlik: I think that `unapply` will be missing in `Hash` to make it work but I agree :+1:

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/ValidationErrorRoutes.scala
```diff
@@ -0,0 +1,99 @@
...
+
+object models {
+  @derive(encoder)
+  case class RejectedDataApplicationBlock(roundId: RoundId, updatesHashes: NonEmptyList[Hash])
+  object RejectedDataApplicationBlock {
+    def apply(a: DataApplicationBlock): RejectedDataApplicationBlock =
+      RejectedDataApplicationBlock(a.roundId, a.updatesHashes)
+  }
+
+  implicit val currencySnapshotEventEncoder: Encoder[CurrencySnapshotEvent] =
```
kpudlik, line 51: How it differs from the encoder that is in `CurrencySnapshotEvent`'s companion object?

kpudlik (dismissed):

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/ValidationErrorRoutes.scala
effects-ai: The issue is that `Decoder[DataUpdate]` is not defined in the current scope here, so I created a wrapper (`RejectedDataApplicationBlock`) for `DataApplicationBlockEvent` that holds only the update hashes, making it easier to serialize.

doc/api/models/schemas/ValidationError.yml
```diff
@@ -0,0 +1,10 @@
+type: object
+properties:
+  message:
+    type: string
+    example: "snapshot ordinal unavailable"
+  description:
+    type: string
+    example: "Snapshot ordinal unavailable"
+  details:
+    type: object
```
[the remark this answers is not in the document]
effects-ai: Yes. But details field is optional. I added an example.

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/TransactionValidationErrorRoutes.scala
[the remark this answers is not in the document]
effects-ai: There is `unapply` in `HashVar` which does the job.

doc/api/currency/l0/public.yml
```diff
@@ -278,6 +278,73 @@ paths:
             Request-Signature:
               $ref: '../../models/headers/Request-Signature.yml'
 
+  /snapshots/block/errors:
```
marcinwadon, line 281: I've just spotted it. Probably it should be `/snapshots/blocks/errors` (`blocks` as plural) as there is not a single block in snapshot but many. @AlexBrandes what do you think?

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/ValidationErrorRoutes.scala
marcinwadon: Shouldn't it be in `RejectedDataApplicationBlock` companion's object then? Having it in `models` may produce false-positives if I'd import models (as I will have a wrong encoder imported).

ryle-goehausen (approved):
Looks like most all comments have been addressed apart from apidoc related one, which should be minor. If you can rebase off new dev merged branch this should be ready to merge otherwise.

doc/api/currency/l0/public.yml
effects-ai: Fixed.

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/ValidationErrorRoutes.scala
marcinwadon: @effects-ai bump

AlexBrandes (dismissed):

ryle-goehausen (approved):
approved assuming tests pass

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/http/routes/ValidationErrorRoutes.scala
effects-ai: Fixed.

ryle-goehausen (approved):

PR #991, opened 2025-01-07
ci: workflow to publish sdk artifact to sonatype central

scasplte2:
## Overview
This PR introduces a workflow and sbt plugin to automate publication of the `sdk` module. The workflow is modeled after the release workflow and relies on the sbt plugin `TessellationCiRelease` which is derived from [sbt-ci-release](https://github.com/sbt/sbt-ci-release). 

A custom version of the release plugin was used to avoid conflicts due to Tessellation's use of `GitVersioningPlugin` which was causing inconsistencies. This custom version of the plugin removes functionality of the original so it may only be run via Github Actions and will only publish to Sonatype Central.

## Changes
- Updated project_template to only depend `sdk` module
- Added artifact publication step to the release workflow
- Added necessary project information for publishing to build.sbt
- Updated `sdk` project definition to include internal module classfiles, source, docs, and transitive dependencies
- Add `TessellationCiRelease` plugin to handle artifact generation, signing, and publishing
- Renamed `application.conf` within `dag-l0, dag-l1, currency-l0, & currency-l1` to avoid file conflicts in `sdk` bundling
- Updated primary IOApp in each of `dag-l0, dag-l1, currency-l0, & currency-l1` to load the app specific config file
- Added `networkStateAfterJoining` val to `TessellationIoApp` so it is available at runtime for Joining program
- Removed `stateAfterJoining` from `NodeSharedConfig` as this parameter is app specific and not available in the scope of loading this config
- Added plugins `sbt-git, sbt-pgp, sbt-sonatype` as they are used by the `TessellationCiRelease` process

## Testing
- Ran the workflow locally via `act` and successfully built sdk artifact and pushed to Sonatype Central (there was an intentional error in the bundle pushed to Sonatype to avoid actual publication)


.github/workflows/publish-sdk.yml
```diff
@@ -0,0 +1,38 @@
...
+jobs:
+  publish:
+    if: startsWith(github.ref, 'refs/tags/')
+    name: Build, test, and publish
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Check if project version is release version
+        run: test "v$(cut -d \" -f 2 version.sbt)" = "${{ github.ref_name }}"
```
marcinwadon, line 17: I think you've made it against the obsolete version of the code. we don't support `version.sbt` right now. IMO you should prepare it against the feature/allow-spends branch that will be a base for v3. We don't want to publish old version of the code to the public repo.

build.sbt
```diff
@@ -551,6 +563,7 @@ lazy val sdk = (project in file("modules/sdk"))
     scalafixCommonSettings,
     commonSettings,
     commonTestSettings,
+    publish / skip := false,
```
marcinwadon, line 566: do I understand it correctly that only the sdk module would be published? if so, that's correct :+1:

marcinwadon (changes requested):

scasplte2:
There is one point that may be worth a discussion which is that `sbt-ci-release` depends on several other plugins, specifically:
[sbt-dynver](https://github.com/dwijnand/sbt-dynver): sets the version number based on your git history
[sbt-pgp](https://github.com/sbt/sbt-pgp): to cryptographically sign the artifacts before publishing
[sbt-sonatype](https://github.com/xerial/sbt-sonatype): to publish artifacts to Sonatype
[sbt-git](https://github.com/sbt/sbt-git): to automatically populate scmInfo

However, we are using the [GitVersioningPlugin](https://github.com/rallyhealth/sbt-git-versioning) for setting our version for other tasks and the two plugins automatically generate differing formats for non-tagged commits. I don't think this will be an issue since this workflow should only be run on tagged commits, in which case the two tools will derive the same version label but I wanted to get y'all's thoughts on it.

build.sbt
scasplte2: yes, only sdk will publish (this was confirmed on the Sonatype end as well). It is a bit of an odd syntax but this line is overriding the `publish / skip := true` value set in `commonSettings` (which is used by nearly every module). If we'd prefer to update all the modules definitions instead, I can do that.

.github/workflows/publish-sdk.yml
scasplte2: oh, you're right, I had checked out a previous tag for testing and was thinking that `version.sbt` was being auto-generated. I'll update this.

scasplte2: @marcinwadon it seems like this test was replaced by the snippet below in the release workflow but I am not sure I follow what the circumstance would be to abort the release.

```
      - name: Check if the commit already has a tag starting with "v"
        id: check_tag
        run: |
          TAG=$(git tag --contains HEAD | grep -E '^v')
          if [ -n "$TAG" ]; then
            echo "This commit has a tag starting with 'v': $TAG"
            echo "TAG_EXISTS=true" >> "$GITHUB_ENV"
          else
            echo "This commit does not have a tag starting with 'v'."
            echo "TAG_EXISTS=false" >> "$GITHUB_ENV"
          fi

      - name: Check branch and abort if necessary
        if: env.TAG_EXISTS == 'true' && github.ref_name == 'release/integrationnet'
        run: |
          echo "Aborting the action because the release should already exist."
          exit 1
```

Perhaps I could just remove the check as Sonatype will enforce that there is a unique artifact for a given tag. I originally added this as a measure to ensure correctness of the tag that is used during publication since I (incorrectly) thought `version.sbt` was being auto-generated.

build.sbt
marcinwadon: No, it's fine for me. Just wanted to make sure

marcinwadon (approved):

.github/workflows/publish-sdk.yml
marcinwadon: Ok, so basically when we were discussing the new release flow, we decided that everything that goes to mainnet, should go through integrationnet (and previously through the testnet). By saying that, I mean creating a tag version for testnet environment, then the same code being pushed to integrationnet, and then the same code being pushed to mainnet. It shouldn't trigger the release flow again and again, because it's the same code and should be released once. So here's the check for that. As you already working with this, please make a double check if the check is correct because I feel my PR was not verified correctly.

scasplte2:
> Hey @scasplte2 I'm seen that the actions are failing due:
> ```
> Exception in thread "main" java.lang.ExceptionInInitializerError
> 	at io.constellationnetwork.dag.l0.Main.main(Main.scala)
> Caused by: java.lang.IllegalArgumentException: Predicate failed: "0.0.0+1-b8de29ef+20250108-0020-SNAPSHOT".matches("^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$").
> 	at eu.timepit.refined.internal.RefinePartiallyApplied.$anonfun$unsafeFrom$1(RefinePartiallyApplied.scala:21)
> 	at scala.util.Either.fold(Either.scala:190)
> 	at eu.timepit.refined.internal.RefinePartiallyApplied.unsafeFrom(RefinePartiallyApplied.scala:21)
> 	at io.constellationnetwork.schema.semver$TessellationVersion$.unsafeFrom(semver.scala:30)
> 	at io.constellationnetwork.dag.l0.Main$.<init>(Main.scala:46)
> 	at io.constellationnetwork.dag.l0.Main$.<clinit>(Main.scala:42)
> 	... 1 more
> ```
> 
> probably it's a version checker

Yes, definitely the semver check not liking the tag it is being given. I suspect this may be a result of the conflicting plugin issue since this want happening before. Thanks for looking into it @IPadawans

.github/workflows/publish-sdk.yml
IPadawans: Maybe we can keep testnet version as we have nowadays, where we generate locally to test on metagraphs, and we can release only official versions of mainnet, WDYT? I mean we want that public metagraph developers use official versions

marcinwadon:
> > Hey @scasplte2 I'm seen that the actions are failing due:
> > ```
> > Exception in thread "main" java.lang.ExceptionInInitializerError
> > 	at io.constellationnetwork.dag.l0.Main.main(Main.scala)
> > Caused by: java.lang.IllegalArgumentException: Predicate failed: "0.0.0+1-b8de29ef+20250108-0020-SNAPSHOT".matches("^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$").
> > 	at eu.timepit.refined.internal.RefinePartiallyApplied.$anonfun$unsafeFrom$1(RefinePartiallyApplied.scala:21)
> > 	at scala.util.Either.fold(Either.scala:190)
> > 	at eu.timepit.refined.internal.RefinePartiallyApplied.unsafeFrom(RefinePartiallyApplied.scala:21)
> > 	at io.constellationnetwork.schema.semver$TessellationVersion$.unsafeFrom(semver.scala:30)
> > 	at io.constellationnetwork.dag.l0.Main$.<init>(Main.scala:46)
> > 	at io.constellationnetwork.dag.l0.Main$.<clinit>(Main.scala:42)
> > 	... 1 more
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > probably it's a version checker
> 
> Yes, definitely the semver check not liking the tag it is being given. I suspect this may be a result of the conflicting plugin issue since this want happening before. Thanks for looking into it @IPadawans

I think it's just a version that is assigned for testing that does not match the predicate 🤔

.github/workflows/publish-sdk.yml
scasplte2: If we tag a version of the codebase and release to testnet, then I would consider that an official version. I agree that we should continue to use locally built artifacts as a first step before tagging but if we don't maintain a 1:1 correspondence between releases and published artifacts, then it would seem the next problem becomes "How do external devs know which version of `tessellation-sdk` is compatible with the executable packages `dag-l0, dag-l1, keytool, and wallet` published in a release"?

project/TessellationCiRelease.scala
```diff
@@ -0,0 +1,164 @@
...
+    publishTo := sonatypePublishToBundle.value
+  )
+
+  private def getVersion(state: State): String =
+    (ThisBuild / version).get(Project.extract(state).structure.data) match {
+      case Some(v) => v
+      case None    => throw new NoSuchFieldError("version")
+    }
+
+  private def isSnapshotVersion(v: String): Boolean = v.endsWith("-SNAPSHOT")
```
marcinwadon, line 134: The version may end with `-SNAPSHOT-dirty` and it should be also considered a snapshot version 🤔 instead of `endsWith` I'd use `contains("-SNAPSHOT")`

.github/workflows/publish-sdk.yml
```diff
@@ -0,0 +1,36 @@
...
+        with:
+          JAVA_VERSION: openjdk@1.11.0
+        uses: "./.github/templates/setup_java_and_scala"
+
+      - name: Run linter checks and unit tests
+        run: sbt 'scalafixAll --check --rules OrganizeImports;scalafmtCheckAll;test'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Publish to Sonatype Central
```
marcinwadon, line 29: As agreed on the meeting, please make it a part of the current release process

.github/workflows/publish-sdk.yml
scasplte2: moved the artifact publication to be the final step of the release process as discussed

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/app/TessellationIOApp.scala
```diff
@@ -72,6 +73,8 @@ abstract class TessellationIOApp[A <: CliMethod](
     */
   val kryoRegistrar: Map[Class[_], KryoRegistrationId[KryoRegistrationIdRange]]
 
+  val networkStateAfterJoining: NodeState
```
scasplte2, line 76: @marcinwadon we may wish to discuss this variable as I am not really a fan of having to pull it out to this level. It was done because previously we were relying on the assembly process to merge the various `application.conf` files for each app for us. Since the config parameters were merged, the config reader (line 94) had access to the networkState parameter. However, since I renamed the app configs to avoid collisions within the bundled `sdk` jar, this networkState parameter is no longer merged into an `application.conf` file and thus is not present at this layer. I couldn't think of any other way to get this information so I added this `val` and specify it within each app.

If you (or anyone else) can think of a way to avoid this I would be appreciative.

modules/currency-l1/src/main/scala/io/constellationnetwork/currency/l1/CurrencyL1App.scala
```diff
@@ -77,11 +77,18 @@ abstract class CurrencyL1App(
...
+  val networkStateAfterJoining: NodeState = NodeState.Ready
+
   def run(method: Run, nodeShared: NodeShared[IO, Run]): Resource[IO, Unit] = {
     import nodeShared._
 
     for {
-      cfgR <- ConfigSource.default.loadF[IO, AppConfigReader]().asResource
+      cfgR <- ConfigSource
+        .resources("currency-l1.conf")
+        .withFallback(ConfigSource.resources("dag-l1.conf"))
```
scasplte2, line 88: @marcinwadon this was a consequence of the assembly merging the `application.conf` files. I could replicate the settings from the `dag-l1.conf` into `currency-l1.conf` but went with this approach instead so that any changes to parameters within `dag-l1.conf` would also propagate to CurrencyL1. This is more closely aligned with the previous behavior rather than duplicating the parameters between the two configs.

.github/workflows/test-currency-transactions.yml
```diff
@@ -41,7 +41,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          sbt -Dversion.override=99.99.99 --error clean compile sdk/publishM2 shared/publishM2 kernel/publishM2 keytool/publishM2 nodeShared/publishM2 currencyL0/publishM2 currencyL1/publishM2 dagL1/publishM2
```
scasplte2, line 44: publication is skipped for all other modules but sdk, so these additional commands wouldn't create any artifacts.

.github/workflows/release.yml
```diff
@@ -144,3 +144,12 @@ jobs:
...
             ./mainnet/mainnet-seedlist
+
+      - name: Publish SDK module to Sonatype Central
+        run: sbt tessellation-ci-release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
+          PGP_SECRET: ${{ secrets.PGP_SECRET }}
+          SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
+          SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
```
scasplte2, line 155: the end of the workflow seemed the most reasonable but I'm happy to move somewhere else if there is a preference

ryle-goehausen (approved):

IPadawans (approved):

PR #994, opened 2025-01-15
feat: majority alignment and allow spends on global layer

marcinwadon:


marcinwadon:
FYI - it's missing API doc update, but other than that - ready for review

modules/dag-l1/src/main/scala/io/constellationnetwork/dag/l1/Main.scala
```diff
@@ -158,6 +159,12 @@ object Main
         )
         .asResource
 
+      alignment = Alignment.make[IO, GlobalSnapshotStateProof, GlobalIncrementalSnapshot, GlobalSnapshotInfo, Run](
```
scasplte2, line 162: just a small thing I noticed, but it is a little confusing to have a `class Alignment` and a `sealed trait Alignment` both defined in `dag-l1`  ([trait](https://github.com/Constellation-Labs/tessellation/blob/ce02a6fa52d0016849c13cc9e1666b682ae82eaa/modules/dag-l1/src/main/scala/io/constellationnetwork/dag/l1/domain/snapshot/programs/SnapshotProcessor.scala#L421))

marcinwadon: Good catch. I've just changed it to `GlobalSnapshotAlignment`

IPadawans (approved):

scasplte2 (approved):
lgtm

PR #999, opened 2025-02-03
feat: update node parameters process

effects-ai:


modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/http/routes/NodeParametersRoutes.scala
```diff
@@ -0,0 +1,82 @@
...
+      req
+        .as[Signed[UpdateNodeParameters]]
+        .map(mkCell)
+        .flatMap(_.run())
+        .flatMap {
+          case Left(_)  => BadRequest()
+          case Right(_) => Ok()
+        }
+    case req @ GET -> Root / nodeId =>
+      nodeId
```
marcinwadon, line 64: `nodeId` should be created as a http4s Var (take a look how `Address` is being used in http routes)

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/http/routes/NodeParametersRoutes.scala
```diff
@@ -0,0 +1,82 @@
...
+        .flatMap { nodeId =>
+          nodeStorage.getNodeState
+            .map(validStateForSnapshotReturn)
+            .ifM(
+              resolveEncoder[F, UpdateNodeParameters](req) { implicit enc =>
+                latestNodeParameters(nodeId).flatMap {
+                  case Some(params) => Ok(params)
+                  case _            => NotFound()
+                }
+              },
```
marcinwadon, line 76: do you need to support multiple encoders here?

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/package.scala
```diff
@@ -19,7 +20,10 @@ package object l0 {
   type DagL0KryoRegistrationId = KryoRegistrationId[DagL0KryoRegistrationIdRange]
 
   val dagL0KryoRegistrar: Map[Class[_], DagL0KryoRegistrationId] = Map(
-    classOf[StateChannelOutput] -> 700
+    classOf[StateChannelOutput] -> 700,
+    classOf[StateChannelEvent] -> 701,
+    classOf[DAGEvent] -> 702,
+    classOf[UpdateNodeParametersEvent] -> 703
```
marcinwadon, line 26: kryo should not be used anymore. please don't add new classes to registration

modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshot.scala
```diff
@@ -41,6 +43,7 @@ case class GlobalIncrementalSnapshot(
   nextFacilitators: NonEmptyList[PeerId],
   tips: SnapshotTips,
   stateProof: GlobalSnapshotStateProof,
+  updateNodeParameters: SortedMap[Id, Signed[UpdateNodeParameters]],
```
marcinwadon, line 46: when you add a new field to schema, please use wrap it into `Option` field, so the change will be backward-compatible using json serializer

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/package.scala
effects-ai: But kryo is used in unit test. Without these lines `GlobalSnapshotEventCutterSuite` and `GlobalSnapshotConsensusFunctionsSuite` fail

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/http/routes/NodeParametersRoutes.scala
effects-ai: Fixed

modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshot.scala
effects-ai: Fixed

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/http/routes/NodeParametersRoutes.scala
effects-ai: Not really, fixed.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/package.scala
marcinwadon: Yes, this is what I fixed in my PR. Kryo was incorrectly used there in tests.

modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshotInfo.scala
```diff
@@ -29,10 +31,12 @@ import io.circe.disjunctionCodecs._
 case class GlobalSnapshotInfoV1(
   lastStateChannelSnapshotHashes: SortedMap[Address, Hash],
   lastTxRefs: SortedMap[Address, TransactionReference],
-  balances: SortedMap[Address, Balance]
+  balances: SortedMap[Address, Balance],
+  updateNodeParameters: Option[SortedMap[Id, Signed[UpdateNodeParameters]]]
 ) extends SnapshotInfo[GlobalSnapshotStateProof] {
   def stateProof[F[_]: Sync: Hasher](ordinal: SnapshotOrdinal): F[GlobalSnapshotStateProof] =
     GlobalSnapshotInfoV1.toGlobalSnapshotInfo(this).stateProof[F](ordinal)
+  // def updateNodeParameters: Option[SortedMap[Id, Signed[UpdateNodeParameters]]] = None
```
marcinwadon, line 39: leftover?

modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshotInfo.scala
```diff
@@ -70,7 +75,8 @@ case class GlobalSnapshotInfoV2(
   lastCurrencySnapshots: SortedMap[Address, Either[Signed[
     CurrencySnapshot
   ], (Signed[CurrencyIncrementalSnapshotV1], CurrencySnapshotInfoV1)]],
-  lastCurrencySnapshotsProofs: SortedMap[Address, Proof]
+  lastCurrencySnapshotsProofs: SortedMap[Address, Proof],
+  updateNodeParameters: Option[SortedMap[Id, Signed[UpdateNodeParameters]]]
```
marcinwadon, line 79: other versions like `GlobalSnapshotInfoV1` or `GlobalSnapshotInfoV2` are locked for kryo deserialization (it's an older schema), so you should not modify their properties. it's that way to be backward compatible and be able to deserialize historical snapshots.

modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshotInfo.scala
```diff
@@ -146,6 +155,7 @@ object GlobalSnapshotInfo {
     SortedMap.empty,
     SortedMap.empty,
     Some(SortedMap.empty),
-    Some(SortedMap.empty)
+    Some(SortedMap.empty),
+    None
```
marcinwadon, line 159: probably `Some(SortedMap.empty)`

modules/shared/src/main/scala/io/constellationnetwork/schema/node.scala
```diff
@@ -116,4 +118,25 @@ object node {
...
+
+  @derive(eqv, show, encoder, decoder)
+  case class DelegatedStakeRewardParameters(rewardFraction: Int) {
+    def rewardFractionDouble: Double = rewardFraction.toDouble / DelegatedStakeRewardParameters.MaxRewardFraction
+  }
+
+  object DelegatedStakeRewardParameters {
+    val MaxRewardFraction = 100_000_000
+    val MinRewardValue = 5_000_000
+    val MaxRewardValue = 10_000_000
```
marcinwadon, line 130: should it be parametrized from config?

modules/shared/src/test/scala/io/constellationnetwork/security/HashSuite.scala
```diff
@@ -38,13 +38,14 @@ object HashSuite extends MutableIOSuite with Checkers {
...
         )
       }
     }
 
   test("ensure backward compatibility") { implicit res =>
-    def oldHash = Hash("6512e0fdd9e2b870ff6124b86744ad8e1eedb7cd4281fd7b9a36a0457e1bfdcb")
+    def oldHash = Hash("79403710a0d048dc3400bb6f400f013495e2a76c8903801e4659f202733e004f")
+//    def oldHash = Hash("6512e0fdd9e2b870ff6124b86744ad8e1eedb7cd4281fd7b9a36a0457e1bfdcb")
 
     def genesis = GlobalSnapshot.mkGenesis(Map.empty, EpochProgress.MinValue)
```
marcinwadon, line 51: please don't modify this test. It's there to verify backward compatibility. If the test is failing, then the code is not backward compatible

marcinwadon (changes requested):

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/package.scala
effects-ai: Done.

modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshotInfo.scala
effects-ai: Removed.

modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshotInfo.scala
effects-ai: Fixed.

modules/shared/src/test/scala/io/constellationnetwork/security/HashSuite.scala
effects-ai: I reverted it back

modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshot.scala
```diff
@@ -91,7 +95,7 @@ object GlobalSnapshot {
       SortedSet.empty,
       startingEpochProgress,
       nextFacilitators,
-      GlobalSnapshotInfoV1(SortedMap.empty, SortedMap.empty, SortedMap.from(balances)),
+      GlobalSnapshotInfoV1(SortedMap.empty, SortedMap.empty, SortedMap.from(balances), None),
```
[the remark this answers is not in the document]
effects-ai: Fixed

modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshotInfo.scala
effects-ai: Fixed

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/http/routes/NodeParametersRoutes.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  protected val prefixPath: InternalUrlPrefix = "/node-params"
+
+  private def validStateForSnapshotReturn(state: NodeState): Boolean = state === NodeState.Ready
+
+  private def readLatestFullSnapshot(maybeOrdinal: Option[SnapshotOrdinal]): F[Option[Signed[GlobalSnapshot]]] =
+    maybeOrdinal match {
+      case Some(ordinal) => fullGlobalSnapshotStorage.read(ordinal)
+      case _             => Option.empty[Signed[GlobalSnapshot]].pure[F]
+    }
```
marcinwadon, line 40: you can use operations like `.map`, `.flatMap` or `.traverse`. Since Option is traversable, you don't need to pattern match to remap or traverse it.

```scala
maybeOrdinal.traverse(fullGlobalSnapshotStorage.read)
```

modules/dag-l0/src/test/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotSerializationSuite.scala
```diff
@@ -39,7 +34,7 @@ object GlobalSnapshotSerializationSuite extends MutableIOSuite with Checkers {
     implicit0(j: JsonSerializer[IO]) <- JsonSerializer.forSync[IO].asResource
     hk = Hasher.forKryo[IO]
   } yield (ks, hk)
-
+  /*
```
marcinwadon, line 37: why?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
```diff
@@ -216,6 +224,31 @@ object GlobalSnapshotAcceptanceManager {
             (maybeMerkleTree, updatedLastCurrencySnapshotProofs).tupled
         }
 
+        updateNodeParameters = unpEvents
+          .filter(e =>
+            e.value.delegatedStakeRewardParameters.rewardFraction >= MinRewardValue && e.value.delegatedStakeRewardParameters.rewardFraction >= MaxRewardValue
```
marcinwadon, line 229: probably typo in second `>=`. otherwise, the first check does not make sense

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
```diff
@@ -216,6 +224,31 @@ object GlobalSnapshotAcceptanceManager {
...
+          SortedMap.empty[Id, (Signed[UpdateNodeParameters], SnapshotOrdinal)]
+        )
+
+        cutUpdateNodeParameters = updateNodeParameters.toList.sortWith { (a, b) =>
+          def ordinalDiff(id: Id): Long = {
+            val ordinalValue = ordinal.value.value
+            val lastOrdinalValue = lastSnapshotUpdateNodeParameters.get(id).map(_._2).getOrElse(ordinal).value.value
+            ordinalValue - lastOrdinalValue
+          }
+          ordinalDiff(a._1) <= ordinalDiff(b._1)
```
marcinwadon, line 244: what is `_1` and `_2`? can you use descriptive names for those?

modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshot.scala
```diff
@@ -173,7 +178,8 @@ object GlobalSnapshot {
         genesis.tips,
         stateProof,
         Some(SortedSet.empty),
-        Some(SortedMap.empty)
+        Some(SortedMap.empty),
+        genesis.info.updateNodeParameters.map(_.map { case (k, v) => (k, v._1) })
```
marcinwadon, line 182: please don't touch `GlobalSnapshot`. its schema is locked as we already made a genesis and we switched to incremental snapshots

modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshotInfo.scala
```diff
@@ -159,7 +163,8 @@ case class GlobalSnapshotInfo(
   lastCurrencySnapshotsProofs: SortedMap[Address, Proof],
   activeAllowSpends: Option[SortedMap[Option[Address], SortedMap[Address, SortedSet[Signed[AllowSpend]]]]],
   tokenLockBalances: Option[SortedMap[Address, SortedMap[Address, Balance]]],
-  lastAllowSpendRefs: Option[SortedMap[Address, AllowSpendReference]]
+  lastAllowSpendRefs: Option[SortedMap[Address, AllowSpendReference]],
+  updateNodeParameters: Option[SortedMap[Id, (Signed[UpdateNodeParameters], SnapshotOrdinal)]]
```
marcinwadon, line 167: when you update anything in GlobalSnapshotInfo, please remember to update the referenced state proof. I've written about that on slack lately

modules/shared/src/main/scala/io/constellationnetwork/schema/node.scala
```diff
@@ -116,4 +118,25 @@ object node {
...
+  }
+
+  @derive(eqv, show, encoder, decoder)
+  case class NodeMetadataParameters(name: String, description: String)
+
+  @derive(eqv, show, encoder, decoder, order, ordering)
+  case class UpdateNodeParameters(
+    delegatedStakeRewardParameters: DelegatedStakeRewardParameters,
+    nodeMetadataParameters: NodeMetadataParameters,
+    lastReference: TransactionReference
```
marcinwadon, line 140: why `Transaction` reference? how does it relate to DAG transactions?

modules/shared/src/main/scala/io/constellationnetwork/schema/node.scala
effects-ai: Done.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotEvent.scala
```diff
@@ -14,6 +15,12 @@ object event {
   @derive(eqv, decoder, encoder, show)
   sealed trait GlobalSnapshotEvent
 
+//  object GlobalSnapshotEvent {
+//    implicit val decoder: Decoder[GlobalSnapshotEvent] = deriveDecoder
+//
+//    implicit val encoder: Encoder[GlobalSnapshotEvent] = deriveEncoder
+//  }
+
```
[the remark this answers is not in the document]
effects-ai: Removed.

modules/shared/src/main/scala/io/constellationnetwork/schema/node.scala
effects-ai: Which type is it supposed to be?

modules/shared/src/main/scala/io/constellationnetwork/schema/node.scala
effects-ai: Yes.

modules/dag-l0/src/test/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotSerializationSuite.scala
effects-ai: I don't remember. God catch though!

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
effects-ai: yes, that's a typo

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotEvent.scala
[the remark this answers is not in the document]
effects-ai: yes

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/http/routes/NodeParametersRoutes.scala
effects-ai: Done

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
effects-ai: Fixed

modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshot.scala
effects-ai: Fixed

modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshotInfo.scala
effects-ai: Fixed

modules/shared/src/main/scala/io/constellationnetwork/schema/node.scala
ryle-goehausen: I think the pattern we're generally following for most of the other classes is to use a uniquely typed last reference per type (see AllowSpend for example for one of the new types.) It's not the greatest due to the fact we lack a common type, but I think you should follow that pattern for now.

ryle-goehausen (changes requested):
LGTM apart from unique lastref type

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/domain/snapshot/programs/UpdateNodeParametersCutter.scala
```diff
@@ -0,0 +1,60 @@
...
+
+trait UpdateNodeParametersCutter[F[_]] {
+  def cut(
+    events: List[UpdateNodeParametersEvent],
+    lastSnapshotContext: GlobalSnapshotInfo,
+    ordinal: SnapshotOrdinal
+  ): F[List[UpdateNodeParametersEvent]]
+}
+
+object UpdateNodeParametersCutter {
```
marcinwadon, line 25: Why it is a separate cutter for `UpdateNodeParameters`? There is already an event cutter used by global consensus and currency consensus. If you want to add "UpdateNodeParameters" here, you should make it in already existing event cutter.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/http/routes/NodeParametersRoutes.scala
```diff
@@ -0,0 +1,66 @@
...
+  private def latestNodeParameters(nodeId: Id): F[Option[(Signed[UpdateNodeParameters], SnapshotOrdinal)]] =
+    for {
+      maybeOrdinal <- snapshotStorage.headSnapshot.map(_.map(_.ordinal))
+      maybeSnapshot <- readLatestFullSnapshot(maybeOrdinal)
+    } yield maybeSnapshot.flatMap(_.value.info.updateNodeParameters.flatMap(_.get(nodeId)))
+
+  protected val public: HttpRoutes[F] = HttpRoutes.of[F] {
+    case req @ POST -> Root =>
+      req
+        .as[Signed[UpdateNodeParameters]]
```
marcinwadon, line 47: @ryle-goehausen should `UpdateNodeParameters` be validated on endpoint (like we do for transactions, allow spends, token locks, etc.?)

effects-ai: Fixed

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/domain/snapshot/programs/UpdateNodeParametersCutter.scala
effects-ai: Because, the cutting logic is different. First, in case of `UpdateNodeParameters` the hard limit is in number items, whereas in the other cases it's bytes. Second, in the former case we use frequency based cutting.

marcinwadon: Ok, I got it, but IMO it should take into account other events too. Because the idea behind event cutter is to keep snapshot size limited. If the original event cutter will choose data up to the limit, all update node parameters would result in snapshot above the size limit.

ryle-goehausen: Original intention behind some of choices in the event cutter was to ensure that no one event type overrides the entire snapshot -- so we need even under saturation conditions for some of these requests to be processed. Idea here is we don't need a very sophisticated mechanism to handle this -- so long as we have a rough requirement that some of these events make it in.

There's no requirement for a heuristic as all these are associated with collateral -- hence slashing in event of spam / ddos.

ryle-goehausen (approved):
rebase required

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/domain/snapshot/programs/UpdateNodeParametersCutter.scala
marcinwadon: Ok, so simply saying it's fine to have data up to the size limit of the snapshot and also additional 100 node parameters updates. If that's what you mean then fine, but it would be great to clearly communicate that with the team, so on review we would realize that it was intended.

ryle-goehausen (approved):

PR #1001, opened 2025-02-05
Return empty Snaphot fee info if unable to deserialize

gclaramunt:
Warn and return empty SnaphotFeeInfo if unable to deserialize. 
A few old global snapshots fail to deserialize totally if the currency snapshot fee can't be deserialized. Instead of throwing an error, we log a warning and return  SnapshotFeesInfo.empty


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -71,7 +71,9 @@ object GlobalSnapshotStateChannelEventsProcessor {
           case hash if hash == Hash.empty => SnapshotFeesInfo.empty.pure // genesis
           case _ =>
             deserialize[Signed[CurrencyIncrementalSnapshot]](event.snapshotBinary).flatMap {
-              case None => new Exception("Could not get snapshot after deserializing").raiseError[F, SnapshotFeesInfo]
+              case None =>
+                logger.warn(s"Could not get snapshot fee info after deserializing event $event, using empty snapshot fees");
+                SnapshotFeesInfo.empty.pure
```
scasplte2, line 76: I can't speak to if this is acceptable behavior but wouldn't you need `logger.warn(...) >> SnapshotFeesInfo.empty.pure[F]` to have the log output to console?

gclaramunt: :facepalm: that got lost in a reformatting

marcinwadon (approved):

PR #1002, opened 2025-02-06
fix: revert changes to SharedConfig via arg in IO app constr

scasplte2:
Demonstrates a method for how we could revert to the previous behavior of `TessellationIOApp` where the shared node config has access to the application-specific level parameters.


marcinwadon (approved):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/app/TessellationIOApp.scala
```diff
@@ -55,7 +54,8 @@ abstract class TessellationIOApp[A <: CliMethod](
   header: String,
   clusterId: ClusterId,
   helpFlag: Boolean = true,
-  version: TessellationVersion = TessellationVersion.unsafeFrom("0.0.1")
+  version: TessellationVersion = TessellationVersion.unsafeFrom("0.0.1"),
+  createConfig: () => IO[ConfigObjectSource]
```
gclaramunt, line 58: I think you can replace it with `createConfig: => IO[ConfigObjectSource]` and (lazily) get the result. Doesn't matter much here because you call it only once.

scasplte2:
Thinking through @gclaramunt suggestion, I didn't care for having to define the app config string in multiple places. Instead, I use the constructor arg to pass the required filenames to the underlying TessellationIOApp, where a function now resides to generate the compiled config source object that should be interpreted as the different types of config files (i.e. NodeShared, AppConfig). 

If there are any strong opinions about not having a `val` in the constructor of the `abstract class` I'm happy to move it. Otherwise, this seemed the most concise approach.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/app/TessellationIOApp.scala
```diff
@@ -45,17 +46,18 @@ import eu.timepit.refined.pureconfig._
...
+import pureconfig.{ConfigReader, ConfigSource}
 
 abstract class TessellationIOApp[A <: CliMethod](
   name: String,
   header: String,
   clusterId: ClusterId,
   helpFlag: Boolean = true,
-  version: TessellationVersion = TessellationVersion.unsafeFrom("0.0.1")
+  version: TessellationVersion = TessellationVersion.unsafeFrom("0.0.1"),
+  val configFiles: List[String]
```
gclaramunt, line 60: I have a slight preference for
```scala
extends CommandIOApp(
      name,
      header,
      helpFlag,
      version.version.value
    ) {

  def configFiles: List[String]
```
and 
```scala
object Main
    extends TessellationIOApp[Run](
      name = "dag-l0",
      header = "Tessellation Node",
      version = TessellationVersion.unsafeFrom(BuildInfo.version),
      clusterId = ClusterId("6d7f1d6a-213a-4148-9d45-d7200f555ecf")
    ) {

  val configFiles = List("dag-l0.conf")
```
but you have to keep in mind initialization order and  probably you can say the same of all the other parameters of TessellationIOApp, so is fine:)

gclaramunt (approved):
LGTM

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/app/TessellationIOApp.scala
marcinwadon: @scasplte2 it's totally up to you

PR #1003, opened 2025-02-07
feat: allow spends eviction

IPadawans:
### Changes
* Added allow-spend eviction logic.
* An allow-spend should be evicted when its epoch progress expires or when a spend transaction is associated with it.
* If evicted due to expiration, a SharedArtifact expiration event should be emitted, and the full balance should be returned to the user.
* If evicted due to a spend transaction, only the remaining unspent balance (the difference between the allow-spend amount and the spend transaction amount) should be returned to the user.

### DIFF
--- a/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensusStateAdvancer.scala
+++ b/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensusStateAdvancer.scala
@@ -15,8 +15,10 @@ import io.constellationnetwork.currency.l0.snapshot.services.StateChannelSnapsho
 import io.constellationnetwork.currency.schema.currency.CurrencySnapshotContext
 import io.constellationnetwork.ext.collection.FoldableOps.pickMajority
 import io.constellationnetwork.ext.crypto._
+import io.constellationnetwork.node.shared.config.types.LastGlobalSnapshotsSyncConfig
 import io.constellationnetwork.node.shared.domain.gossip.Gossip
 import io.constellationnetwork.node.shared.domain.node.NodeStorage
+import io.constellationnetwork.node.shared.domain.snapshot.storage.LastSyncGlobalSnapshotStorage
 import io.constellationnetwork.node.shared.infrastructure.consensus.ConsensusStateUpdater._
 import io.constellationnetwork.node.shared.infrastructure.consensus._
 import io.constellationnetwork.node.shared.infrastructure.consensus.declaration._
@@ -29,7 +31,7 @@ import io.constellationnetwork.node.shared.snapshot.currency._
 import io.constellationnetwork.schema.currencyMessage.fetchStakingAddress
 import io.constellationnetwork.schema.peer.PeerId
 import io.constellationnetwork.security.signature.Signed
-import io.constellationnetwork.security.signature.signature.{Signature, _}
+import io.constellationnetwork.security.signature.signature._
 import io.constellationnetwork.security.{HasherSelector, SecurityProvider}
 import io.constellationnetwork.syntax.sortedCollection._
 
@@ -50,6 +52,7 @@ abstract class CurrencySnapshotConsensusStateAdvancer[F[_]]
 object CurrencySnapshotConsensusStateAdvancer {
 
   def make[F[_]: Async: SecurityProvider: Metrics: HasherSelector](
+    lastGlobalSnapshotsSyncConfig: LastGlobalSnapshotsSyncConfig,
     keyPair: KeyPair,
     consensusStorage: CurrencyConsensusStorage[F],
     consensusFns: CurrencySnapshotConsensusFunctions[F],
@@ -58,7 +61,8 @@ object CurrencySnapshotConsensusStateAdvancer {
     maybeDataApplication: Option[BaseDataApplicationL0Service[F]],
     restartService: RestartService[F, _],
     nodeStorage: NodeStorage[F],
-    leavingDelay: FiniteDuration
+    leavingDelay: FiniteDuration,
+    lastGlobalSnapshotStorage: LastSyncGlobalSnapshotStorage[F]
   ): CurrencySnapshotConsensusStateAdvancer[F] =
     new CurrencySnapshotConsensusStateAdvancer[F] {
       val logger = Slf4jLogger.getLogger[F]
@@ -105,6 +109,9 @@ object CurrencySnapshotConsensusStateAdvancer {
                             for {
                               peerEvents <- consensusStorage.pullEvents(bound)
                               events = peerEvents.toList.flatMap(_._2).map(_._2).toSet
+                              lastNGlobalSnapshotsSynchronized <- lastGlobalSnapshotStorage.getLastNSynchronized(
+                                lastGlobalSnapshotsSyncConfig.minGlobalSnapshotsToParticipateConsensus.value
+                              )
                               (artifact, context, returnedEvents) <- consensusFns
                                 .createProposalArtifact(
                                   state.key,
@@ -113,7 +120,8 @@ object CurrencySnapshotConsensusStateAdvancer {
                                   hasher,
                                   majorityTrigger,
                                   events,
-                                  state.facilitators.value.toSet
+                                  state.facilitators.value.toSet,
+                                  lastNGlobalSnapshotsSynchronized
                                 )
                               returnedPeerEvents = peerEvents.map {
                                 case (peerId, events) =>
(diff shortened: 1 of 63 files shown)


modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotTraverse.scala
```diff
@@ -110,7 +110,9 @@ object GlobalSnapshotTraverse {
             case ((lastCtx, lastInc), hash) =>
               loadIncOrErr(hash).flatMap { inc =>
                 HasherSelector[F].forOrdinal(inc.ordinal) { implicit hasher =>
-                  contextFns.createContext(lastCtx, lastInc, inc).map(_ -> inc)
+                  lastInc.toHashed.flatMap { lastIncHashed =>
```
scasplte2, line 113: could we avoid the Hash operation here by querying `inc.lastSnapshotHash`?

modules/dag-l1/src/main/scala/io/constellationnetwork/dag/l1/domain/snapshot/programs/DAGSnapshotProcessor.scala
```diff
@@ -36,29 +39,42 @@ object DAGSnapshotProcessor {
...
-              transactionStorage,
-              lastGlobalSnapshotStorage,
-              addressStorage
-            )
-          )
+      )(implicit hasher: Hasher[F]): F[SnapshotProcessingResult] = {
+        val snapshotOrdinal = snapshot match {
+          case Left(value) =>
+            val (snapshot, _) = value
+            snapshot.ordinal
```
scasplte2, line 46: this could be simplified to `case Left((snapshot, _)) => snapshot.ordinal`

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/config/types.scala
```diff
@@ -155,4 +157,6 @@ object types {
   case class AllowSpendsConfig(lastValidEpochProgress: MinMax)
 
   case class TokenLocksConfig(minEpochProgressesToLock: NonNegLong)
+
+  case class LastGlobalSnapshotsSyncConfig(minGlobalSnapshotsToParticipateConsensus: PosLong)
```
scasplte2, line 161: You are casting this to an `Int` everyplace this is called so perhaps it should be a `PosInt` or maybe even a smaller range like `PosByte` which would limit from 1 -> 127.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/snapshot/storage/LastNGlobalSnapshotStorage.scala
```diff
@@ -0,0 +1,8 @@
+package io.constellationnetwork.node.shared.domain.snapshot.storage
+
+import io.constellationnetwork.schema.{GlobalIncrementalSnapshot, GlobalSnapshotInfo, SnapshotOrdinal}
+import io.constellationnetwork.security.Hashed
+
+trait LastNGlobalSnapshotStorage[F[_]] extends LastSnapshotStorage[F, GlobalIncrementalSnapshot, GlobalSnapshotInfo] {
```
scasplte2, line 6: This seems like what I had in mind when I mentioned adding something to replace `LastSnapshotStorage[F, GlobalIncrementalSnapshot, GlobalSnapshotInfo] with LastSynchronizedGlobalSnapshotStorage[F]`

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
```diff
@@ -179,6 +186,46 @@ object CurrencySnapshotAcceptanceManager {
         facilitators
       )
 
+      offset = NonNegLong(2L)
```
scasplte2, line 189: can we move this magic number to a config?

scasplte2 (commented):
Overall it seems reasonable to me. I noticed that this is calling `lastGlobalSnapshotsSyncConfig.minGlobalSnapshotsToParticipateConsensus.value.toInt` quite a number of times so I suggested embedding the value into the creation of a synchronized storage object but there are several different interfaces for snapshot storage that are being used so this may not be possible.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotTraverse.scala
IPadawans: Not actually because we need the snapahot hashed, not just the hash

modules/dag-l1/src/main/scala/io/constellationnetwork/dag/l1/domain/snapshot/programs/DAGSnapshotProcessor.scala
IPadawans: 👍🏼

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/config/types.scala
IPadawans: 👍🏼

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotTraverse.scala
scasplte2: but `inc.lastSnapshotHash == lastIncHashed` no? Perhaps I am misunderstanding the difference of "the snapahot hashed, not just the hash"

IPadawans: It's because the lastSnapshotHash is a string, and the lastIncHashed is Hashed[GlobalIncrementalSnapshot]

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensusStateAdvancer.scala
```diff
@@ -58,7 +63,9 @@ object CurrencySnapshotConsensusStateAdvancer {
     maybeDataApplication: Option[BaseDataApplicationL0Service[F]],
     restartService: RestartService[F, _],
     nodeStorage: NodeStorage[F],
-    leavingDelay: FiniteDuration
+    leavingDelay: FiniteDuration,
+    lastGlobalSnapshotStorage: LastSnapshotStorage[F, GlobalIncrementalSnapshot, GlobalSnapshotInfo]
+      with LastSynchronizedGlobalSnapshotStorage[F]
```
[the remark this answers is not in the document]
IPadawans: 👍🏼

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/programs/Download.scala
```diff
@@ -42,14 +45,17 @@ import retry._
...
     clusterStorage: ClusterStorage[F],
     currencySnapshotContextFns: CurrencySnapshotContextFunctions[F],
     nodeStorage: NodeStorage[F],
     consensus: CurrencySnapshotConsensus[F],
     peerSelect: PeerSelect[F],
     identifierStorage: IdentifierStorage[F],
-    maybeDataApplication: Option[BaseDataApplicationL0Service[F]]
+    maybeDataApplication: Option[BaseDataApplicationL0Service[F]],
+    lastGlobalSnapshotStorage: LastSnapshotStorage[F, GlobalIncrementalSnapshot, GlobalSnapshotInfo]
+      with LastSynchronizedGlobalSnapshotStorage[F]
```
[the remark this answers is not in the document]
IPadawans: 👍🏼

IPadawans:
> Overall it seems reasonable to me. I noticed that this is calling `lastGlobalSnapshotsSyncConfig.minGlobalSnapshotsToParticipateConsensus.value.toInt` quite a number of times so I suggested embedding the value into the creation of a synchronized storage object but there are several different interfaces for snapshot storage that are being used so this may not be possible.

Yes, this is called in several different interfaces, so I think it will be hard to put in the creation of a synchronized storage object

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
IPadawans: :+1

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/snapshot/storage/LastNGlobalSnapshotStorage.scala
IPadawans: :+1

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotTraverse.scala
scasplte2: oh, I see 🤦‍♂️  `Hashed[T] != Hash` My mistake, thanks for clarifying

scasplte2 (approved):
looks good to me

ryle-goehausen (approved):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/LastSyncGlobalSnapshotStorage.scala
```diff
@@ -93,9 +94,47 @@ object LastSynchronizedGlobalSnapshotStorage {
...
+        currencySnapshotStorage.head.flatMap {
+          case Some((_, info)) =>
+            info.globalSnapshotSyncView.flatTraverse { syncView =>
+              val lastOrdinalOpt = syncView.values
+                .map(_.globalSnapshotOrdinal)
+                .groupBy(identity)
+                .toList
+                .sortBy { case (ordinal, occurrences) => (-occurrences.size, ordinal.value.value) }
+                .map { case (ordinal, _) => ordinal }
+                .maxOption
```
kpudlik, line 109: You can shorten it a little bit by `maxByOption`:

```
.maxByOption { case (ordinal, occurrences) => (occurrences.size, -ordinal.value.value) }
.map { case (ordinal, _) => ordinal }
```

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/LastSyncGlobalSnapshotStorage.scala
```diff
@@ -93,9 +94,47 @@ object LastSynchronizedGlobalSnapshotStorage {
...
+                  }.toSet.toList
+
+                  selectedOrdinals.traverse(get).map { results =>
+                    val snapshots = results.flatten
+                    if (snapshots.isEmpty) None else Some(snapshots)
+                  }
+
+                case None =>
+                  Async[F].pure(None)
+              }
```
kpudlik, line 124: You can just `flatTraverse` and thanks to that you do not need to explicitly return this `None` in lines 122-123:

```scala
              lastOrdinalOpt.flatTraverse { lastOrdinal =>
                val selectedOrdinals = (0 until n).flatMap { i =>
                  NonNegLong.from(lastOrdinal.value.value - i.toLong).toOption.flatMap(SnapshotOrdinal(_))
                }.toSet.toList

                selectedOrdinals
                  .traverse(get)
                  .map { results =>
                    val snapshots = results.flatten
                    if (snapshots.isEmpty) None else Some(snapshots)
                  }
              }
```

PR #1004, opened 2025-02-10
feat: price oracle data update

effects-ai:


ryle-goehausen:
@effects-ai might want to rebase this again so we can review new diff from dev merge

marcinwadon (changes requested):
As we spoke on the meeting, this change should be a part of metagraph, not tessellation codebase.

ryle-goehausen:
>As we spoke on the meeting, this change should be a part of metagraph, not tessellation codebase.

regarding this -- part of the schema must be merged to GL0, similar to AllowSpend types -- any of the component schemas related to the INTERNAL AMM types can be proposed here but should NOT be merged. We only need to merge one new type. This distinction should be equivalent to the AllowSpend / SpendTransaction vs. internal AMM Order Data Update types.

modules/shared/src/main/scala/io/constellationnetwork/schema/priceOracle.scala
```diff
@@ -0,0 +1,18 @@
...
+
+import derevo.cats.{order, show}
+import derevo.circe.magnolia.{decoder, encoder}
+import derevo.derive
+import eu.timepit.refined.types.numeric.NonNegLong
+
+object priceOracle {
+
+  @derive(decoder, encoder, order, show)
+  case class PricingUpdate(currencyId: CurrencyId, fractionalValue: NonNegLong, markOverride: Boolean, lastReference: TransactionReference)
```
ryle-goehausen, line 15: This update should be in the AMM itself through the Daemon -- we need the other type committed here from ticket, which should represent the GL0 type (which does not have references or overrides, it's purely an oracle price update (the other declared type from doc.)

ryle-goehausen (changes requested):
This case class should go in the new repo, the other type belongs here not this one.

ryle-goehausen:
This PR can be left for now -- blocked by other tasks. Will close and reference on existing ticket for later work

PR #1008, opened 2025-02-13
feat: adding brotli serialization for fee transactions

IPadawans:
### Changes
+ Changing the serializer of feeTxn to use the same as AllowSpends, TokenLocks, and CurrencyTxn. All these transactions use brotli, so I've added brotli to feeTxn as well, so we can have all in the same pattern


marcinwadon (dismissed):
image

Isn't `JsonBinarySerializer` the brotli one?

PR #1009, opened 2025-02-14
feat: adding last n global snapshot sync to dag-l1

IPadawans:
### Changes
+ Adding last n global snapshot sync to dag-l1

### DIFF
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/LastNGlobalSnapshotStorage.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/LastNGlobalSnapshotStorage.scala
@@ -32,9 +32,15 @@ object LastNGlobalSnapshotStorage {
       def set(snapshot: Hashed[GlobalIncrementalSnapshot], state: GlobalSnapshotInfo): F[Unit] =
         snapshotsR.modify { snapshots =>
           snapshots.lastOption match {
-            case Some((_, (latest, _))) if isNextSnapshot(latest, snapshot.signed.value) =>
+            case Some((_, (latest, _))) =>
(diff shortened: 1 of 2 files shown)


marcinwadon (dismissed):

marcinwadon:
I'm not really sure if it's correct but I approve since I need this change

IPadawans:
> I'm not really sure if it's correct but I approve since I need this change

Sure, if you find any issues during your tests, let me know so we can fix. It seems to be working at the first look, but who knows the other scenarios

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/LastNGlobalSnapshotStorage.scala
```diff
@@ -34,7 +34,7 @@ object LastNGlobalSnapshotStorage {
           snapshots.lastOption match {
             case Some((_, (latest, _))) if isNextSnapshot(latest, snapshot.signed.value) =>
               (snapshots.updated(snapshot.ordinal, (snapshot, state)), Applicative[F].unit)
-            case _ => (snapshots, MonadThrow[F].raiseError[Unit](new Throwable("Failure during putting new global snapshot!")))
+            case _ => (snapshots.updated(snapshot.ordinal, (snapshot, state)), Applicative[F].unit)
```
marcinwadon, line 37: Does it mean that we ignore the chain correctness right now? This check was to avoid storing forked snapshots.

IPadawans: Hmmm that's the reason then, then for clarifying, I'll rollback this part to ensure the correctness

PR #1010, opened 2025-02-14
feat: fixing dag-l1 last n snapshot sync

IPadawans:


marcinwadon (changes requested):
@IPadawans please fix the merge and IMO it would be good to go

PR #1017, opened 2025-02-18
feat: delegated stake transaction process

effects-ai:


modules/shared/src/main/scala/io/constellationnetwork/schema/stake.scala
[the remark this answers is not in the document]
effects-ai: I renamed it to `delegatedStake`.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/domain/cell/AlgebraCommand.scala
```diff
@@ -12,5 +14,7 @@ object AlgebraCommand {
   case class EnqueueStateChannelSnapshot(snapshot: StateChannelOutput) extends AlgebraCommand
   case class EnqueueDAGL1Data(data: Signed[Block]) extends AlgebraCommand
   case class EnqueueUpdateNodeParameters(data: Signed[UpdateNodeParameters]) extends AlgebraCommand
+  case class EnqueueDelegatedStake(data: DelegatedStakeOutput) extends AlgebraCommand
+  case class EnqueueNodeCollateral(data: NodeCollateralOutput) extends AlgebraCommand
```
ryle-goehausen, line 18: It might be a lot easier to have both of these types extend some new direct type (similar to the way we treat the data artifact additions in the ML0 for the changes for SpendTransaction / AllowSpend,

so rather than dealing with this queue in multiple places, you could have a single new type here, indicating a direct GL0 embedded artifact. That way, we don't need to modify any of these functions again in the future.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/domain/cell/CoalgebraCommand.scala
```diff
@@ -11,4 +13,6 @@ object CoalgebraCommand {
   case class ProcessDAGL1(data: Signed[Block]) extends CoalgebraCommand
   case class ProcessStateChannelSnapshot(snapshot: StateChannelOutput) extends CoalgebraCommand
   case class ProcessUpdateNodeParameters(updateNodeParameters: Signed[UpdateNodeParameters]) extends CoalgebraCommand
+  case class ProcessDelegatedStake(createDelegatedStake: DelegatedStakeOutput) extends CoalgebraCommand
```
ryle-goehausen, line 16: Same issue as above, this should be a unified type since we're processing them all the same way

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/domain/cell/L0Cell.scala
```diff
@@ -16,24 +22,23 @@ import io.constellationnetwork.statechannel.StateChannelOutput
...
-import L0Cell.{Algebra, Coalgebra}
-import L0CellInput._
-
 sealed trait L0CellInput
 
 object L0CellInput {
   case class HandleDAGL1(data: Signed[Block]) extends L0CellInput
   case class HandleStateChannelSnapshot(snapshot: StateChannelOutput) extends L0CellInput
   case class HandleUpdateNodeParameters(updateNodeParameters: Signed[UpdateNodeParameters]) extends L0CellInput
+  case class HandleDelegatedStake(delegatedStake: DelegatedStakeOutput) extends L0CellInput
```
ryle-goehausen, line 31: Same here

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/domain/cell/L0Cell.scala
```diff
@@ -16,24 +22,23 @@ import io.constellationnetwork.statechannel.StateChannelOutput
...
 }
 
 class L0Cell[F[_]: Async](
   data: L0CellInput,
   l1OutputQueue: Queue[F, Signed[Block]],
   stateChannelOutputQueue: Queue[F, StateChannelOutput],
-  updateNodeParametersQueue: Queue[F, Signed[UpdateNodeParameters]]
+  updateNodeParametersQueue: Queue[F, Signed[UpdateNodeParameters]],
+  delegatedStakeOutputQueue: Queue[F, DelegatedStakeOutput],
+  nodeCollateralOutputQueue: Queue[F, NodeCollateralOutput]
```
ryle-goehausen, line 41: Same here

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/domain/cell/L0Cell.scala
```diff
@@ -47,6 +52,8 @@ class L0Cell[F[_]: Async](
                 Algebra.enqueueDAGL1Data(l1OutputQueue)(data)
               case EnqueueUpdateNodeParameters(updateNodeParameters) =>
                 Algebra.enqueueUpdateNodeParameters(updateNodeParametersQueue)(updateNodeParameters)
+              case EnqueueDelegatedStake(data) => Algebra.enqueueDelegatedStake(delegatedStakeOutputQueue)(data)
+              case EnqueueNodeCollateral(data) => Algebra.enqueueNodeCollateral(nodeCollateralOutputQueue)(data)
```
ryle-goehausen, line 56: Same here (will skip references after this that apply to same issue, but its repeated.)

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/domain/delegatedStake.scala
```diff
@@ -0,0 +1,12 @@
+package io.constellationnetwork.dag.l0.domain
+
+import io.constellationnetwork.schema.delegatedStake.UpdateDelegatedStake
+import io.constellationnetwork.security.signature.Signed
+
+object delegatedStake {
+
+  sealed trait DelegatedStakeOutput
```
ryle-goehausen, line 8: This is maybe not the best naming here, although I do appreciate what you're trying to do by wrapping these Signed[T] types in an event class.

Since these are essentially wrappers around the submit endpoints, maybe Request or Event is a better name? I think output is a little confusing here.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/domain/nodeCollateral.scala
```diff
@@ -0,0 +1,12 @@
+package io.constellationnetwork.dag.l0.domain
+
+import io.constellationnetwork.schema.nodeCollateral.UpdateNodeCollateral
+import io.constellationnetwork.security.signature.Signed
+
+object nodeCollateral {
+
+  sealed trait NodeCollateralOutput
```
ryle-goehausen, line 8: Same, this needs a better name, Request or Event would be more appropriate here since it represents a user submitted value, not a calculated value.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/http/routes/DelegatedStakesRoutes.scala
```diff
@@ -0,0 +1,121 @@
...
+) extends Http4sDsl[F]
+    with PublicRoutes[F] {
+
+  import org.http4s.circe.CirceEntityCodec.circeEntityDecoder
+
+  private val logger = Slf4jLogger.getLoggerFromName[F]("DelegatedStakesLogger")
+
+  protected val prefixPath: InternalUrlPrefix = "/"
+
+  private def readLatestFullSnapshot(maybeOrdinal: Option[SnapshotOrdinal]): F[Option[Signed[GlobalSnapshot]]] =
```
ryle-goehausen, line 45: Both this and the below function are copy/pasted between NodeParams / NodeCollateral routes and this, please abstract any common functionality here.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/http/routes/DelegatedStakesRoutes.scala
```diff
@@ -0,0 +1,121 @@
...
+    maybeOrdinal.traverse(fullGlobalSnapshotStorage.read).map(_.flatten)
+
+  private def getLatestFullSnapshot: F[Option[Signed[GlobalSnapshot]]] =
+    for {
+      maybeOrdinal <- snapshotStorage.headSnapshot.map(_.map(_.ordinal))
+      maybeSnapshot <- readLatestFullSnapshot(maybeOrdinal)
+    } yield maybeSnapshot
+
+  @derive(eqv, show, encoder)
+  case class DelegatedStakesInfo(
```
ryle-goehausen, line 55: It might make more sense to locate this alongside your other schemas (not directly embedded in the class,) in case it is ever used elsewhere, I realize that's unlikely for now, but it's easier for someone trying to find all the schemas related to this feature.

Additionally, for all the input / output types, please implement the model yml, here's an example api doc update PR for reference which includes new type definitions: https://github.com/Constellation-Labs/tessellation/pull/1034

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/delegatedStake/UpdateDelegatedStakeValidator.scala
```diff
@@ -0,0 +1,198 @@
...
+        } else {
+          TooManySignatures(signed.proofs).invalidNec
+        }
+        result.pure[F]
+      }
+
+      private def validateNodeId[A <: UpdateDelegatedStake](
+        signed: Signed[A]
+      ): UpdateDelegatedStakeValidationErrorOr[Signed[A]] =
+        if (seedlist.forall(_.exists(_.peerId == signed.nodeId))) {
```
ryle-goehausen, line 89: Nitpick comment: I realize elsewhere in the code we're following this pattern as treating seedlist as an Option type and enforcing it elsewhere, but it would be much better if the tests mocked up a seedlist based on the generators, rather than allowing this to be optional everywhere, since it's essentially always required.

It's not required for you to change this, but if you can see a place to introduce a pattern it would be helpful.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/http/routes/DelegatedStakesRoutes.scala
```diff
@@ -0,0 +1,121 @@
...
+            .as[Signed[UpdateDelegatedStake.Create]]
+            .flatMap(signed => validator.validateCreateDelegatedStake(signed, lastSnapshot.info, lastSnapshot.epochProgress))
+            .flatTap {
+              case Valid(signed) =>
+                logger.info(s"Accepted create delegated stake from ${signed.proofs.map(_.id).map(PeerId.fromId(_))}")
+              case Invalid(errors) =>
+                logger.warn(s"Invalid create delegated stake: $errors")
+            }
+            .flatMap {
+              case Valid(signed)   => mkCell(CreateDelegatedStakeOutput(signed)).run()
```
ryle-goehausen, line 75: Nitpick: this is potentially out of scope of the PR, but anywhere you can bypass this mkCell and just directly enqueue, all the better. We're not using the rest of that functionality anyways

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/http/routes/NodeCollateralRoutes.scala
```diff
@@ -0,0 +1,121 @@
...
+) extends Http4sDsl[F]
+    with PublicRoutes[F] {
+
+  import org.http4s.circe.CirceEntityCodec.circeEntityDecoder
+
+  private val logger = Slf4jLogger.getLoggerFromName[F]("NodeCollateralsLogger")
+
+  protected val prefixPath: InternalUrlPrefix = "/"
+
+  private def readLatestFullSnapshot(maybeOrdinal: Option[SnapshotOrdinal]): F[Option[Signed[GlobalSnapshot]]] =
```
ryle-goehausen, line 45: Duplicate code

ryle-goehausen: Also, these are already on dev, please update this PR so this delta doesn't show

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotConsensusFunctions.scala
```diff
@@ -69,7 +72,25 @@ object GlobalSnapshotConsensusFunctions {
       val tokenLockEvents = artifact.tokenLockBlocks.map(_.toList.map(TokenLockEvent(_))).getOrElse(List.empty)
       val unpEvents =
         artifact.updateNodeParameters.getOrElse(SortedMap.empty[Id, Signed[UpdateNodeParameters]]).values.map(UpdateNodeParametersEvent(_))
-      val events: Set[GlobalSnapshotEvent] = dagEvents ++ scEvents ++ allowSpendEvents ++ unpEvents ++ tokenLockEvents
+      val cdsEvents = artifact.activeDelegatedStakes
```
ryle-goehausen, line 75: This is a perfect example of why we should have an abstracted artifact/event type here for all of these new types, so we can avoid this.

modules/node-shared/src/main/resources/application.conf
```diff
@@ -146,4 +146,10 @@ validation-error-storage {
 delegated-staking {
   min-reward-fraction:  5000000
   max-reward-fraction: 10000000
+  # (10 years * 365.25 days * 24 hours * 60 minutes * 60 seconds) / 43 seconds per epoch
+  # in epochs
+  min-epoch-progresses-to-lock: 7338977
+  # 21 days * 24 hours * 60 minutes * 4.5 snapshots per minute
+  # in snapshot ordinals
+  withdrawal-time-limit = 136080
```
ryle-goehausen, line 154: Since we'll need another configuration specific to testing (haven't reviewed full PR so if this has already been done please ignore this comment.) But -- we need another conf / or some way to introduce acceptable parameters for testing this, i.e. a very small amount.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/delegatedStake/UpdateDelegatedStakeValidator.scala
```diff
@@ -0,0 +1,198 @@
...
+  )(implicit hasher: Hasher[F]): UpdateDelegatedStakeValidator[F] =
+    new UpdateDelegatedStakeValidator[F] {
+
+      def validateCreateDelegatedStake(
+        signed: Signed[UpdateDelegatedStake.Create],
+        lastContext: GlobalSnapshotInfo,
+        lastGlobalEpochProgress: EpochProgress
+      ): F[UpdateDelegatedStakeValidationErrorOr[Signed[UpdateDelegatedStake.Create]]] =
+        for {
+          numberOfSignaturesV <- validateNumberOfSignatures(signed)
```
ryle-goehausen, line 54: Nitpick: not required to change, but most of the functionality in these validators is shared, any way we can abstract here?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/nodeCollateral/UpdateNodeCollateralValidator.scala
```diff
@@ -0,0 +1,203 @@
...
+  ): UpdateNodeCollateralValidator[F] =
+    new UpdateNodeCollateralValidator[F] {
+
+      def validateCreateNodeCollateral(
+        signed: Signed[UpdateNodeCollateral.Create],
+        lastContext: GlobalSnapshotInfo,
+        lastGlobalEpochProgress: EpochProgress
+      ): F[UpdateNodeCollateralValidationErrorOr[Signed[UpdateNodeCollateral.Create]]] =
+        for {
+          numberOfSignaturesV <- validateNumberOfSignatures(signed)
```
ryle-goehausen, line 58: Same problem here, we have almost identical validator functions duplicated here.

ryle-goehausen (changes requested):

ryle-goehausen:
@effects-ai please merge or rebase the dev changes here for a smaller diff.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/delegatedStake/UpdateDelegatedStakeValidator.scala
```diff
@@ -0,0 +1,198 @@
...
+            .getOrElse(address, SortedSet.empty[Signed[TokenLock]])
+          for {
+            tokenLocksWithReferences <- tokenLocks.toList.traverse(t => TokenLockReference.of(t).map(r => (t, r)))
+          } yield
+            tokenLocksWithReferences.find { case (_, r) => r === signed.tokenLockReference } match {
+              case Some((tokenLock, _)) =>
+                if (signed.stake.amount.value.value > tokenLock.amount.value.value) {
+                  false
+                } else {
+                  val validUnlockEpochProgress =
```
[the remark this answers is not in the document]
effects-ai: TokenLock.unlockEpoch is not optional. How can we can tell it has no expiration? 

Here I check if it unlocks in at least in 10 years (configurable).

ryle-goehausen (changes requested):
Please fix API doc related issues first, then can cover rest of functionality in separate PR

modules/shared/src/main/scala/io/constellationnetwork/schema/nodeCollateral.scala
```diff
@@ -0,0 +1,83 @@
...
+  sealed trait UpdateNodeCollateral {
+    def nodeId: PeerId
+  }
+  case object UpdateNodeCollateral {
+    @derive(eqv, show, encoder, decoder)
+    case class Create(collateral: NodeCollateral, tokenLockRef: Hash) extends UpdateNodeCollateral {
+      def nodeId: PeerId = collateral.nodeId
+    }
+    @derive(eqv, show, encoder, decoder)
+    case class Withdraw(collateral: NodeCollateral, collateralRef: Hash) extends UpdateNodeCollateral {
```
ryle-goehausen, line 62: remove this collateral field

ryle-goehausen (changes requested):
@derive(eqv, show, encoder, decoder)
    case class Create(stake: DelegatedStake, tokenLockRef: Hash) extends UpdateDelegatedStake {
      def nodeId: PeerId = stake.nodeId
    }
    
    
    Realized there's also something missing here. Since we need to add overwrite on create functionality for changing the nodeId, we're missing this from ticket:
    
>For this type, we need a new functionality added, which is ‘overwrite’ with a new nodeId. This should ‘change’ the nodeId of an existing active delegated stake.

>In order for this new ‘overwrite’ to work properly, all the amounts / token lock refs must remain identical, only the nodeId can change, and importantly, we need a new field which is a DelegatedStakeLastRef(hash,ordinal)

>We need an endpoint added to get the lastRef for this create also now, to prevent replay.

Please update the API docs and schema to support a lastRef field for these Create types. (This is not the same as the token reference.)

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/http/routes/DelegatedStakesRoutes.scala
```diff
@@ -0,0 +1,143 @@
...
+  nodeStorage: NodeStorage[F],
+  withdrawalTimeLimit: NonNegLong
+) extends Http4sDsl[F]
+    with PublicRoutes[F] {
+
+  import org.http4s.circe.CirceEntityCodec.circeEntityDecoder
+
+  private val logger = Slf4jLogger.getLoggerFromName[F]("DelegatedStakesLogger")
+
+  protected val prefixPath: InternalUrlPrefix = "/"
```
marcinwadon, line 45: you repeat "delegated-stakes" in each route. it should be the prefix here

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/http/routes/NodeCollateralRoutes.scala
```diff
@@ -0,0 +1,145 @@
...
+  nodeStorage: NodeStorage[F],
+  withdrawalTimeLimit: NonNegLong
+) extends Http4sDsl[F]
+    with PublicRoutes[F] {
+
+  import org.http4s.circe.CirceEntityCodec.circeEntityDecoder
+
+  private val logger = Slf4jLogger.getLoggerFromName[F]("NodeCollateralsLogger")
+
+  protected val prefixPath: InternalUrlPrefix = "/"
```
marcinwadon, line 46: as above, node-collateral should be a prefix here

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/delegatedStake/UpdateDelegatedStakeAcceptanceManager.scala
```diff
@@ -0,0 +1,84 @@
...
+        creates: List[Signed[UpdateDelegatedStake.Create]],
+        withdrawals: List[Signed[UpdateDelegatedStake.Withdraw]],
+        lastSnapshotContext: GlobalSnapshotInfo,
+        lastGlobalEpochProgress: EpochProgress,
+        lastSnapshotOrdinal: SnapshotOrdinal
+      ): F[UpdateDelegatedStakeAcceptanceResult] = {
+
+        def partitionAccepted[A](validated: List[UpdateDelegatedStakeValidationErrorOr[A]], signed: List[A]) =
+          validated
+            .zip(signed)
```
marcinwadon, line 46: `Validated` monad may contain the the type that is being verified. here is that `[A]`. I think zipping like this is error prone, as you would have to ensure that both lists are equally the same length and both are ordered. instead, you can the the value from validated monad. so, having the list of validated elements, you should partition them by valid/invalid at that's it

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/delegatedStake/UpdateDelegatedStakeAcceptanceResult.scala
```diff
@@ -0,0 +1,18 @@
...
+
+import io.constellationnetwork.node.shared.domain.delegatedStake.UpdateDelegatedStakeValidator.UpdateDelegatedStakeValidationError
+import io.constellationnetwork.schema.SnapshotOrdinal
+import io.constellationnetwork.schema.address.Address
+import io.constellationnetwork.schema.delegatedStake.UpdateDelegatedStake
+import io.constellationnetwork.security.signature.Signed
+
+case class UpdateDelegatedStakeAcceptanceResult(
+  acceptedCreates: SortedMap[Address, List[(Signed[UpdateDelegatedStake.Create], SnapshotOrdinal)]],
+  notAcceptedCreates: List[(Signed[UpdateDelegatedStake.Create], NonEmptyChain[UpdateDelegatedStakeValidationError])],
```
marcinwadon, line 15: as above, it should not be a tuple but a list of validated elements (all invalid in that case). so you can take the value from it or validation errors.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
```diff
@@ -249,7 +249,7 @@ object CurrencySnapshotAcceptanceManager {
       )
 
       acceptedTokenLocks = incomingTokenLocks
-        .filter(itl => itl.unlockEpoch >= epochProgress)
+        .filter(itl => itl.unlockEpoch.forall(_ >= epochProgress))
```
marcinwadon, line 252: `forall` returns `true` for `None`. is that correct?

doc/api/models/schemas/CreateNodeCollateral.yml
```diff
@@ -0,0 +1,8 @@
+type: object
+properties:
+  stake:
+    $ref: 'NodeCollateral.yml'
+  tokenLockRef:
+    $ref: 'Hash.yml'
+  collateralRef:
```
ryle-goehausen, line 7: This reference actually needs an ordinal -- the tokenLockRef doesn't, but there needs to be a unique type with a hash/ordinal for this one unfortunately

doc/api/models/schemas/CreateDelegatedStake.yml
```diff
@@ -0,0 +1,8 @@
+type: object
+properties:
+  stake:
+    $ref: 'DelegatedStake.yml'
+  tokenLockRef:
+    $ref: 'Hash.yml'
+  stakeRef:
```
ryle-goehausen, line 7: Same here, needs a new type with an ordinal -- not the same as token lock here.

ryle-goehausen (changes requested):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
ryle-goehausen: I think this makes sense, unlock is being validated only if there's an unlock provided, otherwise it has no criteria.

ryle-goehausen (approved):
Remainder of changes in next PR

PR #1019, opened 2025-02-25
fix: fixing last epoch progress on allow spend and token lock services

IPadawans:


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/swap/ContextualAllowSpendValidator.scala
```diff
@@ -108,9 +110,17 @@ object ContextualAllowSpendValidator {
...
           allowSpend.asRight
         } else
-          TooFarLastValidEpochProgress(allowSpend.lastValidEpochProgress).asLeft
+          TooFarLastValidEpochProgress(allowSpend.lastValidEpochProgress, currentEpochProgress).asLeft
       }
 
+      private def validateCurrencyId(
+        allowSpend: Hashed[AllowSpend]
+      ): Either[ContextualAllowSpendValidationError, Hashed[AllowSpend]] =
+        if (allowSpend.currency.exists(currencyId.contains(_)) || allowSpend.currency.isEmpty) {
```
marcinwadon, line 119: it should be determined by the layer. if it's on DAG, then only empty is allowed. if it's on currency, then only currency is allowed 🤔

modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala
```diff
@@ -42,7 +42,7 @@ object swap {
 
   @derive(decoder, encoder, order, show)
   @newtype
-  case class AllowSpendFee(value: PosLong)
+  case class AllowSpendFee(value: NonNegLong)
```
marcinwadon, line 45: is it a fix for last epoch progress on allow spend and token lock services? if not, it should be a separate commit 🤔

IPadawans: This is a fix for the testing on tesntet, we need to test with 0 fee, so as a simple solution I've put together in this PR

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/swap/ContextualAllowSpendValidator.scala
IPadawans: So if the provided currencyId is none, it means that we're on DAG layer and we can accept none, otherwise, we should validate? right?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/swap/ContextualAllowSpendValidator.scala
```diff
@@ -108,9 +110,15 @@ object ContextualAllowSpendValidator {
...
           allowSpend.asRight
         } else
-          TooFarLastValidEpochProgress(allowSpend.lastValidEpochProgress).asLeft
+          TooFarLastValidEpochProgress(allowSpend.lastValidEpochProgress, currentEpochProgress).asLeft
       }
 
+      private def validateCurrencyId(
+        allowSpend: Hashed[AllowSpend]
+      ): Either[ContextualAllowSpendValidationError, Hashed[AllowSpend]] =
+        if (allowSpend.currency.forall(currencyId.contains)) allowSpend.asRight
```
marcinwadon, line 119: when `currencyId` is `Some` and `allowSpend.currency` is `None`, it returns `true` but should not 🤔 
image

IPadawans: Thanks for pointing this, 👍🏼

marcinwadon: cool :+1:
to be more verbose, it could be sth like this:

```
val predicate = currencyId match {
  case Some(id) => allowSpend.currency.contains(id)
  case None => allowSpend.currency.isEmpty
}

if (predicate) valid else invalid
```

marcinwadon (approved):

PR #1021, opened 2025-02-25
feat: adding a flag to skip artifact validation in currency incremental snapshots

IPadawans:
### Changes
+ Adding a flag to disable the validation of artifacts on `createContext` function, used by BlockExplorer

### DIFF
--- a/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/modules/Services.scala
+++ b/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/modules/Services.scala
@@ -102,6 +102,7 @@ object Services {
       )
 
       validator = CurrencySnapshotValidator.make[F](
+        sharedCfg.environment,
         creator,
         signedValidator,
         maybeRewards,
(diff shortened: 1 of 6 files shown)


marcinwadon:
Where do you pass true? I didn't spot it. Also, what's the purpose of this? If we want to skip some validations, we should turn off a single validator, not whole validation. If skip flag would be set to true, we basically won't check the correctness of chain and it's insecure to rely on such data. TBH I'm strongly against

IPadawans:
@marcinwadon this will be true only on the block explorer side, not on tessellation. The purpose is due to the issue we had today. We can discuss disabling a few validations and how to proceed, but we discussed this in today's meeting.
cc: @AlexBrandes

IPadawans:
@marcinwadon Do you think we should skip only the validations of allowSpends, balances, and tokenLocks?

IPadawans:
@marcinwadon I'm also not understanding this part:
```
 If skip flag would be set to true, we basically won't check the correctness of chain and it's insecure to rely on such data
 ```
 if the flag is set to true we're skipping only the validation of the artifact, no? I mean the fields of the artifact

marcinwadon:
@IPadawans could you summarize the meeting discussion? It's hard to discuss it then without context. Anyway, snapshot streaming should reflect the network validation. If you want to skip something, you should skip it on the network too. Otherwise, the snapshot streaming may desync with network but since the block explorer is a single centralized source of truth, it's really important to keep it in sync. Anyway, I strongly believe that this PR is just a workaround and the issue should be resolved differently, by making a proper validation in the snapshot streaming.

marcinwadon:
> @marcinwadon I'm also not understanding this part:
> 
> ```
> 
>  If skip flag would be set to true, we basically won't check the correctness of chain and it's insecure to rely on such data
> 
>  ```
> 
>  if the flag is set to true we're skipping only the validation of the artifact, no? I mean the fields of the artifact

And "artifact" there is a whole snapshot.

ryle-goehausen (dismissed):
This is the non-hybrid solution which will allow us to proceed with testnet development while we discuss tradeoffs of further solutions. Need CI passing though,

IPadawans:
The CI tests are passing, I'll merge

PR #1022, opened 2025-02-26
feat: adding global snapshot sync to schema

IPadawans:


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/LastSyncGlobalSnapshotStorage.scala
```diff
@@ -99,32 +102,44 @@ object LastSyncGlobalSnapshotStorage {
...
+              info.globalSnapshotSyncView.fold(
+                Async[F].pure(Option.empty[List[Hashed[GlobalIncrementalSnapshot]]])
+              ) { syncView =>
+                val mostFrequentOrdinal = syncView.values
+                  .map(_.globalSnapshotOrdinal)
+                  .groupBy(identity)
+                  .toList
+                  .sortBy { case (ordinal, occurrences) => (-occurrences.size, ordinal.value.value) }
+                  .map { case (ordinal, _) => ordinal }
+                  .headOption
```
scasplte2, line 115: Could this lead to a fork if we are using the most frequency global ordinal instead of requiring a majority? Though perhaps I am misunderstanding the wider context of when this is called.

scasplte2 (approved):
minor question to clarify for myself, otherwise LGTM

PR #1025, opened 2025-03-03
fix: adding new config fields-added-ordinals on config to read historical snapshots

IPadawans:
### Changes
+ The new globalSyncView will break for traversing historical snapshots when rolling back since we fetch the missing global snapshots directly from the peers. So, I'm disabling the validation of this field, such as we did with dataAppplication (I think?)

### DIFF
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
@@ -2,7 +2,6 @@ package io.constellationnetwork.node.shared.infrastructure.snapshot
 
 import cats.MonadThrow
 import cats.data.NonEmptyList
-import cats.data.Validated.{Invalid, Valid}
 import cats.effect.Async
 import cats.syntax.all._
 
@@ -15,7 +14,6 @@ import io.constellationnetwork.merkletree.Proof
 import io.constellationnetwork.merkletree.syntax._
 import io.constellationnetwork.node.shared.domain.block.processing._
 import io.constellationnetwork.node.shared.domain.node.UpdateNodeParametersAcceptanceManager
-import io.constellationnetwork.node.shared.domain.snapshot.services.GlobalL0Service
 import io.constellationnetwork.node.shared.domain.statechannel.StateChannelAcceptanceResult
 import io.constellationnetwork.node.shared.domain.statechannel.StateChannelAcceptanceResult.CurrencySnapshotWithState
 import io.constellationnetwork.node.shared.domain.swap.SpendActionValidator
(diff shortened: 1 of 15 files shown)


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotCreator.scala
```diff
@@ -271,7 +271,7 @@ object CurrencySnapshotCreator {
             currencySnapshotAcceptanceResult.sharedArtifacts.some,
             currencySnapshotAcceptanceResult.allowSpendBlock.accepted.toSortedSet.some,
             currencySnapshotAcceptanceResult.tokenLockBlock.accepted.toSortedSet.some,
-            currencySnapshotAcceptanceResult.globalSyncView.some
+            currencySnapshotAcceptanceResult.globalSyncView
```
scasplte2, line 274: can this be required here? I thought schema changes were required to be optional for historic compatibility

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
```diff
@@ -341,10 +341,10 @@ object GlobalSnapshotAcceptanceManager {
           updatedLastCurrencySnapshots,
           updatedLastCurrencySnapshotProofs,
           updatedAllowSpends.some,
-          updatedGlobalTokenLocks.some,
+          if (updatedGlobalTokenLocks.nonEmpty) Some(updatedGlobalTokenLocks) else None,
           updatedTokenLockBalances.some,
           updatedAllowSpendRefs.some,
-          updatedTokenLockRefs.some,
+          if (updatedTokenLockRefs.nonEmpty) Some(updatedTokenLockRefs) else None,
```
marcinwadon, line 347: `None` should not be assigned here probably, as `None` is exclusively for historical snapshots

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotCreator.scala
marcinwadon: yes, exactly @scasplte2 :+1:

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala

IPadawans:
The failing test will probably be solved in this PR:
https://github.com/Constellation-Labs/tessellation/pull/1030
I think we should merge this one to unblock testnet

scasplte2 (approved):
lgtm

marcinwadon:
> The failing test will probably be solved in this PR: #1030 I think we should merge this one to unblock testnet

Can you help me understand what part of this PR fixes the "InvalidSignature" error in #1030?

marcinwadon:
@IPadawans I think that #1030 should be merged first. It fixes the currency field (as described https://constellationnetwork.atlassian.net/jira/software/c/projects/PROT/boards/21?selectedIssue=PROT-1000)

ryle-goehausen (approved):
On unrelated note, I didn't see full error in detail, but if it's the distinction between a None and a Some(Map.empty) -- in any future cases it's always ideal to rely on a 'start' ordinal for this -- but if it doesn't change the end result of the new behavior (i.e. None vs. Map.empty) we can ALSO make the assumption that None is the same as Map.empty.

PR #1031, opened 2025-03-06
test: verify spend balances

marcinwadon:
- check doubled allow spend transactions
- check expired allow spend
- verify spend balances
- double use allow spend workflow

### DIFF
--- a/.github/action_scripts/send_transactions/allow-spends-and-spend-transactions.js
+++ b/.github/action_scripts/send_transactions/allow-spends-and-spend-transactions.js
@@ -94,7 +94,8 @@ const createInvalidParentAllowSpendTransaction = async (sourceAccount, ammAddres
         fee,
         lastValidEpochProgress: currentEpochProgress + CONSTANTS.EPOCH_PROGRESS_BUFFER,
         parent: invalidParent,
-        source: sourceAccount.address
+        source: sourceAccount.address,
+        currency: isCurrency ? CONSTANTS.CURRENCY_TOKEN_ID : null
     };
 };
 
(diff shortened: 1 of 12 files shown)


.github/action_scripts/send_transactions/allow-spends-and-spend-transactions.js
```diff
@@ -417,6 +419,196 @@ const executeInvalidTransactionWorkflow = async (createTransactionFn, workflowNa
...
+    const proof1 = await generateProof(allowSpend, sourcePrivateKey, sourceAccount, SerializerType.BROTLI);
+    const proof2 = await generateProof(allowSpend, sourcePrivateKey, sourceAccount, SerializerType.BROTLI);
+    
+    logWorkflow.info(`Created allow spend: ${JSON.stringify(allowSpend)}`);
+    logWorkflow.info(`Generated two proofs for the same allow spend`);
+
+    try {
+        const body1 = { value: allowSpend, proofs: [proof1] };
+        logWorkflow.info(`Sending allow spend to DAG L1: ${dagL1Url}`);
+        const response1 = await axios.post(`${dagL1Url}/allow-spends`, body1);
```
ryle-goehausen, line 476: const [response1, response2] = await Promise.all([request1, request2]);

Change this to only await so both requests are sent at the same time (better test of conflict detection if they're received at identical times.)

.github/action_scripts/send_transactions/allow-spends-and-spend-transactions.js
```diff
@@ -417,6 +419,196 @@ const executeInvalidTransactionWorkflow = async (createTransactionFn, workflowNa
...
+            urls.globalL0Url
+        );
+        
+        const isExtendedDagL1HashAccepted = await verifyDoubleSpendInL0(
+            result.address, 
+            result.extendedDagL1Hash, 
+            urls.globalL0Url
+        );
+        
+        const exactlyOneAccepted = (isDagL1HashAccepted && !isExtendedDagL1HashAccepted) || (!isDagL1HashAccepted && isExtendedDagL1HashAccepted);
```
ryle-goehausen, line 561: As long as we're returning the number of allow spends in the state, we should also be checking that to verify it matches what we expect. In case this test is run with other tests, at the beginning we should check the number and verify the count is incremented by 1.

ryle-goehausen (changes requested):

.github/action_scripts/send_transactions/allow-spends-and-spend-transactions.js
marcinwadon: > As long as we're returning the number of allow spends in the state

We're not returning the number of allow spends in the state

> we should also be checking that to verify it matches what we expect

And this is exactly what this code does 🤷‍♂️ 

Why we should care about other allow spends if there are any? What we need to verify is that only one of both allow spend has been accepted.

.github/action_scripts/send_transactions/allow-spends-and-spend-transactions.js
marcinwadon: done

.github/action_scripts/send_transactions/allow-spends-and-spend-transactions.js
marcinwadon: I've changed to verifying the number of allow spends in the state

ryle-goehausen (approved):

ryle-goehausen (approved):

PR #1032, opened 2025-03-06
fix: fixing spend action balance updates

IPadawans:


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/swap/SpendActionValidator.scala
```diff
@@ -53,23 +53,44 @@ object SpendActionValidator {
...
+                case Some(signedAllowSpend) =>
+                  if (signedAllowSpend.currency =!= spendTransaction.currency)
+                    InvalidCurrency(
+                      s"Currency mismatch: expected ${signedAllowSpend.currency}, found ${spendTransaction.currency}"
+                    ).invalidNec[SpendTransaction]
+                  else if (signedAllowSpend.destination =!= spendTransaction.destination)
+                    InvalidDestinationAddress(
+                      s"Destination address ${spendTransaction.destination} not found in active allow spends"
+                    ).invalidNec[SpendTransaction]
+                  else
```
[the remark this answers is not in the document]
IPadawans: I think we missed this validation, thanks a lot for pointing this

marcinwadon:
I think that the change with `destination` address you've made is incorrect. Destination of allow spend is the AMM metagraph id and destination in spend transaction is the receiver, not the AMM metagraph id. I've explained more on slack.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
```diff
@@ -764,80 +747,90 @@ object CurrencySnapshotAcceptanceManager {
...
-                  currentSourceBalance.plus(diffBetweenAllowSpendAndSpendTransaction).getOrElse(currentSourceBalance)
-                )
-            )
+                .updated(destinationAddress, updatedDestinationBalance)
+                .updated(sourceAddress, updatedSourceBalance)
+
+            case None =>
+              val updatedDestinationBalance = currentDestinationBalance
+                .minus(spendTransactionAmount)
+                .getOrElse(currentDestinationBalance)
```
marcinwadon, line 827: I think, based on what we've discussed on slack, for `None` case, the source is the metagraph address and the `spendTransactionAmount` should be deducted from the metagraph address.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
```diff
@@ -764,80 +747,90 @@ object CurrencySnapshotAcceptanceManager {
...
+              val currentSourceBalance = acc.getOrElse(sourceAddress, Balance.empty)
+              val balanceToReturnToAddress = allowSpend.amount.value.value - spendTransactionAmount.value.value
+
+              val updatedDestinationBalance = currentDestinationBalance
+                .plus(spendTransactionAmount)
+                .getOrElse(currentDestinationBalance)
+
+              val updatedSourceBalance = currentSourceBalance
+                .plus(Amount(NonNegLong.from(balanceToReturnToAddress).getOrElse(NonNegLong.MinValue)))
+                .getOrElse(currentDestinationBalance)
```
marcinwadon, line 818: shouldn't it be `currentSourceBalance`?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
```diff
@@ -509,6 +552,59 @@ object GlobalSnapshotAcceptanceManager {
...
+                .updated(destinationAddress, updatedDestinationBalance)
+                .updated(sourceAddress, updatedSourceBalance)
+
+            case None =>
+              val updatedDestinationBalance = currentDestinationBalance
+                .plus(spendTransactionAmount)
+                .getOrElse(currentDestinationBalance)
+
+              acc
+                .updated(destinationAddress, updatedDestinationBalance)
```
marcinwadon, line 602: as above

IPadawans: @marcinwadon In currency we can do this:
```
 val sourceAddress = currencyId
              val currentSourceBalance = acc.getOrElse(sourceAddress, Balance.empty)

              val updatedDestinationBalance = currentDestinationBalance
                .plus(spendTransactionAmount)
                .getOrElse(currentDestinationBalance)

              val updatedSourceBalance = currentSourceBalance
                .minus(spendTransactionAmount)
                .getOrElse(currentSourceBalance)

              acc
                .updated(destinationAddress, updatedDestinationBalance)
                .updated(sourceAddress, updatedSourceBalance)
```
but I'm not sure how we can do in global layer

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
IPadawans: 👍🏼

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
IPadawans: 👍🏼

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
```diff
@@ -509,6 +552,69 @@ object GlobalSnapshotAcceptanceManager {
...
+                  val updatedSourceBalance = currentSourceBalance
+                    .plus(Amount(NonNegLong.from(balanceToReturnToAddress).getOrElse(NonNegLong.MinValue)))
+                    .getOrElse(currentSourceBalance)
+
+                  innerAcc
+                    .updated(destinationAddress, updatedDestinationBalance)
+                    .updated(sourceAddress, updatedSourceBalance)
+
+                case None =>
+                  val currentSourceBalance = innerAcc.getOrElse(address, Balance.empty)
```
ryle-goehausen, line 600: I'm still double-checking this function, but this part concerns me the most to verify if it is correct.

`allAcceptedSpendTransactions: Map[Address, List[SpendAction]]`
might be useful to add some indication that this Address corresponds to the metagraph address associated with the CIS where this SpendAction was accepted.


`val currencySpendTransactions = spendActions.flatMap(spendAction => List(spendAction.input, spendAction.output))`

if we're mapping over both the input, and the output, which might have different currency denominations,

`currentBalances: SortedMap[Address, Balance],` 

but the balance map is only over [Address,Balance] rather than [CurrencyId,[Address,Balance]]

are we updating the correct balance? I need to review more of the code to be sure but this is worrying just on the types themselves.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/swap/SpendActionValidator.scala
```diff
@@ -53,23 +53,49 @@ object SpendActionValidator {
...
+                  else if (signedAllowSpend.amount.value.value < spendTransaction.amount.value.value)
+                    SpendAmountGreaterThanAllowed(
+                      s"Spend amount: ${spendTransaction.amount} greater than allowed: ${signedAllowSpend.amount}"
+                    ).invalidNec[SpendTransaction]
+                  else
+                    spendTransaction.validNec[SpendActionValidationError]
+              }
+            }
+
+        case None => spendTransaction.validNec[SpendActionValidationError].pure[F]
```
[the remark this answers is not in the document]
ryle-goehausen: So I realize we might need to update different balances in different cases in different places, so what might be easiest here is to keep track of a balance Map[CurrencyId, Map[Address, Balance]] and then ONLY apply the changes to each currency id where appropriate outside of the return value of this function.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
ryle-goehausen: So I realize we might need to update different balances in different cases in different places, so what might be easiest here is to keep track of a balance Map[CurrencyId, Map[Address, Balance]] and then ONLY apply the changes to each currency id where appropriate outside of the return value of this function.

IPadawans: Hey, the balance is [Address, Balance] because we're handling Global Balance only, the currency balance is on CurrencySnapshotAcceptanceManager

IPadawans:
```
allAcceptedSpendTransactions: Map[Address, List[SpendAction]]
might be useful to add some indication that this Address corresponds to the metagraph address associated with the CIS where this SpendAction was accepted.
```

Hmm yeah, we probably need a better naming

ryle-goehausen: `val currencySpendTransactions = spendActions.flatMap(spendAction => List(spendAction.input, spendAction.output))`

So either we need a filter in this to only update when currency id is DAG None, or we need to abstract this function to Map of currency id and then filter after the balances are updated like in my earlier comment

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
```diff
@@ -509,6 +552,69 @@ object GlobalSnapshotAcceptanceManager {
...
+
+      for {
+        allGlobalAllowSpends <- (incomingGlobalAllowSpends |+| lastActiveGlobalAllowSpends).toList.traverse {
+          case (address, allowSpends) =>
+            allowSpends.toList.traverse(_.toHashed).map(address -> _)
+        }.map(_.toSortedMap)
+
+        response = allAcceptedSpendTransactions.foldLeft(currentBalances) {
+          case (acc, (address, spendActions)) =>
+            val currencySpendTransactions = spendActions.flatMap(spendAction => List(spendAction.input, spendAction.output))
```
ryle-goehausen, line 571: This line is main issue, no filter on currencyid here

IPadawans: :+1

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
IPadawans: :+1

ryle-goehausen (dismissed):
I think this is fixed now, the currency acceptance definitely looks correct.

This part on GL0 I'm pretty sure is correct but need to review more to be certain.
      allAcceptedSpendTransactionsByCurrencyId: Map[Address, List[SpendAction]]

either way I think we should merge this to start testing more.

ryle-goehausen (approved):

PR #1033, opened 2025-03-06
fix: fixing majority alignment of token locks and allow spends

marcus-girolneto:
### Changes
+ During testing, I encountered a scenario where an allowSpend expired while holding the user’s full balance. Even after the balance was returned to the user, subsequent allowSpend transactions were blocked due to insufficient balance.
After investigating, I discovered that the function `advanceMajorityRefs` was never being called for allowSpends and token locks
I’ve fixed this issue, ensuring that balances are correctly restored and available for new transactions.

### DIFF
--- a/modules/dag-l1/src/main/scala/io/constellationnetwork/dag/l1/domain/snapshot/programs/SnapshotProcessor.scala
+++ b/modules/dag-l1/src/main/scala/io/constellationnetwork/dag/l1/domain/snapshot/programs/SnapshotProcessor.scala
@@ -2,28 +2,29 @@ package io.constellationnetwork.dag.l1.domain.snapshot.programs
 
 import cats.data.NonEmptyList
 import cats.effect.Async
-import cats.syntax.applicativeError._
-import cats.syntax.either._
-import cats.syntax.flatMap._
-import cats.syntax.functor._
-import cats.syntax.traverse._
+import cats.syntax.all._
 import cats.{Applicative, MonadThrow}
 
+import scala.collection.immutable.{SortedMap, SortedSet}
 import scala.util.control.NoStackTrace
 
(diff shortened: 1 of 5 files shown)


marcinwadon (approved):

modules/dag-l1/src/main/scala/io/constellationnetwork/dag/l1/domain/snapshot/programs/SnapshotProcessor.scala
```diff
@@ -2,28 +2,34 @@ package io.constellationnetwork.dag.l1.domain.snapshot.programs
 
 import cats.data.NonEmptyList
 import cats.effect.Async
+import cats.implicits.catsSyntaxTuple3Semigroupal
 import cats.syntax.applicativeError._
 import cats.syntax.either._
 import cats.syntax.flatMap._
 import cats.syntax.functor._
 import cats.syntax.traverse._
```
kpudlik, line 10: `import cats.syntax.all._`

kpudlik: Cats do not recommend mixing both.

kpudlik (changes requested):

modules/dag-l1/src/main/scala/io/constellationnetwork/dag/l1/domain/snapshot/programs/SnapshotProcessor.scala
IPadawans: 👍🏼

marcinwadon (approved):

PR #1034, opened 2025-03-07
docs: update node params endpoints

AlexBrandes:
Adds missing endpoints for node parameter updates (PR #999).

### DIFF
--- a/doc/api/dag/l0/public.yml
+++ b/doc/api/dag/l0/public.yml
@@ -12,6 +12,7 @@ tags:
   - name: DAG
   - name: State channel
   - name: Metric
+  - name: Node parameters
 
 paths:
   /cluster/info:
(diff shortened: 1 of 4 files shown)


doc/api/dag/l0/public.yml
```diff
@@ -338,3 +339,89 @@ paths:
...
+      tags:
+        - Node parameters
+      description: Submit node parameter metadata and delegated staking configuration
+      requestBody:
+        description: Signed node parameters
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '../../models/schemas/SignedUpdateNodeParameters.yml'
```
ryle-goehausen, line 354:
```
  private def getLatestNodeParameters(nodeId: Id): F[Option[(Signed[UpdateNodeParameters], SnapshotOrdinal)]] =
    for {
      maybeSnapshot <- getLatestFullSnapshot
    } yield maybeSnapshot.flatMap(_.value.info.updateNodeParameters.flatMap(_.get(nodeId)))
    case GET -> Root / PeerIdVar(nodeId) =>
      nodeStorage.getNodeState
        .map(validStateForSnapshotReturn)
        .ifM(
          getLatestNodeParameters(nodeId.toId).flatMap {
            case Some(params) => Ok(params)
            case _            => NotFound()
          },
          ServiceUnavailable()
        )
```
This is currently incorrect for the code on develop, since it returns this tuple structure. Since we're going to change this anyways here: https://constellationnetwork.atlassian.net/browse/PROT-1064 to

```
case class NodeParamsInfo(
  latest: Signed[UpdateNodeParams]
  lastRef: UpdateNodeParametersReference
  acceptedOrdinal: SnapshotOrdinal
)
```

We should just implement the docs against the (as yet unmerged) API

ryle-goehausen (changes requested):
Wrong response type on /node-params

ryle-goehausen:
@effects-ai please confirm on this PR that all relevant docs from here are merged / updated properly so we can close.

PR #1035, opened 2025-03-07
chore: add missing ordering instances and filter empty collections 

gclaramunt:
Changes needed to validate kryo-serialzed currency snapshots

### DIFF
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotCreator.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotCreator.scala
@@ -266,12 +266,12 @@ object CurrencySnapshotCreator {
             Option.when(currencySnapshotAcceptanceResult.info.lastMessages.nonEmpty)(
               currencySnapshotAcceptanceResult.messages.accepted.toSortedSet
             ),
-            currencySnapshotAcceptanceResult.globalSnapshotSync.accepted.toSortedSet.some,
+            currencySnapshotAcceptanceResult.globalSnapshotSync.accepted.toSortedSet.some.filter(_.nonEmpty),
             currencySnapshotAcceptanceResult.feeTransactions,
             currencySnapshotAcceptanceResult.sharedArtifacts.some,
-            currencySnapshotAcceptanceResult.allowSpendBlock.accepted.toSortedSet.some,
-            currencySnapshotAcceptanceResult.tokenLockBlock.accepted.toSortedSet.some,
-            currencySnapshotAcceptanceResult.globalSyncView.some
+            currencySnapshotAcceptanceResult.allowSpendBlock.accepted.toSortedSet.some.filter(_.nonEmpty),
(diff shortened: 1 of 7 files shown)


modules/shared/src/main/scala/io/constellationnetwork/schema/currencyMessage.scala
```diff
@@ -29,6 +30,8 @@ object currencyMessage {
 
     case object Owner extends MessageType("Owner")
     case object Staking extends MessageType("Staking")
+
+    implicit object OrderingInstance extends OrderBasedOrdering[MessageType]
```
marcinwadon, line 34: I'm not sure but there is already an ordering instance specified in `@derive(.., ordering)`, so you probably should remove the derivation if you want to use an explicitly defined one

ryle-goehausen: @gclaramunt in your tests did you find any difference between using the derive ordering macro vs. the scala builtin ordering implicit? We seem to have a mixed usage of either derive order, ordering, or derive order + implicit object declared order, but this is right we shouldn't be doing both

ryle-goehausen (changes requested):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotCreator.scala
```diff
@@ -266,12 +266,12 @@ object CurrencySnapshotCreator {
...
-            currencySnapshotAcceptanceResult.globalSnapshotSync.accepted.toSortedSet.some,
+            currencySnapshotAcceptanceResult.globalSnapshotSync.accepted.toSortedSet.some.filter(_.nonEmpty),
             currencySnapshotAcceptanceResult.feeTransactions,
             currencySnapshotAcceptanceResult.sharedArtifacts.some,
-            currencySnapshotAcceptanceResult.allowSpendBlock.accepted.toSortedSet.some,
-            currencySnapshotAcceptanceResult.tokenLockBlock.accepted.toSortedSet.some,
-            currencySnapshotAcceptanceResult.globalSyncView.some
+            currencySnapshotAcceptanceResult.allowSpendBlock.accepted.toSortedSet.some.filter(_.nonEmpty),
+            currencySnapshotAcceptanceResult.tokenLockBlock.accepted.toSortedSet.some.filter(_.nonEmpty),
+            currencySnapshotAcceptanceResult.globalSyncView.some.filter(_.ordinal > SnapshotOrdinal.MinValue)
```
ryle-goehausen, line 274: The other filters seem totally fine, but is there an explicit reason for this? The type itself here is already implementing NonNegLong validator? Is there a specific reason to exclude 0?

AlexBrandes:
@gclaramunt Is this PR still relevant?

ryle-goehausen:
@AlexBrandes some variant of this needs to be merged to fix the kryo serializer on the historical traverse for new nodes joining -- since that's the bug that was hit during snapshot streaming. So it's gotta be added to the registry for L0/L1

gclaramunt:
I'll probably update it to use the configuration to decide to use None or Some(empty) (we already have the ordinals in the config)

ryle-goehausen:
Closing this PR due to deprecation of registry

PR #1038, opened 2025-03-12
feat: updating spend actions fields to a non-empty-list of spend transactions

IPadawans:


ryle-goehausen (dismissed):
As on slack discussion, we don't need max length validation yet due to existing max size limitation.

ryle-goehausen (approved):

PR #1039, opened 2025-03-12
feat: adding artifacts field to GIS schema

IPadawans:
Screenshot 2025-03-12 at 17 16 15
Screenshot 2025-03-12 at 17 19 45

I've also tested rolling back with snapshots from testnet and it worked normally

### DIFF
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
@@ -32,7 +32,7 @@ import io.constellationnetwork.node.shared.domain.tokenlock.block.{
 import io.constellationnetwork.schema.ID.Id
 import io.constellationnetwork.schema._
 import io.constellationnetwork.schema.address.Address
-import io.constellationnetwork.schema.artifact.{SharedArtifact, SpendAction, SpendTransaction}
+import io.constellationnetwork.schema.artifact._
 import io.constellationnetwork.schema.balance.{Amount, Balance}
 import io.constellationnetwork.schema.epoch.EpochProgress
 import io.constellationnetwork.schema.node.UpdateNodeParameters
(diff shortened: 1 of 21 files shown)


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
```diff
@@ -381,6 +380,18 @@ object GlobalSnapshotAcceptanceManager {
 
         stateProof <- gsi.stateProof(maybeMerkleTree)
 
+        allowSpendsExpiredEvents <- emitAllowSpendsExpired(
+          lastActiveAllowSpends.getOrElse(None, SortedMap.empty[Address, SortedSet[Signed[AllowSpend]]]).collect {
```
ryle-goehausen, line 384: val expiredGlobalAllowSpends = lastActiveGlobalAllowSpends.collect {
        case (address, allowSpends) => address -> allowSpends.filter(_.lastValidEpochProgress < epochProgress)
      }

So it looks like this code is copied from here, which shows it's correct. But I think it might be cleaner instead to change the return type above where its getting acceptedAllowSpends to rely on this state mutation above, so its only defined in one place.

Also, I realize this filter is copied from there, but collect on a partial function where we're only ever matching one case statement and with no predicate and  filtering the values should instead be a view mapValues -- I tried to change it here but there's an overlap on filter def, so would have to modify imports or choose a different function. 

Neither of these changes are actually required, they're stylistic, so don't have to change this if not necessary.

ryle-goehausen (approved):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala

.github/workflows/setup-environment.yml
```diff
@@ -43,14 +56,14 @@ jobs:
         with:
           name: hypergraph-${{ runner.os }}-${{ hashFiles('**/build.sbt', '**/project/*.sbt', '**/project/build.properties') }}
           path: .github/code/hypergraph
-          retention-days: 7
+          retention-days: 1
```
marcinwadon, line 59: why only a single day? it means that tomorrow morning, everything would have to be built from scratch 🤔

IPadawans: The caches are consuming a lot of storage, I can increase but I think 7 days can be too much

ryle-goehausen (approved):

PR #1047, opened 2025-03-17
fix: delegated stakes api changes

effects-ai:


doc/api/models/schemas/CreateDelegatedStake.yml
```diff
@@ -1,7 +1,19 @@
...
+    description: Non-negative long value representing the delegated stake amount
+  fee:
+    type: integer
+    format: int64
+    example: 256
+    minimum: 0
+    description: Optional non-negative long value representing the delegated stake fee
   tokenLockRef:
     $ref: 'Hash.yml'
   stakeRef:
```
ryle-goehausen, line 19: Change this field name to `parent` for consistency

doc/api/models/schemas/CreateNodeCollateral.yml
```diff
@@ -1,7 +1,19 @@
...
+    description: Non-negative long value representing the node collateral amount
+  fee:
+    type: integer
+    format: int64
+    example: 123
+    minimum: 0
+    description: Optional non-negative long value representing the node collateral fee
   tokenLockRef:
     $ref: 'Hash.yml'
   collateralRef:
```
ryle-goehausen, line 19: Change this to parent

ryle-goehausen (changes requested):
"Add API docs for the stake lastRef endpoint – for getting the “parent” of CreateDelegatedStake. prefix / last-reference / address"

This is the last remaining item, even if the route is not added on this PR, lets get the API docs for it merged so we have types for UI to build against.

PR #1048, opened 2025-03-18
fix: fixing CurrencySnapshotAcceptanceManager

IPadawans:
### Changes
+ Since spendActions is a field in the snapshot schema, and we do not aggregate, if we skip on ordinal between lastGlobalView and lastSyncGlobalSnapshot, we have a chance to miss the spend actions. 
+ This PR aims to get all spend actions between this interval

### DIFF
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
@@ -86,7 +86,8 @@ trait CurrencySnapshotAcceptanceManager[F[_]] {
     calculateRewardsFn: SortedSet[Signed[Transaction]] => F[SortedSet[RewardTransaction]],
     facilitators: Set[PeerId],
     lastGlobalSnapshots: Option[List[Hashed[GlobalIncrementalSnapshot]]],
-    getGlobalSnapshotByOrdinal: SnapshotOrdinal => F[Option[Hashed[GlobalIncrementalSnapshot]]]
+    getGlobalSnapshotByOrdinal: SnapshotOrdinal => F[Option[Hashed[GlobalIncrementalSnapshot]]],
+    lastGlobalSyncView: Option[GlobalSyncView]
   )(implicit hasher: Hasher[F]): F[CurrencySnapshotAcceptanceResult]
 }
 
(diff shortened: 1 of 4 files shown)


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
```diff
@@ -438,6 +440,47 @@ object CurrencySnapshotAcceptanceManager {
...
+                      snapshot.spendActions.getOrElse(SortedMap.empty[Address, List[SpendAction]])
+                    case None =>
+                      SortedMap.empty[Address, List[SpendAction]]
+                  }
+              }
+            }.map(_.foldLeft(SortedMap.empty[Address, List[SpendAction]]) { (acc, spendMap) =>
+              acc |+| spendMap
+            })
+      }
+
```
scasplte2, line 483: minor modification to avoid the `traverse` + `foldLeft` and use a Map to avoid `.find`, otherwise seems reasonable

```
    private def getLastSpendActions(
                                     lastGlobalSyncView: Option[GlobalSyncView],
                                     maybeLastGlobalSnapshot: Option[Hashed[GlobalIncrementalSnapshot]],
                                     lastGlobalSnapshots: Option[List[Hashed[GlobalIncrementalSnapshot]]],
                                     getGlobalSnapshotByOrdinal: SnapshotOrdinal => F[Option[Hashed[GlobalIncrementalSnapshot]]]
                                   ): F[SortedMap[Address, List[SpendAction]]] =
      maybeLastGlobalSnapshot match {
        case None => SortedMap.empty[Address, List[SpendAction]].pure[F]
        case Some(lastGlobalSnapshot) =>
          val lastSyncOrdinal = lastGlobalSyncView.map(_.ordinal).getOrElse(lastGlobalSnapshot.ordinal)

          if (lastGlobalSnapshot.ordinal.value.value < lastSyncOrdinal.value.value) {
            SortedMap.empty[Address, List[SpendAction]].pure[F]
          } else {
            val snapshotsMap = lastGlobalSnapshots.map { snapshots =>
              snapshots.map(s => (s.ordinal, s)).toMap
            }.getOrElse(Map.empty)

            val start = lastSyncOrdinal.value.value
            val end = lastGlobalSnapshot.ordinal.value.value
            val snapshotOrdinals = (start to end)
              .map(l => SnapshotOrdinal(NonNegLong.unsafeFrom(l)))
              .toList

            Slf4jLogger.getLogger[F].debug(s"--- [CURRENCY] Fetching spend actions from global snapshots: $snapshotOrdinals") >>
              snapshotOrdinals.foldMapM { ordinal =>
                snapshotsMap.get(ordinal) match {
                  case Some(snapshot) =>
                    snapshot.spendActions.getOrElse(SortedMap.empty[Address, List[SpendAction]]).pure[F]
                  case None =>
                    getGlobalSnapshotByOrdinal(ordinal).map(_.flatMap(_.spendActions)
                      .getOrElse(SortedMap.empty[Address, List[SpendAction]]))
                }
              }
          }
      }
 ```

kpudlik (approved):

PR #1051, opened 2025-03-18
feat: allow overwriting delegated stake's node id

effects-ai:


ryle-goehausen (approved):
This mostly looks correct, but I don't see any diff here related to updating the state, I am presuming that is already in overwrite mode? (I.e. replace the map entry for that address / token lock) Can you confirm that is the case before merging?

PR #1060, opened 2025-03-24
feat: adding source to delegated state types

IPadawans:
### Changes
+ Adding source field to delegated staking types and updating the tests/validations

### DIFF
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/delegatedStake/UpdateDelegatedStakeValidator.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/delegatedStake/UpdateDelegatedStakeValidator.scala
@@ -10,7 +10,6 @@ import io.constellationnetwork.ext.cats.syntax.validated._
 import io.constellationnetwork.node.shared.domain.delegatedStake.UpdateDelegatedStakeValidator.UpdateDelegatedStakeValidationErrorOr
 import io.constellationnetwork.node.shared.domain.seedlist.SeedlistEntry
 import io.constellationnetwork.schema.address.Address
-import io.constellationnetwork.schema.delegatedStake.DelegatedStakeReference._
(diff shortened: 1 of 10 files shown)


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/delegatedStake/UpdateDelegatedStakeValidator.scala
```diff
@@ -51,10 +51,10 @@ object UpdateDelegatedStakeValidator {
         for {
           numberOfSignaturesV <- validateNumberOfSignatures(signed)
           signaturesV <- signedValidator
-            .validateSignatures(signed)
+            .isSignedExclusivelyBy(signed, signed.source)
```
ryle-goehausen, line 54:
```    
    def isSignedExclusivelyBy[A: Encoder](
      signed: Signed[A],
      signerAddress: Address
    ): F[SignedValidationErrorOr[Signed[A]]] =
      signed.proofs.existsM { proof =>
        proof.id.hex.toPublicKey.map { signerPk =>
          signerPk.toAddress =!= signerAddress
        }
      }.ifM(
        NotSignedExclusivelyByAddressOwner
          .asInstanceOf[SignedValidationError]
          .invalidNec[Signed[A]]
          .pure[F],
        signed.validNec[SignedValidationError].pure[F]
      )
```
I think you need both of these functions, it doesn't look like this is actually checking the signature? Also, there's a missing impl of this for GlobalSnapshotConsensusFunctions so verify this is in right impl.

I would re-add the validate and then keep this function call everywhere.

ryle-goehausen (changes requested):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/delegatedStake/UpdateDelegatedStakeValidator.scala
IPadawans: 👍🏼

ryle-goehausen (approved):

PR #1065, opened 2025-03-25
fix: validating node param last ref per nodeId

IPadawans:


ryle-goehausen (approved):
Realizing now we have an issue in terms of using the `source` Address field externally, but the Id field here internally, which is okay for now but might be a little confusing. (Since we're really using Id everywhere else)

PR #1067, opened 2025-03-26
feat: adding stakeRef to DelegatedStakeInfo schema

IPadawans:


doc/api/models/schemas/DelegatedStakeInfo.yml
```diff
@@ -16,6 +16,8 @@ properties:
     format: int64
     example: 123
     minimum: 0
+  stakeRef:
```
AlexBrandes, line 19: The name ‘stakeRef’ doesn’t make sense to me here because it’s a reference to itself (ID, hash, etc.). I think we should follow the naming convention we have with Transaction and name this field ‘hash’

AlexBrandes (approved):

PR #1071, opened 2025-03-27
fix: updating fee deduction on managers

IPadawans:


kpudlik (approved):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
```diff
@@ -799,8 +798,7 @@ object CurrencySnapshotAcceptanceManager {
             unexpired.foldLeft(acc.getOrElse(address, Balance.empty)) { (currentBalance, allowSpend) =>
```
kpudlik, line 798: I think it is still error prone. The logic of applying `unexpired` and `expired` is splitted into 2 separate `foldLeft` calls. So it can be that applying unexpired allow spend on address succeeds but applying expired one fails. Even if we make it in a single foldLeft then it still won't be transactional 🤔 .

The question is if it should be transactional per address so all expried+unexpired allow spends for address should be applied or none.

kpudlik (changes requested):

kpudlik (approved):

PR #1072, opened 2025-03-27
test: explicit full allow spend amount spend

marcinwadon:


marcinwadon (commented):
Actually I need to change one more thing before merging

marcinwadon:
> Actually I need to change one more thing before merging

Done

IPadawans (approved):

PR #1077, opened 2025-04-02
feat: update snapshot info for tracking delegated rewards

scasplte2:


modules/shared/src/main/scala/io/constellationnetwork/schema/delegatedStake.scala
```diff
@@ -105,6 +107,34 @@ object delegatedStake {
     pendingWithdrawals: List[DelegatedStakeInfo]
   )
 
+  @derive(decoder, encoder, eqv, show)
+  case class DelegatedStakeAccount(
+    activeStakes: List[(Signed[UpdateDelegatedStake.Create], SnapshotOrdinal)],
+    accruedRewards: Map[Address, Balance]
+  )
+
```
ryle-goehausen, line 115:
```
  activeDelegatedStakes: Option[SortedMap[Address, DelegatedStakeAccount]],
  delegatedStakesWithdrawals: Option[SortedMap[Address, List[PendingWithdrawal]]],
```
So we already have this keyed by address ( the delegator address,)

I would change this case class to be 
```
case class DelegatedStakeAccount(
create: Signed[UpdateDelegatedStake.Create]
acceptedAt: SnapshotOrdinal,
accruedRewardBalance: Balance
)
```

I'm unsure if you had a different meaning here for the Map of Address? But you could change it then so the state is:

  activeDelegatedStakes: Option[SortedMap[Address, List[DelegatedStakeAccount]]],

modules/shared/src/main/scala/io/constellationnetwork/schema/delegatedStake.scala
```diff
@@ -105,6 +107,34 @@ object delegatedStake {
...
+  object DelegatedStakeAccount {
+    val empty: DelegatedStakeAccount = DelegatedStakeAccount(List.empty, Map.empty[Address, Balance])
+
+    implicit val delegatedStakeWithBalanceOrdering: Ordering[DelegatedStakeAccount] =
+      new Ordering[DelegatedStakeAccount] {
+        def compare(x: DelegatedStakeAccount, y: DelegatedStakeAccount): Int = {
+          val xTotalBalance = x.accruedRewards.values.foldLeft(Balance.empty.value.value)((l, r) => l + r.value.value)
+          val yTotalBalance = y.accruedRewards.values.foldLeft(Balance.empty.value.value)((l, r) => l + r.value.value)
+
+          Ordering[Long].compare(xTotalBalance, yTotalBalance)
```
ryle-goehausen, line 125: Why the discrepancy in ordering in state? It seems like createdAt ordering for PendingWithdrawal makes a lot more sense to me for an internal state ordering. Is this ordering applied on the state object? Or for the REST API?

ryle-goehausen:
@IPadawans one more thing on this PR before testnet redeploy, I don't know if we need to update the config ordinals for these field activations? If the way you're planning to redeploy works with the configs as is, then we can leave this, otherwise we might need another set of config ordinals related to these changes (but since we changed the fields entirely .. probably okay as is?)

modules/shared/src/main/scala/io/constellationnetwork/schema/delegatedStake.scala
scasplte2: The original implementation used a `SortedMap`, this was simply a straightforward method to provide an `Ordering` instance as needed for that data structure.

modules/shared/src/main/scala/io/constellationnetwork/schema/delegatedStake.scala
scasplte2: thinking over it again, either way works really so I will see what is most straightforward

ryle-goehausen: >While we could move to the List[DelegatedStakeAccount] the accruedRewards makes it more obvious which operators are contributing. I was also trying to minimize modifications to prior function calls by using the same tuple-list.

So its really just an organization thing like you said, either way works, the thing is we already have a primary Address on the outer map which is equal to delegator address, and then we enforce this 1-1 correspondence so there's guaranteed uniqueness for (DelegatorAddress, NodeId) -- which means that list contains at MOST 1 entry per NodeId. That's not made obvious in the type signature, only the validator functions, so you're technically representing the enforcement thru a map... but still. I feel its a more natural code expression to group the balance with the data we want to associate with that balance for analytics purposes.

I.e. on the frontend or anywhere else, the natural "granular unit" of balance, is an address-nodeId delegation pair, which accrues a corresponding reward. That's the most raw format of it -- we can aggregate the balance in either way but IMO tracking it on the original chain data item is natural.

modules/shared/src/main/scala/io/constellationnetwork/schema/delegatedStake.scala
ryle-goehausen: Okay, I think that works. I'm pretty sure the sorting on state is never used anywhere. And since we're reconstructing the state map everywhere it's not really an issue. I could see this potentially being an issue if we eventually migrate the state to a separate store for the proof construction, because it shuffles based not on acceptance order, but that really doesn't matter yet so this is fine I think, can resolve this.

modules/shared/src/main/scala/io/constellationnetwork/schema/delegatedStake.scala
scasplte2: yeah, it is symmetric to have the List on the outside so I went with that

ryle-goehausen (approved):
lgtm let's get marcus to verify this is correct for testnet deploy otherwise seems ready for tomorrow

IPadawans:
> @IPadawans one more thing on this PR before testnet redeploy, I don't know if we need to update the config ordinals for these field activations? If the way you're planning to redeploy works with the configs as is, then we can leave this, otherwise we might need another set of config ordinals related to these changes (but since we changed the fields entirely .. probably okay as is?)

Hey, we already have the fields for that, I've created. Since we just changed the inner types, the PR seems to be fine

PR #1078, opened 2025-04-03
fix: node-params endpoint shows all avaialble nodes

Serpion-ua:
/node-params endpoint now shows information about all available nodes, even if the node is not in a cluster (basic information shows then only)


IPadawans (approved):
LGTM

PR #1080, opened 2025-04-07
fix: adding parallelism on currency snapshot acceptance

IPadawans:
### Changes
+ Adding parallelism on the GlobalSnapshotStateChannelEventsProcessor, since we can accept different metagraph snapshots in parallel
+ Updated the acceptRewards function to not merge the 2 balances map directly in memory, and use a optimized structure

### DIFF
--- a/.github/workflows/check-scala-code.yml
+++ b/.github/workflows/check-scala-code.yml
@@ -17,7 +17,7 @@ jobs:
         java:
           - openjdk@1.11.0
     steps:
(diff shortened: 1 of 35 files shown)


gclaramunt (dismissed):

IPadawans:
The failing tests are only when saving the artifacts on the action, this is probably solved in tessellation v3. All the transaction tests are passing.

ryle-goehausen (approved):

PR #1081, opened 2025-04-08
feat: delegated rewards

scasplte2:


modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/config/types.scala
```diff
@@ -95,7 +114,9 @@ object types {
       OneTimeReward(EpochProgress(1928500L), treasureWalletMetanomics1, TransactionAmount(150_000_000_000_000_00L)),
       OneTimeReward(EpochProgress(1928500L), treasureWalletMetanomics2, TransactionAmount(150_000_000_000_000_00L)),
       OneTimeReward(EpochProgress(1928500L), treasureWalletMetanomics3, TransactionAmount(150_000_000_000_000_00L))
-    )
+    ),
+    // Fallback reward pool in case no emission formula is configured
+    fallbackRewardPoolPerEpoch: Amount = Amount(10000000000L)
```
ryle-goehausen, line 119: Does this ever actually get triggered? Is this dev only?

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/config/types.scala
```diff
@@ -114,47 +135,63 @@ object types {
...
+            stardustNewPrimary -> NonNegFraction.unsafeFrom(5L, 100L),
+            stardustSecondary -> NonNegFraction.unsafeFrom(5L, 100L),
+            testnet -> NonNegFraction.unsafeFrom(3L, 100L),
+            dataPool -> NonNegFraction.unsafeFrom(55L, 100L),
+            integrationNet -> NonNegFraction.unsafeFrom(15L, 100L)
+          ),
+          validatorsWeight = NonNegFraction.unsafeFrom(17L, 100L),
+          delegatorsWeight = NonNegFraction.unsafeFrom(0L, 100L)
+        )
+      case epoch if epoch < EpochProgress(2000000L) => // TODO - change value at launch - Marks transition to new base case
```
ryle-goehausen, line 170: Need to review the code in the repo itself to double check, but is this accounting for separate values for integration vs. mainnet?

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/config/types.scala
```diff
@@ -114,47 +135,63 @@ object types {
...
+            integrationNet -> NonNegFraction.unsafeFrom(88L, 1000L),
+            treasureWalletMetanomics2 -> NonNegFraction.unsafeFrom(30L, 100L)
           ),
-          remainingWeight = Weight(24L) // facilitators
+          validatorsWeight = NonNegFraction.unsafeFrom(88L, 1000L),
+          delegatorsWeight = NonNegFraction.unsafeFrom(45L, 100L)
         )
     }
 
+    // setting an epoch progress value here will override the emission formula calculation
```
ryle-goehausen, line 194: Same question, case for intNet vs. mainnet

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/rewards/DelegatedRewardsDistributor.scala
```diff
@@ -0,0 +1,330 @@
...
+      val iInitial = emConfig.iInitial
+      val lambda = emConfig.lambda
+      val epochsPerYear = emConfig.epochsPerYear.value
+      val yearInitialEpoch = emConfig.yearInitial.value.value
+      val secondsPerEpoch = emConfig.secondsPerEpoch.value
+      val totalSupply = emConfig.totalSupply.value
+      val iImpact = emConfig.iImpact
+
+      if (emConfig.dagPrices.values.isEmpty) {
+        Slf4jLogger.getLogger[F].error("Empty DAG price configuration").as(Amount(NonNegLong.unsafeFrom(0L)))
```
ryle-goehausen, line 91: Wondering here if we want a default that represents the current value? Not sure we can even get to this point normally or what the consequences of 0 are, maybe better not to have it or is this proceeding assuming 0?

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/config/types.scala
```diff
@@ -78,15 +78,34 @@ object types {
...
+    */
+  case class EmissionConfig(
+    epochsPerYear: PosLong, // Number of epochs in a year (yardstick for time-based calculations)
+    yearInitial: EpochProgress, // Initial reference year epoch
+    iTarget: NonNegFraction, // Target inflation rate (0.5%)
+    iInitial: NonNegFraction, // Initial inflation rate (6%)
+    lambda: NonNegFraction, // Decay constant (0.1)
+    dagPrices: SortedMap[EpochProgress, NonNegFraction], // DAG price data - to be replaced by oracle in the future
+    iImpact: NonNegFraction, // Impact factor (0.35)
+    secondsPerEpoch: PosLong, // Seconds per epoch scaling factor (empirically measured as 43 seconds)
```
ryle-goehausen, line 98: I only see one reference to this value, and the reference has no usages beyond an initial val assignment. I think we can remove this -- epochsPerYear should account for the 43 second average as part of the yardstick definition, so I'm not sure if this is an artifact from an earlier branch.

ryle-goehausen:
@scasplte2 I'm unsure if the current E2E failures are due to arg parsing ? OR just need to be rebased against dev.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/config/types.scala
```diff
@@ -114,47 +135,63 @@ object types {
...
         ProgramsDistributionConfig(
           weights = Map(
-            stardustNewPrimary -> Weight(7L),
-            testnet -> Weight(7L),
-            dataPool -> Weight(42L),
-            integrationNet -> Weight(20L)
+            stardustNewPrimary -> NonNegFraction.unsafeFrom(5L, 100L),
+            testnet -> NonNegFraction.unsafeFrom(24L, 1000L),
+            integrationNet -> NonNegFraction.unsafeFrom(88L, 1000L),
+            treasureWalletMetanomics2 -> NonNegFraction.unsafeFrom(30L, 100L)
```
ryle-goehausen, line 187: I would give this a specific distinct variable name (a duplicate), just to clarify it -- even if the address value is the same as ticket.

ryle-goehausen (dismissed):
Summary of outstanding issues for tomorrow morning:
* Need to ensure the override is implemented for IntNet on yardstick and beginning epoch of changes.
* Rebase against develop
* Verify the E2E is passing (most likely due to args or merge)

You can disregard all the stylistic comments -- they are not necessary to fix.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/config/types.scala
scasplte2: the order of application of the different rewards is controlled by the function below in the `DelegatedRewardDistributor` but there is no distinction of different rewards amounts mapping to `AppEnvironment` in this config. 

```scala
      /** Calculate the total rewards to mint for this epoch based on appropriate rewards mechanism. The calculation follows this priority:
       *   1. Use the emission formula + pre-configured rewards (if configured) and epoch is past the transition point
       *   1. Use the emission formula and epoch is past the transition point
       *   1. Use preconfigured rewards per epoch mapping for epochs before the emission formula transition (v1 & v2)
       *   1. Return zero rewards if no configuration is available
       */
      def calculateTotalRewardsToMint(epochProgress: EpochProgress): F[Amount] = {

        implicit val rewardOrdering: Ordering[(EpochProgress, Amount)] =
          (x: (EpochProgress, Amount), y: (EpochProgress, Amount)) => x._1.compare(y._1)

        // Get environment-specific configurations
        val envEmissionConfig = delegatedRewards.emissionConfig.get(environment)

        // Get the transition epoch for this environment (if available)
        val transitionEpoch = envEmissionConfig.map(_.asOfEpoch).getOrElse(EpochProgress.MaxValue)

        // First check for static rewards per epoch defined for this environment
        rewardCfg.rewardsPerEpoch.filter {
          case (epoch, _) =>
            epoch.value <= epochProgress.value && epochProgress.value < transitionEpoch.value
        }.maxOption match {
          case Some((_, amount)) if envEmissionConfig.exists(ec => epochProgress.value >= ec.asOfEpoch.value) =>
            calculateEmissionRewards(epochProgress, envEmissionConfig.get).map(_.value + amount.value).map(sum => Amount(NonNegLong.unsafeFrom(sum)))

          case None if envEmissionConfig.exists(ec => epochProgress.value >= ec.asOfEpoch.value) =>
            calculateEmissionRewards(epochProgress, envEmissionConfig.get)

          case Some((_, amount)) =>
            amount.pure[F]

          // Return zero rewards if no environment-specific config exists
          case _ =>
            Slf4jLogger
              .getLogger[F]
              .warn(s"No emission config found for $environment, returning zero rewards")
              .as(Amount(NonNegLong.unsafeFrom(0L)))
        }
      }
      ```

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/config/types.scala
scasplte2: this was removed, there is a fallback in case no emission config is found and the epoch is not covered by the provided `rewardsPerEpoch`. If that val is correct, we should not hit this case.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/rewards/DelegatedRewardsDistributor.scala
scasplte2: I've moved to providing the config values directly in source for the time-being so we should be able to guarantee a price. I was thinking `0` was more for testing.

IPadawans (approved):

PR #1091, opened 2025-04-09
fix: fixing schema validation from the new tessellation v3 fields

IPadawans:


scasplte2 (approved):

scasplte2:
@IPadawans I've created a refactor to rewards that targets this branch so may be worth pulling those in or just pointing the larger PR directly to int deploy. Just a heads up

PR #1095, opened 2025-04-11
fix: adding invalid balances removal for metagraphs

IPadawans:


modules/node-shared/src/main/scala/org/tessellation/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
```diff
@@ -137,9 +138,19 @@ object CurrencySnapshotAcceptanceManager {
...
 
+      updatedBalancesByInvalidAddressChecks = metagraphsInvalidAddresses
+        .get(lastSnapshotContext.address)
+        .map { info =>
+          if (info.snapshotOrdinal === snapshotOrdinal)
+            updatedBalancesByRewards
+          else
+            info.filterFunction(updatedBalancesByRewards)
+        }
+        .getOrElse(updatedBalancesByRewards)
```
scasplte2, line 149: @IPadawans I think this needs to be flipped, no?
```scala
        .map { info =>
          if (info.snapshotOrdinal === snapshotOrdinal)
            info.filterFunction(updatedBalancesByRewards)
          else
            updatedBalancesByRewards
        }
        .getOrElse(updatedBalancesByRewards)
```

ryle-goehausen (approved):

scasplte2 (approved):

PR #1099, opened 2025-04-14
feat: add logging & metrics to global snapshot consensus

scasplte2:


modules/dag-l0/src/main/scala/org/tessellation/dag/l0/infrastructure/snapshot/GlobalSnapshotConsensusStateAdvancer.scala
```diff
@@ -41,230 +44,329 @@ abstract class GlobalSnapshotConsensusStateAdvancer[F[_]]
...
     consensusStorage: GlobalConsensusStorage[F],
     globalSnapshotStorage: SnapshotStorage[F, GlobalSnapshotArtifact, GlobalSnapshotContext],
     consensusFns: GlobalSnapshotConsensusFunctions[F],
     gossip: Gossip[F]
-  ): GlobalSnapshotConsensusStateAdvancer[F] = new GlobalSnapshotConsensusStateAdvancer[F] {
-    val logger = Slf4jLogger.getLogger[F]
-    val facilitatorsObservationName = "facilitators"
+  ): F[GlobalSnapshotConsensusStateAdvancer[F]] =
+    (ConsensusMetricsLogger.make[F], GlobalSnapshotMetrics.make[F], CurrencySnapshotAnalyzer.make[F]).mapN {
+      (metricsLogger, globalMetrics, currencyAnalyzer) =>
```
IPadawans, line 55: unused logger, is this intentional?

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/infrastructure/snapshot/GlobalSnapshotConsensusStateAdvancer.scala
```diff
@@ -41,230 +44,329 @@ abstract class GlobalSnapshotConsensusStateAdvancer[F[_]]
...
+
+          // Helper method to collect metrics in a non-invasive way
+          private def collectMetrics(
+            context: GlobalSnapshotContext,
+            key: GlobalSnapshotKey,
+            startTime: FiniteDuration
+          ): F[Unit] =
+            Clock[F].monotonic.flatMap { currentTime =>
+              context match {
+                case globalInfo: org.tessellation.schema.GlobalSnapshotInfo =>
```
IPadawans, line 68: maybe import this part:
`org.tessellation.schema`
to use only GlobalSnapshotInfo

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/infrastructure/snapshot/GlobalSnapshotConsensusStateAdvancer.scala
```diff
@@ -41,230 +44,329 @@ abstract class GlobalSnapshotConsensusStateAdvancer[F[_]]
...
+                      // Get the CurrencySnapshotInfo from the snapshot
+                      currencySnapshot.fold(
+                        _ => Async[F].unit, // Skip full snapshots
+                        {
+                          case (_, currencyInfo) =>
+                            // Log balance map size metrics
+                            val balanceSize = currencyInfo.balances.size
+
+                            // Track metrics for the currency snapshot info
+                            currencyInfo.balances.get(address).traverse_ { _ =>
```
IPadawans, line 81: can we maybe use `parTraverse`? Since we're just collecting infos to have the metrics

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/infrastructure/snapshot/GlobalSnapshotConsensusStateAdvancer.scala
```diff
@@ -41,230 +44,329 @@ abstract class GlobalSnapshotConsensusStateAdvancer[F[_]]
...
+                                      lastArtifact,
+                                      state.lastOutcome.finished.context,
+                                      HasherSelector[F].getForOrdinal(lastArtifact.ordinal),
+                                      majorityTrigger,
+                                      events,
+                                      state.facilitators.value.toSet
+                                    )
+                                }
+                                (artifact, context, returnedEvents) = proposalResult
+                                _ <- collectMetrics(context, state.key, startTime)
```
IPadawans, line 181: Should we collect the metrics only on the `collectingFacilities` state? I mean on the `creatingProposal` we should collect also or skip?

marcus-girolneto (approved):
I think this is great, I've just put a couple of small comments, more questions than other things. But, in overall, LGTM

modules/dag-l0/src/main/scala/org/tessellation/dag/l0/infrastructure/snapshot/GlobalSnapshotConsensusStateAdvancer.scala
scasplte2: left over from dev, thank you for pointing it out. I ending up simplifying this a bit into a single metric logger

AlexBrandes (commented):
Can we test this on IntNet before we test on MainNet? I think since MainNet is stable now, we can go back to only adding new features to the V3 release.

scasplte2:
> Can we test this on IntNet before we test on MainNet? I think since MainNet is stable now, we can go back to only adding new features to the V3 release.

converted back to a draft so I can rebase on `develop`. Due to the namespace change this is essentially a by-hand situation so I will re-submit once the work is complete.

ryle-goehausen (changes requested):
So biggest issue here with this PR is it's touching a lot of mission critical code, and the changes are a huge number of lines due mostly to indentation changes.

As much as possible, if we can split this into a few different PRs, I think that would help most of all, and as you're rebasing on dev, if you can attempt to preserve original indentations / functions where-ever possible, it would make it easier to review. 

I realize that's not possible in all places, due to the insertion of the metrics type, but it may be possible to write it such that there's a single func change at the top that tuples the metrics in, and then the rest of function stays same. Not always straightforward but it would help when you do the rebase to keep in consideration.

scasplte2:
closing since this is stale. I will incorporate the feedback and metrics into future work

PR #1100, opened 2025-04-14
fix: delegated stakes duplication

IPadawans:


scasplte2 (approved):

IPadawans:
@effects-ai the part you've changed it's only on the rewards distribution, no? I mean if you take a look in the `GlobalSnapshotAcceptanceManager` you have the logic to accept/reject delegated stakes, specifically on `acceptDelegatedStakes`, don't we have to add sth there to avoid duplication?

effects-ai (approved):
LGTM

PR #1104, opened 2025-04-17
feat: add latest reward diffs to global incr snapshot 

scasplte2:
## Overview
Addresses bug impacting snapshot streaming in v3 by moving latest assigned delegated rewards into time trigger incremental snapshots. This allows for context recreation for streaming.


IPadawans (dismissed):

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/rewards/GlobalDelegatedRewardsDistributor.scala
```diff
@@ -105,12 +67,60 @@ object DelegatedRewardsDistributor {
...
+
+      val reservedWeight = reservedAddressWeights.map(_._2).sum
+      val validatorsWeight = pctConfig.validatorsWeight.toBigDecimal
+      val delegatorsWeight = pctConfig.delegatorsWeight.toBigDecimal
+      val totalWeight = reservedWeight + validatorsWeight + delegatorsWeight
+
+      val delegatorFlatInflationPercentage = delegatedRewardsConfig.flatInflationRate.toBigDecimal
+
+      val reservedAddressRewards =
+        if (totalWeight <= 0) List.empty
```
scasplte2, line 109: I think this logic should be updated to return zero if the totalRewards is zero not the totalWeight. I will change this

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotContextFunctions.scala
```diff
@@ -34,8 +34,98 @@ import eu.timepit.refined.types.all.NonNegLong
...
-  def make[F[_]: Async: Parallel: HasherSelector](snapshotAcceptanceManager: GlobalSnapshotAcceptanceManager[F]) =
+  def make[F[_]: Async: Parallel: HasherSelector: SecurityProvider](
+    snapshotAcceptanceManager: GlobalSnapshotAcceptanceManager[F],
+    updateDelegatedStakeAcceptanceManager: UpdateDelegatedStakeAcceptanceManager[F],
+    withdrawalTimeLimit: EpochProgress,
+    tessellation3MigrationStartingOrdinal: SnapshotOrdinal
+  ) =
     new GlobalSnapshotContextFunctions[F] {
+
+      private def acceptDelegatedStakes(
```
scasplte2, line 45: we should put a todo so that we can not repeat the function here and in `GlobalSnapshotConsensusFunctions`

ryle-goehausen (approved):

PR #1108, opened 2025-04-21
fix: remove active stakes after withdrawal

effects-ai:


modules/shared/src/main/scala/io/constellationnetwork/schema/delegatedStake.scala
```diff
@@ -97,7 +97,7 @@ object delegatedStake {
   @derive(eqv, show, encoder)
   case class DelegatedStakeInfo(
     nodeId: PeerId,
-    acceptedOrdinal: SnapshotOrdinal,
+    acceptedOrdinal: Option[SnapshotOrdinal],
```
ryle-goehausen, line 100: So I think if we change the other state information, we can still preserve this information for determining the snapshot ID associated with the actual update. (It requires one change.) We need to either do that and revert this, or change the API docs to reflect the change.

ryle-goehausen (changes requested):
Two issues:

First error, after withdrawal, not removing empty list for address in snapshot info
activeDelegatedStakes":{"DAG1vmb6wbdKgMRite7nTmp5Di8mT5ZqjRw6KNTc":[]}
Second, after withdrawal, same thing for token locks.
"activeTokenLocks":{"DAG0LpUnSFwUmwMSeLry5R3xgL5MFf5rXStBfQLZ":[]}

ryle-goehausen (approved):

PR #1110, opened 2025-04-23
test: delegated stake test fragment & wallet methods

ryle-goehausen:
reproducible method for bash script for manual debugging of delegated stake operations

Should be possibly to reproduce entirely locally without any external dependencies / key pairs.

Added new wallet methods, skipped several validators / test code is low quality but this is mostly throwaway for tracking our internal tests so we have reproducible environment.

Update node params is being accepted, presently token lock is having issue here. I based the test changes off the latest PR branch from @effects-ai


IPadawans (approved):
This is a great idea, and thanks for doing that. This can help with testing. LGTM, the failure task it's just due to uppercase in the commit message

PR #1116, opened 2025-04-28
test: l1 structured logging example metrics for join debug

ryle-goehausen:


modules/dag-l1/src/main/scala/io/constellationnetwork/dag/l1/domain/consensus/block/Validator.scala
```diff
@@ -53,15 +56,25 @@ object Validator {
     blockStorage: BlockStorage[F],
     transactionStorage: TransactionStorage[F],
     peersCount: PosInt,
-    tipsCount: PosInt
+    tipsCount: PosInt,
+    tokenLockStorage: TokenLockStorage[F]
```
scasplte2, line 60: this seems kind of off-topic to inject token lock storage into this place just to count. I understand for debugging purposes but it seems out of place unless we want to include more information about node A's view of node B instead of just specifically token locks.

ryle-goehausen: fixed, I think I added this because I presumed it was a token lock failure, but the error is actually just related to how quickly the join is started, which is where the other statements help.

scasplte2 (approved):

PR #1118, opened 2025-04-28
fix: cleanup empty token lock lists

effects-ai:


ryle-goehausen (approved):

ryle-goehausen:
note, this needs to be updated in currency manager as well. but this fixes current issue at least

PR #1128, opened 2025-05-01
fix: allow for accepted creates that modify existing delegations

scasplte2:


ryle-goehausen:
@scasplte2 just double checking, this branch doesn't have the 'skip one rewards' fix in it yet right? Will wait to review until confirm that's in.

scasplte2:
> @scasplte2 just double checking, this branch doesn't have the 'skip one rewards' fix in it yet right? Will wait to review until confirm that's in.

@ryle-goehausen yes, I forgot to come back and convert back to draft. Doing so now

scasplte2:
there still seems to be an issue, so converting back once again :/

ryle-goehausen (approved):
One nice unit / e2e test we could add for this (or as separate ticket,) would be verifying that a modify request can only be updated by the correct matching address.

.github/action_scripts/delegated_staking/delegated-staking.js
```diff
@@ -688,7 +694,7 @@ const checkUpdateDelegatedStake = async (urls, account, stakeHash, nodeId) => {
             nodeId,
             amount: originalStake.amount,
             tokenLockRef: originalStake.tokenLockRef,
-            // rewardAmount: originalStake.rewardAmount, // balance is transferred
+            rewardAmount: originalStake.rewardAmount, // balance is transferred
```
ryle-goehausen, line 697: More of a question / comment here, are we asserting that the rewardAmount remains EXACTLY the same during the update? So if another reward round runs before this assertion, the test will fail? 

It seems like a greater than equals is required? Or to verify the exact state at every ordinal to know that we've actually transferred correct amount. I could see this becoming flaky.

AlexBrandes: It's possible this could become flaky but I think it's unlikely. We hit this endpoint as soon as we get a successful response from the gL0 that the stake was created which would be before the snapshot that will contain the new stake position has been created, then we recheck every second if not initially found. Somehow a timed snapshot with additional rewards would need to undergo consensus and update the state in less than a second to have this be greater than the transferred reward amount the first time we check.

PR #1131, opened 2025-05-05
test: working e2e for delegated stake with docker

ryle-goehausen:


scasplte2 (approved):
was able to build and run the clusters locally and had the test execute successfully :tada:

PR #1133, opened 2025-05-06
feat: price oracle common types

effects-ai:


scasplte2:
@greptileai please review

modules/shared/src/main/scala/io/constellationnetwork/schema/priceOracle.scala
```diff
@@ -0,0 +1,51 @@
...
+
+    def of(hashedTransaction: Hashed[PricingUpdate]): PricingUpdateReference =
+      PricingUpdateReference(hashedTransaction.ordinal, hashedTransaction.hash)
+
+    def of[F[_]: Async: Hasher](signedTransaction: Signed[PricingUpdate]): F[PricingUpdateReference] =
+      signedTransaction.value.hash.map(PricingUpdateReference(signedTransaction.ordinal, _))
+  }
+
+  @derive(eqv, show, encoder, decoder)
+  case class PriceRecord(currencyId: CurrencyId, fractionalValue: NonNegFraction, updatedAt: EpochProgress)
```
scasplte2, line 49: Do we have a standard for how this fraction is defined? As in I previously assumed price was defined as (number of DAG per dollar) so `NonNegFraction.from(25, 1)` would mean a price of $0.04 per DAG. If we are tracking this per currencyId are we always going to reference price to USD or to DAG? Also, can we ensure that the numerator and denominator aren't flipped (maybe due to a developer misunderstanding who defines NonNegFraction.from(1, 25) since that number would be 1/25 = 0.04 but would be read as 1 dollar per 25 DAG.

modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshotInfo.scala
```diff
@@ -219,7 +230,8 @@ case class GlobalSnapshotInfo(
   activeDelegatedStakes: Option[SortedMap[Address, List[DelegatedStakeRecord]]],
   delegatedStakesWithdrawals: Option[SortedMap[Address, List[PendingDelegatedStakeWithdrawal]]],
   activeNodeCollaterals: Option[SortedMap[Address, List[NodeCollateralRecord]]],
-  nodeCollateralWithdrawals: Option[SortedMap[Address, List[PendingNodeCollateralWithdrawal]]]
+  nodeCollateralWithdrawals: Option[SortedMap[Address, List[PendingNodeCollateralWithdrawal]]],
+  priceState: Option[SortedMap[CurrencyId, PriceRecord]]
```
scasplte2, line 234: not super familiar with the reqs for this but are we wanting to track the price of any ML0 or just for DAG?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
```diff
@@ -528,7 +530,8 @@ object GlobalSnapshotAcceptanceManager {
           if (ordinal < tessellation3MigrationStartingOrdinal) none else updatedCreateDelegatedStakesCleaned.some,
           if (ordinal < tessellation3MigrationStartingOrdinal) none else updatedWithdrawDelegatedStakesCleaned.some,
           if (ordinal < tessellation3MigrationStartingOrdinal) none else updatedCreateNodeCollateralsCleaned.some,
-          if (ordinal < tessellation3MigrationStartingOrdinal) none else updatedWithdrawNodeCollateralsCleaned.some
+          if (ordinal < tessellation3MigrationStartingOrdinal) none else updatedWithdrawNodeCollateralsCleaned.some,
+          if (ordinal < tessellation3MigrationStartingOrdinal) none else updatedPriceState.some
```
IPadawans, line 534: This is wrong, we already rolled back networks using this ordinal, in your case you'll need to introduce a new configuration and select one ordinal to start storing this value, otherwise, the state will break

modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshotInfo.scala
IPadawans, line 234: Maybe the naming isn't that good, and this will ensure that we always have a currency and their priceRecord. If we want to trace DAG, we will probably want 
`Option[SortedMap[Option[CurrencyId], PriceRecord]]` where None is DAG
also `priceState` seems redundant since this is already a state, maybe we can change to sth like:
`currencyPrices`, or `currencyCurrentPrices`. Just suggestions

marcus-girolneto (changes requested):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
scasplte2: @ryle-goehausen @AlexBrandes 
Marcus makes a good point, we have now encountered this ordinal based logic switching in 3 places I can think of
1. hashing
2. rewards (both classic for different distribution percentages & the global switch between algorithms)
3. global v3 switch with introduction of associated functionality

We likely should introduce a context bound as a constraint on the `F[_]` with an associated algebra to handle the implementation of this logic for us instead of the ad-hoc approaches that have worked so far.

ryle-goehausen: Yeah, I think you're right here @scasplte2  -- but I'd go a step further too, while its great to have trigger ordinals, we should introduce a common pattern that defaults to None for historical changes -- such that we don't need as many flags in the code to handle this case.

There's no difference between a schema field that is None ( and doesn't modify the hash ) and the prior schema, we really need to take advantage of that, and only rely on the ordinals as behavior switching triggers. I don't think we have an established pattern of this anywhere.

modules/shared/src/main/scala/io/constellationnetwork/schema/artifact.scala
```diff
@@ -40,4 +41,10 @@ object artifact {
   case class AllowSpendExpiration(
     allowSpendRef: Hash
   ) extends SharedArtifact
+
+  @derive(decoder, encoder, order, ordering, show)
+  case class PricingUpdate(currencyId: CurrencyId, fractionalValue: NonNegFraction, markOverride: Boolean, parent: PricingUpdateReference)
```
ryle-goehausen, line 46: Same here, needs Option[CurrencyId]

modules/shared/src/main/scala/io/constellationnetwork/schema/priceOracle.scala
ryle-goehausen: So I think datum/USD is most ideal from our discussion. But also, I think this record should be defined exclusively in terms of the latest value of the PricingUpdate and the updatedAt, since it's redundant fields otherwise (and the ref storage should also be included here so it's not in another lastref field.

So PriceRecord(latest: PricingUpdate, updatedAt: EpochProgress)

ryle-goehausen (changes requested):

ryle-goehausen (commented):
Putting a summary of meeting discussion here for reference. We decided to avoid using datum/USD -- as this introduces a future uncertainty around the definition of the Amount type as it pertains to external network amount definitions. Since we lack a unified union type to represent all amounts for future oracles, datum/USD doesn't generalize well to future additional types.

Instead, we should preserve the fraction class approach, using it to represent a price in the same currency / quote currency as the oracle originally returns, in this case DAG/USD. The additional information we need to include, is the numerator / denominator CurrencyId.

CurrencyId must be changed, such that it does not refine as an address validated type, but refines as a type which either validates against a strictly defined enum, or the existing address type. We can introduce DAG as one of the enum variants for now, so long as we maintain compatibility with the existing presumption that None=DAG everywhere else in the codebase. It's advisable to translate this type internally to None for any existing interface that requires it in that format, and to define a USD currency type.

It is not required to introduce a new generic Amount type. It's sufficient to use the existing fraction structures, to represent data to the same precision we are expecting from the oracle endpoints (without excess gain or less of the expected precision.) This represents the data as a price point of the same original precision as the oracle, with the additional enriched information associated with the two currency identifiers used for the price -- this assumes in the future, that any calculations we perform, will need to be able to understand the underlying conversions to their respective datum equivalents.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotConsensus.scala
```diff
@@ -89,6 +89,7 @@ object GlobalSnapshotConsensus {
       feeCalculator = FeeCalculator.make(feeConfigs)
       snapshotAcceptanceManager = GlobalSnapshotAcceptanceManager.make(
         sharedCfg.fieldsAddedOrdinals.tessellation3Migration.getOrElse(sharedCfg.environment, SnapshotOrdinal.MinValue),
+        sharedCfg.fieldsAddedOrdinals.tessellation301Migration.getOrElse(sharedCfg.environment, SnapshotOrdinal.MinValue),
```
scasplte2, line 92: reasonable idea to add new ordinals for these sorts of things but if that is required then we may want to plan out and group functionality like with v3 (since I could see this blowing up quickly) though I think that is more an overall strategy that @ryle-goehausen or @AlexBrandes might have opinions on how to approach

scasplte2 (commented):
just one question about how we continue to "turn on" functionality through ordinals otherwise looks reasonable to me

modules/shared/src/main/scala/io/constellationnetwork/schema/swap.scala
```diff
@@ -48,7 +48,7 @@ object swap {
     implicit def toAmount(fee: AllowSpendFee): Amount = Amount(fee.value)
   }
 
-  @derive(decoder, encoder, order, show, ordering)
+  @derive(decoder, encoder, keyDecoder, keyEncoder, order, show, ordering)
   @newtype
   case class CurrencyId(value: Address)
```
ryle-goehausen, line 53: unit test the validation here for refined

modules/shared/src/main/scala/io/constellationnetwork/schema/priceOracle.scala
```diff
@@ -0,0 +1,126 @@
...
+  sealed trait TokenId
+  case class ExternalToken(id: PartnerChain) extends TokenId
+  case class HypergraphToken(id: MemberChain) extends TokenId
+
+  object TokenId {
+    implicit val keyDecoder: KeyDecoder[TokenId] = new KeyDecoder[TokenId] {
+      def apply(key: String): Option[TokenId] =
+        // Split the key into type and value parts
+        key.split(":") match {
+          case Array("ExternalToken", id) =>
```
ryle-goehausen, line 50: outstanding question for discussion

modules/shared/src/main/scala/io/constellationnetwork/schema/priceOracle.scala
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  /** An asset pair representation.
+    *
+    * Base asset: This is the first asset in the pair. It's the one you are buying or selling.
+    *
+    * Quote asset: This is the second asset in the pair. It shows how much of this asset is needed to buy one unit of the base asset.
+    */
+  @derive(decoder, encoder, keyEncoder, keyDecoder, show, order, ordering)
+  case class TokenPair(base: TokenId, quote: TokenId)
```
ryle-goehausen, line 72: Perfect

ryle-goehausen (dismissed):

modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshotInfo.scala
```diff
@@ -60,6 +64,7 @@ object GlobalSnapshotInfoV1 {
       Some(SortedMap.empty),
       Some(SortedMap.empty),
       Some(SortedMap.empty),
+      Some(SortedMap.empty),
```
ryle-goehausen, line 61: not sure this is the only place, but ideally we treat empty map as none.

modules/node-shared/src/main/resources/application.conf
```diff
@@ -132,6 +132,12 @@ fields-added-ordinals {
     integrationnet: 3330000,
     dev: 0
   }
+  tessellation-301-migration {
+    mainnet: 0,
+    testnet: 2700000,
```
effects-ai: Hey @IPadawans, what ordinal should I put here for the next testnet release?

ryle-goehausen (commented):
Outstanding issues from discussion: 

calculated state needs epoch progress for price update accepted at.

need a validator on whitelist to prevent random spam of price updates.

modules/node-shared/src/main/resources/application.conf
ryle-goehausen: see other comments I made on PR -- but I'd really like if we can try and establish a pattern here, of not using Some(emptyMap) for new state transitions -- because it breaks the hasher.

I think we won't need to coordinate in the future on 'exact' perfect ordinals nearly as much, if we can guarantee that no data is produced before a certain time period, it doesn't matter exactly which release the state change is made on that way.

@effects-ai can you try to implement the changes here, such that we ALWAYS produce a None instead of a Some(emptyMap) in the event that no data is received?

@IPadawans also, we do need to figure out what ordinals make sense as a 'first pass' -- since I know we're doing another release train before this one, so if you can provide a sensible default guess so @effects-ai can continue his testing on testnet at least, that would be helpful. Otherwise, we can attempt to guess at an ordinal for testnet and then disable intnet / mainnet entirely (extremely higher ordinal or something.)

marcus-girolneto: This will be the ordinal that should be used to decide if we will fill as nothing of something, so you can provide the ordinal to work immediately after the release is done, which will require stop the network get the ordinal, and then generating the version with this ordinal, or you can estimate and provide a slightly future ordinal. You just need to have in mind that we will fill the field only after crossing this ordinal

modules/currency-l1/src/test/scala/io/constellationnetwork/currency/l1/domain/snapshots/programs/CurrencySnapshotProcessorSuite.scala
```diff
@@ -59,8 +59,7 @@ object CurrencySnapshotProcessorSuite extends SimpleIOSuite with TransactionGene
           )
           currencySnapshotAcceptanceManager <- CurrencySnapshotAcceptanceManager
             .make(
-              FieldsAddedOrdinals(Map.empty, Map.empty, Map.empty),
-              Dev,
+              SnapshotOrdinal.MinValue,
```
marcus-girolneto: Please merge with develop to use the new pattern:
FieldsAddedOrdinals and Dev

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotConsensus.scala
marcus-girolneto: Please merge with develop to use the new pattern:
FieldsAddedOrdinals and Dev

modules/dag-l0/src/test/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotTraverseSuite.scala
```diff
@@ -301,8 +301,7 @@ object GlobalSnapshotTraverseSuite extends MutableIOSuite with Checkers {
     for {
       validationErrorStorage <- CurrencySnapshotEventValidationErrorStorage.make(TestValidationErrorStorageMaxSize)
       currencySnapshotAcceptanceManager <- CurrencySnapshotAcceptanceManager.make(
-        FieldsAddedOrdinals(Map.empty, Map.empty, Map.empty),
-        Dev,
+        SnapshotOrdinal.MinValue,
```
marcus-girolneto: Please merge with develop to use the new pattern:
FieldsAddedOrdinals and Dev

marcus-girolneto (changes requested):
It mostly looks good, but you need to merge develop first because I'm seeing several spots where this code is deleting/changing develop behavior

marcus-girolneto (approved):
LGTM, we just need to wait for the tests to pass

PR #1139, opened 2025-05-08
test: refactor delegated staking e2e tests

AlexBrandes:


.github/action_scripts/delegated_staking/lib.js
scasplte2, line 1: could this `lib` be used for tests outside of `delegated_staking`? If so, then my preference would be to move it a layer up in the organization so tests of `node_params` and `delegated_rewards` might import that functionality and extend it to their specific use case.

.github/action_scripts/delegated_staking/delegated-staking.js
```diff
@@ -414,152 +320,50 @@ const checkCreateNodeParameters = async (urls) => {
   )
   logWorkflow.info('Update second node params is OK')
 
-  const bothNodesParams = await getNodeParams(urls)
```
ryle-goehausen, line 417: This file is getting extremely large, it makes the most sense to split the node params request into a different file since they're somewhat independent.

.github/action_scripts/delegated_staking/lib.js
```diff
@@ -0,0 +1,355 @@
...
+        Pragma: 'no-cache',
+        Expires: '0',
+      },
+    },
+  )
+  checkOk(response)
+  return response.data
+}
+
+const fetchSnapshot = async (urls, ordinal) => {
```
ryle-goehausen, line 290: I think a utility like this that doesn't pertain to delegated staking doesn't belong in this file

.github/action_scripts/delegated_staking/delegated-staking.js
AlexBrandes: Cool - I like that. It makes it more explicit that node params are being tested also.

AlexBrandes:
I'm going to go ahead and merge this to bank these improvements and I'll make a larger refactor in a separate PR.

.github/action_scripts/delegated_staking/lib.js
AlexBrandes: Yeah, I think a bigger refactor to create reusable methods would be a good idea. Most of the tests rewrite the same helper functions, etc.

PR #1143, opened 2025-05-13
feat: add rewards-info endpoint

scasplte2:


doc/api/models/schemas/NextDagPrice.yml
```diff
@@ -0,0 +1,12 @@
+type: object
+description: Information about the next DAG price and when it will take effect
+properties:
+  price:
+    type: integer
+    format: int64
+    description: The next DAG price (in the same format as used internally in configuration)
```
ryle-goehausen, line 7: Since this is user facing API docs, I would explicitly describe the format. I think we decided on datum/USD?

doc/api/models/schemas/RewardsInfo.yml
```diff
@@ -0,0 +1,43 @@
...
+description: Information about rewards configuration on the network for calculating APY
+properties:
+  epochsPerYear:
+    type: integer
+    format: int64
+    description: Number of epochs per year (based on current GS epoch configuration)
+  currentDagPrice:
+    type: integer
+    format: int64
+    description: Current DAG price (in the same format as used internally in configuration)
```
ryle-goehausen, line 11: Same here

doc/api/models/schemas/RewardsInfo.yml
```diff
@@ -0,0 +1,43 @@
...
+    description: Total amount delegated across all active delegated stakes, including accrued rewards
+  latestAverageRewardPerDag:
+    type: integer
+    format: int64
+    description: Average reward rate per DAG per epoch in datums
+  totalDagAmount:
+    type: integer
+    format: int64
+    description: Current total supply of all DAG balances in datums
+  totalRewardPerEpoch:
```
ryle-goehausen, line 27: Realizing now this might not be the most descriptive name (but it's same as in ticket.) Maybe a better name here would be lastTotalIncrementalRewards or something like that? Or we leave it as is, and communicate it as a "rate" of change. Not sure actually.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/http/routes/DelegatedStakesRoutes.scala
```diff
@@ -140,6 +147,144 @@ final case class DelegatedStakesRoutes[F[_]: Async: Hasher](
...
+    lastSnapshotInfo: GlobalSnapshotInfo
+  ): F[RewardsInfo] =
+    for {
+      emConfig <- delegatedRewardsConfig.emissionConfig
+        .get(environment)
+        .pure[F]
+        .flatMap(Async[F].fromOption(_, new RuntimeException(s"Could not retrieve emission config for env: $environment")))
+
+      (_, _, totalDelegateStake, currentTotalSupply) <- processDelegations(lastSnapshotInfo)
+      latestDelegateRewardsNoCommission <- getLatestDelegateRewardTotal(lastSnapshot, lastSnapshotInfo)
```
ryle-goehausen, line 171: Maybe worth a longer discussion at some point on reducing the load on this API by pre-calculating this, but since we're doing this everywhere else probably not worth implementing here.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/http/routes/DelegatedStakesRoutes.scala
```diff
@@ -140,6 +147,144 @@ final case class DelegatedStakesRoutes[F[_]: Async: Hasher](
...
+        val stakeAmount = record.event.value.amount.value.value
+        val rewardsAmount = record.rewards.value
+        (stakeAcc + stakeAmount, rewardsAcc + rewardsAmount)
+    }
+
+    (
+      toAmount(totalStakeLocked),
+      toAmount(totalActiveRewards),
+      toAmount(totalStakeLocked + totalActiveRewards), // totalDelegateStake
+      toAmount(totalSpendableSupply + totalPendingSupply + totalActiveRewards) // currentTotalSupply
```
ryle-goehausen, line 212: I don't think this method should be a part of this class, for instance, it's missing balances from AllowSpends. It makes way more sense to declare a method on top of the state to sum the balances.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/http/routes/DelegatedStakesRoutes.scala
```diff
@@ -140,6 +147,144 @@ final case class DelegatedStakesRoutes[F[_]: Async: Hasher](
...
+      .getOrElse(SortedMap.empty[PeerId, Map[Address, Amount]])
+
+    val nodeParams = info.updateNodeParameters
+      .getOrElse(SortedMap.empty[ID.Id, (Signed[UpdateNodeParameters], SnapshotOrdinal)])
+
+    val calcFullReward: (Long, (PeerId, Map[Address, Amount])) => Long = {
+      case (acc, (peerId, rewards)) =>
+        val nodeCommission = BigDecimal(
+          nodeParams.get(peerId.toId).map(_._1.delegatedStakeRewardParameters.reward).getOrElse(0.0)
+        )
```
ryle-goehausen, line 227: Similar issue here as above, where this method (or conversion,) should be associated with the class itself, such that we get back a decimal related to the class instance (and can therefore see how it was originally declared.

I'm also wondering if this should just be wrapped in the other fraction type which does the conversion there. It doesn't feel great having decimal logic encapsulated in places like this.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/http/routes/DelegatedStakesRoutes.scala
```diff
@@ -140,6 +147,144 @@ final case class DelegatedStakesRoutes[F[_]: Async: Hasher](
...
+      .foldLeft(0L)(calcFullReward)
+      .pure[F]
+      .flatMap(toAmount)
+  }
+
+  private def calculateAverageReward(latestRewards: Amount, totalStakedAmount: Amount): F[Amount] =
+    if (totalStakedAmount.value.value === 0) Amount.empty.pure[F]
+    else
+      toAmount(
+        (BigDecimal(latestRewards.value.value) / BigDecimal(totalStakedAmount.value.value))
```
ryle-goehausen, line 246: Similar thing, it'd be great if we can hide the decimal interactions behind an API we control so it's more strictly verifiable.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/http/routes/DelegatedStakesRoutes.scala
```diff
@@ -200,5 +345,17 @@ final case class DelegatedStakesRoutes[F[_]: Async: Hasher](
...
         case None => ServiceUnavailable()
       }
+
+    case GET -> Root / "rewards-info" =>
+      snapshotStorage.head.flatMap {
+        case Some((snapshot, info)) =>
+          getRewardsInfo(snapshot, info)
+            .flatMap(Ok(_))
+            .handleErrorWith { error =>
+              logger.error(error)(s"Error getting rewards info: ${error.getMessage}") >>
```
ryle-goehausen, line 355: Maybe over-reaching here, but it'd be nice to track failures like this in a metric, so we don't have to scan logs.

doc/api/models/schemas/NextDagPrice.yml
ryle-goehausen: This is now changed to DAG/USD but as a fraction, we need the doc here to correspond to whatever fractional basis format that we're planning to return here (Or whatever it's presently represented as.)

scasplte2:
@greptileai please review again

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/modules/Services.scala
```diff
@@ -122,6 +123,12 @@ object Services {
         )
       getOrdinal = storages.globalSnapshot.headSnapshot.map(_.map(_.ordinal))
       trustUpdaterService = TrustStorageUpdater.make(getOrdinal, sharedServices.gossip, storages.trust)
+
+      rewardsService = RewardsService(
+        classicRewards,
+        delegatorRewards,
+        DefaultDelegatedRewardsConfigProvider.getConfig()
+      )
```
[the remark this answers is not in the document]
scasplte2: refactored approach

doc/api/models/schemas/RewardsInfo.yml
scasplte2: updated to datum/usd with the assumption of the minimum supported price of 1 datum/usd. As above, I think we could revisit this with the updated types

ryle-goehausen (approved):

PR #1147, opened 2025-05-15
feat: add operation for refined types non neg long and epoch progress

Serpion-ua:


ryle-goehausen (approved):
Discussed in meeting already, useful for |+| additions.

PR #1148, opened 2025-05-16
fix: validating state per block and setting last currency snapshot

IPadawans:
### Changes
+ This PR aims to fix the forking issue with currency L0.
+ After investigation, I've noticed 2 inconsistencies on the  code:
1st -> We were validating the same state to all blocks, and we were validating all blocks/updates at once. So I've fixed this part
2nd -> We were missing the lastCurrencySnapshot to validator nodes joining the network. In some metagraphs, such as DO,R we use this information to combine, and if we don't have this, when we join the consensus, we produce inconsistent artifacts

### DIFF
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/DataApplicationSnapshotAcceptanceManager.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/DataApplicationSnapshotAcceptanceManager.scala
@@ -1,6 +1,7 @@
 package io.constellationnetwork.node.shared.infrastructure.snapshot
 
 import cats.Applicative
+import cats.data.Validated.{Invalid, Valid}
 import cats.data.{NonEmptyList, OptionT}
 import cats.effect.Async
 import cats.syntax.all._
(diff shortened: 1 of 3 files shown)


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/DataApplicationSnapshotAcceptanceManager.scala
```diff
@@ -150,42 +151,69 @@ object DataApplicationSnapshotAcceptanceManager {
...
+                  else
+                    logger.debug(s"Block ${dataBlock.value.roundId} is invalid: ${validation.fold(_.toList.mkString(", "), _ => "")}")
+                }.flatMap {
+                  case Valid(_) =>
+                    val dataTransactionsAsList = dataTransactions.toList
+                    val dataUpdates = getDataUpdates(dataTransactionsAsList)
+                    val feeTransactions = getFeeTransactions(dataTransactionsAsList)
+
+                    logger.debug(s"Block ${dataBlock.value.roundId} is valid") >>
+                      service.combine(currentState, dataUpdates).map { newState =>
```
ryle-goehausen, line 185: For anyone else reviewing, this is and

`.foldLeftM(initialResult) {`

are the most important changes. We discussed this on slack already, but it wasn't properly updating the state

ryle-goehausen (approved):

PR #1153, opened 2025-05-19
feat: removing collateral check on context functions

marcus-girolneto:
### Changes
+ Adding the parameter `shouldValidateCollateral` on block validations.
+ This parameter, by default, is true, and we will validate the collateral always in the snapshot creation moment, in both layers: currency and DAG
+ However, the functions that only create the context, don't need to perform this validation since the snapshot is already created and we're only building the context

### Tests
**Before the change:**
If only the global layer had the required 250K collateral and the metagraph didn’t, the snapshot was incorrectly rejected.
Screenshot 2025-05-19 at 14 44 07
**After the change:**
With debug prints added, we confirmed that collateral is now validated only during snapshot creation, not when building the context. The behavior is now correct.
Screenshot 2025-05-19 at 15 15 49
Screenshot 2025-05-19 at 15 16 01


modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
```diff
@@ -89,7 +89,8 @@ object CurrencySnapshotConsensusFunctions {
           None,
           None,
           lastGlobalSnapshots,
-          getGlobalSnapshotByOrdinal
+          getGlobalSnapshotByOrdinal,
+          shouldValidateCollateral = true
```
ryle-goehausen, line 93: Haven't looked at full PR yet, but is this supposed to be true within this file if its the currency snapshot functions?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/block/processing/BlockAcceptanceManager.scala
```diff
@@ -9,13 +9,15 @@ trait BlockAcceptanceManager[F[_]] {
   def acceptBlocksIteratively(
     blocks: List[Signed[Block]],
     context: BlockAcceptanceContext[F],
-    snapshotOrdinal: SnapshotOrdinal
+    snapshotOrdinal: SnapshotOrdinal,
+    shouldValidateCollateral: Boolean = true
```
ryle-goehausen, line 13: Is this code shared by cl0 and gl0?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/block/processing/BlockAcceptanceLogic.scala
```diff
@@ -187,25 +179,30 @@ object BlockAcceptanceLogic {
 
   def processSignatures[F[_]: Async: SecurityProvider](
     signedBlock: Signed[Block],
-    context: BlockAcceptanceContext[F]
+    context: BlockAcceptanceContext[F],
+    shouldValidateCollateral: Boolean = true
```
ryle-goehausen, line 183: is this only defaulting to true for gl0?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/block/processing/BlockAcceptanceLogic.scala
```diff
@@ -187,25 +179,30 @@ object BlockAcceptanceLogic {
...
-        .map(_.id.toPeerId)
-        .toList
-        .traverse(_.toAddress)
-        .flatMap(
-          _.filterA(address =>
-            context.getBalance(address).map { balances =>
-              !balances.getOrElse(Balance.empty).satisfiesCollateral(context.getCollateral)
-            }
+      if (!shouldValidateCollateral) {
+        ().asRight[BlockNotAcceptedReason].pure
```
ryle-goehausen, line 187: I'm confused by this, so if we shouldn't validate collateral, then this always is a rejection?

ryle-goehausen, line 187: I'm confused by this, so if we shouldn't validate collateral, then this always is a rejection?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
```diff
@@ -89,7 +89,8 @@ trait CurrencySnapshotAcceptanceManager[F[_]] {
     facilitators: Set[PeerId],
     lastGlobalSnapshots: Option[List[Hashed[GlobalIncrementalSnapshot]]],
     getGlobalSnapshotByOrdinal: SnapshotOrdinal => F[Option[Hashed[GlobalIncrementalSnapshot]]],
-    lastGlobalSyncView: Option[GlobalSyncView]
+    lastGlobalSyncView: Option[GlobalSyncView],
+    shouldValidateCollateral: Boolean
```
ryle-goehausen, line 93: For currency shouldn't this always be false?

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
marcus-girolneto: Yes @ryle-goehausen this is the only place where we call on ML0 to run the consensus, so if we have collateral set on metagraph layer, we will respect the validation here

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/block/processing/BlockAcceptanceManager.scala
marcus-girolneto: Yes it's shared

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/block/processing/BlockAcceptanceLogic.scala
marcus-girolneto: It's defaulting to true for all I think

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/block/processing/BlockAcceptanceLogic.scala
marcus-girolneto: No, we call the `asRight`  providing that the right portion is the failure, but we're returning the Unit here

marcus-girolneto: No, we call the asRight providing that the right portion is the failure, but we're returning the Unit here

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
marcus-girolneto: No because we will validate the currency collateral if needed, but ignore on the construction of snapshot on Global

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
ryle-goehausen: ah right, sorry -- so that makes sense to me now, the error is only on the GL0 validating ML0 -- within ML0, it will still default to checking the environment variable and skipping this true if not present (iirc). So this 'true' isn't actually "force validation"

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/block/processing/BlockAcceptanceManager.scala
ryle-goehausen: Okay so if it's shared, then won't the GL0 invoke the currency portion of this and default to true ? I haven't traced execution of this to see if when the GL0 invokes this function wrt to CL0, does it override this default? It may be better not to use a default value for anything that is potentially invoked by both -- this may actually be fine I'm just not actually certain.

If we had a test to cover this (I'll check rest of PR,) then that wouldn't be required.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/block/processing/BlockAcceptanceLogic.scala
ryle-goehausen: I think maybe we should add a unit test of GL0 invocating currency acceptance. Not sure how hard that would be, maybe easier to do as E2E ? Problem is I'm not sure if this is being overridden by GL0 recreating CL0 context.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/block/processing/BlockAcceptanceLogic.scala
ryle-goehausen: My mistake, this is fine, didn't read closely enough.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
ryle-goehausen: You're right

modules/node-shared/src/test/scala/io/constellationnetwork/node/shared/infrastructure/block/processing/BlockAcceptanceLogicSuite.scala
```diff
@@ -73,6 +73,15 @@ object BlockAcceptanceLogicSuite extends MutableIOSuite with Checkers {
     }
   }
 
+  test("accept block with signers without collateral when skipped validation") { implicit sc =>
```
ryle-goehausen, line 76: The most important test is invocation of the GL0 state update function, which has a currency snapshot embedded, with no collateral provided. That's basically the only condition we need to cover, as the E2E can cover the other case.

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
marcus-girolneto: yes, the ML0 will still validate the collateral but GL0 no

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/block/processing/BlockAcceptanceManager.scala
marcus-girolneto: This should be provided as true always IMO, on CurrencyManager is the place where we call this, and on currency layer we provide as false. I think all validations should explicitly be true by default, and only in places where we want to skip we manually provide false

ryle-goehausen (approved):
As discussed on slack, it's sufficient to test this on testnet, since we have a pretty easy way to reproduce the actual problem this is solving, and I believe all the default params are already covered by existing E2E tests, so this is good enough I think.

PR #1160, opened 2025-05-22
fix: parallelizing getLastSpendActions

marcus-girolneto:


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
```diff
@@ -492,50 +492,50 @@ object CurrencySnapshotAcceptanceManager {
...
-                    )
+              val (cached, missing) = snapshotOrdinals.partition(snapshotCache.contains)
+
+              val fromCache: List[SortedMap[Address, List[SpendAction]]] =
+                cached.flatMap(ordinal => snapshotCache.get(ordinal).flatMap(_.spendActions).toList)
+
+              val fetchMissing: F[List[SortedMap[Address, List[SpendAction]]]] =
+                missing.parTraverse { ordinal =>
+                  getGlobalSnapshotByOrdinal(ordinal)
+                    .map(_.flatMap(_.spendActions).getOrElse(SortedMap.empty))
```
scasplte2, line 529: If one of the missing values errors out while trying to retrieve it, then should this whole operation be skipped or is should just that element be skipped? I'm just thinking that this returns an `F[_]` wrapped value which might error so can we handle the case of missing one or do we need the guarantee that all elements have been fetched?

ryle-goehausen: Need to review this more to confirm, but the case we need to support is error on not everything fetched. It can't be considered healthy if it misses a single fetch, but the inner getOrElse empty map is fine. If we successfully fetch the ordinal, but the spendactions are empty, that's okay.

If any of the fetches fail, we should either have a retry in the fetcher which waterfalls to error out, and propagate the errors up so a bulk fetch errors on any individual failure. I think? these map statements are doing that but need to confirm

PR #1163, opened 2025-05-26
fix: calling combine with empty updates

marcus-girolneto:


scasplte2:
@IPadawans what is the purpose of running the combine function on empty blocks?

gclaramunt (approved):

scasplte2 (approved):

AlexBrandes (changes requested):
Didn't we disable this for a reason? What's the purpose of bringing it back?

marcus-girolneto:
@AlexBrandes, we didn't disable it on purpose; this was a bug that I introduced when I changed the behavior of validating all blocks at once to validate separately. You can check the history of the file. This is the bug that's throwing errors on AMM Ci tests, because we need to call combine even without updates

marcus-girolneto:
The combine function should be called always, with or without updates, since several metagraphs use this function to update state according to the epochProgress even without updates

AlexBrandes (approved):

PR #1166, opened 2025-05-27
pullGlobalSnapshots with optional ordinal

gclaramunt:


ryle-goehausen (changes requested):
I think we should change the branch to merge here to develop so we go through the normal release cycle here

PR #1224, opened 2025-06-02
feat: adding last global snapshot with currency snapshot

marcus-girolneto:
### Changes
+ Adding the information of the last global snapshot that contains the currency snapshots, per metagraphId.
+ This will be stored on state and will saves us several rollback time, since we don't need to get the latest global and traverse back until the first global with currency snapshot

### Tests
I've performed some tests locally, and it worked as expected:
Screenshot 2025-06-02 at 13 27 39

### NOTE
We need to update purconfig ordinals to match exactly the starting point


modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshotInfo.scala
```diff
@@ -138,11 +141,12 @@ object GlobalSnapshotStateProof {
...
       Option[Hash],
       Option[Hash],
+      Option[Hash],
       Option[Hash]
     )
   ) => GlobalSnapshotStateProof = {
-    case (x1, x2, x3, x4, x5, x6, x7, x8, x9, x10, x11, x12, x13, x14) =>
-      GlobalSnapshotStateProof.apply(x1, x2, x3, x4, x5, x6, x7, x8, x9, x10, x11, x12, x13, x14)
+    case (x1, x2, x3, x4, x5, x6, x7, x8, x9, x10, x11, x12, x13, x14, x15) =>
+      GlobalSnapshotStateProof.apply(x1, x2, x3, x4, x5, x6, x7, x8, x9, x10, x11, x12, x13, x14, x15)
```
scasplte2, line 149: We're going to hit a Scala limit at 22 elements I believe

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
```diff
@@ -103,7 +106,8 @@ object GlobalSnapshotAcceptanceManager {
   case object InvalidMerkleTree extends NoStackTrace
 
   def make[F[_]: Async: Parallel: HasherSelector: SecurityProvider](
-    tessellation3MigrationStartingOrdinal: SnapshotOrdinal,
+    fieldsAddedOrdinals: FieldsAddedOrdinals,
```
scasplte2, line 109: I think we should revisit this at some point and try to define an algebra that provides this logic in the F context so we can expand the specific service selection logic component and then in this portion of the code just use the result of that selection.

For now this is fine but there seems to be an obvious pattern emerging that we should pull together into a single definition, if possible.

modules/node-shared/src/main/resources/application.conf
```diff
@@ -132,6 +132,12 @@ fields-added-ordinals {
     integrationnet: 3330000,
     dev: 0
   }
+  global-snapshots-with-currency-snapshots {
+    mainnet: 4409045,
+    testnet: 2497000,
+    integrationnet: 3330000,
+    dev: 0
+  }
```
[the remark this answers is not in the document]
marcus-girolneto: it will be updated right before the merging

modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshotInfo.scala
```diff
@@ -86,6 +87,7 @@ case class GlobalSnapshotStateProofV1(
       None,
       None,
       None,
+      None,
       None
```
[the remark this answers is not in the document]
marcus-girolneto: It's already updated

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/programs/Rollback.scala
```diff
@@ -49,14 +51,37 @@ object Rollback {
...
+                  logger.warn(s"Global snapshot ordinal ${value.ordinal} not found, using current global snapshot") >>
+                    globalSnapshot.pure[F]
+              }
+              .handleErrorWith { e =>
+                logger.error(e)(
+                  s"Could not fetch global snapshot ordinal: ${value.ordinal}, starting from latest global snapshot"
+                ) >> globalSnapshot.pure[F]
+              }
+        case None => globalSnapshot.pure[F]
+      }
```
scasplte2, line 83: slightly refactored to avoid some pattern matching but overall still pretty complex 

```
      globalSnapshotStartingPoint: Hashed[GlobalIncrementalSnapshot] <- lastGlobalSnapshotWithCurrencySnapshot
        .fold(globalSnapshot.pure[F]) { value =>
          (logger.info(
            s"Using global snapshot at ordinal ${value.ordinal} as the starting point, which includes the last currency snapshot for metagraph ${identifier.show}"
          ) >> globalL0Service.pullGlobalSnapshot(value.ordinal))
            .flatTap(_.fold(
              logger.warn(s"Global snapshot ordinal ${value.ordinal} not found, using current global snapshot")
            )(_ => Applicative[F].unit))
            .map(_.getOrElse(globalSnapshot))
            .handleErrorWith {
              logger.error(_)(
                s"Could not fetch global snapshot ordinal: ${value.ordinal}, starting from latest global snapshot"
              ) >> globalSnapshot.pure[F]
            }
        }
```

scasplte2 (approved):
worth a try :tada:

modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshotInfo.scala
marcus-girolneto: 🤔 We probably need to think of sth to fix this

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
marcus-girolneto: Hmm, we definitely will use this logic a lot to backward compatibility

PR #1228, opened 2025-06-05
feat: price oracle support

effects-ai:


modules/node-shared/src/main/resources/application.conf
```diff
@@ -174,3 +174,10 @@ delegated-staking {
     dev: 3
   }
 }
+
+price-oracle {
+  # null means all currency ids are allowed
+  allowed-currency-ids = null
```
ryle-goehausen: I would change this to metagraph-id for better consistency.

ryle-goehausen (commented):
It looks like this is missing the logic for averaging for fixed time windows, which is enough to test, but I don't know if we need to merge it in yet ? Unless there's more schema updates related to the testnet code.

Those changes will require us to change the state again, so maybe best to wait for those?

modules/node-shared/src/main/resources/application.conf
effects-ai: Done

ryle-goehausen (dismissed):

modules/node-shared/src/main/resources/application.conf
```diff
@@ -182,3 +182,10 @@ delegated-staking {
 metagraphs-sync {
  max-unapplied-global-change-ordinals: 100
 }
+
+price-oracle {
+  # null means all metagraph ids are allowed
+  allowed-metagraph-ids = null
```
ryle-goehausen, line 188: I think this may already have come up before -- but just to confirm, we need a way to handle verifying the correct metagraph id functionality works on the testnet ticket. Since we need to wait for the other API endpoints to be merged to do the full deployment, I think this PR is okay to merge. But please account for this in the JIRA ticket, since we need to verify allowed metagraph ids is tested on testnet -- if there's already another way to override this, then that's fine as part of a custom deployment -- but we'll need a separate PR for the mainnet ids and that should be captured somewhere (either in ticket or separate PR)

ryle-goehausen (approved):

PR #1229, opened 2025-06-05
test: include metagraph test in docker

ryle-goehausen:
note: this worked for me locally a few times, but when you run it with 3*every service layer, it has unpredictable behavior in the tests, especially regarding the join process. As far as I can tell, this doesn't break the prior work done, so 3GL0/ 3 dag-l1 seems to still work, but it doesn't behave predictably with all of them running on local server. I think this has actually reproduced the flakiness problem of the E2E suite locally at least though. I believe the error / flakiness is actually primarily caused by the join process, as varying the parameters there seems to produce different results (the initial join delay being the most important one.)

Either way, at one point the tests passed and all services joined correctly, so while I need to keep debugging it should be okay. Most importantly, when supplying a `--metagraph` parameter I set the default service counts as low as possible (1 GL0, 1ML0). And there's parameters for number of nodes. The metagraph was tested on the project template.

This also covers the debugging scripts for mainnet, and the readme updates.


docker/bin/build-production.sh
```diff
@@ -9,7 +9,7 @@ mkdir -p ./docker/jars/
 # Duplicate copy overwrites with dirty version if only compiling one module
 # Order is deliberate here for reruns
 
-for module in "dag-l0" "dag-l1" "keytool" "wallet"
+for module in "dag-l0" "gl1" "keytool" "wallet"
```
scasplte2, line 12: If we're already calling one layer `dag-l0` why introduce a new label of `gl1` instead of `dag-l1`? I can understand updating `ml0, ml1, dl1` that I have no problem with but the `dag-l0, dag-l1` is more consistent to the code base modules and I would really love to all just settle on a convention for labeling these dang applications.

docker/bin/set-env.sh
```diff
@@ -97,7 +161,7 @@ for arg in "$@"; do
     --dag-l0-port-prefix=*)
       export DAG_L0_PORT_PREFIX="${arg#*=}"
       ;;
-    --dag-l1-port-prefix=*)
+    --gl1-port-prefix=*)
```
scasplte2, line 164: here as well

docker/bin/set-env.sh
```diff
@@ -122,3 +240,64 @@ exit_func() {
...
+echo "RELEASE_TAG: $RELEASE_TAG"
+
+
+# If a specific metagraph is provided, set sensible defaults
+if [ -n "$METAGRAPH" ]; then
+  if [ -z "$NUM_GL0_NODES" ]; then
+    export NUM_GL0_NODES=1
+  fi
+  if [ -z "$NUM_GL1_NODES" ]; then
+    export NUM_GL1_NODES=3
```
scasplte2, line 254: this one is internal so :man_shrugging:

docker/docker-compose.local-bridge.yaml
```diff
@@ -1,8 +1,8 @@
 services:
-  global-l0:
+  gl0:
     networks:
       tessellation-network:
-  dag-l1:
+  gl1:
```
scasplte2, line 5: if internal I would keep as gl0 and gl1 but if a metagraph dev might see then dag-l0, dag-l1 would be my preference

scasplte2 (dismissed):
I would like to run locally to confirm in my env but otherwise looks good to me

docker/bin/build-production.sh
AlexBrandes: Yeah, I agree here. Greptile actually caught some of the variations too. 

I would suggest: dag-l0, dag-l1, ml0, cl1, dl1

ryle-goehausen: So there's two things going on here, one, I think ^ the diff above is a legitimate mistake, since the modules all need to be named to match the corresponding folder structure, irrespective of what we decide to call the service identifier.

Second question, for the metagraph services, we have a convention of "L0/L1/data-l1" which is repeated, which could either be interpreted as ML0/ML1/DL1 or as CL0/CL1/DL1 since we're using both naming conventions.

I do like the syntax for the metagraphs where we can just use a 3 letter identifier, since it makes service referencing extremely easy and consistent. So if we choose ML0/CL1/DL1 as the consistent metagraph service identifiers for metagraphs, it makes sense to use a similar notation for our main service.

We're also continuously using GL0 everywhere (especially in our class names,) to reference the module called "dag-l0" which is not great. We have GlobalSnapshotInfo and GlobalIncrementalSnapshot as our main 'outputs' of this service, so the convention there is a little confusing. At least in the dag-l1 it makes sense to use that because it only ever processes dag transactions like a CL1.

So if we choose ML0/CL1/DL1 for the metagraphs, I think it makes sense to use a consistent notation for the dag-l0 / global snapshots. Hence GL0/GL1 -- GL1 is perhaps less canonical however, but D is already taken.

If we want to drop ML0/CL1/DL1 syntax, and expand it to a longer service identifier, like metagraph-l0 and currency-l1 and data-l1, that'd work too, then we could use a different syntax. I just don't like the inconsistency which is why I did the 3 letter service ID.

ryle-goehausen: Also, my personal vote is for the ML0/CL1/DL1 syntax, because it makes the references much shorter when doing scripts, i.e. --num-ml0 or `docker logs ml0` for instance. It's just easier to deal with. But I'd rather have a consistent naming convention, so if we do short names, we should have uniformly short service identifiers. If we do long ones, same.

docker/bin/set-env.sh
ryle-goehausen: The only ones that matter IMO are the module names which cannot be easily changed without refactoring, I think our identifier syntax should be uniform in all of these scripts, whichever method we choose, short or long service references.

docker/bin/build-production.sh
AlexBrandes: I think I can get behind the switch to "gL1" if we do gL0, mL0, cL1, dL1. It at least logically makes sense as the only global L1 layer.

ryle-goehausen: Okay, I'll try to adapt it for that. I realize your suggestion of CL1 makes a lot more sense than ML1, I didn't even think of that at the time of working on this. I think there's a separate concern for our 'public facing' docker compose as well. I had to litter this docker compose with a LOT of non-sense variables as part of the debugging process. So maybe it's better here to separate the 'production' or public facing docker compose in the future with longer names.

I've seen both conventions in other projects, i.e. long names and short, depending on use case. But maybe if we have a non-developer version, longer names would make sense for someone running this on a larger system with many more containers.

AlexBrandes (commented):
I've run this a few times and the tests always run but I also always get failed tests for different reasons. It seems close but I'm not sure this is totally usable yet.

ryle-goehausen:
@AlexBrandes are the failures when you run it with all services involved ? Or just L0/L1? I already spotted an error in some environment variables I'll try to retest this after the other metrics stuff.

docker/docker-compose.test.yaml
```diff
@@ -1,17 +1,17 @@
 services:
-  global-l0:
+  gl0:
     # merge in an extra network just for tests
     networks:
       tessellation_common:
         aliases:
-          - global-l0{COxNTAINER_NAME_SUFFIX:-}
+          - gl0{COxNTAINER_NAME_SUFFIX:-}
```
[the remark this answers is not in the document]
ryle-goehausen: This is a good one, I think this might have been breaking the tests. It's very obviously wrong, not sure when I committed this.

ryle-goehausen:
Branch updated to latest working version. I tested this with test --use-test-metagraph - all the tests were run on mainnet version which is nearby. This branch is just a clone of folder from other branch with a merge to develop. It shouldn't affect any core code though just the /docker folder. The lint failure is due to the merge commit so I'm going to ignore that.

ryle-goehausen:
I pushed another commit on here to address issue James found, the --publish flag was not being supplied by default. I set it to only publish by default in the event that the template metagraph jars were not yet compiled -- which we could potentially adjust in future as needed.

James got this test to pass with the test metagraph, but there is still a scenario where it fails due to low memory -- I think a followup PR to add more memory warnings / docker limits might help prevent a SIGKILL in the event of insufficient host memory.

scasplte2 (approved):
lgtm

PR #1232, opened 2025-06-11
feat: gl0 state transition function for monthly price update

effects-ai:


modules/node-shared/src/main/resources/application.conf
```diff
@@ -184,4 +184,11 @@ price-oracle {
   allowed-metagraph-ids = null
   # allowed-metagraph-ids = ["DAG2deV7SKjC2xhTzmpX2ASmvEPzuzkSjXPRcSDy"]
   min-epochs-between-updates = 0
+  update-window-epochs = 733395
```
ryle-goehausen: Where is this number coming from ? It doesn't appear to have an asOfEpoch ? Is this bearing a common derivation from the other rewards epochs per year yardstick ? We should rely on that.

modules/node-shared/src/main/resources/application.conf
```diff
@@ -184,4 +184,11 @@ price-oracle {
...
+  initial-prices = {
+    "DAG::USD": 0.0400638
+  }
+  initial-upcoming-prices = {
+    "DAG::USD": 0.0400638
```
ryle-goehausen: Why not rely on these values being derived from the existing rewards config? Same question as above, there should already be a way to parse these with an asOfEpoch

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/config/types.scala
```diff
@@ -233,5 +234,11 @@ object types {
...
   case class ValidationErrorStorageConfig(maxSize: PosInt)
 
-  case class PriceOracleConfig(allowedMetagraphIds: Option[List[Address]], minEpochsBetweenUpdates: NonNegLong)
+  case class PriceOracleConfig(
+    allowedMetagraphIds: Option[List[Address]],
+    minEpochsBetweenUpdates: NonNegLong,
+    updateWindowEpochs: NonNegLong,
+    initialPrices: Map[TokenPair, BigDecimal],
+    initialUpcomingPrices: Map[TokenPair, BigDecimal]
+  )
```
ryle-goehausen, line 242: This configuration needs some unification with the rewards configuration -- in terms of the values described above.

Additionally, the window period is missing an asOfEpoch -- we cannot change the window size except as of a specified epoch.

ryle-goehausen: See the Rewards definitions for examples, we also need different values on different environments. So we need asOfEpoch for any new values (a list,) and an environment it applies to.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/priceOracle/PriceStateUpdater.scala
```diff
@@ -0,0 +1,116 @@
...
+                    for {
+                      initialCurrentPrice <- initialCurrentPricingUpdate(tokenPair, config)
+                      initialUpcomingPrice <- initialUpcomingPricingUpdate(tokenPair, config)
+                    } yield
+                      PriceRecord(
+                        currentPrice = initialCurrentPrice,
+                        upcomingPrice = initialUpcomingPrice,
+                        currentSum = aggregatedPricingUpdate,
+                        currentNumEvents = PosInt(1),
+                        nextWindowChange = epochProgress |+| EpochProgress(config.updateWindowEpochs),
```
ryle-goehausen: We need a dependence on the asOfEpoch value associated with the config, unless elsewhere we modify the config, its generally speaking a dynamic value defined in terms of a function, i.e. `configFunc(currentEpochProgress)`

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/priceOracle/PriceStateUpdater.scala
```diff
@@ -0,0 +1,116 @@
...
+                      ).pure[F]
+                    } else {
+                      val newCurrentNumEvents = lastPriceRecord.currentNumEvents |+| PosInt(1)
+                      for {
+                        newCurrentSum <- aggregateUpdates(
+                          NonEmptyList.of(lastPriceRecord.currentSum, aggregatedPricingUpdate),
+                          newCurrentNumEvents.some
+                        )
+                      } yield
+                        lastPriceRecord.copy(currentSum = newCurrentSum, currentNumEvents = newCurrentNumEvents, updatedAt = epochProgress)
```
ryle-goehausen: Use focus/replace pattern instead

ryle-goehausen (changes requested):
Please unify this PR with the prior one, since in standup meeting we discussed only doing this as a single feature PR. **edit nevermind, this is PR to other PR branch.

modules/node-shared/src/main/resources/application.conf
effects-ai: I created `EmissionConfigEntry.epochsPerMonth()` method that derives the value from the rewards epochs per year.

modules/node-shared/src/main/resources/application.conf
effects-ai: It get's the initial prices from the existing rewards config now.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/config/types.scala
effects-ai: see the comment above

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/priceOracle/PriceStateUpdater.scala
effects-ai: it uses DelegatedRewardsConfig to get the update window.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/priceOracle/PriceStateUpdater.scala
effects-ai: Done

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/config/MainnetRewardsConfig.scala
```diff
@@ -0,0 +1,101 @@
...
+import io.constellationnetwork.schema.NonNegFraction
+import io.constellationnetwork.schema.address.Address
+import io.constellationnetwork.schema.balance.Amount
+import io.constellationnetwork.schema.epoch.EpochProgress
+import io.constellationnetwork.schema.transaction.TransactionAmount
+
+import eu.timepit.refined.auto._
+import eu.timepit.refined.types.numeric.PosLong
+
+object MainnetRewardsConfig {
```
ryle-goehausen, line 19: I would not change this entire class in a new type as part of this PR, as it makes it difficult to tell which values have actually changed here, but this is more of a stylistic comment since I don't think anything has changed here? Can you verify which values have changed here?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/config/types.scala
ryle-goehausen: I think this is still wrong, the config value can change dynamically, which makes it a function of epochProgress. So if we change this in the future, it'll switch as of a particular epochProgress, meaning the config is a runtime function to select it. You could select the environment here, but it should still take the current epoch progress as a parameter to return the window value to be passed to the validator. As far as I can tell, this config is in the shared config on the base PR, which means it's rendered at the beginning / start.

ryle-goehausen (changes requested):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/config/MainnetRewardsConfig.scala
effects-ai: I moved `MainnetRewardsConfig` and `DefaultDelegatedRewardsConfigProvider` from the `dag-l0` module to the `node-shared` module. This was necessary because `GlobalSnapshotAcceptanceManager` needed access to DefaultDelegatedRewardsConfigProvider during its creation, but it wasn't accessible in its original location. The functionality remains exactly the same - this was purely a module reorganization to improve accessibility.

effects-ai:
@greptileai review it again

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/config/types.scala
effects-ai: The current implementation uses `EmissionConfigEntry` to determine the window value, with the ability to select different values based on the environment. Do we need changes in `EmissionConfigEntry` and/or `DelegatedRewardsConfig`?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/config/MainnetRewardsConfig.scala
ryle-goehausen: Oh, that's 100% fine then, will resolve, didn't realize.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/config/types.scala
ryle-goehausen: Need to think about this, I think there may still be an issue same as I remarked above.

`    asOfEpoch: EpochProgress,`

EmissionConfigEntry uses this asOfEpoch dynamically. So during the download / historical traverse, it will select the exact emission config as is appropriate to the epochProgress. When you do 

```
    def getEmissionConfig: F[EmissionConfigEntry] =
      delegatedRewardsConfigProvider
        .getConfig()
        .emissionConfig
        .get(environment)
```
Are you choosing a particular emission config entry based on the asOfEpoch at node startup? Because the validator here needs to support multiple options. It needs to validate historical based on one emission config, and live based on another, and if the live version updates partway through, it needs to adjust the config based on that.

So for example, in Marcus' other PR, we have a validator which has never been used, and which we don't ever expect to change. But these rewards configs / yardsticks we 100% expect to change over time (as we already have several for other values.) So it makes sense to capture that as a function of epochProgress => config

effects-ai:
@greptileai review it again

ryle-goehausen (approved):

PR #1234, opened 2025-06-13
feat: histogram and metrics middleware

ryle-goehausen:
added jol and custom middleware, new histogram method

JOL is a better way to measure in memory object sizes than relying on serialization, if we need more granularity than the external monitoring dashboard. It's not used in this PR just imported and defined. I'll make separate PR with a CL_METRICS_TRACE flag to prevent using it unless explicitly enabled in case we need to investigate in the future in more detail for objects in flight (separate from the state/info).

The current timer using micrometer doesn't integrate properly with prometheus / grafana. We're not capturing default bucket level information in prometheus as is normally intended by micrometer, and we're only capturing max/count/sum with no way to aggregate by buckets as far as I can tell (I don't see them being published in the right format in the /metrics endpoint at least.)

There's another branch ryle/mainnet-debug-mgdocker where I'm actually testing / using these, I've been adjusting the histogram buckets based on the routes I applied them to as part of testing, it appears to at least be returning somewhat correct information.

I tried to use the http4s-prometheus-metrics, that has a serious problem with versions and the way that we're implementing micrometer metrics. There's also metrics4s and others, but all of them would require us manually merging the registry due to the way we've created a custom route to serve the micrometer metrics. I tried a few libraries but they were all more effort to integrate than it was to just recreate directly. This custom middleware also very importantly captures the actual origin IPs. It may need some more headers though, the actual ip is definitely populated though, and I saw request labels being formatted correctly.

I doubt I've perfectly captured the regex for normalizing routes, but its at least mostly correct so far in tests. Also I had to bypass Refined due to compiler error on Seq[] and it doesn't seem right to register every string metric as an individual val since that's almost worse?

No functional changes are made as part of this PR, I am trying to merge code from the debug branch back so please review for correctness and I'll test on a separate branch.

### DIFF
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/http/p2p/middlewares/MetricsMiddleware.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/http/p2p/middlewares/MetricsMiddleware.scala
@@ -1,0 +1,104 @@
+package io.constellationnetwork.node.shared.http.p2p.middlewares
+
+import cats.data.Kleisli
+import cats.effect.kernel.Async
+import cats.syntax.all._
+import eu.timepit.refined.auto._
+import eu.timepit.refined.api.Refined
+import io.constellationnetwork.node.shared.infrastructure.metrics.Metrics
+import org.http4s.HttpRoutes
+import org.http4s.headers.`X-Forwarded-For`
+import scala.concurrent.duration.FiniteDuration
+import java.util.concurrent.TimeUnit
+
+object MetricsMiddleware {
+
+  def apply[F[_]: Async: Metrics](): HttpRoutes[F] => HttpRoutes[F] = { routes =>
+    Kleisli { req =>
+      val startTime = System.nanoTime()
+      routes(req).semiflatMap { response =>
+        val scriptName = req.scriptName.renderString
+        val endTime = System.nanoTime()
+        val duration = FiniteDuration(endTime - startTime, TimeUnit.NANOSECONDS)
+        // Extract route path and normalize it for metric naming
+        val routePath = normalizeRoutePath(req.pathInfo.renderString)
+        val actualIp = req.remote.map(_.host.toString).getOrElse("unknown")
+        val forwardedIp = req.headers
+          .get[`X-Forwarded-For`]
+          .map(_.values.head.toString.split(",").head.trim)
+          .getOrElse("none")
+
+        // Cannot compile time infer type for Seq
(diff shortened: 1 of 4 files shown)


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/http/p2p/middlewares/MetricsMiddleware.scala
```diff
@@ -0,0 +1,104 @@
...
+import org.http4s.HttpRoutes
+import org.http4s.headers.`X-Forwarded-For`
+import scala.concurrent.duration.FiniteDuration
+import java.util.concurrent.TimeUnit
+
+object MetricsMiddleware {
+
+  def apply[F[_]: Async: Metrics](): HttpRoutes[F] => HttpRoutes[F] = { routes =>
+    Kleisli { req =>
+      val startTime = System.nanoTime()
```
scasplte2: Should we use `Clock[F]` instead of `System` to grab the current instant?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/http/p2p/middlewares/MetricsMiddleware.scala
```diff
@@ -0,0 +1,104 @@
+package io.constellationnetwork.node.shared.http.p2p.middlewares
+
+import cats.data.Kleisli
+import cats.effect.kernel.Async
```
scasplte2: using the `cats.effect.kernel.Async` sometimes causes compilation issues (at least in intellij). Can we switch this to `cats.effect.Async` instead?

scasplte2 (commented):
Seems like a reasonable addition to me

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/http/p2p/middlewares/MetricsMiddleware.scala
ryle-goehausen: fixed

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/http/p2p/middlewares/MetricsMiddleware.scala
ryle-goehausen: fixed

ryle-goehausen:
Updated to restrict the number of metrics we collect heavily. Labels removed from histogram metrics apart from route.

marcus-girolneto (approved):
LGTM

PR #1235, opened 2025-06-13
feat: adding token lock limits for the currency layer

marcus-girolneto:
### Changes
+ Adding an overridable function on CL1 to set the token lock limits


ryle-goehausen (commented):
Will this break historical validation ? Unclear if this is being used so far and if we want to gate it behind an epochProgress feature flag, and/or environment flag similar to rewards config.

marcus-girolneto:
> Will this break historical validation ? Unclear if this is being used so far and if we want to gate it behind an epochProgress feature flag, and/or environment flag similar to rewards config.

@ryle-goehausen No, we haven’t used tokenLock at all so far on the currency layer — maybe only on testnet, but if that’s the case, we can roll it back. So I think it’s safe to proceed without using the ordinal here.

By default, we won’t have any limits on the metagraph. I’ll configure this later on the AMM metagraph to apply the limits that João provided.

ryle-goehausen (approved):

PR #1237, opened 2025-06-16
fix: adding retry mechanism to global snapshot download on currency manager

marcus-girolneto:


ryle-goehausen (commented):
@IPadawans does this contain the fix that, within the GL0 accept/validate recreateFn, it passes the global snapshot storage ?

ryle-goehausen (commented):
@IPadawans does this contain the fix that, within the GL0 accept/validate recreateFn, it passes the global snapshot storage ?

ryle-goehausen (commented):
@IPadawans does this contain the fix that, within the GL0 accept/validate recreateFn, it passes the global snapshot storage ?

marcus-girolneto:
> @IPadawans does this contain the fix that, within the GL0 accept/validate recreateFn, it passes the global snapshot storage ?

This retains the same behavior as before — and as currently used on mainnet — where we pass lastNGlobalSnapshots and a network fetch function to the currency snapshot acceptance manager.

The issue with the current version is that if it fails to fetch a global snapshot that isn’t in the cache, it simply terminates the process without retrying.

This PR adds a retry mechanism to address that issue. It preserves the existing logic (until version 3.4.0 is deployed), but now attempts to re-fetch the global snapshot instead of failing immediately.

ryle-goehausen (approved):
Ok yeah, since we already have that behavior I think this okay to go irrespective of the storage optimization.

PR #1241, opened 2025-06-18
fix: multiple delegated stakes for the same node id and address combination

effects-ai:


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/delegatedStake/UpdateDelegatedStakeAcceptanceManager.scala
```diff
@@ -35,6 +33,24 @@ trait UpdateDelegatedStakeAcceptanceManager[F[_]] {
 }
 
 object UpdateDelegatedStakeAcceptanceManager {
+
+  private type CreateDelegatedStakeAcceptanceResult = (
```
ryle-goehausen: I know we've used type alias patterns in other areas, but this might be easier to read if we instead used a case class here

ryle-goehausen (commented):
I think this is correct but you should update the node js test for delegated stake to attempt duplicate submission at same time.

ryle-goehausen (commented):
Stale PR, as discussed, needs merge to main branch and address comments before merge. Otherwise seems fine.

effects-ai:
> I think this is correct but you should update the node js test for delegated stake to attempt duplicate submission at same time.

Done

ryle-goehausen (approved):

PR #1242, opened 2025-06-20
feat: add node shared access for implementations of currencyl0app

Serpion-ua:


modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/CurrencyL0App.scala
```diff
@@ -38,6 +38,9 @@ import pureconfig.module.enumeratum._
 trait OverridableL0 extends TessellationIOApp[Run] {
   def dataApplication: Option[Resource[IO, BaseDataApplicationL0Service[IO]]] = None
 
+  def dataApplication(nodeShared: NodeShared[IO, Run]): Option[Resource[IO, BaseDataApplicationL0Service[IO]]] =
```
effects-ai, line 41: `nodeShared` is not used anywhere,  is it supposed to be used in metagraphs?

Serpion-ua: yes

effects-ai: I see. LGTM

ryle-goehausen (approved):

ryle-goehausen:
@Serpion-ua this is a backwards compatibility breaking change, but looks correct, lets discuss more in meeting when to release. ** Edit nevermind -- my mistake, this is not breaking since it's just overloading.

AlexBrandes (dismissed):
Since this is a breaking change - let's discuss whether this is the best option or if there are other ways to solve this.

ryle-goehausen:
@IPadawans did we already implement an alternative to this PR ? If so we can close, but it shouldn't harm to do this.

effects-ai (approved):
👍

PR #1243, opened 2025-06-23
chore: lock wallets

effects-ai:


ryle-goehausen (dismissed):
Reminder, this is using the wrong mainnet configuration bucket for the one time reward (the old config)

effects-ai:
@greptileai review it again

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/rewards/GlobalDelegatedRewardsDistributor.scala
```diff
@@ -466,6 +476,8 @@ object GlobalDelegatedRewardsDistributor {
             lastSnapshotContext,
             delegatedStakeDiffs.acceptedCreates
           )
+        oneTimeRewardsTxs = SortedSet.from(oneTimeRewards.map(otr => RewardTransaction(otr.address, otr.amount)))
+        nodeOperatorRewardsTxs = nodeOperatorRewards ++ oneTimeRewardsTxs
```
scasplte2, line 480: this is a little concerning to me since the one time rewards don't necessarily go to operators, they can be to any address. There are several variables here of type `SorteedSet[RewardsTransaction]` so I would suggest to just add another or add this to `reservedRewards` since these mintings are closer to that kind of reward.

Ultimately, if this just needs to get done though, this should be functional.

scasplte2 (approved):
minor comment but should be functional

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/rewards/GlobalDelegatedRewardsDistributor.scala
ryle-goehausen: I agree, but I'm also unclear how this will impact the block explorer stuff ? I don't think we need an explicit event for this, but this might be okay. Worth checking with @gclaramunt to see if this'll affect any endpoints.

ryle-goehausen (commented):
Check with Gabriel first to clarify if this breaks anything, otherwise agree with james.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/rewards/GlobalDelegatedRewardsDistributor.scala
gclaramunt: if it doesn't add a new collection to `io.constellationnetwork.schema.GlobalIncrementalSnapshot` it should be fine. If  I understand correctly, this rewards will come in `rewards: SortedSet[RewardTransaction],` right? I've recently changed snapshot streaming to accept multiple rewards for the same address in the same snapshot, so it should accept it.

PR #1247, opened 2025-06-26
feat: exit on fork on follower

ryle-goehausen:


scasplte2:
What is the motivation for this change? Should a node really restart if it becomes a "follower"? What even is a follower?

scasplte2:
> What is the motivation for this change? Should a node really restart if it becomes a "follower"? What even is a follower?

Ryle and I discussed that this work enables a feature flag which would allow nodes to more readily identify when they are in a minority fork and then rejoin the majority.

ryle-goehausen:
@scasplte2 to add more context to this PR for others, the intention of this is to re-create some of the behavior of nodectl but for the docker compose runner. This is not intended to be run unless someone manually sets this flag, in which case, it will refuse to participate in a cluster that is missing a specified peer. Intention here is an alternative fork detection process, as the current fork detection is failing to re-discover other clusters.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/domain/snapshot/programs/Download.scala
```diff
@@ -398,6 +399,10 @@ object Download {
     def fetchSnapshot(hash: Option[Hash], ordinal: SnapshotOrdinal)(implicit hasher: Hasher[F]): F[Signed[GlobalIncrementalSnapshot]] =
       clusterStorage.getResponsivePeers
         .map(NodeState.ready)
+        .map { peers =>
+          ForkDetect.exitOnCheck("CL_EXIT_ON_FOLLOWER_DOWNLOAD", () => peers.map(_.id))
+          peers
+        }
```
scasplte2: I think this would rather be
```scala
    .flatTap { peers =>
      ForkDetect.exitOnCheck("CL_EXIT_ON_FOLLOWER_DOWNLOAD", () => peers.map(_.id))
    }
 ```

File: modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotConsensusStateAdvancer.scala [hunk 2]
```diff
@@ -134,6 +135,11 @@ object GlobalSnapshotConsensusStateAdvancer {
                             hash <- HasherSelector[F].forOrdinal(artifact.ordinal)(implicit hasher => artifact.hash)
                             effect = gossip.spread(ConsensusPeerDeclaration(state.key, Proposal(hash, facilitatorsHash))) *>
                               gossip.spreadCommon(ConsensusArtifact(state.key, artifact))
+                            facilitators = state.facilitators.value
+                            _ = {
+                              ForkDetect.exitOnCheck("CL_EXIT_ON_FOLLOWER_ADVANCER", () => facilitators.toSet)
+                              ()
+                            }
```
scasplte2: simplify to
```scala
_ <- ForkDetect.exitOnCheck("CL_EXIT_ON_FOLLOWER_ADVANCER", () => facilitators.toSet)
```

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/fork/ForkDetect.scala
```diff
@@ -5,11 +5,37 @@ import cats.syntax.contravariantSemigroupal._
...
+          val hasFollowerPeer = facilitators().contains(peerId)
+          if (!hasFollowerPeer) {
+            // Println to avoid logger shutdown
+            println(s"Exit due to missing follower peer on $flag")
+            System.exit(1)
+          }
+        case _ =>
+      }
+    }
+
```
scasplte2: I know it is more cumbersome but these are side-effecting and should be `F[_]` wrapped
```scala
  def hasFlag[F[_]: Sync](flag: String): F[Boolean] =
    Sync[F].delay(sys.env.get(flag).contains("true"))

  def exitOnFeature[F[_]: Sync](flag: String): F[Unit] =
    hasFlag(flag).flatMap(_.guard[F].whenA(
      Slf4jLogger.getLogger[F].error(s"Exit due to feature flag on $flag") *>
      Sync[F].delay(System.exit(1))
    ))

  def exitOnCheck[F[_]: Sync](flag: String, facilitators: () => Set[PeerId]): F[Unit] =
    hasFlag(flag).flatMap(_.guard[F].whenA(
      Sync[F].delay(sys.env.get("CL_FOLLOWER_ID")).flatMap(_.traverse_ { id =>
        for {
          peerId <- PeerId(Hex(id)).pure[F]
          peers <- Sync[F].delay(facilitators())
          hasFollowerPeer = peers.contains(peerId)
          _ <- (!hasFollowerPeer).guard[F].whenA(
            Slf4jLogger.getLogger[F].error(s"Exit due to missing follower peer on $flag") *>
            Sync[F].delay(System.exit(1))
          )
        } yield ()
      })
    ))
 ````

File: modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/GossipRoundRunner.scala [hunk 4]
```diff
@@ -68,6 +69,7 @@ object GossipRoundRunner {
           for {
             _ <- Temporal[F].sleep(cfg.interval)
             allPeers <- clusterStorage.getResponsivePeers
+            _ = ForkDetect.exitOnCheck("CL_EXIT_ON_FOLLOWER_GOSSIP", () => allPeers.map(_.id))
```
scasplte2: just 
```scala
_ <- ForkDetect.exitOnCheck("CL_EXIT_ON_FOLLOWER_GOSSIP", () => allPeers.map(_.id))
```

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ConsensusStateUpdater.scala
```diff
@@ -225,6 +226,7 @@ object ConsensusStateUpdater {
       val isForked = majorityObservationHash =!= ownObservationHash
 
       if (isForked) {
+        ForkDetect.exitOnFeature("CL_EXIT_ON_FORK")
```
scasplte2: chain to after the fork recovery
```scala
  pickMajority(observations.values.toList).traverse { majorityObservationHash =>
    val isForked = majorityObservationHash =!= ownObservationHash

    if (isForked) {
      val majorityForkPeers = observations.collect {
        case (peerId, observationHash) if observationHash === majorityObservationHash => peerId
      }.toList

      val forkRecovery = Slf4jLogger
        .getLogger[F]
        .warn(s"Different hash observations [$observationName]. This node is in fork") >>
        nodeStorage.setNodeState(NodeState.Leaving) >>
        Temporal[F].sleep(leavingDelay) >>
        nodeStorage.setNodeState(NodeState.Offline) >>
        Temporal[F].sleep(5.seconds) >>
        restartService.signalNodeForkedRestart(majorityForkPeers) >>
        ForkDetect.exitOnFeature[F]("CL_EXIT_ON_FORK")

      Temporal[F].start(forkRecovery).void

    } else Applicative[F].unit
  }.void
  ```

**Review by scasplte2 — CHANGES_REQUESTED** [review]
My preference would be to create side-effecting functions. All of the invocations can be easily updated to use the `F[_]` context

File: modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/fork/ForkDetect.scala (hunk 3 above)
Comment (ryle-goehausen) [reply modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/fork/ForkDetect.scala]: System.exit(1) breaks the logger and the F context because any invocation AFAIK makes any other threads inherently unreliable after it is invoked. Hence the use of println. I can wrap this in an F context for prettiness maybe, but it's guarded already by the feature flag making it redundant. We could have a proper shutdown initiation hook, but given the failures of the 'soft restart' code I just wanted something guarantee to force a panic.

Any invocation of System.exit anywhere will break the entire F chain though, it's not a 'gentle' shutdown. I could see an argument for using F context here to substitute a trait in here, but I'm not sure there's a purpose yet. There is probably a better cats shutdown hook to invoke honestly, but I only was using this for debugging so I haven't looked into best way, and I wasn't sure what was causing the other threads to hang.

Comment (scasplte2) [reply modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/fork/ForkDetect.scala]: Fair point about shutdown within the cats but this isn't a question of prettiness to me. The invocations have to be weird things like 
```
_ = {
   <method>
   ()
}
```
in order to work properly in the type system. This breaks the mental model of chaining operations together with this sudden !!REMEMBER YOU'RE IN THE JVM!! imperative operation of executing two lines back to back. 

Could you include a type parameter in the function definition that would "return" a given type to satisfy the compiler but within the function invocation you just kill the main process? Like `_ = <method[Unit]>` then it just looks like another function call and the method itself can document that it kills the primary thread.

ryle-goehausen: Yeah, so I think your other suggestions actually make perfect sense to keep it in F types for the constructions. The one thing I'm hesistant to do is to use the F logger, because I don't know if it guarantees to log before the shutdown initiates. So I think you're right about this part:       `Sync[F].delay(sys.env.get("CL_FOLLOWER_ID")).flatMap(_.traverse_ { id =>`
because technically that does belong in an F context to prevent the sys.env access from delaying. Lemme refactor it a little bit more to be better cats, but then I think I need to leave the println / other thing as they are now.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ConsensusStateUpdater.scala
ryle-goehausen: This was deliberate, the ForkDetect process had several issues I did not want to debug. I actually can't even get this code to trigger all the time when there is a fork (so it does not always detect this,) and the signal itself is completely broken, so running that portion of the code after doesn't help.

But I think you are right about the order on the nodeState ; because  there's no reason not to indicate we're leaving even if the signal has an issue. So I think it should go here: 

```        Temporal[F].sleep(5.seconds) >>```
Right after that, rather than after the signal.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/fork/ForkDetect.scala
ryle-goehausen: @scasplte2 please re-review, I took some of your suggestions but not all, guard doesn't play well with sync without other helpers, so it's just a traverse and flatMaps, but not for the panic.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ConsensusStateUpdater.scala
ryle-goehausen: updated

scasplte2 (approved):
great! lgtm

PR #1250, opened 2025-06-26
test: override jar or version hash

ryle-goehausen:


scasplte2:
@gclaramunt would appreciate your perspective here since you added the jar hash functionality

gclaramunt:
LGTM, reported hashes are mostly for information purposes and we don't verify them (a node can lie about the hash).

scasplte2 (approved):

PR #1252, opened 2025-06-26
fix: develop test compile error NoOpMetrics

ryle-goehausen:


ryle-goehausen:
Thought the linting error was due to commit message / titles on last merge, was due to test only classpath failure. This is fix for it.

PR #1254, opened 2025-06-26
test: clean configs memory error

ryle-goehausen:


ryle-goehausen:
James hit memory related errors which I can't replicate on 2 other machines. SIGKILL causes node dir to be owned by root, which requires a fallback to sudo -- I removed all memory settings out of caution.

scasplte2 (approved):
tested locally and successfully ran `just test`

PR #1255, opened 2025-06-26
feat: add per-environment price oracle configuration

effects-ai:


modules/node-shared/src/main/resources/application.conf
```diff
@@ -184,8 +184,21 @@ metagraphs-sync {
 }
 
 price-oracle {
-  # null means all metagraph ids are allowed
-  allowed-metagraph-ids = null
-  # allowed-metagraph-ids = ["DAG2deV7SKjC2xhTzmpX2ASmvEPzuzkSjXPRcSDy"]
-  min-epochs-between-updates = 0
+  mainnet: {
+    allowed-metagraph-ids = []
+    min-epochs-between-updates = 50
```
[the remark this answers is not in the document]
effects-ai: Yes, this the intended behavior for production for now.

modules/node-shared/src/main/resources/application.conf
```diff
@@ -184,8 +184,21 @@ metagraphs-sync {
...
+  testnet: {
+    allowed-metagraph-ids = null
+    min-epochs-between-updates = 0
+  }
+  integrationnet: {
+    allowed-metagraph-ids = []
+    min-epochs-between-updates = 50
+  }
+  dev: {
+    # null means all metagraph ids are allowed
```
ryle-goehausen, line 202: I think we need to explicitly enable this setting to verify the functionality works.

ryle-goehausen (commented):
This PR is good if it's required for the initial testnet metagraph setup; in which case, it requires a follow-up PR to add an enforcement check ideally on the metagraph id so we can verify it's ID specific.

modules/node-shared/src/main/resources/application.conf
effects-ai: Done

ryle-goehausen (approved):
E2E tests are likely flakey, please check.

PR #1258, opened 2025-06-30
fix: changing from incomingCurrencySnapshots to currencySnapshots in places that uses state

marcus-girolneto:


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
```diff
@@ -332,13 +333,9 @@ object GlobalSnapshotAcceptanceManager {
...
 
-        activeAllowSpendsFromCurrencySnapshots = incomingCurrencySnapshots.flatMap {
-          case (address, snapshots) =>
-            snapshots.reverse.collectFirst {
-              case Right((_, info)) if info.activeAllowSpends.isDefined =>
-                address -> info.activeAllowSpends.get
-            }
-        }
+        activeAllowSpendsFromCurrencySnapshots = currencySnapshots.map { case (key, value) => (key, value) }
+          .mapFilter(_.toOption.flatMap { case (_, info) => info.activeAllowSpends })
```
scasplte2: the first partial function is not needed here
```
activeAllowSpendsFromCurrencySnapshots = currencySnapshots
  .mapFilter(_.toOption.flatMap { case (_, info) => info.activeAllowSpends })
```

scasplte2 (commented):
what is the motivation for the change here?

marcus-girolneto:
> what is the motivation for the change here?

I'm returning to the older behavior using the right field, that's the reason I've changed this

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
marcus-girolneto: 👍🏼

scasplte2 (approved):
lgtm

PR #1260, opened 2025-07-07
feat: adding globalSnapshotsProcessed artifact and updating metagraphSyncData

marcus-girolneto:


ryle-goehausen (approved):
as far as I can tell, I think this is correct in the sense that the GL0 still maintains the finality balance condition of the 'newer' value, while this only acts as a validator function the CS is consistent. (which is what I was referring to on slack)

ryle-goehausen:
It may be useful to add some comments or docs indicating why this change was made and what it affects.

marcus-girolneto:
> It may be useful to add some comments or docs indicating why this change was made and what it affects.

I'll add a comment and open a 2nd PR

marcus-girolneto:
https://github.com/Constellation-Labs/tessellation/pull/1261

PR #1262, opened 2025-07-08
fix: fixing a bug where the globalSnapshotProcessed is always filled, and improving logs

marcus-girolneto:


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
```diff
@@ -330,12 +330,23 @@ object CurrencySnapshotAcceptanceManager {
...
 
+      globalSyncView = maybeLastGlobalSyncView match {
+        case Some(lastGlobalSyncView) =>
+          if (lastGlobalSyncView.ordinal < lastSyncGlobalSnapshot.ordinal) {
+            GlobalSyncView(lastSyncGlobalSnapshot.ordinal, lastSyncGlobalSnapshot.hash, lastSyncGlobalSnapshot.epochProgress)
+          } else {
+            lastGlobalSyncView
+          }
+        case None => GlobalSyncView(lastSyncGlobalSnapshot.ordinal, lastSyncGlobalSnapshot.hash, lastSyncGlobalSnapshot.epochProgress)
+      }
```
scasplte2: minor suggestion but continuous pattern matching (the boxing / unboxing) may eventually impact performance so it would be my preference for us to use functional approach instead

```scala
globalSyncView = maybeLastGlobalSyncView
  .filter(_.ordinal >= lastSyncGlobalSnapshot.ordinal)
  .getOrElse(
    GlobalSyncView(
      lastSyncGlobalSnapshot.ordinal, 
      lastSyncGlobalSnapshot.hash, 
      lastSyncGlobalSnapshot.epochProgress
    )
  )
```

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
```diff
@@ -614,14 +622,14 @@ object CurrencySnapshotAcceptanceManager {
             case None => (emptySpendActions, emptyProcessedGlobalSnapshots).pure[F]
             case Some(syncDataInfo) =>
               val unappliedOrdinals = syncDataInfo.unappliedGlobalChangeOrdinals
-                .filter(_ < lastSyncGlobalSnapshot.ordinal)
+                .filter(_ <= globalSnapshotViewOrdinal)
```
scasplte2, line 625: what happens if the global snapshot that is synced advances multiple steps between currency snapshots?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
```diff
@@ -1168,34 +1168,36 @@ object GlobalSnapshotAcceptanceManager {
...
-        )
+            updatedFromSpendActions = updateFromSpendActions(
+              updatedFromSnapshots,
+              acceptedSpendActions,
+              currentGlobalOrdinal
+            )
 
-        existingData ++ updatedFromSpendActions
+          } yield existingData ++ updatedFromSpendActions
       }
```
scasplte2: we should stick with the `.map` here. Excessive pattern matching will be detrimental. The logic seems nearly identical.

effects-ai (dismissed):
LGTM

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
```diff
@@ -1204,16 +1206,23 @@ object GlobalSnapshotAcceptanceManager {
...
+              .replace(currentOrdinal)
+              .focus(_.globalEpochProgressLastAcceptedOn)
+              .replace(currentEpochProgress)
+              .focus(_.unappliedGlobalChangeOrdinals)
+              .replace(updatedUnappliedGlobalChangeOrdinals)
+
+            address -> updatedInfo
+          }
+      }
+    } yield SortedMap.from(updatedList)
```
scasplte2: the logic is very difficult for me to follow here. I think the for-comps are hurting you more than helping. My suggestion

```scala
private def updateFromCurrencySnapshots(
  existingData: SortedMap[Address, MetagraphSyncDataInfo],
  snapshots: SortedMap[Address, CurrencySnapshotWithState],
  globalSnapshotsProcessed: Map[Address, List[GlobalSnapshotsProcessed]],
  currentOrdinal: SnapshotOrdinal,
  currentEpochProgress: EpochProgress
): F[SortedMap[Address, MetagraphSyncDataInfo]] = {
  
  snapshots.toList.traverse {
    case (address, _) =>
      val currentInfo = existingData.getOrElse(address, MetagraphSyncDataInfo.empty)
      val metagraphGlobalSnapshotsProcessed =
        globalSnapshotsProcessed.getOrElse(address, List.empty).flatMap(_.ordinals).toSet
      val updatedUnappliedGlobalChangeOrdinals =
        currentInfo.unappliedGlobalChangeOrdinals.diff(metagraphGlobalSnapshotsProcessed)
      
      val updatedInfo = currentInfo
        .focus(_.globalOrdinalLastAcceptedOn)
        .replace(currentOrdinal)
        .focus(_.globalEpochProgressLastAcceptedOn)
        .replace(currentEpochProgress)
        .focus(_.unappliedGlobalChangeOrdinals)
        .replace(updatedUnappliedGlobalChangeOrdinals)
      
      logger.info(
        s"[CURRENCY=$address][GlobalOrdinal=$currentOrdinal] Updated unapplied global change ordinals: $updatedUnappliedGlobalChangeOrdinals"
      ) >> (address -> updatedInfo).pure[F]
  }.map(SortedMap.from)
}
```

scasplte2 (changes requested):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
marcus-girolneto: 👍🏼

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
marcus-girolneto: This is not a problem because we're still getting the snapshots below the lastSyncGlobal ordinal and then we add these snapshots in the field to be marked as processed

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
marcus-girolneto: 👍🏼

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
marcus-girolneto: 👍🏼

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
marcus-girolneto: The goal of this point is not getting the unapplied ordinals directly from the lastGlobalSnapshotInfo, because it's not synchronized and can lead to divergences, so we're getting only the available one that is below or equal the synchronization to get the same information in all nodes

scasplte2: ok, I realized I might be misunderstanding. Thanks for the explanantion

scasplte2 (approved):
lgtm

PR #1263, opened 2025-07-08
fix: fixing wrong metagraph sync data update

marcus-girolneto:


marcus-girolneto:
Previously, we're updating all the metagraphs, including offline, so I've updated to process only the incoming ones

ryle-goehausen (dismissed):

marcus-girolneto:
@greptileai review it again

ryle-goehausen (approved):

PR #1264, opened 2025-07-09
test: fix publish error, bulk submit test, env

ryle-goehausen:
@scasplte2 hit error on publish due to jars missing and publish happening in wrong order, I changed the defaults.

This also adds the bulk submit test I'm using for debugging, but is not invoked yet since it's a very long running test. Separate PRs will add the functionality to introduce a failure middleware into the gossip server.


docker/bin/set-env.sh
```diff
@@ -223,6 +223,9 @@ for arg in "$@"; do
       ;;
     --use-test-metagraph)
       export USE_TEST_METAGRAPH=true
+      ;;    
+    --fail)
+      export SET_FAILURE_BREAKPOINT_TIME=true
       ;;
```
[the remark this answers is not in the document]
ryle-goehausen: I think this test should be more generic, since it's the only example

docker/bin/compose-runner.sh
```diff
@@ -198,6 +201,15 @@ show_time "Cluster became healthy"
...
 
+# Long running debug test if needed
+# echo "------------------------------------------------"
+# echo "Running bulk submit test"
+# echo "------------------------------------------------"
+
+# cd $PROJECT_ROOT/.github/action_scripts/send_transactions
+# node bulk-submit-test.js $DAG_L0_PORT_PREFIX $DAG_L1_PORT_PREFIX
+# show_time "Bulk submit test completed"
+
```
scasplte2, line 212: could use an env var to run or skip?

scasplte2 (approved):
lgtm

docker/bin/compose-runner.sh
ryle-goehausen: I'll try to fix this after, I need to add limiters to the test so it reproduces the issue in a short enough period I think. It shouldn't be run at all right now unless for debugging purposes, as it runs for like 10 minutes constantly spamming transactions. Ideally it would only run long enough to capture the problem and demonstrate it's fixed.

PR #1268, opened 2025-07-09
feat: enable dag-l0 request / client metrics

ryle-goehausen:
I tested this on mainnet to verify it's not adding too many values to the /metrics endpoints. This enables the response metrics and then also the client request metrics.


scasplte2 (approved):
lgtm, my only concern is if the middleware could cause a performance degradation? Is there any reasonable way to load test before and after this change?

PR #1272, opened 2025-07-11
feat: adding force global sync event

marcus-girolneto:


marcus-girolneto:
This is a "blank" event that will be used only to trigger EventTrigger consensus. So in this case, it basically creates a new event type extending CurrencySnapshotEvent and then enqueues this event according to the conditions of `checkIfShouldForceEventTrigger` function. 
This will trigger the EventType consensus

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/snapshot/currency.scala
```diff
@@ -51,5 +51,14 @@ object currency {
...
   case class GlobalSnapshotSyncEvent(value: Signed[GlobalSnapshotSync]) extends CurrencySnapshotEvent
 
+  /** Represents a wrapper event used to force the synchronization of a global snapshot.
+    *
+    * @param value
+    *   The signed `GlobalSnapshotSync` message that must be preserved as a field to allow hashing of the full event content. This hash is
+    *   required during validation in the `CurrencySnapshotEventValidationErrorStorage` class.
+    */
+  @derive(encoder, decoder, eqv)
+  case class ForceGlobalSyncEvent(value: Signed[GlobalSnapshotSync]) extends CurrencySnapshotEvent
```
scasplte2: Is this for testing or something that you envision continuing to use? If testing then I have no concerns but if this is planned to stick around then my preference would be to figure out why the `Signed[GlobalSnapshotSync]` field is being required. It seems fragile and cumbersome that an error class would dictate the structure of this case class.

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/StateChannel.scala
```diff
@@ -54,6 +55,11 @@ object StateChannel {
           case None => Applicative[F].unit
         }
 
+      def checkIfShouldForceEventTrigger(snapshot: GlobalIncrementalSnapshot, currencyId: Address): Boolean =
+        snapshot.spendActions.fold(false) { spendActions =>
```
scasplte2, line 59: we're forcing the entire global snapshot sync but only querying into the `spendActions`?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/snapshot/currency.scala
marcus-girolneto: This field is required because it’s used in the hash calculation within `CurrencySnapshotEventValidationErrorStorage.`
We’re explicitly forcing the event trigger based on global snapshot sync conditions, as determined by the `checkIfShouldForceEventTrigger` function.

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/StateChannel.scala
marcus-girolneto: This class determines whether an event trigger should be forced. Initially, it’s used to trigger events only when there are SpendActions defined for the current metagraph. This behavior was established by Ryle.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/snapshot/currency.scala
scasplte2: I see that the the error storage class uses
```scala
      case ForceGlobalSyncEvent(sync)           => sync.hash.map(List(_))
```
but if the `value: Signed[GlobalSnapshotSync]` is only there to satisfy that why not just remove the value from the case class and make
```scala
      case ForceGlobalSyncEvent(sync)           => List(Hash("00"))
```
or something like that?

marcus-girolneto: Yeah 👍🏼

scasplte2 (approved):
tests passed previously so I presume they will once again. LGTM!

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/snapshot/currency.scala
ryle-goehausen: So if this is just for passing around the information about what data triggered the event, this is 100% fine, but are you certain this will NOT embed this information in the CIS ? It's the desired behavior that the event is not included in CIS, which I think ? Is what will happen since we're not explicitly encoding it. As long as this is only passed around internally, it should be fine I think? It may cause issue if it gets passed during the pooling phase though. Ideally the pooling operation on events doesn't include this information when sending it to peers, because the existence of any sync event will naturally cause the CIS to update internally; meaning its only required as node internal information.

PR #1273, opened 2025-07-11
fix: fixing acceptMetagraphSyncData merging

marcus-girolneto:


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
```diff
@@ -1212,26 +1217,41 @@ object GlobalSnapshotAcceptanceManager {
...
     private def updateFromSpendActions(
       currentData: SortedMap[Address, MetagraphSyncDataInfo],
       spendActions: Map[Address, List[SpendAction]],
       currentOrdinal: SnapshotOrdinal
-    ): SortedMap[Address, MetagraphSyncDataInfo] = {
-      val currencySpendTransactions = extractCurrencySpendTransactions(spendActions)
+    ): F[SortedMap[Address, MetagraphSyncDataInfo]] = {
+      val allCurrencySpendTransactions = spendActions.toList.flatMap {
+        case (address, actions) => extractCurrencySpendTransactions(Map(address -> actions))
+      }
```
scasplte2: This series of operations takes a Map to a List then back to a Map. It seems the same as
```scala
    val allCurrencySpendTransactions = extractCurrencySpendTransactions(spendActions)
 ```

scasplte2 (changes requested):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala
scasplte2: Or do you need to separately process each entry for some reason?

marcus-girolneto: No, you're right, this is a leftover. I've fixed

scasplte2 (approved):

PR #1274, opened 2025-07-13
refactor: restructure delegated stakes price calculation to use latest global context info

effects-ai:


modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/rewards/GlobalDelegatedRewardsDistributor.scala
```diff
@@ -239,12 +241,13 @@ object GlobalDelegatedRewardsDistributor {
...
+      lastSnapshotContext: GlobalSnapshotInfo
     ): NonNegFraction =
-      dagPrices.filter { case (epoch, _) => epoch.value <= epochProgress.value }
-        .maxByOption(_._1.value.value)
-        .map(_._2)
+      lastSnapshotContext.priceState
+        .getOrElse(SortedMap.empty[TokenPair, PriceRecord])
+        .get(TokenPair.DAG_USD)
+        .map(priceRecord => priceRecord.currentPrice.price.value)
         .getOrElse(dagPrices.head._2)
```
scasplte2: it may be worthwhile to update this default option. I was comfortable doing it before since I could guarantee that a price was included in the `DelegatedRewardsConfigProvider` but how can we ensure the map is not empty in this approach?

effects-ai: Done

scasplte2 (approved):
lgtm :fire:

PR #1275, opened 2025-07-14
fix: adding GlobalIncrementalSnapshot and GlobalSnapshotStateProof to kryo registrar

marcus-girolneto:
### Changes
+ We need to add the classes GlobalIncrementalSnapshot and GlobalSnapshotStateProof to Kryo registrar for backward compatibility, in testnet as one example, if we don't have all snapshots lower or equal `1933590` will fail when decoding and get the hash, because this is the ordinal that we switched to JSON serializer.


ryle-goehausen (approved):
This is same problem that snapshot streaming hit earlier.

scasplte2 (approved):

scasplte2 (approved):

PR #1282, opened 2025-07-24
refactor: add new fields to delegated stake records

effects-ai:


gclaramunt:
If I understand correctly, the process would be:
create the original token lock
delegate it (new DelegateStakeRecord)
create a new token lock to replace the original
(a new DelegateStakeRecord will replace the previous one with the new lock ref in currentLockRef and new amount in current Amount) ?

effects-ai:
> If I understand correctly, the process would be: create the original token lock delegate it (new DelegateStakeRecord) create a new token lock to replace the original (a new DelegateStakeRecord will replace the previous one with the new lock ref in currentLockRef and new amount in current Amount) ?

Exactly.

modules/shared/src/main/scala/io/constellationnetwork/schema/tokenLock.scala
```diff
@@ -86,7 +86,8 @@ object tokenLock {
     fee: TokenLockFee,
     parent: TokenLockReference,
     currencyId: Option[CurrencyId],
-    unlockEpoch: Option[EpochProgress]
+    unlockEpoch: Option[EpochProgress],
+    replaceTokenLockRef: Option[TokenLockReference]
```
ryle-goehausen, line 90: This was decided in meeting

ryle-goehausen (dismissed):

ryle-goehausen (approved):

PR #1283, opened 2025-07-25
fix: halting consensus on failures

ryle-goehausen:
to reproduce the failure and attempted fix; run `just test --fail --cleanup` 

The cleanup arg stops the logs from accumulating after so they can be inspected easier.

Test results below:

Ordinal change deltas:
ℹ️    [1] Ordinal 2 → 4: 3.1s
ℹ️    [2] Ordinal 4 → 5: 5.1s
ℹ️    [3] Ordinal 5 → 6: 9.1s
ℹ️    [4] Ordinal 6 → 7: 2.0s
ℹ️    [5] Ordinal 7 → 8: 5.1s
ℹ️    [6] Ordinal 8 → 9: 1.5s
ℹ️    [7] Ordinal 9 → 10: 3.5s
ℹ️    [8] Ordinal 10 → 11: 6.1s
ℹ️    [9] Ordinal 11 → 12: 1.0s
ℹ️    [10] Ordinal 12 → 13: 3.0s
ℹ️    [11] Ordinal 13 → 14: 4.6s
ℹ️    [12] Ordinal 14 → 15: 13.7s
ℹ️    [13] Ordinal 15 → 16: 15.2s
ℹ️    [14] Ordinal 16 → 17: 14.7s
ℹ️    [15] Ordinal 17 → 18: 15.2s
ℹ️  
Max ordinal delta: 15.2s
ℹ️  Average ordinal delta: 6.9s
ℹ️  
Final delta (test end to last ordinal): 19.2s
ℹ️  
=== Timing Assertions ===
ℹ️  ✅ PASS: All 15 ordinal deltas were under 30s
ℹ️  ✅ PASS: Final delta was under 30s: 19.2s
ℹ️  
=== Signer Analysis ===
ℹ️  Total signers: 15
ℹ️  Signers: 1, 1, 3, 3, 3, 3, 3, 3, 3, 3, 3, 2, 2, 2, 2
ℹ️  
✅ Bulk submit test passed!
Bulk submit test completed took: 122 seconds - total time: 224 seconds


modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensusStateAdvancer.scala
```diff
@@ -195,8 +195,8 @@ object CurrencySnapshotConsensusStateAdvancer {
 
                 case CollectingSignatures(majorityArtifactInfo, majorityTrigger, candidates, ownFacilitatorsHash) =>
                   for {
-                    maybeAllSignatures <- maybeGetAllDeclarations(state, resources, config)(_.signature)
-                    maybeAllFacilities <- maybeGetAllDeclarations(state, resources, config)(_.facility)
+                    maybeAllSignatures <- maybeGetAllDeclarations(state, resources, config)(_.signature, _.signaturesLatestUnique)
+                    maybeAllFacilities <- maybeGetAllDeclarations(state, resources, config)(_.facility, _.facilitiesLatestUnique)
                     maybeGlobalSnapshotOrdinal =
```
ryle-goehausen, line 208: The part above this I am a bit uncertain of; since we're dealing with 2 types of 'getters' here, the previous one passed; but this one is being copied as part of the new state instantiation, so I'm not 100% certain this logic above is correct in the event of a timeout, despite the test passing.

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensusStateAdvancer.scala
```diff
@@ -277,7 +277,9 @@ object CurrencySnapshotConsensusStateAdvancer {
                     ) =>
                   {
                     for {
-                      maybeAllBinarySignatures <- OptionT.liftF(maybeGetAllDeclarations(state, resources, config)(_.binarySignature))
+                      maybeAllBinarySignatures <- OptionT.liftF(
+                        maybeGetAllDeclarations(state, resources, config)(_.binarySignature, _.signaturesLatestUnique)
+                      )
                       binarySignatures <- OptionT.fromOption[F](maybeAllBinarySignatures)
```
ryle-goehausen, line 292: ^ Also need to double check this part; not sure I accounted for this correctly in the addSignature portion ?

modules/dag-l0/src/main/resources/dag-l0.conf
```diff
@@ -12,7 +12,7 @@ snapshot {
     declaration-timeout = 50 seconds
     declaration-range-limit = 3
     lock-duration = 10 seconds
-    peers-declaration-timeout = 10 seconds
+    peers-declaration-timeout = 3 seconds
```
ryle-goehausen, line 15: This is deliberate; the problem here is there's 3 rounds of this in total; so 10 seconds adds a 30 second delay; since the other 'slow phases' like state transitions are already 5+ seconds, this is a non-trivial amount. This 3 second delay I'm certain of is actually after the unique event change now, as opposed to any resource change.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotConsensusStateAdvancer.scala
```diff
@@ -257,6 +261,15 @@ object GlobalSnapshotConsensusStateAdvancer {
...
+                                    numBlocks = signedArtifact.blocks.size
+                                    numTransactions = signedArtifact.blocks.map(_.block.transactions.size).sum
+                                    _ <- logger.debug(
+                                      s"Adding GlobalSnapshot to storage for " +
+                                        s"ordinal=${signedArtifact.ordinal.value.value} with hash=${hashedSnapshot.hash.value} " +
+                                        s"numSigners=${signedArtifact.proofs.size} " +
+                                        s"numBlocks=$numBlocks " +
+                                        s"numTransactions=$numTransactions"
+                                    )
                                     _ <- lastNGlobalSnapshotStorage.set(hashedSnapshot, majorityArtifactInfo.context)
```
ryle-goehausen, line 273: This is also deliberate log statement; it's very useful for gauging exact round change time for measuring other log statements.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/http/p2p/middlewares/FailureMiddleware.scala
```diff
@@ -0,0 +1,41 @@
...
+object FailureMiddleware {
+
+  def failureSimulatorFeatureCheck[F[_]: Async](envVarName: String): F[Unit] =
+    Async[F].delay {
+      sys.env.get(envVarName).map(_.toLong)
+    }.flatMap {
+      case Some(failTime) =>
+        // Monotonic is off by about 30 seconds here, ideal to switch to in future.
+        // Clock[F].monotonic.flatMap { monotonic =>
+        Clock[F].realTime.flatMap { currentTime =>
```
ryle-goehausen, line 21: I needed something I could map to event log timings easily. The compose runner set a delay of like +90 seconds (before docker started,) and this was showing ~120-130 seconds at the same trigger time.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -142,6 +142,10 @@ object ConsensusManager {
               .trySetInitialConsensusOutcome(outcome)
               .ifM(
                 nodeStorage.tryModifyState(Observing, WaitingForReady) >>
+                  // FIX: Start time trigger immediately after download
+                  // Previously, time trigger was only scheduled after successful consensus completion.
+                  // If the first consensus attempt failed, no periodic retry mechanism was active.
+                  scheduleFacility >>
                   internalFacilitateWith(none),
```
ryle-goehausen, line 145: I might need to revert this change above ^, I'm not certain if this was just an artifact of the testing process locally.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -164,20 +168,29 @@ object ConsensusManager {
...
+                logger.debug(s"Time trigger firing at ${currentTime}") >>
+                  internalFacilitateWith(TimeTrigger.some)
+              }
+            _ <- runTimeTriggerLoop // Continue the loop recursively
+          } yield ()
+
+          runTimeTriggerLoop
+            .handleErrorWith(logger.error(_)(s"Error in time trigger loop, restarting"))
+            .foreverM // Ensure it runs forever even if errors occur
+        }.void
```
ryle-goehausen, line 181: This is probably the most concerning section of code / change I made. This most likely needs to be changed; however, this was* one of the main things that allowed the local test to not halt (but I also made a ton of other changes which might not require this one now?). Probably worth reverting and trying variations.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ConsensusManager.scala
```diff
@@ -198,47 +211,78 @@ object ConsensusManager {
...
+                            case Some(_) =>
+                              logger.trace(s"Found existing state for {key=${nextKey.show}}, checking for updates") >>
+                                internalCheckForStateUpdate(nextKey, resources)
+                            case None =>
+                              logger.trace(s"No existing state for {key=${nextKey.show}}, nothing to do") >>
+                                Applicative[F].unit
+                          }
+                    }
+                }
+          }
```
ryle-goehausen, line 229: This is same logic just lots of tracing statements to figure out why the stall trigger code was never being reached.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ConsensusResources.scala
```diff
@@ -21,19 +21,34 @@ case class ConsensusResources[A, Kind](
...
-  updatedAt: FiniteDuration
+  updatedAt: FiniteDuration,
+  createdAt: FiniteDuration,
+  facilities: Option[FiniteDuration],
+  facilitiesLatestUnique: Option[FiniteDuration],
+  proposals: Option[FiniteDuration],
+  proposalsLatestUnique: Option[FiniteDuration],
+  signatures: Option[FiniteDuration],
+  signaturesLatestUnique: Option[FiniteDuration]
 )
```
ryle-goehausen, line 32: These represent first event occurrence and latest unique change, we probably don't need so many but its useful for event log tracing.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ConsensusResources.scala
```diff
@@ -21,19 +21,34 @@ case class ConsensusResources[A, Kind](
...
 
 object ConsensusResources {
   def empty[F[_]: Async, A, Kind]: F[ConsensusResources[A, Kind]] = for {
-    time <- Clock[F].monotonic
+    updatedAt <- Clock[F].realTime
+    createdAt <- Clock[F].realTime
```
ryle-goehausen, line 37: Deliberately not monotonic; such that we can map it to other log statements

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ConsensusStateAdvancer.scala
```diff
@@ -57,13 +62,40 @@ trait ConsensusStateAdvancer[F[_], Key, Artifact, Context, Status, Outcome, Kind
...
-      now <- Clock[F].monotonic
-      elapsed = now - resources.updatedAt
+      now <- Clock[F].realTime
+      started = resources.createdAt
+      latestUnique = staleTimer(resources)
+      uniqueDelta = latestUnique.map(_ - started)
+      updatedAtDelta = now - resources.updatedAt
+      elapsed = now - latestUnique.getOrElse(started)
       isStale = elapsed > config.peersDeclarationTimeout
+      _ <- logger.info(
```
ryle-goehausen, line 73: Primarily for debugging event changes on intnet

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ConsensusStorage.scala
```diff
@@ -293,18 +297,60 @@ object ConsensusStorage {
           } yield bound.toMap
 
         def addFacility(peerId: PeerId, key: Key, facility: Facility): F[Option[ConsensusResources[Artifact, Kind]]] =
-          updatePeerDeclaration(key, peerId) { peerDeclaration =>
-            peerDeclaration.focus(_.facility).modify(_.orElse(facility.some))
+          Clock[F].realTime.flatMap { realTime =>
+            logger.trace(s"addFacility called for key=$key, peerId=$peerId, facility=$facility") >>
+              updatePeerDeclarationWithChange(key, peerId, "facility") { peerDeclaration =>
```
ryle-goehausen, line 302: This is how I'm finding unique events

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ConsensusStorage.scala
```diff
@@ -353,6 +399,51 @@ object ConsensusStorage {
             }
           }
```
ryle-goehausen, line 401: Function below can be used for event log tracing at much better granularity than grafana. It does emit a lot of log noise but it's the main information we need.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/update/UnlockConsensusUpdate.scala
```diff
@@ -39,8 +39,10 @@ object UnlockConsensusUpdate {
                 .getOrElse(acc)
             }
 
-          val keepThreshold = (_facilitators.get(state).value.size + 1) / 2
-          val removeThreshold = _facilitators.get(state).value.size / 2 + 1
+          val facilitatorCount = _facilitators.get(state).value.size
+          // For small networks, be more lenient with thresholds
+          val keepThreshold = if (facilitatorCount <= 3) 1 else (facilitatorCount + 1) / 2
+          val removeThreshold = if (facilitatorCount <= 3) 2 else facilitatorCount / 2 + 1
```
ryle-goehausen, line 46: This may no longer be necessary

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/GossipRoundRunner.scala
```diff
@@ -51,19 +53,34 @@ object GossipRoundRunner {
...
-              .flatMap {
-                case (duration, _) => metrics.recordRoundDuration(duration, roundLabel)
-              }
-              .flatMap(_ => metrics.incrementGossipRoundSucceeded)
-              .handleErrorWith { err =>
-                logger.error(s"Error running gossip round {peer=${peer.show}, reason=${err.show}") >>
-                  localHealthcheck.start(peer)
-              },
-            selectedPeersR.update(_.excl(peer))
-          )
```
ryle-goehausen, line 66: This is a bit overkill in here, I think this function actually is fixed correctly due to other changes, but useful information here.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/healthcheck/LocalHealthcheck.scala
```diff
@@ -13,20 +13,25 @@ import cats.{Applicative, Show}
...
 import org.typelevel.log4cats.slf4j.Slf4jLogger
 import retry._
 
 object LocalHealthcheck {
   def make[F[_]: Async: Supervisor](nodeClient: NodeClient[F], clusterStorage: ClusterStorage[F]): F[LocalHealthcheck[F]] = {
     def mkPeersR = MapRef.ofConcurrentHashMap[F, PeerId, F[Fiber[F, Throwable, Unit]]]()
-    def retryPolicy: RetryPolicy[F] = RetryPolicies.fibonacciBackoff[F](2.seconds)
+    def retryPolicy: RetryPolicy[F] = RetryPolicies.limitRetries[F](3).join(RetryPolicies.constantDelay[F](5.seconds))
+    // TODO: Ensure new retry policy to prevent dead forks does not cause issue, previously using:
+    // def retryPolicy: RetryPolicy[F] = RetryPolicies.fibonacciBackoff[F](2.seconds)
```
ryle-goehausen, line 34: This may be unnecessary, I think test still works without this, but these 2-3 day fibonnaci retries are just wrong to begin with.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ConsensusManager.scala
ryle-goehausen: I reverted this; other fixes work without this, still needs changes to this section later.

marcus-girolneto (dismissed):
@ryle-goehausen It's a lot of changes in several places of the consensus. However, I think it's good enough so we can test it on integrationnet, I'm feeling comfortable with this

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/http/p2p/middlewares/FailureMiddleware.scala
scasplte2: I wonder if it has anything to do with running in a container environment? Could that additional overhead somehow mess with the assumptions of the cats scheduler?

scasplte2 (dismissed):
agreed, it is a lot but seems reasonable to try out

effects-ai (dismissed):

ryle-goehausen:
Closing this since other PRs replicate same solution now.

PR #1288, opened 2025-07-31
fix: fixing last global snapshot sync logic

marcus-girolneto:


modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/StateChannel.scala
```diff
@@ -227,61 +229,86 @@ object StateChannel {
...
+                      lastSyncGlobalSnapshot1 <- storages.lastSyncGlobalSnapshot.get
+
+                      lastSyncGlobalSnapshotOrdinal = lastSyncGlobalSnapshot.map(_._1.ordinal)
+                      lastGlobalSnapshotCombined <- sharedStorages.lastGlobalSnapshot.getCombined
+
+                      // The current and last global snapshot ordinal should be filled to not start a infinity loop of events
+                      _ <-
+                        if (
+                          currentLastSyncGlobalSnapshotOrdinal.isDefined && lastSyncGlobalSnapshotOrdinal.isDefined &&
+                          currentLastSyncGlobalSnapshotOrdinal.get < lastSyncGlobalSnapshotOrdinal.get
```
scasplte2, line 251: I don't think we should use `.get` here unless you can guarantee that it will not be `None` otherwise you will get a `NoSuchElementException`

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/StateChannel.scala
```diff
@@ -227,61 +229,86 @@ object StateChannel {
...
+                Applicative[F].unit
+              case nonEmptySnapshots =>
+                nonEmptySnapshots.tailRecM {
+                  case Nil =>
+                    for {
+                      currentLastSyncGlobalSnapshot <- currentLastSyncGlobalSnapshotRef.get
+                      currentLastSyncGlobalSnapshotOrdinal = currentLastSyncGlobalSnapshot.map(_.ordinal)
+
+                      lastSyncGlobalSnapshot <- storages.lastSyncGlobalSnapshot.getLastSynchronizedCombined
+                      lastSyncGlobalSnapshot1 <- storages.lastSyncGlobalSnapshot.get
```
scasplte2: what is this? the variable name is very opaque. Why would these two calls give different things?

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/StateChannel.scala
scasplte2: Perhaps this is a slightly cleaner version of the `Nil` case

```scala
case Nil =>
  (for {
    currentOrdinal <- OptionT(currentLastSyncGlobalSnapshotRef.get).map(_.ordinal)
    lastSync <- OptionT(storages.lastSyncGlobalSnapshot.getLastSynchronizedCombined)
    lastSyncOrdinal = lastSync._1.ordinal
    
    // Only proceed if current < last (prevents infinite loop)
    _ <- OptionT.whenM[F, Unit](
      Applicative[F].pure(currentOrdinal < lastSyncOrdinal)
    )(())
    
    combined <- OptionT(sharedStorages.lastGlobalSnapshot.getCombined)
    (latestSnapshot, latestState) = combined
    
    _ <- OptionT.liftF(
      logger.info("Trying to force event trigger") >>
      maybeForceEventTrigger(latestSnapshot, latestState, Some(lastSyncOrdinal))
    )
  } yield ()).value.flatMap { _ =>
    // Always update the ref with the latest sync snapshot
    storages.lastSyncGlobalSnapshot.getLastSynchronizedCombined.flatMap { lastSync =>
      currentLastSyncGlobalSnapshotRef.set(lastSync.map(_._1))
    }
  }.as(().asRight[List[Hashed[GlobalIncrementalSnapshot]]])
  ```

but is there a test to see if this works?

scasplte2 (dismissed):
talked with marcus over comments and have a better understand, LGTM

PR #1304, opened 2025-08-21
fix: adding check alignment stream

marcus-girolneto:


modules/dag-l1/src/main/scala/io/constellationnetwork/dag/l1/GlobalSnapshotAlignment.scala
```diff
@@ -121,20 +168,34 @@ class GlobalSnapshotAlignment[F[_]: Async: HasherSelector: SecurityProvider, P <
...
+    val checkAlignmentStream = checkAlignment.handleErrorWith { e =>
+      Stream.eval(logger.error(e)("Check alignment stream terminated unexpectedly, restarting")) >>
+        checkAlignment
+    }
+
+    Stream
+      .emits(List(l0PeerDiscoveryStream, globalSnapshotProcessingStream, checkAlignmentStream))
+      .covary[F]
+      .parJoinUnbounded
+  }
```
scasplte2: since each stream already has error hadnling, I think you can remove the `.handleErrorWith` on each of these vals. I'm also not sure if you need `.emits`. I think this would work just as well

```scala
def runtime(): Stream[F, Unit] = 
  Stream(l0PeerDiscovery, globalSnapshotProcessing, checkAlignment)
    .covary[F]
    .parJoinUnbounded
```

scasplte2 (dismissed):
minor simplification but looks good overall

scasplte2 (approved):
lgtm

PR #1321, opened 2025-09-17
fix: delegate reward info calculation now works with non-epoch snapshots

effects-ai:


effects-ai:
@greptile-apps please review it again

modules/shared/src/main/scala/io/constellationnetwork/schema/delegatedStake.scala
```diff
@@ -160,9 +160,9 @@ object delegatedStake {
     currentDagPrice: Amount,
     nextDagPrice: NextDagPrice,
     totalDelegatedAmount: Amount,
-    latestAverageRewardPerDag: Amount,
+    latestAverageRewardPerDag: BigDecimal,
```
scasplte2, line 163: doesn't `Amount` enforce this to be positive definite?

scasplte2 (approved):
nicely done :tada: much nicer abstraction than my haphazard attempt :smile:

modules/shared/src/main/scala/io/constellationnetwork/schema/delegatedStake.scala
effects-ai: `latestAverageRewardPerDag` (and `totalRewardsPerYearEstimate`) is now below 1, so I changed it's type to `BigDecimal` so we can have at least some value in the UI.

PR #1328, opened 2025-09-25
feat: (dataApp) add optional update hashes to incremental snapshot

scasplte2:
## Summary
Added support for tracking data update hashes in currency snapshots by introducing an optional `updateHashes` field to the `DataApplicationPart` schema. This enhancement allows data applications to optionally provide a hash function for their data updates, enabling better traceability and verification of data updates within snapshots.

## Changes
- Added `updateHashes: Option[SortedSet[Hash]]` field to `DataApplicationPart` for tracking update hashes
- Created `DataApplicationPartV1` to maintain backward compatibility with existing serialized data
- Added `hashDataUpdate` method to `BaseDataApplicationL0ContextualOps` and `DataApplicationL0ContextualOps` traits with the signature  `Option[DataUpdate => F[Hash]]` to make update hashing optional with default value of `None`
- Updated `DataApplicationSnapshotAcceptanceManager` to:
  - Check if `hashDataUpdate` is provided by the data application
  - Compute hashes for all data updates when the function is available
  - Include the computed hashes in the `DataApplicationPart` as a `SortedSet[Hash]`
  - Skip hash computation when no hash function is provided (backwards compatible behavior)
- Registered `DataApplicationPartV1` with Kryo serialization at ID 631 for backward compatibility
- Updated `CurrencyIncrementalSnapshot` to use the new `DataApplicationPart` with the optional field
- Ensured `CurrencyIncrementalSnapshotV1` continues using `DataApplicationPartV1` for deserializing legacy data

## Testing
- Local unit and e2e tests passing
- Verified backward compatibility with existing Kryo-serialized test data
- Confirmed state proof calculations remain consistent for historical snapshots
- Tested with data applications both with and without `hashDataUpdate` implementation

### DIFF
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/DataApplicationSnapshotAcceptanceManager.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/DataApplicationSnapshotAcceptanceManager.scala
@@ -24,6 +24,7 @@ import io.constellationnetwork.schema.artifact.{SharedArtifact, TokenUnlock}
 import io.constellationnetwork.security.hash.Hash
 import io.constellationnetwork.security.signature.Signed
 import io.constellationnetwork.security.{Hasher, SecurityProvider}
+import io.constellationnetwork.syntax.sortedCollection.sortedSetSyntax
 
 import org.typelevel.log4cats.slf4j.Slf4jLogger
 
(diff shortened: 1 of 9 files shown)


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/DataApplicationSnapshotAcceptanceManager.scala
```diff
@@ -248,9 +249,22 @@ object DataApplicationSnapshotAcceptanceManager {
         )
 
         sharedArtifacts = newDataState.sharedArtifacts ++ tokenUnlocks
+
+        updateHashes <- OptionT.liftF(
```
effects-ai, line 253: Where are the computed hashes used later on? I cannot find any usage of them...

scasplte2: there is currently no usage, the inclusion is informational for the moment so that snapshot streaming could expose the data update hashes via the DAG explorer. Currently only currency transactions can be shown because of the metagraph-specific serialization / hashing functionality. So snapshot streaming cannot compute these hashes without knowing the algo from the metagraph. Instead, I just have the metagraph compute and insert the hashes on snapshot creation. This seemed the easiest approach.

effects-ai (dismissed):

marcus-girolneto (approved):

PR #1331, opened 2025-09-30
feat: (scalafix) add custom rule for sum on set via scalafix module

scasplte2:
This PR introduces a custom scalafix rule to prevent the usage of `.sum` over Set-like collections. This builds on the previous work done in https://github.com/Constellation-Labs/tessellation/pull/1330#pullrequestreview-3285693323. Using the current state of `develop` this rule provides on error on compilation like blow
```
[info] done compiling
[info] Running scalafix on 5 Scala sources
[info] Running scalafix on 267 Scala sources
[error] /home/scas/git/tessellation/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotContextFunctions.scala:214:80: error: [NoSetSum.NoSetSum] Calling .sum on a Set discards duplicate values, leading to incorrect results. Use .toList.map(...).sum or compute the sum before converting to Set.
[error]                       totalEmittedRewardsAmount = Amount(NonNegLong.unsafeFrom(txs.map(_.amount.value.value).sum))
[error]                                                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error] /home/scas/git/tessellation/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotContextFunctions.scala:225:80: error: [NoSetSum.NoSetSum] Calling .sum on a Set discards duplicate values, leading to incorrect results. Use .toList.map(...).sum or compute the sum before converting to Set.
[error]                       totalEmittedRewardsAmount = Amount(NonNegLong.unsafeFrom(txs.map(_.amount.value.value).sum))
[error]                                                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error] (nodeShared / Compile / scalafix) scalafix.sbt.ScalafixFailed: LinterError
[error] Total time: 183 s (03:03), completed Sep 30, 2025, 12:02:44 PM
[IJ]
```

The usage of the `onLoad` global hook is to ensure the custom rules are present in the local ivy cache for running against the compiled application code.


gclaramunt:
nice, I was going the same route

gclaramunt (changes requested):
actually the rule should be no `map` to Sets

gclaramunt:
maybe we could also add a rule to warn on `++` on Map

gclaramunt (dismissed):
I would allow sum on Sets, but otherwise, LGTM

scasplte2:
I've refactored the original approach to separate out the custom scalafix rules. These are now located in a separate artifact that can be pulled into multiple projects. These projects can then enable the rules they find helpful.

The artifact is published to https://central.sonatype.com/artifact/io.constellationnetwork/constellation-scalafix-rules_2.13

@greptileai review again please

marcus-girolneto (approved):

PR #1333, opened 2025-10-02
fix: add support of history traversal

Serpion-ua:


.jvmopts
```diff
@@ -4,3 +4,4 @@
 -XX:ReservedCodeCacheSize=500M
 -XX:+TieredCompilation
 -XX:+UseParallelGC
+-Xss8M
```
marcus-girolneto: Why do you need this?

marcus-girolneto (changes requested):

.jvmopts
Serpion-ua: @IPadawans removed

scasplte2 (approved):
lgtm

PR #1334, opened 2025-10-06
feat: adding plugin system for metagraphs

marcus-girolneto:


ryle-ai:
## PR #1334 Review: Plugin System for Metagraphs

### Executive Summary

This PR introduces a plugin architecture for metagraphs, enabling modular composition of data application functionality. After thorough analysis of the ~54 changed files and 2,502 additions, I find that **Greptile's concerns are partially valid but somewhat overblown**. The type casting concerns are real but largely unavoidable given JVM type erasure constraints, and the implementation includes reasonable safeguards.

### Assessment of Key Files Flagged by Greptile

#### 1. `MasterPluginWrapper.scala` (Greptile Score: 2/5)

**Greptile's Concern:** "type-erased wrapper with mutable state, extensive unchecked casting"

**Analysis:**
- **No mutable state** in `MasterPluginWrapper` itself - it wraps an immutable `MasterPlugin`
- The `@unchecked` pattern matches (e.g., `case s: POnChain @unchecked`) are a standard Scala idiom for dealing with JVM type erasure
- The casting pattern follows a consistent fail-fast approach with `RuntimeException` on type mismatches

**Verdict:** The casting is necessary due to type erasure, properly guarded with runtime checks, and follows a consistent pattern. **Score adjustment: 3/5**

#### 2. `PluginWrapper.scala` (Greptile Score: 2/5)

**Greptile's Concern:** "mutable state, type safety concerns"

**Analysis:**
- **Mutable state exists** in `PluginRegistry` via `cats.effect.Ref` - this is **thread-safe** because `Ref` provides atomic, referentially transparent mutable state
- The casting on line 139 is concerning:
  ```scala
  _ <- pluginsRef.update(_ :+ wrapper.asInstanceOf[PluginWrapper[F, PUpdate, POnChain, PCalculated]])
  ```

**Thread Safety Verdict:** Using `Ref` is the correct functional approach for shared mutable state. **No thread safety issues.**

**Type Safety Concern:** The cast could theoretically allow registration of incompatible plugins - a valid concern.

**Score adjustment: 2.5/5**

#### 3. `BaseDataApplicationL1ContextualOps.scala` (Greptile Score: 2/5)

**Critical Finding - Unchecked Casts Without Validation:**
```scala
// Line 51: Direct cast without type check
service.validateFee(gsOrdinal)(dataUpdate.asInstanceOf[Signed[D]], maybeFeeTransaction)

// Line 57: Direct cast without type check  
service.estimateFee(gsOrdinal)(update.asInstanceOf[D])
```

These casts occur **without** the pattern match guard used elsewhere. If a `DataUpdate` of the wrong concrete type is passed, this will throw `ClassCastException` at runtime.

**Verdict:** This is a **legitimate concern**. **Score: 2/5** - Greptile's rating is appropriate here.

#### 4. `DataApplicationBlock.scala` (Greptile Score: 2/5)

**Analysis:**
- Simple case class with no mutable state
- No type casting
- Straightforward Circe encoder/decoder derivation

**Verdict:** No issues found. **Score adjustment: 4/5** - Greptile appears to have flagged this incorrectly.

### Critical Issues Found

#### Issue 1: Inconsistent Type Safety Patterns (MEDIUM)

Some locations use pattern matching with fallback, others use direct casting without guards. 

**Recommendation:** Standardize on the pattern-matching approach with proper error handling.

#### Issue 2: Missing Type Validation in Plugin Registration (MEDIUM)

In `PluginRegistry.register()`, the cast could silently corrupt the registry if incompatible plugins are registered.

**Recommendation:** Add runtime type checking via `ClassTag` to validate plugin compatibility at registration time.

#### Issue 3: Encoder `contramap` Throws Exceptions (LOW)

Throwing exceptions in encoders breaks the principle of total functions for JSON serialization. Consider returning a sentinel error JSON value instead.

### Breaking Changes for Existing Metagraph Developers

The breaking changes are **minimal and well-documented** via template updates:

1. **Import path change** (cosmetic): `dataApplication.dataApplication.*` → `dataApplication.*`
2. **New method requirement** - `hashDataUpdate` now expected
3. **`routesPrefix` signature change** - Now returns `F[ExternalUrlPrefix]` instead of `ExternalUrlPrefix`

### Summary Table

| File | Greptile Score | My Assessment | Revised Score |
|------|---------------|---------------|---------------|
| MasterPluginWrapper.scala | 2/5 | Casting necessary, properly guarded | 3/5 |
| PluginWrapper.scala | 2/5 | Thread-safe, but one unsafe cast | 2.5/5 |
| PluginOnlyDataApplicationL0Service.scala | 2/5 | Clean implementation | 3.5/5 |
| BaseDataApplicationL1ContextualOps.scala | 2/5 | Valid concern - unguarded casts | 2/5 |
| DataApplicationBlock.scala | 2/5 | No issues found | 4/5 |

### Recommendations

1. **Before merge:** Standardize type safety patterns in `BaseDataApplicationL1ContextualOps.scala` and `BaseDataApplicationL0ContextualOps.scala`
2. **Before merge:** Add runtime type validation to `PluginRegistry.register()`
3. **Post-merge follow-up:** Add comprehensive integration tests for plugin lifecycle with type mismatches
4. **Documentation:** Add migration guide for existing metagraph developers

### Verdict

**Conditional Approval** - The architecture is sound and the type casting is largely unavoidable on the JVM. However, the inconsistent handling of type mismatches (some graceful, some throwing) should be addressed before merge to prevent production runtime errors.

---
*Automated review by ryle-ai*

modules/shared/src/main/scala/io/constellationnetwork/currency/dataApplication/services/BaseDataApplicationL0Service.scala
```diff
@@ -0,0 +1,184 @@
...
+
+      def combine(state: DataState.Base, updates: List[Signed[DataUpdate]])(
+        implicit context: L0NodeContext[F]
+      ): F[DataState.Base] =
+        service.pluginRegistry match {
+          case Some(registry) =>
+            for {
+              pluginState <- registry.combine(state.onChain, state.calculated, updates)
+              finalState <- ctx.combine(DataState(pluginState._1, pluginState._2), updates)
+            } yield finalState
```
ryle-ai, line 148: I think this has a subtle bug with double-application of updates. When `pluginRegistry` is `Some`, we call `registry.combine(state.onChain, state.calculated, updates)` here, which runs all plugins + master on the full `updates` list. Then we take `pluginState` and immediately pass the *same* `updates` to `ctx.combine` on the next line -- which calls `service.combine` (the user's `DataApplicationL0Service` implementation) with those same updates again.

So every update gets processed twice: once through the plugin registry, once through the base contextual ops. For the existing (non-plugin) path this is fine since `pluginRegistry` is `None`, but for plugin-based metagraphs that also override `combine` in their `DataApplicationL0Service`, the state mutation will be applied twice.

I believe the intent here is that when plugins are present, the plugin registry should be the *sole* combiner, and `ctx.combine` should be bypassed -- or the updates should be partitioned so each handler only sees its own subset. Worth clarifying the expected behavior here before merge.

🤖 *ryle-ai automated review*

modules/shared/src/main/scala/io/constellationnetwork/currency/dataApplication/services/PluginOnlyDataApplicationL0Service.scala
```diff
@@ -0,0 +1,202 @@
...
+
+  // ========== Combine: all plugins (master + features) ==========
+
+  override def combine(
+    state: DataState.Base,
+    updates: List[Signed[DataUpdate]]
+  )(implicit context: L0NodeContext[F]): F[DataState.Base] =
+    registry.combine(state.onChain, state.calculated, updates).map {
+      case (newOnChain, newCalculated, artifacts) =>
+        DataState(newOnChain, newCalculated)
```
ryle-ai, line 117: The `SharedArtifact` list produced by `registry.combine` is silently dropped here -- we destructure `(newOnChain, newCalculated, artifacts)` but then construct `DataState(newOnChain, newCalculated)` without passing `artifacts`. This means any `SharedArtifact` produced by plugin `combine` calls will never make it into the snapshot.

Looking at how `DataApplicationSnapshotAcceptanceManager` uses the result (line ~250 in that file: `sharedArtifacts = newDataState.sharedArtifacts ++ tokenUnlocks`), the artifacts need to be in `DataState.sharedArtifacts` to be included. I think this should be `DataState(newOnChain, newCalculated, artifacts.toSortedSet)` or similar -- otherwise `PluginLifecycle.combine` returning `SharedArtifact`s is effectively dead code.

🤖 *ryle-ai automated review*

modules/shared/src/main/scala/io/constellationnetwork/currency/dataApplication/services/PluginOnlyDataApplicationL1Service.scala
```diff
@@ -0,0 +1,140 @@
...
+      case None         => Async[F].raiseError(new RuntimeException("No master plugin registered"))
+    }
+}
+
+object PluginOnlyDataApplicationL1Service {
+  def make[F[_]: Async, U <: DataUpdate, POnChain <: DataOnChainState, PCalculated <: DataCalculatedState](
+    masterPlugin: io.constellationnetwork.currency.dataApplication.plugin.MasterPlugin[F, U, POnChain, PCalculated]
+  ): F[PluginOnlyDataApplicationL1Service[F]] =
+    for {
+      registry <- PluginRegistry.make[F, U, POnChain, PCalculated]
```
ryle-ai, line 128: So each of `PluginOnlyDataApplicationL0Service.make` and `PluginOnlyDataApplicationL1Service.make` creates its own `PluginRegistry` and calls `registerMaster` independently. This means the L0 and L1 services each have a completely separate registry instance with separate plugin state.

I think this is probably intentional -- the L0 and L1 run in different JVM processes -- but it means the caller has to remember to register feature plugins (via `registry.register`) on *both* service instances separately, and there's nothing in the API that enforces or even hints at this. If someone registers a plugin on the L0 side but not L1, their L1 validation will silently pass updates that L0 will then reject during snapshot acceptance.

Might be worth documenting this requirement, or providing a shared factory method that constructs both services from the same plugin set.

🤖 *ryle-ai automated review*

modules/shared/src/main/scala/io/constellationnetwork/currency/dataApplication/plugin/PluginWrapper.scala
```diff
@@ -0,0 +1,336 @@
...
+
+  def combine(
+    onChainState: POnChain,
+    calculatedState: PCalculated,
+    updates: List[Signed[PUpdate]]
+  )(implicit context: L0NodeContext[F]): F[(POnChain, PCalculated, List[SharedArtifact])] =
+    for {
+      plugins <- pluginsRef.get
+
+      stateAfterPlugins <- plugins.foldLeftM((onChainState, calculatedState, List.empty[SharedArtifact])) {
```
ryle-ai, line 253: The `foldLeftM` here means plugin `combine` results are sequentially threaded -- each plugin receives the *accumulated* state from all previous plugins. This makes the result depend on registration order, which is determined by the order of `pluginsRef.update(_ :+ wrapper)` calls in `register()`.

That's probably fine as a design choice, but it's worth noting that the ordering semantics are implicit and not documented anywhere. If two plugins modify the same portion of `POnChain`, the last-registered plugin wins silently. For the `validateData` path above (line 228), all plugins get the *original* state -- so there's an asymmetry between validation (sees original state) and combine (sees accumulated state from prior plugins). I think this is intentional but should be called out in docs since it's a non-obvious subtlety for plugin authors.

🤖 *ryle-ai automated review*

modules/shared/src/main/scala/io/constellationnetwork/currency/dataApplication/plugin/PluginWrapper.scala
```diff
@@ -0,0 +1,336 @@
...
+          plugin.combine(currentOnChain, currentCalculated, updates).map {
+            case (newOnChain, newCalculated, newArtifacts) =>
+              (newOnChain, newCalculated, artifacts ++ newArtifacts)
+          }
+      }
+
+      finalState <- getMasterPlugin.flatMap {
+        case Some(master) =>
+          val (onChain, calculated, artifacts) = stateAfterPlugins
+          master.combine(onChain, calculated, updates).map {
```
ryle-ai, line 264: I think there's an ordering question here too -- `master.combine` runs *after* all feature plugins (it receives `stateAfterPlugins`), but for `validateData` above, master validation runs after plugin validations are collected but they all see the *same* original state. This asymmetry in the combine path means the master plugin effectively has "last write wins" semantics on the combined state, which is probably the right call architecturally -- but again worth documenting for plugin developers.

Also, the master receives the *same full updates list* that the feature plugins already processed. The `handles()` filtering inside each plugin/master prevents double-processing of the same update type, but only if `handles()` correctly partitions -- if a master and a feature plugin both claim to handle the same update type, that update's combine logic runs twice on the state.

🤖 *ryle-ai automated review*

modules/shared/src/main/scala/io/constellationnetwork/currency/dataApplication/PluginSystem.scala
```diff
@@ -0,0 +1,26 @@
+package io.constellationnetwork.currency.dataApplication
+
+import cats.data.NonEmptyList
+
+import io.constellationnetwork.currency.dataApplication.DataApplicationValidationErrorOr
+import io.constellationnetwork.security.signature.Signed
+
+trait PluginSystem {
```
ryle-ai, line 8: Is `PluginSystem` used anywhere? I can't find any implementations or references outside this file. It looks like it was maybe an earlier design sketch that got superseded by `MetagraphPlugin` + `PluginLifecycle` + `PluginWrapper`. If it's not used, I'd remove it to avoid confusion -- having two parallel plugin abstractions in the same package is going to confuse developers.

🤖 *ryle-ai automated review*

modules/shared/src/main/scala/io/constellationnetwork/currency/dataApplication/services/PluginOnlyDataApplicationL0Service.scala
```diff
@@ -0,0 +1,202 @@
...
+    registry.validateData(state.onChain, state.calculated, updates)
+
+  override def validateFee(
+    gsOrdinal: SnapshotOrdinal
+  )(dataUpdate: Signed[DataUpdate], maybeFeeTransaction: Option[Signed[FeeTransaction]])(
+    implicit context: L0NodeContext[F],
+    A: Applicative[F]
+  ): F[DataApplicationValidationErrorOr[Unit]] = {
+    import cats.data.Validated
+    Validated.validNec[DataApplicationValidationError, Unit](()).pure[F]
```
ryle-ai, line 101: `validateFee` here always returns `validNec` -- every fee validation passes unconditionally. For the existing `DataApplicationL0Service` path, fee validation is delegated to the user's service implementation via `ctx.validateFee`. But for plugin-based metagraphs, there's no mechanism for plugins to participate in fee validation at all.

The same pattern appears in `PluginOnlyDataApplicationL1Service` (line 92). This means plugin-based metagraphs effectively have no fee validation -- which is fine if that's the intended design (fees are opt-in), but worth a comment explaining why this is intentionally a no-op, since the existing non-plugin path does validate fees.

🤖 *ryle-ai automated review*

modules/shared/src/main/scala/io/constellationnetwork/currency/dataApplication/plugin/MasterPluginWrapper.scala
```diff
@@ -0,0 +1,246 @@
...
+  def onGlobalSnapshotPull(snapshot: Hashed[GlobalIncrementalSnapshot], context: GlobalSnapshotInfo)(implicit A: Applicative[F]) =
+    masterPlugin.onGlobalSnapshotPull(snapshot, context)
+
+  def getTokenUnlocks(
+    stateToProcess: DataState[DataOnChainState, DataCalculatedState]
+  )(implicit context: L0NodeContext[F], hasher: Hasher[F]): F[SortedSet[TokenUnlock]] =
+    masterPlugin.getTokenUnlocks(stateToProcess)
+
+  def getCalculatedState(implicit context: L0NodeContext[F]) =
+    masterPlugin.getCalculatedState.map { case (ord, calc) => (ord, calc.asInstanceOf[DataCalculatedState]) }
```
ryle-ai, line 207: I think `getCalculatedState` here has an unnecessary `.asInstanceOf[DataCalculatedState]` -- `masterPlugin.getCalculatedState` returns `F[(SnapshotOrdinal, PCalculated)]` where `PCalculated <: DataCalculatedState`, so the `.widen` pattern used elsewhere in this file (e.g., `deserializeState` at line 141, `deserializeCalculatedState` at line 165) would work here too. The `asInstanceOf` isn't unsafe in this case since the bound guarantees the cast, but it breaks the pattern established in the rest of the wrapper and suppresses compiler warnings that would catch a real type error.

🤖 *ryle-ai automated review*

ryle-ai (commented):
Deeper follow-up review -- I went through the actual codebase to cross-reference the plugin system integration points with the existing consensus pipeline and snapshot acceptance flow. Found a few issues that I think are worth addressing before merge, mostly around the combine path and shared artifact propagation.

ryle-ai:
I think there's a bug in the plugin combine path that silently drops shared artifacts. In `BaseDataApplicationL0Service.apply`, when `pluginRegistry = Some(registry)`:

```scala
pluginState <- registry.combine(state.onChain, state.calculated, updates)
finalState <- ctx.combine(DataState(pluginState._1, pluginState._2), updates)
```

`registry.combine` returns `F[(POnChain, PCalculated, List[SharedArtifact])]`, but `DataState(pluginState._1, pluginState._2)` only takes the first two elements -- `pluginState._3` (the `List[SharedArtifact]`) gets dropped. `DataState`'s third parameter defaults to `SortedSet.empty`, so any shared artifacts produced by plugins are gone before `ctx.combine` ever sees them.

Same issue exists in `PluginOnlyDataApplicationL0Service.combine` -- the `registry.combine(...).map { case (newOnChain, newCalculated, artifacts) => DataState(newOnChain, newCalculated) }` explicitly destructures `artifacts` and then doesn't use it.

Separately -- I believe there's a state visibility inconsistency between `PluginRegistry.validateData` and `PluginRegistry.combine`. `validateData` runs all plugin validations via `traverse` (each plugin sees the original input state), while `combine` uses `foldLeftM` (each plugin sees state from the previous plugin's combine). This might be intentional -- validation maybe shouldn't depend on ordering -- but it's worth calling out since it means a plugin could validate against state that's different from what it actually receives during combine.

ryle-ai (commented):
Follow-up to the double-combine issue I flagged earlier -- I think there's another bug in the same area. `registry.combine` returns a `(OnChainState, CalculatedState, List[SharedArtifact])` tuple, but both the legacy path in `BaseDataApplicationL0Service` and the pure plugin path in `PluginBasedDataApplicationL0Service` construct `DataState(pluginState._1, pluginState._2)` -- dropping the third element entirely. Since `DataState.sharedArtifacts` defaults to `SortedSet.empty`, any `SharedArtifact`s produced by plugins during combine are silently lost.

I looked at the `PluginBasedDataApplicationL0Service` version and it actually pattern-matches `artifacts` but then doesn't use it: `case (newOnChain, newCalculated, artifacts) => DataState(newOnChain, newCalculated)`. I believe this should be `DataState(newOnChain, newCalculated, SortedSet.from(artifacts))` or similar -- assuming `SharedArtifact` has an `Ordering` defined (which it does given the `SortedSet` in the case class).

This might not matter yet if no plugins are currently producing shared artifacts, but it seems like a latent issue that could be confusing to debug later. Probably worth fixing while the plugin system code is still fresh.

Also noticed a secondary concern -- `PluginRegistry.combine` folds plugins sequentially (each plugin sees state from the previous one), but `PluginRegistry.validateData` runs all validations in parallel against the original state. If plugin B's validation depends on state changes from plugin A's combine, the validation will pass against stale state. Might be intentional for performance reasons, but the different state visibility semantics between validate and combine should probably be documented somewhere.

ryle-ai (commented):
Quick correction to my earlier comment -- I want to reframe both points.

The sharedArtifacts drop is worth noting as a latent issue but probably not blocking for this PR. If no plugins currently produce shared artifacts (which I believe is the case), this is a future concern rather than an active bug. The pattern-match that drops the third tuple element should probably be fixed eventually so it doesn't surprise someone later, but it's not urgent for this change.

On the validate-vs-combine state visibility difference -- this is more of a general concern about the plugin registry's parallel semantics than something specific to this PR. It's the same class of issue as Seq vs ParSeq execution ordering, and the parallel validation APIs haven't been implemented yet anyway. Something to keep in mind as the plugin system evolves, but not something this PR needs to address.

Apologies for framing these as more urgent than they are in the previous comment.

PR #1335, opened 2025-10-08
feat: add default calculated state ref to data application service

scasplte2:
## Summary

This PR adds default calculated state management to the `DataApplicationL0Service` trait, eliminating the need for metagraph developers to implement their own `CalculatedStateService`. The default implementation uses a `Ref`-based approach to maintain calculated state across snapshot processing, simplifying the development experience and reducing boilerplate code.

## Changes

- **Added** default `getCalculatedState` and `setCalculatedState` method implementations in `DataApplicationL0Service` using `cats.effect.Ref` for in-memory state management
- **Modified** method signatures to use `implicit async: Async[F]` instead of `implicit context: L0NodeContext[F]` since calculated state management doesn't require node context
- **Removed** `getCalculatedState` and `setCalculatedState` from `BaseDataApplicationL0ContextualOps` and `DataApplicationL0ContextualOps` traits as they are not contextual operations
- **Updated** `BaseDataApplicationL0Service` to properly delegate to the typed service implementation with correct implicit parameters
- **Removed** `CalculatedStateService` from metagraph project template as it's no longer needed
- **Updated** metagraph project template `Main.scala` to use the default calculated state implementation directly
- **Simplified** `CustomRoutes` to call `getCalculatedState` directly instead of going through a separate service

## Testing

- Verify that `shared` module compiles successfully with the new default implementation
- Validate that metagraph template projects can build without `CalculatedStateService`
- Confirm that calculated state is properly maintained across snapshot processing using the default `Ref`-based implementation
- Test that metagraph developers can still override `getCalculatedState` and `setCalculatedState` for custom state management needs

---
## Calculated State Refactor: Core Issues with Previous Design

The previous design required metagraph developers to implement `getCalculatedState` and `setCalculatedState`, creating the appearance that they control state persistence. This led to three critical issues:

1. **Misplaced responsibility** - Developers implemented infrastructure instead of business logic
2. **Dangerous flexibility** - The API permitted implementations that break framework assumptions
3. **Hidden synchronization hazards** - External storage creates consistency problems during normal operation and especially during rollback/recovery

### The Framework's Implicit Contract

Tessellation's calculated state system has specific requirements that were not enforced by the old API:

```scala
// Old API - looks flexible
trait DataApplicationL0ContextualOps[F[_], D, DON, DOF] {
  def getCalculatedState(implicit context: L0NodeContext[F]): F[(SnapshotOrdinal, DOF)]
  def setCalculatedState(ordinal: SnapshotOrdinal, state: DOF)(implicit context: L0NodeContext[F]): F[Boolean]
}
```

**Implied contract** (not documented):
- `getCalculatedState` must be **fast and consistent** (read from memory)
- `setCalculatedState` must be **idempotent and synchronous**
- State must be **immediately available** after `setCalculatedState` returns
- Failures in `setCalculatedState` must **propagate to snapshot rejection**
- **State must support rollback via `deleteAbove(ordinal)`** (critical for recovery)

### Why Database-Backed State Breaks the Contract

Consider a developer attempting to use PostgreSQL for queryable state history:

```scala
override def setCalculatedState(
  ordinal: SnapshotOrdinal,
  state: UsageUpdateCalculatedState
)(implicit context: L0NodeContext[F]): F[Boolean] =
  for {
    _ <- stateRef.set((ordinal, state))
    _ <- postgresClient.upsert(ordinal, state)  // Additional persistence
  } yield true
```

#### Problem 1: Write Amplification Creates Inconsistency Window

**Framework's state flow** (`DataApplicationSnapshotAcceptanceManager.scala`):
```scala
newCalculatedState <- service.combine(currentState, dataUpdates)
_ <- service.setCalculatedState(artifact.ordinal, newCalculatedState)         // Step 1
_ <- calculatedStateStorage.write(artifact.ordinal, newCalculatedState)(...)  // Step 2
```

**With database-backed implementation**:
- Step 1 writes to: In-memory Ref + PostgreSQL
- Step 2 writes to: Tessellation's disk storage

**Three copies, three write operations, multiple failure modes**:

| Failure Point | Ref State | PostgreSQL State | Disk State | Consistency Issue |
|---------------|-----------|------------------|------------|-------------------|
| Postgres write fails | Old | Old | Old | Clean failure - retry ✅ |
| Postgres succeeds, disk fails | New | New | Old | **Node state diverges from persisted state** ❌ |
| Disk succeeds, Postgres fails | Old | Old | New | **Ref inconsistent with disk** ❌ |
| Node crashes after Postgres | New | New | Old | **On restart: disk overwrites Postgres** ❌ |

#### Problem 2: Rollback Cannot Delete External State

**Rollback flow** (`DataApplicationTraverse.scala:323-326`):
```scala
dataApplication.setCalculatedState(latestOrdinal, latestState.calculated) >>
  globalSnapshotsWithStateLocalFileSystemStorage.deleteAbove(globalOrdinal) >>
  globalSnapshotsWithStateDeltasLocalFileSystemStorage.deleteAbove(globalOrdinal) >>
  calculatedStateStorage.deleteAbove(latestOrdinal)  // Deletes disk files > latestOrdinal
```

**What rollback does**:
1. Sets calculated state to recovered ordinal (e.g., ordinal 100)
2. Deletes all disk-stored calculated states > ordinal 100 (e.g., 101-110)

**With PostgreSQL-backed state**:

```scala
override def setCalculatedState(ordinal: SnapshotOrdinal, state: DOF): F[Boolean] =
  for {
    _ <- stateRef.set((ordinal, state))
    _ <- postgresClient.upsert(ordinal, state)
  } yield true
```

**What happens during rollback**:

| Step | Ref | Disk | PostgreSQL | Consistency |
|------|-----|------|------------|-------------|
| Normal operation at ordinal 110 | S110 | S110 | S110 | ✅ |
| **Corruption detected, rollback to 100** | - | - | - | - |
| Framework calls `setCalculatedState(100, S100)` | S100 | - | **S100 (UPSERT!)** | 😱 |
| Framework calls `deleteAbove(100)` on disk | S100 | ✅ S101-110 deleted | ❌ S101-110 **STILL EXIST** | ❌ |
| **Final state** | S100 | Files: S100 | **Rows: S100-S110** | **CORRUPTED** |

**The critical issue**: `deleteAbove` only knows about Tessellation's disk storage. The framework has **no way to tell the metagraph's database to delete ordinals > 100**.

**Consequences**:
- PostgreSQL now contains "future" states (101-110) that were never actually accepted
- Queries return inconsistent data: "What's the state at ordinal 105?" → Returns S105, which was rolled back
- Historical queries are permanently corrupted - no way to know which states were valid
- Developer must **manually** detect and clean up PostgreSQL, but framework doesn't tell them rollback happened

#### Problem 3: Rollback Idempotency Violations

**Rollback can happen multiple times** during recovery:

```scala
// Rollback.scala:165
currencySnapshotCleanupStorage.cleanupAbove(lastIncremental.ordinal)
```

**Timeline of repeated rollback**:

| Event | Ref | Disk | PostgreSQL |
|-------|-----|------|------------|
| Normal state at ordinal 110 | S110 | S110 | S110 |
| **First rollback to 100** | S100 | S100 | S100, **S101-110** ❌ |
| Node restarts, traversal finds corruption | - | - | - |
| **Second rollback to 95** | S95 | S95 | S95, **S96-110** ❌ |
| Developer queries "state at ordinal 100" | - | - | **Returns S100** ❌ |

**PostgreSQL now has**:
- Ordinals 96-110: States that were rolled back (should not exist)
- No way to distinguish "valid historical state" from "rolled back state"
- Ordinals 100-110 were written, rolled back, but never deleted

**How would developer fix this?**
They can't, without:
1. Implementing their own `deleteAbove` in PostgreSQL
2. Framework calling it during rollback (but framework doesn't know about their database)
3. Tracking "valid vs rolled back" state in metadata

#### Problem 4: Read-After-Rollback Consistency Violations

**After rollback**:

```scala
// Framework restores to ordinal 100
dataApplication.setCalculatedState(100, S100)
calculatedStateStorage.deleteAbove(100)
```

**Application code queries database**:
```scala
def getDeviceHistory(address: Address): F[List[(SnapshotOrdinal, DeviceInfo)]] =
  postgresClient.query(
    sql"SELECT ordinal, device_info FROM calculated_state WHERE address = $address ORDER BY ordinal"
  )
```

**Returns**:
- Ordinal 100: DeviceInfo(count=5)
- Ordinal 101: DeviceInfo(count=6) ← **ROLLED BACK STATE**
- Ordinal 102: DeviceInfo(count=7) ← **ROLLED BACK STATE**
- ...
- Ordinal 110: DeviceInfo(count=15) ← **ROLLED BACK STATE**

**User sees**: "Device count increased from 5 to 15", but reality is node is at ordinal 100 with count=5.

**Silent data corruption** - no error, no warning, just wrong results.

#### Problem 5: No Extension Point for Cleanup

**Current API doesn't provide hooks for cleanup**:

```scala
trait DataApplicationL0ContextualOps[F[_], D, DON, DOF] {
  def getCalculatedState(implicit context: L0NodeContext[F]): F[(SnapshotOrdinal, DOF)]
  def setCalculatedState(ordinal: SnapshotOrdinal, state: DOF)(implicit context: L0NodeContext[F]): F[Boolean]
  // ❌ No deleteAbove method for metagraph to implement
  // ❌ No rollback notification
  // ❌ No cleanup hook
}
```

**Framework code** (`DataApplicationTraverse.scala:326`):
```scala
calculatedStateStorage.deleteAbove(latestOrdinal)  // Only cleans framework's disk storage
```

**No way for metagraph developer to be notified** that they should clean their PostgreSQL tables.

### Real-World Rollback Scenario

**Timeline of subtle corruption**:

| Time | Event | Ref | Disk | PostgreSQL |
|------|-------|-----|------|------------|
| T0 | Process snapshots 1-110 | S110 | S110 | S110 |
| T1 | **Disk corruption detected at ordinal 105** | S110 | Corrupted | S110 |
| T2 | Node initiates rollback to ordinal 100 | - | - | S110 |
| T3 | Framework calls `setCalculatedState(100, S100)` | S100 | - | **S100 (upsert)** |
| T4 | Framework calls `deleteAbove(100)` on disk | S100 | ✅ 101-110 deleted | ❌ 101-110 remain |
| T5 | Node resumes from ordinal 100 | S100 | S100 | S100-S110 |
| T6 | Process snapshot 101 (new S101') | S101' | S101' | **S101' (overwrites), S102-S110 (orphaned)** |
| T7 | Developer queries ordinal 105 | - | - | **Returns old S105** ❌ |

**Final state**:
- Current ordinal: 101
- Disk has: S100, S101'
- PostgreSQL has: S100, S101' (updated), **S102-S110 (orphaned from rolled-back future)**

**Developer debugging**:
- "Why does database show state at ordinal 105 when node is at ordinal 101?"
- "Did we process snapshot 105 or not?"
- "How do I query 'real' state vs 'rolled back' state?"

No good answers without manual database inspection and cleanup.

## Secondary Issue: Unnecessary Boilerplate

Every metagraph implements identical code:

```scala
// 48 lines of infrastructure code
trait CalculatedStateService[F[_]] { ... }
object CalculatedStateService {
  def make[F[_]: Async]: F[CalculatedStateService[F]] =
    Ref.of[F, CalculatedState](CalculatedState.empty).map { stateRef =>
      new CalculatedStateService[F] {
        // Identical across all metagraphs
      }
    }
}
```

**Developer confusion**:
- "Why am I implementing this?"
- "Is this where I add my business logic?"
- "Can I optimize this?" (No - don't touch it)

## Solution: Framework-Managed State

**New design**:
```scala
trait DataApplicationL0Service[F[_], D, DON, DOF] {
  def genesis: DataState[DON, DOF]

  // Business logic only - no storage concerns
  def combine(state: DataState[DON, DOF], updates: List[Signed[D]]): F[DataState[DON, DOF]]
  def validateData(state: DataState[DON, DOF], updates: NonEmptyList[Signed[D]]): F[ValidationResult]

  // Framework-provided default implementation
  final def getCalculatedState(implicit async: Async[F]): F[(SnapshotOrdinal, DOF)] =
    latestCalculatedStateRef.flatMap(_.get)

  final def setCalculatedState(ordinal: SnapshotOrdinal, state: DOF)(implicit async: Async[F]): F[Boolean] =
    latestCalculatedStateRef.flatMap(_.set((ordinal, state)).as(true))

  private def latestCalculatedStateRef(implicit async: Async[F]): F[Ref[F, (SnapshotOrdinal, DOF)]] =
    Ref.of[F, (SnapshotOrdinal, DOF)]((SnapshotOrdinal.MinValue, genesis.calculated))
}
```

**Benefits**:
1. **Eliminates dangerous flexibility** - Developers cannot accidentally implement broken state management
2. **Removes boilerplate** - 48 lines of identical code per metagraph eliminated
3. **Clearer responsibilities** - State storage is framework concern, business logic is metagraph concern
4. **Still extensible** - Developers can override for auxiliary storage:

```scala
// Safe extension pattern
override def setCalculatedState(ordinal: SnapshotOrdinal, state: DOF)(implicit async: Async[F]): F[Boolean] =
  for {
    result <- super.setCalculatedState(ordinal, state)  // Delegate to framework
    _ <- metricsDB.recordStateSize(state.devices.size)  // Auxiliary writes OK here
  } yield result
```

This is safe because:
- Framework's Ref is updated first (maintains contract)
- Auxiliary writes are clearly additive, not replacements
- Failures in auxiliary writes propagate correctly
- **Developer is responsible for their own cleanup** (framework doesn't manage it)

**Note**: If developers want queryable history in PostgreSQL, they should:
1. Override `setCalculatedState` to write to both Ref (via super) and database
2. **Handle their own cleanup** during issues (framework won't clean their database)
3. Accept triple storage (Ref + Disk + Database) as a tradeoff for queryability

## Conclusion

The previous design's flexibility was a **footgun** especially during rollback:

- Appeared to give developers control over state storage
- Actually required developers to maintain framework invariants they couldn't see
- Permitted implementations that **silently break during rollback** - the framework deletes its state but not external databases
- No extension points for cleanup operations during rollback

The refactor **removes dangerous flexibility** while maintaining extensibility for legitimate use cases. Developers implement business logic; framework handles infrastructure. If developers want auxiliary storage, they must handle their own lifecycle management.


scasplte2:
@greptileai review the new approach

AlexBrandes (changes requested):
We had hoped to expand the ability for developers to choose calculated state stores in the future that were something other than in-memory. In-memory state is seen as a potential scaling issue for applications that require lots of data to be available to the metagraph processes. So with that framing, my preference would be to find ways to expand support for other data stores rather than remove support for them. 

Would providing access to implement a custom `deleteAbove` method and clear documentation about the expectations of a custom data store cover the majority of the concerns? We could also potentially update error handling to allow downstream (from the postgres setter) errors to be caught and rolled back. 

On the issue of the boilerplate, I agree that the current implementation isn't very developer friendly. Is there a way to provide a sensible default that can be overridden if needed?

scasplte2:
> We had hoped to expand the ability for developers to choose calculated state stores in the future that were something other than in-memory. In-memory state is seen as a potential scaling issue for applications that require lots of data to be available to the metagraph processes. So with that framing, my preference would be to find ways to expand support for other data stores rather than remove support for them.
> 
> Would providing access to implement a custom `deleteAbove` method and clear documentation about the expectations of a custom data store cover the majority of the concerns? We could also potentially update error handling to allow downstream (from the postgres setter) errors to be caught and rolled back.
> 
> On the issue of the boilerplate, I agree that the current implementation isn't very developer friendly. Is there a way to provide a sensible default that can be overridden if needed?

@AlexBrandes Calculated state is back by a filesystem store within tessellation so the in-memory version is simply acting as a cache with the current approach. The metagraph functionality would be the exact same if instead of moving the Ref internally I completely removed the in-memory store here as it doesn't provide any additional benefit other than fast-access for validate / combine operations.

I agree that the ability to expand the data storage to a large scale would be beneficial but this still leaves the synchronization issue between the filesystem store and the external store. If we want to expose a more scalable store then shouldn't we eliminate the filesystem store in order to have a single source of truth?

AlexBrandes:
> @AlexBrandes Calculated state is back by a filesystem store within tessellation so the in-memory version is simply acting as a cache with the current approach. The metagraph functionality would be the exact same if instead of moving the Ref internally I completely removed the in-memory store here as it doesn't provide any additional benefit other than fast-access for validate / combine operations.
> 
> I agree that the ability to expand the data storage to a large scale would be beneficial but this still leaves the synchronization issue between the filesystem store and the external store. If we want to expose a more scalable store then shouldn't we eliminate the filesystem store in order to have a single source of truth?

I don't see the synchronization issue as particularly significant because the main problem is that we're missing hooks into at least one important event in the storage lifecycle (rollback). We could just add that hook to solve that problem imo. 

The vision for expanding calculated state to use non-disk data stores is that calculated state could become arbitrarily large, like it is in standard database-backed web2 applications. Current calculated state could contain every custom DataUpdate and state transition that has ever been processed, for example. That's not possible with calculated state as defined as an in-memory object. Calculated state would need to be set in the database as change deltas (new records) indexed by snapshot ordinal and then there would be the requirement of querying the data only in slices (not full object) to get the relevant data you need at that moment. 

So I see two places to improve to make that possible: 
- For `setCalculatedState` we need to address some of the error handling issues that you brought up and add custom `deleteAbove` support. 
- `getCalculatedState` as it's currently implemented is really more like `getLatestCalculatedState` - makes sense, we can keep it and maybe remove support for customizing this. Then we just need additional methods to query state like `getCalculatedStateSlice` or `queryCalculatedState` that would support accessing data from stores that are too large for memory. 

I don't think we need to replace the disk-backed storage because that's also just a cache (for calc state, not snapshots). We don't expect metagraphs to store every calculated state file for all snapshots and we actually need to introduce tools to delete old ones so this storage doesn't grow indefinitely. We're handling that with custom bash scripts right now which is pretty non-ideal.

scasplte2:
> I don't see the synchronization issue as particularly significant because the main problem is that we're missing hooks into at least one important event in the storage lifecycle (rollback). We could just add that hook to solve that problem imo.
> 
> The vision for expanding calculated state to use non-disk data stores is that calculated state could become arbitrarily large, like it is in standard database-backed web2 applications. Current calculated state could contain every custom DataUpdate and state transition that has ever been processed, for example. That's not possible with calculated state as defined as an in-memory object. Calculated state would need to be set in the database as change deltas (new records) indexed by snapshot ordinal and then there would be the requirement of querying the data only in slices (not full object) to get the relevant data you need at that moment.
> 
> So I see two places to improve to make that possible:
> 
> * For `setCalculatedState` we need to address some of the error handling issues that you brought up and add custom `deleteAbove` support.
> * `getCalculatedState` as it's currently implemented is really more like `getLatestCalculatedState` - makes sense, we can keep it and maybe remove support for customizing this. Then we just need additional methods to query state like `getCalculatedStateSlice` or `queryCalculatedState` that would support accessing data from stores that are too large for memory.
> 
> I don't think we need to replace the disk-backed storage because that's also just a cache (for calc state, not snapshots). We don't expect metagraphs to store every calculated state file for all snapshots and we actually need to introduce tools to delete old ones so this storage doesn't grow indefinitely. We're handling that with custom bash scripts right now which is pretty non-ideal.

As a metagraph developer, my preference would be to maintain a single massive SQL database with canonical data that is populated from an indexer of my state as it appears on the hypergraph via the explorer. This seems more cost effective than the idea of hooking into the calculated state methods since then each node needs to maintain their own potentially massive SQL store. The motivation to this change was my own desire to use a disk backed store for calculated state but it was not clear to me how synchronization was maintained so upon investigation I discovered it is being used mostly as a cache. 

My opinion would be that increasing the complexity of the data application interface with additional methods presents an even greater barrier to understanding how to build metagraphs effectively, so my vote would be to simplify the interface and build out indexing as a part of the metagraph stack (along with monitoring) or offer it as a service.

Perhaps even Marcus' idea of a default `getCalculatedStateAtOrdinal` could be used by an indexer when it sees a new metagraph snapshot on the hypergraph. That event could signal to the indexer to pick a randomly selected ML0 peer and query the calculated state to insert into the datastore. Or if it pulls a state that doesn't match the hash included in the snapshot it can try another peer and even potentially inform the previous peer that it is on a fork. Or instead of a default route it could be an Auth'd route to maintain privacy.

scasplte2:
closing due to disinterest

PR #1338, opened 2025-10-10
fix: removing stale process on ConsensusStateAdvancer temporarily

marcus-girolneto:
I've seen in several metagraphs that when we achieve this step, the nodes are turning offline immediately and not able to maybe pick majority and resume the execution. This can explain the large forking events


ryle-goehausen (approved):
Both this and the soft-reset code are broken, I don't think this will make things substantially worse, so ideal to test as improvement. The PR with timeouts main intention was to bypass all of these checks.

scasplte2 (approved):

PR #1343, opened 2025-10-14
fix: remove deprecated guava dependency

effects-ai:


ryle-goehausen:
@effects-ai not urgent, but what is the status of this? I cannot remember where we left off here, is this a dependency conflict? Is it fixable? Maybe could wait til we do the JVM 17 migration?

PR #1344, opened 2025-10-15
fix: fixing storage initialization when the majority peers are set up

marcus-girolneto:


marcus-girolneto:
@greptileai please review it again

Serpion-ua (approved):

PR #1347, opened 2025-10-16
refactor: use blocking operations for serialization tasks

effects-ai:


scasplte2 (approved):
I think this looks good if we know that Brotli needs blocking

PR #1353, opened 2025-10-29
refactor: refactoring GlobalSnapshotAcceptanceManager

marcus-girolneto:


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/global/ArtifactEmissionManager.scala
```diff
@@ -0,0 +1,73 @@
...
+import io.constellationnetwork.schema._
+import io.constellationnetwork.schema.address.Address
+import io.constellationnetwork.schema.artifact._
+import io.constellationnetwork.schema.swap._
+import io.constellationnetwork.schema.tokenLock._
+import io.constellationnetwork.security.Hasher
+import io.constellationnetwork.security.signature.Signed
+import io.constellationnetwork.syntax.sortedCollection.sortedSetSyntax
+
+trait ArtifactEmissionManager[F[_]] {
```
scasplte2, line 18: what is the consideration for whether to create a `trait` with ` make` holding the standard implementation vs. creating a `class` that holds the implementation directly and the make is like a convenience method? Here you went with trait but the Ops managers were all classes. It will work as is, I'm just wondering if there is a rationale.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/currency/BalanceOpsManager.scala
```diff
@@ -0,0 +1,111 @@
...
+import io.constellationnetwork.security.signature.Signed
+
+import eu.timepit.refined.types.numeric.NonNegLong
+import org.typelevel.log4cats.SelfAwareStructuredLogger
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+
+class BalanceOpsManager[F[_]: Async](
+  feeTransactionValidator: FeeTransactionValidator[F]
+) {
+  val logger: SelfAwareStructuredLogger[F] = Slf4jLogger.getLoggerFromName[F]("BalanceOps")
```
scasplte2: I thought you were moving away from hardcoded logger names? Could we use `getLoggerfromClass(this)` instead?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/global/DelegatedStakeStateManager.scala
```diff
@@ -0,0 +1,71 @@
...
+
+trait DelegatedStakeStateManager[F[_]] {
+  def acceptDelegatedStakes(
+    lastSnapshotContext: GlobalSnapshotInfo,
+    epochProgress: EpochProgress,
+    withdrawalTimeLimit: EpochProgress
+  ): (
+    SortedMap[Address, SortedSet[DelegatedStakeRecord]],
+    SortedMap[Address, SortedSet[PendingDelegatedStakeWithdrawal]],
+    SortedMap[Address, SortedSet[PendingDelegatedStakeWithdrawal]]
```
scasplte2, line 20: probably out of scope, but this may benefit from a case class to delineate the purpose of two elements with the same type signature.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/global/NodeCollateralStateManager.scala
```diff
@@ -0,0 +1,129 @@
...
+  def make[F[_]: Async](): NodeCollateralStateManager[F] = new NodeCollateralStateManager[F] {
+
+    def acceptNodeCollaterals(
+      lastSnapshotContext: GlobalSnapshotInfo,
+      epochProgress: EpochProgress,
+      withdrawalTimeLimit: EpochProgress
+    ): (
+      SortedMap[Address, SortedSet[NodeCollateralRecord]],
+      SortedMap[Address, SortedSet[PendingNodeCollateralWithdrawal]],
+      SortedMap[Address, SortedSet[PendingNodeCollateralWithdrawal]]
```
scasplte2, line 50: same here as in previous comment, maybe this could be a shared class as the 
```scala
(
  existing: SortedMap[Address, A],
  unexpired: SortedMap[Address, B],
  expired: SortedMap[Address, B]
)
```
seems to be a repeated pattern

modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshotInfo.scala
```diff
@@ -238,23 +238,32 @@ case class GlobalSnapshotInfo(
...
+        tokenLockBalances.parTraverse(_.hash),
+        lastAllowSpendRefs.parTraverse(_.hash),
+        lastTokenLockRefs.parTraverse(_.hash),
+        updateNodeParameters.parTraverse(_.hash),
+        activeDelegatedStakes.parTraverse(_.hash),
+        delegatedStakesWithdrawals.parTraverse(_.hash),
+        activeNodeCollaterals.parTraverse(_.hash),
+        nodeCollateralWithdrawals.parTraverse(_.hash),
+        priceState.parTraverse(_.hash),
+        metagraphSyncData.parTraverse(_.hash)
```
scasplte2: does parallelization help here? We are dealing with the object as a whole and not iterating over a list of elements so my intuition would say that adding parallelization here adds overhead to this computation. Now if we oculd group (activeAllowSpend, activeTokenLocks, tokenLockBalances, etc) into a list or a tuple and then parallel execution of the hashes within that, then I think it might be applicable.

scasplte2 (commented):
looks good for the most part but I had a couple of questions

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/statechannel/StateChannelValidator.scala
```diff
@@ -52,6 +54,13 @@ object StateChannelValidator {
       case (address, Right((_, info))) => address -> info.lastMessages.toList.flatMap(_.values.map(_.address).toList).toSet
     }
 
+  def getFeeAddresses(
+    lastCurrencySnapshots: SortedMap[Address, Either[Signed[CurrencySnapshot], (Signed[CurrencyIncrementalSnapshot], CurrencySnapshotInfo)]]
```
ryle-goehausen, line 58: I'm not sure we can fix this quite yet, but huge arguments like this are indicative of an issue where we're passing large maps around everywhere instead of having some kind of wrapper type. A getter like this should really belong on some kind of bigger object, rather than being it's own isolated instance that takes a huge argument and then does a small function on it.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/currency/AllowSpendOpsManager.scala
```diff
@@ -0,0 +1,222 @@
...
+      unexpiredAllowSpends = (incomingCurrencyAllowSpends |+| expiredAllowSpends).foldLeft(existentCurrencyAllowSpends) {
+        case (acc, (address, allowSpends)) =>
+          val lastAddressAllowSpends = acc.getOrElse(address, SortedSet.empty[Signed[AllowSpend]])
+          val unexpired = (lastAddressAllowSpends ++ allowSpends).filter(_.value.lastValidEpochProgress >= epochProgress)
+          acc + (address -> unexpired)
+      }
+
+      result <- unexpiredAllowSpends.toList.foldLeftM(unexpiredAllowSpends) {
+        case (acc, (address, allowSpends)) =>
+          allowSpends.toList.traverse(_.toHashed).map { hashedAllowSpends =>
```
ryle-goehausen, line 52: This isn't necessary to change, but an example of potential duplicate hashing here.

ryle-goehausen, line 52: This isn't necessary to change, but an example of potential duplicate hashing here.

ryle-goehausen (dismissed):
Generally I think this is fine as an improvement for now; but in the future I think ideally we want to transition to more trait abstractions and/or getter methods associated with specific instances / traits, rather than static methods with tons of arguments. We have this same issue in a lot of places though.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/global/ArtifactEmissionManager.scala
marcus-girolneto: I'm following the pattern that we already have in other parts, and this makes it flexible, maybe for other implementations?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/currency/BalanceOpsManager.scala
marcus-girolneto: 👍🏼

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/global/DelegatedStakeStateManager.scala
marcus-girolneto: 👍🏼

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/global/NodeCollateralStateManager.scala
marcus-girolneto: 👍🏼

modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshotInfo.scala
marcus-girolneto: no, not it all it was a leftover

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/statechannel/StateChannelValidator.scala
marcus-girolneto: Yeah, maybe we can separate this in another PR futurely

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/currency/AllowSpendOpsManager.scala
marcus-girolneto: Yeah, maybe we can separate this in another PR futurely

marcus-girolneto: Yeah, maybe we can separate this in another PR futurely

scasplte2 (approved):
LGTM 🚀 

I think we may want to revisit the ordinal routing logic at some point as the current pattern in GlobalSnapshotAcceptanceManager doesn't seem the most scalable, but that is a problem for another day.

PR #1354, opened 2025-10-29
feat: implementing custom circe printer and caching snapshot requests

marcus-girolneto:


modules/shared/src/main/scala/io/constellationnetwork/security/Hasher.scala
```diff
@@ -80,8 +80,8 @@ object Hasher {
           case d: Encodable[_] => map(d.toEncode)
           case _               => map(data)
         })
+        .map(Hash.fromBytes)
```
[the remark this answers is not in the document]
marcus-girolneto: Reverting

modules/shared/src/main/scala/io/constellationnetwork/security/Hasher.scala
```diff
@@ -100,7 +100,7 @@ object Hasher {
           JsonSerializer[F].serialize(d.toEncode)(d.jsonEncoder)
         case _ =>
           JsonSerializer[F].serialize[A](data)
-      }).flatMap(Hash.fromBytesForSync[F])
+      }).map(Hash.fromBytes)
```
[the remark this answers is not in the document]
marcus-girolneto: Reverting

marcus-girolneto:
@greptileai please review it again

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/LatestSnapshotCacheRefStorage.scala
```diff
@@ -0,0 +1,101 @@
...
+  override def getByOrdinal(ordinal: SnapshotOrdinal): F[Option[CachedEntry]] =
+    cache.get.map(_.recentSnaps.get(ordinal))
+
+  override def update(snapshot: Signed[S], info: C): F[Unit] =
+    Async[F].defer {
+      val ordinal = snapshot.ordinal
+      val combined = (snapshot, info)
+
+      val snapshotBytes = printer.printToByteArray(snapshot.asJson, StandardCharsets.UTF_8)
+      val combinedBytes = printer.printToByteArray(combined.asJson, StandardCharsets.UTF_8)
```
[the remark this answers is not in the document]
marcus-girolneto: 👍🏼

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/http/routes/SnapshotRoutes.scala
```diff
@@ -33,124 +37,151 @@ final case class SnapshotRoutes[F[_]: Async, S <: Snapshot: Encoder, C: Encoder]
...
+
+  private def whenNodeReady[A](action: F[Response[F]]): F[Response[F]] =
+    nodeStorage.getNodeState.flatMap {
+      case NodeState.Ready => action
+      case _               => ServiceUnavailable(("message" ->> "Node is not ready yet") :: HNil)
+    }
+
+  private def streamJson[A: Encoder](data: A): Stream[F, Byte] = {
+    val bytes = printer.printToByteArray(data.asJson, StandardCharsets.UTF_8)
+    Stream.emits(bytes).covary[F]
```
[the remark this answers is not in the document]
marcus-girolneto: 👍🏼

marcus-girolneto:
@greptileai please review it again

marcus-girolneto:
@greptileai please review it again

marcus-girolneto:
@greptileai please review it again

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/http/routes/SnapshotRoutes.scala
```diff
@@ -33,124 +37,149 @@ final case class SnapshotRoutes[F[_]: Async, S <: Snapshot: Encoder, C: Encoder]
...
-          resolveEncoder[F, Signed[S]](req) { implicit enc =>
-            snapshotStorage.headSnapshot.flatMap {
-              case Some(snapshot) => Ok(snapshot)
-              case _              => NotFound()
-            }
-          },
-          serviceUnavailableNodeNotReady
-        )
+      whenNodeReady {
+        resolveEncoder[F, Signed[S]](req) { _ =>
```
scasplte2: this `resolveEncoder` doesn't seem like it is necessary if there is only a single element in the `List[EntityEncoder[F, A]]` (jsonEncoders above) perhaps it is deprecated due to a new approach but I only see this `SnapshotRoutes` using it and like I said, if only one element then perhaps we could simplify the approach.

But maybe I am missing something here.

modules/shared/src/main/scala/io/constellationnetwork/json/JsonBrotliBinarySerializer.scala
```diff
@@ -19,28 +21,36 @@ trait JsonBrotliBinarySerializer[F[_]] {
 
 object JsonBrotliBinarySerializer {
   private val compressionLevel = 2
+  private val parser = JawnParser(allowDuplicateKeys = false)
```
scasplte2, line 25: is this same behavior enforced by the custom printer?

modules/shared/src/main/scala/io/constellationnetwork/json/JsonSerializer.scala
```diff
@@ -14,7 +14,7 @@ object JsonSerializer {
   def apply[F[_]: JsonSerializer]: JsonSerializer[F] = implicitly
 
   def forSync[F[_]: Sync]: F[JsonSerializer[F]] = {
-    def printer = Printer(dropNullValues = true, indent = "", sortKeys = true)
+    def printer = CustomPrinter(dropNullValues = true, indent = "", sortKeys = true)
```
scasplte2: why is this version of the printer different then the one I get on line 29 in `JsonBrotliBinarySerializer`? https://github.com/Constellation-Labs/tessellation/pull/1354/files#diff-ecab8e1f156b7684528aff8b64d5f1a88f0d10dcbab5046b043b007b4d4f49c3R29

I see that you're just doing what was done before but if we use these two different printers for different ops then perhaps we should indicate their intended usage via a different name? Having different printer behavior with the `forSync` seems like it could lead to confusion

modules/shared/src/main/scala/io/constellationnetwork/merkletree/MerkleTree.scala
```diff
@@ -128,59 +120,34 @@ object MerkleTree {
...
-          prevLevelStart = 0
-        )
-      ) { state =>
-        if (state.levelLen <= 0) {
-          state.mt.asRight[State].pure[F]
-        } else {
-          go(state).map(_.asLeft[MerkleTree])
-        }
-      }
+      go()
```
scasplte2, line 150: why do this? the mutable version could blow up the stack couldn't it? Don't we want to prefer tail-recursion over direct recursion?

modules/shared/src/test/scala/io/constellationnetwork/security/HashSuite.scala
```diff
@@ -79,88 +77,6 @@ object HashSuite extends MutableIOSuite with Checkers {
...
-
-  test("fromBytes and fromBytesForSync produce identical results") { implicit res =>
-    val testData = "Hello, World!".getBytes(StandardCharsets.UTF_8)
-
-    Hash.fromBytesForSync[IO](testData).map { asyncHash =>
-      val syncHash = Hash.fromBytes(testData)
-      expect.eql(asyncHash, syncHash)
-    }
-  }
-
```
scasplte2, line 163: the behavior of the hasher shouldn't change should it? Why remove these tests?

modules/shared/src/main/scala/io/constellationnetwork/json/CustomPrinter.scala
```diff
@@ -0,0 +1,622 @@
...
+      if (required > bytes.length) {
+        val copy = new Array[Byte](math.max(required, bytes.length * 2))
+        System.arraycopy(bytes, 0, copy, 0, index)
+        bytes = copy
+      }
+    }
+
+    def append(c: Char): Appendable = {
+      if (c < 128) {
+        ensureToFit(1)
```
effects-ai: The major difference between the `CustomPrinter` and the stock `Printer` is that the former doesn't try to convert UTF-8 chars into bytes all the time, right?

modules/shared/src/main/scala/io/constellationnetwork/json/CustomPrinter.scala
```diff
@@ -0,0 +1,622 @@
...
+
+import scala.annotation.switch
+
+import io.circe._
+
+/** A pretty-printer for JSON values.
+  *
+  * (Original documentation truncated for brevity)
+  */
+case class CustomPrinter(
```
ryle-goehausen: https://github.com/plokhotnyuk/jsoniter-scala

Have you investigated pre-built libraries and compared them against this? If we can find an optimized library, it's much better than maintaining a solution here.

If this works for now though, seems fine to use it, but I'd much rather find an existing solution. If we can find any SIMD related json libs in Scala (not sure how popular yet,) that's another huge win potentially. But macros to avoid traversal could help us a lot.

ryle-goehausen (commented):
I agree with james re: tests. 

The only other issue I see, there's a lot of Array[Byte] -- it's hard to verify the printer code to see if it's potentially converting large json lists into mutable updates on proper byte classes. That could still be an issue in here even if we're appending element by element.

ryle-goehausen (commented):
I agree with james re: tests. 

The only other issue I see, there's a lot of Array[Byte] -- it's hard to verify the printer code to see if it's potentially converting large json lists into mutable updates on proper byte classes. That could still be an issue in here even if we're appending element by element.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/http/routes/SnapshotRoutes.scala
```diff
@@ -33,124 +37,149 @@ final case class SnapshotRoutes[F[_]: Async, S <: Snapshot: Encoder, C: Encoder]
...
-    ServiceUnavailable(("message" ->> "Node is not ready yet") :: HNil)
-
-  private def validStateForSnapshotReturn(state: NodeState): Boolean = state === NodeState.Ready
+  private val printer = CustomPrinter(dropNullValues = false, indent = "")
+  private val jsonContentType: `Content-Type` =
+    `Content-Type`(MediaType.application.json, Charset.`UTF-8`)
+
+  /** Safe blocking serialization (runs on blocking pool) */
+  private def serializeJson[A: Encoder](value: A): F[Array[Byte]] =
+    Async[F].blocking(printer.printToByteArray(value.asJson, StandardCharsets.UTF_8))
```
ryle-goehausen: Did you compare this directly to sync.delay ? Afaik this is CPU only, so its not 'supposed' to be async, but we may have another bug that makes this better to do async

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/http/routes/SnapshotRoutes.scala
```diff
@@ -33,124 +37,149 @@ final case class SnapshotRoutes[F[_]: Async, S <: Snapshot: Encoder, C: Encoder]
...
+
+  /** Converts an effectful JSON result into a stream */
+  private def streamJson[A: Encoder](data: A): Stream[F, Byte] =
+    Stream.eval(serializeJson(data)).flatMap(bytes => Stream.emits(bytes).covary[F])
```
ryle-goehausen: I think this still emits each individual payload as one chunk, rather than streaming the intermediary outputs.

Also I would double check all the Stream.emits code vs. Stream.chunks ; depending on where it's used it may not be efficient, although I think this usage autoconverts to chunks inside.

PR #1355, opened 2025-10-30
chore: add the latest updates from develop

effects-ai:


ryle-goehausen (approved):
most likely need to review the feature branch independently based on diff afterwards, since this is so large.

PR #1356, opened 2025-10-30
chore: one time mint

effects-ai:


ryle-goehausen (approved):
believe this is correct, should have another approval though to confirm amounts / address.

PR #1364, opened 2025-11-10
feat!: updating to Java 21

marcus-girolneto:


marcus-girolneto:
@greptileai please review

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/node/L0NodeContext.scala
```diff
@@ -38,17 +38,17 @@ object L0NodeContext {
 
     def getLastCurrencySnapshot: F[Option[Hashed[CurrencyIncrementalSnapshot]]] =
       OptionT(snapshotStorage.headSnapshot)
-        .semiflatMap(snapshot => hasherSelector.forOrdinal(snapshot.ordinal)(implicit hasher => snapshot.toHashed))
+        .semiflatMap(snapshot => hasherSelector.withCurrent(implicit hasher => snapshot.toHashed))
```
ryle-goehausen, line 41: Why remove the ordinal here ? Even if we only want one hasher right now, shouldn't we still pass ordinal ?

ryle-goehausen (approved):

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/node/L0NodeContext.scala
marcus-girolneto: It's not removing the ordinal, it's just removing the hasher per ordinal

PR #1373, opened 2025-12-02
feat: adding fsm to consensus

marcus-girolneto:


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/RumorHandlerWithQueue.scala
```diff
@@ -0,0 +1,87 @@
...
+  *       │
+  *       ▼
+  *   queue.offer(RumorReceived(rumor))
+  * }}}
+  */
+object RumorHandlerWithQueue {
+
+  def peer[F[_]: Async, A: TypeTag: Decoder](
+    queue: Queue[F, ConsensusCommand],
+    selfOriginPolicy: OriginPolicy = IncludeSelfOrigin
```
scasplte2, line 53: I only see this handler constructed in `ConsensusRumorHandlers` where the default arg here is always used. Should this handler process messages that originated from itself?

I see the origin policy also used in `RumorHandler` so maybe this is just following the other handlers or maybe I am misunderstanding how the policy is used.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/state/StateTransitions.scala
```diff
@@ -0,0 +1,211 @@
...
+        case Some((_, newState)) =>
+          advancer.getConsensusOutcome(newState) match {
+            case Some((prevKey, outcome)) =>
+              finalizeAndNotify(newState, prevKey, outcome)
+
+            case None =>
+              log.debug(s"State updated for key=$key")
+          }
+      }
+    }
```
scasplte2: stylistic preference but this is slightly easier to read in my opinion
```scala
  def checkUpdate(key: Key): F[Unit] =
    for {
      resources <- storage.getResources(key)
      maybeUpdate <- updater.tryUpdateConsensus(key, resources)
      _ <- maybeUpdate.traverse_ { case (_, newState) =>
        advancer.getConsensusOutcome(newState)
          .map { case (prevKey, outcome) => finalizeAndNotify(newState, prevKey, outcome) }
          .getOrElse(log.debug(s"State updated for key=$key"))
      }
    } yield ()
```

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/state/StateTransitions.scala
```diff
@@ -0,0 +1,211 @@
...
+      outcome <- fetchOutcomeFromCluster(key, artifact, context)
+        .flatMap(_.liftTo[F](new Throwable(s"[DownloadInit] Could not observe outcome for key=$key")))
+      success <- storage.trySetInitialConsensusOutcome(outcome)
+      _ <-
+        if (success)
+          ctx.nodeStorage.tryModifyState(NodeState.Observing, NodeState.WaitingForReady) >>
+            ctx.nodeStorage.setJoiningGracePeriod >>
+            queue.offer(StartRound(none))
+        else
+          new Throwable(s"[DownloadInit] Failed to initialize consensus storage").raiseError[F, Unit]
```
scasplte2: could avoid the intermediate allocation (though it is a minor deal)

```scala
  def initFromDownload(key: Key, artifact: Signed[Artifact], context: Ctx): F[Unit] =
    for {
      _ <- log.info(s"[DownloadInit] Initializing consensus at key=$key")
      outcome <- fetchOutcomeFromCluster(key, artifact, context)
        .flatMap(_.liftTo[F](new Throwable(s"[DownloadInit] Could not observe outcome for key=$key")))
      _ <- storage.trySetInitialConsensusOutcome(outcome).ifM(
        ifFalse = new Throwable(s"[DownloadInit] Failed to initialize consensus storage").raiseError[F, Unit],
        ifTrue = ctx.nodeStorage.tryModifyState(NodeState.Observing, NodeState.WaitingForReady) >>
          ctx.nodeStorage.setJoiningGracePeriod >>
          queue.offer(StartRound(none))
      )
    } yield ()
```

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/state/StateTransitions.scala
```diff
@@ -0,0 +1,211 @@
...
+
+  def initFromRollback(key: Key, outcome: Outcome): F[Unit] =
+    for {
+      _ <- log.info(s"[RollbackInit] Initializing consensus after rollback at key=$key")
+      _ <- storage.trySetInitialConsensusOutcome(outcome)
+      _ <- queue.offer(StartRound(TimeTrigger.some))
+    } yield ()
+
+  private def fetchOutcomeFromCluster(key: Key, artifact: Signed[Artifact], context: Ctx): F[Option[Outcome]] = {
+    val retryPolicy = limitRetries(10).join(constantDelay(3.seconds))
```
scasplte2, line 154: put in conf?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/ConsensusRoundRunner.scala
```diff
@@ -0,0 +1,209 @@
...
+                            Left(nextState)
+                          case None =>
+                            Right(())
+                        }
+                  }
+              }
+          }
+      }
+    Async[F].tailRecM(initialState)(step)
+  }
```
scasplte2: what about ?

```scala

  private def stallDetectionLoop(
    key: Key,
    initialState: ConsensusState[Key, Status, Outcome, Kind]
  ): F[Unit] = {
    type CState = ConsensusState[Key, Status, Outcome, Kind]

    def smartWaitUntil(
      condition: ConsensusResources[Artifact, Kind] => Boolean,
      timeout: FiniteDuration
    ): F[Unit] =
      Stream
        .repeatEval(storage.getResources(key))
        .metered(1.seconds)
        .timeout(timeout)
        .takeThrough(res => !condition(res))
        .compile
        .drain
        .handleError(_ => ())

    def allDeclared(state: CState)(resources: ConsensusResources[Artifact, Kind]): Boolean = {
      val active = state.facilitators.value.toSet -- state.withdrawnFacilitators.value
      resources.peerDeclarationsMap.keySet.size == active.size
    }

    def checkAndContinue: F[Either[CState, Unit]] =
      queue.offer(ConsensusCommand.CheckUpdate(key)) >>
        storage.getState(key).map(_.toLeft(()))

    def handleStall(currentState: CState): F[Either[CState, Unit]] =
      updater.tryLockConsensus(key, currentState).flatMap {
        case Some((_, lockedState)) =>
          logger.debug(s"Stall detection: locked round $key") >>
            smartWaitUntil(allDeclared(lockedState), config.lockDuration) >>
            spreadAckIfCollecting(key, lockedState) >>
            checkAndContinue

        case None =>
          smartWaitUntil(allDeclared(currentState), config.declarationTimeout) >>
            checkAndContinue
      }

    def step(state: CState): F[Either[CState, Unit]] =
      OptionT(storage.getState(key))
        .filter(s => advancer.getConsensusOutcome(s).isEmpty)
        .semiflatMap(handleStall)
        .getOrElse(Right(()))

    Async[F].tailRecM(initialState)(step).handleErrorWith { _ =>
      OptionT(storage.getState(key))
        .semiflatMap(s => Async[F].tailRecM(s)(step))
        .getOrElse(())
    }
  }
```

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/state/ConsensusFSM.scala
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+  private def startRound(trigger: Option[ConsensusTrigger]): F[Unit] =
+    isRunning.get.flatMap {
+      case true => log.debug(s"Ignoring StartRound($trigger) — round already running")
+      case false =>
+        log.info(s"Starting consensus round with trigger=$trigger") >>
+          isRunning.set(true) >>
+          roundRunner.runRound(trigger)
+    }
```
scasplte2:
```scala
  private def startRound(trigger: Option[ConsensusTrigger]): F[Unit] =
    isRunning.get.ifM(
      ifTrue = log.debug(s"Ignoring StartRound($trigger) — round already running"),
      ifFalse = log.info(s"Starting consensus round with trigger=$trigger") >>
        isRunning.set(true) >>
        roundRunner.runRound(trigger)
    )
```

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/state/ConsensusFSM.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  private def completeRound(preAction: F[Unit]): F[Unit] =
+    for {
+      _ <- preAction
+      _ <- isRunning.set(false)
+      next <- pending.pullNext
+      _ <- next.traverse_ {
+        case TriggerPriority.Time  => ctx.queue.offer(StartRound(Some(TimeTrigger)))
+        case TriggerPriority.Event => ctx.queue.offer(StartRound(Some(EventTrigger)))
+      }
```
scasplte2, line 110: if a round takes too long, could multiple Time trigger "pile up" within PendingTriggers?

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotConsensusStateAdvancer.scala
scasplte2, line 1: this is way easier to get an understanding of IMO

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/ConsensusCommand.scala
```diff
@@ -0,0 +1,51 @@
...
+object ConsensusCommand {
+  case class RumorReceived(rumor: Either[PeerRumor[_], CommonRumor[_]]) extends ConsensusCommand
+  case class StartRound(trigger: Option[ConsensusTrigger]) extends ConsensusCommand
+  case object TimeTick extends ConsensusCommand
+  case object FacilitateByEvent extends ConsensusCommand
+  case class CheckUpdate(key: Any) extends ConsensusCommand
+  case object RoundCompleted extends ConsensusCommand
+  case class InternalScheduled(inner: ConsensusCommand) extends ConsensusCommand
+  case class PeerObserved(peer: Peer) extends ConsensusCommand
+  case class InitializeFromDownload(key: Any, artifact: Any, context: Any) extends ConsensusCommand
```
scasplte2, line 46: not really a fan of the Any's and `asInstanceOf` castings. Seems like we're asking for problems but this is an internal mechanism so maybe its alright

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/ConsensusManager.scala
```diff
@@ -0,0 +1,70 @@
...
+  def make[F[_]: Async, Event, Key, Artifact, Context, Status, Outcome, Kind](
+    queue: Queue[F, ConsensusCommand],
+    storage: ConsensusStorage[F, Event, Key, Artifact, Context, Status, Outcome, Kind],
+    nodeStorage: NodeStorage[F]
+  ): F[ConsensusManager[F, Event, Key, Artifact, Context, Status, Outcome, Kind]] =
+    Async[F].pure {
+      new ConsensusManager[F, Event, Key, Artifact, Context, Status, Outcome, Kind] {
+
+        def registerForConsensus(observationKey: Key): F[Unit] =
+          storage.trySetObservationKey(observationKey).flatMap {
```
scasplte2: can replace with `ifM`

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/ConsensusEventLoop.scala
```diff
@@ -0,0 +1,152 @@
...
+          .handleErrorWith(e => Stream.eval(ctx.logger.error(e)("Peer registration failed")))
+
+      val leavingStream: Stream[F, Unit] =
+        nodeStorage.nodeStates
+          .filter(_ === NodeState.Leaving)
+          .evalMap(_ => manager.withdrawFromConsensus)
+          .handleErrorWith(e => Stream.eval(ctx.logger.error(e)("Error handling Leaving state")))
+
+      val run: Stream[F, Unit] =
+        Stream(commandStream, peerRegistrationStream, leavingStream).parJoinUnbounded
```
scasplte2, line 140: nice 🔥

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/RumorHandlerWithQueue.scala
marcus-girolneto: I've just followed the same example of the other rumor handler, so I don't know deeper about it. And yes, this is only used in the consensusHandler, because is the place where we're using the FSM/Queue system

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/state/StateTransitions.scala
marcus-girolneto: 👍🏼

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/state/StateTransitions.scala
marcus-girolneto: 👍🏼

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/state/StateTransitions.scala
marcus-girolneto: I think it's not worth moving to conf, maybe it was left hardcoded because it's a simple thing, just a retry policy

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/ConsensusRoundRunner.scala
marcus-girolneto: I've updated this code, please take a look again

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/state/ConsensusFSM.scala
marcus-girolneto: 👍🏼

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/state/ConsensusFSM.scala
marcus-girolneto: It can, but the handleBusy will handle this, putting the timeTrigger as the next priority
```
  private def handleWhileBusy(cmd: ConsensusCommand): F[Unit] =
    cmd match {
      case FacilitateByEvent             => pending.setEvent()
      case TimeTick                      => pending.setTime()
```

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotConsensusStateAdvancer.scala
marcus-girolneto: 🎉

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/ConsensusCommand.scala
marcus-girolneto: 👍🏼

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/ConsensusManager.scala
marcus-girolneto: 👍🏼

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/ConsensusEventLoop.scala
marcus-girolneto: 🎉

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/state/ConsensusFSM.scala
scasplte2: I see, I was worried that a single session could stack up multiple time triggers but since the flag is just flipped once I think this is a safe way to do it.

scasplte2 (approved):

PR #1376, opened 2025-12-09
chore: upgrade deps and remove unnecessary imports

scasplte2:


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/metrics/Metrics.scala
```diff
@@ -330,7 +329,7 @@ object Metrics {
           genericRecordDistribution(key, value, tagsWithBucket)
         }
         def getAllAsText: F[String] = Async[F].delay {
-          registry.scrape(TextFormat.CONTENT_TYPE_OPENMETRICS_100)
+          registry.scrape("application/openmetrics-text; version=1.0.0; charset=utf-8")
```
scasplte2, line 332: previous string enum was made private, string value here follows documentation suggestion from
https://docs.micrometer.io/micrometer/reference/implementations/prometheus.html#_scrape_format

marcus-girolneto (dismissed):
LGTM

marcus-girolneto (approved):

PR #1378, opened 2025-12-10
fix: adding default ports for nodepilot

marcus-girolneto:


scasplte2 (approved):
This is makes the l1 layer a required arg right? If so, won't that impact all startup scripts and runtime configurations?

scasplte2 (commented):
apologies, I did not mean to hit approve

marcus-girolneto:
> This is makes the l1 layer a required arg right? If so, won't that impact all startup scripts and runtime configurations?

Not exactly, the L1 layers will decide automatically which layer we should provide. DAGL1 by default is always DAGL1, and the currencyL1, if we provide the dataApplication it will be DATAL1, and if not CURRENCYL1.
It shouldn't impact it all on startup; it's just one automation if we didn't want to provide the ports and follow the pattern of node pilot

scasplte2:
ok, if no impact to startup then I think there is no problem :+1:

PR #1380, opened 2025-12-11
fix: updating fork detection to use snapshot hash

marcus-girolneto:


modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensusStateAdvancer.scala
```diff
@@ -204,7 +205,7 @@ object CurrencySnapshotConsensusStateAdvancer {
       )(implicit hasher: Hasher[F]): F[Option[Transition]] =
         for {
           maybeProposals <- maybeGetAllDeclarations(state, resources)(_.proposal)
-          _ <- maybeProposals.traverse_(checkForkByFacilitatorsHash(_, status.facilitatorsHash))
+          _ <- maybeProposals.traverse_(checkForkByLastSnapshotHash(_, status.lastSnapshotHash))
```
scasplte2, line 208: Not specific to this change, but why are we returning `F[Unit]` here? Shouldn't the fork gate proceeding to the next step and if a node finds it is forked it drops out of the current round and not progress to signature phase?

agent analysis
---
recoverIfForking does:
  1. Checks if this node's hash differs from the majority
  2. If forked, it starts the fork recovery in a background fiber (Temporal[F].start(forkRecovery).void on line 253)
  3. Returns F[Unit] immediately without waiting

  Since the fork recovery is fire-and-forget. After checkForkByLastSnapshotHash returns, the consensus continues to toProposalsPhase even though this node just detected it's forked. The recovery process (setting state to Leaving → Offline → potentially restarting) runs concurrently in the background.

  This means a forked node will:
  1. Detect the fork
  2. Continue participating in consensus (potentially creating/signing invalid artifacts)
  3. Eventually get shut down by the background fiber

  Why this might be intentional (but still questionable):
  - The Leaving state should theoretically prevent the node from being included in peer lists
  - Other nodes might reject proposals from a node in Leaving state

  Why it's probably problematic:
  - There's a race condition window where the forked node continues processing
  - The traverse_ discards the result, so there's no way to short-circuit the for-comprehension based on fork detection
  - If you want to halt consensus participation immediately upon fork detection, you'd need to raise an error or return an Option/Either that
  stops the flow

scasplte2 (commented):
change looks fine but I don't follow how the fork detection scheme is meant to actually function

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensusStateAdvancer.scala
marcus-girolneto: It seems that we handle it here
```
        val forkRecovery = Slf4jLogger
          .getLogger[F]
          .warn(s"Different hash observations [$observationName]. This node is in fork") >>
          nodeStorage.setNodeState(NodeState.Leaving) >>
          Temporal[F].sleep(leavingDelay) >>
          nodeStorage.setNodeState(NodeState.Offline) >>
          Temporal[F].sleep(5.seconds) >>
          ExitOnFork.exitOnFeature("CL_EXIT_ON_FORK") >>
          restartService.signalNodeForkedRestart(majorityForkPeers)
        ```

marcus-girolneto: It seems that it sends a signal to stop the current execution

scasplte2: yeah, but that is spawning a separate thread that will not prevent the first one from continuing, because we don't wait on a value to return from the intermediate computation the runner thread just fires it off then we have a race condition. I think I'll just try to address in another PR as I used the modification here to improve the entropy in the facilitators subset PR (which seems to be working now)

scasplte2 (approved):

PR #1392, opened 2026-01-12
refactor: extend global state proof with mptRoot: Option[Hash]

scasplte2:
This PR refactors the previous global state proof selector migration that introduced new case class types for `GlobalSnapshotInfo, GlobalIncrementalSnapshot, & GlobalSnapshotStateProof` and instead extends the previous `GlobalSnapshotStateProof` with a backwards compatible `mptRoot: Option[Hash]` parameter. Since circe nulls the optional field by default, this refactored approached avoids a multitude of schema migration issues in the persistence, downloading, and validation of snapshot data prior to the hard fork. Additionally, the use of a `StateProofSelector` allows the network to perform a "live" transition over the state proof transition boundary during normal chain operations.

This forking functionality has been successfully tested in a local cluster with 3 nodes under various scenarios
1. Genesis network (no historical data, chain begins from genesis and progresses through the fork)
2. Coordinated fork (all nodes have historical data up to the transition ordinal, then restart and continue progress post fork)
3. Partial coordinated fork (some nodes have all historical data up to the transition ordinal, other nodes may be further behind or have no data, upon restart the network progress post fork)

This work also adds `run-rollback` as an option with the `just` launcher to allow local clusters to resume a building a chain from data available on disk.


marcus-girolneto (approved):
LGTM 🔥

PR #1398, opened 2026-01-21
fix: improving mpt logic

marcus-girolneto:


modules/shared/src/main/scala/io/constellationnetwork/security/mpt/MerklePatriciaTrie.scala
```diff
@@ -2,22 +2,111 @@ package io.constellationnetwork.security.mpt
...
+    } yield new MerklePatriciaNode.Branch(updatedPaths, Some(hash))
+
+  private def computeExtensionHash[F[_]: Async: Hasher: Parallel](
+    ext: MerklePatriciaNode.Extension,
+    counter: cats.effect.Ref[F, Int]
+  ): F[MerklePatriciaNode] =
+    for {
+      // First compute child's hash
+      updatedChild <- computeAndCacheHash[F](ext.child, counter)
+      updatedBranch = updatedChild.asInstanceOf[MerklePatriciaNode.Branch]
```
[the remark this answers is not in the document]
marcus-girolneto: No, om tjos case it will be branch

modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshotInfo.scala
```diff
@@ -191,14 +208,17 @@ case class GlobalSnapshotInfo(
...
-      case MerklePatriciaFormat =>
-        statefulMPTProducer.build.flatMap {
-          case Left(value) =>
+  ): F[GlobalSnapshotStateProof] = stateProofSelector.select(ordinal) match {
+    case LegacyFormat =>
+      lastCurrencySnapshots.merkleTree[F].flatMap(stateProof(_))
+    case MerklePatriciaFormat =>
+      statefulMPTProducer.build.flatMap {
+        case Left(value) =>
+          stateProof(ordinal).handleErrorWith { _ =>
```
scasplte2: This looks problematic to call `stateProof(ordinal)` here again (infinite loop) but I also think it could be unnecessary. If you tried to build the `statefulMPT` and got an error than you can just handle the error directly without calling `stateProof(ordinal)` again.

Furthermore, `MerklePatriciaError` already extends Throwable so there doesn't seem to be a need to unwrap and rewrap the error since you can just `.raiseError` directly on the Left value. Like

```scala
    case MerklePatriciaFormat =>
      statefulMPTProducer.build.flatMap {
        case Left(err) => err.raiseError[F, GlobalSnapshotStateProof]
        case Right(trie) =>
          MerklePatriciaTrie.getRootHash[F](trie).map { mptRoot =>
            GlobalSnapshotStateProof
              .apply(
                Hash.empty,
                Hash.empty,
                Hash.empty,
                None,
                None,
                None,
                None,
                None,
                None,
                None,
                None,
                None,
                None,
                None,
                None,
                None,
                Some(mptRoot.value)
              )
          }
      }
```

modules/shared/src/main/scala/io/constellationnetwork/validator/StateProofValidator.scala
scasplte2, line 1: I noticed that this file seem inappropriately named and placed. Perhaps it was changed from `MerkleTreeValidator` to `StateProofValidator` at some previous commit but since we're modifying it here perhaps we can clean this up and refactor it to a place that makes more sense.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/domain/snapshot/programs/Download.scala
```diff
@@ -371,8 +370,21 @@ object Download {
...
+                _ <- logger.info("Performing initial sync of MPT")
+                kvPairs <- hasherSelector.withCurrent(implicit h => context.allStateEntries[F])
+                _ <- mptStore.syncFull(kvPairs, lastSnapshot.ordinal)
+              } yield ()
+            } else {
+              ().pure[F]
+            }
+          _ <- persistLastSnapshot
+          result <- processNextOrFinish
+        } yield result
```
scasplte2, line 386: what do you think about this refactor? Makes it a little more clear what is going on IMO
```scala
def performInitialSync: F[Unit] =
  for {
    _       <- logger.info("Performing initial sync of MPT")
    kvPairs <- hasherSelector.withCurrent(implicit h => context.allStateEntries[F])
    _       <- mptStore.syncFull(kvPairs, lastSnapshot.ordinal)
  } yield ()

for {
  mptEntries <- mptStore.underlying.entries
  _          <- performInitialSync.whenA(mptEntries.isEmpty)
  _          <- persistLastSnapshot
  result     <- processNextOrFinish
} yield result
```

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotTraverse.scala
```diff
@@ -110,8 +110,8 @@ object GlobalSnapshotTraverse {
 
           firstInfoCalculatedProof <- HasherSelector[F].withCurrent { implicit hasher =>
             hasher.getLogic(firstInc.ordinal) match {
-              case KryoHash => GlobalSnapshotInfoV2.fromGlobalSnapshotInfo(firstInfo).stateProof(firstInc.ordinal)
-              case JsonHash => firstInfo.stateProof(firstInc.ordinal)
+              case KryoHash => GlobalSnapshotInfoV2.fromGlobalSnapshotInfo(firstInfo).stateProof(mptStore.underlying, firstInc.ordinal)
```
scasplte2: probably unavoidable but this is a bit confusing since the `KryoHash` era uses the legacy state proof format and not the MPT

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/SnapshotDownloadStorage.scala
```diff
@@ -77,23 +80,33 @@ object SnapshotDownloadStorage {
...
-                  .raiseError[F, Option[(Signed[GlobalIncrementalSnapshot], GlobalSnapshotInfo)]]
-              )
+            for {
+              kvPairs <- info match {
+                case Left(value)  => value.toGlobalSnapshotInfo.allStateEntries[F]
+                case Right(value) => value.allStateEntries[F]
+              }
+              _ <- mptStore.syncFull(kvPairs, ordinal)
+              result <- info
+                .bitraverse(_.stateProof(mptStore.underlying, ordinal), _.stateProof(mptStore.underlying, ordinal))
```
scasplte2: same issue here with applying the updated state proof function signature to the legacy formats. A Left return from the ops above should not need to compute `allStateEntries` since it uses the legacy format. Maybe the simplest solution is to add a `stateProofLegacy` and `stateProofMPT` methods to `StateProofValidator`

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotContextFunctions.scala
```diff
@@ -23,7 +24,7 @@ abstract class CurrencySnapshotContextFunctions[F[_]]
     extends SnapshotContextFunctions[F, CurrencyIncrementalSnapshot, CurrencySnapshotContext]
 
 object CurrencySnapshotContextFunctions {
-  def make[F[_]: Async: Parallel](validator: CurrencySnapshotValidator[F])(
+  def make[F[_]: Async: Parallel](validator: CurrencySnapshotValidator[F], mptStore: MptStore[F, GlobalStateKey])(
```
scasplte2: this seems like we're leaking abstractions here as the Currency consensus process (CurrencySnapshotInfo, etc.) doesn't use the MPT state proof mechanism and would need a dedicated CurrencyStateKey type if we did want to have the metagraphs use MPT state proofs

modules/shared/src/main/scala/io/constellationnetwork/currency/schema/currency.scala
scasplte2, line 1: I don't think these modifications are necessary since currency consensus uses an alternative state proof protocol

modules/shared/src/main/scala/io/constellationnetwork/schema/mpt/GlobalStateConverter.scala
```diff
@@ -232,28 +233,48 @@ object GlobalStateConverter {
...
+                              case (key, value) => GlobalStateKey.toHex[F](key).map(_ -> value)
+                            }
+                            newAcc = acc ++ batchResult.toMap
+                            _ <- Async[F].cede
+                          } yield newAcc
+                      }
+                  }
+                trie <- MerklePatriciaTrie.makeParallel[F, Json](hexMap)
+                root <- MerklePatriciaTrie.getRootHash[F](trie)
+              } yield root
```
scasplte2: Perhaps a slightly more readable version

```scala
def toHexMap(kvPairs: Map[GlobalStateKey, Json]): F[Map[Hex, Json]] = {
  def convertBatch(batch: List[(GlobalStateKey, Json)]): F[List[(Hex, Json)]] =
    batch.parTraverse { case (key, value) => 
      GlobalStateKey.toHex[F](key).map(_ -> value) 
    }

  if (kvPairs.size <= BatchSize)
    convertBatch(kvPairs.toList).map(_.toMap)
  else
    kvPairs.toList
      .grouped(BatchSize)
      .toList
      .foldLeftM(Map.empty[Hex, Json]) { (acc, batch) =>
        convertBatch(batch).map(acc ++ _) <* Async[F].cede
      }
}

def buildMptRoot(hexMap: Map[Hex, Json]): F[MptRoot] =
  for {
    trie <- MerklePatriciaTrie.makeParallel[F, Json](hexMap)
    root <- MerklePatriciaTrie.getRootHash[F](trie)
  } yield root

for {
  kvPairs <- kvPairsF
  mptRoot <- 
    if (kvPairs.isEmpty)
      logger.info("Empty map, returning empty hash").as(MptRoot(Hash.empty))
    else
      toHexMap(kvPairs).flatMap(buildMptRoot)
} yield mptRoot
```

modules/shared/src/main/scala/io/constellationnetwork/schema/mpt/GlobalStateConverter.scala
```diff
@@ -266,8 +287,25 @@ object GlobalStateConverter {
...
       def syncFromStateChanges(acc: StateChangesAccumulator, snapshotOrdinal: SnapshotOrdinal)(
         implicit stateProofSelector: StateProofSelector
-      ): F[Unit] =
-        acc.toStateEntries[F].flatMap(store.sync[Json](_, snapshotOrdinal))
+      ): F[Unit] = {
+        val BatchSize = 5000
+
+        for {
+          entries <- acc.toStateEntries[F]
+          syncStart <- Async[F].realTime
```
scasplte2: debug remnant?

modules/shared/src/main/scala/io/constellationnetwork/schema/mpt/MptStore.scala
```diff
@@ -1,102 +1,271 @@
...
   def remove(keys: List[K]): F[Unit]
-
   def contains(key: K): F[Boolean]
-
   def clear: F[Unit]
-
   def build: F[Either[MerklePatriciaError, MerklePatriciaTrie]]
 
-  def sync[V: Encoder](newState: Map[K, V], ordinal: SnapshotOrdinal): F[Unit]
+  def buildWithRootHash: F[Either[MerklePatriciaError, (MerklePatriciaTrie, MptRoot)]]
```
scasplte2: why not just modify build?

modules/shared/src/main/scala/io/constellationnetwork/schema/mpt/MptStore.scala
```diff
@@ -1,102 +1,271 @@
...
+    for {
+      trieRef <- Ref.of[F, Option[MerklePatriciaTrie]](None)
+    } yield new Impl[F, K](producer, toHex, trieRef)
 
-  private final class Impl[F[_]: Async: Parallel: Hasher, K](
+  private final class Impl[F[_]: Async: Parallel: Hasher: JsonSerializer, K](
     producer: StatefulMerklePatriciaProducer[F],
-    toHex: K => F[Hex]
+    toHex: K => F[Hex],
+    trieRef: Ref[F, Option[MerklePatriciaTrie]]
```
scasplte2: this seems repetitive as your implementations of `StatefulMerklePatriciaProducer` each have a `Ref[F, Option[MerklePatriciaTrie]]` already so this would cause two different classes to hold different Refs containing the trie. One option might be to expose `trieRef` in the trait definition of the stateful producer. to just reuse it in the Impl instead of making another Ref.

modules/shared/src/main/scala/io/constellationnetwork/schema/mpt/MptStore.scala
```diff
@@ -1,102 +1,271 @@
...
 
-    private def persistAndCutoff(ordinal: SnapshotOrdinal): F[Unit] =
+    private val logger = Slf4jLogger.getLoggerFromName[F](this.getClass.getName)
+    private val BatchSize = 5000
+
+    private def persistAndCutoffAsync(ordinal: SnapshotOrdinal): F[Unit] =
       producer match {
-        case p: StatefulWithPersistenceMerklePatriciaProducer[F] =>
-          p.persist(ordinal) >> p.applyCutoff(ordinal)
+        case p: StatefulWithPersistenceMerklePatriciaProducer[F @unchecked] =>
```
scasplte2: please avoid unchecked, why not require the more specific type for producer in the function signature, since it extends the original Stateful producer then it should have all the same methods.

modules/shared/src/main/scala/io/constellationnetwork/schema/mpt/MptStore.scala
```diff
@@ -1,102 +1,271 @@
...
+          Async[F].start {
+            p.persist(ordinal) >> p.applyCutoff(ordinal)
+          }.void
         case _ =>
-          Async[F].unit
+          ().pure
+      }
+
+    /** Serialize value directly to bytes without intermediate Json AST */
+    private def serializeToBytes[V: Encoder](v: V): F[Array[Byte]] =
```
scasplte2: we might want to ask @ryle-goehausen about this but I would prefer to use the `JsonSerializer` byte serialization instead of defining a different binary encoding for the trie.

modules/shared/src/main/scala/io/constellationnetwork/schema/mpt/MptStore.scala
```diff
@@ -1,102 +1,271 @@
...
+            } <* Async[F].cede
+          }
+      }
+
+    private def deserializeBytes[V: Decoder](bytes: Array[Byte]): F[Option[V]] =
+      if (bytes == null || bytes.isEmpty) {
+        logger.warn("Attempted to deserialize null or empty bytes") >>
+          none[V].pure[F]
+      } else {
+        JsonSerializer[F].deserialize[Json](bytes).flatMap {
```
scasplte2, line 92: Does this work with the different serializer used above?

modules/shared/src/main/scala/io/constellationnetwork/security/mpt/producer/FileSystemMerklePatriciaProducer.scala
```diff
@@ -137,23 +217,108 @@ class FileSystemMerklePatriciaProducer[F[_]: Async: Parallel: Hasher](
...
+      loaded <- storage.readTrie(ordinal).flatMap {
+        case Some((trie, state)) =>
+          for {
+            _ <- logger.info(s"Found persisted trie with ${state.size} entries")
+            _ <- stateRef.set(state)
+            _ <- trieRef.set(Some(trie))
+            _ <- pendingInsertsRef.set(Map.empty)
+            _ <- pendingRemovesRef.set(List.empty)
+            end <- Async[F].realTime
+            _ <- logger.info(s"Trie loaded successfully in ${(end - start).toMillis}ms")
```
scasplte2: remnant debug logs?

modules/shared/src/main/scala/io/constellationnetwork/security/mpt/producer/InMemoryMerklePatriciaProducer.scala
```diff
@@ -4,158 +4,211 @@ import cats.Parallel
...
+  ): F[Either[MerklePatriciaError, MerklePatriciaTrie]] =
+    for {
+      afterRemoves <-
+        if (removes.isEmpty) currentTrie.asRight[MerklePatriciaError].pure[F]
+        else parallelProducer.remove(currentTrie, removes)
+
+      result <- afterRemoves match {
+        case Left(err) =>
+          val errResult: Either[MerklePatriciaError, MerklePatriciaTrie] = err.asLeft[MerklePatriciaTrie]
+          errResult.pure[F]
```
scasplte2: can use right bias of `.map` to avoid this pattern matching
```scala
result <- afterRemoves.map { trieAfterRemoves => 
       ...
    }
```

modules/shared/src/main/scala/io/constellationnetwork/security/mpt/producer/ParallelMerklePatriciaProducer.scala
```diff
@@ -3,99 +3,111 @@ package io.constellationnetwork.security.mpt.producer
...
+        hashEnd <- Async[F].realTime
+        _ <- logger.info(s"Phase 2 - batchComputeDataHashesFromBytes took ${(hashEnd - hashStart).toMillis}ms")
+
+        buildStart <- Async[F].realTime
+        root <- buildTreeFromBytes(entries, dataHashes, 0, entries.length, 0)
+        buildEnd <- Async[F].realTime
+        _ <- logger.info(s"Phase 3 - buildTreeFromBytes took ${(buildEnd - buildStart).toMillis}ms")
+
+        totalEnd <- Async[F].realTime
+        _ <- logger.info(s"Total MPT creation (bytes) took ${(totalEnd - totalStart).toMillis}ms")
```
scasplte2: can cleanup timing here or add to logger.debug perhaps

modules/shared/src/main/scala/io/constellationnetwork/security/mpt/producer/ParallelMerklePatriciaProducer.scala
```diff
@@ -104,117 +116,623 @@ class ParallelMerklePatriciaProducer[F[_]: Hasher: Async: Parallel] extends Merk
...
+      case ext: MerklePatriciaNode.Extension =>
+        MerklePatriciaNode.Extension.fromCompact[F](newPath ++ ext.sharedPath, ext.child).widen
+      case b: MerklePatriciaNode.Branch =>
+        MerklePatriciaNode.Extension.fromCompact[F](newPath, b).widen
+    }
+  }
+
+  // ==================== Legacy Build Tree (Json-based) ====================
+
+  private def batchComputeDataHashes(entries: Array[(CompactNibblePath, Json)]): F[Array[String]] = {
```
scasplte2: my preference would be to standardize around Array[Byte] and just remove the Json methods

modules/shared/src/main/scala/io/constellationnetwork/security/mpt/producer/InMemoryMerklePatriciaProducer.scala
```diff
@@ -4,158 +4,200 @@ import cats.Parallel
...
+                err.asLeft[MerklePatriciaTrie].pure[F]
+              case Right(finalTrie) =>
+                (trieRef.set(Some(finalTrie)) >>
+                  pendingInsertsRef.set(Map.empty) >>
+                  pendingRemovesRef.set(List.empty)).as(finalTrie.asRight[MerklePatriciaError])
+            }
+          }
       }
-    }
+    } yield result
```
scasplte2: what do you think about

```scala
  private def incrementalUpdate(
                                 currentTrie: MerklePatriciaTrie,
                                 inserts: Map[Hex, Array[Byte]],
                                 removes: List[Hex]
                               ): F[Either[MerklePatriciaError, MerklePatriciaTrie]] = {

    def applyRemoves(trie: MerklePatriciaTrie): EitherT[F, MerklePatriciaError, MerklePatriciaTrie] =
    if (removes.isEmpty) EitherT.rightT(trie)
    else EitherT(parallelProducer.remove(trie, removes))

    def applyInserts(trie: MerklePatriciaTrie): EitherT[F, MerklePatriciaError, MerklePatriciaTrie] =
      if (inserts.isEmpty) EitherT.rightT(trie)
      else EitherT(parallelProducer.insertFromBytes(trie, inserts))

    def clearPending(trie: MerklePatriciaTrie): EitherT[F, MerklePatriciaError, Unit] =
      EitherT.liftF(
        trieRef.set(Some(trie)) >>
          pendingInsertsRef.set(Map.empty) >>
          pendingRemovesRef.set(List.empty)
      )

    (for {
      afterRemoves <- applyRemoves(currentTrie)
      finalTrie    <- applyInserts(afterRemoves)
      _            <- clearPending(finalTrie)
    } yield finalTrie).value
  }
```

EitherT will let you avoid matching on the Left and just rewrapping since its is right biased.

modules/shared/src/main/scala/io/constellationnetwork/security/mpt/producer/StatelessMerklePatriciaProducer.scala
```diff
@@ -1,52 +1,180 @@
...
+    if (data.isEmpty) {
+      current.asRight[MerklePatriciaError].pure[F]
+    } else {
+      (for {
+        dataStoreRef <- Ref.of[F, Map[Hash, Array[Byte]]](current.dataStore)
+        result <- insertMultipleWithStore(current.trie.rootNode, data.toList, dataStoreRef)
+        finalDataStore <- dataStoreRef.get
+      } yield result.map(node => MerklePatriciaTrieWithData(MerklePatriciaTrie(node), finalDataStore)))
+        .handleError(e => OperationError(e.getMessage).asLeft)
+    }
```
scasplte2: looks like you could remove these as they are no in use

modules/shared/src/main/scala/io/constellationnetwork/security/mpt/storages/MptStateStorage.scala
```diff
@@ -0,0 +1,105 @@
...
+
+  // Codecs for Map[Hex, Array[Byte]]
+  private implicit val byteArrayEncoder: Encoder[Array[Byte]] =
+    Encoder.encodeString.contramap(Base64.getEncoder.encodeToString)
+
+  private implicit val byteArrayDecoder: Decoder[Array[Byte]] =
+    Decoder.decodeString.emap { str =>
+      try Right(Base64.getDecoder.decode(str))
+      catch { case e: IllegalArgumentException => Left(s"Invalid Base64: ${e.getMessage}") }
+    }
```
scasplte2: I don't understand why we need Encoder, Decoder instances for Array[Byte]. Shouldn't we just use JsonSerializer to go from Object -> Bytes and vice versa?

modules/shared/src/main/scala/io/constellationnetwork/security/mpt/MerklePatriciaNode.scala
```diff
@@ -19,32 +17,100 @@ import io.circe.syntax.EncoderOps
...
 case class MptRoot(value: Hash) extends AnyVal
 
+/** Immutable Merkle Patricia Trie node with pre-computed digest. Hash is computed at construction time - no caching or invalidation needed.
+  *
+  * MEMORY OPTIMIZATION: Paths are stored as CompactNibblePath (2 nibbles per byte) instead of Seq[Nibble] (1 object per nibble). This
+  * reduces memory by ~20-40x for path storage.
+  */
 sealed trait MerklePatriciaNode {
   def digest: Hash
+  def dataDigest: Hash
```
scasplte2: I don't think you want to add `dataDigest` to the node trait, it should only apply to the Leaf. This will reduce the memory footprint of intermediate nodes because they don't need Hash.empty in Braches or extensions.

scasplte2 (changes requested):

modules/shared/src/main/scala/io/constellationnetwork/validator/StateProofValidator.scala
marcus-girolneto: 👍🏼

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/domain/snapshot/programs/Download.scala
marcus-girolneto: 👍🏼

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotTraverse.scala
marcus-girolneto: 👍🏼

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/SnapshotDownloadStorage.scala
marcus-girolneto: 👍🏼

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotContextFunctions.scala
marcus-girolneto: I think we can remove

modules/shared/src/main/scala/io/constellationnetwork/currency/schema/currency.scala
marcus-girolneto: We need this because this implements the `stateProof` functions, where I've added more functions and changed the signature. Currently, the new functions just call the old ones

modules/shared/src/main/scala/io/constellationnetwork/schema/mpt/GlobalStateConverter.scala
marcus-girolneto: 👍🏼

modules/shared/src/main/scala/io/constellationnetwork/schema/mpt/GlobalStateConverter.scala
marcus-girolneto: yes, thanks

modules/shared/src/main/scala/io/constellationnetwork/schema/mpt/MptStore.scala
marcus-girolneto: This can be removed

modules/shared/src/main/scala/io/constellationnetwork/schema/mpt/MptStore.scala
marcus-girolneto: 👍🏼

modules/shared/src/main/scala/io/constellationnetwork/schema/mpt/MptStore.scala
marcus-girolneto: 👍🏼

modules/shared/src/main/scala/io/constellationnetwork/schema/mpt/MptStore.scala
marcus-girolneto: 👍🏼

modules/shared/src/main/scala/io/constellationnetwork/schema/mpt/MptStore.scala
marcus-girolneto: 👍🏼

modules/shared/src/main/scala/io/constellationnetwork/security/mpt/producer/FileSystemMerklePatriciaProducer.scala
marcus-girolneto: Yes

modules/shared/src/main/scala/io/constellationnetwork/security/mpt/producer/InMemoryMerklePatriciaProducer.scala
marcus-girolneto: 👍🏼

modules/shared/src/main/scala/io/constellationnetwork/security/mpt/producer/ParallelMerklePatriciaProducer.scala
marcus-girolneto: 👍🏼

modules/shared/src/main/scala/io/constellationnetwork/security/mpt/producer/ParallelMerklePatriciaProducer.scala
marcus-girolneto: 👍🏼

modules/shared/src/main/scala/io/constellationnetwork/security/mpt/producer/InMemoryMerklePatriciaProducer.scala
marcus-girolneto: 👍🏼

modules/shared/src/main/scala/io/constellationnetwork/security/mpt/producer/StatelessMerklePatriciaProducer.scala
marcus-girolneto: 👍🏼

modules/shared/src/main/scala/io/constellationnetwork/security/mpt/storages/MptStateStorage.scala
marcus-girolneto: You're right, it was a leftover

modules/shared/src/main/scala/io/constellationnetwork/security/mpt/MerklePatriciaNode.scala
marcus-girolneto: 👍🏼

scasplte2:
@greptileai review the latest changes as if it is a new PR

scasplte2 (dismissed):

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/StateChannel.scala
```diff
@@ -118,85 +112,111 @@ object StateChannel {
...
             _ <- enqueueConsensusEventFn(globalSyncEvent).run()
             _ <- storages.lastGlobalSnapshotSync.set(globalSyncEvent.value)
           } yield ()
         case (Some(_), None) =>
-          if (!isStartupCall) {
-            Logger[F].warn("Couldn't send GlobalSnapshotSyncEvent. Session is missing.")
-          } else {
-            ().pure
-          }
+          Logger[F].warn("Couldn't send GlobalSnapshotSyncEvent. Session is missing.")
```
scasplte2: can we use `logger.warn(...)` instead of the `Logger` type constraint. Logger[F] labels logs as coming from where the Logger instance is created. Using the logger val will put the logs labeled as coming from this class (easier to debug)

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/StateChannel.scala
```diff
@@ -118,85 +112,111 @@ object StateChannel {
...
-          Logger[F].error(
-            "Couldn't construct GlobalSnapshotSyncEvent. Last sent reference and session are missing"
-          )
+          Logger[F].error("Couldn't construct GlobalSnapshotSyncEvent. Last sent reference and session are missing")
       }
     }
 
+    def ensureMptInitialized(ordinal: SnapshotOrdinal, state: GlobalSnapshotInfo): F[Unit] =
+      sharedStorages.mptStore.isEmpty.ifM(
+        HasherSelector[F].withCurrent { implicit hasher =>
```
scasplte2: could we maybe just add `Hasher` as a type constraint to the `make` function instead of generating a new hasher in multiple places, not really a big deal though

scasplte2 (approved):
🚀

PR #1402, opened 2026-01-26
feat: initial working event mempool

scasplte2:


ryle-ai (commented):
## Review Summary

Great architectural improvement decoupling event gossip from consensus rounds. The hash-based coordination is cleaner than the ordinal-bound approach.

### Key Concerns

| Severity | Issue |
|----------|-------|
| **HIGH** | TTL eviction not implemented - `maxEventAge` config exists but is never enforced. Mempool can fill with stuck events → liveness risk |
| **HIGH** | Unanimous intersection - Single slow node blocks all events from inclusion |
| MEDIUM | Rate limiting missing on gossip endpoints |
| MEDIUM | Gossip polling frequency should be environment-configurable |
| LOW | Observability - recommend ClickHouse structured event tracing |

### Positive Notes
- Auth properly applied via `PeerAuthMiddleware` chain
- Clean generic abstractions (`EventMempool[F, Event, Key]`)

See inline comments for details.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/mempool/EventMempool.scala
```diff
@@ -0,0 +1,259 @@
...
+    * @param events
+    *   The signed events to add
+    * @return
+    *   List of results in the same order as input
+    */
+  def addBatch(events: List[Signed[Event]]): F[List[Either[MempoolRejectionReason, MempoolEntry[Event, Key]]]]
+
+  /** Get all event hashes currently in the mempool.
+    *
+    * Used for hash-based consensus bounds - each facilitator declares which events they have by hash.
```
ryle-ai: I think the TTL eviction logic isn't actually implemented here -- `maxEventAge` configuration exists and `MempoolEntry` stores `receivedAt: Instant`, but there's no code that uses them for eviction. Events are only removed via `clearIncluded()` (after consensus successfully includes them) or `remove()` (explicit removal).

The concern is that if an event never gets included in consensus -- for example because unanimous intersection always fails since one node never receives it -- that event stays in the mempool indefinitely. Once `maxSize` is reached, new legitimate events get rejected with `MempoolFull`. That's a liveness issue since the system could stop accepting new events entirely.

I think this needs either a background eviction task that removes events older than `maxEventAge`, or at least documentation explaining why it's acceptable to defer this.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotConsensusStateAdvancer.scala
```diff
@@ -150,26 +152,48 @@ object GlobalSnapshotConsensusStateAdvancer {
       state: GlobalSnapshotConsensusState,
       facilities: SortedMap[PeerId, Facility]
     ): F[Option[Transition]] = {
-      val (bound, candidates, triggers) = facilities.foldMap(f => (f.upperBound, f.candidates.value, f.trigger.toList))
+      val (candidates, triggers) = facilities.foldMap(f => (f.candidates.value, f.trigger.toList))
+
+      // Compute hash intersection for consensus safety - only include events ALL facilitators have
+      val allHashSets = facilities.values.map(_.eventHashes).toList
+      val commonHashes = allHashSets.reduceOption(_ intersect _).getOrElse(Set.empty[Hash])
```
ryle-ai: Unanimous intersection means a single slow or temporarily partitioned node will block all events from being included in consensus. In production with real network variability, this could significantly reduce throughput or cause consensus rounds with empty event sets.

I'm curious about a few things here -- what's the expected behavior when one validator is lagging? Is there a plan to move to majority-based inclusion (>50% threshold) as described in the design docs, and has this been observed in testing? The risk is that even brief network hiccups could stall the whole pipeline if every node must agree on the event set.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/http/routes/EventGossipRoutes.scala
```diff
@@ -0,0 +1,77 @@
...
+final case class EventGossipRoutes[F[_]: Async, Event: Encoder: Decoder, Key](
+  mempool: EventMempool[F, Event, Key],
+  triggerEventConsensus: Option[F[Unit]] = None
+) extends Http4sDsl[F]
+    with P2PRoutes[F] {
+
+  protected val prefixPath: InternalUrlPrefix = "/events"
+
+  protected val p2p: HttpRoutes[F] = HttpRoutes.of[F] {
+    case req @ POST -> Root / "push" =>
```
ryle-ai, line 40: Authentication is properly enforced via `PeerAuthMiddleware`, but there's no rate limiting on the `/events/push` endpoint. A compromised peer (or one with valid credentials but malicious behavior) could flood the mempool with events. The `maxSize` check in EventMempool helps prevent unbounded growth, but it doesn't prevent CPU and bandwidth exhaustion from processing a high volume of invalid events before they get rejected.

Same concern applies to `/events/iwant` -- a peer could repeatedly request large event sets to exhaust bandwidth. I think per-peer rate limiting would be worth adding here, similar to how other high-volume endpoints tend to handle this.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/event/EventGossipDaemon.scala
```diff
@@ -0,0 +1,362 @@
...
+/** Information about the current mesh state for monitoring.
+  */
+case class MeshInfo(
+  meshSize: Int,
+  meshPeers: Set[PeerId],
+  seenHashCount: Int
+)
+
+/** Configuration for the event gossip daemon.
+  */
```
ryle-ai, line 72: The daemon runs a heartbeat every 1 second (mesh maintenance, score decay) and a pull loop every 2 seconds (IHAVE/IWANT to non-mesh peers). For a network of 100 validators with `gossipFactor = 3`, that works out to roughly 30 heartbeat operations/second network-wide and about 150 pull operations/minute to non-mesh peers. That seems reasonable, but I think these defaults should be environment-configurable -- mainnet may want different tuning than testnet.

Are these values based on simulation/testing, or are they initial estimates? Worth monitoring either way once this hits a real network.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/mempool/EventMempool.scala
```diff
@@ -0,0 +1,259 @@
...
+  * consensus. The mempool is purely a hash-based storage layer.
+  *
+  * @tparam F
+  *   Effect type
+  * @tparam Event
+  *   The event type
+  * @tparam Key
+  *   The state key type for conflict detection
+  */
+trait EventMempool[F[_], Event, Key] {
```
ryle-ai, line 32: The design docs mention metrics were deferred, which makes sense for an initial implementation. For production though, I think structured event tracing (e.g., to ClickHouse) would help a lot with diagnosing issues -- things like mempool size over time, event add/clear rates, gossip push/pull latencies, and hash intersection sizes per consensus round. Without this kind of observability, debugging mempool growth or gossip delays in a live network is going to be pretty painful.

PR #1409, opened 2026-02-02
chore: sdc metagraph source nodes: cns, validVote, starVault

ffox77:


ryle-ai (approved):
Seedlist-only change adding 9 new source node keys for the SDC metagraph (cns, validVote, starVault). Also fixes the missing trailing newline -- good catch.

One minor note -- the file still ends without a trailing newline (`\ No newline at end of file`). Probably fine for the seedlist parser since it splits on newlines, but worth being aware of if tooling ever cares about POSIX text file compliance.

🤖 *ryle-ai automated review*

marcus-girolneto (approved):
LGTM

ryle-ai (commented):
Saw the new commit adding Valid Vote, CNS, and StarVault registrations -- same pattern as the existing entries, 3 keys each with matching seedlist additions. LGTM.

Very minor nit I think -- the indentation on the three new blocks drifts by +1 space each (Valid Vote at 11 spaces, CNS at 12, StarVault at 13) compared to the existing entries which use 10. Not a blocker at all but figured I'd mention it in case scalafmt or similar would catch it.

ryle-goehausen (approved):

PR #1419, opened 2026-02-03
fix: add error recovery to syncFullIfNeeded and fromProducer optimiza…

scasplte2:
…tion

- MptStore.syncFullIfNeeded: Reset ordinal on sync failure to allow retry
- GlobalSnapshotStateProof.fromProducer: Direct creation from synced producer
- GlobalSnapshotAcceptanceManager: Use fromProducer after sync (avoids cache lookup)

Builds on top of PR #1417 fixes.


scasplte2:
@ryle-ai is this a worthwhile update? I think I may be overthinking it here.

PR #1420, opened 2026-02-04
Fix/currency token lock replacement

scasplte2:
## Summary
Adds comprehensive E2E edge case tests for the token lock replacement + delegated stake interaction.

Changes

Core Feature

• Remove currency token lock replacement restriction - Currency token locks can now be replaced (previously only DAG locks supported replacement)
• Fix globalActiveTokenLocksByRef lookup - Include currency token locks from currency snapshots in the replacement lookup map (fixes "NothingToReplace" errors for currency locks)
• Allow same-currency replacement - DAG→DAG and currency→same-currency replacements now work correctly
Files Changed

• ContextualTokenLockValidator.scala - Remove ReplacementIsNotSupported check for currency locks
• TokenLockValidator.scala - Same removal
• TokenLockStateManager.scala - Update filter to allow same-currency replacement
• GlobalSnapshotAcceptanceManager.scala - Include currency locks in globalActiveTokenLocksByRef
Test Metagraph Updates

• Add UsageUpdateWithTokenLockReplacement data type for testing replacement flow
• Update Combiners.scala and LifecycleSharedFunctions.scala to handle new type
CI Test Coverage

New E2E test workflow (test-token-lock-replacement.yml) covering:

1. Replace with same amount → ReplacementLowerThanCurrentTokenLock ❌
2. Replace with less amount → ReplacementLowerThanCurrentTokenLock ❌
3. Replace non-existent ref → NothingToReplace ❌
4. Replace with minimum increase (+1 datum) → ✅
5. Multiple sequential replacements (3x) → ✅
6. Replace while stake in withdrawal → ✅
Testing

• All edge case tests pass on local cluster
• Verified delegated stake currentTokenLockRef and currentAmount update correctly after replacement
Notes

The e2e-test/ directory contains local debugging scripts used during development. These are not part of CI but may be useful for manual testing.

### DIFF
--- a/.github/action_scripts/delegated_staking/token-lock-replacement-edge-cases.js
+++ b/.github/action_scripts/delegated_staking/token-lock-replacement-edge-cases.js
@@ -1,0 +1,485 @@
+/**
+ * Token Lock Replacement Edge Cases - Extended tests for delegated stake + token lock replacement
+ *
+ * Tests edge cases not covered by the main delegated-staking.js tests:
+ * 1. Replace with same amount (should fail - ReplacementLowerThanCurrentTokenLock)
+ * 2. Replace with less amount (should fail - ReplacementLowerThanCurrentTokenLock)
+ * 3. Replace non-existent token lock ref (should fail - NothingToReplace)
+ * 4. Replace with minimum valid increase (+1 datum)
+ * 5. Multiple sequential replacements (3 in a row)
+ * 6. Replace while stake is in withdrawal (pendingWithdrawals)
+ *
+ * Usage:
+ * - CI: node token-lock-replacement-edge-cases.js 90 91 testTokenLockReplacementEdgeCases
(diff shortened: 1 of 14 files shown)


modules/shared/src/test/scala/io/constellationnetwork/json/BrotliCompareApp.scala
```diff
@@ -0,0 +1,120 @@
...
+import eu.timepit.refined.types.numeric.{NonNegLong, PosLong}
+import io.circe.Printer
+import io.circe.syntax._
+
+/**
+ * Test app to compare brotli serialization between Tessellation and dag4.js
+ * 
+ * Run with: sbt "shared/test:runMain io.constellationnetwork.json.BrotliCompareApp"
+ */
+object BrotliCompareApp extends IOApp.Simple {
```
scasplte2: just a test app that I thought might be useful to keep as a comparison but we can remove if y'all think it is clutter

scasplte2:
Currency token locks are disabled until an adequate signalling mechanism is derived to allow metagraph state to react to token lock replacements

PR #1422, opened 2026-02-05
feat: add MPT field-level digest tracking for debugging non-determinism

scasplte2:


ryle-ai (commented):
## Review by @ryle-ai

### Summary
This PR adds debugging utilities to track and compare per-field digests from the Merkle Patricia Trie (MPT), primarily for debugging non-determinism issues between consensus (incremental sync) and re-traversal (full sync). The implementation is well-structured and the core digest comparison logic is correct.

### Issue Found

**Test will fail for empty trie case**
`modules/shared/src/test/scala/io/constellationnetwork/security/mpt/MptFieldDigestsSuite.scala:45-50`

```scala
test("getSubtrieDigest returns Hash.empty for empty trie") {
  for {
    trie <- MerklePatriciaTrie.make[IO, Json](Map.empty)  // This throws RuntimeException!
    digest <- MptFieldDigests.getSubtrieDigest[IO](trie, Hex("0000000002"))
  } yield expect.eql(Hash.empty, digest)
}
```

`StatelessMerklePatriciaProducer.createWithData` throws `RuntimeException("Empty data provided")` when given an empty map. This test will fail at runtime.

**Recommendation:** Either remove this test if empty trie is not a valid use case, or create a valid trie with a single entry and test a prefix that doesn't exist.

### Observations (Non-blocking)

1. **Subtrie digest stability**: The test at line 131-159 correctly documents that adding entries under different prefixes can reorganize branch nodes and change digests - this is expected MPT behavior.

2. **Environment variable check**: Using `sys.env.getOrElse("MPT_FIELD_DIGESTS_LOGGING", "false")` is acceptable for debugging code.

3. **Cache eviction performance**: Sorting up to 51 entries for eviction is acceptable for the cache size of 50.

### Positive Aspects

1. **Thread Safety**: Proper use of `Ref` for concurrent state management in cache operations.
2. **Memory bounded**: Cache limited to 50 entries via `MaxCacheSize`.
3. **Correct digest comparison logic**: `compareAndLogDifferences` and `compareEntrySets` properly identify key and value differences.
4. **No production impact**: Diagnostic code is guarded by `whenA(stateProofInvalid)` and the env var.
5. **Good logging structure**: Well-organized debug output with clear separators.

### Verdict
The test issue should be addressed, but this is non-blocking for debugging purposes. The core functionality is sound.

🤖 *ryle-ai automated review*

ryle-ai (commented):
## Re-Review: MPT Field-Level Digest Tracking

**Previous review at:** commit `f76781dc`
**Current review at:** commit `a3793b3c` (latest)

### Previously Noted Issue — ✅ Resolved

**Empty trie test case** — the problematic test that would throw `RuntimeException("Empty data provided")` has been removed. Correct fix since empty tries are not a valid production use case.

### New Code Assessment (a3793b3c)

The new commit adds `MptIncrementalVsFullSyncSuite.scala` — a comprehensive test suite for verifying incremental vs full sync consistency. Test architecture is solid:
- Proper `MutableIOSuite` with shared resources
- Tests the core hypothesis that `StateChangesAccumulator` produces identical results to full rebuild
- Covers entry-level comparison (adds, removes, updates)

### Minor Observation (non-blocking)

**MptIncrementalVsFullSyncSuite.scala:42-44**
```scala
// Generate deterministic test addresses
private val addr1 = addressGen.sample.get
private val addr2 = addressGen.sample.get
```

Comment says "deterministic" but `addressGen.sample.get` uses unseeded ScalaCheck generators — produces random values each run. Not a bug (tests will pass regardless), but either remove the misleading comment or use seeded generators / hardcoded addresses for true reproducibility.

### No Critical Issues

Test-only changes. No production code modified. Debug logging is acceptable for the diagnostic purpose of this PR.

🤖 *ryle-ai automated review*

ryle-ai (commented):
Deeper code-level review after cross-referencing the full MPT implementation, `GlobalStateConverter`, and the producer hierarchy.

🤖 *ryle-ai automated review*

modules/shared/src/main/scala/io/constellationnetwork/security/mpt/MptFieldDigests.scala
```diff
@@ -0,0 +1,320 @@
...
+
+          case branch: MerklePatriciaNode.Branch =>
+            val nextNibble = remainingPrefix.head
+            branch.paths.get(nextNibble) match {
+              case Some(child) => traverse(child, remainingPrefix.tail)
+              case None        => Hash.empty // No data under this prefix
+            }
+        }
+
+    traverse(trie.rootNode, prefixNibbles).pure[F]
```
ryle-ai, line 66: I think this subtrie traversal logic has a subtle issue with how MPT extension nodes interact with field prefixes.

Consider: the key format is `[networkNamespace 2 hex][fieldId 8 hex][contractNamespace][userNamespace]`. For hypergraph data, the prefix is `00` + 8 hex chars (e.g., `0000000002` for Balances). But when the MPT has entries across many fields, the root is likely a Branch node, and the first nibble `0` (from `00...`) routes to a child. That child might be an Extension node whose `shared` path spans across the field ID boundary -- meaning the extension covers nibbles from both the namespace AND the field ID.

In that case, when `shared.startsWith(remainingPrefix)` is true, you return `ext.child.digest` -- but that child may contain entries from *multiple* fields if the extension only partially covers the field ID portion. The digest you'd get would be an aggregate of entries from multiple fields, not just the one you're querying.

For debugging purposes this is probably fine -- if two fields share a common prefix path through the trie, they'll produce the same digest and any divergence will still surface. But I think the method's doc comment ("returns the digest of the first node fully contained within the prefix") is slightly misleading about what it actually returns in the extension-spans-prefix case. Worth a clarifying comment.

modules/shared/src/main/scala/io/constellationnetwork/security/mpt/producer/FileSystemMerklePatriciaProducer.scala
```diff
@@ -101,25 +102,34 @@ class FileSystemMerklePatriciaProducer[F[_]: Async: Parallel: Hasher: JsonSerial
   override def buildForOrdinal(ordinal: SnapshotOrdinal): F[Either[MerklePatriciaError, MerklePatriciaTrie]] =
     build.flatMap {
       case Right(trie) =>
-        val rootHash = trie.rootHash
-        (cacheRootHash(ordinal, rootHash) >> lastBuiltOrdinalRef.set(Some(ordinal)))
-          .as(trie.asRight[MerklePatriciaError])
+        for {
```
ryle-ai, line 105: I see `extractAllFieldDigests` is now called on every `buildForOrdinal` invocation (which happens every consensus round). Looking at the implementation, it traverses the trie 19 times (once per `GlobalStateFieldId`), each time walking from the root following the prefix nibbles.

For the current field count (19) this is fine -- it's O(19 * prefix_depth) which is bounded and cheap. But I think it's worth noting that if the number of fields grows (e.g., with metagraph-specific field IDs), this could become non-trivial. The alternative would be to compute all field digests in a single traversal pass, but that's premature optimization for now.

The bigger question -- is this diagnostic data actually consumed anywhere in production, or is it purely for the env-var-gated debug logging? If it's only used when `MPT_FIELD_DIGESTS_LOGGING=true`, you might want to make the `extractAllFieldDigests` call conditional on that flag too, so you're not paying the cost on every consensus round in production.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotTraverse.scala
```diff
@@ -127,6 +128,55 @@ object GlobalSnapshotTraverse {
...
+                hexMap <- entries.toList.traverse {
+                  case (k, v) => GlobalStateKey.toHex[F](k).map(_ -> v)
+                }.map(_.toMap)
+                trie <- MerklePatriciaTrie.makeParallel[F, Json](hexMap)
+                digests <- MptFieldDigests.extractAllFieldDigests[F](trie)
+              } yield (trie, digests)
+            }
+
+            // Log retraversal field digests
+            _ <- logger.info(s"--- Retraversal Field Digests ---")
```
ryle-ai, line 157: This is a solid debugging approach -- when `stateProofInvalid`, rebuild a trie from `firstInfo.allStateEntries` and compare field-by-field against the consensus cache.

I think the key insight here is that `firstInfo.allStateEntries` goes through `GlobalStateConverter.toAllStateKeyValuePairs`, which serializes `DelegatedStakeRecord` (and everything else) via circe encoders. If the consensus path's `StateChangesAccumulator` has a `DelegatedStakeRecord` with `currentTokenLockRef = Some(x)` but the retraversal path has `currentTokenLockRef = None` for the same record, the JSON serialization will differ and the field digest for `ActiveDelegatedStakes` will show up as mismatched. This is exactly the kind of thing PR #1423 fixes.

One thing I notice -- `MerklePatriciaTrie.makeParallel[F, Json](hexMap)` at line 262 creates a fresh stateless trie from the hex map. This is a full rebuild from scratch, which is the correct thing to do for comparison -- but it means you're building a potentially large trie (all global state entries) every time validation fails. I'd add a guard or at least a log line indicating the entry count before building, since on mainnet this could be a significant amount of data.

Also -- `mptStore.underlying.getOrdinalCache(firstInc.ordinal)` relies on the consensus cache still being in memory. With `MaxCacheSize = 50`, if the retraversal is more than 50 ordinals behind, the cache entry will have been evicted. The code handles this gracefully (the `consensusCache.traverse_` just no-ops), but it's worth noting this tool is most useful for recent divergences, not historical debugging.

modules/shared/src/main/scala/io/constellationnetwork/security/mpt/producer/InMemoryMerklePatriciaProducer.scala
```diff
@@ -58,25 +59,34 @@ class InMemoryMerklePatriciaProducer[F[_]: Async: Hasher: Parallel: JsonSerializ
   override def buildForOrdinal(ordinal: SnapshotOrdinal): F[Either[MerklePatriciaError, MerklePatriciaTrie]] =
     build.flatMap {
       case Right(trie) =>
-        val rootHash = trie.rootHash
-        (cacheRootHash(ordinal, rootHash) >> lastBuiltOrdinalRef.set(Some(ordinal)))
-          .as(trie.asRight[MerklePatriciaError])
+        for {
```
ryle-ai, line 62: Same pattern as `FileSystemMerklePatriciaProducer` -- `extractAllFieldDigests` runs on every `buildForOrdinal`. Since `InMemoryMerklePatriciaProducer` is used by the `snapshot-streaming` project (I can see it wired up in `SnapshotProcessor.scala`), this means the streaming service will also pay the cost of field digest extraction on every ordinal it processes.

For the streaming service this is likely fine since it's not on the consensus hot path, but I think it would be cleaner to have the `MptOrdinalCache` construction be opt-in -- e.g., a parameter on the producer that controls whether field digests are computed and cached. For now this is acceptable since the extraction itself is cheap (19 prefix lookups).

scasplte2:
closed in lieu of https://github.com/Constellation-Labs/tessellation/pull/1425

PR #1423, opened 2026-02-05
Test/mpt incremental vs full sync

scasplte2:


ryle-ai (commented):
## Review: Test/MPT Incremental vs Full Sync

**Commit reviewed:** `afc2db47` (latest of 9)
**Reviewer context:** Full diff review + cross-reference with consensus path

### Critical Bug Fix (afc2db47) — ✅ Correct

The fix in `GlobalSnapshotContextFunctions.createContext` correctly applies the `DelegatedStakeRecord` transformation to the download/streaming path:

```scala
r.copy(
  currentTokenLockRef = r.currentTokenLockRef.orElse(r.tokenLockRef.some),
  currentAmount = r.currentAmount.orElse(r.amount.some)
)
```

This mirrors the consensus path in `GlobalSnapshotConsensusFunctions` and is applied consistently in **3 locations**:
1. `acceptDelegatedStakes` — existing delegated stakes
2. Before `setSumFixOrdinal` — rewards distribution
3. After `setSumFixOrdinal` — rewards distribution (else branch)

**Root cause confirmed:** Streaming/download nodes were computing different mptRoot during catch-up because the download path was missing these field transformations. Consensus path applied them, download path didn't → non-deterministic MPT root.

### Diagnostic Logging — ✅ Properly Guarded

All diagnostic logging is gated behind:
```scala
Option(System.getenv("MPT_STATE_DEBUG")).contains("true")
```

Logging locations cover consensus, retraversal, and download sync paths — exactly the comparison points needed for debugging.

### Test Suites — ✅ Comprehensive

The 4 test suites (MptIncrementalVsFullSync, MptStateMutationDebug, MptFileSystemProducerDebug, MptFieldDigests) provide excellent coverage:
- Incremental vs full sync comparison
- State mutation sequences (create/update/remove)
- FileSystemMerklePatriciaProducer persistence + reload
- Serialization round-trip tests

All test patterns are correct — no empty trie issues.

### 🔴 Issue: `wallet.jar` Binary Committed

Commit `a361168f` includes a ~91MB `wallet.jar` binary file. This should be removed before merge — it bloats the repository permanently. Add `*.jar` or `wallet.jar` to `.gitignore`.

### Docker Configuration — ✅ Good

The `gl0-stream` service profile for streaming node testing is properly configured with separate key/ports (9030-9032) and `MPT_STATE_DEBUG` enabled.

### Recommendation

**APPROVE with required change:** Remove `wallet.jar` from the commit history before merging. The bug fix itself is critical and correct — it should be prioritized for deployment.

This is the kind of subtle state consistency bug that causes production forks. Good debugging methodology here — the test suites that prove the conversion logic is correct in isolation, combined with the diagnostic logging, narrowed it down to the download path transformation gap.

ryle-ai (commented):
Re-reviewing new commits since afc2db47 (4 new commits: a9b8ef74..eb26c609).

**wallet.jar (~91MB binary) is still in this PR.** I flagged this in my previous review -- this needs to be removed before merge. `git rm wallet.jar` + force push, or at minimum add it to `.gitignore` and strip it from history. Binaries in the repo will bloat the clone size for everyone permanently.

On the substantive code changes:

**`GlobalSnapshot.scala` ordinal fix** -- using `genesis.ordinal.next` for the stateProof in `mkFirstIncrementalSnapshot` makes sense. The first incremental snapshot lives at `ordinal.next`, so the proof should be computed at that ordinal. The previous off-by-one was a real bug.

**`GlobalSnapshotTraverse.scala` RETRAVERSAL fix** -- applying the `currentTokenLockRef`/`currentAmount` fallback transformation in the rollback path mirrors what was already done for the download path in the earlier commit. This is the right fix -- without it, the MPT inputs during retraversal would diverge from what consensus produced. One thing I notice is that this transformation logic (`r.copy(currentTokenLockRef = r.currentTokenLockRef.orElse(...), currentAmount = r.currentAmount.orElse(...))`) now lives in at least three places -- consensus acceptance, download, and retraversal. I think extracting it to a shared method on `DelegatedStakeRecord` or `GlobalSnapshotInfo` would reduce the risk of these drifting apart in the future. Not blocking, but worth a follow-up.

**Expanded debug logging** -- properly guarded behind `MPT_STATE_DEBUG=true`, no concerns. The additional fields (balances, lastTxRefs, tokenLocks, allowSpends) will be useful for diagnosing future mismatches.

**Test script** -- `mpt_rollback_comprehensive_test.sh` is a good addition for E2E validation. References `docker/jars/wallet.jar` so it depends on the binary being available at that path -- just making sure that's intentional and not coupled to the committed `wallet.jar` at the repo root.

TL;DR: The code fixes look correct. Please remove `wallet.jar` from the PR.

🤖 *ryle-ai automated review*

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotTraverse.scala
```diff
@@ -107,11 +109,72 @@ object GlobalSnapshotTraverse {
           _ <- logger.info(s"Rollback hash candidate: ${hashCandidate.show}")
           firstInc <- loadIncOrErr(incHashesNec.head)
 
-          firstInfo <- loadFullOrIncOrErr(hashCandidate).flatMap {
+          firstInfoLoaded <- loadFullOrIncOrErr(hashCandidate).flatMap {
             case Left(globalIncrementalSnapshot) => loadInfoOrErr(globalIncrementalSnapshot.ordinal)
             case Right(globalSnapshot)           => globalSnapshot.info.toGlobalSnapshotInfo.pure[F]
           }
```
ryle-ai, line 116: So the transformation applied here on `firstInfoLoaded` looks correct for aligning the retraversal path with consensus -- but I think there's a subtle concern re: the `firstInfo` path vs the `incHashesNec.tail.foldLeftM` path below.

This transformation fires for `firstInfo` specifically (the info loaded from the candidate snapshot), but the `foldLeftM` at the bottom of `loadChain` calls `contextFns.createContext` for subsequent snapshots. `createContext` is `GlobalSnapshotContextFunctions.createContext`, which is separately patched in this PR to apply the same transformation. So the chain is covered -- but I wanted to confirm: is there any code path where an intermediate `info` gets persisted *before* the transformation runs? Looking at `setInitialFetchingGL0` and `setInitial` below, those use `firstInfo` which at that point already has the transformation applied, so that looks fine.

The main thing I'd highlight -- this transformation logic (`r.copy(currentTokenLockRef = ..., currentAmount = ...)`) now lives in at minimum 3 places: here, `GlobalSnapshotContextFunctions` (2 branches), and presumably the consensus acceptance path. I think it would be cleaner to extract this to a method on `DelegatedStakeRecord` or a utility on `GlobalSnapshotInfo` -- something like `info.withNormalizedDelegatedStakes(ordinal, threshold)` -- to avoid future drift. Not blocking, but this is the kind of thing where a 4th copy will eventually show up and someone will forget to update one of them.

modules/shared/src/main/scala/io/constellationnetwork/schema/GlobalSnapshot.scala
```diff
@@ -76,7 +76,7 @@ object GlobalSnapshot {
   def mkFirstIncrementalSnapshot[F[_]: Parallel: Async: Hasher: JsonSerializer](
     genesis: Hashed[GlobalSnapshot]
   )(implicit stateProofSelector: GlobalStateProofSelector): F[GlobalIncrementalSnapshot] =
-    genesis.info.toGlobalSnapshotInfo.stateProof[F](genesis.ordinal).map { stateProof =>
+    genesis.info.toGlobalSnapshotInfo.stateProof[F](genesis.ordinal.next).map { stateProof =>
```
ryle-ai, line 79: I think this is correct -- `mkFirstIncrementalSnapshot` creates the snapshot at `genesis.ordinal.next`, so computing the stateProof at `genesis.ordinal.next` matches. Previously using `genesis.ordinal` for the proof but `genesis.ordinal.next` for the snapshot ordinal meant the proof selector / hasher logic could pick a different hashing strategy between the two, which would cause a mismatch on any ordinal where the hash logic transitions.

One thing worth double checking -- this function is called during genesis initialization in tests and potentially during the initial network bootstrap. Are there any persisted genesis snapshots in testnet/mainnet whose first incremental snapshot was computed with the old ordinal? If so, nodes doing a full retraversal from genesis would now compute a different state proof for that first incremental and fail validation. I believe this should be fine since the `stateProof` is always recomputed during traversal via `contextFns.createContext`, but worth verifying against actual persisted data.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotContextFunctions.scala
```diff
@@ -248,10 +248,23 @@ object GlobalSnapshotContextFunctions {
...
+                    // Apply same transformation as consensus path for ordinals > incrementalDelegatedStakingStartingOrdinal
+                    val transformedCreateDelegatedStakes =
+                      if (signedArtifact.ordinal > incrementalDelegatedStakingStartingOrdinal) {
+                        updatedCreateDelegatedStakes.view.mapValues { records =>
+                          records.map { r =>
+                            r.copy(
+                              currentTokenLockRef = r.currentTokenLockRef.orElse(r.tokenLockRef.some),
+                              currentAmount = r.currentAmount.orElse(r.amount.some)
+                            )
+                          }
```
ryle-ai, line 260: I notice the transformation is duplicated in both the `< setSumFixOrdinal` branch and the `else` branch. The logic is identical in both -- the only difference between the branches is the `totalEmittedRewardsAmount` calculation (`.distinct.sum` vs `.sum`). I think you could lift the `transformedCreateDelegatedStakes` computation above the `if/else if/else` to avoid the duplication -- the ordinal guard `> incrementalDelegatedStakingStartingOrdinal` is independent of the `setSumFixOrdinal` check.

Not blocking, but reduces the surface area for future copy-paste inconsistency -- especially given this same pattern already appears in 3+ locations across the codebase.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/global/GlobalSnapshotAcceptanceManager.scala
```diff
@@ -1131,9 +1131,50 @@ object GlobalSnapshotAcceptanceManager {
...
+            // Diagnostic logging for MPT state comparison (enabled via MPT_STATE_DEBUG=true)
+            _ <- Sync[F].delay(Option(System.getenv("MPT_STATE_DEBUG")).contains("true")).flatMap { debugEnabled =>
+              if (debugEnabled) {
+                logger.info(s"[MPT_DEBUG][CONSENSUS] ordinal=$ordinal") >>
+                  // Balances summary
+                  logger.info(s"[MPT_DEBUG][CONSENSUS] balances entries: ${stateChangesAccumulator.balances.size}") >>
+                  // LastTxRefs summary
+                  logger.info(s"[MPT_DEBUG][CONSENSUS] lastTxRefs entries: ${stateChangesAccumulator.lastTxRefs.size}") >>
+                  // Token locks
+                  logger.info(s"[MPT_DEBUG][CONSENSUS] activeTokenLocks entries: ${stateChangesAccumulator.activeTokenLocks.size}") >>
```
ryle-ai, line 1143: I think the debug logging placement here is correct -- it's after the `stateChangesAccumulator` is fully built and before `syncFromStateChanges`. This captures the final state that the MPT sees.

One thing I'd note for future debugging: the `stateChangesAccumulator.activeDelegatedStakes` logged here is the *transformed* version (with `currentTokenLockRef`/`currentAmount` set), since the transformation happens in the acceptance manager's `accept` method prior to building the accumulator. But the download path's debug logging at `Download.scala` logs `context.activeDelegatedStakes` which is the *pre-transformation* version (from the snapshot info). So if you're comparing the two log outputs to find divergence, keep in mind they're logging at different points in the transformation pipeline -- you'd need to compare the consensus log output with the *post-transformation* download/retraversal state, not the raw snapshot info.

ryle-ai (commented):
Some deeper code-level observations from cross-referencing the full codebase -- focusing on areas the previous reviews didn't cover.

🤖 *ryle-ai automated review*

scasplte2:
included via #1426

PR #1424, opened 2026-02-05
feat: add --hypergraph-release flag to use published JARs

ottobot-ai:
Reopening PR #1413 (was accidentally closed).

Adds support for running a metagraph cluster against a stable tessellation release instead of building from source.

## Usage
```bash
just up --metagraph=/path/to/mymetagraph --hypergraph-release=v4.0.0-rc.2
```

**What it does:**
1. Downloads pre-built JARs (gl0, gl1, keytool, wallet) from the specified GitHub release
2. Skips sbt sdk/publishLocal — SDK resolves from Maven Central
3. Sets TESSELLATION_VERSION from the release tag so metagraph builds against correct SDK version
4. Metagraph still builds from source (targeting the specified SDK version)

**Benefits:**
- ~30x faster startup — 23 seconds vs ~12 minutes with full compile
- Stable targeting — Metagraph developers can target known-good tessellation releases

**Testing:**
Tested with template metagraph against v4.0.0-rc.2 — full cluster (gl0, gl1×3, ml0, cl1×3, dl1×3) running in 23 seconds.

**Security improvements (from previous review):**
- Download and verify .sha256 checksums from GitHub releases
- Validate release tag format before download
- Explicit version precedence documentation

### DIFF
--- a/docker/bin/download-release-jars.sh
+++ b/docker/bin/download-release-jars.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+#
+# Downloads tessellation JARs from a GitHub release
+# Usage: ./download-release-jars.sh <release-tag> <output-dir>
+# Example: ./download-release-jars.sh v3.5.11 ./docker/jars
+#
+
+set -e
+
+RELEASE_TAG="${1:-}"
+OUTPUT_DIR="${2:-./docker/jars}"
+
+if [ -z "$RELEASE_TAG" ]; then
+    echo "Error: Release tag required"
+    echo "Usage: $0 <release-tag> [output-dir]"
+    echo "Example: $0 v3.5.11 ./docker/jars"
+    exit 1
+fi
+
+REPO="Constellation-Labs/tessellation"
+BASE_URL="https://github.com/${REPO}/releases/download/${RELEASE_TAG}"
+
+# JAR mappings: release-name -> local-name
+declare -A JAR_MAP=(
+    ["cl-node.jar"]="gl0.jar"
+    ["cl-dag-l1.jar"]="gl1.jar"
+    ["cl-keytool.jar"]="keytool.jar"
+    ["cl-wallet.jar"]="wallet.jar"
+)
+
+mkdir -p "$OUTPUT_DIR"
+
+echo "Downloading tessellation JARs from release ${RELEASE_TAG}..."
+
+for release_name in "${!JAR_MAP[@]}"; do
+    local_name="${JAR_MAP[$release_name]}"
+    url="${BASE_URL}/${release_name}"
+    dest="${OUTPUT_DIR}/${local_name}"
+    
+    echo "  Downloading ${release_name} -> ${local_name}"
+    
+    if ! curl -fsSL -o "$dest" "$url"; then
+        echo "Error: Failed to download ${url}"
+        echo "Check that release ${RELEASE_TAG} exists and contains ${release_name}"
+        exit 1
+    fi
+    
+    # Verify download (basic size check)
+    size=$(stat -f%z "$dest" 2>/dev/null || stat -c%s "$dest" 2>/dev/null)
+    if [ "$size" -lt 1000 ]; then
+        echo "Error: Downloaded file ${local_name} is too small (${size} bytes)"
+        echo "The release may not contain this JAR or download failed"
+        rm -f "$dest"
+        exit 1
+    fi
+    
+    echo "    ✓ ${local_name} (${size} bytes)"
+done
+
+echo "Successfully downloaded all JARs to ${OUTPUT_DIR}"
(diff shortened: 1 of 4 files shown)


ryle-ai (approved):
Nice work on the second iteration. The checksum verification and tag format validation address the concerns from the previous round. Verified the JAR name mapping (`cl-node.jar` -> `gl0.jar`, `cl-dag-l1.jar` -> `gl1.jar`, etc.) matches the actual release assets on `v4.0.0-rc.2`.

A couple observations:

**Version precedence logic in `set-env.sh`** -- the `EXPLICIT_TESSELLATION_VERSION` pattern for preserving the env var across arg parsing is a bit unusual but I think it works. The ordering (explicit env > `--hypergraph-release` > `--version` > SNAPSHOT) makes sense for the use case. Just want to flag that this is a behavior change for existing users who set `TESSELLATION_VERSION` env var -- previously it was set early and could be overwritten by `--version`, now the env var always wins. I think that's actually the *correct* behavior but worth noting.

**`sdk/publishLocal` skip** -- the comment says "SDK should be available from Maven" which is true for published releases, but if someone uses this with a pre-release tag that hasn't been published to Maven yet, the metagraph `sbt` build will fail on dependency resolution. Might be worth a warning echo there rather than a silent skip -- something like "ensure SDK version X.Y.Z is available on Maven Central". Not a blocker, just a future footgun.

**`download-release-jars.sh`** -- the `stat -f%z` (macOS) with `stat -c%s` (Linux) fallback is the right cross-platform approach. Checksum verification looks solid.

🤖 *ryle-ai automated review*

docker/bin/download-release-jars.sh
```diff
@@ -0,0 +1,88 @@
...
+    # Download and verify checksum
+    echo "    Verifying checksum..."
+    if ! expected_checksum=$(curl -fsSL "$checksum_url" | awk '{print $1}'); then
+        echo "Error: Failed to download checksum from ${checksum_url}"
+        rm -f "$dest"
+        exit 1
+    fi
+    
+    actual_checksum=$(sha256sum "$dest" | awk '{print $1}')
+
```
ryle-ai, line 74: I think there's a portability issue here -- `sha256sum` isn't available on macOS by default (it ships `shasum -a 256` instead). The script already handles the macOS vs Linux difference for `stat` on line 85 with the fallback pattern, but this line doesn't do the same for checksums.

The existing `deploy-metagraph.sh` script in `ded/deploy/scripts/metagraph/` uses `shasum -a 256` as the canonical command (and `require_cmd shasum`). I think it'd be worth matching that convention for consistency, something like:

```bash
actual_checksum=$(shasum -a 256 "$dest" 2>/dev/null || sha256sum "$dest" 2>/dev/null | awk '{print $1}')
```

Or just use `shasum -a 256` since that works on both platforms with coreutils installed.

🤖 *ryle-ai automated review*

docker/bin/download-release-jars.sh
```diff
@@ -0,0 +1,88 @@
...
+    local_name="${JAR_MAP[$release_name]}"
+    url="${BASE_URL}/${release_name}"
+    checksum_url="${BASE_URL}/${release_name%.jar}.sha256"
+    dest="${OUTPUT_DIR}/${local_name}"
+    
+    echo "  Downloading ${release_name} -> ${local_name}"
+    
+    # Download JAR
+    if ! curl -fsSL -o "$dest" "$url"; then
+        echo "Error: Failed to download ${url}"
```
ryle-ai, line 52: So `curl -fsSL -o "$dest"` writes directly to the final destination -- if curl fails partway through (network timeout, partial download), you'll end up with a truncated file at the expected path. The checksum will catch it downstream, but then you've got partial state where some JARs are fine and others aren't.

I think the more robust pattern would be to download to a temp file first and atomic-rename on success:

```bash
dest_tmp="${dest}.download.$$"
if ! curl -fsSL -o "$dest_tmp" "$url"; then
    rm -f "$dest_tmp"
    ...
fi
# checksum verification on $dest_tmp, then:
mv "$dest_tmp" "$dest"
```

Not a blocker -- the checksum verification will catch most issues -- but this is the pattern I've seen in other download scripts that need to be resilient to interrupts.

🤖 *ryle-ai automated review*

docker/bin/set-env.sh
```diff
@@ -190,6 +192,25 @@ exit_func() {
 echo "BUILD_ONLY: $BUILD_ONLY"
 echo "RELEASE_TAG: $RELEASE_TAG"
 
+# Set TESSELLATION_VERSION with explicit precedence (after args are parsed):
+# 1. TESSELLATION_VERSION env var (explicit override) - highest priority
```
ryle-ai, line 196: The precedence logic here is clean. One thing I think is worth noting for context -- the euclid-development-environment's Dockerfile (in \`infra/metagraph-ubuntu/Dockerfile\`) already has its own JAR download logic for tag-based builds using the same release URL pattern and JAR naming convention. That path flows through \`euclid.json\` → \`get-information.sh\` → docker build-args, which is separate from this in-repo tooling.

So this PR is correctly scoped to the in-repo docker dev environment -- just want to flag the parallel path in euclid in case someone asks why they need \`--hypergraph-release\` when euclid already handles released versions.

🤖 *ryle-ai automated review*

scasplte2:
closed in lieu of https://github.com/Constellation-Labs/tessellation/pull/1434

PR #1425, opened 2026-02-06
fix: mpt stateproof fixes

scasplte2:
Incorporates changes and comments from
https://github.com/Constellation-Labs/tessellation/pull/1422


scasplte2:
fixed via #1426

PR #1426, opened 2026-02-06
fix: fixing stateProof broken issue

marcus-girolneto:


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotContextFunctions.scala
```diff
@@ -248,10 +269,23 @@ object GlobalSnapshotContextFunctions {
...
+                      if (signedArtifact.ordinal > incrementalDelegatedStakingStartingOrdinal) {
+                        updatedCreateDelegatedStakes.view.mapValues { records =>
+                          records.map { r =>
+                            r.copy(
+                              currentTokenLockRef = r.currentTokenLockRef.orElse(r.tokenLockRef.some),
+                              currentAmount = r.currentAmount.orElse(r.amount.some)
+                            )
+                          }
+                        }.to(SortedMap)
+                      } else updatedCreateDelegatedStakes
```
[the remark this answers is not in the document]
marcus-girolneto: 👍🏼

scasplte2 (approved):

ryle-ai (commented):
The three core MPT fixes here are solid and clearly address the root causes of the StateProof mismatch. The proof sync fix, the insertion ordering fix, and the `getOrEmpty` fix all make sense together -- each one independently could cause divergence between the incremental and full-build paths, so I think fixing all three at once is the right call.

A couple of things I want to flag:

**Duplicated transformation block in `GlobalSnapshotContextFunctions.scala`** -- The `transformedCreateDelegatedStakes` block is copy-pasted in both the `ordinal.value < setSumFixOrdinal.value` and the `else` branches of the historical snapshot validation path. I believe the logic is the same in both cases, and if it ever needs to change it would need to be changed in two places. I think this should be extracted to a helper method or a `val` computed once before the branch -- it's not blocking but it's the kind of thing that drifts silently.

**Unused import** -- `java.security.MessageDigest` is imported in `FileSystemMerklePatriciaProducer.scala` but never referenced. Minor cleanup.

**Re: fatal error removal in `GlobalSnapshotAlignment`** -- removing the hard crash for `NoPeersWithMajorityHash` / `NoMajoritySnapshotData` and letting the stream restart is a reasonable change. These conditions can be transient (especially during network partitions or bootstrap), and crashing the process was overly aggressive. I think this is fine for now -- worth revisiting if we see nodes silently spinning in alignment loops without making progress, but that's a separate concern.

Test coverage is thorough -- the suites do a good job of exercising the exact production scenario (incremental build, full rebuild, rollback, variable-length keys, insertion order sensitivity).

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotContextFunctions.scala
```diff
@@ -94,29 +98,44 @@ object GlobalSnapshotContextFunctions {
...
+          (unexpiredWithdrawals, expiredWithdrawals) = existingWithdrawals.foldLeft(
+            (
+              SortedMap.empty[Address, SortedSet[PendingDelegatedStakeWithdrawal]],
+              SortedMap.empty[Address, SortedSet[PendingDelegatedStakeWithdrawal]]
+            )
+          ) {
+            case ((unexpiredAcc, expiredAcc), (address, withdrawals)) =>
+              val (expired, unexpired) = withdrawals.partition {
+                case PendingDelegatedStakeWithdrawal(_, _, _, withdrawalEpoch, _, _) =>
+                  isWithdrawalExpired(withdrawalEpoch)
```
ryle-ai, line 125: This transformation block (the `if (signedArtifact.ordinal > incrementalDelegatedStakingStartingOrdinal)` section computing `transformedCreateDelegatedStakes`) appears identically in the `else` branch below around line 294. I think it should be lifted out and computed once before the `if/else if/else` chain to avoid the duplication -- both branches apply the same `currentTokenLockRef.orElse / currentAmount.orElse` normalization, and if the logic needs to change for any reason later, having it in one place would be safer.

modules/shared/src/main/scala/io/constellationnetwork/security/mpt/producer/FileSystemMerklePatriciaProducer.scala
```diff
@@ -1,5 +1,7 @@
 package io.constellationnetwork.security.mpt.producer
 
+import java.security.MessageDigest
```
ryle-ai, line 3: This import doesn't appear to be used anywhere in the file -- I think it might be left over from a debugging session. Can be cleaned up.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotContextFunctions.scala
```diff
@@ -81,7 +85,7 @@ object GlobalSnapshotContextFunctions {
```
ryle-ai, line 80: I think the token lock replacement logic here is correct -- matching replacement token locks by their `replaceTokenLockRef` against the existing records' `tokenLockRef`, and updating both `currentTokenLockRef` and `currentAmount` on the record. One thing I want to verify -- does `record.tokenLockRef` here return the *original* tokenLockRef (from the event), or the *current* one? Looking at the DelegatedStakeRecord definition, the `tokenLockRef` method uses `currentTokenLockRef.getOrElse(event.tokenLockRef)`, which means after the first replacement, `record.tokenLockRef` would return the *current* ref. So a second replacement would need to reference the *current* hash, not the original one. I believe that's the intended behavior since `replaceTokenLockRef` points to the lock being replaced, but worth confirming this works correctly for chained replacements.

PR #1427, opened 2026-02-06
test: add token lock replacement edge cases

scasplte2:
tests fail because secrets usage is disabled for public contributions so trying with a tessellation branch. References
https://github.com/Constellation-Labs/tessellation/pull/1421


ottobot-ai:
Closing - duplicate of #1421, which is being reopened from tessellation branch to enable CI tests.

ryle-ai (commented):
Hey @scasplte2 -- solid set of edge case tests here. Went through the test file alongside `ContextualTokenLockValidator.scala` and the existing `delegated-staking.js` patterns. Some thoughts:

re: the error matching approach in `createTokenLockExpectError` -- the string-based error matching (`errorStr.includes(expectedErrorSubstring)`) is somewhat fragile. The dag4 SDK wraps errors in different ways depending on the response format, and I see you're accounting for that with the `isValidationError` heuristic, which is fine. But I think there's a subtle issue: if the API ever returns a different error that happens to contain "NothingToReplace" or "TokenLock" as a substring of a longer message, the `isValidationError` check would pass incorrectly and you'd silently miss a real network error. Probably unlikely in practice but worth noting -- maybe tightening the check to look for the specific error key rather than substring matching on the whole message would be more robust. Not blocking though.

re: test 6 (testReplaceWhileInWithdrawal) -- I think there's a potential timing issue here. You create the stake, wait 5 seconds, then immediately withdraw. The existing `delegated-staking.js` tests use `withRetry` to wait for stake inclusion in the global snapshot before performing operations on it. Here you're relying on a fixed `sleep(5000)` which could be flaky in CI if snapshot production is slow. I'd suggest using `withRetry` to confirm the stake is in `activeDelegatedStakes` before calling `withdrawDelegatedStake`, similar to what `testReplaceSameAmount` does. Same concern with the withdrawal confirmation -- you verify it's in `pendingWithdrawals` (good), but then immediately try the replacement. Given that the replacement check happens on L1 state, there might be a propagation delay. Probably fine most of the time, but could be a source of flakiness.

re: missing edge case -- looking at the Scala validator, there's a `ReplacementIsNotSupported` error path that fires when `currencyId` is set. The PR covers `NothingToReplace` and `ReplacementLowerThanCurrentTokenLock` but not `ReplacementIsNotSupported`. I get that currency token locks might not be easy to set up in the E2E environment, but it might be worth a TODO comment at least so it's tracked.

re: the workflow file -- looks structurally correct and matches the pattern from `test-delegated-staking.yml`. One thing I noticed is the `npm install` step installs packages directly rather than using the `package-lock.json` that's already there (the cache key points to `.github/action_scripts/package-lock.json`). The existing `test-delegated-staking.yml` does this same thing so I think it's a pre-existing pattern, but it means the cache isn't really being used effectively since you're installing individually rather than via `npm ci`. Not something to fix in this PR necessarily, but worth keeping in mind.

re: key usage -- you're using `PRIVATE_KEYS.key3` while the existing delegated-staking tests use `key4`. That's smart for avoiding conflicts when both test suites run concurrently. Worth documenting somewhere (comment in the file or in the shared constants) which keys are "reserved" for which test suites, because if someone adds another test and grabs key3 there'll be weird cross-test contamination.

The sequential replacement test (test 5) is the strongest part of this PR I think -- chaining 3 replacements and verifying the lock ref updates correctly after each one exercises a real workflow that could surface state management bugs. The 10-second GL0 sync waits between replacements seem reasonable.

Overall this looks good to merge. The timing concern in test 6 is the main thing I'd want addressed -- the rest is minor.

.github/action_scripts/delegated_staking/token-lock-replacement-edge-cases.js
```diff
@@ -0,0 +1,532 @@
...
+ * 1. Replace with same amount (should fail - ReplacementLowerThanCurrentTokenLock)
+ * 2. Replace with less amount (should fail - ReplacementLowerThanCurrentTokenLock)
+ * 3. Replace non-existent token lock ref (should fail - NothingToReplace)
+ * 4. Replace with minimum valid increase (+1 datum)
+ * 5. Multiple sequential replacements (3 in a row)
+ * 6. Replace while stake is in withdrawal (pendingWithdrawals)
+ *
+ * Usage:
+ * - CI: node token-lock-replacement-edge-cases.js 90 91 testTokenLockReplacementEdgeCases
+ * - Local: RUN_ENV=local node token-lock-replacement-edge-cases.js 90 91 testTokenLockReplacementEdgeCases
```
effects-ai: For consistency:
```suggestion
 * - Local: RUN_ENV=local node .github/action_scripts/delegated_staking/token-lock-replacement-edge-cases.js 90 91 testTokenLockReplacementEdgeCases

```

ryle-ai (commented):
Follow-up after reviewing the 9 new commits since the initial review.

The timing concern in test 6 (testReplaceWhileInWithdrawal) has been addressed -- the fixed `sleep(5000)` before withdrawal was replaced with `withRetry` polling on `activeDelegatedStakes`, which is the right approach. There's still a `sleep(5000)` after the withdrawal for L1 sync before attempting the replacement, which is reasonable given that L1 state sync doesn't have a clean polling target here.

The error matching was also tightened up in the review feedback commit. Instead of the heuristic `isValidationError` string check, it now validates `error.response?.status === 400` first and pulls the error message from `error.response?.data` rather than pattern-matching on the whole `error.message`. This is a better approach -- it separates transport errors from validation errors at the HTTP level rather than guessing from string contents.

The TODO for `ReplacementIsNotSupported` was added, and the key reservation is documented in the file header. Both were minor items and this is fine.

The most interesting development here is that test 6 uncovered actual protocol behavior that contradicted the initial assumption. The test originally expected replacement to fail with `NothingToReplace` during withdrawal, but it turned out token locks remain in `activeTokenLocks` during the withdrawal period and replacement is allowed. The test was updated to reflect this, and the subsequent commit ("add L1 sync wait before replacement during withdrawal") added the necessary delay for L1 to sync GL0 state before attempting the replacement. Good that the test surfaced this -- the comments explaining the behavior are clear.

One thing to flag: there's a `fix.patch` file committed to the PR (visible in the diff). This looks like a leftover from development and should be removed before merging.

scasplte2:
included with https://github.com/Constellation-Labs/tessellation/pull/1431 to include the test improvements

PR #1431, opened 2026-02-09
ci: improve cluster robustness with wait-for-ready and retry

scasplte2:
Addresses frequent CI E2E test failures where nodes get stuck in 'Observing' state due to race conditions during cluster formation on GitHub runners.

Changes:
- Added -wait_for_ready=true flag to wait for node to reach 'Ready' state
- Added configurable -max_attempts and -interval parameters
- Increased cluster check timeout from 120s to 240s
- Added retry logic (5 attempts with 5s backoff) for cluster/join requests
- Better error handling for unreachable nodes

Fixes argument parsing for double-dash args (--url -> url instead of -url).

### DIFF
--- a/.github/action_scripts/check_clusters/shared.js
+++ b/.github/action_scripts/check_clusters/shared.js
@@ -1,12 +1,14 @@
 const axios = require('axios');
 
-const fetchData = async (url) => {
-  for (let idx = 0; idx < 12; idx++) {
+const fetchData = async (url, maxRetries = 12, retryIntervalMs = 5000) => {
+  let lastError = null;
+  for (let idx = 0; idx < maxRetries; idx++) {
     try {
       const response = await axios.get(url, {
         headers: {
           Accept: 'application/json'
-        }
+        },
+        timeout: 10000
       });
 
       const { data } = response
@@ -14,9 +16,14 @@ const fetchData = async (url) => {
 
       return data
     } catch (e) {
-      throw Error(`Error when fetching data: ${e.message}`);
+      lastError = e;
+      console.log(`fetchData attempt ${idx + 1}/${maxRetries} failed for ${url}: ${e.message}`);
+      if (idx < maxRetries - 1) {
+        await sleep(retryIntervalMs);
+      }
     }
   }
+  throw Error(`Error fetching data from ${url} after ${maxRetries} attempts: ${lastError?.message}`);
 };
 
 const sleep = (ms) => {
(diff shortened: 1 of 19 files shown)


scasplte2:
@greptileai provide an updated review please

ryle-ai (commented):
The core problem this addresses is real and well-diagnosed -- the old `fetchData` in `shared.js` had a retry loop that threw on the first error, so it never actually retried. And the old `check_if_node_started.js` did `.split('=')[1]` which would truncate URLs with query params. Both good catches.

The `fetchData` fix is clean. Adding the axios timeout, accumulating `lastError`, logging each attempt, and only throwing after exhaustion -- all straightforward. The `checkIfNodeIsReady` bump from 12 to 24 attempts (120s to 240s) with a try/catch around the `fetchData` call makes sense too, since a node that's still starting shouldn't be treated as a permanent failure.

The `parseArgs` refactor in `check_if_node_started.js` using `indexOf`/`slice` instead of `split('=')` is the right fix for preserving URLs. The new `wait_for_ready` mode is a good addition -- the old code considered any 200 response as "started" regardless of node state, which was the root cause of the Observing-stuck race condition. Now genesis and initial validator nodes wait for Ready before downstream validators attempt to join.

The join retry blocks across all the validator actions (dag-l0, dag-l1, currency-l1, data-l1, metagraph-l0) are identical in structure. This is a lot of duplicated bash -- the same ~25 lines of curl-retry logic copy-pasted into 5 files. I'd consider extracting it into a shared script (maybe `.github/action_scripts/join_cluster.sh` that takes the node ID, CLI port, P2P port, and cluster name as args). That said, I understand this is CI glue code where copy-paste is sometimes the pragmatic choice, so I wouldn't block on it.

One thing worth noting: there's a nested retry situation in the `checkIfNodeIsReady` path. `checkIfNodeIsReady` calls `fetchData` up to 24 times, and `fetchData` itself retries up to 12 times with 5s sleeps. That means a single `checkIfNodeIsReady` call could theoretically take 24 * (12 * 5s) = 1440 seconds in the worst case before failing. In practice this probably doesn't matter because if the node is truly down, `fetchData` will fail fast on connection refused rather than timing out, but it's worth being aware of. You might want to pass a lower `maxRetries` to `fetchData` when calling it from `checkIfNodeIsReady` (like 1 or 2) since the outer loop is already handling the retry-over-time logic.

Similarly, the `check_if_node_started.js` script (used before join) doesn't set `maxRetries` or pass a retry config for its axios calls. Its default of 30 attempts * 10s = 300s is reasonable on its own, but it has the same nested-retry consideration with the 5s timeout it sets on axios.

The temp file cleanup in the join retry blocks is handled correctly on the success path (`rm -f` after break) and after logging on the failure path. Good.

One minor thing: several of the modified action.yml files end without a trailing newline (I count ~18 "No newline at end of file" warnings in the diff). Not a functional issue, but worth fixing for cleanliness.

Overall this is a solid CI reliability improvement. The changes address real failure modes (fetchData not actually retrying, join requests failing transiently, nodes not being in Ready state before cluster join) with reasonable timeout/retry parameters.

marcus-girolneto (approved):

PR #1434, opened 2026-02-10
feat: add --hypergraph-release flag to use published JARs

scasplte2:
replaces https://github.com/Constellation-Labs/tessellation/pull/1424 with a local tessellation branch

### DIFF
--- a/docker/bin/assembly.sh
+++ b/docker/bin/assembly.sh
@@ -6,67 +6,92 @@ assemble_all() {
 
 show_time "Starting assembly"
 
-if [[ "$INCLUDE_L0" == "false" && "$INCLUDE_L1" == "false" && "$SKIP_ASSEMBLY" == "false" ]]; then
-  assemble_all
+# If HYPERGRAPH_RELEASE is set, download pre-built JARs instead of building from source
+if [ -n "$HYPERGRAPH_RELEASE" ]; then
+  echo "Using pre-built hypergraph JARs from release: ${HYPERGRAPH_RELEASE}"
+  rm -rf ./docker/jars/ > /dev/null 2>&1 || true
+  mkdir -p ./docker/jars/
+  
+  ./docker/bin/download-release-jars.sh "$HYPERGRAPH_RELEASE" ./docker/jars/
+  
+  show_time "Downloaded release JARs"
+  
+  # Skip the rest of hypergraph assembly, jump to metagraph handling
+  SKIP_HYPERGRAPH_BUILD=true
 else
-  missing=false
-
-  for module in dag-l0 dag-l1 keytool wallet; do
-    set +e
-    jar_path=$(ls -1t modules/"$module"/target/scala-2.13/tessellation-"$module"-assembly*.jar 2>/dev/null | head -n1)
-    set -e
-    if [ -z "$jar_path" ]; then
-      echo "⚠️  Missing JAR for module: $module"
-      missing=true
-      break
-    fi
-  done
-
-  if [ "$missing" = true ]; then
-    echo "▶️  One or more modules is missing. Cannot skip assembly. Running full assembly"
-    assemble_all
-  else
-    if [ "$SKIP_ASSEMBLY" == "false" ]; then
-      override_set=false
-      if [ "$INCLUDE_L0" == "true" ]; then
-        echo "Assembling L0"
-        sbt dagL0/assembly
-        override_set=true
-      fi
-      if [ "$INCLUDE_L1" == "true" ]; then
-        echo "Assembling L1"
-        sbt dagL1/assembly
-        override_set=true
-      fi
-      if [ "$override_set" == "false" ]; then
-        echo "Assembling L0 according to default behavior"
-        sbt dagL0/assembly
-      fi
-    else
-      echo "Found existing assemblies, and skip assembly was set to true"
-    fi
-  fi
+  SKIP_HYPERGRAPH_BUILD=false
 fi
 
-show_time "SBT Assembly"
+if [ "$SKIP_HYPERGRAPH_BUILD" != "true" ]; then
+  # Build hypergraph JARs from source
+  if [[ "$INCLUDE_L0" == "false" && "$INCLUDE_L1" == "false" && "$SKIP_ASSEMBLY" == "false" ]]; then
+    assemble_all
+  else
+    missing=false
 
-rm -rf ./docker/jars/ > /dev/null 2>&1 || true;
-mkdir -p ./docker/jars/
+    for module in dag-l0 dag-l1 keytool wallet; do
+      set +e
+      jar_path=$(ls -1t modules/"$module"/target/scala-2.13/tessellation-"$module"-assembly*.jar 2>/dev/null | head -n1)
+      set -e
+      if [ -z "$jar_path" ]; then
+        echo "⚠️  Missing JAR for module: $module"
+        missing=true
+        break
+      fi
+    done
 
-for module in "dag-l0" "dag-l1" "keytool" "wallet"
-do
-  path=$(ls -1t modules/${module}/target/scala-2.13/tessellation-${module}-assembly*.jar | head -n1)
-  dest="$PROJECT_ROOT/docker/jars/${module}.jar"
-  cp $path $dest
-done
+    if [ "$missing" = true ]; then
+      echo "▶️  One or more modules is missing. Cannot skip assembly. Running full assembly"
+      assemble_all
+    else
+      if [ "$SKIP_ASSEMBLY" == "false" ]; then
+        override_set=false
+        if [ "$INCLUDE_L0" == "true" ]; then
+          echo "Assembling L0"
+          sbt dagL0/assembly
+          override_set=true
+        fi
+        if [ "$INCLUDE_L1" == "true" ]; then
+          echo "Assembling L1"
+          sbt dagL1/assembly
+          override_set=true
+        fi
+        if [ "$override_set" == "false" ]; then
+          echo "Assembling L0 according to default behavior"
+          sbt dagL0/assembly
+        fi
+      else
+        echo "Found existing assemblies, and skip assembly was set to true"
+      fi
+    fi
+  fi
 
-mv ./docker/jars/dag-l0.jar ./docker/jars/gl0.jar
-mv ./docker/jars/dag-l1.jar ./docker/jars/gl1.jar
+  show_time "SBT Assembly"
+
+  rm -rf ./docker/jars/ > /dev/null 2>&1 || true;
+  mkdir -p ./docker/jars/
+
+  for module in "dag-l0" "dag-l1" "keytool" "wallet"
+  do
+    path=$(ls -1t modules/${module}/target/scala-2.13/tessellation-${module}-assembly*.jar | head -n1)
+    dest="$PROJECT_ROOT/docker/jars/${module}.jar"
+    cp "$path" "$dest"
+  done
+
+  mv ./docker/jars/dag-l0.jar ./docker/jars/gl0.jar
+  mv ./docker/jars/dag-l1.jar ./docker/jars/gl1.jar
+fi
 
 
 if [ "$PUBLISH" == "true" ]; then
-  echo "Publishing local"
-  sbt --error sdk/publishLocal
+  if [ -n "$HYPERGRAPH_RELEASE" ]; then
+    echo "Skipping sdk/publishLocal - using release ${HYPERGRAPH_RELEASE}"
+    echo "  Note: Metagraph builds require tessellation-sdk ${HYPERGRAPH_RELEASE#v} from Maven Central."
+    echo "  SNAPSHOT versions are not published - use a tagged release."
+  else
+    echo "Publishing local"
+    sbt --error sdk/publishLocal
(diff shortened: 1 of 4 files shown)


ryle-ai (commented):
## Review: `--hypergraph-release` flag for pre-built JAR support

Overall this is a well-structured change. The TESSELLATION_VERSION precedence fix in `set-env.sh` is a genuine improvement that also fixes a latent bug where `--version=` args were parsed *after* `TESSELLATION_VERSION` was already set. A few observations:

### Metagraph builds may fail silently when SDK isn't on Maven Central

**`docker/bin/assembly.sh` lines 131-134 (in the PR diff)**

```bash
if [ -n "$HYPERGRAPH_RELEASE" ]; then
    echo "Skipping sdk/publishLocal - using release $HYPERGRAPH_RELEASE (SDK should be available from Maven)"
```

When `--hypergraph-release` is combined with `--metagraph` and `--publish`, the script skips `sdk/publishLocal`. The metagraph build (via `Dependencies.scala`) resolves `tessellation-sdk` at `TESSELLATION_VERSION` from Maven Central / `Resolver.mavenLocal`.

This works for versions actually published to Maven Central, but the tag regex accepts pre-release tags like `v3.5.11-rc1`. If someone passes a tag that matches the regex but isn't published to a remote resolver, the metagraph `sbt` assembly will fail with an opaque dependency resolution error -- no clear indication that the problem is a missing `publishLocal`.

Suggestion: either add a guard that warns/errors when `HYPERGRAPH_RELEASE` + `PUBLISH` + `METAGRAPH` are all set, or improve the echo message to explicitly say something like:

```bash
echo "WARNING: Skipping sdk/publishLocal. Metagraph builds require tessellation-sdk ${HYPERGRAPH_RELEASE#v} to be available from Maven Central."
```

This way the developer knows immediately where to look if their metagraph build fails.

### Minor: uninitialized `RELEASE_TAG` reference in new precedence block

**`docker/bin/set-env.sh` around the new precedence block (line ~204 in the PR diff)**

```bash
elif [ -n "$RELEASE_TAG" ]; then
```

`RELEASE_TAG` has no default initialization (unlike `HYPERGRAPH_RELEASE` which gets `${HYPERGRAPH_RELEASE:-""}`). This works today because `set -u` / `nounset` is not enabled, but for consistency and future-proofing, consider `${RELEASE_TAG:-}` here.

### Notes on the Greptile bot comments

For the author's reference: the bot's `stat` portability complaint on `download-release-jars.sh:85` is incorrect -- the `||` fallback pattern (`stat -f%z ... 2>/dev/null || stat -c%s ... 2>/dev/null`) handles cross-platform correctly. The associative array iteration order comment is also a non-issue since the downloads are independent and idempotent. The `cp` quoting issue was already fixed in this PR.

ryle-ai (commented):
### Phase 2 re-check doesn't handle `--version` / `RELEASE_TAG`

The Phase 2 re-check after arg parsing (line 205) only handles `--hypergraph-release`. But `--version` has the same timing problem: `RELEASE_TAG` is checked on line 57 (before arg parsing) when it hasn't been set yet, and there's no Phase 2 re-check for it.

This means `--version=v3.5.0` alone never propagates into `TESSELLATION_VERSION` -- it stays at `99.99.99-SNAPSHOT`. This is technically a pre-existing bug on `develop`, but the Phase 2 pattern makes the inconsistency more visible: you fixed the timing problem for `--hypergraph-release` but not for `--version`.

Consider extending the Phase 2 block:
```bash
if [ -n "$HYPERGRAPH_RELEASE" ] && [ "$TESSELLATION_VERSION" = "99.99.99-SNAPSHOT" ]; then
    export TESSELLATION_VERSION="${HYPERGRAPH_RELEASE#v}"
elif [ -n "$RELEASE_TAG" ] && [ "$TESSELLATION_VERSION" = "99.99.99-SNAPSHOT" ]; then
    export TESSELLATION_VERSION="${RELEASE_TAG#v}"
fi
```

### Missing `pipefail` in download-release-jars.sh masks curl failures

See inline comment on line 59.

docker/bin/download-release-jars.sh
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env bash
+#
+# Downloads tessellation JARs from a GitHub release with integrity verification
+# Usage: ./download-release-jars.sh <release-tag> <output-dir>
+# Example: ./download-release-jars.sh v3.5.11 ./docker/jars
+#
+
+set -e
```
ryle-ai: Consider `set -eo pipefail` here. Without `pipefail`, the checksum download pipeline on line 59 (`curl ... | awk ...`) reports success even when curl fails -- the pipeline exit code is awk's (0), not curl's. The downstream checksum comparison still catches the problem (empty expected vs real actual), but the error message is misleading ("Checksum verification failed" rather than "Failed to download checksum").

docker/bin/set-env.sh
```diff
@@ -190,6 +192,25 @@ exit_func() {
...
+# 1. TESSELLATION_VERSION env var (explicit override) - highest priority
+# 2. --hypergraph-release flag
+# 3. --version / RELEASE_TAG
+# 4. 99.99.99-SNAPSHOT (default, build from source)
+if [ -n "${EXPLICIT_TESSELLATION_VERSION:-}" ]; then
+    export TESSELLATION_VERSION="$EXPLICIT_TESSELLATION_VERSION"
+    echo "Setting TESSELLATION_VERSION=$TESSELLATION_VERSION (from environment)"
+elif [ -n "$HYPERGRAPH_RELEASE" ]; then
+    export TESSELLATION_VERSION="${HYPERGRAPH_RELEASE#v}"
+    echo "Setting TESSELLATION_VERSION=$TESSELLATION_VERSION (from --hypergraph-release)"
```
ryle-ai, line 208: This re-check handles the timing problem where `--hypergraph-release` is parsed after `TESSELLATION_VERSION` is already set. But `--version` / `RELEASE_TAG` has the exact same timing problem (line 57 checks `RELEASE_TAG` before arg parsing sets it on line 134), and there's no equivalent re-check here for it. If someone runs `./compose-runner.sh --version=v3.5.0`, `TESSELLATION_VERSION` stays `99.99.99-SNAPSHOT`.

Worth adding an `elif` branch for `RELEASE_TAG` to fix both timing issues consistently.

ryle-ai (commented):
### Missing `pipefail` masks curl failures in checksum download

`download-release-jars.sh` uses `set -e` but not `pipefail`. On line 68, `curl -fsSL "$checksum_url" | awk '{print $1}'` reports success even when curl fails, because the pipeline exit code is awk's (0). The `if !` guard on the pipeline won't catch HTTP errors from curl.

The downstream checksum comparison (line 76) still catches the problem indirectly -- empty `expected_checksum` won't match the actual hash -- but the error message says "Checksum verification failed" instead of "Failed to download checksum", which sends the developer looking in the wrong direction.

Adding `set -eo pipefail` at the top would fix this.

### `sdk/publishLocal` skip + metagraph build = opaque failure

When `--hypergraph-release` is combined with `--metagraph`, the script skips `sdk/publishLocal` (assembly.sh line 134). But `PUBLISH` is auto-set to `true` when `METAGRAPH` is present (set-env.sh line 271-275), and the metagraph's SBT build resolves `tessellation-sdk` at `TESSELLATION_VERSION` from Maven resolvers.

For versions published to Maven Central (via the `sonatype-publish` job in release.yml), this works. But the tag regex accepts pre-release tags like `v3.5.11-rc1`, and those may not be on Maven Central. A developer passing a pre-release tag with `--metagraph` would get an opaque SBT dependency resolution failure with no indication that `publishLocal` was skipped.

The echo on line 134 could be more explicit: `"WARNING: Skipping sdk/publishLocal. Metagraph builds require tessellation-sdk ${HYPERGRAPH_RELEASE#v} to be available from Maven Central."`

docker/bin/download-release-jars.sh
ryle-ai: `set -eo pipefail` here. Without `pipefail`, the checksum download pipeline on line 68 (`curl ... | awk ...`) reports success even when curl fails -- the pipeline exit code is awk's (0), not curl's. The downstream comparison still catches mismatches, but the error message is misleading ("Checksum verification failed" instead of "Failed to download checksum").

docker/bin/assembly.sh
```diff
@@ -6,67 +6,90 @@ assemble_all() {
...
-    set +e
-    jar_path=$(ls -1t modules/"$module"/target/scala-2.13/tessellation-"$module"-assembly*.jar 2>/dev/null | head -n1)
-    set -e
-    if [ -z "$jar_path" ]; then
-      echo "⚠️  Missing JAR for module: $module"
-      missing=true
-      break
-    fi
-  done
+  SKIP_HYPERGRAPH_BUILD=false
```
ryle-ai, line 22: When `--hypergraph-release` is set, this path downloads JARs into `./docker/jars/` and then sets `SKIP_HYPERGRAPH_BUILD=true`, which skips the source-build block entirely. But it also means we skip the `mv ./docker/jars/dag-l0.jar ./docker/jars/gl0.jar` renames (line 81-82). That's correct because the download script already names the files `gl0.jar`, `gl1.jar` etc. -- just noting this is a subtle coupling between the `JAR_MAP` in `download-release-jars.sh` and the rename logic here. If either side changes naming independently, jars will be wrong or missing with no error.

docker/bin/assembly.sh
```diff
@@ -6,67 +6,90 @@ assemble_all() {
...
-do
-  path=$(ls -1t modules/${module}/target/scala-2.13/tessellation-${module}-assembly*.jar | head -n1)
-  dest="$PROJECT_ROOT/docker/jars/${module}.jar"
-  cp $path $dest
-done
+  for module in "dag-l0" "dag-l1" "keytool" "wallet"
+  do
+    path=$(ls -1t modules/${module}/target/scala-2.13/tessellation-${module}-assembly*.jar | head -n1)
+    dest="$PROJECT_ROOT/docker/jars/${module}.jar"
+    cp "$path" "$dest"
```
ryle-ai, line 78: The echo says SDK "should be available from Maven" but this isn't guaranteed for pre-release tags (the tag regex accepts `v3.5.11-rc1` etc.). If someone passes a pre-release `--hypergraph-release` tag with `--metagraph`, SBT will fail with an opaque dependency resolution error. Consider making the warning more explicit:

```bash
echo "WARNING: Skipping sdk/publishLocal. Metagraph builds require tessellation-sdk ${HYPERGRAPH_RELEASE#v} to be available from Maven Central."
```

ottobot-ai:
## Addressed ryle-ai Review Feedback

Commit `5b7c2463` on `ottobot-ai/tessellation:feat/hypergraph-release-v2`

| Review Comment | Fix |
|----------------|-----|
| **download-release-jars.sh:8** — Missing `pipefail` masks curl failures | Changed `set -e` to `set -eo pipefail` so curl errors in `curl ... \| awk` pipeline propagate correctly |
| **assembly.sh:78** — Warning should be explicit about Maven Central | Changed to: `WARNING: Skipping sdk/publishLocal. Metagraph builds require tessellation-sdk ${HYPERGRAPH_RELEASE#v} to be available from Maven Central.` |
| **set-env.sh** — RELEASE_TAG has no default initialization | Added `export RELEASE_TAG=${RELEASE_TAG:-""}` for consistency with HYPERGRAPH_RELEASE |

### Note on Phase 2 re-check comment

The review mentioned that `--version`/`RELEASE_TAG` has the same timing problem as `--hypergraph-release`. Looking at the current code, this is **already addressed** — the elif chain at line 206-208 handles `RELEASE_TAG`:

```bash
elif [ -n "$RELEASE_TAG" ]; then
    export TESSELLATION_VERSION="${RELEASE_TAG#v}"
    echo "Setting TESSELLATION_VERSION=$TESSELLATION_VERSION (from --version/RELEASE_TAG)"
```

This runs after arg parsing, so `--version=v3.5.0` correctly propagates into `TESSELLATION_VERSION`.

docker/bin/download-release-jars.sh
scasplte2: - Changed 'set -e' to 'set -eo pipefail'
   - Ensures curl errors in 'curl ... | awk' propagate correctly

ottobot-ai:
**Updated** — Fixed the warning message. Now reads:

```
WARNING: Skipping sdk/publishLocal. If building a metagraph, tessellation-sdk X.Y.Z must be published (Maven Central or local).
```

Clarifies:
- Node JARs (gl0, gl1, etc.) → GitHub releases
- tessellation-sdk (metagraph dependency) → Maven Central or publishLocal

Commit: `cd88c906`

docker/bin/assembly.sh
scasplte2: - Changed warning to explicitly state Maven Central requirement
   - Helps developers diagnose opaque SBT dependency resolution failures

docker/bin/assembly.sh
scasplte2: - Added 'export RELEASE_TAG=${RELEASE_TAG:-""}' at top
   - Consistent with HYPERGRAPH_RELEASE initialization pattern

ryle-goehausen (approved):

PR #1436, opened 2026-02-10
fix: fixing consensus facilitators select

marcus-girolneto:


ottobot-ai (approved):
Looks good! The fix correctly addresses the bug where facilitators removed during unlock consensus weren't being excluded from the next round's selection.

**Analysis:**
- The `removedFacilitators` from `lastOutcome` contains peers that were voted out during the previous round's unlock consensus
- These should be excluded from the next round (which this PR does)
- Re-entry is still possible in subsequent rounds since each new `ConsensusState` starts with `removedFacilitators = empty`
- The removed `ClusterStorage` import was dead code — correctly cleaned up

**Optional suggestions (non-blocking):**

1. **Log peer IDs, not just count** — helps debugging in production:
```scala
s", excludedFromPreviousRound=${previouslyRemoved.map(_.show).mkString(",")}"
```

2. **Consider INFO level for exclusions** — operationally significant:
```scala
_ <- previouslyRemoved.traverse_(peerId =>
  logger.info(s"Excluding facilitator ${peerId.show} (removed in previous round) for key=$key")
)
```

Otherwise, LGTM! ✅

scasplte2 (approved):
lgtm

PR #1439, opened 2026-02-11
fix: comprehensive file I/O error handling

scasplte2:
PROT-1461 - follow up on file cleaning issue stalling network progress

### DIFF
--- a/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/infrastructure/snapshot/CurrencySnapshotCleanupStorage.scala
+++ b/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/infrastructure/snapshot/CurrencySnapshotCleanupStorage.scala
@@ -1,5 +1,7 @@
 package io.constellationnetwork.currency.l0.infrastructure.snapshot
 
+import java.nio.file.NoSuchFileException
+
 import cats.effect.Async
 import cats.syntax.all._
 
@@ -47,9 +49,13 @@ object CurrencySnapshotCleanupStorage {
 
         val cleanupAboveOrdinal = persistedStorage
           .cleanupAboveOrdinal(ordinal, deletePersisted)
-          .handleErrorWith { err =>
-            logger.error(err)("Error during cleanup snapshot of the metagraph") >>
-              CurrencyCleanupError.raiseError[F, Unit]
+          .handleErrorWith {
+            case _: java.nio.file.NoSuchFileException =>
+              // Files already cleaned up - not an error
+              Async[F].unit
+            case err =>
+              logger.error(s"Error during cleanup snapshot of the metagraph: ${err.getMessage}") >>
+                CurrencyCleanupError.raiseError[F, Unit]
           }
 
         val verify =
--- a/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/infrastructure/snapshot/CurrencySnapshotCleanupStorage.scala
+++ b/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/infrastructure/snapshot/CurrencySnapshotCleanupStorage.scala
@@ -54,7 +54,7 @@ object CurrencySnapshotCleanupStorage {
               // Files already cleaned up - not an error
               Async[F].unit
             case err =>
-              logger.error(s"Error during cleanup snapshot of the metagraph: ${err.getMessage}") >>
+              logger.error(err)("Error during cleanup snapshot of the metagraph") >>
                 CurrencyCleanupError.raiseError[F, Unit]
           }
 
--- a/modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/SnapshotDownloadStorage.scala
+++ b/modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/SnapshotDownloadStorage.scala
@@ -154,12 +154,12 @@ object SnapshotDownloadStorage {
             .handleErrorWith {
               case _: java.nio.file.NoSuchFileException =>
                 // Files already deleted - not an error
-                Async[F].unit
+                logger.debug(s"Snapshot_info files above ${ordinal.show} already deleted or do not exist")
               case err =>
-                logger.error(s"Error while deleting snapshot_info files above ${ordinal.show}: ${err.getMessage}") >>
+                logger.error(err)(s"Error while deleting snapshot_info files above ${ordinal.show}") >>
                   Async[F].raiseError(err)
             }
-          _ <- logger.info(s"Successfully deleted snapshot_info files above ordinal ${ordinal.show}")
+          _ <- logger.debug(s"Completed snapshot_info cleanup above ordinal ${ordinal.show}")
         } yield ()
 
         val cleanupAboveOrdinal = persistedStorage.cleanupAboveOrdinal(ordinal, movePersistedToTmp)
--- a/modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/SnapshotDownloadStorage.scala
+++ b/modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/SnapshotDownloadStorage.scala
@@ -1,5 +1,7 @@
 package io.constellationnetwork.dag.l0.infrastructure.snapshot
 
+import java.nio.file.NoSuchFileException
+
 import cats.Parallel
 import cats.effect.Async
 import cats.syntax.all._
@@ -149,10 +151,14 @@ object SnapshotDownloadStorage {
           _ <- logger.info(s"Starting cleanup above ordinal ${ordinal.show}")
           _ <- snapshotInfoStorage
             .deleteAbove(ordinal)
-            .handleErrorWith(err =>
-              logger.error(err)(s"Error while deleting snapshot_info files above ${ordinal.show}") >>
-                Async[F].raiseError(err)
-            )
+            .handleErrorWith {
+              case _: java.nio.file.NoSuchFileException =>
+                // Files already deleted - not an error
+                Async[F].unit
+              case err =>
+                logger.error(s"Error while deleting snapshot_info files above ${ordinal.show}: ${err.getMessage}") >>
+                  Async[F].raiseError(err)
+            }
           _ <- logger.info(s"Successfully deleted snapshot_info files above ordinal ${ordinal.show}")
         } yield ()
 
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/allowance_list/Loader.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/allowance_list/Loader.scala
@@ -1,7 +1,9 @@
 package io.constellationnetwork.node.shared.infrastructure.allowance_list
 
+import java.nio.file.NoSuchFileException
+
 import cats.effect.Async
-import cats.syntax.functor._
+import cats.syntax.all._
 
 import io.constellationnetwork.domain.allowance_list.AllowanceListEntry
 import io.constellationnetwork.env.env.AllowanceListPath
@@ -10,6 +12,7 @@ import fs2.data.csv._
 import fs2.io.file.Files
 import fs2.text
 import io.estatico.newtype.ops._
+import org.typelevel.log4cats.slf4j.Slf4jLogger
 
 trait Loader[F[_]] {
   def load(path: AllowanceListPath): F[Set[AllowanceListEntry]]
@@ -18,7 +21,8 @@ trait Loader[F[_]] {
 object Loader {
 
   def make[F[_]: Async]: Loader[F] =
-    (path: AllowanceListPath) =>
+    (path: AllowanceListPath) => {
+      val logger = Slf4jLogger.getLoggerFromName[F]("AllowanceListLoader")
       Files
         .forAsync[F]
         .readAll(path.coerce)
@@ -29,4 +33,11 @@ object Loader {
         .compile
         .toList
         .map(_.toSet)
+        .handleErrorWith {
+          case _: NoSuchFileException =>
+            logger.warn(s"Allowance list file not found at ${path.coerce}, using empty set") >>
+              Set.empty[AllowanceListEntry].pure[F]
+          case e => Async[F].raiseError(e)
+        }
+    }
 }
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/genesis/GenesisFS.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/genesis/GenesisFS.scala
@@ -1,9 +1,11 @@
 package io.constellationnetwork.node.shared.infrastructure.genesis
 
+import java.nio.file.NoSuchFileException
+
 import cats.effect.Async
-import cats.syntax.either._
-import cats.syntax.flatMap._
-import cats.syntax.functor._
+import cats.syntax.all._
+
+import scala.util.control.NoStackTrace
 
 import io.constellationnetwork.ext.json._
 import io.constellationnetwork.json.JsonSerializer
@@ -20,8 +22,15 @@ import io.circe.{Decoder, Encoder}
 
 object GenesisFS {
 
+  case class GenesisFileNotFound(path: Path, fileType: String) extends NoStackTrace {
+    override def getMessage: String = s"Genesis $fileType file not found at $path"
+  }
+
   def make[F[_]: Async: JsonSerializer, S <: FullSnapshot[_, _]: Encoder: Decoder]: GenesisFS[F, S] = new GenesisFS[F, S] {
     def write(genesis: Signed[S], identifier: Address, path: Path): F[Unit] = {
+      // Ensure directory exists before writing
+      val ensureDir = Files.forAsync[F].createDirectories(path)
+
       val writeGenesis = Stream
         .evalSeq(genesis.toBinaryF.map(_.toSeq))
         .through(Files.forAsync[F].writeAll(path / "genesis.snapshot"))
@@ -31,7 +40,7 @@ object GenesisFS {
         .through(text.utf8.encode)
         .through(Files.forAsync[F].writeAll(path / "genesis.address"))
 
-      writeGenesis.merge(writeIdentifier).compile.drain
+      ensureDir >> writeGenesis.merge(writeIdentifier).compile.drain
     }
 
     def loadBalances(path: Path): F[Set[types.GenesisAccount]] =
@@ -48,6 +57,9 @@ object GenesisFS {
         .compile
         .toList
         .map(_.toSet)
+        .adaptError {
+          case _: NoSuchFileException => GenesisFileNotFound(path, "balances")
+        }
 
     def loadSignedGenesis(path: Path): F[Signed[S]] =
       Files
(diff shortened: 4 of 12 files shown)


ryle-ai (commented):
## File I/O Error Handling Review

Thorough review of all 12 changed files, including reading surrounding source code, callers, validators, and the existing error handling patterns.

### What I Verified

- Read the full source of `LocalFileSystemStorage`, all its subclasses, and the `SnapshotLocalFileSystemStorage` class
- Traced the seedlist/allowance-list loading through `TessellationIOApp` into `SignedValidator.validateSignaturesWithSeedlist` to assess security implications
- Checked the `delete()` method already had the same `NoSuchFileException` handling pattern, confirming this PR extends a pre-existing pattern to `exists()`, `link()`, and `move()`
- Verified that `SnapshotLocalFileSystemStorage.write()` calls `link()` on line 56, which is affected by the new silent error handling
- Read the `CurrencySnapshotCleanupStorage` and `SnapshotDownloadStorage` cleanup pipelines end-to-end including verification steps
- Checked wallet `readFromJsonFile`/`writeToJsonFile` callers in the wallet CLI

### Overall Assessment

This is a well-structured defensive error handling PR that addresses real concurrent-deletion race conditions during snapshot cleanup. The core pattern (catch `NoSuchFileException` on file operations that race with cleanup) is sound and follows the existing precedent set by the `delete()` method.

I have a few substantive observations below. The seedlist/allowance-list loader changes and the stack trace loss in logging are worth discussing, but neither is a blocker.

### Key Observations

**1. Seedlist/AllowanceList loaders returning empty sets on missing files** -- The current calling pattern in `TessellationIOApp` is `seedlistPath.traverse(loader.load)`. Before this PR, a configured-but-missing file would crash the node at startup. After this PR, it silently returns `Some(emptySet)`. I traced this through `SignedValidator.validateSignaturesWithSeedlist` and confirmed that `Some(emptySet)` causes ALL signers to fail validation (since `diff(emptySet)` returns the full signer set). So this is not a security bypass -- it's actually *more* restrictive than `None`. However, it means a misconfigured seedlist path would cause the node to reject all peers silently rather than fail fast at startup with a clear error. The warn-level logging helps, but a node operator might not notice until the node is already running and rejecting everything. This is a reasonable tradeoff for preventing crash-on-missing-optional-file, but worth being aware of.

**2. `link()` error suppression affects `SnapshotLocalFileSystemStorage.write()`** -- The `write()` method at line 56 calls `link(hashName, ordinalName)`. With the new error handling, if the source hash file was concurrently removed between the `write(hashName, snapshot)` call and the `link()` call, the link silently fails and the snapshot appears written but has no ordinal reference. This is a very narrow race window and in practice the hash file wouldn't be deleted during a write operation (cleanup targets old ordinals, not files being actively written). But it's worth noting that the base-class `link()` error handling is quite broad -- it applies to all callers, not just cleanup paths.

**3. Stack traces lost in logging format change** -- Several places change from `logger.error(err)("message")` to `logger.error(s"message: ${err.getMessage}")`. The old format passes the exception object to SLF4J, which preserves the full stack trace in log output. The new format only includes the exception message string. For the `NoSuchFileException` cases this doesn't matter (those are expected and handled). But the catch-all `case err =>` branches for unexpected errors also use the new format, which means unexpected failures lose their stack traces. This could make production debugging harder for the non-expected error cases.

**4. Log level inconsistency across methods** -- Minor: `delete()` logs `NoSuchFileException` at `info` level, while the new `move()` and `link()` handlers log at `warn` level. Not a problem, but worth being consistent if you care about log-level filtering.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/seedlist/Loader.scala
```diff
@@ -29,4 +33,11 @@ object Loader {
         .compile
         .toList
         .map(_.toSet)
+        .handleErrorWith {
+          case _: NoSuchFileException =>
```
ryle-ai, line 37: When a seedlist path is configured but the file is missing, this now returns `Set.empty` wrapped in `Some(...)` by the calling `.traverse()` in `TessellationIOApp`. I traced this through to `SignedValidator.validateSignaturesWithSeedlist` -- `Some(emptySet)` causes all signers to fail the `diff` check, so every peer gets rejected. That's secure (not a bypass), but it means a misconfigured path silently makes the node reject all peers rather than failing fast at startup with a clear error.

The warn log helps, but you might want to consider whether this should remain a hard failure for seedlist specifically (since a configured seedlist that can't be loaded is more likely a deployment mistake than a transient condition).

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/allowance_list/Loader.scala
```diff
@@ -29,4 +33,11 @@ object Loader {
         .compile
         .toList
         .map(_.toSet)
+        .handleErrorWith {
+          case _: NoSuchFileException =>
```
ryle-ai, line 37: Same observation as the seedlist loader -- a configured but missing allowance list file now silently returns an empty set. In `GlobalSnapshotStateChannelAcceptanceManager.filterWithAllowedPeers`, `Some(emptyNonEmptySet)` would filter out all signatures, while `None` allows all through. So this is safe from a security perspective (empty = block all, not allow all).

But it converts a clear startup failure into a subtle runtime behavior change. If the operator configures a path but the file is absent, the node starts but rejects all state channels.

modules/shared/src/main/scala/io/constellationnetwork/storage/LocalFileSystemStorage.scala
```diff
@@ -102,13 +105,26 @@ abstract class LocalFileSystemStorage[F[_], A](baseDir: Path)(
       case (src, dst) =>
         F.blocking(dst.parent.createDirectoryIfNotExists(createParents = true)) >>
           F.blocking(src.linkTo(dst)).void
-    }.flatten
+    }.flatten.handleErrorWith {
+      case _: NoSuchFileException =>
+        logger.warn(s"Cannot link $fileName to $to: source file does not exist or was removed")
+      case _: FileAlreadyExistsException =>
+        logger.debug(s"Link $to already exists, skipping")
+      case e => F.raiseError(e)
```
ryle-ai, line 113: The `link()` error handling here is quite broad since it applies to all callers of the base class. In particular, `SnapshotLocalFileSystemStorage.write()` calls `link(hashName, ordinalName)` at line 56 -- if a `NoSuchFileException` occurs during a write operation (not just cleanup), the link would silently not be created and the snapshot would lack its ordinal reference.

In practice this race window is extremely narrow during writes (cleanup doesn't target files being actively written), so I think it's fine. Just noting that this is a base-class change that affects all paths, not just cleanup.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
```diff
@@ -164,9 +168,15 @@ abstract class SnapshotLocalFileSystemStorage[
...
                 for {
                   hashed <- snapshot.toHashed
-                  _ <- movePersistedToTmp(hashed.hash, hashed.ordinal).handleErrorWith { err =>
-                    logger.warn(err)(s"Failed to move persisted to tmp for ordinal=${snapshot.ordinal}, hash=${hashed.hash}")
-                    Async[F].raiseError(err)
+                  _ <- movePersistedToTmp(hashed.hash, hashed.ordinal).handleErrorWith {
+                    case _: java.nio.file.NoSuchFileException =>
+                      // File already moved/deleted - this is expected during concurrent cleanup
+                      logger.debug(s"File already removed during cleanup for ordinal=${snapshot.ordinal}")
+                    case err =>
```
ryle-ai, line 175: The change from `logger.warn(err)(...)` to `logger.warn(s"...: ${err.getMessage}")` drops the full stack trace from the log output for the non-`NoSuchFileException` case. The `case err =>` branch handles *unexpected* errors where the stack trace is most valuable for debugging. Consider keeping the exception object in the logging call for unexpected errors:
```scala
case err =>
  logger.warn(err)(s"Failed to move persisted to tmp for ordinal=${snapshot.ordinal}, hash=${hashed.hash}")
  Async[F].raiseError(err)
```

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/infrastructure/snapshot/CurrencySnapshotCleanupStorage.scala
```diff
@@ -47,9 +49,13 @@ object CurrencySnapshotCleanupStorage {
...
           .cleanupAboveOrdinal(ordinal, deletePersisted)
-          .handleErrorWith { err =>
-            logger.error(err)("Error during cleanup snapshot of the metagraph") >>
-              CurrencyCleanupError.raiseError[F, Unit]
+          .handleErrorWith {
+            case _: java.nio.file.NoSuchFileException =>
+              // Files already cleaned up - not an error
+              Async[F].unit
+            case err =>
+              logger.error(s"Error during cleanup snapshot of the metagraph: ${err.getMessage}") >>
```
ryle-ai: Same stack-trace observation: the `case err =>` branch now uses `logger.error(s"...: ${err.getMessage}")` instead of `logger.error(err)("...")`. For unexpected cleanup errors (not `NoSuchFileException`), losing the stack trace makes production debugging harder. Consider preserving the exception object in the error logging for the catch-all case.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/SnapshotDownloadStorage.scala
```diff
@@ -149,10 +151,14 @@ object SnapshotDownloadStorage {
...
-            .handleErrorWith(err =>
-              logger.error(err)(s"Error while deleting snapshot_info files above ${ordinal.show}") >>
-                Async[F].raiseError(err)
-            )
+            .handleErrorWith {
+              case _: java.nio.file.NoSuchFileException =>
+                // Files already deleted - not an error
+                Async[F].unit
+              case err =>
+                logger.error(s"Error while deleting snapshot_info files above ${ordinal.show}: ${err.getMessage}") >>
```
ryle-ai: Same pattern: `logger.error(s"...: ${err.getMessage}")` loses the stack trace for non-`NoSuchFileException` errors. The `NoSuchFileException` branch is good -- those are expected and don't need stack traces. But the catch-all should preserve the exception for diagnostics.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
```diff
@@ -187,7 +201,7 @@ abstract class SnapshotLocalFileSystemStorage[
     ordinal: SnapshotOrdinal,
     movePersistedToTmp: (Hash, SnapshotOrdinal) => F[Unit]
   )(implicit hs: HasherSelector[F], kryoSerializer: KryoSerializer[F]): F[Unit] = for {
-    _ <- logger.info(s"Searching for persisted files above ordinal ${ordinal.show}")
+    _ <- logger.debug(s"Searching for persisted files above ordinal ${ordinal.show}")
```
ryle-ai, line 204: Good change reducing these from `info` to `debug`. During cleanup of large ordinal ranges, these were likely generating substantial log volume with limited diagnostic value.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/genesis/GenesisFS.scala
```diff
@@ -20,8 +22,15 @@ import io.circe.{Decoder, Encoder}
...
 object GenesisFS {
 
+  case class GenesisFileNotFound(path: Path, fileType: String) extends NoStackTrace {
+    override def getMessage: String = s"Genesis $fileType file not found at $path"
+  }
+
   def make[F[_]: Async: JsonSerializer, S <: FullSnapshot[_, _]: Encoder: Decoder]: GenesisFS[F, S] = new GenesisFS[F, S] {
     def write(genesis: Signed[S], identifier: Address, path: Path): F[Unit] = {
+      // Ensure directory exists before writing
+      val ensureDir = Files.forAsync[F].createDirectories(path)
```
ryle-ai, line 32: The `ensureDir` call uses `Files.forAsync[F].createDirectories(path)` where `path` is the genesis directory. This is the right approach -- `writeAll` downstream doesn't create parent directories, so this prevents failures when the genesis path doesn't exist yet.

modules/wallet/src/main/scala/io/constellationnetwork/wallet/file/io.scala
```diff
@@ -19,14 +22,23 @@ object io {
       .through(decoder[F, A])
       .compile
       .last
+      .handleErrorWith {
+        case _: NoSuchFileException => none[A].pure[F]
+        case e                      => Async[F].raiseError(e)
```
ryle-ai: `readFromJsonFile` now returns `None` for missing files instead of throwing. I checked the wallet CLI callers -- most use `.map(_.get)` or `OptionT(...).getOrRaise(...)` patterns that still fail appropriately when the file is truly required. This is a clean change.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/SnapshotLocalFileSystemStorage.scala
```diff
@@ -175,9 +185,13 @@ abstract class SnapshotLocalFileSystemStorage[
           }
         } yield ()
 
-        operation.handleErrorWith { err =>
-          logger.error(err)(s"Failed to process file with ordinal $fileOrdinal") >>
+        operation.handleErrorWith {
+          case _: java.nio.file.NoSuchFileException =>
```
ryle-ai, line 189: The outer `operation.handleErrorWith` now distinguishes `NoSuchFileException` (returns unit silently) from other errors (logs warning, returns unit). Both cases swallow the error -- the only difference is logging level. This is consistent with the pre-existing behavior where all errors in `processFileChunk` were swallowed to allow cleanup to continue. The `NoSuchFileException` case just gets quieter treatment, which makes sense.

modules/wallet/src/main/scala/io/constellationnetwork/wallet/file/io.scala
```diff
@@ -19,14 +22,23 @@ object io {
       .through(decoder[F, A])
       .compile
       .last
+      .handleErrorWith {
```
ryle-ai: Three callers in `Main.scala` (lines 118-119, 162-163, 191-192) use the pattern:
```scala
readFromJsonFile[IO, SomeType](path).map(x => x.get)
```

Before this PR, a missing file throws `NoSuchFileException` from `Files.readAll`, which propagates out with a clear message like `NoSuchFileException: /path/to/reference.json`. After this PR, a missing file returns `None`, and `.get` on `None` throws `NoSuchElementException: None.get` — a much less descriptive error for the wallet CLI user.

Also, the caller at line 247 wraps `readFromJsonFile` in `.handleErrorWith` to log the error before re-raising:
```scala
readFromJsonFile[F, Signed[Transaction]](path)
  .handleErrorWith(e =>
    logger.error(e)(s"Error while reading previous transaction from path $path") >> ...
  )
```
After this PR, a missing file returns `None` instead of raising, so the `.handleErrorWith` never fires and the error log is silently skipped.

Neither is a crash or data corruption issue — the wallet CLI still fails with an error for the `.get` callers and proceeds safely for the `.handleErrorWith` caller. But the error messages become less useful for the user.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/SnapshotDownloadStorage.scala
```diff
@@ -149,10 +151,14 @@ object SnapshotDownloadStorage {
...
+                  Async[F].raiseError(err)
+            }
           _ <- logger.info(s"Successfully deleted snapshot_info files above ordinal ${ordinal.show}")
```
ryle-ai: After the new `NoSuchFileException` handler returns `Async[F].unit`, the for-comprehension continues here and logs `"Successfully deleted snapshot_info files above ordinal ..."`. This fires even when no deletion occurred (e.g., the directory didn't exist yet on a fresh node, or files were already removed by a concurrent cleanup). Not a functional issue, but misleading for operators reviewing logs — they'd see a success message when the operation was actually a no-op.

ryle-ai (commented):
### Deep review: additional observations

Two issues not covered in the prior review, both in the wallet module and the SnapshotDownloadStorage cleanup flow.

ottobot-ai:
## Response to Review Feedback

Thanks for the thorough review @ryle-ai! I've pushed fixes addressing the actionable items.

### ✅ Fixed (commit 18b9c0d6)

**1. Stack traces preserved for unexpected errors**
Changed error logging back to `logger.error(err)(...)` / `logger.warn(err)(...)` pattern in:
- `CurrencySnapshotCleanupStorage.scala` 
- `SnapshotDownloadStorage.scala`
- `SnapshotLocalFileSystemStorage.scala`

**2. Misleading success log fixed**
`SnapshotDownloadStorage.cleanupAbove`: Changed from `logger.info("Successfully deleted...")` to `logger.debug("Completed snapshot_info cleanup...")` - no longer claims success when the operation was a no-op.

---

### 📝 Acknowledged (no code changes needed)

**Seedlist/AllowanceList empty set behavior**
Agreed this is a tradeoff worth being aware of. A misconfigured path now silently rejects all peers rather than failing fast. The warn-level logging helps, but operators should monitor for these warnings. This behavior is consistent with the "optional file" semantics - if you configure a seedlist path, ensure the file exists.

**`link()` error suppression breadth**  
Understood that the base-class change affects all callers including writes. The race window for concurrent cleanup during active writes is extremely narrow in practice, and cleanup targets old ordinals not files being actively written. Acceptable tradeoff for preventing cleanup stalls.

**Wallet `readFromJsonFile` returning `None`**
Valid observation - callers using `.get` will get less descriptive `NoSuchElementException` vs `NoSuchFileException`. The affected callers in the wallet CLI will still fail appropriately, just with a less helpful message. This is a minor UX regression for rare error cases, not a correctness issue.

---

Fixes available at: `ottobot-ai/tessellation:fix/comprehensive-file-io-error-handling-v2`

James will need to pull and push to the PR branch since fork commits can't access CI secrets.

ottobot-ai:
**Update:** Fixed 3 additional places in `SnapshotLocalFileSystemStorage.scala` that were also using `${err.getMessage}` instead of `logger.warn(err)(...)`. All error logging in the PR-touched files now preserves full stack traces.

Branch updated: `ottobot-ai/tessellation:fix/comprehensive-file-io-error-handling-v2` (commit f68a876e)

ottobot-ai:
**Update 2:** Addressed wallet file comments.

Reverted the `readFromJsonFile` error handling change — the original behavior (propagating `NoSuchFileException`) is correct here. The callers use `.get` and `.handleErrorWith` patterns that expect exceptions, so returning `None` was a regression:

```scala
// Before PR: Clear error
NoSuchFileException: /path/to/reference.json

// After PR change: Unhelpful  
NoSuchElementException: None.get
```

Kept the `writeToJsonFile` improvement (ensure parent directory exists).

Branch updated: `ottobot-ai/tessellation:fix/comprehensive-file-io-error-handling-v2` (commit 4c6d6b90)

marcus-girolneto (approved):

PR #1441, opened 2026-02-12
feat: clickhouse metrics

effects-ai:


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/metrics/CompositeMetrics.scala
```diff
@@ -0,0 +1,120 @@
...
+
+    def incrementCounterBy(key: MetricKey, value: Double, tags: TagSeq): F[Unit] =
+      prometheus.incrementCounterBy(key, value, tags) >> writeToClickHouse(key, value, MetricType.Counter, tags)
+
+    // Time
+    def recordTime(key: MetricKey, duration: FiniteDuration, tags: TagSeq): F[Unit] =
+      prometheus.recordTime(key, duration, tags) >> writeToClickHouse(key, duration.toMillis.toDouble, MetricType.Histogram, tags)
+
+    def timedMetric[A](operation: F[A], metricKey: MetricKey, tags: TagSeq): F[A] =
+      prometheus.timedMetric(operation, metricKey, tags) // ClickHouse write happens via recordTime
```
ryle-ai: Bug: this delegates to `prometheus.timedMetric`, which internally calls `prometheus.recordTime` (not `this.recordTime`). ClickHouse never receives the timing data despite the comment claiming otherwise.

Fix by implementing the timing logic here:
```scala
def timedMetric[A](operation: F[A], metricKey: MetricKey, tags: TagSeq): F[A] =
  Async[F].realTime.flatMap { start =>
    operation.flatTap { _ =>
      Async[F].realTime.flatMap { end =>
        val duration = FiniteDuration(end.toNanos - start.toNanos, TimeUnit.NANOSECONDS)
        recordTime(metricKey, duration, tags)
      }
    }
  }
```
This ensures `this.recordTime` is called, which writes to both Prometheus and ClickHouse.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/metrics/MetricsFactory.scala
```diff
@@ -0,0 +1,79 @@
...
+    environment: AppEnvironment,
+    appConfig: ClickHouseAppConfig,
+    logger: SelfAwareStructuredLogger[F]
+  ): Resource[F, Option[MetricsSink[F]]] =
+    appConfig.metricsTableName match {
+      case Some(metricsTableName) =>
+        val configResult = ClickHouseConfig.fromAppConfig(appConfig)
+        configResult match {
+          case Right(Some(baseConfig)) =>
+            val metricsConfig = baseConfig.copy(tableName = metricsTableName)
```
ryle-ai: `metricsTableName` is copied into `baseConfig.tableName` without going through `ClickHouseConfig.validateId`. The original `tableName` was validated at line 61 of `ClickHouseConfig.fromAppConfig`, but this substitution happens after that validation.

Add validation before the copy:
```scala
val metricsConfig = ClickHouseConfig.validateId("metricsTableName", metricsTableName) match {
  case Right(validated) => baseConfig.copy(tableName = validated)
  case Left(err) => // handle error or raise
}
```
(This would require making `validateId` public or adding a dedicated method.)

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/metrics/ClickHouseMetricsSink.scala
```diff
@@ -0,0 +1,187 @@
...
+      case MetricType.Distribution => "distribution"
+    }
+
+    private def tagsToMapString(tags: TagSeq): String = {
+      val pairs = tags.map { case (k, v) => s"'${escapeSingleQuotes(k.value)}':'${escapeSingleQuotes(v)}'" }.mkString(",")
+      s"{$pairs}"
+    }
+
+    private def escapeSingleQuotes(s: String): String =
+      s.replace("'", "\\'")
```
ryle-ai: ClickHouse's default string escaping uses doubled single quotes (`''`), not backslash (`\'`). The backslash escape only works when `allow_experimental_analyzer` or specific `sql_mode` settings are enabled. Since the tag values are passed as a string literal in a `setString` call (not a parameterized map value), this could produce malformed queries if a tag value contains single quotes.

The existing `ClickHouseSink` avoids this entirely by using only parameterized values. Consider either:
- Using `s.replace("'", "''")` for ClickHouse-standard escaping, or
- Finding a way to pass the map as a parameterized value through the JDBC driver

ryle-ai (changes requested):
## `timedMetric` silently drops ClickHouse writes

`CompositeMetrics.timedMetric` delegates directly to `prometheus.timedMetric`, with the comment "ClickHouse write happens via recordTime." This is incorrect. The Prometheus implementation of `timedMetric` calls `this.recordTime(...)` internally -- but `this` is the Prometheus instance, not the CompositeMetrics wrapper. The ClickHouse sink never receives timing data from `timedMetric` calls.

Fix: implement `timedMetric` in CompositeMetrics directly (time the operation, then call `this.recordTime`).

## `metricsTableName` bypasses identifier validation

`MetricsFactory.makeClickHouseSink` calls `ClickHouseConfig.fromAppConfig(appConfig)` and then does `baseConfig.copy(tableName = metricsTableName)`. The problem is that `fromAppConfig` validates the original `tableName` through `validateId("tableName", _)` which enforces `^[a-zA-Z_][a-zA-Z0-9_]{0,63}$`, but `metricsTableName` is substituted *after* validation. The value goes straight from the environment variable into DDL and INSERT string interpolation without any identifier check.

The risk is bounded by this being an env-var-controlled config rather than user input, but it's inconsistent with how every other identifier in `ClickHouseConfig` is validated and it would be trivial to add the same `validateId` call.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/metrics/CompositeMetrics.scala
ryle-ai: `prometheus.timedMetric` calls `this.recordTime` inside the Prometheus impl -- `this` is the Prometheus instance, not this CompositeMetrics wrapper. ClickHouse never gets timing data.

Fix by implementing the timing directly:
```scala
def timedMetric[A](operation: F[A], metricKey: MetricKey, tags: TagSeq): F[A] =
  Async[F].realTime.flatMap { start =>
    operation.flatTap { _ =>
      Async[F].realTime.flatMap { end =>
        val duration = FiniteDuration(end.toNanos - start.toNanos, TimeUnit.NANOSECONDS)
        recordTime(metricKey, duration, tags)
      }
    }
  }
```
No callers today, but this will bite someone eventually.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/metrics/MetricsFactory.scala
ryle-ai: `metricsTableName` goes straight from the env var into `ClickHouseConfig.tableName` without `validateId`. The original `tableName` was validated at `ClickHouseConfig.fromAppConfig` line 61, but this substitution happens after. The value ends up in DDL and INSERT string interpolation unvalidated.

Since `validateId` is private, simplest fix is to add an inline check here:
```scala
val metricsConfig = if (metricsTableName.matches("^[a-zA-Z_][a-zA-Z0-9_]{0,63}$"))
  baseConfig.copy(tableName = metricsTableName)
else
  // handle error
```
Or make `validateId` package-accessible.

ryle-ai (commented):
## `timedMetric` silently skips ClickHouse

`CompositeMetrics.timedMetric` delegates to `prometheus.timedMetric`, with a comment claiming ClickHouse writes happen via `recordTime`. This is incorrect -- the Prometheus implementation's `timedMetric` (Metrics.scala:277) calls `recordTime` on itself (`this`), which is the Prometheus instance, not the CompositeMetrics wrapper. ClickHouse never receives timing data.

Nobody calls `timedMetric` in the codebase today so this has zero current impact, but it's a trap waiting for the first caller. Fix is straightforward: implement the timing logic directly in CompositeMetrics so `this.recordTime` is called.

## `metricsTableName` skips identifier validation

`MetricsFactory.makeClickHouseSink` calls `ClickHouseConfig.fromAppConfig(appConfig)` which validates `tableName` through `validateId` (enforces `^[a-zA-Z_][a-zA-Z0-9_]{0,63}$`), then immediately replaces it with `baseConfig.copy(tableName = metricsTableName)` -- the unvalidated value from the env var. That value flows into `createTableDDL` and `insertSql` via string interpolation.

Risk is bounded by this being an env-var rather than user input, but it's inconsistent with how every other identifier in `ClickHouseConfig` is validated. Since `validateId` is private, you'd need to either make it package-private or add a one-liner wrapper.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/metrics/CompositeMetrics.scala
ryle-ai: `prometheus.timedMetric` calls `recordTime` on the Prometheus instance (`this` = Prometheus, not this CompositeMetrics wrapper). The comment "ClickHouse write happens via recordTime" is incorrect. No callers today and the method is deprecated upstream, but the comment is actively misleading — either implement timing here directly or update the comment to reflect reality.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/metrics/MetricsFactory.scala
```diff
@@ -0,0 +1,79 @@
...
+    nodeId: PeerId,
+    environment: AppEnvironment,
+    appConfig: ClickHouseAppConfig,
+    logger: SelfAwareStructuredLogger[F]
+  ): Resource[F, Option[MetricsSink[F]]] =
+    appConfig.metricsTableName match {
+      case Some(metricsTableName) =>
+        val configResult = ClickHouseConfig.fromAppConfig(appConfig)
+        configResult match {
+          case Right(Some(baseConfig)) =>
```
ryle-ai: `metricsTableName` goes into `baseConfig.copy(tableName = ...)` without the `validateId` check that `fromAppConfig` applies to the original `tableName`. The value then hits `createTableDDL` and `insertSql` via string interpolation. Since `validateId` is private to `ClickHouseConfig`, simplest fix is an inline check:
```scala
if (!metricsTableName.matches("^[a-zA-Z_][a-zA-Z0-9_]{0,63}$"))
  // return Left / raise error
```

ryle-ai (commented):
## `timedMetric` delegation and unvalidated table name

Two issues, both previously flagged and still unaddressed:

1. `CompositeMetrics.timedMetric` delegates to `prometheus.timedMetric`, which calls `this.recordTime` on the Prometheus instance — ClickHouse never sees timing data. The comment claiming otherwise is wrong. Nobody calls `timedMetric` today (and it's marked deprecated in the Prometheus impl), so zero current impact, but the misleading comment should at minimum be corrected.

2. `MetricsFactory.makeClickHouseSink` calls `ClickHouseConfig.fromAppConfig(appConfig)` — which validates `tableName` through `validateId` (enforces `^[a-zA-Z_][a-zA-Z0-9_]{0,63}$`) — then immediately replaces it with the unvalidated `metricsTableName` from the env var. That value flows into `createTableDDL` and `insertSql` via string interpolation. Risk is bounded (env-var, not user input), but the existing `ClickHouseSink` avoids this by validating all identifiers before interpolation. One-liner fix: apply the same regex check before the `copy`.

modules/node-shared/src/main/resources/application.conf
```diff
@@ -271,15 +271,16 @@ click-house-config {
...
  retention-period-in-days = 30
  error-pause-duration = 30 seconds
 
- host = ${?CLICKHOUSE_HOST}
- user = ${?CLICKHOUSE_USER}
- password = ${?CLICKHOUSE_PASSWORD}
- table-name = ${?CLICKHOUSE_TABLE_NAME}
+ host = "ajg2gs104i.us-west-2.aws.clickhouse.cloud"
+ user = "default"
+ password = "o0.hBE28MYNLl"
```
ryle-ai: Hardcoded plaintext password for a ClickHouse Cloud instance. The rest of the codebase uses `${?ENV_VAR}` for credentials — this breaks that pattern and will be visible in git history permanently. Restore the env-var references:
```
 host = ${?CLICKHOUSE_HOST}
 user = ${?CLICKHOUSE_USER}
 password = ${?CLICKHOUSE_PASSWORD}
 table-name = ${?CLICKHOUSE_TABLE_NAME}
 metrics-table-name = ${?CLICKHOUSE_METRICS_TABLE_NAME}
```

docker/grafana/provisioning/datasources/clickhouse.yaml
```diff
@@ -0,0 +1,16 @@
...
+    editable: true
+    jsonData:
+      host: ajg2gs104i.us-west-2.aws.clickhouse.cloud
+      port: 8443
+      protocol: http
+      secure: true
+      username: default
+      defaultDatabase: default
+    secureJsonData:
+      password: o0.hBE28MYNLl
```
ryle-ai: Same hardcoded credentials as in `application.conf`. This file should either use Grafana environment variable substitution (`$CLICKHOUSE_PASSWORD`) or be excluded from source control via `.gitignore`.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/metrics/CompositeMetrics.scala
ryle-ai: This delegates to `prometheus.timedMetric`, which calls `this.recordTime` on the Prometheus instance — not this CompositeMetrics wrapper. ClickHouse never receives timing data. The comment claiming otherwise is incorrect.

Nobody calls `timedMetric` today (it's deprecated upstream), but either implement the timing here directly so `this.recordTime` is called, or at minimum correct the comment so it doesn't mislead the next person who tries to use it.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/metrics/MetricsFactory.scala
```diff
@@ -0,0 +1,82 @@
...
+            val metricsConfig = baseConfig.copy(tableName = metricsTableName)
```
ryle-ai: `metricsTableName` goes into `baseConfig.copy(tableName = ...)` without the `validateId` check that `fromAppConfig` applies to the original `tableName`. The value then flows into `createTableDDL` and `insertSql` via string interpolation.

Risk is bounded (env-var, not user input), but the existing `ClickHouseSink` validates all identifiers before interpolation. Since `validateId` is private, simplest fix:
```scala
if (!metricsTableName.matches("^[a-zA-Z_][a-zA-Z0-9_]{0,63}$"))
  // raise error / return None
```

ryle-ai (dismissed):
## Hardcoded ClickHouse credentials

`application.conf` replaces the `${?CLICKHOUSE_HOST}`, `${?CLICKHOUSE_USER}`, `${?CLICKHOUSE_PASSWORD}`, and `${?CLICKHOUSE_TABLE_NAME}` env-var references with hardcoded values — including a plaintext password for a ClickHouse Cloud instance. The same credentials are duplicated in `docker/grafana/provisioning/datasources/clickhouse.yaml`.

These will persist in git history even if removed in a follow-up commit. Restore the `${?ENV_VAR}` pattern and inject credentials at deployment time. If you need defaults for local dev, use obviously-fake placeholders.

Two other issues have been flagged in three prior review rounds and remain unaddressed — see inline comments on `CompositeMetrics.timedMetric` and `MetricsFactory` table name validation.

modules/node-shared/src/main/resources/application.conf
```diff
@@ -268,10 +268,14 @@ snapshot-binary-sender-timeouts {
 }
 
 click-house-config {
+ logs-enabled = false
```
ryle-ai: This defaults ClickHouse logs to disabled. Previously, the code attempted ClickHouse logging unconditionally for DagL0/DagL1 layers (falling back to SLF4J on failure). With this default, existing deployments with ClickHouse configured will silently lose log forwarding unless they add `CLICKHOUSE_LOGS_ENABLED=true`. If intentional, worth noting as a breaking change for operators.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/metrics/CompositeMetrics.scala
```diff
@@ -0,0 +1,120 @@
...
+
```
ryle-ai, line 89: Still present from prior reviews: `prometheus.timedMetric` internally calls `this.recordTime` where `this` is the Prometheus instance (Metrics.scala:277), not this CompositeMetrics wrapper. ClickHouse never receives timing data. The comment is incorrect.

No callers today (method is deprecated upstream), but either fix the delegation or correct the comment before someone relies on it.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/metrics/MetricsFactory.scala
```diff
@@ -0,0 +1,86 @@
...
+    (appConfig.metricsEnabled, appConfig.metricsTableName) match {
+      case (false, _) =>
+        Resource.eval(
+          logger.info("ClickHouse metrics disabled via configuration. Metrics will only go to Prometheus.")
+        ) >> Resource.pure(none[MetricsSink[F]])
+      case (_, Some(metricsTableName)) =>
+        val configResult = ClickHouseConfig.fromAppConfig(appConfig)
+        configResult match {
+          case Right(Some(baseConfig)) =>
+            val metricsConfig = baseConfig.copy(tableName = metricsTableName)
```
ryle-ai: Still present from prior reviews: `metricsTableName` replaces the validated `tableName` without going through `validateId` (which is private to `ClickHouseConfig`). The value flows into `createTableDDL` and `insertSql` via string interpolation.

One-liner fix: `if (!metricsTableName.matches("^[a-zA-Z_][a-zA-Z0-9_]{0,63}$"))` before the copy.

ryle-ai (commented):
Two issues from prior reviews remain unaddressed:

1. **`CompositeMetrics.timedMetric`** delegates to the Prometheus instance's `timedMetric`, which calls `recordTime` on itself — ClickHouse never receives timing data. The inline comment claiming otherwise is wrong. No callers today, but the comment is a trap.

2. **`metricsTableName`** bypasses `validateId` in `MetricsFactory.makeClickHouseSink`. The value goes into DDL and INSERT via string interpolation without the identifier check that `fromAppConfig` applies to the original `tableName`. Bounded risk (env-var source), but inconsistent with the rest of `ClickHouseConfig`.

New: `logs-enabled = false` as the default is a behavior change — ClickHouse log forwarding was previously attempted unconditionally for DagL0/DagL1 when configured (falling back to SLF4J on failure). Existing deployments will silently lose log forwarding unless `CLICKHOUSE_LOGS_ENABLED=true` is set.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/config/types.scala
```diff
@@ -191,7 +185,8 @@ object types {
     host: Option[String],
     user: Option[String],
     password: Option[String],
-    tableName: Option[String],
+    logsTableName: Option[String],
```
marcus-constellation-ai, line 189: `logsTableName` maps to HOCON key `logs-table-name` (PureConfig KebabCase default), but `application.conf` still has `table-name = ${?CLICKHOUSE_TABLE_NAME}`. `logsTableName` will always be `None` for any existing deployment, silently breaking ClickHouse log forwarding without any error. Add `logs-table-name = ${?CLICKHOUSE_LOGS_TABLE_NAME}` to `application.conf` and document the migration.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/config/types.scala
```diff
@@ -191,7 +185,8 @@ object types {
...
+    metricsTableName: Option[String],
```
marcus-constellation-ai, line 190: Similarly, `metricsTableName` maps to `metrics-table-name` but no such key exists in `application.conf`. Metrics ClickHouse sink will never activate unless operators discover and set a new, undocumented env var. Add `metrics-table-name = ${?CLICKHOUSE_METRICS_TABLE_NAME}` to `application.conf`.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/metrics/ClickHouseMetricsSink.scala
```diff
@@ -0,0 +1,187 @@
...
+       |    node_id LowCardinality(String),
+       |    network_id LowCardinality(String),
+       |    metric_name LowCardinality(String),
+       |    metric_type Enum8('gauge' = 1, 'counter' = 2, 'histogram' = 3, 'distribution' = 4),
+       |    value Float64,
+       |    tags Map(String, String)
+       |) ENGINE = MergeTree()
+       |PARTITION BY (network_id, toYYYYMM(timestamp))
+       |ORDER BY (node_id, metric_name, timestamp)
+       |TTL toDateTime(timestamp) + INTERVAL 90 DAY""".stripMargin
```
marcus-constellation-ai: TTL is hardcoded to 90 days, but `ClickHouseConfig` has `retentionPeriodInDays`. The log sink presumably uses the configured value. For consistency, use `config.retentionPeriodInDays` here: `TTL toDateTime(timestamp) + INTERVAL ${config.retentionPeriodInDays} DAY`.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/metrics/ClickHouseMetricsSink.scala
marcus-constellation-ai: `s.replace("'", "\\'")` uses backslash escaping for the Map literal that's passed via `setString`. ClickHouse's SQL-standard escape for single quotes in string literals is `''` (doubled), not `\'`. The backslash form relies on `allow_backslash_escaping_in_strings = 1` (the ClickHouse default), but this can be disabled. Consider `s.replace("'", "''")` to follow the SQL standard.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/metrics/MicrometerRegistryScraper.scala
```diff
@@ -0,0 +1,81 @@
...
+      meters.traverse_ { meter =>
+        val id = meter.getId
+        val baseName = id.getName.replace('.', '_')
+
+        if (baseName.startsWith("dag_"))
+          Async[F].unit
+        else {
+          val tags = id.getTags.asScala.toList.map { tag =>
+            (unsafeLabelName(tag.getKey.replace('.', '_')), tag.getValue)
+          }
```
marcus-constellation-ai, line 30: `unsafeLabelName` calls `Refined.unsafeApply` and throws an unchecked exception if the tag key doesn't match `[a-z0-9]+(?:_[a-z0-9]+)*` after dot replacement. Current JVM binder keys are all lowercase so this is fine today, but any future binder or common-tag with hyphens or uppercase would crash the entire scrape loop. Consider wrapping with `refineV[MatchesRegex[...]](...).toOption` and skipping non-conforming tag keys.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/metrics/ClickHouseMetricsSink.scala
```diff
@@ -0,0 +1,187 @@
...
+  ): Resource[F, Unit] = {
+    val flush: F[Unit] = for {
+      now <- Async[F].realTime.map(_.toMillis)
+      paused <- pausedUntil.get
+      _ <-
+        if (now < paused) Async[F].unit
+        else
+          queue.tryTakeN(Some(config.batchSize)).flatMap { entries =>
+            val (ready, notReady) = entries.partition(_.retryAfter <= now)
+            notReady.traverse_(e => queue.tryOffer(e)) >>
```
marcus-constellation-ai: `notReady.traverse_(e => queue.tryOffer(e))` silently drops items if the bounded queue is full during an error pause. These are entries with remaining retries that were already dequeued. Consider logging a warning when `tryOffer` returns `false` here, so dropped retryable metrics are at least observable.

marcus-constellation-ai (changes requested):
## Review

### Breaking: `application.conf` not updated for renamed config field

The PR renames `tableName: Option[String]` to `logsTableName: Option[String]` and adds `metricsTableName: Option[String]` in `ClickHouseAppConfig`. With PureConfig's default KebabCase mapping:
- `logsTableName` → expects HOCON key `logs-table-name`
- `metricsTableName` → expects HOCON key `metrics-table-name`

But `application.conf` still has:
```hocon
table-name = ${?CLICKHOUSE_TABLE_NAME}
```

This means:
1. `logsTableName` will always be `None` because `logs-table-name` doesn't exist in application.conf. Existing deployments that set `CLICKHOUSE_TABLE_NAME` will silently lose ClickHouse log forwarding.
2. `metricsTableName` is also never set via application.conf (no `metrics-table-name` key or env var binding).

The fix requires adding to `application.conf`:
```hocon
logs-table-name = ${?CLICKHOUSE_LOGS_TABLE_NAME}
metrics-table-name = ${?CLICKHOUSE_METRICS_TABLE_NAME}
```
and documenting the migration from `CLICKHOUSE_TABLE_NAME` to `CLICKHOUSE_LOGS_TABLE_NAME`.

### Non-blocking issues

**Hardcoded TTL in createTableDDL** (`ClickHouseMetricsSink.scala`): The DDL hardcodes `INTERVAL 90 DAY` while `ClickHouseConfig` has a `retentionPeriodInDays` field already used by the log sink. This inconsistency means the metrics table ignores the configured retention period.

**`tagsToMapString` backslash escaping**: `escapeSingleQuotes` uses `s.replace("'", "\\'")` (backslash escape). ClickHouse's standard SQL escaping uses doubled single-quotes (`''`). The backslash form works when `allow_backslash_escaping_in_strings` is enabled (ClickHouse default), but could fail in stricter SQL-compatibility modes.

**`unsafeLabelName` on external metric tag keys** (`MicrometerRegistryScraper.scala`): `unsafeLabelName` calls `Refined.unsafeApply` and throws if the tag key doesn't match `[a-z0-9]+(?:_[a-z0-9]+)*` after dot-to-underscore substitution. Current JVM binder tag keys are all lowercase (area, id, state, level, etc.) so this works today, but adding any binder or library with non-conforming tag keys (e.g. hyphens, uppercase) would cause an unchecked exception in the scrape loop.

**Silent metric loss on retry re-queue** (`ClickHouseMetricsSink.scala`): In `startFlusher`, `notReady` items taken from the bounded queue are put back via `tryOffer`. If the queue is near capacity during an error pause, `tryOffer` silently drops them — items with remaining retries are permanently lost with no log warning.

ryle-goehausen (approved):

PR #1442, opened 2026-02-12
fix: fallback to random peer with retries

scasplte2:


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/cluster/programs/L0PeerDiscovery.scala
```diff
@@ -43,17 +42,37 @@ sealed abstract class L0PeerDiscovery[F[_]: Sync: Random] private (
...
-        getPeersFrom(p)
-          .map(_.filter(peer => lastFacilitators.contains(peer.id)).toSortedSet.toNes)
-          .flatMap(_.traverse_(l0ClusterStorage.setPeers))
-      })
+      .flatMap(_.headOption.flatTraverse(l0ClusterStorage.getPeer))
+      .flatMap {
+        case Some(facilitatorPeer) =>
+          getPeersFrom(facilitatorPeer)
+            .map(_.filter(peer => lastFacilitators.contains(peer.id)))
+            .flatMap(l0ClusterStorage.addPeers)
```
ryle-ai: The old code here did `setPeers` which replaced the entire L0 peer set with only the current facilitators -- acting as implicit pruning. `addPeers` merges without removing stale entries, and `L0ClusterStorage` has no `removePeer` method, so old facilitators that drop out of the active set will persist indefinitely.

This affects `getRandomPeer` which is used by `GlobalL0Service`, `BinaryPoster`, and the `discoverFrom` fallback -- all of which could end up selecting a dead peer.

Was the switch to `addPeers` intentional here, or a side-effect of simplifying the filter/toNes/traverse_ chain? If intentional, it might be worth adding a comment explaining why accumulation is acceptable.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/cluster/programs/L0PeerDiscovery.scala
```diff
@@ -43,17 +42,37 @@ sealed abstract class L0PeerDiscovery[F[_]: Sync: Random] private (
...
+        case None =>
+          logger.warn("No known peers found among last facilitators, falling back to random peer") >>
+            l0ClusterStorage.getPeers
+              .map(_.toNonEmptyList.toList)
+              .flatMap(Random[F].shuffleList)
+              .flatMap(peers => tryDiscoverFromPeers(peers.take(2)))
```
ryle-ai, line 57: In `tryDiscoverFromPeers`, the `getPeersFrom` result is not filtered by `lastFacilitators` before calling `addPeers`. The facilitator-found path (line 49) filters to only facilitator peers, but this fallback path adds **all** ready peers from the contacted node into L0ClusterStorage. This seems intentional (we're in the "no known facilitators" fallback), but it widens the peer set more aggressively than the primary path does.

ryle-ai (commented):
## `setPeers` → `addPeers`: stale peer accumulation

The fix for the unsafe `.head` call and the new fallback path are good improvements. But switching from `setPeers` to `addPeers` in the facilitator-found path (line 50) changes the peer management semantics in a way that deserves attention.

`L0ClusterStorage` has no `removePeer` method. The old `setPeers` call was the **only mechanism** that pruned the L0 peer set -- every 10 seconds, `discover` would replace the stored peers with only the current facilitators. With `addPeers`, peers are merged in and never removed. Over time, the storage accumulates every peer that was ever a facilitator.

This matters because `getRandomPeer` draws from L0ClusterStorage and is used in critical paths: `GlobalL0Service.pullLatestSnapshotFromRandomPeer`, `BinaryPoster` peer selection, and the initial `discoverFrom` fallback. Stale peers in the pool increase the probability of selecting a dead node, which means more retries and latency in snapshot pulling and binary posting.

The fallback path itself (new `tryDiscoverFromPeers`) is a clear improvement over crashing on `.head` with an empty list. The retry-across-peers pattern is reasonable. Just flagging the accumulation concern as something to think about -- it may be fine in practice if facilitator sets are stable, but it's a behavioral change worth being intentional about.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/cluster/programs/L0PeerDiscovery.scala
scasplte2: The facilitator path (happy path) has been reverted to use setPeers, matching the original semantics when we successfully find facilitators, storage is replaced with the current set, preventing stale accumulation.

The fallback path intentionally uses addPeers to widen the peer pool when no facilitators are known in storage. Any peers accumulated here are temporary, the next successful discover call that hits the facilitator path will call setPeers and sweep them out. That said, a follow-up PR to add peer pruning/TTL to L0ClusterStorage would harden this further.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/cluster/programs/L0PeerDiscovery.scala
scasplte2: This is intentional. The fallback fires when none of our stored L0 peers match the current facilitators — so filtering the discovered results by facilitators would likely yield nothing useful and defeat the purpose. The fallback's goal is to bootstrap by widening the pool so future discovery cycles have more peers to work with. The retry across up to 2 distinct peers adds resilience if the first random peer is unresponsive. Once a subsequent discover call succeeds via the facilitator path, setPeers narrows the storage back to just the current facilitators.

marcus-girolneto (approved):

PR #1443, opened 2026-02-13
chore: adding metrics for incremental SC

marcus-girolneto:


scasplte2 (dismissed):

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotConsensusStateAdvancer.scala
```diff
@@ -493,11 +494,16 @@ object GlobalSnapshotConsensusStateAdvancer {
           val totalBytes = binaries.toList.map(_.value.content.length.toLong).sum
           val totalFee = binaries.toList.map(_.value.fee.value.value).sum
 
+          val nowEpochSecond = Instant.now().getEpochSecond.toDouble
```
ryle-ai: `Instant.now()` is a side effect that should be suspended in `F`. The timestamp gets captured at expression-construction time rather than when the effect chain actually executes. In practice the difference is probably negligible here, but it's inconsistent with the rest of the codebase which uniformly uses `Clock[F]`/`Async[F]` for time.

Something like:
```scala
Async[F].realTimeInstant.map(_.getEpochSecond.toDouble).flatMap { nowEpochSecond =>
  Metrics[F].updateGauge("dag_global_snapshot_sc_address_last_activity_epoch", nowEpochSecond, addrTag) >>
    Metrics[F].incrementCounterBy("dag_global_snapshot_sc_address_submissions_count", binariesCount, addrTag)
}
```

This also avoids the `java.time.Instant` import.

ryle-ai (commented):
`Instant.now()` is a side effect evaluated eagerly outside of `F`. The rest of the codebase uses `Clock[F].realTimeInstant` or `Async[F].realTime` for current time — see `ValidationErrorStorage`, `SnapshotProcessor`, `ClickHouseSink`, etc. Since `Async` is already in scope here, this should be straightforward to fix.

PR #1444, opened 2026-02-17
feat: add dag transaction generator tool

scasplte2:


modules/tools/src/main/scala/io/constellationnetwork/tools/TransactionSender.scala
```diff
@@ -0,0 +1,259 @@
...
+                    case (sent, errors, now) =>
+                      val elapsed = (now - startTime).toSeconds.max(1)
+                      val actualTps = sent.toDouble / elapsed
+                      IO.println(
+                        s"  [progress] sent=$sent errors=$errors elapsed=${elapsed}s actual_tps=${f"$actualTps%.1f"}"
+                      )
+                  }
+                }
+
+              def nextRecipient: IO[Address] = iterRef.modify(it => (it, it.next()))
```
ryle-ai: Wrapping a mutable `scala.collection.Iterator` in a `Ref` is an anti-pattern. `Ref.modify` expects a pure function `A => (A, B)`, but `it.next()` mutates the iterator as a side effect. The function returns the same iterator reference it received (`(it, it.next())`), so the `Ref` never actually sees a state change -- it's just being used as a lock around a side-effecting call.

This works today because `sendOne` is called sequentially within each stream phase, so there's no real contention on `iterRef`. But it's fragile: if someone later adds concurrency (e.g. parallel sends), the mutable iterator would silently corrupt.

A `Ref[IO, Int]` counter is simpler and correct:
```scala
Ref.of[IO, Int](0).flatMap { idxRef =>
  def nextRecipient: IO[Address] =
    idxRef.getAndUpdate(_ + 1).map(i => recipients(i % recipients.size))
  ...
}
```
This eliminates the mutable state entirely and would remain correct under concurrency.

ryle-ai (commented):
The Greptile bot flagged several issues (ref advanced before submission, silent fallback to empty ref, unvalidated config values, infinite spin on zero counts) -- all of which appear to be addressed in this version of the code. The ref is now set only on successful submission, `getLastReference` propagates errors via `.adaptError`, config values are validated at startup, and the zero-count guard is in place.

One non-blocking observation on the `Iterator`/`Ref` pattern used for recipient round-robin -- see inline comment.

ottobot-ai:
## Addressed ryle-ai review: mutable Iterator anti-pattern

**Commit:** `8329c630` on `ottobot-ai/tessellation:feat/transaction-sender-fixes`

Replaced the mutable `Iterator` with a pure `Ref[IO, Int]` counter:

```scala
// Before (anti-pattern - mutable state in Ref)
Ref.of[IO, Iterator[Address]](recipientIterator).flatMap { iterRef =>
  def nextRecipient: IO[Address] = iterRef.modify(it => (it, it.next()))

// After (pure - no mutable state)
Ref.of[IO, Int](0).flatMap { recipientIdxRef =>
  def nextRecipient: IO[Address] =
    recipientIdxRef.getAndUpdate(_ + 1).map(i => recipients(i % recipients.size))
```

This eliminates mutable state entirely and is correct under concurrency.

Cherry-pick: `git cherry-pick 8329c630`

modules/tools/src/main/scala/io/constellationnetwork/tools/TransactionSender.scala
```diff
@@ -0,0 +1,258 @@
...
+    val privateKeyInt = new BigInteger(cleanHex, 16)
+    require(
+      privateKeyInt.compareTo(BigInteger.ZERO) > 0 && privateKeyInt.compareTo(curveParams.getN) < 0,
+      "Private key is out of valid range for secp256k1 curve"
+    )
+
+    val kf = KeyFactory.getInstance(secKey.ECDSA, SecurityProvider[F].provider)
+
+    val privateKeySpec = new ECPrivateKeySpec(privateKeyInt, curveParams)
+    val privateKey = kf.generatePrivate(privateKeySpec)
```
ryle-ai, line 257: `Ref.modify` requires a **pure** function `A => (A, B)`. Here, `it.next()` is a side effect that mutates the iterator, and the "new state" returned is the same object reference `it` — so from `Ref`'s perspective, the state never changes.

This is unsafe because `Ref.modify` uses a CAS loop internally. If the CAS retries (possible under concurrency, or even theoretically with `IO` fiber scheduling), the function runs again on the same mutable iterator, calling `it.next()` a second time and skipping a recipient. It works today only because `sendOne` is called sequentially within each stream phase, but the code is one `parEvalMap` away from silently corrupting.

Replace with a pure `Ref[IO, Int]` counter:
```scala
Ref.of[IO, Int](0).flatMap { recipientIdxRef =>
  def nextRecipient: IO[Address] =
    recipientIdxRef.getAndUpdate(_ + 1).map(i => recipients(i % recipients.size))
```
This eliminates the mutable state entirely and is correct under any concurrency model. (I see ottobot-ai already has this fix on a separate branch — worth cherry-picking.)

modules/tools/src/main/scala/io/constellationnetwork/tools/TransactionSender.scala
```diff
@@ -0,0 +1,258 @@
...
+  ): IO[Either[String, String]] = {
+    val uri = Uri.unsafeFromString(s"$l1BaseUrl/transactions")
+    val request = Request[IO](Method.POST, uri).withEntity(signedTx)
+
+    client.run(request).use { response =>
+      response.status match {
+        case Status.Ok =>
+          response
+            .as[Json]
+            .map(json => Right(json.hcursor.downField("hash").as[String].getOrElse("unknown")))
```
ryle-ai, line 234: When `burstTps` is configured as 0 (user wants to skip burst but forgot to also set `burstCount=0`), `config.burstTps.max(1)` silently corrects it to 1 TPS. This is fine defensively, but `steadyIntervalSeconds.max(1)` on the next line means a configured value of 0 seconds between steady transactions gets silently bumped to 1 second — which could be confusing if someone intentionally wants no delay.

Non-blocking — just noting the silent clamping could surprise users. A startup warning or a doc comment in the example config would help.

ryle-ai (dismissed):
The mutable `Iterator` inside a `Ref` is a correctness issue that should be fixed before merge. One other observation inline.

modules/tools/src/main/scala/io/constellationnetwork/tools/TransactionSender.scala
scasplte2: done in previous commit

ottobot-ai:
## Addressed ryle-ai review (round 2)

**Commits on** `ottobot-ai/tessellation:feat/transaction-sender-fixes`:

| Comment | Commit | Fix |
|---------|--------|-----|
| Line 250: mutable Iterator in Ref | `8329c630` | Replaced with pure `Ref[IO, Int]` counter |
| Line 227: silent clamping of intervals | `a9085d37` | Added startup warnings when values are clamped |

**New warnings:**
```
Warning: burstTps=0 clamped to 1 TPS (set burstCount=0 to skip burst phase)
Warning: steadyIntervalSeconds=0 clamped to 1s (set steadyCount=0 to skip steady phase)
```

modules/tools/src/main/scala/io/constellationnetwork/tools/TransactionSender.scala
```diff
@@ -0,0 +1,258 @@
...
+                    source = source,
+                    destination = destination,
+                    amount = TransactionAmount(PosLong.unsafeFrom(config.amountDatum)),
+                    fee = TransactionFee(NonNegLong.unsafeFrom(config.feeDatum)),
+                    parent = parentRef,
+                    salt = salt
+                  )
+                  signedTx <- Signed.forAsyncHasher[IO, Transaction](tx, keyPair)
+                  txHash <- signedTx.value.hash
+                  newRef = TransactionReference(parentRef.ordinal.next, txHash)
```
ryle-ai, line 156: The `run` method takes `Hasher[IO]` and `SecurityProvider[IO]` explicitly but is hardcoded to `IO` throughout. The existing tools module pattern (see `postTransaction`, `getLastReference`, `sendTransactions` in Main.scala) uses `F[_]: Async: Hasher: SecurityProvider` to keep things polymorphic. Not blocking for a standalone CLI tool, but if this code ever gets extracted into a shared utility it'll need refactoring.

ryle-ai (commented):
The major issues from earlier review rounds (ref advanced before submission, silent empty-ref fallback, unvalidated config, mutable Iterator in Ref) all appear to be fixed in the current version. The tool is well-structured and the streaming approach is appropriate.

One observation on the Main.scala restructuring: the `TxSenderCmd` branch correctly avoids initializing JsonSerializer and the JSON hasher, using `Hasher.forKryo` instead — which is required for backward-compatible transaction hashing per the warning in `Transaction.toEncode`. Good call.

Minor note: `TransactionReference.of[F]` already does the `hash → TransactionReference(ordinal, hash)` construction that `sendOne` does manually. Not blocking, but using it would reduce the surface area for ordinal-computation bugs if the `Transaction.ordinal` derivation ever changes.

modules/tools/src/main/scala/io/constellationnetwork/tools/TransactionSender.scala
```diff
@@ -0,0 +1,258 @@
...
+
+    val curveParams = ECNamedCurveTable.getParameterSpec(secKey.secp256k)
+    val privateKeyInt = new BigInteger(cleanHex, 16)
+    require(
+      privateKeyInt.compareTo(BigInteger.ZERO) > 0 && privateKeyInt.compareTo(curveParams.getN) < 0,
+      "Private key is out of valid range for secp256k1 curve"
+    )
+
+    val kf = KeyFactory.getInstance(secKey.ECDSA, SecurityProvider[F].provider)
+
```
ryle-ai, line 255: With infinite cycles enabled, `recipientIdxRef` (an `Int`) will overflow after ~2.1 billion transactions, at which point `i % recipients.size` produces a negative index and `recipients(negative)` throws `IndexOutOfBoundsException`. At 10 TPS that's ~6.8 years so probably not urgent, but a `Long` counter or `Math.floorMod` would make it bulletproof:
```scala
recipientIdxRef.getAndUpdate(_ + 1).map(i => recipients(Math.floorMod(i, recipients.size)))
```

ottobot-ai:
Fixed Int overflow: `9cd9b7b8` — now uses `Math.floorMod(i, recipients.size)` for bulletproof infinite cycling.

Re: polymorphic `F[_]` — agreed it would be cleaner, but leaving as `IO` for now since it's a standalone CLI tool.

ryle-goehausen (approved):

PR #1446, opened 2026-02-18
feat: maintain local view of peer availability

scasplte2:


modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/domain/snapshot/programs/Download.scala
```diff
@@ -452,13 +453,15 @@ object Download {
...
+                      .warn(e)(s"Unable to retrieve snapshot at ordinal ${ordinal.show} from peer ${peer.show}")
+                      .as(none[Hashed[GlobalIncrementalSnapshot]])
                 )
-                .map {
-                  case Some(snapshot) if hash.forall(_ === snapshot.hash) => snapshot.signed.some.asRight[Agg]
-                  case _                                                  => (tail, none[Success]).asLeft[Result]
+                .flatMap {
+                  case Some(snapshot) if hash.forall(_ === snapshot.hash) =>
+                    peerAvailability.recordSuccess(peer).as(snapshot.signed.some.asRight[Agg])
+                  case _ => (tail, none[Success]).asLeft[Result].pure[F]
```
ryle-ai, line 464: When a peer returns a valid snapshot but with a wrong hash, this `case _` fires without calling `recordFailure`. The error path above already records failure via `handleErrorWith`, but a peer that consistently serves stale or mismatched data will accumulate no negative signal and won't be deprioritized.

Same pattern in `currency-l0/Download.scala` and in `fetchGenesis` below (though genesis has no hash check so it's fine there).

Worth considering whether hash mismatch should count as a failure — a peer returning wrong hashes is arguably less useful than one that errors outright, since it still costs a full round-trip.

ryle-ai (commented):
The tiered peer classification approach (Good/Unknown/Bad with probabilistic reprobe) is a solid design. One behavioral gap worth discussing: hash mismatches during download aren't tracked by the availability system.

scasplte2:
pending discussion on if this feature should be further fleshed out or re-aligned with the recent work in #1449 do you have a preference @ryle-goehausen ? Should we close this or flesh it out?

ryle-ai (commented):
Two design-level concerns with the current `PeerAvailability` implementation, relevant to the question of whether to flesh this out or realign with #1449.

The previous inline comment about hash mismatches not calling `recordFailure` is still applicable -- a peer serving wrong hashes accumulates no negative signal.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/PeerAvailability.scala
```diff
@@ -0,0 +1,92 @@
...
+    lastFailure: Option[FiniteDuration]
+  )
+
+  private sealed trait Tier
+  private case object Good extends Tier
+  private case object Unknown extends Tier
+  private case object Bad extends Tier
+
+  def make[F[_]: Async: Random]: F[PeerAvailabilityAlgebra[F]] =
+    Ref.of[F, Map[PeerId, Entry]](Map.empty).map { ref =>
```
ryle-ai, line 33: This `Map[PeerId, Entry]` grows without bound. Entries are added on first interaction but never removed. In a network with peer churn (nodes joining, leaving, rotating IPs), this accumulates dead entries indefinitely. After weeks of uptime with 100+ peers cycling through, you'd have thousands of stale entries that are periodically classified as `Unknown` (due to staleness) and given reprobe chances -- wasting download attempts on peers that no longer exist.

If this feature moves forward, the map needs some form of eviction. Simplest approach: filter out entries in `sortByAvailability` for `PeerId`s not present in the input peer list (since those peers are no longer in the cluster).

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/PeerAvailability.scala
```diff
@@ -0,0 +1,92 @@
...
+            shuffledBad <- Random[F].shuffleList(bad)
+          } yield shuffledGood ++ shuffledUnknown ++ shuffledBad
+
+        private def classifyPeer(now: FiniteDuration, maybeEntry: Option[Entry]): F[Tier] =
+          maybeEntry match {
+            case None => (Unknown: Tier).pure[F]
+            case Some(entry) =>
+              val isStale = entry.lastFailure.exists(lf => (now - lf) > stalenessThreshold)
+
+              val isGood = entry.successCount > 0 &&
```
ryle-ai, line 79: The `successCount` and `failureCount` fields accumulate but are never meaningfully used -- `successCount > 0` is true after the first success and stays true forever. The actual classification is purely **last-event-wins**: whichever of `lastSuccess` or `lastFailure` is more recent determines the tier.

This means:
- A peer with 100 failures then 1 success → `Good`
- A peer with 100 successes then 1 failure → `Bad`
- A peer that succeeded once an hour ago and hasn't been seen since → `Good` forever (no decay)

If the intent is to track availability over time, a ratio or sliding window would be more robust. If the intent is purely recency-based, the counts can be removed to simplify the model and avoid giving the impression they influence behavior.

PR #1447, opened 2026-02-18
chore: delegated stake test script

effects-ai:


ryle-ai (commented):
This test script duplicates most of the functionality already in `.github/action_scripts/delegated_staking/lib.js` and `delegated-staking.js`, but imports from a nonexistent `./lib.cjs`. A few concrete issues below.

Broadly: the existing CI test suite at `.github/action_scripts/delegated_staking/` already covers the full delegated staking lifecycle (create node params, create token lock, create/update/increase/withdraw delegated stake) with proper retry logic (`withRetry`), assertion helpers, and structured logging. This script reimplements most of that with custom polling loops and no shared utilities. If the intent is a standalone CLI tool for manual testing against live networks, that's fine, but it should be self-contained and actually runnable.

test-scripts/delegated_stake/delegated_stake.js
```diff
@@ -0,0 +1,998 @@
...
+    fetchSnapshot,
+    assertRewardTxnInSnapshot,
+    assertTokenUnlockInSnapshot,
+  } = require('./lib.cjs');
+    
+
+// Initialize elliptic curve for key generation
+const ec = new EC.ec('secp256k1');
+
+/**
```
ryle-ai, line 29: **`./lib.cjs` does not exist in the repository.** The functions imported here (`checkOk`, `postNodeParamsNodeId`, `getAccountDelegatedStakes`, `createTokenLock`, etc.) live in `.github/action_scripts/delegated_staking/lib.js` (CommonJS, but `.js` extension). This script will fail immediately at startup with `MODULE_NOT_FOUND`.

If this is meant to be a standalone CLI tool, either fix the import path or inline the needed functions. If it's meant to work alongside the CI scripts, it should import from the correct location.

test-scripts/delegated_stake/delegated_stake.js
```diff
@@ -0,0 +1,998 @@
...
+    
+    if (response && response.status === 200) {
+        console.log(`✅ Node parameters created successfully!`);
+        console.log(`Response status: ${response}`);
+    } else {
+        // console.log(`Response: ${JSON.stringify(response, null, 2)}`);
+        console.log(`⚠️  Node parameters response: ${response.statusText || 'unknown'}`);
+    }
+    
+    return {
```
ryle-ai, line 150: `${response}` will stringify the axios response object to `[object Object]`. Should be `${response.status}` to match the log label.

test-scripts/delegated_stake/delegated_stake.js
```diff
@@ -0,0 +1,998 @@
...
+            networkVersion: '2.0',
+            testnet: isTestnet,
+            l0Url: `http://${options.ip}:9000`,
+            l1Url: `http://${options.ip}:9010`,
+        });
+    } else {
+        dag4.account.connect({
+            networkVersion: '2.0',
+            testnet: isTestnet,
+        });
```
ryle-ai, line 196: Port inconsistency: when `options.ip` is provided, L1 uses port `9010`, but the standard DAG L1 public port is `9100` (confirmed in `euclid-development-environment/infra/ansible/local/playbooks/vars.ansible.yml` and `.github/action_scripts/shared/network.js`). Port `9010` is used in the Docker test script (`delegated_stake_docker_test.sh`) which uses a different container port mapping -- if this CLI tool is meant for non-Docker networks, `9100` is correct.

test-scripts/delegated_stake/delegated_stake.js
```diff
@@ -0,0 +1,998 @@
...
+            source: sourceAddress,
+            amount: lockAmountDatum,
+            tokenL1Url: urls.dagL1Url,
+            unlockEpoch: options.unlockEpoch || null,
+            currencyId: null,
+            replaceTokenLockRef: options.replaceTokenLockRef || null,
+            fee: options.fee || 0,
+        });
+        console.log(`Token lock: ${JSON.stringify(result, null, 2)}`);
+
```
ryle-ai, line 253: `options.unlockEpoch || null` will coerce `0` to `null`. Use `options.unlockEpoch ?? null` (nullish coalescing). Same applies to `options.replaceTokenLockRef || null` on line 258 and `options.fee || 0` on line 259, though those are less likely to hit in practice.

test-scripts/delegated_stake/delegated_stake.js
```diff
@@ -0,0 +1,998 @@
...
+    const rewardFraction = args.rewardFraction;
+
+    if (!privateKeyHex) {
+        console.log('Usage: node ds.js --key <privateKeyHex> [--amount <amount>] [--network <network>] [--ip <ip>]');
+        console.log('Options:');
+        console.log('  --key, --privateKey    Private key in hex format (required)');
+        console.log('  --network              Network: testnet, integrationnet, mainnet (default: testnet)');
+        console.log('  --ip                   Custom node IP address (optional)');
+        console.log('  --command              Command: createTokenLock, replaceTokenLock, listTokenLocks, createDelegatedStake, withdrawDelegatedStake, smokeTest (optional)');
+        console.log('  --amount               Amount in DAG (optional)');
```
ryle-ai, line 457: Usage text says `node ds.js` but the file is `delegated_stake.js`. Same on lines 473-475.

test-scripts/delegated_stake/delegated_stake.js
```diff
@@ -0,0 +1,998 @@
...
+            if (amountArg === 'max') {
+                // Get current balance
+                const balance = await dag4.network.getAddressBalance(address);
+                smokeTestAmount = balance.balance / 100000000;
+                console.log(`Using max balance: ${smokeTestAmount} DAG`);
+            } else {
+                smokeTestAmount = parseFloat(amountArg);
+                if (isNaN(smokeTestAmount)) {
+                    console.log('--amount must be a number or "max"');
+                    process.exit(1);
```
ryle-ai, line 566: When `amount` is `'max'`, the entire account balance is used. But the smoke test then tries to create a 5000 DAG token lock (line 871) followed by a replacement lock for the full amount (line 930). The replacement lock calls `dag4.network.getAddressBalance` to get the *current* balance after the initial lock, then uses `Math.min(amount, currentBalanceDag)`. Since `amount` here is the *original* full balance (before the 5000 DAG lock), and `currentBalanceDag` is the balance *after* locking 5000, the replacement amount will be `currentBalanceDag` (= original - 5000). This means the replacement lock is actually *smaller* than the original balance. Is that the intended behavior? The variable name `amount` is overloaded between "total test budget" and "replacement lock amount" which makes the flow hard to follow.

test-scripts/delegated_stake/delegated_stake.js
marcus-constellation-ai, line 29: `./lib.cjs` does not exist. The shared library is at `.github/action_scripts/delegated_staking/lib.js` (CommonJS `module.exports`). The correct path from `test-scripts/` would be `../.github/action_scripts/delegated_staking/lib.js`, and `createRequire` can require `.js` files with CommonJS format directly — the `.cjs` extension is not needed. As-is, the script crashes immediately with `MODULE_NOT_FOUND`.

test-scripts/delegated_stake/delegated_stake.js
```diff
@@ -0,0 +1,998 @@
...
+    for (let attempt = 1; attempt <= maxAttempts; attempt++) {
+        const tokenLocks = await getActiveTokenLocks(account, urls);
+        for (const lock of tokenLocks) {
+            const hash = await computeHash(lock);
+            if (hash === expectedHash) {
+                console.log(`✅ Token lock found after ${attempt} attempt(s)`);
+                return lock;
+            }
+        }
+        console.log(`   Attempt ${attempt}/${maxAttempts} - not found yet...`);
```
marcus-constellation-ai, line 682: The hash comparison is wrong. `computeHash(lock)` runs brotliCompress+sha256 on the raw API response object, but `expectedHash` is the transaction hash returned by `postTokenLock`. The `/token-locks/${address}` API already returns a `hash` field on each lock object (this is how `waitForTokenLockInclusion` in `lib.js` works: `response.data.find(l => l.hash === lockHash)`). Replace this with `lock.hash === expectedHash`.

test-scripts/delegated_stake/delegated_stake.js
```diff
@@ -0,0 +1,998 @@
...
+async function main() {
+    // Get parameters from named command line arguments
+    const args = parseArgs();
+    const privateKeyHex = args.key || args.privateKey;
+    const amount = args.amount ? parseFloat(args.amount) : undefined;
+    const tokenLockHash = args.tokenLockHash;
+    const nodeId = args.nodeId;
+    const stakeHash = args.stakeHash;
+    const network = args.network || 'testnet';
+    const ip = args.ip;
```
marcus-constellation-ai, line 443: Passing private keys as command-line arguments exposes them in the process table (`ps aux`) and shell history. Consider reading the key from a file path argument or an environment variable instead.

test-scripts/delegated_stake/delegated_stake.js
```diff
@@ -0,0 +1,998 @@
...
+    const privateKeyBuffer = Buffer.from(privateKeyHex, 'hex');
+  
+    try {
+        const privateKeyString = privateKeyBuffer.toString('hex');
+        const keyPair = ec.keyFromPrivate(privateKeyBuffer);
+    
+        const uncompressedPublicKey = keyPair.getPublic(false, 'hex'); // Uncompressed format
+        const nodeId = uncompressedPublicKey.slice(2); // Remove the '0x04' prefix
+    
+        const account = dag4.createAccount(privateKeyString);
```
marcus-constellation-ai, line 108: `Buffer.from(privateKeyHex, 'hex').toString('hex')` is identical to `privateKeyHex`. The roundtrip serves no purpose; just use `privateKeyHex` directly as `privateKeyString`.

test-scripts/delegated_stake/delegated_stake.js
```diff
@@ -0,0 +1,998 @@
...
+    checkOk,
+    checkBadRequest,
+    dagToDatum,
+    getPrivateKeyAndNodeIdFromFile,
+    postNodeParamsNodeId,
+    // createDelegatedStake,
+    // withdrawDelegatedStake,
+    getAccountDelegatedStakes,
+    assertDelegatedStakes,
+    fetchStakeWithRewardsBalance,
```
marcus-constellation-ai, line 16: Commented-out imports (`createDelegatedStake`, `withdrawDelegatedStake`) should be removed. The script defines its own versions of these functions locally; leaving them commented out creates confusion about what is actually used.

test-scripts/delegated_stake.js
```diff
@@ -0,0 +1,998 @@
...
+                console.log('No active token locks found.');
+            }
+
+/*
+            // Download combined snapshot from API
+            const combinedSnapshot = await downloadCombinedSnapshot(urls);
+            const tokenLocks = await getActiveTokenLocksTransactions(combinedSnapshot, address);
+            
+            if (tokenLocks && tokenLocks.length > 0) {
+                for (let index = 0; index < tokenLocks.length; index++) {
```
marcus-constellation-ai: Large commented-out block (lines 531–553) should be removed before merge. Dead code inside a new file signals unfinished work.

marcus-constellation-ai (changes requested):
The script is not runnable as written due to a broken import path, and has a second correctness issue with hash matching in `pollForTokenLock`. The ryle-ai reviewer's high-level observation about duplication with `.github/action_scripts/delegated_staking/` is also accurate — the existing library has ordinal-aware retry helpers (`waitForStakeInclusion`, `waitForStakeWithdrawal`, `waitForTokenLockInclusion`) that the new polling loops don't use.

**Blocking:**
- `require('./lib.cjs')` references a file that does not exist anywhere in the repository. The actual shared library is `.github/action_scripts/delegated_staking/lib.js`. The script will crash immediately on startup before executing any logic.

**Significant correctness issue:**
- `pollForTokenLock` computes the hash of the full API response object via `computeHash(lock)` and compares it to `expectedHash` (which is the transaction hash returned by `postTokenLock`). The existing `waitForTokenLockInclusion` in `lib.js` shows the API already returns a `hash` field on each lock object — the correct comparison is `lock.hash === expectedHash`, not a client-side brotli+sha256 of the object.

**Other observations:**
- Private key is accepted directly from CLI argument (`--key`), making it visible in the process list and shell history on shared systems.
- `extractKeysAndAccount` converts `privateKeyHex → Buffer → hex` and stores the result as `privateKeyString`, but this is a no-op: the result equals the original `privateKeyHex`.
- Commented-out import lines and a ~22-line dead code block remain in the file.

scasplte2 (approved):

PR #1448, opened 2026-02-18
feat: migrating GlobalSnapshotInfo to use MptStore

marcus-girolneto:


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/collateral/MptStoreCollateral.scala
```diff
@@ -0,0 +1,29 @@
...
+    config: CollateralConfig,
+    mptStore: MptStore[F, GlobalStateKey]
+  ): Collateral[F] =
+    new Collateral[F] {
+
+      def hasCollateral(peerId: PeerId): F[Boolean] =
+        peerId.value.toPublicKey
+          .map(_.toAddress)
+          .flatMap(mptStore.getBalance)
+          .map(_.forall(_.satisfiesCollateral(config.amount)))
```
ryle-ai, line 30: `.forall` on `Option[Balance]` returns `true` when the balance is `None` (address not found in MptStore). This matches the old `Collateral.make` behavior which used `.getOrElse(true)` -- meaning nodes with no balance entry are treated as having sufficient collateral. Just confirming: is this the intended behavior? The old code had the same semantics, but it's worth being explicit about since "unknown address = has collateral" is a non-obvious default.

ryle-ai (commented):
## MptStore Migration Review

This is a large, systematic migration that moves state reads from in-memory `GlobalSnapshotInfo` fields to `MptStore` (Merkle Patricia Trie). The architectural direction is sound and the migration is generally well-executed. A few observations below.

### StateChannelEventsProcessor fee deduction complexity

The restructured fee deduction logic in `GlobalSnapshotStateChannelEventsProcessor` (the `tailRecM` loop) is significantly harder to follow than the original. The nested `fold` + `flatMap` composition that replaced a straightforward `map` block is correct but dense. The original had:

```scala
val balance = balanceUpdate.get(feeAddress)
  .orElse(lastGlobalSnapshotInfo.balances.get(feeAddress))
  .getOrElse(Balance.empty)
```

The new version:
```scala
val localBalance = balanceUpdate.get(feeAddress)
localBalance.fold(mptStore.getBalance(feeAddress).map(_.getOrElse(Balance.empty)))(_.pure[F]).map { ... }
```

These are semantically equivalent (MptStore is synced from the same GlobalSnapshotInfo before processing), but the new version is noticeably harder to reason about. Not blocking, but worth noting for future maintainers.

### Partial migration in AddressService

`getBalance` is migrated to MptStore but `getTotalSupply`, `getWalletCount`, `getFilteredOutTotalSupply`, `getCirculatedSupply` etc. still read from `state.balances` directly. This could produce inconsistencies if MptStore and the in-memory snapshot ever diverge (e.g. during the transition period). Worth tracking as a follow-up.

### Validator duplication

Both `UpdateDelegatedStakeValidator` and `UpdateNodeCollateralValidator` now have two near-identical implementations (MptStore-backed + fallback). This is a lot of duplicated validation logic (~200+ lines each). If the fallback is only needed temporarily during the migration, consider a timeline for removing it to avoid maintaining parallel validation paths long-term.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/delegatedStake/UpdateDelegatedStakeValidator.scala
```diff
@@ -44,6 +46,233 @@ object UpdateDelegatedStakeValidator {
...
+          signed.validNec
+        } else {
+          UnauthorizedNode(signed.nodeId).invalidNec
+        }
+
+      private def validateNodeId(
+        signed: Signed[UpdateDelegatedStake.Create],
+        lastContext: GlobalSnapshotInfo
+      ): F[UpdateDelegatedStakeValidationErrorOr[Signed[UpdateDelegatedStake.Create]]] = {
+        val nodeIdParams = lastContext.updateNodeParameters
```
ryle-ai, line 154: `validateParent` sorts by `.createdAt` and takes `.lastOption`. The fallback implementation (line ~107 in the original) uses `maxBy(_.createdAt)`. These are logically equivalent for finding the latest record, but note that the `NodeCollateralValidator` equivalent sorts by `.event.ordinal` instead (matching its own fallback). If `createdAt` and `event.ordinal` don't always have the same ordering, this distinction matters. Worth a sanity check that `createdAt` is the correct sort key here and not `.event.ordinal` as used in the routes (`DelegatedStakesRoutes.getLastReference` sorts by `.event.ordinal`).

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/delegatedStake/UpdateDelegatedStakeValidator.scala
```diff
@@ -44,6 +46,233 @@ object UpdateDelegatedStakeValidator {
...
+          }
+        }
+      }
+
+      private def validatePendingWithdrawal(
+        signed: Signed[UpdateDelegatedStake.Create]
+      ): F[UpdateDelegatedStakeValidationErrorOr[Signed[UpdateDelegatedStake.Create]]] =
+        mptStore.getDelegatedStakeWithdrawals(signed.source).map { maybeWithdrawals =>
+          val withdrawalRef = maybeWithdrawals
+            .getOrElse(SortedSet.empty[PendingDelegatedStakeWithdrawal])
```
ryle-ai, line 178: `validateNodeId` reads `activeDelegatedStakes` from `mptStore` but still reads `updateNodeParameters` from `lastContext` (line 172-174). This is correct since `updateNodeParameters` is not stored in MptStore, but the mixed data source pattern is subtle. A brief comment explaining why `lastContext` is still needed here would help future readers.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/global/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -201,105 +207,120 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
-
-                            maybeBalanceUpdate match {
-                              case Some(newBalanceUpdate) =>
-                                ((nel.prepend((head, (snapshot, state).asRight.some)), newBalanceUpdate).some, tail).asLeft
-                              case None if !isFeeRequired =>
-                                ((nel.prepend((head, (snapshot, state).asRight.some)), balanceUpdate).some, tail).asLeft[Result]
-                              case None => // balance update unsuccessful or impossible? we can't accept any more snapshots
-                                current.asRight
+                              (nel.prepend((head, snapshot.asLeft.some)), balanceUpdate).some,
+                              tail
```
ryle-ai: The initial state construction changed from:
```scala
lastGlobalSnapshotInfo.lastCurrencySnapshots.get(address)
  .map(init => (stubBinary, init.some))
```
to querying `getIncrementalCurrencySnapshot` and `getCurrencySnapshotInfo` separately from MptStore. This works because `convertCurrencySnapshots` in `GlobalStateConverter` stores both entries for `Left(fullSnapshot)` cases (converting full -> incremental + info). However, the new code always constructs a `Right(incremental, info)` -- it can never produce the `Left(fullSnapshot)` case that the old code could. This means the `tailRecM` branch handling `Some(Left(fullSnapshot))` (lines ~284-297) will never be hit from the initial state. This is functionally fine because the "first incremental after full snapshot" path still works, but it's a subtle semantic change worth being aware of.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/global/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -201,105 +207,120 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
-                            ) >> Async[F].pure(current.asRight)
-                          }
-                        case None => // again we only let it through if fee is not required
-                          if (isFeeRequired)
-                            Async[F].pure(current.asRight) // was: none.asRight but why clean it out rather than using current state?
-                          else ((nel.prepend((head, lastCurrState)), balanceUpdate).some, tail).asLeft.pure[F]
-                      }
+                          case None => // again we only let it through if fee is not required
+                            if (isFeeRequired)
+                              Async[F].pure(current.asRight) // was: none.asRight but why clean it out rather than using current state?
```
ryle-ai: The fee deduction path changed from `.map { state => ... }` to `.flatMap { state => ... }`. This is necessary because the balance lookup is now async (`mptStore.getBalance`). The restructured logic with `maybeFeeAddress.fold(...)` is correct but significantly harder to follow than the original pattern-match approach. Consider adding a comment explaining the three cases:
1. Fee not required -> accept unconditionally
2. Fee required, no fee address -> reject
3. Fee required with fee address -> check balance in local accumulator first, then MptStore

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/collateral/MptStoreCollateral.scala
marcus-girolneto: This is intentional — matches the old Collateral.make behavior (getOrElse(true)). The semantics are: if an address has no balance entry in MptStore, it's treated as having sufficient collateral. This is the pre-existing behavior, not a regression

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/delegatedStake/UpdateDelegatedStakeValidator.scala
marcus-girolneto: This is the previous logic, so I think it's fine

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/global/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -201,105 +207,120 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
+                                    current.asRight[Agg].pure[F]
+                                ) { feeAddress =>
+                                  val localBalance = balanceUpdate.get(feeAddress)
+                                  localBalance.fold(mptStore.getBalance(feeAddress).map(_.getOrElse(Balance.empty)))(_.pure[F]).map {
+                                    balance =>
+                                      (balance.minus(head.fee).toOption.map(uBalance => balanceUpdate + (feeAddress -> uBalance)) match {
+                                        case Some(newBalanceUpdate) =>
+                                          ((nel.prepend((head, (snapshot, state).asRight.some)), newBalanceUpdate).some, tail)
+                                            .asLeft[Result]
+                                        case None if !isFeeRequired =>
```
ryle-ai: Dead code: this `case None if !isFeeRequired =>` branch is unreachable. We're inside the `fold`'s `Some(feeAddress)` handler, which only executes when `maybeFeeAddress.filter(_ => isFeeRequired)` is `Some` — meaning `isFeeRequired` is always `true` here. The guard `!isFeeRequired` can never be true.

Not a bug (falls through to `case None =>` which correctly rejects), but worth removing to reduce confusion. The old code had this as part of a single match over `maybeBalanceUpdate` where both the "fee not required" and "fee required but insufficient" cases could land on `None`. After the refactor split them into separate fold branches, this guard became vestigial.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/global/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -201,105 +207,120 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
+                              maybeFeeAddress
+                                .filter(_ => isFeeRequired)
+                                .fold(
+                                  if (!isFeeRequired)
+                                    ((nel.prepend((head, (snapshot, state).asRight.some)), balanceUpdate).some, tail).asLeft[Result].pure[F]
+                                  else
+                                    current.asRight[Agg].pure[F]
+                                ) { feeAddress =>
+                                  val localBalance = balanceUpdate.get(feeAddress)
+                                  localBalance.fold(mptStore.getBalance(feeAddress).map(_.getOrElse(Balance.empty)))(_.pure[F]).map {
```
ryle-ai: This line is dense. `localBalance.fold(mptStore.getBalance(feeAddress).map(_.getOrElse(Balance.empty)))(_.pure[F])` — the first argument to `fold` is the "None" case (fetch from MptStore), the second is the "Some" case (use local accumulator). Semantically equivalent to the old `balanceUpdate.get(feeAddress).orElse(lastGlobalSnapshotInfo.balances.get(feeAddress)).getOrElse(Balance.empty)`, but harder to parse at a glance. A named helper or brief comment would help future readers.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/global/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -201,105 +207,120 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
-                    case Some(snapshot) => // full snapshot - we don't subtract fee
-                      (
-                        (NonEmptyList.one((head, snapshot.asLeft.some)), emptyBalanceUpdate).some,
-                        tail
-                      ).asLeft
-                    case None => // no full snapshot yet - we only accept the binary if fee is not required
-                      if (isFeeRequired) none.asRight
-                      else ((NonEmptyList.one((head, none)), emptyBalanceUpdate).some, tail).asLeft
-                  }
+            val initialStateF: F[Option[MetagraphAcceptanceResult]] = for {
```
ryle-ai: The old code handled both `Left(fullSnapshot)` and `Right(incremental, info)` in the initial state. The new code always constructs `Right(incremental, info)` because `convertCurrencySnapshots` in `GlobalStateConverter` converts full snapshots to incremental form during sync. This means the `case (_, Some(Left(fullSnapshot)))` branch at line ~254 will never fire from the initial state — only from a deserialized binary later in the loop. This is functionally correct (verified that `syncFromGlobalSnapshotInfo` always stores both `LastIncrementalCurrencySnapshots` and `LastCurrencySnapshotInfo` for both Left and Right cases), but it's a subtle semantic change worth being aware of.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/nodeCollateral/UpdateNodeCollateralValidator.scala
```diff
@@ -58,6 +60,220 @@ object UpdateNodeCollateralValidator {
...
+
+        def tokenLockValid(address: Address): F[Boolean] =
+          mptStore.getActiveTokenLocks(address).flatMap { maybeTokenLocks =>
+            val tokenLocks = maybeTokenLocks.getOrElse(SortedSet.empty[Signed[TokenLock]])
+            for {
+              tokenLocksWithReferences <- tokenLocks.toList.traverse(t => TokenLockReference.of(t).map(r => (t, r)))
+            } yield
+              tokenLocksWithReferences.find { case (_, r) => r.hash === signed.tokenLockRef } match {
+                case Some((tokenLock, _)) =>
+                  signed.amount.value.value === tokenLock.amount.value.value && tokenLock.unlockEpoch.isEmpty
```
ryle-ai: Pre-existing inconsistency: `UpdateDelegatedStakeValidator.tokenLockValid` checks `address === tokenLock.source` (verifying the token lock belongs to the signing address), but this method does not. The fallback version has the same gap, so this isn't a regression — but since you're writing a fresh MptStore-backed implementation anyway, this would be a good time to add `address === tokenLock.source` here for parity. Without it, the validation relies on the MptStore key scoping (token locks are stored per-address), but the returned `Signed[TokenLock]` objects contain their own `.source` field which could theoretically differ.

ryle-ai (commented):
### Fee deduction dead code in StateChannelEventsProcessor

The refactored fee deduction logic in `GlobalSnapshotStateChannelEventsProcessor.processCurrencySnapshots` has a `case None if !isFeeRequired =>` branch (line 296) inside the `fold`'s `Some(feeAddress)` handler. This branch is **unreachable**: we only enter the `Some(feeAddress)` branch when `maybeFeeAddress.filter(_ => isFeeRequired)` is `Some`, which means `isFeeRequired` is `true`. So `!isFeeRequired` is always `false` inside this branch. The code is correct (it falls through to `case None =>` which rejects), but the dead branch makes the logic harder to reason about. Looks like it was carried over mechanically from the old flat pattern-match where all three cases (fee not required, no address, insufficient balance) were handled together.

### Validator source address check inconsistency

The `UpdateDelegatedStakeValidator.tokenLockValid` checks `address === tokenLock.source` but `UpdateNodeCollateralValidator.tokenLockValid` does not. This is a **pre-existing inconsistency** (the fallback implementations have the same gap), so it's not introduced by this PR. But since this migration is a good opportunity to audit and align the two validators, it's worth considering whether node collateral should also verify token lock ownership. The MptStore reads token locks by address key, which provides some implicit scoping, but the data returned is the full `Signed[TokenLock]` which could theoretically have a different `.source`.

ryle-ai (commented):
## MptStore Migration Review

This is a well-structured, systematic migration of GlobalSnapshotInfo reads to MptStore. The approach is sound — each component that previously read from in-memory SortedMap fields now reads from the Merkle Patricia Trie store, with fallback patterns where appropriate. Tests are thoroughly updated.

A few things worth flagging:

### Behavioral change in staking balance for full-snapshot metagraphs

The old `fetchStakingBalance` function explicitly returns `Balance.empty` for `Left(fullSnapshot)` cases (i.e., metagraphs that only have a full currency snapshot, no incremental yet). The new MptStore-based code queries `getCurrencySnapshotInfo(address)` which, based on `convertCurrencySnapshots` in GlobalStateConverter, stores `CurrencySnapshotInfo` for *both* Left and Right cases. This means a metagraph that previously had its staking balance treated as zero will now potentially have a real staking balance resolved.

This affects fee calculations in `SnapshotBinaryFeeCalculator`, `StateChannelService`, and `GlobalSnapshotStateChannelEventsProcessor.buildSnapshotFeesInfo`. It may be intentional (the old behavior looks like a simplification that the MptStore migration corrects), but it's a subtle change that could affect fee thresholds for metagraphs in the full-snapshot-only state. Worth confirming this is the desired behavior.

### Mixed data sources in event processing

`getFeeAddresses` (used at the top of `process()` in GlobalSnapshotStateChannelEventsProcessor) still reads directly from `lastGlobalSnapshotInfo.lastCurrencySnapshots`, while `buildSnapshotFeesInfo` now reads staking balance from MptStore. Within the same processing flow, fee address lookups come from the old data source while staking balances come from the new one. This is likely fine during the transition, but worth noting as a follow-up migration item.

### Dead code in fee deduction logic

In `processCurrencySnapshots`, the `case None if !isFeeRequired =>` branch inside the `fold`'s `Some(feeAddress)` handler (the handler that only runs when `isFeeRequired` is true) is unreachable. The `maybeFeeAddress.filter(_ => isFeeRequired).fold(...)` ensures the `Some(feeAddress)` branch only executes when `isFeeRequired == true`, making the `!isFeeRequired` guard dead code. This was already flagged by another reviewer — just confirming it's a real issue, not a blocker.

### initialState now only handles incremental snapshots

The old `initialState` in `processCurrencySnapshots` read from `lastGlobalSnapshotInfo.lastCurrencySnapshots.get(address)` which returns the full Either (Left for full snapshots, Right for incrementals). The new code only queries `getIncrementalCurrencySnapshot` and `getCurrencySnapshotInfo`, and returns `None` unless both are present. This means if a metagraph's last known state is a full snapshot (Left case), the initial state will be `None` and processing will start from scratch. This is a change in behavior — previously the full snapshot was carried forward as initial state. I'd like to understand if this is intentional.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/global/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -201,105 +207,120 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
+              maybeIncremental <- mptStore.getIncrementalCurrencySnapshot(address)
+              maybeInfo <- mptStore.getCurrencySnapshotInfo(address)
+            } yield
+              (maybeIncremental, maybeInfo) match {
+                case (Some(incremental), Some(info)) =>
+                  val init: Either[Signed[CurrencySnapshot], (Signed[CurrencyIncrementalSnapshot], CurrencySnapshotInfo)] =
+                    (incremental, info).asRight
+                  Some((NonEmptyList.one((stubBinary, init.some)), Map.empty[Address, Balance]))
+                case _ => None
+              }
```
ryle-ai: The old `initialState` read from `lastGlobalSnapshotInfo.lastCurrencySnapshots.get(address)` which returned `Either[Signed[CurrencySnapshot], (Signed[CurrencyIncrementalSnapshot], CurrencySnapshotInfo)]` — handling both full and incremental snapshots. The new code only queries `getIncrementalCurrencySnapshot` and `getCurrencySnapshotInfo`, returning `None` unless both are present.

This means metagraphs whose last known state is a full snapshot (`Left` case) will now get `initialState = None` instead of `Some(...)`, and processing will fall through to the `case (None, head :: tail)` branch which tries to deserialize a full snapshot from the binary. This should be fine if the next binary in the chain is indeed a full snapshot, but it loses the previously-carried-forward full snapshot context.

Is this the intended behavior? The old code would seed the initial state with the full snapshot, meaning the next binary only needed to be the first incremental. With this change, if there's no incremental in MptStore, the processor starts from scratch.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/global/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -201,105 +207,120 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
+                            ).flatMap { state =>
+                              val maybeFeeAddress = state.lastMessages.flatMap(_.get(MessageType.Owner)).map(_.address)
+
+                              maybeFeeAddress
+                                .filter(_ => isFeeRequired)
+                                .fold(
+                                  if (!isFeeRequired)
+                                    ((nel.prepend((head, (snapshot, state).asRight.some)), balanceUpdate).some, tail).asLeft[Result].pure[F]
+                                  else
+                                    current.asRight[Agg].pure[F]
```
ryle-ai: `case None if !isFeeRequired =>` is unreachable here. This branch is inside the `Some(feeAddress)` handler of `maybeFeeAddress.filter(_ => isFeeRequired).fold(...)`, which only executes when `isFeeRequired == true`. The `!isFeeRequired` guard can never be true in this context.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/domain/snapshot/programs/SnapshotBinaryFeeCalculator.scala
```diff
@@ -19,19 +22,29 @@ trait SnapshotBinaryFeeCalculator[F[_]] {
...
-        .pure[F]
-        .map(fetchStakingBalance(_, info))
-        .flatMap { balance =>
+  def make[F[_]: Async](
+    feeCalculator: FeeCalculator[F],
+    mptStore: MptStore[F, GlobalStateKey]
+  ): SnapshotBinaryFeeCalculator[F] =
+    (event: StateChannelEvent, _: GlobalSnapshotInfo, ordinal: SnapshotOrdinal) =>
+      for {
+        maybeCurrencyInfo <- mptStore.getCurrencySnapshotInfo(event.value.address)
```
ryle-ai, line 42: Minor: the old `fetchStakingBalance` returned `Balance.empty` for metagraphs with only a full snapshot (`Left` case in `lastCurrencySnapshots`). The new code queries `getCurrencySnapshotInfo` which stores info for both full and incremental cases. This could yield a non-empty staking balance where the old code returned zero, subtly affecting fee calculations for metagraphs in the full-snapshot-only state.

If this is intentional (correcting the old simplification), no action needed — just flagging the behavioral difference.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/domain/statechannel/StateChannelService.scala
```diff
@@ -34,12 +39,20 @@ object StateChannelService {
         globalSnapshotAndState: (Signed[GlobalIncrementalSnapshot], GlobalSnapshotInfo)
       )(implicit hasher: Hasher[F]): F[Either[NonEmptyList[StateChannelValidationError], Unit]] = {
         val (snapshot, state) = globalSnapshotAndState
```
ryle-ai, line 41: `getFeeAddresses(state)` still reads from `state.lastCurrencySnapshots` directly (not MptStore), while the staking balance on line 49 now comes from MptStore. Within the same validation flow, fee address data and staking balance data come from different sources. Worth noting as a follow-up migration item.

ryle-ai (commented):
### Behavioral change: staking balance for full-snapshot metagraphs

The old `fetchStakingBalance` explicitly returned `Balance.empty` for `Left(fullSnapshot)` cases in `lastCurrencySnapshots`. The new MptStore-based code queries `getCurrencySnapshotInfo(address)` which returns `Some(info)` for *both* Left and Right cases, because `convertCurrencySnapshots` in `GlobalStateConverter` converts full snapshots to incremental form and stores `CurrencySnapshotInfo` for both during sync.

This means metagraphs whose last known state is a full snapshot will now potentially have a real staking balance resolved, where the old code always returned zero. This affects fee calculations in `SnapshotBinaryFeeCalculator`, `StateChannelService`, and `GlobalSnapshotStateChannelEventsProcessor.buildSnapshotFeesInfo`.

If this is intentional (correcting an old simplification), no action needed. But if fee thresholds were calibrated around the old behavior, this could subtly change which metagraph binaries pass fee validation. Worth confirming.

### Security improvement: `address === tokenLock.source` added to NodeCollateral

The PR adds the `address === tokenLock.source` check to `UpdateNodeCollateralValidator.tokenLockValid` in both the MptStore-backed and fallback implementations. This was a pre-existing inconsistency (DelegatedStake had it, NodeCollateral didn't). Good fix to include here.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/domain/snapshot/programs/SnapshotBinaryFeeCalculator.scala
```diff
@@ -9,29 +9,47 @@ import scala.collection.immutable.SortedMap
...
-        .pure[F]
-        .map(fetchStakingBalance(_, info))
-        .flatMap { balance =>
+  // Staking balance note: for metagraphs with only a full snapshot (Left case in the old
+  // lastCurrencySnapshots), getCurrencySnapshotInfo returns CurrencySnapshotInfo with
+  // lastMessages = None (via toCurrencySnapshotInfo), so fetchStakingAddress returns None
+  // and we get Balance.empty — same as old fetchStakingBalance which returned Balance.empty
+  // for Left(fullSnapshot).
+  def make[F[_]: Async](
+    feeCalculator: FeeCalculator[F],
```
ryle-ai, line 37: The old `fetchStakingBalance` returned `Balance.empty` for metagraphs with only a full snapshot (`Left` case in `lastCurrencySnapshots`). The new code queries `getCurrencySnapshotInfo` which stores info for both full and incremental cases (see `convertCurrencySnapshots` in `GlobalStateConverter` -- it calls `fullSnapshot.info.toCurrencySnapshotInfo` for Left cases). This means a metagraph in the full-snapshot-only state could now yield a non-empty staking balance, subtly changing fee calculations.

If this is intentional (correcting the old simplification where full snapshots were ignored), no action needed -- just flagging the behavioral difference.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/global/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -201,105 +223,131 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
-                      else ((NonEmptyList.one((head, none)), emptyBalanceUpdate).some, tail).asLeft
-                  }
+            // MptStore always stores both incremental snapshot and CurrencySnapshotInfo, even for
+            // metagraphs originally in the Left(fullSnapshot) state — convertCurrencySnapshots
+            // normalizes full snapshots via CurrencyIncrementalSnapshot.fromCurrencySnapshot and
+            // toCurrencySnapshotInfo. So initialState always produces Right(incremental, info)
+            // when the metagraph exists, preserving the same processing flow. The Left(fullSnapshot)
+            // branch in the tailRecM below only triggers for brand-new metagraphs submitting their
+            // first genesis snapshot within this batch (initialState = None → deserialize full).
+            val initialStateF: F[Option[MetagraphAcceptanceResult]] = for {
```
ryle-ai: The old `initialState` read `lastGlobalSnapshotInfo.lastCurrencySnapshots.get(address)` which returned the full `Either[Signed[CurrencySnapshot], (Signed[CurrencyIncrementalSnapshot], CurrencySnapshotInfo)]`. The new code only queries `getIncrementalCurrencySnapshot` and `getCurrencySnapshotInfo`, constructing `Right(incremental, info)`. This means:

1. The `case (_, Some(Left(fullSnapshot)))` branch (line ~240) will never fire from the initial state -- only from a deserialized binary later in the loop.
2. Metagraphs whose last known state was a full snapshot (`Left` case) get `initialState = None` instead of `Some(Left(...))`, so processing falls through to the `case (None, head :: tail)` branch which tries to deserialize a full snapshot from the binary.

This is functionally fine if the next binary in the chain is indeed a full snapshot (and `syncFromGlobalSnapshotInfo` always stores both incremental + info for Left cases). But it's a subtle semantic change worth being aware of -- the previously-carried-forward full snapshot context is lost, and the processor starts from scratch for those metagraphs.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/global/GlobalSnapshotStateChannelEventsProcessor.scala
```diff
@@ -201,105 +223,131 @@ object GlobalSnapshotStateChannelEventsProcessor {
...
+                              val maybeFeeAddress = state.lastMessages.flatMap(_.get(MessageType.Owner)).map(_.address)
+
+                              // Fee deduction: if fee is required, we need a fee address (owner address from
+                              // currency messages). Without one we reject. With one, we check the local balance
+                              // accumulator first (to account for fees already deducted earlier in this batch),
+                              // falling back to MptStore for the initial balance lookup.
+                              maybeFeeAddress
+                                .filter(_ => isFeeRequired)
+                                .fold(
+                                  if (!isFeeRequired)
```
ryle-ai: Dead code: `!isFeeRequired` is always `false` here. We're inside the `Some(feeAddress)` handler of `maybeFeeAddress.filter(_ => isFeeRequired).fold(...)`, which only executes when `isFeeRequired == true`.

This whole block (lines 311-314) can be simplified to:
```scala
current.asRight[Agg].pure[F]
```
The `if (!isFeeRequired)` branch was carried over from the old flat pattern-match where the fee-not-required case was handled in the same `None` match arm. After the refactor split them into separate `fold` branches, it became vestigial.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/delegatedStake/UpdateDelegatedStakeValidator.scala
```diff
@@ -109,10 +348,12 @@ object UpdateDelegatedStakeValidator {
         lastContext: GlobalSnapshotInfo
       ): F[UpdateDelegatedStakeValidationErrorOr[Signed[UpdateDelegatedStake.Create]]] =
         for {
+          // Sort by event.ordinal (account-local sequence) to find the latest reference,
+          // consistent with DelegatedStakesRoutes.getLastReference and NodeCollateralValidator.
           lastRef <- lastContext.activeDelegatedStakes
             .getOrElse(SortedMap.empty[Address, List[DelegatedStakeRecord]])
             .get(signed.source)
-            .flatMap(stakes => Option.when(stakes.nonEmpty)(stakes.maxBy(_.createdAt)))
+            .flatMap(stakes => Option.when(stakes.nonEmpty)(stakes.maxBy(_.event.ordinal)))
```
ryle-ai, line 356: This changes the fallback (non-MptStore) sort key from `.createdAt` to `.event.ordinal`. This aligns with the MptStore implementation and `DelegatedStakesRoutes.getLastReference`, which is good — but note this is a behavioral change to the pre-existing fallback code path, not just a migration refactor. If `createdAt` (epoch progress) and `event.ordinal` (account-local sequence) ever disagree on which record is "latest", this could change validation outcomes for currency-l0/l1 modules that still use the fallback.

ryle-ai (commented):
## MptStore Migration Review

Systematic, well-structured migration of GlobalSnapshotInfo reads to MptStore. The approach of per-key lookups is sound for the fields that are accessed by address, and the fallback patterns for modules without MptStore access are reasonable.

I verified the staking balance behavioral equivalence claim in the comments: `toCurrencySnapshotInfo` sets `lastMessages = None` for full snapshots, so `fetchStakingAddress` returns `None` → `Balance.empty`, matching old `fetchStakingBalance` for the `Left(fullSnapshot)` case. The comments in `SnapshotBinaryFeeCalculator` and `StateChannelService` accurately describe this.

Two things worth addressing:

### Fallback `validateParent` sort key changed

The fallback `UpdateDelegatedStakeValidator.validateParent` changed from `maxBy(_.createdAt)` to `maxBy(_.event.ordinal)`. This is a behavioral change to the existing (non-MptStore) code path, not just a migration. If `createdAt` and `event.ordinal` can disagree on ordering, this could change which record is treated as the latest reference for create validation. The MptStore implementation already uses `.event.ordinal` (correctly matching routes), so this change aligns the fallback — but it's worth being aware this touches the pre-MptStore behavior too.

### Dead code in fee deduction

The `if (!isFeeRequired)` branch in `processCurrencySnapshots` is unreachable — it's inside the `fold`'s `Some(feeAddress)` handler which only executes when `maybeFeeAddress.filter(_ => isFeeRequired)` is `Some`, so `isFeeRequired` is always true there. The `else` branch (reject) is the only one that can fire. Should be simplified to just `current.asRight[Agg].pure[F]`.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/nodeCollateral/UpdateNodeCollateralValidator.scala
```diff
@@ -58,6 +60,228 @@ object UpdateNodeCollateralValidator {
...
+        } yield uniqueV.productR(parentV)
+      }
+
+      private def validateTokenLock(
+        signed: Signed[UpdateNodeCollateral.Create]
+      ): F[UpdateNodeCollateralValidationErrorOr[Signed[UpdateNodeCollateral.Create]]] = {
+
+        def tokenLockAvailable(address: Address): F[Boolean] =
+          for {
+            maybeDelegatedStakes <- mptStore.getDelegatedStakes(signed.source)
```
ryle-ai, line 228: Minor inconsistency: `tokenLockAvailable` uses `signed.source` directly here for delegated stakes but uses the `address` parameter on the next line for node collaterals. The same pattern exists in the fallback. Not a bug since `tokenLockAvailable(signed.source)` is always called with `signed.source`, but using `address` consistently (or dropping the parameter) would be cleaner.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/delegatedStake/UpdateDelegatedStakeValidator.scala
ryle-ai, line 356: This changes the sort key from `_.createdAt` to `_.event.ordinal` in the **fallback** (non-MptStore) implementation. The MptStore version already uses `event.ordinal`, so this aligns them — but note this affects the pre-existing code path used by currency-l0/l1 modules that don't have MptStore access. If `createdAt` and `event.ordinal` can disagree on ordering for any records, this could change which record is treated as the latest reference.

ryle-ai (commented):
## MptStore Migration: Observations

Systematic migration of GlobalSnapshotInfo reads to MptStore. The approach is consistent and tests are thoroughly updated. A few things worth noting:

### Behavioral change in fallback `validateParent`

The fallback `UpdateDelegatedStakeValidator.validateParent` changed from `maxBy(_.createdAt)` to `maxBy(_.event.ordinal)`. This aligns with the MptStore implementation and `DelegatedStakesRoutes.getLastReference`, but it's a behavioral change to the pre-MptStore code path still used by currency-l0/l1 modules. If `createdAt` (epoch progress) and `event.ordinal` (account-local sequence number) can ever disagree on which record is "latest", this could change validation outcomes for those modules. Worth confirming that's intentional since it's a change beyond the MptStore migration.

### Security fix: `address === tokenLock.source` added to NodeCollateral

Good change — both the MptStore-backed and fallback `UpdateNodeCollateralValidator.tokenLockValid` now verify token lock ownership, matching the DelegatedStake validator. This was a pre-existing gap.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/snapshot/services/GlobalL0Service.scala
```diff
@@ -67,7 +67,7 @@ object GlobalL0Service {
       private val numConcurrentQueries = 10
       private val logger = Slf4jLogger.getLoggerFromName[F](this.getClass.getName)
       private val maybeMajorityPeerIds = maybeMajorityPeerIdSet.map(_.toNonEmptyList)
-      private val ordinalRange = 0L to 3L
+      private val ordinalRange = 0L to 20L
```
ryle-ai, line 70: `ordinalRange` expanded from `0L to 3L` to `0L to 20L` — this means a joining node will now accept snapshots up to 20 ordinals behind the majority, vs 3 before. This is a substantial relaxation of the validation window during `pullLatestSnapshot`. Was this motivated by sync failures with the tighter window? Seems unrelated to the MptStore migration and worth its own PR or at least a note in the description.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/snapshot/services/GlobalL0Service.scala
```diff
@@ -138,12 +138,13 @@ object GlobalL0Service {
           _ <- logger.trace(s"${peers.map(_.id.show)}")
 
           majorityPeers <- getL0Peers(majorityPeerIds)
+          cachedMajorityOrdinal <- getMajorityOrdinal(majorityPeers)
```
ryle-ai, line 141: `getMajorityOrdinal` is now called once before the `tailRecM` peer loop and cached, where previously it was called per-peer inside `majorityOrdinalValidation`. This avoids redundant network round-trips to majority peers on each iteration, which is a good optimization. With the expanded ordinal range (20 vs 3), staleness of the cached value is unlikely to cause false rejections. Just noting the behavioral change since it's bundled here.

ryle-ai (commented):
### GlobalL0Service changes bundled into MptStore migration

This PR bundles two unrelated changes to `GlobalL0Service` alongside the MptStore migration: expanding `ordinalRange` from `0L to 3L` to `0L to 20L` and caching the majority ordinal before the peer loop. These are behavioral changes to snapshot pull validation that have nothing to do with migrating reads from `GlobalSnapshotInfo` to `MptStore`. Worth splitting out or at minimum documenting the rationale — a 6.7x expansion of the ordinal acceptance window during node sync is a significant parameter change.

### NodeCollateralValidator.tokenLockAvailable parameter inconsistency

The MptStore-backed `tokenLockAvailable` in `UpdateNodeCollateralValidator` uses `signed.source` directly for the delegated stakes lookup instead of the `address` parameter, while the equivalent in `UpdateDelegatedStakeValidator` correctly uses the `address` parameter for both lookups. Not a bug since the method is always called with `signed.source`, but worth aligning for consistency.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/nodeCollateral/UpdateNodeCollateralValidator.scala
ryle-ai, line 228: `mptStore.getDelegatedStakes(signed.source)` — this uses `signed.source` directly instead of the `address` parameter. Compare with `UpdateDelegatedStakeValidator.tokenLockAvailable` which correctly uses `address` for both lookups. Not a bug (always called with `signed.source`), but the inconsistency is a maintenance risk if the call site ever changes.

ryle-goehausen (approved):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/snapshot/services/GlobalL0Service.scala
ryle-ai, line 70: `0L to 3L` → `0L to 20L` is a significant relaxation of the ordinal acceptance window during node sync. A joining node will now accept snapshots up to 20 ordinals behind the majority, vs 3 before. What motivated this change? It seems unrelated to the MptStore migration and could use its own PR or at least a note explaining the rationale.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/delegatedStake/UpdateDelegatedStakeValidator.scala
ryle-ai, line 356: This changes the **fallback** (non-MptStore) sort key from `_.createdAt` to `_.event.ordinal`. The MptStore version already uses `event.ordinal`, so this aligns them — but it's a behavioral change to the pre-existing code path used by currency-l0/l1 modules that don't have MptStore access. If `createdAt` and `event.ordinal` can disagree on ordering for any records, this could change validation outcomes for those modules.

ryle-ai (commented):
### Unrelated behavioral changes bundled into MptStore migration

The MptStore migration itself is systematic and well-executed. However, this PR bundles two unrelated behavioral changes that affect node sync validation:

1. **`ordinalRange` expanded from `0L to 3L` to `0L to 20L`** in `GlobalL0Service` — a 6.7x expansion of the ordinal acceptance window during `pullLatestSnapshot`. This means a joining node will now accept snapshots up to 20 ordinals behind the majority, vs 3 before. This is a significant parameter change with potential network-level implications that has nothing to do with migrating reads from GlobalSnapshotInfo to MptStore.

2. **Fallback `validateParent` sort key changed** from `_.createdAt` to `_.event.ordinal` in `UpdateDelegatedStakeValidator` — this modifies the pre-MptStore code path still used by currency-l0/l1 modules. If `createdAt` (epoch progress) and `event.ordinal` (account-local sequence number) can ever disagree on ordering, this could change which record is treated as the latest reference for those modules.

Neither change is documented in the PR description. Worth splitting out or at minimum noting the rationale.

### `tokenLockAvailable` parameter inconsistency in `UpdateNodeCollateralValidator`

The MptStore-backed `tokenLockAvailable` in `UpdateNodeCollateralValidator` uses `mptStore.getDelegatedStakes(signed.source)` directly instead of the `address` parameter, while the equivalent in `UpdateDelegatedStakeValidator` correctly uses `address` for both lookups. The fallback implementation has the same inconsistency, so this isn't introduced by the PR — but since you're writing a fresh MptStore-backed implementation anyway, aligning them costs nothing.

PR #1449, opened 2026-02-19
fix: optimizing gossip

marcus-girolneto:


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/GossipDaemon.scala
```diff
@@ -177,13 +177,22 @@ object GossipDaemon {
...
+              .queryPeerRumors(PeerRumorInquiryRequest(ordinals))
+              .run(peer)
+              .evalMap(rumor => hasherSelector.withCurrent(implicit hasher => rumor.toHashed))
+              .enqueueUnterminated(rumorQueue)
+              .compile
+              .drain
+          }
+        }
+
+      private def limitOrdinals(ordinals: Map[PeerId, Ordinal]): F[Map[PeerId, Ordinal]] =
```
ryle-ai: This interacts poorly with the server-side handler in `GossipRoutes.scala`. When the server receives a `PeerRumorInquiryRequest`, it computes `additionalOrdinals` for every peer it knows about that's *not* in the request:

```scala
// GossipRoutes.scala:42
additionalOrdinals = localPeerIds.diff(inquiryOrdinals.keySet)
  .toList.map(_ -> Ordinal.MinValue)
additionalRumors <- peerRumorChain(additionalOrdinals)
```

So when `limitOrdinals` drops peers from the request to reduce its size, the server interprets those missing peers as "the client doesn't know about this peer" and attempts to send all their rumors from `Ordinal.MinValue`. In practice the response data is bounded (established peers return empty chains from `getPeerRumorsFromCursor` because their oldest stored counter isn't `MinValue`), but the server still has to iterate and look up every omitted peer. On a network with hundreds of known peers and `maxOrdinalsPerRequest = 100`, that's hundreds of extra per-peer lookups on the server side per round — partially undoing the optimization.

More importantly, for any relatively new peers (< `peerRumorsCapacity` rumors produced), the server *will* send their full rumor chains, which the client already has and will reject as `CounterTooLow`. This wastes bandwidth proportional to the number of new peers omitted.

Consider either adjusting the server-side route to not fill in `additionalOrdinals` (or making that behavior opt-in via a request flag), or accounting for this interaction when choosing the default `max-ordinals-per-request` value.

ryle-ai (commented):
The caching optimization in `RumorStorage` (`lastOrdinalsR`) is clean and correctly synchronized across all mutation paths. The env var overrides for config are a good operational improvement.

The main concern is an interaction between `limitOrdinals` and the server-side gossip route handler that may undermine the optimization — see inline comment.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/GossipDaemon.scala
scasplte2: I think this is a reasonable concern

scasplte2 (commented):
looks good but I think ryle-ai brings up a good point that is worth looking into. For small network tests, this is probably fine but could hit issues at scale I think.

ryle-ai (commented):
Agreeing with the concern raised about `limitOrdinals` and the server-side `additionalOrdinals` backfill in `GossipRoutes`. The protocol currently conflates "I don't know about this peer" with "I deliberately skipped this peer to limit request size" -- both result in the peer being absent from the ordinals map, and the server treats both the same way (backfills from `Ordinal.MinValue`). Consider extending `PeerRumorInquiryRequest` with a flag like `skipAdditionalPeers: Boolean = false` so the server can skip the `additionalOrdinals` computation when the client is deliberately limiting. This would be backwards-compatible (default `false` preserves current behavior) and avoids the server sending up to `peerRumorsCapacity` (50) duplicate rumors per excluded newer peer per round.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/GossipDaemon.scala
```diff
@@ -177,13 +177,22 @@ object GossipDaemon {
...
             .toMap
 
-          gossipClient
-            .queryPeerRumors(PeerRumorInquiryRequest(nextOrdinals))
-            .run(peer)
-            .evalMap(rumor => hasherSelector.withCurrent(implicit hasher => rumor.toHashed))
-            .enqueueUnterminated(rumorQueue)
-            .compile
-            .drain
+          limitOrdinals(nextOrdinals).flatMap { ordinals =>
```
ryle-ai: When `limitOrdinals` drops peers here, the server-side handler at `GossipRoutes.scala:42` interprets the missing peers as unknown and backfills from `Ordinal.MinValue`. For established peers (>50 stored rumors) this is cheap (returns empty chain), but for newer peers (<50 rumors) the server sends their entire stored chain, all of which the client already has and rejects as `CounterTooLow`. With `fanout=2` and up to 10 concurrent rounds, each independently sampling a different random subset, the overhead is multiplied across concurrent requests.

If reducing request payload size is the goal here, this would pair well with a server-side change to make the `additionalOrdinals` backfill opt-out.

scasplte2 (approved):
lgtm

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/config/types.scala
```diff
@@ -130,7 +130,8 @@ object types {
   case class GossipRoundConfig(
     fanout: PosInt,
     interval: FiniteDuration,
-    maxConcurrentRounds: PosInt
+    maxConcurrentRounds: PosInt,
+    maxOrdinalsPerRequest: Option[PosInt] = None
```
marcus-constellation-ai, line 134: `maxOrdinalsPerRequest` is only meaningful for peer rounds, but `GossipRoundConfig` is shared between `peerRound` and `commonRound`. `commonRound` will silently carry `None` here. This works today because `limitOrdinals` is only called from `peerRound`, but the shared struct makes the intent harder to follow. Consider a separate `PeerRoundConfig`.

marcus-constellation-ai (commented):
Overall the PR is well-structured. The `lastOrdinalsR` cache is a good optimization, the `skipAdditionalPeers` flag correctly addresses the concern raised in prior reviews, and the env-var overrides are a good operational improvement. A few observations below.

**Client timeout reduction**: Cutting `client` from 30 s to 15 s while simultaneously raising `max-concurrent-rounds` from 8 → 10 and slowing the interval from 0.2 s → 0.5 s is a trade-off that makes sense for steady-state operation, but a node catching up after downtime may be streaming many rumors per request. Worth verifying that 15 s is sufficient when `getPeerRumorsFromCursor` returns large chains across many peers.

**`maxOrdinalsPerRequest` on shared config struct**: The field is added to `GossipRoundConfig`, which is also used for `commonRound`. `commonRound` will always see `None` (no key in config, Scala default used). The field is semantically irrelevant to common rounds and creates a misleading API surface. Consider a dedicated `PeerRoundConfig`.

**Non-atomic consistency between `peerRumorsR` and `lastOrdinalsR`**: In `addPeerRumorIfConsecutive`, `peerRumorsR(rumor.origin).modify` atomically updates the chain, but the returned `lastOrdinalsR.update(...)` effect runs as a separate, subsequent `flatMap`. A concurrent `getLastPeerOrdinals` call in that window returns stale ordinals. For gossip's eventual-consistency model this is acceptable, but it is a semantic change from the old implementation where both reads came from the same `MapRef` (still non-atomic across peers, but at least consistent per-peer).

**`localPeerIds` always fetched even when `skipAdditional = true`**: `rumorStorage.getPeerIds` is a cheap `Ref` read so this is not a performance concern, just a minor code clarity point.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/GossipDaemon.scala
```diff
@@ -177,13 +177,24 @@ object GossipDaemon {
...
+                .enqueueUnterminated(rumorQueue)
+                .compile
+                .drain
+          }
+        }
+
+      private def limitOrdinals(ordinals: Map[PeerId, Ordinal]): F[(Map[PeerId, Ordinal], Boolean)] =
+        cfg.peerRound.maxOrdinalsPerRequest match {
+          case Some(max) if ordinals.size > max.value =>
+            Random[F].shuffleList(ordinals.toList).map(_.take(max.value).toMap -> true)
```
marcus-constellation-ai, line 196: `Random[F].shuffleList` produces a uniformly random ordering, so over many rounds all peers should eventually be covered. However, there is no back-pressure or priority for long-excluded peers: a peer could theoretically be excluded many consecutive rounds by chance. If `maxOrdinalsPerRequest` is set close to the actual peer count, the probability of repeated exclusion becomes significant. Worth documenting this trade-off in the config description.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/RumorStorage.scala
```diff
@@ -119,15 +119,10 @@ object RumorStorage {
...
     } yield
       new RumorStorage[F] {
-        def getLastPeerOrdinals: F[Map[PeerId, Ordinal]] = peerRumorsR.keys.flatMap { peerIds =>
-          peerIds.flatTraverse { peerId =>
-            peerRumorsR(peerId).get.map { maybeWrapper =>
-              maybeWrapper.map(peerId -> _.chain.head.ordinal).toList
-            }
-          }.map(_.toMap)
-        }
+        def getLastPeerOrdinals: F[Map[PeerId, Ordinal]] = lastOrdinalsR.get
```
marcus-constellation-ai, line 125: `getLastPeerOrdinals` now reads from the separate `lastOrdinalsR` ref rather than from `peerRumorsR`. Because the `lastOrdinalsR.update` inside `addPeerRumorIfConsecutive` is executed as an effect *after* the `peerRumorsR.modify` completes, there is a brief window where the two refs are inconsistent. Concurrent callers of `getLastPeerOrdinals` in that window will see the ordinal *before* the new rumor was added even though `peerRumorsR` already reflects it. Acceptable for gossip, but differs subtly from the old behavior where both reads originated from the same `MapRef`.

modules/node-shared/src/main/resources/application.conf
```diff
@@ -1,24 +1,37 @@
...
+      interval = ${?CL_GOSSIP_COMMON_ROUND_INTERVAL}
       max-concurrent-rounds = 4
+      max-concurrent-rounds = ${?CL_GOSSIP_COMMON_ROUND_MAX_CONCURRENT}
     }
   }
   timeouts {
     routes = 30 seconds
-    client = 30 seconds
+    routes = ${?CL_GOSSIP_ROUTES_TIMEOUT}
+    client = 15 seconds
```
marcus-constellation-ai, line 33: Client timeout halved from 30 s to 15 s. This is fine for steady-state gossip, but during catch-up (e.g. after a node restart), `peerRumorChain` may need to stream a large number of rumor entries for many peers. Verify that 15 s is sufficient under those conditions, or document that `CL_GOSSIP_CLIENT_TIMEOUT` should be raised in catch-up scenarios.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/http/routes/GossipRoutes.scala
```diff
@@ -37,9 +37,12 @@ final case class GossipRoutes[F[_]: Async](
       for {
         inquiryRequest <- req.as[PeerRumorInquiryRequest]
         inquiryOrdinals = inquiryRequest.ordinals
+        skipAdditional = inquiryRequest.skipAdditionalPeers.getOrElse(false)
         localPeerIds <- rumorStorage.getPeerIds
```
marcus-constellation-ai, line 41: `rumorStorage.getPeerIds` is always called even when `skipAdditional = true`, where `localPeerIds` ends up unused. The call is cheap (a `Ref` read), so this is not a performance concern, but structuring the `for`-comprehension to skip the call when `skipAdditional` is already true would make the control flow clearer.

PR #1450, opened 2026-02-20
fix: enforce a gate on setting owner msg on startup in public networks

scasplte2:
addresses a race condition where the owner message was not set until after ordinal 2 was created

### DIFF
--- a/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/CurrencyL0App.scala
+++ b/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/CurrencyL0App.scala
@@ -16,6 +16,7 @@ import io.constellationnetwork.currency.l0.modules._
 import io.constellationnetwork.currency.l0.node.L0NodeContext
 import io.constellationnetwork.currency.l0.snapshot.schema.{CurrencyConsensusOutcome, Finished}
 import io.constellationnetwork.currency.schema.currency._
+import io.constellationnetwork.env.AppEnvironment
 import io.constellationnetwork.ext.cats.effect.ResourceIO
 import io.constellationnetwork.ext.kryo._
 import io.constellationnetwork.node.shared.app._
@@ -31,6 +32,7 @@ import io.constellationnetwork.node.shared.snapshot.currency.CurrencySnapshotEve
 import io.constellationnetwork.node.shared.{NodeSharedOrSharedRegistrationIdRange, nodeSharedKryoRegistrar}
 import io.constellationnetwork.schema.artifact.SharedArtifact
 import io.constellationnetwork.schema.cluster.ClusterId
+import io.constellationnetwork.schema.gossip.{Ordinal => GossipOrdinal}
 import io.constellationnetwork.schema.node.NodeState
 import io.constellationnetwork.schema.semver.{MetagraphVersion, TessellationVersion}
 import io.constellationnetwork.security._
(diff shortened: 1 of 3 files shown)


marcus-girolneto (dismissed):

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/CurrencyL0App.scala
```diff
@@ -417,6 +419,19 @@ abstract class CurrencyL0App(
...
                       mkCell
                     )
+                    _ <-
+                      if (cfg.environment =!= AppEnvironment.Dev) {
+                        for {
+                          _ <- logger.info(s"Setting owner address filled on path: ${m.metagraphOwnerMessagePath}")
+                          maybeOwnerEvent <- services.currencyMessages.setInitialCurrencyOwner(m.metagraphOwnerMessagePath)
+                          _ <- logger.info(s"Owner address set")
+                          _ <- maybeOwnerEvent.traverse_ { event =>
+                            services.consensus.storage.addEvents(
```
ryle-ai, line 429: `setInitialCurrencyOwner` also enqueues this event to `l1Output` via `mkCell.run()`. The publisher daemon's `containsEvent` check will filter it, so no double-processing — but the enqueue is wasted work. If the intent is to bypass gossip entirely for non-Dev genesis, the service shouldn't be calling `mkCell` at all in this code path.

ryle-ai (commented):
In the non-Dev genesis path, `setInitialCurrencyOwner` still calls `mkCell(event).run()` internally (enqueuing to `l1Output`), and then the caller *also* injects the returned event into consensus storage via `addEvents`. The `SnapshotEventsPublisherDaemon`'s `containsEvent` filter prevents the `l1Output` copy from being double-processed, so this is safe — but the `mkCell` side effect is dead work in this path. Consider splitting the service into validate-and-return vs validate-and-process-via-cell, so each call site uses only the injection path it needs.

scasplte2:
@ryle-ai please review the latest update

ottobot-ai (approved):
## Review: LGTM ✅

The second commit (`141d6a3`) addresses ryle-ai's feedback cleanly.

### What changed

| Before | After |
|--------|-------|
| `setInitialCurrencyOwner` called `mkCell(event).run()` internally | `validateInitialCurrencyOwner` returns pure `Option[CurrencySnapshotEvent]` |
| Single code path for all environments | Environment-specific injection: non-dev → `addEvents`, dev → `mkCell` |
| `mkCell` was a service parameter | `mkCell` removed from `CurrencyMessagesService` |

### Race condition fix

The fix correctly gates the timing:

```
Non-dev genesis:
  loadGenesis → initStorages → validateOwner → addEvents(MinValue) → hashSnapshot → startFacilitating

Dev genesis:
  loadGenesis → initStorages → hashSnapshot → startFacilitating → ... → validateOwner → mkCell
```

In non-dev, owner event is injected at `GossipOrdinal.MinValue` **before** consensus starts, ensuring it's present before ordinal 2.

### No dead work

ryle-ai's concern is resolved: `validateInitialCurrencyOwner` no longer has the `mkCell` side effect — it's a pure validate-and-return. Each call site uses only the injection path it needs.

### Minor observation

The `containsEvent` filter in `SnapshotEventsPublisherDaemon` is still a good safety net, but with this refactor it's not relied upon for correctness in the owner message path.

**Ship it.** 🚀

marcus-girolneto (approved):

PR #1453, opened 2026-02-23
chore: github actions for just test

effects-ai:


.github/workflows/e2e-just-test.yml
```diff
@@ -0,0 +1,110 @@
...
+          curl -sL "https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x2EE0EA64E40A89B84B2DF73499E82A75642AC823" | sudo gpg --dearmor -o /usr/share/keyrings/sbt-archive-keyring.gpg
+          echo "deb [signed-by=/usr/share/keyrings/sbt-archive-keyring.gpg] https://repo.scala-sbt.org/scalasbt/debian all main" | sudo tee /etc/apt/sources.list.d/sbt.list
+          sudo apt-get update
+          sudo apt-get install -y sbt
+
+      - name: Create runner-logs directory
+        run: mkdir -p runner-logs
+
+      - name: Run E2E tests
+        run: just test
```
scasplte2, line 77: Does this only test the one e2e at the moment? Could we target tests individually with a CLI arg and then use multiple invocations of just test to keep the parallel runners for e2e in the CI?

scasplte2 (approved):

.github/workflows/e2e-just-test.yml
```diff
@@ -0,0 +1,110 @@
...
+      - name: Installing node dependencies
+        shell: bash
+        run: |
+          npm i @stardust-collective/dag4@v2.8.0 
+          npm i js-sha256
+          npm i axios
+          npm i brotli
+          npm i zod
+          npm i elliptic
+
```
marcus-constellation-ai, line 47: These packages are installed in the project root working directory, but they are never used there. `compose-runner.sh` installs the node dependencies it actually needs in `.github/action_scripts/` (line 207 of that script). The test scripts resolve modules relative to `.github/action_scripts/`, so this entire step has no effect on test execution. Additionally, `compose-runner.sh` does not install `elliptic`, yet `delegated_staking/lib.js` requires it directly — if it isn't a transitive dep of `dag4`, that will fail at runtime.

.github/workflows/e2e-just-test.yml
```diff
@@ -0,0 +1,110 @@
...
+      - name: Install sbt
+        run: |
+          echo "deb https://repo.scala-sbt.org/scalasbt/debian all main" | sudo tee /etc/apt/sources.list.d/sbt.list
+          curl -sL "https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x2EE0EA64E40A89B84B2DF73499E82A75642AC823" | sudo gpg --dearmor -o /usr/share/keyrings/sbt-archive-keyring.gpg
+          echo "deb [signed-by=/usr/share/keyrings/sbt-archive-keyring.gpg] https://repo.scala-sbt.org/scalasbt/debian all main" | sudo tee /etc/apt/sources.list.d/sbt.list
+          sudo apt-get update
+          sudo apt-get install -y sbt
+
+      - name: Create runner-logs directory
```
marcus-constellation-ai, line 73: This line writes an unsigned apt source entry to `sbt.list`, but line 75 immediately overwrites it with the correctly signed-by entry. The write on this line is dead code — likely a copy-paste artifact from `install_dependencies.sh`.

.github/workflows/e2e-just-test.yml
```diff
@@ -0,0 +1,110 @@
...
+
+      - name: Install just
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://just.systems/install.sh | bash -s -- --to /usr/local/bin
+          echo "/usr/local/bin" >> $GITHUB_PATH
+
+      - name: Install sbt
+        run: |
+          echo "deb https://repo.scala-sbt.org/scalasbt/debian all main" | sudo tee /etc/apt/sources.list.d/sbt.list
+          curl -sL "https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x2EE0EA64E40A89B84B2DF73499E82A75642AC823" | sudo gpg --dearmor -o /usr/share/keyrings/sbt-archive-keyring.gpg
```
marcus-constellation-ai, line 68: Piping a remote install script directly to bash (`curl ... | bash`) is a known supply chain risk — already raised by greptile-apps. The other workflows don't use `just` at all; worth considering whether downloading and verifying the script first, or pinning to a specific release hash, is acceptable for this CI context.

marcus-constellation-ai (commented):
The workflow structure is mostly fine, but there are a couple of issues worth addressing.

**Node deps installed in wrong directory**: The "Installing node dependencies" step runs `npm i` in the project root (the default working directory), but the test scripts live in `.github/action_scripts/` — which is where `compose-runner.sh` already installs packages at line 207 (`npm i @stardust-collective/dag4 js-sha256 axios brotli zod`). The packages installed at the project root are never on the require resolution path for the test scripts. This step is a no-op from the tests' perspective.

Note that `compose-runner.sh` also does not install `elliptic`, yet `delegated_staking/lib.js` has a direct `require('elliptic')`. This would work only if `elliptic` comes in as a transitive dependency of `@stardust-collective/dag4`. The empty `package-lock.json` in `action_scripts/` makes this hard to verify, and a missing `elliptic` would be a silent test failure.

**Duplicate apt source entry**: Lines 73 and 75 both write to `/etc/apt/sources.list.d/sbt.list`. Line 73 writes an unsigned entry that line 75 immediately overwrites with the correctly signed-by entry. Line 73 is dead code (likely a copy-paste artifact from `install_dependencies.sh`).

marcus-girolneto:

marcus-girolneto:

.github/workflows/e2e-just-test.yml
```diff
@@ -0,0 +1,110 @@
...
+
+jobs:
+  e2e-just-test:
+    runs-on: ubuntu-latest-m
+    timeout-minutes: 40
+
+    env:
+      BASH_DEBUG_MODE: ${{ github.event.inputs.debug_enabled || 'false' }}
+
+    steps:
```
marcus-constellation-ai, line 23: `ubuntu-latest-m` is a GitHub larger runner and is not used in any other workflow in this repo. It requires org-level enablement. If not configured for `Constellation-Labs`, this job will never start on PR triggers. Consider `ubuntu-22.04` to match the rest of the workflow suite, or confirm `ubuntu-latest-m` is already provisioned.

.github/workflows/e2e-just-test.yml
```diff
@@ -0,0 +1,110 @@
...
+          npm i axios
+          npm i brotli
+          npm i zod
+          npm i elliptic
+
+      - name: Cache SBT dependencies
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.sbt
```
marcus-constellation-ai, line 52: The existing review calls this step a no-op, but `elliptic` here is doing real work. `compose-runner.sh` installs packages to `.github/action_scripts/node_modules/` without `elliptic`, and `lib.js` has a direct `require('elliptic')`. Node resolves it by walking up to the project root where this step installs it. The step is not useless — but the mechanism is fragile and the intent is invisible. The right fix is adding `elliptic` to `compose-runner.sh`'s `npm i` invocation so the dep is co-located with the code that needs it.

marcus-constellation-ai (commented):
**Merge Confidence: 3/5** 🟡

Two issues worth addressing; existing inline reviews have flagged others (dead apt line 73, `curl|bash` on line 68) which I won't repeat.

1. **`ubuntu-latest-m` runner** — Every other workflow in this repo targets `ubuntu-22.04` or `ubuntu-latest`. `ubuntu-latest-m` is a GitHub [larger runner](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners) that must be explicitly enabled at the org level and incurs extra billing. If it hasn't been provisioned for `Constellation-Labs`, every PR trigger will silently queue forever. Verify the runner label is available, or use `ubuntu-22.04` to match the established pattern.

2. **`elliptic` resolution is fragile, and the existing review misdescribes it** — The comment on line 47 says the npm install step "has no effect," but this is incorrect. `compose-runner.sh` runs `npm i @stardust-collective/dag4 js-sha256 axios brotli zod` in `.github/action_scripts/` — without `elliptic`. `lib.js` calls `require('elliptic')` directly. The only reason this resolves at all is that Node.js traverses up the directory tree and finds `elliptic` installed at the project root by this workflow step. If someone removes the step in response to the "no-op" characterization, the tests will throw `MODULE_NOT_FOUND` (assuming `elliptic` is not a transitive dep of `dag4` — the empty `package-lock.json` makes this unverifiable). The correct fix is to add `elliptic` to `compose-runner.sh`'s own install command, making the dependency explicit where the module is actually loaded from.

PR #1455, opened 2026-02-24
fix: bug in xor distance being biased by MSB

scasplte2:


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/FacilitatorSelector.scala
```diff
@@ -1,46 +1,45 @@
...
   * ==Algorithm==
   *
   * Given a set of candidate facilitators and entropy from the last consensus outcome, selects a deterministic subset by:
   *
-  *   1. Computing XOR distance between each peer's ID hash and the entropy hash 2. Ordering peers by this distance (ascending) 3. Taking
-  *      the first M peers (where M = maxFacilitatorCount)
+  *   1. Computing a per-peer score as `SHA-256(entropy ++ peerId)` 2. Ordering peers by this score (ascending) 3. Taking the first M peers
+  *      (where M = maxFacilitatorCount)
+  *
+  * This is a '''rendezvous hashing''' (Highest Random Weight) scheme. Each peer's score is independently derived from the entropy, producing
```
marcus-constellation-ai, line 20: "Highest Random Weight" is incorrect here. The implementation takes the M **lowest** scores (ascending sort, `take(maxCount)`), which is equivalent to Lowest Random Weight. Classic HRW selects the node with the highest weight. The algorithm is unbiased either way, but the label is misleading and should be corrected or removed.

marcus-constellation-ai (commented):
The rendezvous hashing approach is algorithmically sound and correctly addresses the MSB bias in XOR distance selection. A few issues worth noting:

**"Highest Random Weight" naming is wrong.** The implementation sorts scores ascending and takes the first M (i.e., peers with the *lowest* scores). Classic HRW hashing selects the node with the *highest* random weight. The label in the scaladoc is inverted. The algorithm is still uniform and correct — it just shouldn't be called "Highest Random Weight."

**No tests.** This changes the core consensus facilitator selection algorithm with zero test coverage (none existed before either). For a security-critical path that determines which nodes participate in each consensus round, at minimum there should be a unit test verifying determinism, correct subset size, and that all candidates are returned when count ≤ maxCount.

**Implicit protocol-breaking change.** Nodes on old vs. new software will compute different facilitator sets from the same inputs. This is expected for a bug fix, but the coordinated-upgrade requirement isn't documented anywhere in the PR or code.

marcus-girolneto (approved):

PR #1456, opened 2026-02-25
ci: align testnet release with intnet & mainnet

scasplte2:


scasplte2:
@marcus-constellation-ai any thoughts?

.github/workflows/testnet-release.yml
```diff
@@ -34,15 +52,32 @@ jobs:
       - name: Get the pre-release version
         id: get_version
         run: |
-          sbt "writeVersion ./version"
-          echo "VERSION=$(cat ./version)" >> "$GITHUB_OUTPUT"
-          echo "Release-As: $(cat ./version)" >> $GITHUB_STEP_SUMMARY
-          rm ./version
+          {
+            echo 'VERSION<<EOF'
+            npx commit-and-tag-version --dry-run --prerelease alpha | grep -oP 'tagging release v\K(.*)'
```
marcus-constellation-ai: If `commit-and-tag-version` output doesn't contain the expected `tagging release v...` line (tool not installed, format change, or no version bump detected), `VERSION` will be empty string. The heredoc syntax sets it silently to `""` rather than failing. A subsequent `git tag -a v -m "Release "` creates a nonsensical `v` tag. Consider validating here:
```bash
VERSION=$(npx commit-and-tag-version --dry-run --prerelease alpha | grep -oP 'tagging release v\K(.*)')
[ -n "$VERSION" ] || { echo "ERROR: could not extract version"; exit 1; }
```
or add the guard at the top of the `Create release tag` run block.

.github/workflows/testnet-release.yml
```diff
@@ -34,15 +52,32 @@ jobs:
...
+          git config user.email 41898282+github-actions[bot]@users.noreply.github.com
+
+      - name: Create release tag
+        id: create_tag
+        env:
+          VERSION: ${{ steps.get_version.outputs.VERSION }}
+        run: |
+          echo "Release-As: $VERSION" >> $GITHUB_STEP_SUMMARY
+          git tag -a v$VERSION -m "Release $VERSION"
+          git push origin v$VERSION -f
```
marcus-constellation-ai: The `check_tag` pre-check prevents re-running on the *same* commit, but `-f` will silently overwrite the tag if a different commit on `release/testnet` produces the same `commit-and-tag-version` output (e.g., only `chore:` or `ci:` commits since the last tagged release). Dropping `-f` would make this conflict fail loudly rather than silently move the tag. The same pattern exists in `release.yml` for integrationnet — tracking it there would fix both.

marcus-constellation-ai (commented):
**Merge Confidence: 4/5** 🟢

Solid alignment with the intnet/mainnet pattern. Two issues worth addressing before this ships.

1. **`get_version` — empty `VERSION` is unguarded**: `npx commit-and-tag-version --dry-run --prerelease alpha | grep -oP 'tagging release v\K(.*)'` produces empty output if the grep finds no match (unexpected tool output format, first run with no version history, npx resolution failure). `VERSION` then silently becomes an empty string. The subsequent `git tag -a v -m "Release "` creates a bogus `v` tag and `git push origin v -f` pushes it. The Copy dependencies and S3 steps would then operate on paths without a version component. Add a guard in the `Create release tag` step: `[ -n "$VERSION" ] || { echo "ERROR: Failed to extract version from commit-and-tag-version"; exit 1; }`. The same gap exists in `release.yml` for integrationnet.

2. **`git push origin v$VERSION -f` can silently retarget an existing tag**: The pre-check (`check_tag`) prevents re-running on the same commit, but doesn't prevent the scenario where two different commits on `release/testnet` produce the same computed version (e.g., only `chore:` commits between runs). In that case the force-push silently moves the version tag from the first commit to the second with no error signal. Dropping `-f` would cause the push to fail loudly in this case, making the conflict explicit. The same pattern exists in `release.yml`.

scasplte2:
@marcus-constellation-ai everything addressed to your satisfaction?

.github/workflows/testnet-release.yml
```diff
@@ -34,15 +52,30 @@ jobs:
...
+          git config user.email 41898282+github-actions[bot]@users.noreply.github.com
+
+      - name: Create release tag
+        id: create_tag
+        env:
+          VERSION: ${{ steps.get_version.outputs.VERSION }}
+        run: |
+          echo "Release-As: $VERSION" >> $GITHUB_STEP_SUMMARY
+          git tag -a v$VERSION -m "Release $VERSION"
+          git push origin v$VERSION
```
marcus-constellation-ai, line 71: `git push origin v$VERSION` (no `-f`) will fail loudly if the same version was already tagged on a different commit — correct behavior, but the failure lands after tests have run. A remote tag pre-check (`git ls-remote --tags origin "refs/tags/v$VERSION"`) after `get_version` would catch this before the expensive steps. Low-frequency scenario (requires all-`chore:`/`ci:` commits since the last release), so fine to leave as-is.

marcus-constellation-ai (commented):
**Merge Confidence: 4/5** 🟢

Clean CI alignment. Both concerns from the prior review are addressed in the current state: `VERSION` is guarded with an explicit `exit 1`, and the tag push has no `-f` (safer than `release.yml`'s integrationnet path).

1. **Version-conflict late failure**: `check_tag` correctly guards against re-releasing on the same commit, but doesn't catch the case where a different commit on `release/testnet` produces the same version number (e.g., only `chore:`/`ci:` commits since the last release). In that scenario, `git push origin v$VERSION` fails loudly — which is the right behavior — but only after tests have already run (Assembly hasn't started yet at that point). This is a known operational annoyance, not a correctness bug. Mentioned for awareness.

marcus-girolneto (approved):

PR #1457, opened 2026-02-26
ci: separate sdk publish workflow from release

scasplte2:


.github/workflows/publish-sdk.yml
```diff
@@ -0,0 +1,38 @@
+on:
+  push:
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+"       # stable: v4.1.0
```
marcus-constellation-ai, line 4: In GitHub Actions glob patterns, `+` is a literal character, not a "one or more" quantifier — that role belongs to the extglob form `+(...)`. So `v[0-9]+.[0-9]+.[0-9]+` would match a tag like `v4+.1+.0+` (with literal plus signs), not `v4.1.0`. If this is correct, the workflow never fires and SDK publish silently stops working.

Consider `v[0-9]*.[0-9]*.[0-9]*` for stable and `v[0-9]*.[0-9]*.[0-9]*-rc.*` for RCs, or validate by pushing a test tag to a fork before merging.

build.sbt
```diff
@@ -73,6 +73,14 @@ ThisBuild / version := {
 // Disable dynver's Sonatype snapshot suffix (we use +metadata format for dev versions)
 ThisBuild / dynverSonatypeSnapshots := false
 
+// Explicitly derive isSnapshot from git state so ci-release behaves correctly
+// in any context (tag push, branch push, or local invocation)
+ThisBuild / isSnapshot := {
+  dynverGitDescribeOutput.value.forall(out =>
```
marcus-constellation-ai, line 79: `forall` on `None` (no git describe output) returns `true`, making `isSnapshot = true` — safe fallback when git state is unavailable. Correct behavior for the tag-triggered workflow: when HEAD is exactly at a clean tag, `distance == 0` and `!isDirty()` both evaluate to `false`, so the predicate is `false` and `isSnapshot` becomes `false` as intended.

marcus-constellation-ai (commented):
**Merge Confidence: 3/5** 🟡

The design is the right call — tag-triggered publish is cleaner than the old workaround of overriding `GITHUB_REF` on a branch push. Two things worth verifying before merging.

1. **`publish-sdk.yml:4-5`** — The tag filter patterns `v[0-9]+.[0-9]+.[0-9]+` use `+` after a character class. In GitHub Actions' glob (minimatch), `+` outside of extglob syntax `+(...)` is a literal character, not a quantifier. If that's the case here, neither pattern matches a real semver tag like `v4.1.0` — the workflow would silently never trigger, making SDK publish a regression to zero. The safe alternatives are `v[0-9]*.[0-9]*.[0-9]*` (with `*`) or simply `v*` scoped to exclude alphas. Worth running a dry-run push of a test tag against a fork before merging.

2. **`build.sbt:156-160`** — The explicit `isSnapshot` override is necessary. Without it, `dynverSonatypeSnapshots := false` disables the `-SNAPSHOT` suffix in the version string, but the `isSnapshot` key (which `ci-release` reads to choose snapshot vs. release repo) would still be derived from whether the version string ends in `-SNAPSHOT`. Since the custom version format uses `+metadata` notation instead, `isSnapshot` would incorrectly be `false` on every build. The fix — computing `isSnapshot` directly from `dynverGitDescribeOutput` — correctly makes it `false` only when HEAD is exactly at a clean tag. The `forall` on `Option` returning `true` when `None` (no git info → snapshot) is the right safe default.

All other changes are straightforward improvements: removing the tag force-push, generalizing the already-tagged abort to both branches, the version-extraction refactor with explicit error checking, and the documentation.

.github/workflows/publish-sdk.yml
scasplte2: Good catch to flag — this is a common source of confusion. GitHub Actions filter patterns extend standard fnmatch with regex-like quantifiers: + means "one or more
  of the preceding character/class", not a literal +.

  The official https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet documents this explicitly, and even
  includes v[12].[0-9]+.[0-9]+ as a canonical example for matching semver tags like v1.10.1.

  So v[0-9]+.[0-9]+.[0-9]+ correctly matches v4.1.0, and the -rc.* variant correctly matches v4.1.0-rc.1.

build.sbt
scasplte2: Agreed on the analysis — forall on None → true is the intentional safe default, and on a clean tag distance == 0 && !isDirty() → isSnapshot = false as expected.

marcus-girolneto (approved):

PR #1459, opened 2026-02-27
chore: cluster agnostic test runner

effects-ai:


docker/bin/compose-runner.sh
```diff
@@ -41,215 +41,367 @@ echo "Running in top level directory $cur_dir"
...
+      up -d
+    fi
 
-  
-  export PROFILE_ML0_ARG=""
-  if [ "$i" -lt "$NUM_ML0_NODES" ]; then
-    export PROFILE_ML0_ARG="--profile ml0"
-  fi
+    cd ../../
+  done
```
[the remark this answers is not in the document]
scasplte2: I think I was the one that added this. Have you run into this sequencing issue @effects-ai ?

scasplte2:
@ryle-ai review please

ryle-ai (commented):
The cluster-agnostic refactor is a useful direction -- being able to run tests against remote deployments is clearly needed. Two concrete issues to address before merge:

1. **`just dag-tests` references a script that doesn't exist** (`docker/bin/run-dag-tests.sh`). This will fail immediately for anyone who tries to use it.

2. **GL0/GL1 phased startup was removed.** The old code explicitly started GL0 first, waited for cluster readiness, then started GL1 -- with a comment explaining that GL1 crashes if GL0 isn't ready for L0PeerDiscovery (state gets stuck at SessionStarted). The new code starts both profiles simultaneously. The docker-compose files have no `depends_on` between GL0 and GL1, so there's no other ordering guarantee. @scasplte2 already asked about this -- has the underlying application-layer issue been resolved, or has this just not been hit yet in testing?

Notably, the metagraph startup IS properly phased (GL0 wait -> ML0 wait -> CL1/DL1), which makes the GL0->GL1 omission look like an oversight rather than a deliberate decision.

Separately: the CI matrix now runs 8 independent jobs, each doing sbt assembly + docker build + full cluster startup. That's a significant compute cost increase vs. the previous single-job approach. The parallelization benefit is real, but worth confirming the CI budget can absorb 8x the runner-minutes.

justfile
```diff
@@ -52,6 +52,10 @@ debug-main:
 check:
     @bash sbt --error 'scalafixAll --check --rules OrganizeImports;scalafmtCheckAll;test'
 
+# Run non-metagraph e2e tests (DAG cluster check, delegated staking, token lock replacement) against a running environment. Use --test=<name> to select specific tests.
+dag-tests *extra_args:
+	@bash docker/bin/run-dag-tests.sh {{ extra_args }}
```
ryle-ai: `docker/bin/run-dag-tests.sh` does not exist in the repo. This command will fail. Either create the script or inline the compose-runner call:
```suggestion
	@bash docker/bin/compose-runner.sh --test=dag-cluster,delegated-staking,token-lock-replacement {{ extra_args }}
```

docker/bin/compose-runner.sh
```diff
@@ -41,215 +41,414 @@ echo "Running in top level directory $cur_dir"
...
+    cp ../../docker/docker-compose.volumes.yaml . ; \
+    cp ../../docker/docker-compose.metagraph.yaml . ;
+    cp ../../docker/docker-compose.metagraph-test.yaml . ;
+    cp ../../docker/docker-compose.metagraph-genesis.yaml . ;
 
-  export PROFILE_CL1_ARG=""
-  if [ "$i" -lt "$NUM_CL1_NODES" ]; then
-    export PROFILE_CL1_ARG="--profile cl1"
-  fi
+    docker_additional_args="$PROFILE_GL0_ARG $PROFILE_GL1_ARG"
```
ryle-ai: This combines GL0 and GL1 into a single `docker compose up -d`, removing the phased startup that previously existed. The old code had an explicit comment:

> GL1 nodes crash if GL0 isn't ready for L0PeerDiscovery, and the join state machine does not recover (state gets stuck at SessionStarted).

There's no `depends_on` in the docker-compose files between GL0 and GL1, so Docker won't enforce ordering either. The metagraph layers below *do* get proper phased startup (GL0 wait -> ML0 wait -> CL1/DL1), which makes this gap stand out.

If the L0PeerDiscovery crash was fixed at the application layer, a brief note here explaining that would be helpful for future readers. If it wasn't fixed, this needs the phased approach back -- at minimum waiting for GL0 `/cluster/info` to return before starting GL1.

scasplte2:
> Separately: the CI matrix now runs 8 independent jobs, each doing sbt assembly + docker build + full cluster startup. That's a significant compute cost increase vs. the previous single-job approach. The parallelization benefit is real, but worth confirming the CI budget can absorb 8x the runner-minutes.

@effects-ai I believe you could build everything once, cache then place the cached jars in `./docker/jars/` and use `--skip-assembly` to avoid running parallel build steps?

docker/bin/compose-runner.sh
scasplte2: L0 peer discovery has been fixed so maybe the phased startup is no longer necessary

justfile
ryle-ai: `docker/bin/run-dag-tests.sh` doesn't exist. This needs to either be created or inlined as `docker/bin/compose-runner.sh --test=dag-cluster,delegated-staking,token-lock-replacement`.

.github/action_scripts/shared/network.js
```diff
@@ -4,15 +4,16 @@ const { withRetry } = require("./operations");
...
-        currencyL1Url: `http://localhost:${currencyL1PortPrefix}00`,
-        dataL1Url: `http://localhost:${dataL1PortPrefix}00`,
-        extendedDagL1Url: `http://localhost:${dagL1PortPrefix}50`,
-        extendedDataL1Url: `http://localhost:${dataL1PortPrefix}50`
+        globalL0Url: process.env.GL0_URL || `${host}:${dagL0PortPrefix}00`,
+        dagL1Url: process.env.GL1_URL || `${host}:${dagL1PortPrefix}00`,
+        currencyL0Url: process.env.ML0_URL || `${host}:${metagraphL0PortPrefix}00`,
+        currencyL1Url: process.env.CL1_URL || `${host}:${currencyL1PortPrefix}00`,
+        dataL1Url: process.env.DL1_URL || `${host}:${dataL1PortPrefix}00`,
+        extendedDagL1Url: `${host}:${dagL1PortPrefix}50`,
```
ryle-ai: `extendedDagL1Url` and `extendedDataL1Url` (line 16) don't get URL override treatment. The other five URLs all have `process.env.GL0_URL || ...` etc., but these two just use `${host}:${portPrefix}50`. For a remote host, port 9150/9450 probably isn't valid. Consider adding override support for consistency:
```js
extendedDagL1Url: process.env.GL1_URL || `${host}:${dagL1PortPrefix}50`,
extendedDataL1Url: process.env.DL1_URL || `${host}:${dataL1PortPrefix}50`
```
(Or separate env vars if the extended endpoints differ from the primary ones.)

docker/bin/compose-runner.sh
ryle-ai: This combines GL0 and GL1 profiles into `docker_additional_args` and starts both in one `docker compose up -d`. The docker-compose files have no `depends_on` between the two services.

Notably, the metagraph wait at line ~185 only checks `node_count >= 1` before proceeding, whereas the old GL0 wait checked for the full `$NUM_GL0_NODES` cluster. That's probably fine for ML0 (it only needs to reach one GL0 peer), but GL1 also does L0PeerDiscovery against GL0 and gets zero wait time here.

ryle-ai (commented):
The two blocking issues from earlier reviews remain open:

1. `just dag-tests` references `docker/bin/run-dag-tests.sh` which doesn't exist.
2. GL0/GL1 phased startup was removed. The old code explicitly started GL0 first, waited for full cluster readiness, then started GL1 -- with a comment about GL1's join state machine getting stuck at `SessionStarted` if GL0 isn't ready. The new code starts both in a single `docker compose up -d` per node with no ordering guarantee. Meanwhile, the metagraph startup IS properly phased (GL0 wait → ML0 → ML0 wait → CL1/DL1), which makes the GL0→GL1 gap look like an oversight.

One additional observation: in `shared/network.js`, the `extendedDagL1Url` and `extendedDataL1Url` don't get the `process.env.*_URL` override treatment that the other five URLs get. For remote hosts these would resolve to `${TEST_HOST}:${portPrefix}50` which is probably wrong. Not blocking since extended tests aren't in the CI matrix, but it's an inconsistency that would bite someone running 6-node tests against a remote deployment.

scasplte2 (approved):
lgtm

PR #1461, opened 2026-03-06
chore: snapshot streaming tests

effects-ai:


docker/snapshot-streaming/sql/delegated_staking.sql
```diff
@@ -0,0 +1,69 @@
...
+
+CREATE TABLE delegate_stake_rewards (
+	global_snapshot_hash varchar NOT NULL REFERENCES global_snapshots(hash) ON DELETE CASCADE,
+	address varchar NOT NULL REFERENCES addresses(address) ON DELETE CASCADE,
+	node_id varchar NOT NULL,
+	rewards int8 NOT NULL,
+	created_at timestamp DEFAULT now() NOT NULL,
+	updated_at timestamp DEFAULT now() NOT NULL
+);
+CREATE INDEX delegate_stake_rewards_changes_global_snapshot_hash_idx ON public.delegate_stake_balance_changes USING btree (global_snapshot_hash);
```
ryle-ai: These three `CREATE INDEX` statements (lines 66-68) target `public.delegate_stake_balance_changes` but are named `delegate_stake_rewards_changes_*` and follow the `delegate_stake_rewards` table definition. Looks like a copy-paste from the `delegate_stake_balance_changes` block above where the table name wasn't updated.

Result: `delegate_stake_rewards` gets no indexes, `delegate_stake_balance_changes` gets redundant duplicates. Doesn't affect the E2E test (tiny dataset), but should be fixed upstream in the snapshot-streaming repo.

ryle-ai (commented):
`delegated_staking.sql` has three indexes for `delegate_stake_rewards` that accidentally target `delegate_stake_balance_changes` instead (see inline). This is upstream in the snapshot-streaming repo — the build script re-downloads the SQL at runtime so it can't be fixed in this PR alone.

docker/snapshot-streaming/build-snapshot-streaming.sh
```diff
@@ -0,0 +1,92 @@
...
+  exit 1
+fi
+echo "block_explorer migrations ready: $BE_DIR/prisma/migrations"
+
+# --- Download snapshot-streaming supplemental SQL files (branch-specific schema additions) ---
+
+SQL_DIR="$SS_DIR/sql"
+rm -rf "$SQL_DIR"
+echo "Checking for snapshot-streaming supplemental SQL files (branch: $SS_BRANCH)..."
+ss_sql_listing=$(curl -sf "https://api.github.com/repos/Constellation-Labs/snapshot-streaming/contents/sql?ref=${SS_BRANCH}" 2>/dev/null || echo "")
```
ryle-ai: This `curl` to `api.github.com` is unauthenticated, so it's subject to the 60-request/hour IP-based rate limit. In CI the `GITHUB_TOKEN` is in the environment (the build step sets it explicitly) but it's not passed to curl. If rate-limited, the response is a 403, `-sf` makes curl fail silently, and supplemental SQL files are never downloaded. The prisma schema alone might be enough, but if snapshot-streaming writes to columns that only exist in the supplemental SQL, inserts would fail at runtime.

Consider:
```bash
auth_header=""
if [ -n "${GITHUB_TOKEN:-}" ]; then
  auth_header="Authorization: token $GITHUB_TOKEN"
fi
ss_sql_listing=$(curl -sf ${auth_header:+-H "$auth_header"} "https://api.github.com/..." 2>/dev/null || echo "")
```

docker/bin/compose-runner.sh
```diff
@@ -313,6 +325,129 @@ else
...
+  # --- Snapshot-streaming infrastructure ---
+  if [ -z "$SELECTED_TESTS" ] || echo "$SELECTED_TESTS" | tr ',' '\n' | grep -qx "snapshot-streaming"; then
+    echo "================================================"
+    echo "Setting up snapshot-streaming infrastructure"
+    echo "================================================"
+
+    SS_DIR="$PROJECT_ROOT/docker/snapshot-streaming"
+
+    # Build/obtain JAR + SQL
+    source "$SS_DIR/build-snapshot-streaming.sh"
```
ryle-ai, line 338: This is the second `source` of `build-snapshot-streaming.sh` — the first is at line 142 (before the `BUILD_ONLY` exit). In a normal test run both execute. The JAR build is guarded by an existence check, but the block_explorer fetch and SQL download run again unconditionally (the script does `rm -rf "$SQL_DIR"` then re-downloads). A simple guard at the top of the build script (e.g., `if [ -f "$JAR_DEST" ] && [ -d "$BE_DIR/.git" ] && [ -d "$SQL_DIR" ]; then return 0; fi`) would save the duplicate work.

.gitignore
```diff
@@ -42,4 +42,9 @@ docker/jars/
...
 .workspace/
 .claude/*.local.*
-tx-sender.conf
\ No newline at end of file
+tx-sender.conf
+docker/snapshot-streaming/*.jar
+docker/snapshot-streaming/*.conf
+docker/snapshot-streaming/data/
+docker/snapshot-streaming/.build/
+docker/snapshot-streaming/block-explorer/
```
ryle-ai, line 50: `docker/snapshot-streaming/sql/` is missing — the build script downloads SQL files there and `rm -rf`s them on re-run, but a local build would leave them showing up in `git status`.

ryle-ai (commented):
Solid E2E test harness for snapshot-streaming. The two-phase schema approach (prisma base + supplemental SQL) and dynamic SQL discovery via the GitHub API are good design choices. A couple of minor robustness concerns below.

docker/snapshot-streaming/docker-compose.yaml
```diff
@@ -0,0 +1,41 @@
...
+    volumes:
+      - ${SS_JAR_PATH}:/app/snapshot-streaming.jar:ro
+      - ${SS_CONFIG_PATH}:/app/application.conf:ro
+      - ${SS_DATA_PATH}:/app/data
+    networks:
+      tessellation_common:
+        ipv4_address: ${NET_PREFIX}.61
+
+volumes:
+  ss-pgdata:
```
ryle-ai, line 37: Missing explicit `name:` — Docker Compose will create this as `snapshot-streaming_ss-pgdata` (project-name prefix). The cleanup script passes `ss-pgdata` to `docker volume rm`, which won't match. The main `docker-compose.yaml` avoids this with `name: gl0-data${CONTAINER_NAME_SUFFIX:-}`. Add `name: ss-pgdata` here:
```yaml
volumes:
  ss-pgdata:
    name: ss-pgdata
```

.gitignore
ryle-ai, line 50: `docker/snapshot-streaming/sql/` is missing — the build script downloads SQL files there and `rm -rf`s them on re-run, but a local build leaves them in `git status`.

ryle-ai (commented):
The `ss-pgdata` volume in `docker-compose.yaml` lacks an explicit `name:` field, so Docker Compose will create it as `snapshot-streaming_ss-pgdata` (prefixed with the project name derived from the compose file's directory). The cleanup script tries to remove `ss-pgdata`, which won't match. The main `docker-compose.yaml` avoids this by using `name: gl0-data${CONTAINER_NAME_SUFFIX:-}` on its volumes — same pattern should be applied here.

docker/snapshot-streaming/docker-compose.yaml
ryle-ai, line 37: Needs `name: ss-pgdata` to match the cleanup script. Every other volume in this repo uses explicit `name:` to prevent Docker Compose project-name prefixing (see `docker-compose.volumes.yaml`). Without it, `cleanup_container snapshot-streaming-postgres ss-pgdata` silently fails to remove the volume because Docker actually created it as `snapshot-streaming_ss-pgdata`.

```yaml
volumes:
  ss-pgdata:
    name: ss-pgdata
```

.gitignore
ryle-ai, line 50: Missing `docker/snapshot-streaming/sql/` — the build script downloads SQL files there, and they'll show up in `git status` after a local build.

.github/workflows/e2e-just-test.yml
```diff
@@ -56,7 +56,9 @@ jobs:
           sudo apt-get install -y sbt
 
       - name: Build all JARs
-        run: just build --use-test-metagraph --num-gl0=3
+        run: just build --use-test-metagraph --num-gl0=3 --snapshot-streaming-branch=testing --block-explorer-branch=increase_delegated_stakes
```
ryle-ai: `--snapshot-streaming-branch=testing` and `--block-explorer-branch=increase_delegated_stakes` are hardcoded here and in the matrix entry at line 95. If either branch gets deleted from its repo, every CI run that includes the snapshot-streaming test will fail. Are these intended to be long-lived branches, or should this default to `develop` (which `set-env.sh` already does) and only be overridden via workflow dispatch inputs for ad-hoc testing?

ryle-ai (commented):
The `ss-pgdata` volume in the snapshot-streaming docker-compose.yaml needs an explicit `name:` key. The existing `docker-compose.volumes.yaml` uses `name: gl0-data${CONTAINER_NAME_SUFFIX:-}` on every volume specifically to prevent Docker Compose from prefixing the project name. Without it, Docker creates the volume as `snapshot-streaming_ss-pgdata`, but the cleanup script passes bare `ss-pgdata` to `docker volume rm`. On ephemeral CI runners this is harmless, but for local development it leaves stale volumes between runs.

Also: `docker/snapshot-streaming/sql/` is missing from `.gitignore` — the build script downloads SQL files there and `rm -rf`s them on rebuild, but a local build leaves them in `git status`.

effects-ai:
@greptile-apps please review it again

effects-ai:
@ryle-ai please review it again

ottobot-ai (approved):
## Review — PR #1461: Snapshot Streaming E2E Tests

**Author:** effects-ai  
**CI:** ✅ All 12 checks passing

### Summary

Adds E2E testing infrastructure for `snapshot-streaming` indexer. Also consolidates 10+ separate CI workflow files into a single `e2e-just-test.yml` matrix — good architectural improvement.

### Architecture

Solid implementation:
1. **Build** — Clones snapshot-streaming, builds JAR, clones block_explorer for Prisma migrations
2. **Config** — Generates `application.conf` with correct L0 peer and DB connection
3. **Docker** — Postgres 15-alpine + snapshot-streaming container with proper health checks
4. **Test** — Seeds initial snapshot from GL0, waits for indexer to populate DB

### Minor Issues

1. **[build-snapshot-streaming.sh:49-50]** JAR discovery fallback `*.jar` could match non-assembly JARs. Consider `find` with explicit pattern.

2. **[compose-runner.sh:359-360]** proofsHash computation note: Uses SHA256 of JSON instead of Brotli-compressed. Acceptable for E2E (documented in comment).

3. **[generate-config.sh:26-27]** L0 peer IP hardcoded to `${NET_PREFIX}.10:9000`. Consider deriving from `DAG_L0_PORT_PREFIX` for consistency.

4. **[e2e-just-test.yml:105]** `extra-args` undefined for most matrix entries — works but inconsistent.

### Suggestions

- Add seed snapshot metadata (ordinal, hash) to failure diagnostics
- Document minimum CI runner memory requirements (1GB for snapshot-streaming + Postgres + tessellation)
- `ss-pgdata` volume persists after cleanup — add `docker volume rm` if full cleanup desired

### Workflow Consolidation ✓

Deletes 10 separate workflow files, consolidates into `e2e-just-test.yml` matrix. Single build phase, parallel test execution — cleaner architecture.

### Verdict: **APPROVE**

Well-structured addition with good E2E coverage. All CI checks pass including the new snapshot-streaming test (5m50s).

---
*Note: External contributor PR — flagging for @scasplte2 final review before merge.*

scasplte2 (dismissed):

scasplte2 (approved):
nicely done 🔥

PR #1464, opened 2026-03-10
chore: align develop config with release/mainnet

scasplte2:
## Summary
Backport mainnet-specific configuration that was committed directly to `release/mainnet` and never propagated to develop. This alignment is required before force-pushing develop to `release/mainnet` for the v4.0.0 release.

## Changes
* **Added** - Locked address `DAG0y9cvwP9hgTMNoSbXYgjyroCNPirzkHagsuqk` missing from develop
* **Modified** - Set `tessellation-301-migration.mainnet` from `0` to `4915254`
* **Modified** - Set `metagraph-sync-data.mainnet` from `4409045` to `4915254`
* **Modified** - Set `updated-last-sync-global-order.mainnet` from `9999999` to `4915254`
* **Modified** - Set `updated-last-sync-global-from-peers-in-consensus.mainnet` from `9999999` to `4915254`
* **Modified** - Updated `adjustments.json` with additional mainnet balance corrections (PacaSwap)

## Testing
Config-only changes — no behavioral changes. CI should pass as-is. These values match what has been running on mainnet since the v3.5.x releases.

### DIFF
--- a/modules/node-shared/src/main/resources/adjustments.json
+++ b/modules/node-shared/src/main/resources/adjustments.json
@@ -1021,5 +1021,169 @@
         "deduct": -90000000000.0
       }
     ]
+  },
+  {
+    "currencyId": "DAG7X5idd4aLfp4XC6WQdG1eDfR3LGPVEwtUUB2W",
+    "snapshotOrdinal": 145000,
+    "adjustments": [
+      {
+        "address": "DAG09tkvpazCUBp9KZxcYCp8WaphkKFvpH1ZYtGx",
+        "reason": "SpendTransactionSourceNotApplied",
+        "reference": [
+          "80b6b92101f766f84e59873fa27a70a80931a320ea22617cc0a7f1421c7c2705",
+          "bb37bb080605782ec532b0b53f93e5390f1082ceb9b04f2f7f13d44439323e5e"
+        ],
+        "increase": 5234237.0
+      },
+      {
+        "address": "DAG0CQBBTeP7TUQ9JJnsLS4f8ePnWFfoyxMdudr8",
+        "reason": "SpendTransactionSourceNotApplied",
+        "reference": [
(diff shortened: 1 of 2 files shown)


marcus-girolneto (approved):

ryle-ai (commented):
One structural observation about duplicate `currencyId` entries in `adjustments.json` — see inline comment.

modules/node-shared/src/main/resources/adjustments.json
```diff
@@ -1021,5 +1021,169 @@
         "deduct": -90000000000.0
       }
     ]
+  },
+  {
+    "currencyId": "DAG7X5idd4aLfp4XC6WQdG1eDfR3LGPVEwtUUB2W",
```
ryle-ai, line 1026: This `currencyId` already exists at line 564 (ordinal 109991). Similarly, `DAG7Ghth1WhWK83SB3MtXnnHYZbCsmiRTwJrgaW1` duplicates line 273 (ordinal 371419), and `DAG0CyySf35ftDQDQBnd1bdQ9aPyUdacMghpnCuM` duplicates line 3 (ordinal 17103718).

`BalanceAdjustmentLoader.convertToAdjustmentEntries` converts the parsed list to a `Map[Address, BalanceAdjustmentAtOrdinal]` via `.toMap` (line 100), so the last entry per key wins — the three earlier adjustment sets become unreachable. No runtime impact since those ordinals have long been processed on mainnet, but ~1000 lines of adjustment data in this file are now effectively dead.

Not blocking since this matches mainnet. Worth knowing though: the loader doesn't support multiple adjustment events per currency. If a fourth round of adjustments is ever needed for one of these currencies, the third round's data would also become unreachable. At that point it might be worth either extending the loader to support a list of ordinals per currency, or pruning the earlier entries.

PR #1465, opened 2026-03-10
ci: fix stable release changelog comparing against prerelease tags

scasplte2:
## Summary
Fix the release workflow so that stable (mainnet) releases generate a complete changelog comparing against the last stable tag, not the nearest prerelease tag.

## Changes
* **Added** - New workflow step that removes prerelease tags (rc/alpha) from the local CI clone before version computation on `release/mainnet`. This forces `commit-and-tag-version` to walk back to the last stable tag.
* **Modified** - Build Changelog step now passes `--prerelease rc` for integrationnet releases, matching the version computation step.

## Context
The v4.0.0 release changelog only showed commits since `v4.0.0-rc.7` (the nearest ancestor prerelease tag) instead of all changes since `v3.5.12`. This happened because `commit-and-tag-version` uses git ancestry to find the comparison base, and prerelease tags from the integrationnet promotion path were closer ancestors than the previous stable tag.

The fix is minimal and safe — prerelease tags are only deleted from the ephemeral CI runner clone, not from the remote repository. Integrationnet (rc) releases are unaffected since their changelog correctly walks back to the previous rc tag.

## Testing
Workflow-only change. Can be validated by running `commit-and-tag-version --dry-run` locally after deleting prerelease tags and confirming it compares against the last stable tag.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -77,7 +77,7 @@ jobs:
         run: |
           if [[ "${{ github.ref_name }}" == "release/mainnet" ]]; then
             DRY_RUN=$(npx commit-and-tag-version --dry-run)
-          else
+          elif [[ "${{ github.ref_name }}" == "release/integrationnet" ]]; then
             DRY_RUN=$(npx commit-and-tag-version --dry-run --prerelease rc)
           fi
           {
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,6 +49,18 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Remove prerelease tags for stable releases
+        if: github.ref_name == 'release/mainnet'
+        run: |
+          # Remove prerelease tags from the local clone so commit-and-tag-version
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -63,9 +75,14 @@ jobs:
       - name: Build Changelog
         id: get_changelog
         run: |
+          if [[ "${{ github.ref_name }}" == "release/mainnet" ]]; then
+            DRY_RUN=$(npx commit-and-tag-version --dry-run)
+          else
```
ryle-ai: The "Get the release version" step above uses `elif` with an explicit branch check, which means an unexpected branch silently leaves `VERSION` empty and hits the `[ -n "$VERSION" ]` guard (loud failure). This `else` catch-all would silently apply `--prerelease rc` to any non-mainnet branch. The workflow trigger constrains this to two branches so it's functionally fine today, but matching the `elif` pattern would be more defensive and consistent:
```suggestion
          elif [[ "${{ github.ref_name }}" == "release/integrationnet" ]]; then
```

ryle-ai (commented):
Minor consistency nit on the branch-check pattern in Build Changelog -- see inline comment.

ryle-goehausen (approved):

PR #1466, opened 2026-03-10
fix: fixing L1 sync issues and consensus spam logs

marcus-girolneto:


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/ConsensusRoundRunner.scala
```diff
@@ -189,40 +194,65 @@ class ConsensusRoundRunner[F[_]: Async: Metrics, Event, Key: Next, Artifact, Ctx
...
+                  else if (reopened) now // Reset timer after unlock so re-stall waits reStallTimeout
+                  else ms.statusStartTime
                 statusDuration = now - newStatusStartTime
+                newLockedForStatus =
+                  if (statusChanged) false
+                  else if (reopened) false
+                  else ms.lockedForStatus
+                newStallCycleCount =
+                  if (statusChanged) 0
+                  else if (reopened) ms.stallCycleCount
```
ryle-ai: The `reopened` and `else` branches both evaluate to `ms.stallCycleCount`. Was the intent to reset or decrement on reopen? If the current behavior (preserve count in both cases) is intentional, the `reopened` branch is dead code and the expression can be simplified to `if (statusChanged) 0 else ms.stallCycleCount`.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/SnapshotStorage.scala
```diff
@@ -164,7 +164,9 @@ object SnapshotStorage {
                 notPersistedCache.update(current => current + snapshot.ordinal)
             )
           } >>
-          snapshotInfoLocalFileSystemStorage.write(snapshot.ordinal, snapshotInfo) >>
+          snapshotInfoLocalFileSystemStorage.write(snapshot.ordinal, snapshotInfo).handleErrorWith { e =>
```
ryle-ai: Swallowing this error makes the snapshotInfo write consistent with the snapshot binary write above (which already has `handleErrorWith`), but snapshotInfo is critical for rollback: `GlobalSnapshotTraverse` calls `loadInfoOrErr` and will throw if the file is missing. A disk error here means the node keeps running with an in-memory head that has no persisted snapshotInfo, so any future rollback through that ordinal will fail. Might be worth at least tracking these occurrences with a metric so operators can detect the gap before a rollback is needed.

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/programs/Rollback.scala
```diff
@@ -112,6 +113,9 @@ object Rollback {
         .hasCollateral(nodeId)
         .flatMap(OwnCollateralNotSatisfied.raiseError[F, Unit].unlessA)
 
+      _ <- snapshotStorage.prepend(lastIncremental, lastInfo)
```
ryle-ai: `prepend` returns `F[Boolean]` and the return value is discarded here. In `Download.scala` the same call checks the result and raises an error on `false`; in `StoragesInitializer` it logs. If the snapshot storage head is at a different ordinal (e.g. storage wasn't cleared before rollback), `prepend` will silently return `false` and `loadChain` will run against a storage that doesn't have this snapshot as head. Consider checking the return value or at minimum logging it, for parity with the other call sites.

ryle-ai (commented):
After `maxStallCycles` is exhausted, the monitor loop keeps spinning at ~1s intervals with locking suppressed (`alreadyLocked = ... || !withinStallBudget`). There's no outer round timeout or cancellation, so a round that truly can't make progress will poll indefinitely. In practice, facilitator-set changes or arriving acks usually break the deadlock, but a metric or log entry at the point where the stall budget is spent (distinct from the per-stall `dag_consensus_stall_detected` counter) would help operators notice rounds stuck in this state.

marcus-girolneto:
@greptileai please review and give the score again

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/SnapshotStorage.scala
```diff
@@ -164,12 +164,18 @@ object SnapshotStorage {
...
+            .write(snapshot.ordinal, snapshotInfo)
+            .flatMap { _ =>
+              snapshotInfoCutoffQueue.offer(snapshot.ordinal) >>
+                snapshot.ordinal
+                  .partialPreviousN(inMemoryCapacity)
+                  .fold(Applicative[F].unit)(offloadQueue.offer) >>
+                combinedSnapshotCheckpointFileSystemStorage.tryWrite(snapshot.ordinal, snapshot, snapshotInfo, hash)
+            }
+            .handleErrorWith { e =>
+              logger.error(e)(s"Failed writing snapshot info to disk! ordinal=${snapshot.ordinal}. Skipping cutoff and checkpoint.")
```
ryle-ai: This `handleErrorWith` catches errors from the entire `.write(...).flatMap { ... }` chain — so failures in `snapshotInfoCutoffQueue.offer`, `offloadQueue.offer`, or `combinedSnapshotCheckpointFileSystemStorage.tryWrite` (which does disk I/O via `writeJsonTupleStream` and doesn't handle its own errors) are all caught here. The log message attributes everything to "Failed writing snapshot info to disk" which will point operators at the wrong layer if the actual failure is in the checkpoint write or queue offer.

Consider either narrowing the `handleErrorWith` to just the `.write(...)` call (and handling downstream errors separately), or broadening the message to something like "Failed in snapshot info persistence pipeline".

ryle-ai (commented):
One observation on error handling scope in `SnapshotStorage.enqueue`; see inline comment.

scasplte2 (approved):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotContextFunctions.scala
```diff
@@ -323,13 +326,33 @@ object GlobalSnapshotContextFunctions {
             getGlobalSnapshotByOrdinal
           )
 
-        _ <- CannotApplyBlocksError(acceptanceResult.notAccepted.map { case (_, reason) => reason })
-          .raiseError[F, Unit]
+        // For followers (currency-l0, dag-l1, currency-l1), we log warnings instead of raising errors
```
ryle-ai, line 329: The comment says "For followers (currency-l0, dag-l1, currency-l1)" but `GlobalSnapshotContextFunctions.createContext` is in `node-shared` and is also called by dag-l0 during download (`dag-l0/domain/snapshot/programs/Download.scala:231`) and rollback (`GlobalSnapshotTraverse.scala:153`). The rationale still holds there — replayed snapshots were already validated by consensus — but on dag-l0 download the `.handleErrorWith` at line 238 that used to catch acceptance errors and convert them to `InvalidChain` (triggering a peer retry) will no longer fire, since `createContext` now completes successfully with a warning. This means a dag-l0 node downloading from peers will silently accept divergent block/reward acceptance results instead of retrying. Worth updating the comment to acknowledge the full scope.

modules/dag-l0/src/main/resources/dag-l0.conf
```diff
@@ -12,7 +12,9 @@ snapshot {
     declaration-timeout = 45 seconds
     declaration-range-limit = 3
     lock-duration = 10 seconds
-    max-facilitator-count = 20
+    max-facilitator-count = 300
```
ryle-ai, line 15: 15x increase from 20 to 300 — is this intentional as a workaround for the stall issues (effectively making all peers facilitators), or a permanent change? The architecture plan (Priority 7) targets reducing N² overhead via facilitator subset selection, so this trades that off. If it's a temporary measure to stabilize while the stall detection improvements land, a comment or tracking issue would help make that clear.

ryle-ai (commented):
Two notes on the existing bot comments: Greptile's concern about hash validation being outside the retry scope in `fetchAndSetCalculatedState` is a false positive — the `flatTap` hash check is inside the `retryingOnAllErrors { ... }` block, so hash mismatches are retried with a new peer. Similarly, the concern about `setCalculatedState` returning `false` silently collapsing to `None` in `DataApplicationTraverse` is incorrect — the new code raises an error on `false` at line 323.

See inline comments for two observations.

modules/dag-l0/src/main/resources/dag-l0.conf
marcus-girolneto: Temporary

PR #1467, opened 2026-03-11

marcus-girolneto:

Key changes:
- Safe quorum with supermajority gate in ConsensusStateAdvancer
- Adaptive tiered unlock thresholds (DEFER/DEGRADED/FALLBACK/NORMAL)
- Eligible facilitator re-entry after removal (no permanent exclusion)
- Deterministic multi-round removal penalty via consensus outcome
- Per-round fiber lifecycle tracking to prevent orphaned fibers
- Atomic PendingTriggers to fix race condition losing triggers
- Command stream error recovery for critical round-completion commands

### DIFF
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/ConsensusEventLoop.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/ConsensusEventLoop.scala
@@ -129,18 +129,20 @@ object ConsensusEventLoop {
 
       val commandStream: Stream[F, Unit] =
         Stream.repeatEval(queue.take).evalMap { cmd =>
-          fsm.handle(cmd).handleErrorWith { err =>
-            ctx.logger.error(err)(s"Unhandled error processing ${cmd.getClass.getSimpleName}, recovering") >>
-              Metrics[F].incrementCounter("dag_consensus_command_error") >>
-              (cmd match {
-                case _: ConsensusCommand.ConsensusFinished | ConsensusCommand.RoundCompleted =>
-                  // Critical: if round-completion commands fail, FSM stays stuck in BUSY forever.
-                  // Force round completion so the next round can start.
-                  ctx.logger.warn("Forcing round completion after failed ConsensusFinished/RoundCompleted") >>
-                    queue.offer(ConsensusCommand.RoundCompleted)
-                case _ => Async[F].unit
-              })
-          }
+          queue.size.flatMap(sz => Metrics[F].updateGauge("dag_consensus_command_queue_size", sz)) >>
+            fsm.handle(cmd).handleErrorWith { err =>
+              ctx.logger.error(err)(s"Unhandled error processing ${cmd.getClass.getSimpleName}, recovering") >>
+                Metrics[F].incrementCounter("dag_consensus_command_error") >>
+                (cmd match {
+                  case _: ConsensusCommand.ConsensusFinished | ConsensusCommand.RoundCompleted =>
+                    // Critical: if round-completion commands fail, FSM stays stuck in BUSY forever.
+                    // Force round completion so the next round can start.
+                    ctx.logger.warn("Forcing round completion after failed ConsensusFinished/RoundCompleted") >>
+                      Metrics[F].incrementCounter("dag_consensus_forced_round_completion") >>
+                      queue.offer(ConsensusCommand.RoundCompleted)
+                  case _ => Async[F].unit
+                })
+            }
         }
 
       val peerRegistrationStream: Stream[F, Unit] =
@@ -165,11 +167,14 @@ object ConsensusEventLoop {
       BuiltConsensusLoop(run, manager, queue)
     }
 
-  private def collectRegistration[F[_]: Async, Event, Key, Artifact, Ctx, Status, Outcome, Kind](
+  private def collectRegistration[F[_]: Async: Metrics, Event, Key, Artifact, Ctx, Status, Outcome, Kind](
     consensusClient: ConsensusClient[F, Key, Outcome],
     storage: ConsensusStorage[F, Event, Key, Artifact, Ctx, Status, Outcome, Kind]
   )(peer: Peer): F[Unit] =
     consensusClient.getRegistration.run(peer).flatMap { reg =>
-      reg.maybeKey.traverse_(key => storage.registerPeer(peer.id, key))
+      reg.maybeKey.traverse_(key =>
+        storage.registerPeer(peer.id, key) >>
+          Metrics[F].incrementCounter("dag_consensus_peer_registered")
+      )
     }
 }
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/ConsensusEventLoop.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/ConsensusEventLoop.scala
@@ -1,7 +1,8 @@
 package io.constellationnetwork.node.shared.infrastructure.consensus.engine
 
 import cats.Show
-import cats.effect.kernel.Async
+import cats.effect.Fiber
+import cats.effect.kernel.{Async, Deferred, Ref}
 import cats.effect.std.{Queue, Random, Supervisor}
 import cats.kernel.{Eq, Next}
 import cats.syntax.all._
@@ -19,6 +20,7 @@ import io.constellationnetwork.schema.peer.Peer
 import io.constellationnetwork.security.HasherSelector
 import io.constellationnetwork.security.signature.Signed
 
+import eu.timepit.refined.auto._
 import fs2.Stream
 
 /** Builds and wires together all consensus engine components.
@@ -68,7 +70,7 @@ object ConsensusEventLoop {
     Key: Eq: Show: Next,
     Artifact: Eq,
     Ctx: Eq,
-    Status: Eq,
+    Status,
     Outcome,
     Kind
   ](
@@ -108,7 +110,15 @@ object ConsensusEventLoop {
         consensusFunctions,
         consensusClient
       )
-      roundRunner = new ConsensusRoundRunner[F, Event, Key, Artifact, Ctx, Status, Outcome, Kind](ctx)
+      stallDetector = new StallDetector[F, Event, Key, Artifact, Ctx, Status, Outcome, Kind](ctx)
+      roundFibersRef <- Ref.of[F, List[Fiber[F, Throwable, Unit]]](Nil)
+      cancelSignalRef <- Ref.of[F, Option[Deferred[F, Unit]]](None)
+      roundRunner = new ConsensusRoundRunner[F, Event, Key, Artifact, Ctx, Status, Outcome, Kind](
+        ctx,
+        stallDetector,
+        roundFibersRef,
+        cancelSignalRef
+      )
       fsm = new ConsensusFSM[F, Event, Key, Artifact, Ctx, Status, Outcome, Kind](ctx, roundRunner)
       manager <- ConsensusManager.make[F, Event, Key, Artifact, Ctx, Status, Outcome, Kind](
         queue,
@@ -118,7 +128,20 @@ object ConsensusEventLoop {
     } yield {
 
       val commandStream: Stream[F, Unit] =
-        Stream.repeatEval(queue.take).evalMap(fsm.handle)
+        Stream.repeatEval(queue.take).evalMap { cmd =>
+          fsm.handle(cmd).handleErrorWith { err =>
+            ctx.logger.error(err)(s"Unhandled error processing ${cmd.getClass.getSimpleName}, recovering") >>
+              Metrics[F].incrementCounter("dag_consensus_command_error") >>
+              (cmd match {
+                case _: ConsensusCommand.ConsensusFinished | ConsensusCommand.RoundCompleted =>
+                  // Critical: if round-completion commands fail, FSM stays stuck in BUSY forever.
+                  // Force round completion so the next round can start.
+                  ctx.logger.warn("Forcing round completion after failed ConsensusFinished/RoundCompleted") >>
+                    queue.offer(ConsensusCommand.RoundCompleted)
+                case _ => Async[F].unit
+              })
+          }
+        }
 
       val peerRegistrationStream: Stream[F, Unit] =
         clusterStorage.peerChanges.mapFilter {
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/StallDetector.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/StallDetector.scala
@@ -1,0 +1,306 @@
+package io.constellationnetwork.node.shared.infrastructure.consensus.engine
+
+import cats.effect.kernel.{Async, Deferred, Temporal}
+import cats.syntax.all._
+
+import scala.concurrent.duration._
+
+import io.constellationnetwork.node.shared.infrastructure.consensus.ConsensusResources
+import io.constellationnetwork.node.shared.infrastructure.consensus.state._
+import io.constellationnetwork.node.shared.infrastructure.metrics.Metrics
+
+import eu.timepit.refined.auto._
+
+/** Monitors a consensus round for stalls and manages the lock/unlock lifecycle.
+  *
+  * ==Stall Detection Flow==
+  * {{{
+  *   Wait declarationTimeout
+  *     → Lock (Closed) + spread ACKs
+  *     → Wait for unlock (Reopened) via ACK voting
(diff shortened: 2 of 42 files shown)


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/update/UnlockConsensusUpdate.scala
```diff
@@ -55,19 +59,34 @@ object UnlockConsensusUpdate {
...
-                else if (votesRemove >= removeThreshold)
-                  (peerId, false).some
-                else
-                  none
-            }
+          val degradedQuorum = math.ceil(n.toDouble / 5).toInt.max(2)
+          val standardQuorum = math.ceil(n.toDouble / 3).toInt.max(2)
+          val normalKeepThreshold = (n + 1) / 2
+
+          val maybeThresholds: Option[(Int, Int)] = {
```
ryle-ai, line 69: The thresholds in `maybeThresholds` are a function of `voterCount`, which is a locally-observed value (how many ACKs this node has received). Since ACKs propagate asynchronously, two nodes can see different `voterCount` values at the moment they evaluate this block, landing in different tiers with different thresholds. This can cause one node to successfully unlock (removing some peers) while another stays locked or makes a different removal decision.

The discontinuity at the Tier 2→3 boundary is especially sharp: at n=20, crossing from `voterCount=9` to `voterCount=10` causes `keepThreshold` to jump from 5 to 10 and `removeThreshold` from 5 to 11. One late-arriving ACK can push a node from "can decide" to "can't decide."

The pre-PR thresholds (`(n+1)/2` and `n/2+1`) are deterministic because they depend only on `n`, which all nodes agree on. If the goal is to allow progress with fewer voters, consider lowering the fixed thresholds (e.g., based on `n` fractions) rather than making them depend on the locally-variable `voterCount`.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/StallDetector.scala
```diff
@@ -0,0 +1,306 @@
...
+              newStatusStartTime =
+                if (statusChanged) now
+                else if (reopened) now
+                else ms.statusStartTime
+              statusDuration = now - newStatusStartTime
+              newLockedForStatus =
+                if (statusChanged) false
+                else if (reopened) false
+                else ms.lockedForStatus
+              newStallCycleCount = if (statusChanged || reopened) 0 else ms.stallCycleCount
```
ryle-ai: Behavioral change from the old `ConsensusRoundRunner`: the stall cycle count now resets to 0 on `reopened` (successful unlock), not just on `statusChanged`. The old code: `if (statusChanged) 0 else ms.stallCycleCount`. This means the `maxStallCycles` budget refreshes after every successful unlock — a round can cycle through stall→lock→unlock→stall indefinitely as long as each unlock succeeds. The `maxRoundDuration` catches this eventually, but the effective stall budget is now unbounded within the 5-minute window. Was this intentional?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/ConsensusEventLoop.scala
```diff
@@ -118,7 +128,20 @@ object ConsensusEventLoop {
...
+        Stream.repeatEval(queue.take).evalMap { cmd =>
+          fsm.handle(cmd).handleErrorWith { err =>
+            ctx.logger.error(err)(s"Unhandled error processing ${cmd.getClass.getSimpleName}, recovering") >>
+              Metrics[F].incrementCounter("dag_consensus_command_error") >>
+              (cmd match {
+                case _: ConsensusCommand.ConsensusFinished | ConsensusCommand.RoundCompleted =>
+                  // Critical: if round-completion commands fail, FSM stays stuck in BUSY forever.
+                  // Force round completion so the next round can start.
+                  ctx.logger.warn("Forcing round completion after failed ConsensusFinished/RoundCompleted") >>
+                    queue.offer(ConsensusCommand.RoundCompleted)
```
ryle-ai: If `ConsensusFinished` fails partway through (e.g., `afterConsensusFinish` throws after partial execution but before the outcome is stored), re-queuing `RoundCompleted` will run `completeRound` with a no-op preAction — the round gets "completed" but the consensus outcome is silently lost. The next round starts from the previous outcome.

This is still better than the pre-PR behavior (unhandled error kills the command stream), but worth noting that transient failures in `afterConsensusFinish` can cause a completed consensus round to be discarded.

ryle-ai (commented):
## Tiered unlock thresholds depend on locally-observed `voterCount`

The adaptive thresholds in `UnlockConsensusUpdate` are computed from `voterCount` — how many facilitators this specific node has received ACKs from. Since ACKs arrive asynchronously via gossip, different nodes can observe different `voterCount` values at the moment they run the unlock check. Because the thresholds change based on which tier `voterCount` falls into, two nodes can make different keep/remove decisions about the same peer.

Concrete example with n=20: Node A sees `voterCount=9` (Tier 2, removeThresh=5). Node B sees `voterCount=10` (Tier 3, removeThresh=11). For a peer with 5 remove votes among the first 9 voters: Node A removes the peer, Node B can't decide. Node A unlocks with a different facilitator set than Node B. The new `checkForkByFacilitatorsHash` would catch this divergence and trigger fork recovery (restart), which is better than silent corruption — but repeated fork-and-restart cycles from threshold boundary disagreements would be disruptive.

The Tier 2→3 boundary has a severe discontinuity: at n=20, `keepThreshold` jumps from 5 to 10 when gaining one additional voter. This means more participation can paradoxically make decisions *harder*, which is the opposite of the desired property.

The pre-PR code avoids this by using fixed thresholds based only on `n` (which all nodes agree on). Consider keeping thresholds deterministic — e.g., lowering the fixed thresholds based on `n` rather than making them a function of the locally-observed `voterCount`.

## Stall cycle budget reset on Reopened

The stall cycle count now resets to 0 on successful unlock (`Reopened`), where previously it only reset on status change. Combined with `maxStallCycles` increasing from 3 to 5, a flapping peer could cause unlimited stall-unlock-stall cycles as long as each unlock succeeds. The `maxRoundDuration` safety net (5 min) bounds this, but rounds could churn for minutes before abandonment.

## Quorum safety gate

The two-part check in `maybeGetQuorumDeclarations` (receivedCount >= quorumSize AND maxSupport >= quorumSize) is sound. By pigeonhole, two different values can't both have >= quorumSize support when quorumSize > N/2, so `pickMajority` is guaranteed to produce the same result regardless of which quorum-sized subset each node sees. The value extractors (`_.trigger`, `_.hash`, `_.facilitatorsHash`) correctly gate on the fields that downstream `pickMajority` operates on.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/update/UnlockConsensusUpdate.scala
```diff
@@ -55,19 +59,34 @@ object UnlockConsensusUpdate {
                 .getOrElse(acc)
             }
 
-          val keepThreshold = (facilitators.size + 1) / 2
-          val removeThreshold = facilitators.size / 2 + 1
+          val n = facilitators.size
+          val voterCount = facilitators.count(f => acksMap.contains((f, collectingKind)))
```
ryle-ai, line 63: `voterCount` is determined by how many ACKs this node has locally received, which varies across nodes due to async propagation. Two nodes evaluating the same round can see different `voterCount` values and land in different tiers.

Concrete example at n=20: Node A sees `voterCount=9` (DEGRADED tier, `superThreshold = ceil(9*2/3) = 6`, so keepThresh=6, removeThresh=6). Node B sees `voterCount=10` (FALLBACK tier, keepThresh=`(10+1)/2=5`, removeThresh=`10/2+1=6`). A peer with 5 keep votes: Node A says indeterminate (5 < 6), Node B says keep (5 >= 5). They diverge on whether to remove that peer.

The pre-PR thresholds (`(n+1)/2` and `n/2+1`) are deterministic because `n` comes from the consensus state that all nodes agree on. Consider keeping thresholds as a function of `n` only — e.g., lowering fixed thresholds when unlock is attempted (which already implies reduced participation) rather than making them depend on the locally-variable `voterCount`.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/ConsensusEventLoop.scala
ryle-ai: If `ConsensusFinished` fails partway through `afterConsensusFinish` (e.g., persistence error after partial execution), this re-enqueues `RoundCompleted`. The FSM handles `RoundCompleted` via `completeRound(log.debug(...))` — a no-op preAction that succeeds — so `isRunning` gets cleared and the round completes, but the original consensus outcome is silently lost. The next round restarts from the previous outcome.

This is an improvement over no error handling (which would kill the command stream), but consider logging at `error` level when a `ConsensusFinished` falls through to this path, since it means a completed consensus round's result was discarded. The current `warn` level doesn't distinguish between a benign `RoundCompleted` retry and a `ConsensusFinished` outcome loss.

ryle-ai (commented):
## Tiered unlock thresholds depend on locally-observed state

The adaptive thresholds in `UnlockConsensusUpdate` compute keep/remove decisions from `voterCount` — the number of ACKs this specific node has received. Since ACKs propagate asynchronously, different nodes can observe different `voterCount` values, land in different tiers, and make divergent removal decisions about the same peer.

The December 2025 architecture plan explicitly identifies this class of problem: node removal decisions are "fundamentally related to the non-reproducible behavior of nodes" and fixing it "would require a solution similar to AllowSpend, i.e. a data type which is committed on-chain." The tiered thresholds make removal *more* dependent on local observation (voterCount) than the pre-PR deterministic thresholds (based only on `n`, which all nodes agree on). This moves further from the stated architectural direction.

The `checkForkByFacilitatorsHash` safety net catches the resulting divergence and triggers fork recovery (restart), which prevents silent corruption — but repeated fork-and-restart cycles during periods of uneven ACK propagation would be disruptive. The tier 2→3 boundary discontinuity (at n=20, one additional voter jumps `keepThreshold` from 5 to 10) makes this especially sensitive.

## Error recovery can silently lose a consensus outcome

The Greptile concern about an infinite retry loop in the command stream error handler is overstated. When `ConsensusFinished` fails partway through `afterConsensusFinish`, the re-enqueued `RoundCompleted` enters `completeRound` with a simple `log.debug` preAction (not the original `afterConsensusFinish`), which succeeds and breaks the loop. `cleanupRound` is idempotent — after the first call, subsequent calls are no-ops.

The real concern is **outcome loss**: if `afterConsensusFinish` throws after the consensus completed but before the result is persisted, the error recovery completes the round *without* the outcome. The next round starts from the previous outcome, silently discarding a completed consensus round's work. This is still better than the pre-PR behavior (unhandled error kills the command stream entirely), but worth awareness.

## UnlockConsensusUpdate test coverage

`UnlockConsensusUpdateSuite` has only 2 test cases — far too few for what is arguably the highest-risk behavioral change in this PR. There's no coverage of tier boundary behavior, threshold discontinuities, or the scenario where different `voterCount` values produce different keep/remove decisions for the same peer. Given the non-determinism concern above, this is where property-based tests would provide the most value.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/StallDetector.scala
```diff
@@ -0,0 +1,312 @@
...
+              newRoundHadStall = ms.roundHadStall || didLock
+
+              _ <- Metrics[F].updateGauge("dag_consensus_stall_cycle", finalStallCycleCount)
+              _ <- Metrics[F].updateGauge("dag_consensus_stall_declaration_progress", declarationProgress)
+              _ <- Metrics[F].incrementCounter("dag_consensus_stall_restall").whenA(failedStallCycle)
+
+              roundElapsed = now - ms.roundStartTime
+              _ <- Metrics[F].updateGauge("dag_consensus_round_elapsed_seconds", roundElapsed.toSeconds.toInt)
+              roundTimedOut = config.maxRoundDuration.exists(roundElapsed >= _)
+              shouldAbandon = finalStallCycleCount >= config.maxStallCycles || roundTimedOut
```
ryle-ai, line 158: This drops the `&& isLocked` guard from the old `ConsensusRoundRunner` formula:

```
// Old: shouldAbandon = finalStallCycleCount >= config.maxStallCycles && isLocked
// New: shouldAbandon = finalStallCycleCount >= config.maxStallCycles || roundTimedOut
```

Without `&& isLocked`, a round that repeatedly stalls and unlocks (Closed → Reopened cycles) will be abandoned once it hits `maxStallCycles`, even if it's actively making progress in Reopened state. The `StallDetectorSuite` tests the old formula with `&& isLocked` (lines 247, 262, 277, 292), so either the test or this line is wrong.

If dropping `&& isLocked` is intentional ("any round that stalls 5 times is unhealthy regardless"), update the tests to match. If not, add `&& isLocked` back.

modules/node-shared/src/test/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/StallDetectorSuite.scala
```diff
@@ -0,0 +1,495 @@
...
+  test("should NOT abandon when stallCycles exceeded but NOT locked") {
+    IO {
+      val maxRoundDuration = Some(5.minutes)
+      val roundElapsed = 2.minutes
+      val stallCycleCount = 5
+      val maxStallCycles = 5
+      val isLocked = false
+
+      val roundTimedOut = maxRoundDuration.exists(roundElapsed >= _)
+      val shouldAbandon = (stallCycleCount >= maxStallCycles && isLocked) || roundTimedOut
```
ryle-ai, line 277: This test validates a formula that differs from the actual `StallDetector` implementation:

```scala
// Test formula (here):
val shouldAbandon = (stallCycleCount >= maxStallCycles && isLocked) || roundTimedOut

// Actual StallDetector (line 158):
shouldAbandon = finalStallCycleCount >= config.maxStallCycles || roundTimedOut
```

The `&& isLocked` condition exists in the test but not in the implementation. This test asserts "should NOT abandon when cycles exceeded but not locked" — which matches the *old* behavior but not the *new* code. The test passes because it's self-contained (tests its own formula), not because the StallDetector actually behaves this way.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/update/UnlockConsensusUpdate.scala
ryle-ai, line 63: `voterCount` is locally observed (how many ACKs this node received). Two nodes evaluating the same round can see different `voterCount` values and land in different tiers with different thresholds. At `n=20`, one additional ACK can push a node from DEGRADED (`superThreshold=6`) to FALLBACK (`keepThreshold=5`), producing a different keep/remove decision for the same peer.

The `checkForkByFacilitatorsHash` safety net catches the resulting divergence, but the December 2025 architecture docs note that node removal "would require a solution similar to AllowSpend, i.e. a data type which is committed on-chain" for determinism. Making thresholds depend on locally-variable `voterCount` moves further from that direction.

This isn't necessarily blocking — the fork recovery catches it — but it's worth awareness that the tier boundaries create sharp discontinuities where more participation can paradoxically make decisions harder.

ryle-ai (changes requested):
## `shouldAbandon` formula disagrees with its own tests

The `StallDetector` implementation computes:
```scala
shouldAbandon = finalStallCycleCount >= config.maxStallCycles || roundTimedOut
```

But `StallDetectorSuite` tests a different formula:
```scala
val shouldAbandon = (stallCycleCount >= maxStallCycles && isLocked) || roundTimedOut
```

The test preserves the old code's `&& isLocked` guard. The implementation drops it. These produce different results: when `maxStallCycles` is reached but the round is in `Reopened` state (successfully unlocked), the implementation abandons while the test expects it not to.

The test passes because it validates its own reimplemented logic, not the actual `StallDetector`. This gives false confidence about the abandon policy. Either the test or the implementation needs to change — please clarify which formula is intended.

Related: when `shouldAbandon` triggers while not in `Closed` state, `condModifyState` (line 171) is a no-op because it only removes `Closed` state. The `RoundCompleted` still fires so it's not a deadlock, but the abandoned key's state remains in storage until cleaned up by the next round's `tryUpdateLastConsensusOutcomeWithCleanup`.

## Bot review correction

The ryle-ai and greptile reviews both claim `stallCycleCount` now resets on `Reopened`. It doesn't — line 105 of `StallDetector.scala` shows `if (statusChanged) 0 else ms.stallCycleCount` with no `|| reopened` condition. The bots cited code that doesn't match the diff. The actual behavior is that stall cycles accumulate across unlock cycles within the same phase, which is consistent with the `shouldAbandon` formula (if the `&& isLocked` removal is intentional).

PR #1469, opened 2026-03-11

marcus-girolneto:

Key changes:

Safe quorum with supermajority gate in ConsensusStateAdvancer
Adaptive tiered unlock thresholds (DEFER/DEGRADED/FALLBACK/NORMAL)
Eligible facilitator re-entry after removal (no permanent exclusion)
Deterministic multi-round removal penalty via consensus outcome
Per-round fiber lifecycle tracking to prevent orphaned fibers
Atomic PendingTriggers to fix race condition losing triggers

### DIFF
--- a/docs/CONSENSUS_REFACTORING.md
+++ b/docs/CONSENSUS_REFACTORING.md
@@ -65,7 +65,7 @@ snapshot.consensus {
 
 ---
 
-## Adaptive Tiered Unlock Thresholds
+## Deterministic N-Based Unlock Thresholds
 
 ### Problem
 
@@ -75,28 +75,29 @@ The old unlock logic used flat thresholds based on total facilitator count. When
 
 ### Solution
 
-Replaced flat thresholds with **adaptive tiers** in `UnlockConsensusUpdate.scala` based on how many facilitators actually voted (sent ACKs):
+Replaced flat thresholds with **deterministic N-based thresholds** in `UnlockConsensusUpdate.scala`, with a minimum voter gate and a safety floor:
 
-| Tier | Condition | Keep Threshold | Remove Threshold | Rationale |
-|------|-----------|----------------|------------------|-----------|
-| **DEFER** | `voterCount < ceil(N/5)` | — | — | Too few voters; defer to re-stall cycle |
-| **DEGRADED** | `voterCount < ceil(N/3)` | 2/3 supermajority of voters | 2/3 supermajority of voters | Conservative: high bar prevents bad decisions |
-| **FALLBACK** | `voterCount < (N+1)/2` | Simple majority of voters | Simple majority of voters | Moderate: enough voters for reasonable decision |
-| **NORMAL** | `voterCount >= (N+1)/2` | `(N+1)/2` of total | `N/2 + 1` of total | Standard: based on total facilitator count |
+| Phase | Condition | Keep Threshold | Remove Threshold | Rationale |
+|-------|-----------|----------------|------------------|-----------|
+| **DEFER** | `voterCount < ceil(N/3)` | — | — | Too few voters for a safe decision; defer to re-stall cycle |
+| **DECIDE** | `voterCount >= ceil(N/3)` | `(N+1)/2` | `N/2 + 1` | N-based thresholds ensure all nodes make identical decisions |
 
-Where `N` is the total facilitator count and `voterCount` is how many facilitators actually sent ACKs for the current collecting kind.
+Where `N` is the total facilitator count and `voterCount` is how many facilitators actually sent ACKs for the current collecting kind. The minimum voter requirement is `ceil(N/3).max(2)`.
 
-**Key behavior:**
-- DEFER tier means no unlock happens — the re-stall mechanism retries with fresh ACKs, giving more time for ACKs to arrive
-- DEGRADED requires supermajority agreement among actual voters before removing anyone
-- As more voters participate, thresholds relax toward the standard case
+**Key properties:**
+- `keepThreshold + removeThreshold = N + 1 > N`, guaranteeing mutual exclusivity (no peer can be both kept and removed)
+- Thresholds depend only on `N` (total facilitator count), which all nodes agree on — making the decision deterministic across the network
+- DEFER uses `ceil(N/3)` as its minimum voter requirement — a lower bar than `keepThreshold` (`(N+1)/2`), allowing unlock decisions with fewer voters while still requiring meaningful participation. For N=20 this means 7 voters needed (vs 11 with keepThreshold)
+- The stall detector's `maxStallCycles` and `maxRoundDuration` handle liveness: after exhausting stall cycles, the round is abandoned and a new one starts
 - After unlock: `Closed → Reopened`, removed peers are dropped from facilitator list, `advanceStatus` runs again with reduced list
+
+**Safety floor (`MinFacilitatorCount = 2`):** After computing keep/remove decisions, if the number of kept facilitators would fall below 2, the unlock is aborted — the state remains `Closed` and the round is deferred to the stall detector. This prevents a catastrophic scenario where stale ACKs (e.g., from a global latency spike where no declarations arrived before lock) cause all facilitators to be voted out, leaving `facilitatorCount=0` and an irrecoverable cluster.
 
 ### Files Changed
 
 | File | Change |
 |------|--------|
-| `consensus/update/UnlockConsensusUpdate.scala` | Rewritten with 4-tier adaptive thresholds |
+| `consensus/update/UnlockConsensusUpdate.scala` | Rewritten with deterministic N-based thresholds, DEFER gate, and MinFacilitatorCount safety floor |
 
 ---
 
@@ -248,7 +249,7 @@ Wait declarationTimeout
 **Adaptive timeout features:**
 - **Near-completion grace:** When >=75% of declarations received and on first stall cycle, timeout extended by 50% to allow stragglers
 - **Re-stall timeout:** After first lock, subsequent stall cycles use `reStallTimeout` (shorter) instead of `declarationTimeout`
-- **Round abandonment:** Cleans up state and enqueues `RoundCompleted` so the FSM transitions back to IDLE
+- **Round abandonment:** `shouldAbandon = finalStallCycleCount >= maxStallCycles || roundTimedOut`. When triggered, state is removed unconditionally (`case Some(_) =>`) and `RoundCompleted` is enqueued so the FSM transitions back to IDLE. Note: the stall-cycle-exhaustion path does not require `isLocked` — once the budget is spent, the round is abandoned regardless of lock status. This is a deliberate trade-off: after N failed lock/unlock cycles, cutting losses is faster than hoping N+1 works
 
 **Monitoring improvements:**
 - Periodic summary logging (every 10s) showing status, declaration counts, missing peers
@@ -458,15 +459,21 @@ fsm.handle(cmd).handleErrorWith { err =>
     Metrics[F].incrementCounter("dag_consensus_command_error") >>
     (cmd match {
       case _: ConsensusCommand.ConsensusFinished | ConsensusCommand.RoundCompleted =>
-        // Critical: force round completion so FSM doesn't stay stuck in BUSY
+        // Critical: force round completion so FSM doesn't stay stuck in BUSY.
+        // Also offer TimeTick: the forced RoundCompleted calls completeRound without
+        // afterConsensusFinish, so no timer is scheduled for the next round. On solo nodes
+        // with no external events, this would deadlock consensus.
         logger.warn("Forcing round completion after failed ConsensusFinished/RoundCompleted") >>
-          queue.offer(ConsensusCommand.RoundCompleted)
+          queue.offer(ConsensusCommand.RoundCompleted) >>
+          queue.offer(ConsensusCommand.TimeTick)
       case _ => Async[F].unit
     })
 }
 ```
 
 For non-critical commands (rumor processing, peer registration, etc.), the error is logged and the stream continues — same as before.
+
+**Solo-node stall prevention:** The forced `RoundCompleted` calls `completeRound` without `afterConsensusFinish`, so no timer is scheduled for the next round. On solo nodes with no external events (no gossip, no peer messages), consensus would permanently stall. The follow-up `TimeTick` fires once `RoundCompleted` sets `isRunning=false`, starting a new round from IDLE.
 
 ### Files Changed
 
@@ -502,11 +509,11 @@ These files had unused imports, parameters, or methods that were never caught be
 ## Test Results
 
 ```
-Total: 71 tests, 0 failures
-  - StallDetectorSuite:            24 tests  (stall detection, lock/unlock, re-stall cycles, abandonment)
-  - QuorumDeclarationsSuite:       12 tests  (quorum threshold, supermajority gate, split vote deferral)
+Total: 83 tests, 0 failures (430 total project-wide)
+  - StallDetectorSuite:            27 tests  (stall detection, lock/unlock, re-stall cycles, abandonment, abandon guard)
+  - QuorumDeclarationsSuite:       21 tests  (quorum threshold, supermajority gate, split vote deferral, config validation, small cluster)
   - RemovalPenaltySuite:           12 tests  (penalty lifecycle, decrement, expiry, filtering, determinism)
-  - UnlockConsensusUpdateSuite:     9 tests  (adaptive tiers: DEFER, DEGRADED, FALLBACK, NORMAL)
+  - UnlockConsensusUpdateSuite:    10 tests  (N-based thresholds, DEFER gate, mutual exclusivity, determinism, MinFacilitatorCount safety floor)
   - PendingTriggersSuite:           7 tests  (atomic state, priority ordering, concurrency safety)
   - EligibleFacilitatorsSuite:      5 tests  (re-entry after removal, collateral filtering)
 ```
--- a/docs/CONSENSUS_REFACTORING.md
+++ b/docs/CONSENSUS_REFACTORING.md
@@ -1,0 +1,528 @@
+# Consensus FSM Refactoring — Summary of Changes
+
+## Overview
+
+This refactoring targets seven categories of issues in the consensus engine:
+
+1. **Quorum safety** — Quorum declarations lacked a supermajority gate, risking non-deterministic decisions across different node views
+2. **Stall recovery** — Unlock voting used flat thresholds that couldn't adapt when many facilitators were unresponsive
+3. **Facilitator re-entry** — Peers removed during stall recovery were permanently excluded from future rounds
+4. **Removal penalty** — Removed peers could be re-selected one round later, causing repeated stalls when still unresponsive
+5. **Fiber leaks** — Orphaned fibers accumulating over sustained operation
+6. **Race condition** — Non-atomic state in `PendingTriggers` causing lost triggers
+7. **Error recovery** — FSM getting permanently stuck when critical commands fail
+
+All changes are backward-compatible. No protocol changes, no config format changes, no public API changes.
+
+---
+
+## Safe Quorum with Supermajority Gate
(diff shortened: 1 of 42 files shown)


docs/CONSENSUS_REFACTORING.md
```diff
@@ -0,0 +1,528 @@
...
+
+The old unlock logic used flat thresholds based on total facilitator count. When >50% of facilitators were unresponsive (couldn't even send ACKs), the thresholds could never be met. The round stayed permanently locked — a deadlock.
+
+### Solution
+
+Replaced flat thresholds with **adaptive tiers** in `UnlockConsensusUpdate.scala` based on how many facilitators actually voted (sent ACKs):
+
+| Tier | Condition | Keep Threshold | Remove Threshold | Rationale |
+|------|-----------|----------------|------------------|-----------|
+| **DEFER** | `voterCount < ceil(N/5)` | — | — | Too few voters; defer to re-stall cycle |
```
ryle-ai: This 4-tier table doesn't match the implementation. `UnlockConsensusUpdate` only has 2 tiers:

- **DEFER**: `voterCount < ceil(N/3)` — no unlock
- **NORMAL**: `keepThreshold = (N+1)/2`, `removeThreshold = N/2+1` — N-based

The DEGRADED tier (voter-proportional 2/3 supermajority) and FALLBACK tier (voter-proportional simple majority) don't exist in the code. The DEFER boundary is also different: `ceil(N/3)` in code vs `ceil(N/5)` here.

Either update the doc to match the 2-tier implementation, or implement the intermediate tiers if they're intended.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/config/types.scala
```diff
@@ -148,8 +148,27 @@ object types {
...
+    /** Deterministic hash of consensus-critical config values.
+      *
+      * All nodes in a consensus round MUST have the same config to produce the same results. This hash is included in Facility declarations
+      * so that config divergence is detected immediately during the CollectingFacilities phase, rather than causing mysterious forks
+      * downstream.
+      */
+    lazy val deterministicConfigHash: io.constellationnetwork.security.hash.Hash = {
+      val configString =
+        s"maxFacilitatorCount=${maxFacilitatorCount.map(_.value)}," +
+          s"maxStallCycles=$maxStallCycles," +
```
ryle-ai, line 186: This hash uses string interpolation with `Option.toString`, producing strings like `maxFacilitatorCount=None` or `maxFacilitatorCount=Some(20)`. While stable in practice (all nodes run the same Scala/JVM), it has two fragility vectors:

1. Adding a new consensus-critical config field requires remembering to update this string. If someone adds a field to `ConsensusConfig` that affects deterministic behavior but forgets to include it here, nodes with different values for that field will silently diverge.

2. The string format is an implicit contract across all deployed nodes. A Scala version upgrade that changes `Option.toString` (unlikely but possible) would cause every node to compute a different hash during rolling upgrade, triggering mass fork detection.

Consider either (a) a comment listing which fields are consensus-critical and why others are excluded, or (b) deriving the hash from a structured representation (e.g., a tuple of the included values) rather than string interpolation.

ryle-ai (commented):
Solid refactoring of the consensus engine. The quorum/supermajority gate, PendingTriggers atomicity fix, fiber lifecycle management, and multi-round removal penalty are well-designed. Two items to address before merge:

1. **CONSENSUS_REFACTORING.md describes unlock tiers that don't exist in the code.** The doc shows a 4-tier adaptive threshold system (DEFER/DEGRADED/FALLBACK/NORMAL) with voter-count-based thresholds in the intermediate tiers. The actual `UnlockConsensusUpdate` implementation only has 2 tiers: DEFER (below `ceil(N/3)`) and NORMAL (N-based thresholds). The DEGRADED and FALLBACK tiers with voter-proportional thresholds are absent. The DEFER boundary also differs: doc says `ceil(N/5)`, code says `ceil(N/3)`. Since this doc is titled "Summary of Changes", it should describe what was actually implemented.

2. **`deterministicConfigHash` relies on `Option.toString` rendering.** See inline comment.

Minor observation on the abandon path: the old `condModifyState` guard (`lockStatus === Closed`) was removed in favor of unconditional state removal. This is fine for the `maxRoundDuration` safety net, and in practice `maxStallCycles` accumulates enough lock/unlock failures to justify removal regardless of current lock status. Just documenting that the safety semantics changed intentionally.

marcus-girolneto:
@greptileai, please review again and give the score

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/StallDetector.scala
```diff
@@ -0,0 +1,312 @@
...
+              newRoundHadStall = ms.roundHadStall || didLock
+
+              _ <- Metrics[F].updateGauge("dag_consensus_stall_cycle", finalStallCycleCount)
+              _ <- Metrics[F].updateGauge("dag_consensus_stall_declaration_progress", declarationProgress)
+              _ <- Metrics[F].incrementCounter("dag_consensus_stall_restall").whenA(failedStallCycle)
+
+              roundElapsed = now - ms.roundStartTime
+              _ <- Metrics[F].updateGauge("dag_consensus_round_elapsed_seconds", roundElapsed.toSeconds.toInt)
+              roundTimedOut = config.maxRoundDuration.exists(roundElapsed >= _)
+              shouldAbandon = finalStallCycleCount >= config.maxStallCycles || roundTimedOut
```
ryle-ai, line 158: The old code was `shouldAbandon = finalStallCycleCount >= config.maxStallCycles && isLocked`. Dropping the `&& isLocked` guard means: if `finalStallCycleCount` reaches `maxStallCycles` on tick N (while Closed), and between tick N and N+1 a concurrent unlock succeeds (`Closed → Reopened`), tick N+1 carries forward `ms.stallCycleCount >= maxStallCycles` without incrementing (since neither `didLock` nor `failedStallCycle` is true when Reopened), and `shouldAbandon` fires.

The `condModifyState` below then removes the Reopened state (line ~173: `case Some(_) =>`), discarding all collected declarations from the recovering round.

For the `roundTimedOut` branch, unconditional removal makes sense as a hard safety net. For the stall-cycle-exhaustion branch, you could preserve the old behavior with:

```scala
shouldAbandon = (finalStallCycleCount >= config.maxStallCycles && isLocked) || roundTimedOut
```

and restoring the Closed guard on `condModifyState`:

```scala
case Some(s) if roundTimedOut || s.lockStatus === LockStatus.Closed =>
```

This way `roundTimedOut` still kills anything (safety net), but the stall-cycle path gives a just-unlocked round one more chance to advance. Not blocking — the current approach is a defensible "cut losses early" choice, and the removal penalty system means the next round should do better. Just flagging the behavioral change.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/state/ConsensusStateUpdater.scala
```diff
@@ -182,25 +186,50 @@ object ConsensusStateUpdater {
...
+            Metrics[F].incrementCounter("dag_consensus_unlock_transition") >>
+              Metrics[F].updateGauge("dag_consensus_unlock_peers_removed", newlyRemoved.size) >>
+              logger.warn(
+                s"Unlock transition: Closed -> Reopened for key=${currentState.key.show}. " +
+                  s"Removed ${newlyRemoved.size} peers: ${newlyRemoved.map(_.show).mkString(", ")}. " +
+                  s"Remaining facilitators: ${currentState.facilitators.value.size}"
+              )
+          } else Applicative[F].unit
+        }
+
```
ryle-ai, line 204: `dag_consensus_unlock_peers_removed` is set as an absolute gauge on each unlock. Sequential unlocks overwrite each other — a dashboard only sees the most recent value. Since this represents a per-event count (how many peers were removed in *this* unlock), `incrementCounterBy` preserves all events:

```scala
Metrics[F].incrementCounterBy("dag_consensus_unlock_peers_removed", newlyRemoved.size) >>
```

ryle-ai (commented):
The core design of this refactoring is solid — the removal penalty lifecycle, the `PendingTriggers` atomicity fix, the quorum safety gate with its pigeonhole determinism argument, and the fiber lifecycle management are all well-reasoned. Test coverage is thorough. Two specific observations below.

On the `condModifyState` abandon path: I see the test `shouldAbandon when maxStallCycles exceeded even if not locked` confirms the intent to abandon regardless of lock status once the stall budget is exhausted. The trade-off is: after N failed lock/unlock cycles, cutting losses and restarting is probably better than hoping the Nth+1 attempt works. That said, `condModifyState` re-reads state inside the semaphore, so between the `getState` read at the top of `monitorStep` and the `condModifyState` call, a concurrent unlock can transition `Closed → Reopened`. The old `Closed`-only guard let such a round survive; the new `case Some(_)` removes it. The window is narrow and the impact is just a wasted round (not a safety issue), but worth being aware of — see inline comment.

marcus-girolneto:
@greptileai please review again

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/ConsensusEventLoop.scala
```diff
@@ -118,7 +128,22 @@ object ConsensusEventLoop {
...
+            fsm.handle(cmd).handleErrorWith { err =>
+              ctx.logger.error(err)(s"Unhandled error processing ${cmd.getClass.getSimpleName}, recovering") >>
+                Metrics[F].incrementCounter("dag_consensus_command_error") >>
+                (cmd match {
+                  case _: ConsensusCommand.ConsensusFinished | ConsensusCommand.RoundCompleted =>
+                    // Critical: if round-completion commands fail, FSM stays stuck in BUSY forever.
+                    // Force round completion so the next round can start.
+                    ctx.logger.warn("Forcing round completion after failed ConsensusFinished/RoundCompleted") >>
+                      Metrics[F].incrementCounter("dag_consensus_forced_round_completion") >>
+                      queue.offer(ConsensusCommand.RoundCompleted)
```
ryle-ai: This queues `RoundCompleted`, which routes to `completeRound(log.debug("Round completed without outcome"))`. That path calls `cleanupRound` (cancelling the timer fiber spawned by the *previous* round's `afterConsensusFinish`) but never schedules a new timer. If `timeTriggerInterval` hasn't elapsed since the last `scheduleTimeTrigger` call, the timer fiber is killed before it fires, and no replacement is spawned.

In a multi-node network this is fine — external events provide `FacilitateByEvent`. On a solo node (or during testing), consensus stalls permanently until the next external event.

Consider adding a fallback timer schedule in the recovery path:
```scala
case _: ConsensusCommand.ConsensusFinished | ConsensusCommand.RoundCompleted =>
  ctx.logger.warn(...) >>
    Metrics[F].incrementCounter("dag_consensus_forced_round_completion") >>
    queue.offer(ConsensusCommand.RoundCompleted)
    // TODO: also schedule a time trigger to prevent solo-node stall
```

ryle-ai (commented):
The error recovery in `ConsensusEventLoop` (forced `RoundCompleted` after `ConsensusFinished` failure) prevents the permanent BUSY-stuck scenario, which is good. However, the forced `RoundCompleted` follows the `handleWhileBusy(RoundCompleted)` → `completeRound(log.debug(...))` path, which runs `cleanupRound` (cancelling the previous round's timer fiber) but does NOT call `afterConsensusFinish` — so no timer is scheduled for the next round. If the cancelled timer hadn't fired yet and no external events arrive (`FacilitateByEvent`), consensus stalls until something else triggers it. The window is narrow (requires `ConsensusFinished` to throw AND the previous timer to not have fired yet), but on solo nodes or in testing this can deadlock consensus.

docs/CONSENSUS_REFACTORING.md
```diff
@@ -0,0 +1,527 @@
...
+
+| Phase | Condition | Keep Threshold | Remove Threshold | Rationale |
+|-------|-----------|----------------|------------------|-----------|
+| **DEFER** | `voterCount < (N+1)/2` | — | — | Too few voters for any peer to reach keepThreshold; defer to re-stall cycle |
+| **DECIDE** | `voterCount >= (N+1)/2` | `(N+1)/2` | `N/2 + 1` | N-based thresholds ensure all nodes make identical decisions |
+
+Where `N` is the total facilitator count and `voterCount` is how many facilitators actually sent ACKs for the current collecting kind.
+
+**Key properties:**
+- `keepThreshold + removeThreshold = N + 1 > N`, guaranteeing mutual exclusivity (no peer can be both kept and removed)
```
ryle-ai, line 88: The DEFER condition here says `voterCount < (N+1)/2` but the code in `UnlockConsensusUpdate.scala` uses `ceil(N/3).max(2)` — a much lower threshold:

```scala
val minVotersRequired = math.ceil(n.toDouble / 3).toInt.max(2)
```

For N=9: doc says defer below 5 voters, code defers below 3. For N=20: doc says 11, code says 7.

Line 96 ("DEFER uses `keepThreshold` as its minimum") also doesn't match — the code doesn't use `keepThreshold` as the defer boundary.

Since this is titled "Summary of Changes", the table should match what was actually implemented.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/StallDetector.scala
ryle-ai, line 158: Old code: `shouldAbandon = finalStallCycleCount >= config.maxStallCycles && isLocked`
New code: `shouldAbandon = finalStallCycleCount >= config.maxStallCycles || roundTimedOut`

The `&& isLocked` guard let a just-unlocked (`Reopened`) round survive one more monitoring cycle to process `advanceStatus`. Without it: if `finalStallCycleCount` reaches `maxStallCycles` on tick N (while `Closed`), and a concurrent unlock succeeds between tick N and N+1, tick N+1 still carries `finalStallCycleCount >= maxStallCycles` (it only resets on status change, not unlock). Combined with the unconditional `case Some(_)` removal at line ~173, a recovering round gets discarded.

The `roundTimedOut` path removing any state makes sense as a hard safety net. For the stall-cycle-exhaustion path, restoring the old guard would be more conservative:
```scala
shouldAbandon = (finalStallCycleCount >= config.maxStallCycles && isLocked) || roundTimedOut
```
Not blocking — the penalty system means the next round should do better regardless.

ryle-ai (commented):
The DEFER threshold documented in `CONSENSUS_REFACTORING.md` doesn't match the implementation. The doc says `voterCount < (N+1)/2` (i.e., `keepThreshold`) but the code uses `ceil(N/3).max(2)`. For N=9 that's 5 in the doc vs 3 in the code — meaningfully different behavior. See inline comment.

Separately: the `shouldAbandon` path in `StallDetector` intentionally drops the old `&& isLocked` guard, meaning a just-unlocked `Reopened` round can be killed if the stall cycle budget is already exhausted. The test "shouldAbandon when maxStallCycles exceeded even if not locked" confirms this is intentional. The trade-off is defensible (after N failed lock/unlock cycles, cutting losses is faster than hoping N+1 works), but the `condModifyState` at line ~173 compounds it by removing state unconditionally — the old `Closed`-only guard would let a just-unlocked round survive. The window is narrow and the impact is a wasted round (not a safety issue), so not blocking, but worth being aware of during testing.

marcus-girolneto:

modules/node-shared/src/test/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/StallDetectorSuite.scala
```diff
@@ -0,0 +1,501 @@
...
+        else baseTimeout
+
+      expect(!nearCompletion) &&
+      expect.same(35.seconds, effectiveTimeout)
+    }
+  }
+
+  // === Abandon guard: condModifyState removes state unconditionally when shouldAbandon ===
+
+  test("abandon guard: state removal is unconditional when shouldAbandon is true") {
```
ryle-ai, line 473: This test asserts `expect(true)`. It does not call `StallDetector`, does not instantiate any consensus state, and does not exercise the `condModifyState` guard it claims to test.

The comment says "any existing state is removed when shouldAbandon triggers" but the test just checks that `true` is `true`. If someone changes `case Some(_) =>` back to `case Some(s) if s.lockStatus === LockStatus.Closed =>` in `StallDetector.scala`, this test still passes.

All 27 tests in this suite have the same structural problem — they recompute formulas locally instead of calling `StallDetector` methods. The `shouldAbandon` tests (line ~238) duplicate `stallCycleCount >= maxStallCycles || roundTimedOut` inline; the `effectiveTimeout` tests duplicate the timeout selection logic inline; etc.

Consider either (a) extracting the decision functions from `StallDetector` as pure/testable methods and calling them from these tests, or (b) writing integration tests that exercise `StallDetector.monitorStep` with mocked storage/state.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/ConsensusEventLoop.scala
```diff
@@ -118,7 +128,27 @@ object ConsensusEventLoop {
...
+                  case _: ConsensusCommand.ConsensusFinished | ConsensusCommand.RoundCompleted =>
+                    // Critical: if round-completion commands fail, FSM stays stuck in BUSY forever.
+                    // Force round completion so the next round can start.
+                    // Also offer TimeTick: the forced RoundCompleted calls completeRound without
+                    // afterConsensusFinish, so no timer is scheduled for the next round. On solo nodes
+                    // with no external events, this would deadlock consensus. The TimeTick fires once
+                    // RoundCompleted sets isRunning=false, starting a new round from IDLE.
+                    ctx.logger.warn("Forcing round completion after failed ConsensusFinished/RoundCompleted") >>
+                      Metrics[F].incrementCounter("dag_consensus_forced_round_completion") >>
+                      queue.offer(ConsensusCommand.RoundCompleted) >>
```
ryle-ai, line 146: If `completeRound` itself throws (e.g., `cleanupRound` fails during fiber cancellation), this recovery queues another `RoundCompleted` which hits the same error, repeating indefinitely. Each cycle also queues a `TimeTick`, so the command queue grows by ~2 items per cycle with no bound.

A retry counter would cap this:
```scala
val maxRecoveryAttempts = 3
// track in a Ref, increment on each forced completion, reset on successful round start
```

Not blocking — requires `cleanupRound` to throw, which is unlikely — but worth hardening given this is the last-resort recovery path.

ryle-ai (commented):
The core design of this refactoring is solid — the quorum/supermajority gate, PendingTriggers atomicity fix, removal penalty lifecycle, and fiber lifecycle management are all well-reasoned. Two observations:

**StallDetectorSuite tests don't exercise real code.** All 27 tests in `StallDetectorSuite` duplicate formulas from `StallDetector.scala` locally and test the duplicate. None of them instantiate or call `StallDetector`. If someone changes the abandon formula, the near-completion extension logic, or the `condModifyState` guard in `StallDetector.scala`, every test in this suite still passes unchanged. This provides zero regression protection for the most safety-critical new code in the PR. Compare with `QuorumDeclarationsSuite`, which correctly extends `ConsensusStateAdvancer` and exercises the real `maybeGetQuorumDeclarations` method — that's the right pattern.

**Error recovery in `ConsensusEventLoop` has no retry budget.** If `completeRound` fails persistently (e.g., `cleanupRound` throws during fiber cancellation), the recovery at line ~146 queues another `RoundCompleted`, which fails again, queuing another, etc. Each cycle also queues a `TimeTick`. The queue grows unboundedly. In practice this requires fundamental JVM-level breakage, but a simple retry counter (`if retries < 3 then recover else log.error and give up`) would cap the blast radius.

PR #1470, opened 2026-03-11
fix: add MinFacilitatorCount safety floor to prevent facilitatorCount=0

scasplte2:
## Problem

When a global latency spike causes all facilitators to experience simultaneous delays, stale ACKs (containing empty 'who declared' sets) arrive after the network recovers. The unlock voting tallies max removeVotes for every peer, removing ALL facilitators and leaving `facilitatorCount=0` — an irrecoverable cluster death.

## Root Cause

`getAck()` returns the set of peers who have already declared for the current kind. During a latency spike, no declarations arrive before the round locks. Each node's ACK contains an empty or near-empty set. When latency clears and stale ACKs flood in simultaneously, every peer receives max `removeVotes` and zero `keepVotes`, causing all facilitators to be removed.

## Fix

Add a safety floor after computing keep/remove decisions: if `keptFacilitators.size < MinFacilitatorCount` (2), the unlock is aborted. The state remains `Closed` and defers to the stall detector, which will either retry with fresh ACKs or abandon the round after `maxStallCycles`.

**3 lines of production code:**
```scala
val MinFacilitatorCount: Int = 2

// After computing keep/remove partition:
.filter {
  case (_, keptFacilitators) => keptFacilitators.size >= MinFacilitatorCount
}
```

## Testing

Tested against v4.0.0 artifacts on an 8-node local cluster:

| Test | Without fix | With fix |
|------|-------------|----------|
| 20s global latency → clear | 🐛 fac=0 (100% reproducible) | ✅ fac stays, round deferred |
| Majority partition → restore | 🐛 fac=0 | ✅ fac stays |
| 60s spike on 6/8 → clear | 🐛 fac=0 | ✅ fac stays |

**Confirmed reproducible on current develop** (`ad46e5fc`). This is a mainnet risk — any cloud routing hiccup adding 20s+ globally would trigger it.

## Scope

- 1 file changed: `UnlockConsensusUpdate.scala` (3 lines + ScalaDoc)
- 2 new unit tests in `UnlockConsensusUpdateSuite.scala`
- No config changes, no protocol changes, backward compatible

Builds on top of `feat/refactoring-consensus` (adaptive tiered thresholds, facilitator re-entry, removal penalties) which addresses other consensus issues but does not include this safety floor.

### DIFF
--- a/docs/CONSENSUS_REFACTORING.md
+++ b/docs/CONSENSUS_REFACTORING.md
@@ -1,0 +1,528 @@
+# Consensus FSM Refactoring — Summary of Changes
+
+## Overview
+
+This refactoring targets seven categories of issues in the consensus engine:
+
+1. **Quorum safety** — Quorum declarations lacked a supermajority gate, risking non-deterministic decisions across different node views
+2. **Stall recovery** — Unlock voting used flat thresholds that couldn't adapt when many facilitators were unresponsive
+3. **Facilitator re-entry** — Peers removed during stall recovery were permanently excluded from future rounds
+4. **Removal penalty** — Removed peers could be re-selected one round later, causing repeated stalls when still unresponsive
+5. **Fiber leaks** — Orphaned fibers accumulating over sustained operation
+6. **Race condition** — Non-atomic state in `PendingTriggers` causing lost triggers
+7. **Error recovery** — FSM getting permanently stuck when critical commands fail
+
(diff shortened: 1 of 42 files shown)


ryle-ai (commented):
Thorough refactoring of the consensus engine addressing real stability problems (quorum safety, stall recovery, fiber leaks, race conditions). The code is well-structured and the test suite is comprehensive. A few observations below.

PR #1472, opened 2026-03-11
feat: expand local cluster cap from 3 to 9 nodes

scasplte2:
Replace hardcoded `for i in 0 1 2` loops with dynamic `seq 0 \$((MAX_NODES-1))` across docker scripts. `MAX_NODES` computed from `NUM_*_NODES` values, capped at 9.

### Changes
- `set-env.sh`: compute MAX_NODES
- `node-key-env-setup.sh`: generate keys for nodes beyond 0-2
- `compose-runner.sh`: dynamic iteration for all phases
- `docker-env-setup.sh`: env files for all nodes
- `cluster-health-check.sh`: check all nodes
- `tessellation-docker-cleanup.sh`: cleanup up to 9

### Usage
```bash
bash docker/bin/compose-runner.sh --up --num-gl0=8
```

Docker-only changes, no Scala code.

### DIFF
--- a/docker/bin/set-env.sh
+++ b/docker/bin/set-env.sh
@@ -376,6 +376,18 @@ if [ -n "$METAGRAPH" ]; then
     fi
 fi
 
+# Compute MAX_NODES as the maximum of all NUM_*_NODES values (capped at 10)
+# This drives how many node directories, keys, and configs are created
+_max_of() { [ "$1" -gt "$2" ] && echo "$1" || echo "$2"; }
+MAX_NODES=$(_max_of ${NUM_GL0_NODES:-0} ${NUM_GL1_NODES:-0})
+MAX_NODES=$(_max_of $MAX_NODES ${NUM_ML0_NODES:-0})
+MAX_NODES=$(_max_of $MAX_NODES ${NUM_CL1_NODES:-0})
+MAX_NODES=$(_max_of $MAX_NODES ${NUM_DL1_NODES:-0})
+# Ensure at least 3 (legacy default) and at most 9 (single-digit IP/port offset limit)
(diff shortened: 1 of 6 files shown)


docker/bin/set-env.sh
```diff
@@ -376,6 +376,18 @@ if [ -n "$METAGRAPH" ]; then
     fi
 fi
 
+# Compute MAX_NODES as the maximum of all NUM_*_NODES values (capped at 10)
```
ryle-ai, line 379: Comment says "capped at 10" but the code three lines down caps at 9 (`[ "$MAX_NODES" -gt 9 ] && MAX_NODES=9`). The second comment on line 386 correctly says "at most 9". Should be consistent.

```suggestion
# Compute MAX_NODES as the maximum of all NUM_*_NODES values (capped at 9)
```

ryle-ai (commented):
The loop generalizations look correct -- the `NUM_*_NODES` guards in each phase prevent starting containers beyond the configured count, and the port/IP addressing scheme (`${PREFIX}${i}0`, `${NET_PREFIX}.${layer}${i}`) stays valid for single-digit indices 0-8.

One thing to be aware of outside the diff: `tessellation-docker-cleanup.sh` lines 26-27 use `wait $LAST_PID` which only waits on the final backgrounded `cleanup_container` call. With the loop expanded from `0 1 2` to `seq 0 9`, that's 50 background processes where only 1 is awaited before the function returns. The retry loop on `docker network rm` handles the race in practice, but a plain `wait` (no args) would be more correct. Pre-existing issue, not introduced here, but the wider range makes the window larger.

ryle-goehausen (approved):

PR #1473, opened 2026-03-12

marcus-girolneto:

Key changes:

Safe quorum with supermajority gate in ConsensusStateAdvancer
Adaptive tiered unlock thresholds (DEFER/DEGRADED/FALLBACK/NORMAL)
Eligible facilitator re-entry after removal (no permanent exclusion)
Deterministic multi-round removal penalty via consensus outcome
Per-round fiber lifecycle tracking to prevent orphaned fibers
Atomic PendingTriggers to fix race condition losing triggers


marcus-girolneto:
@marcus-constellation-ai

marcus-constellation-ai (commented):
**Merge Confidence: 4/5** 🟢

Well-engineered refactoring with thorough documentation and solid test coverage. Three non-blocking observations:

1. **`Joining.scala:285`** — The `ConsensusConfigMismatch` join-time check is effectively dead code. `Cluster.make` accepts `consensusConfigHash: Option[Hash] = None` as default, and no application-level callers (dag-l0, currency-l0) were updated to pass it. Nodes with mismatched configs can still join and will only be detected via fork recovery during consensus rounds. The consensus-time detection via `checkForkByConsensusConfigHash` in the advancers works correctly. This looks like an intentional phased approach — worth a note in the PR or a follow-on ticket so the wiring doesn't get lost.

2. **`StallDetector.scala:156`** — `shouldAbandon` drops the `isLocked` guard from the old `finalStallCycleCount >= config.maxStallCycles && isLocked`. For the stall-cycle path this is functionally equivalent (the cycle count can only reach `maxStallCycles` when the state was just locked or has been locked). The meaningful behavioral change is the new `|| roundTimedOut` arm: a round actively making progress — including one in `Reopened` state after a successful unlock — is abandoned once the wall-clock ceiling fires, with `condModifyState { case Some(_) }` removing state unconditionally rather than only matching `LockStatus.Closed`. This is documented as intentional and is safe; just confirming the tradeoff is understood.

3. **`PendingTriggersSuite.scala:62`** — The test labelled "concurrent set and pull does not lose triggers" runs 1000 sequential `setEvent()`/`setTime()` calls via `traverse_`, not concurrent fibers. The atomicity guarantee comes correctly from `Ref.getAndSet`, but the test doesn't verify that property under actual concurrency. Minor naming issue; no coverage gap since `Ref` atomicity is a cats-effect guarantee.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/StallDetector.scala
```diff
@@ -0,0 +1,312 @@
...
+                else if (failedStallCycle) newStallCycleCount + 1
+                else newStallCycleCount
+              newRoundHadStall = ms.roundHadStall || didLock
+
+              _ <- Metrics[F].updateGauge("dag_consensus_stall_cycle", finalStallCycleCount)
+              _ <- Metrics[F].updateGauge("dag_consensus_stall_declaration_progress", declarationProgress)
+              _ <- Metrics[F].incrementCounter("dag_consensus_stall_restall").whenA(failedStallCycle)
+
+              roundElapsed = now - ms.roundStartTime
+              _ <- Metrics[F].updateGauge("dag_consensus_round_elapsed_seconds", roundElapsed.toSeconds.toInt)
```
marcus-constellation-ai, line 147: `shouldAbandon` no longer requires `isLocked`. For the stall-cycle exhaustion path this is largely equivalent to the old behaviour (the counter reaches `maxStallCycles` through fresh lock detections, so the state is typically Closed at that point). The substantive change is the new `|| roundTimedOut` arm: once `maxRoundDuration` (5 min) elapses, any state — including a round that recently unlocked and is actively advancing — is abandoned unconditionally. Confirmed this is the intended hard-ceiling semantics from the REFACTORING doc.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/StallDetector.scala
```diff
@@ -0,0 +1,312 @@
...
+                if (roundTimedOut)
+                  s"round timed out after ${roundElapsed.toSeconds}s (max=${config.maxRoundDuration.map(_.toSeconds)}s)"
+                else s"stuck after $finalStallCycleCount stall cycles"
+
+              _ <- (
+                logger.error(
+                  s"Round key=$key $abandonReason, abandoning round"
+                ) >>
+                  Metrics[F].incrementCounter("dag_consensus_round_abandoned") >>
+                  storage
```
marcus-constellation-ai: `case Some(_)` removes state regardless of `LockStatus`. The old guard `case Some(s) if s.lockStatus === LockStatus.Closed` was consistent with `shouldAbandon && isLocked`. Now that `shouldAbandon` can fire for `roundTimedOut` while the round is in any state, the unconditional match is correct — but it's a material change from the previous guarantee that only Closed state was force-removed. Worth keeping in mind if diagnosing unexpected round restarts from the `dag_consensus_round_abandoned` counter.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/domain/cluster/programs/Joining.scala
```diff
@@ -277,6 +278,12 @@ class Joining[
       allowanceListHash <- allowanceList.map(_.map(_.peerId)).hash
       _ <- Applicative[F].unlessA(registrationRequest.allowanceList === allowanceListHash)(AllowanceListDoesNotMatch.raiseError[F, Unit])
 
+      // Validate consensus config hash if both sides provide it (backward-compatible: skip during rolling upgrades)
+      configHashMismatch = (consensusConfigHash, registrationRequest.consensusConfigHash)
+        .mapN(_ =!= _)
+        .getOrElse(false)
+      _ <- ConsensusConfigMismatch.raiseError[F, Unit].whenA(configHashMismatch)
```
marcus-constellation-ai, line 285: `ConsensusConfigMismatch.raiseError` never fires in this PR. `Joining` receives `consensusConfigHash` from `Cluster.make`, which defaults to `None`, and neither dag-l0 nor currency-l0 wiring passes the value. `(None, registrationRequest.consensusConfigHash).mapN(_ =!= _).getOrElse(false)` is always `false`. Consensus-time detection via `checkForkByConsensusConfigHash` (in the state advancers) is wired correctly. If the join-time gate is needed, the dag-l0/currency-l0 `Cluster.make` call sites need to be updated to forward `config.snapshot.consensus.deterministicConfigHash`.

modules/node-shared/src/test/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/PendingTriggersSuite.scala
```diff
@@ -0,0 +1,71 @@
...
+  test("pullNext clears state atomically") {
+    for {
+      pending <- mkPending
+      _ <- pending.setTime()
+      first <- pending.pullNext
+      second <- pending.pullNext
+    } yield expect(first.contains(TriggerPriority.Time)).and(expect(second.isEmpty))
+  }
+
+  test("concurrent set and pull does not lose triggers") {
```
marcus-constellation-ai, line 62: This test uses `traverse_` which sequences effects one at a time — it is not concurrent. After 1000 sequential sets the final state is `TimePending` (odd indices set time, the last even-indexed `setEvent` cannot downgrade it), so `result.isDefined` always holds. The test verifies sequential priority logic correctly, but the "concurrent" label is misleading. The atomicity of `Ref.getAndSet` is a cats-effect guarantee and doesn't need a sequential test to prove it.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/config/types.scala
```diff
@@ -148,8 +148,47 @@ object types {
...
+      *
+      * IMPORTANT: When adding new fields to ConsensusConfig, evaluate whether they affect consensus determinism. If the field changes what
+      * peers decide (facilitator selection, quorum logic, voting thresholds), add it to the hash string below. If it only affects timing or
+      * performance, exclude it.
+      *
+      * Hash.fromBytes applies SHA-256 (via sha256DigestFromBytes), producing a compact 64-char hex digest.
+      */
+    lazy val deterministicConfigHash: io.constellationnetwork.security.hash.Hash = {
+      val configString =
+        s"maxFacilitatorCount=${maxFacilitatorCount.map(_.value)}," +
```
ryle-ai, line 187: The hash string uses `maxFacilitatorCount.map(_.value)` which serializes as `Some(10)` / `None` via `Option.toString`. This works for equality but is an unusual wire format for a consensus-critical hash. If a future refactor changes `maxFacilitatorCount` from `Option[PosInt]` to `Option[Int]` (or similar), the `.map(_.value)` call changes meaning and silently produces a different hash, causing a network fork.

Consider adding a regression test that pins the hash of a reference config to a known value:
```scala
test("deterministicConfigHash is stable") {
  val config = ConsensusConfig(
    timeTriggerInterval = 10.seconds,
    declarationTimeout = 35.seconds,
    // ... all fields with known values
    maxFacilitatorCount = Some(PosInt(10)),
    maxStallCycles = 5,
    quorumThreshold = Some(0.67),
    removalPenaltyRounds = 3
  )
  assertEquals(config.deterministicConfigHash.value, "expected-hex-here")
}
```
This catches accidental format changes at compile time rather than as a production fork.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/ConsensusEventLoop.scala
```diff
@@ -118,7 +128,27 @@ object ConsensusEventLoop {
...
+                  case _: ConsensusCommand.ConsensusFinished | ConsensusCommand.RoundCompleted =>
+                    // Critical: if round-completion commands fail, FSM stays stuck in BUSY forever.
+                    // Force round completion so the next round can start.
+                    // Also offer TimeTick: the forced RoundCompleted calls completeRound without
+                    // afterConsensusFinish, so no timer is scheduled for the next round. On solo nodes
+                    // with no external events, this would deadlock consensus. The TimeTick fires once
+                    // RoundCompleted sets isRunning=false, starting a new round from IDLE.
+                    ctx.logger.warn("Forcing round completion after failed ConsensusFinished/RoundCompleted") >>
+                      Metrics[F].incrementCounter("dag_consensus_forced_round_completion") >>
+                      queue.offer(ConsensusCommand.RoundCompleted) >>
```
ryle-ai, line 150: If the `RoundCompleted` handler itself fails persistently, this creates an unbounded retry loop: `RoundCompleted` fails → recovery queues new `RoundCompleted` → fails again → repeat. The `dag_consensus_forced_round_completion` counter will spike, but there's no circuit breaker to stop the loop.

Minor suggestion: track consecutive forced completions and escalate (e.g., log at error level and skip the re-queue) after a threshold:
```scala
consecutiveForcedRef.modify { n =>
  if (n >= 3) (0, Async[F].unit) // give up, let external health check restart
  else (n + 1, queue.offer(ConsensusCommand.RoundCompleted) >> queue.offer(ConsensusCommand.TimeTick))
}.flatten
```
This is a very unlikely scenario since `cleanupRound` is fiber cancellation + ref clearing, but it makes the failure mode bounded.

ryle-ai (commented):
## Consensus FSM Refactoring Review

This is a substantial and well-structured overhaul of the consensus engine. The quorum safety gate (supermajority + pigeonhole argument), N-based unlock thresholds, multi-round removal penalty, StallDetector extraction, PendingTriggers atomicity fix, and fiber lifecycle management all address real failure modes. The 83 new tests (including property-based suites for quorum declarations, eligible facilitators, and removal penalties) provide strong coverage of the new logic.

Two minor concerns below — neither blocking.

### `deterministicConfigHash` serialization fragility

The hash is computed via string interpolation of Scala values (`Option.toString`, `Double.toString`, `Int.toString`). This is deterministic within a JVM, and the Scaladoc comment explaining which fields to include is excellent. However, the hash format is implicitly part of the consensus protocol — if two nodes running different code produce different strings for the same logical config, they'll fork. Consider adding a pinning test that asserts the hash of a known config equals a hardcoded expected value. This would catch accidental format changes during future refactors.

### Error recovery retry bound

The `handleErrorWith` recovery for `ConsensusFinished`/`RoundCompleted` re-queues `RoundCompleted`, which will re-enter the same code path. If the root cause persists (e.g., a corrupted state ref), this creates an unbounded retry loop. In practice this is extremely unlikely since `cleanupRound` is just fiber cancellation + ref clearing, but a counter-based circuit breaker (e.g., abandon after 3 consecutive forced completions) would make the failure mode explicit rather than relying on the underlying operations being infallible.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/config/types.scala
```diff
@@ -148,8 +148,47 @@ object types {
...
     noProgressTimeout: Option[FiniteDuration] = None,
-    maxStallCycles: Int = 3
-  )
+    maxStallCycles: Int = 3,
+    maxRoundDuration: Option[FiniteDuration] = None,
+    quorumThreshold: Option[Double] = None,
+    removalPenaltyRounds: Int = 0
+  ) {
+    quorumThreshold.foreach { t =>
+      require(t > 0.5 && t <= 1.0, s"quorumThreshold must be in (0.5, 1.0], got $t")
```
marcus-constellation-ai: The lower bound `t > 0.5` is insufficient for the pigeonhole safety guarantee. The documented invariant ("any quorum-sized subset computes the same `pickMajority`") requires `t > 2/3`. With `t ∈ (0.5, 2/3]` you can construct a quorum subset where the majority value holds fewer than half the votes — different nodes pick different outcomes.

Example: `N=20, t=0.55, quorumSize=11`. Dominant value has exactly 11 votes, minority has 9. A subset of 11 containing all 9 minority + 2 majority = 11 nodes: the majority value has 2/11 support, loses `pickMajority`.

Fix: `require(t > 2.0/3.0 && t <= 1.0, ...)`

marcus-constellation-ai (commented):
**Merge Confidence: 4/5** 🟢

Solid, well-documented refactoring with comprehensive test coverage. One correctness gap in the quorum validation bounds is worth fixing before a future config change bites someone.

1. **`types.scala:157`** — `require(t > 0.5 && t <= 1.0)` is mathematically too permissive. The pigeonhole safety guarantee the doc describes — "any quorum-sized subset will compute the same `pickMajority` result" — only holds when `t > 2/3`. With `t ∈ (0.5, 2/3]` the guarantee breaks: for `N=20, t=0.55, quorumSize=11`, a subset of 11 drawn from 11 majority + 9 minority can contain just 2 majority nodes (11-9=2), which won't win a simple majority of 11. A misconfigured `quorum-threshold = 0.6` would silently allow non-deterministic quorum decisions. The current default of 0.67 is safe, but the guard should be `t > 2.0/3.0 && t <= 1.0` to match the documented invariant.

2. **`ConsensusFSM.scala` — cleanup ordering** — The reordering of `cleanupRound` to run before `preAction` in `completeRound` is correct and the comment explains the solo-node deadlock it prevents. Confirmed the logic: `cancelRoundFibers` clears `roundFibersRef` to `Nil`, then `preAction` (via `afterConsensusFinish` → `scheduleTimeTrigger`) re-populates it for the next round, so the new timer fiber is not cancelled by the current round's cleanup.

3. **`Joining.scala` dead code** — As flagged by the existing review: `ConsensusConfigMismatch.raiseError` never fires because no dag-l0/currency-l0 call site passes `consensusConfigHash` to `Cluster.make`. The consensus-time detection via `checkForkByConsensusConfigHash` in the advancers is correctly wired and covers the practical case. Worth a follow-on ticket to wire the join-time gate.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotConsensusStateAdvancer.scala
```diff
@@ -322,22 +335,24 @@ object GlobalSnapshotConsensusStateAdvancer {
...
-                    )
-                  ),
-                  sideEffect = persistAndGossip(signedArtifact, status.majorityArtifactInfo.context)
-                )
-              } yield result
+              // Use the artifact hash (agreed upon during Proposals phase) instead of signedArtifact.hash.
+              // signedArtifact.hash includes signatures, which can differ across nodes when quorum < total
+              // (e.g., some nodes collect 3 signatures, others 4), causing non-deterministic snapshotHash
+              // and deadlocking the next round's Facility phase.
+              val snapshotHash = status.majorityArtifactInfo.hash
```
ryle-ai, line 430: This fix is correct and necessary for quorum mode — but the same fix is missing from the Currency advancer. `CurrencySnapshotConsensusStateAdvancer.buildFinishedTransition` (line 387) still uses `status.signedMajorityArtifact.hash`, which includes per-node-varying signatures. With `quorum-threshold = 0.67` now enabled in `currency-l0.conf`, this will produce different `snapshotHash` across nodes and trigger fork detection every round.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/config/types.scala
ryle-ai: The lower bound `t > 0.5` is too permissive for the pigeonhole guarantee documented above. The invariant — "any quorum-sized subset computes the same `pickMajority` result" — only holds when `t > 2/3`.

Example: `N=20, t=0.55, quorumSize=11`. Value A has 11 votes, value B has 9. A subset of 11 containing all 9 B-voters + 2 A-voters: A has 2/11 support and loses `pickMajority`. Different nodes compute different outcomes.

The deployed default of 0.67 is safe, but `require(t > 2.0/3.0 && t <= 1.0, ...)` would make this guard match the actual mathematical requirement and prevent a future misconfiguration from silently breaking quorum safety.

ryle-ai (changes requested):
## Currency advancer `snapshotHash` not fixed for quorum mode

The Global advancer correctly switches from `signedArtifact.hash` to `status.majorityArtifactInfo.hash` in `buildFinishedTransition` (line 342), with a clear comment explaining why: under quorum mode, different nodes collect different signature subsets, so `Signed(artifact, signaturesNes).hash` is non-deterministic across nodes.

The Currency advancer has the same problem but was not fixed. At `CurrencySnapshotConsensusStateAdvancer.scala:387`:

```scala
snapshotHash <- status.signedMajorityArtifact.hash
```

`signedMajorityArtifact` is `Signed(artifact, signaturesNes)` where `signaturesNes` comes from the CollectingSignatures phase — which now uses `maybeGetQuorumDeclarations`. With `quorum-threshold = 0.67` enabled in `currency-l0.conf`, node A might collect 14 valid signatures while node B collects 18. Their `signedMajorityArtifact.hash` values differ, producing different `snapshotHash` in the `Finished` status. The next round's Facility declaration includes this as `lastSnapshotHash`, triggering `checkForkByLastSnapshotHash` → fork recovery → metagraph restart.

This needs the same fix applied to the Global advancer: use the unsigned artifact hash (agreed upon during the Proposals phase) instead of the signed artifact hash.

ryle-ai (commented):
## Missing `TimeTick` on stall detector round abandonment

The error recovery in `ConsensusEventLoop` (line 146) queues both `RoundCompleted` and `TimeTick` with a comment explaining why: *"the forced RoundCompleted calls completeRound without afterConsensusFinish, so no timer is scheduled for the next round."*

The stall detector's abandon path has the exact same situation — `RoundCompleted` without `afterConsensusFinish` — but omits the `TimeTick`.

Pre-PR, this wasn't an issue because timer fibers were fire-and-forget via `supervisor.supervise(...).void` and were never cancelled. The new `spawnTracked` + `cleanupRound` correctly fixes the fiber leak, but `cleanupRound` now cancels any lingering timer fiber from the previous round's `scheduleTimeTrigger`. If that timer already fired (triggering the now-abandoned round) and no external events arrived during the round, `pending.pullNext` returns `None` and no new round starts.

With current configs (`timeTriggerInterval=43s`, earliest abandon ≈75s) the timer fires well before abandonment, so its `TimeTick` is captured as pending. But the invariant is config-dependent — it breaks if `timeTriggerInterval` is ever tuned above the abandon window, or if a round is abandoned via `maxRoundDuration` before the timer fires.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/StallDetector.scala
```diff
@@ -0,0 +1,312 @@
...
+                    .condModifyState[Unit](key) {
+                      case Some(_) =>
+                        (none[ConsensusState[Key, Status, Outcome, Kind]], ()).some.pure[F]
+                      case _ =>
+                        none.pure[F]
+                    }
+                    .void >>
+                  queue.offer(ConsensusCommand.RoundCompleted)
```
ryle-ai: Same situation as the error recovery in `ConsensusEventLoop` (line 146): `RoundCompleted` enters `completeRound` without `afterConsensusFinish`, so no timer is scheduled for the next round. Consider adding `queue.offer(ConsensusCommand.TimeTick)` here to match:

```scala
queue.offer(ConsensusCommand.RoundCompleted) >>
  queue.offer(ConsensusCommand.TimeTick)
```

With current configs the timer fires before abandonment so this is a latent issue, not a current bug.

ryle-ai (commented):
## Currency advancer `snapshotHash` fix has been applied

The outstanding CHANGES_REQUESTED about `CurrencySnapshotConsensusStateAdvancer.buildFinishedTransition` using `signedMajorityArtifact.hash` (which includes per-node-varying signatures) appears to be resolved in the current diff. Line 404 now reads:

```scala
snapshotHash <- status.signedMajorityArtifact.value.hash
```

`.value` extracts the unsigned `CurrencySnapshotArtifact` from the `Signed` wrapper before hashing, producing a deterministic hash across nodes regardless of which signature subset each node collected. This is semantically equivalent to the Global advancer's approach (`status.majorityArtifactInfo.hash`) — both hash the unsigned artifact. The implementation differs (on-the-fly computation vs pre-computed field) because `CollectingBinarySignatures` carries a `Signed[CurrencySnapshotArtifact]` rather than an `ArtifactInfo` with a pre-computed hash.

## Quorum threshold validation

Note for other reviewers: the `require` guard is `t > 2.0 / 3.0 && t <= 1.0` (line 181 of types.scala), not `t > 0.5` as stated in some earlier comments. The 2/3 bound correctly matches the pigeonhole safety guarantee documented in the Scaladoc.

## Join-time config hash check is dead code

Already flagged but worth confirming: `Joining.consensusConfigHash` defaults to `None` and neither `dag-l0` nor `currency-l0` passes it to `Cluster.make`. The `mapN` guard on line 283 always evaluates to `false`. The consensus-time detection via `checkForkByConsensusConfigHash` in both advancers IS correctly wired (state creators pass `deterministicConfigHash` to Facility declarations). So config mismatches are caught during rounds, not at join time. If the join-time gate is intended for a later phase, a TODO or ticket would prevent the wiring from being forgotten.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/StallDetector.scala
```diff
@@ -0,0 +1,346 @@
...
+                  FiniteDuration((baseEffectiveTimeout.toMillis * config.leaderQualityTimeoutMultiplier).toLong, MILLISECONDS)
+                else baseEffectiveTimeout
+
+              // Handle stall: view change for proposal phase, or count towards abandon
+              didViewChange <- handleStall(
+                key = key,
+                state = state,
+                declarationTimeout = effectiveTimeout,
+                statusDuration = statusDuration,
+                alreadyHandled = newStallCount > 0 && !statusChanged,
```
ryle-ai, line 132: `alreadyHandled = newStallCount > 0 && !statusChanged` is `true` for the entire remaining lifetime of this status phase after the first stall — `newStallCount` is set from `ms.stallCount` (which was 1 after the first handling), and `statusChanged` is `false` (view changes modify `viewNumber`/`leader` but not `status`). So `handleStall` returns `false` on every subsequent timeout, `finalStallCount` stays at 1, and `shouldAbandon` (which requires `>= maxStallCycles(5)`) never fires.

This also means only one view change occurs per proposal phase. If the second leader fails, no further rotation happens.

If the intent is for `maxStallCycles` to work (the config was bumped from 3→5, suggesting it matters), this guard should allow re-triggering after the adjusted timer resets — something like tracking whether the current timeout period has already been handled rather than whether any stall was ever handled for this status.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/StallDetector.scala
```diff
@@ -0,0 +1,346 @@
...
+        ResourcesInfo(
+          hash = resources.peerDeclarationsMap.keySet.hashCode(),
+          declaredCount = resources.peerDeclarationsMap.size,
+          activeCount = active.size,
+          missingPeerIds = Set.empty
+        )
+    }
+  }
+
+  /** Handle a stall condition. Returns true if a view change was performed. */
```
ryle-ai, line 250: Doc says "Returns true if a view change was performed" but the non-proposal branch (line 288) also returns `true` without performing a view change. The caller uses this as `didViewChange` to reset the timer and increment `stallCount`. Consider renaming to `didHandleStall` and updating the doc to match — the current naming makes the monitor loop harder to reason about.

ryle-ai (commented):
## `maxStallCycles` is unreachable in StallDetector

The `alreadyHandled` guard at StallDetector line 132 prevents `stallCount` from incrementing past 1 within any status phase. After the first stall is handled (view change or stall detection), `newStallCount > 0 && !statusChanged` is permanently `true` for the remainder of that phase. Every subsequent timeout fires `handleStall` with `alreadyHandled = true`, which returns `false`, so `finalStallCount` never increases.

This means `shouldAbandon = finalStallCount >= config.maxStallCycles` never evaluates to `true` through the stall path. Only `roundTimedOut` (the `maxRoundDuration` check) can actually abandon a round.

For proposal phases specifically, this limits view changes to exactly one per round. The infrastructure supports multiple view changes (`selectLeader(facilitators, entropy, viewNumber)` with `viewNumber % sorted.size`), and the doc describes `View 0 → View 1 → View 2 → ...`, but the detector only performs one rotation. If the second leader also fails, the round sits for up to `maxRoundDuration` (5 min). The old code abandoned after ~3 × `reStallTimeout` ≈ 30–45s.

This isn't a correctness issue since `maxRoundDuration` catches all cases, but:
- `max-stall-cycles = 5` in both configs is dead configuration (previously 3, increased to 5 as if it mattered)
- Recovery from networks with multiple consecutive unreachable leaders is significantly slower
- The documented stall recovery decision tree ("stallCount >= maxStallCycles → Abandon Round") doesn't match implementation

The fix is to allow re-triggering after the timer resets rather than gating on whether any stall has ever been handled for this status — e.g., track the stall-handled timestamp and only set `alreadyHandled` when `statusDuration` hasn't exceeded a fresh timeout since the last handling.

PR #1474, opened 2026-03-12
refactor: rewrite consensus engine with leader-based protocol and stall detection

marcus-girolneto:
Replace lock-based consensus updates with a leader-based proposal protocol where only the leader spreads proposals and followers validate before signing, eliminating deadlocks from majority-vote artifact resolution.

- Add leader-based proposal flow: leader spreads Proposal + artifact via direct gossip, followers validate leader's artifact before signing
- Add StallDetector with view change: monitors declaration progress, rotates leader on proposal stalls, abandons rounds after max stall cycles
- Add PeerQualityTracker: tracks per-peer completion/view-change rates, surfaces leaderScore in Round STARTED/COMPLETED/monitor logs
- Add FacilitatorSelector: deterministic hash-distance ordering with configurable max facilitator count and leader selection
- Add ConsensusDirectSender: targeted peer-to-peer declaration spreading
- Add multi-line consensus logs with [CONSENSUS:LEADER/FOLLOWER] role prefix, missing-peers-only display, and [CONSENSUS:PROPOSAL] context for acceptance
- Remove duplicate AcceptanceManager logger.debug (keep loggerBundle.app.info)
- Remove UnlockConsensusUpdate (replaced by view change + round abandon)
- Add developer guide to CONSENSUS_REFACTORING.md


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/StallDetector.scala
```diff
@@ -0,0 +1,346 @@
...
+                if (ops.isProposalPhase(state.status) && leaderQuality < config.leaderQualityThreshold)
+                  FiniteDuration((baseEffectiveTimeout.toMillis * config.leaderQualityTimeoutMultiplier).toLong, MILLISECONDS)
+                else baseEffectiveTimeout
+
+              // Handle stall: view change for proposal phase, or count towards abandon
+              didViewChange <- handleStall(
+                key = key,
+                state = state,
+                declarationTimeout = effectiveTimeout,
+                statusDuration = statusDuration,
```
ryle-ai, line 131: `alreadyHandled = newStallCount > 0 && !statusChanged` makes the stall recovery loop single-shot.

After the first stall (view change or non-proposal stall), `stallCount` becomes 1. On subsequent iterations, `newStallCount = 1` (status hasn't changed), so `alreadyHandled = true`, and `shouldHandle` is permanently false. The timer reset from `adjustedStatusStartTime = now` is wasted because the guard never allows re-entry.

The old code's `failedStallCycle` path detected repeated timeouts after the initial lock, incrementing `stallCycleCount` each time. That mechanism was lost in this refactoring.

Simplest fix: for proposal-phase stalls, `alreadyHandled` should only be true during the current timeout window (i.e., between a stall handling and the next `declarationTimeout` expiry). Since `adjustedStatusStartTime` already resets on handling, just removing the `alreadyHandled` guard (or always passing `false`) would work — `statusDuration < declarationTimeout` prevents re-triggering within the same timeout window.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/StallDetector.scala
```diff
@@ -0,0 +1,346 @@
...
+                else baseTimeout
+
+              // Reduce patience for known-bad leaders during proposal phase
+              leaderQuality <- peerQualityTracker.getQualityScore(state.leader)
+              effectiveTimeout =
+                if (ops.isProposalPhase(state.status) && leaderQuality < config.leaderQualityThreshold)
+                  FiniteDuration((baseEffectiveTimeout.toMillis * config.leaderQualityTimeoutMultiplier).toLong, MILLISECONDS)
+                else baseEffectiveTimeout
+
+              // Handle stall: view change for proposal phase, or count towards abandon
```
ryle-ai, line 126: Naming: `didViewChange` is misleading here. For non-proposal stalls, `handleStall` returns `true` (meaning "stall was handled") even though no view change occurred. This same `true` then drives `adjustedStatusStartTime = if (didViewChange) now else ...` and `finalStallCount = if (didViewChange) ...` — the timer reset and count increment logic works correctly for both cases, but a reader would expect `didViewChange = true` to mean an actual view change. Consider renaming to `didHandleStall` or `stallHandled`.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/ConsensusRoutes.scala
```diff
@@ -36,6 +41,15 @@ class ConsensusRoutes[F[_]: Async, Key: Order: Encoder: Decoder, Artifact, Conte
           case _                                                     => Ok(none[GetConsensusOutcomeRequest[Key]])
         }
       } yield result
+    case req @ POST -> Root / "push-rumor" =>
+      HasherSelector[F].withCurrent { implicit hasher =>
+        for {
+          signedRumor <- req.as[Signed[RumorRaw]]
```
ryle-ai, line 47: The `push-rumor` endpoint enqueues the rumor without verifying its signature before offering to the queue. The regular gossip ingestion path has rate limiting (bounded gossip rounds, anti-entropy intervals). Direct push bypasses that, so a compromised peer could queue invalid rumors faster than they get processed and discarded.

Not a blocker since P2P routes require cluster auth and the queue is presumably bounded, but a basic `signedRumor.toHashed[F]` + signature check before `offer` would be cheap defense-in-depth.

ryle-ai (changes requested):
## Stall recovery loop fires only once, then goes dormant

The `alreadyHandled` guard in `StallDetector.monitorStep` permanently prevents further stall handling after the first stall event. This breaks both the multi-view-change mechanism for proposal stalls and the `maxStallCycles` abandon mechanism for all other phases.

Walk-through:
1. First stall fires: `stallCount` goes from 0 to 1, `adjustedStatusStartTime` resets to `now`.
2. Next timeout: `newStallCount = 1` (status unchanged), so `alreadyHandled = 1 > 0 && !false = true`.
3. `shouldHandle = statusDuration >= timeout && !alreadyHandled` → always `false`.
4. `stallCount` stays at 1 forever. Never reaches `maxStallCycles` (5). No further view changes.
5. Only `maxRoundDuration` (5 min) provides eventual recovery.

The old code handled this via the `failedStallCycle` detection path, which checked for repeated timeouts after the first lock. That mechanism was removed during the refactoring without an equivalent replacement.

For **proposal phases**, this means at most one leader rotation per round instead of cycling through facilitators until `maxStallCycles`.

For **non-proposal phases** (facilities, signatures), `stallCount` can never reach `maxStallCycles`, so the abandon-after-N-cycles mechanism is completely inoperative. Every stalled round waits the full `maxRoundDuration` (5 minutes) instead of abandoning after ~50-150 seconds of repeated stalls.

Suggested fix: change `alreadyHandled` to track whether we've already handled a stall *since the last timer reset*, not whether any stall has been handled in this status. After a stall resets `adjustedStatusStartTime`, the next timeout should be eligible for handling again. One approach: set `alreadyHandled = false` (since the timer reset from the previous stall already prevents immediate re-triggering), or introduce a `lastHandledAtStallCount` field in `MonitorState`.

PR #1475, opened 2026-03-16
refactor: consensus engine rewrite with leader-based protocol, TCA filter, and structured logging

marcus-girolneto:
## Summary

- Rewrite consensus engine from all-to-all lock/unlock model to leader-based 3-phase protocol (Facilities → Proposals → Signatures) with deterministic facilitator selection via rendezvous hashing
- Add Ouroboros-inspired Trailing Common Ancestor (TCA) filter that derives facilitator eligibility from historical snapshot proofs — uses early/recent window split to exclude peers that were active but went silent, while allowing new peers to onboard immediately
- Add StallDetector with adaptive phase-aware timeouts, ViewChangeManager for leader re-election on proposal stalls, and AbandonmentTracker for consecutive failure recovery
- Add ConsensusLog structured logging with categories (LIFECYCLE, PHASE, STALL, QUORUM, FORK, FACILITATOR, PROPOSAL, VALIDATION) replacing verbose per-peer rosters with aggregate counts
- Add deterministic config hash in Facility declarations for early fork detection when nodes run different consensus parameters
- Fix leader self-view-change (StallDetector short-circuits when leader == selfId), FSM guard blacklist instead of whitelist, and gossip noise downgrade for peers in download/leaving states

## Motivation

The previous consensus engine suffered from cascading disconnects: when facilitators went offline, surviving nodes couldn't deterministically agree on who to exclude — each node's local timeout observations differed, causing `facilitatorsHash` mismatches and forks. With >50% facilitators unresponsive, rounds entered permanent deadlock.

The TCA filter solves this by deriving exclusions from consensus-agreed data (`Signed[Snapshot].proofs`). The early/recent split ensures new peers joining after a rollback aren't falsely flagged as degraded, which was validated against a live testnet rollback scenario where facilitator count collapsed from 4 to 1 and caused forks due to reward calculation differences.


.github/workflows/testnet-release.yml
```diff
@@ -44,10 +44,10 @@ jobs:
           JAVA_VERSION: 21
         uses: "./.github/templates/actions/setup_java_and_scala"
 
-      - name: Test
-        run: sbt 'scalafixAll --check --rules OrganizeImports;scalafmtCheckAll;test'
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+#      - name: Test
+#        run: sbt 'scalafixAll --check --rules OrganizeImports;scalafmtCheckAll;test'
```
ryle-ai, line 48: Commenting out the full test step (`scalafixAll`, `scalafmtCheckAll`, `test`) means this rewrite ships to testnet with no automated verification. The new test suites (StallDetectorSuite, FacilitatorSelectorSuite, etc.) would not run. This should be restored.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotConsensusStateAdvancer.scala
```diff
@@ -260,6 +767,8 @@ object GlobalSnapshotConsensusStateAdvancer {
         facilitatorsHash <- state.facilitators.value.hash
         signature <- Signature.fromHash(keyPair.getPrivate, majorityInfo.hash)
         _ <- recordProposalAffinity(proposalHashes, status.proposalArtifactInfo.hash)
+        // Round succeeded — discard the proposal savepoint so it won't be restored on the next ordinal
+        _ <- proposalSavepointRef.set(none)
```
ryle-ai, line 771: Savepoint cleared too early. `buildSignatureTransition` is called when transitioning to `CollectingSignatures`, not when the round completes. If the round is abandoned during signature collection and retried at the same ordinal, `buildProposalTransition` finds `None` here and calls `createArtifact()` on a MptStore that already has the previous attempt's mutations applied, producing a wrong artifact. The savepoint should be cleared only after the snapshot is successfully persisted — in `toFinishedPhase` or the storage commit step.

modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotConsensusStateAdvancer.scala
```diff
@@ -97,13 +139,46 @@ object GlobalSnapshotConsensusStateAdvancer {
...
+              val completed =
+                if (state.withdrawnFacilitators.value.contains(pid) || state.removedFacilitators.value.contains(pid)) 0
+                else 1
+              pid -> (completed, 1)
+            }
+          )
+          // Accumulate with previous rounds: merge (completed, participated) tuples
+          val accumulatedQuality: SortedMap[PeerId, (Int, Int)] = {
+            val previous = state.lastOutcome.peerQuality
+            val allPeerIds = (previous.keySet.toList ::: thisRoundQuality.keySet.toList).distinct
```
ryle-ai, line 165: `distinct` on a concatenated list is O(n²). Replace with `(previous.keySet | thisRoundQuality.keySet).toList` for O(n log n). More importantly, `peerQuality` grows without bound: every round appends new peers and never removes old ones. This field is serialized in `GlobalConsensusOutcome` and transmitted in `GetConsensusOutcomeRequest` responses. A decay mechanism (analogous to `PeerQualityTracker.decayThreshold`) will be needed before this becomes a problem in production.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/FacilitatorSelector.scala
```diff
@@ -101,6 +101,79 @@ class FacilitatorSelector private (maxFacilitatorCount: Option[Int]) {
...
+    * @param viewNumber
+    *   Current view number (0 = initial leader, incremented on view change)
+    * @param qualityScores
+    *   Map of peer quality scores (0.0 = worst, 1.0 = best). Peers not in the map default to 1.0.
+    * @param qualityWeight
+    *   Weight of quality in the combined score (0.0 to 1.0). Default 0.3.
+    * @return
+    *   The selected leader PeerId
+    */
+  def selectLeaderWeighted(
```
ryle-ai, line 153: `selectLeaderWeighted` is documented as non-deterministic and explicitly not for production use ("use `selectLeader` for production leader selection"), but it's a public method on a public class. If it's accidentally wired in production code, it will cause divergent leader selections across nodes. Making it package-private (`private[consensus]`) would prevent the mistake without losing test accessibility.

ryle-ai (changes requested):
Large consensus engine rewrite introducing leader-based proposal protocol, TCA-based facilitator filtering, stall detection with view-change, and structured consensus logging. The architecture is generally sound — rendezvous hashing for facilitator selection, phase-adaptive timeouts in `StallDetector`, and the `maybeGetQuorumDeclarations` safe-majority gate are well-designed.

Two issues need attention before this merges.

**CI tests commented out.** The entire `Test` step in `testnet-release.yml` — `scalafixAll`, `scalafmtCheckAll`, and `test` — is commented out. For a 6275-line rewrite of the core consensus engine, shipping to testnet without any automated test verification is a significant risk. The test suite for the new components exists in this PR (StallDetectorSuite, FacilitatorSelectorSuite, etc.), so there's no obvious reason for this to be disabled. This should be un-commented before merging.

**MptStore savepoint cleared too early.** `buildSignatureTransition` clears `proposalSavepointRef` when transitioning to `CollectingSignatures`, with the comment "Round succeeded." But the round hasn't succeeded yet — signing is still pending. If the round is subsequently abandoned during `CollectingSignatures` (stall timeout or `maxRoundDuration`) and retried at the same ordinal, `buildProposalTransition` finds no savepoint to restore and calls `createArtifact()` on an MptStore that already carries mutations from the failed attempt. The artifact for that ordinal would be built on top of a partially-applied state, producing incorrect output. The savepoint should be cleared after the snapshot is committed to storage (in `toFinishedPhase` or the persistence step downstream), not at the proposal→signatures transition.

**Non-blocking:** `peerQuality` in `GlobalConsensusOutcome` accumulates across all historical rounds with no decay. Every round, `accumulatedQuality` is built by concatenating `previous.keySet.toList ::: thisRoundQuality.keySet.toList` and calling `.distinct` (O(n²)). Since the result is stored in `lastOutcome` and forwarded to the next round, the map grows to include every unique peer that ever participated. This affects the serialized size of `GlobalConsensusOutcome` responses in `fetchOutcomeFromCluster`, and memory over a long-running network. The local `PeerQualityTracker` has a `decayThreshold` for this reason; `peerQuality` in the outcome needs the same treatment eventually. The `distinct` call can be replaced with set union (`previous.keySet | thisRoundQuality.keySet`).toList` to eliminate the quadratic behavior in the meantime.

scasplte2:
Closing in favor of #1485, which is the follow-on consensus rewrite that has been extensively tested in the fork-recovery work. Thanks @marcus-girolneto for the original direction here.

PR #1477, opened 2026-03-24
chore: add nightly terraform infrastructure

effects-ai:
Adds Terraform and CI infrastructure for a persistent nightly build cluster on Hetzner Cloud


nightly/terraform/main.tf
```diff
@@ -0,0 +1,150 @@
+terraform {
+  required_version = ">= 1.0"
+
+#   backend "s3" {
```
ryle-ai: The S3 backend is commented out. Deploying as-is stores state locally, which breaks any CI usage and makes shared team operation impossible. This should be uncommented and the bucket created before merging — or the PR should at minimum include a note explaining why local state is intentional for now.

nightly/terraform/main.tf
```diff
@@ -0,0 +1,150 @@
...
+      source_ips = var.allowed_ssh_cidrs
+    }
+  }
+
+  # Tessellation ports (GL0: 9000-9002, GL1: 9010-9012, ML0: 9020-9022, CL1: 9030-9032, DL1: 9040-9042)
+  # Allow from private network and allowed CIDRs
+  rule {
+    direction  = "in"
+    protocol   = "tcp"
+    port       = "9000-9099"
```
ryle-ai: Port 9090 (Prometheus) falls within this 9000–9099 range. The separate 9090-only rule below restricts Prometheus to the private subnet, but firewall rules are additive — the 9000–9099 rule still allows 9090 from `allowed_api_cidrs`. With the current defaults (`0.0.0.0/0`/`::/0`), Prometheus is publicly reachable. Either exclude 9090 from this range (use two rules: 9000–9089 and 9091–9099), or accept that Prometheus exposure is gated entirely on tightening `allowed_api_cidrs` before deployment.

nightly/terraform/components/cluster/variables.tf
```diff
@@ -0,0 +1,61 @@
...
+  description = "Server type for node machines (genesis + validators)"
+  type        = string
+}
+
+variable "streaming_server_type" {
+  description = "Server type for the snapshot streaming machine"
+  type        = string
+}
+
+variable "node_volume_size" {
```
ryle-ai: `node_volume_size`, `subnet_id` (line 32), and `deploy_ssh_public_key` (line 52) are declared here and passed from the root module, but none of them are referenced anywhere in `components/cluster/main.tf`. These are dead variables — either the cluster module should use them, or they should be removed.

nightly/terraform/README.md
```diff
@@ -0,0 +1,191 @@
...
+┌─────────────────────────────────────────────────────────────────┐
+│                  Hetzner Cloud Network (10.0.1.0/24)            │
+│                                                                 │
+│  ┌──────────────┐  ┌──────────────┐  ┌──────────────┐          │
+│  │  Machine 1   │  │  Machine 2   │  │  Machine 3   │          │
+│  │  Genesis     │  │  Validator 1  │  │  Validator 2  │          │
+│  │  10.0.1.11   │  │  10.0.1.12   │  │  10.0.1.13   │          │
+│  │              │  │              │  │              │          │
+│  │  GL0-0       │  │  GL0-1       │  │  GL0-2       │          │
+│  │  GL1-0       │  │  GL1-1       │  │  GL1-1       │          │
```
ryle-ai: Machine 3 shows `GL1-1` but should be `GL1-2`.

ryle-ai (changes requested):
`scripts/deploy-nightly.sh` is documented in the README (directory tree, prerequisites, usage, state management sections) but not included in the PR. The infrastructure can't be deployed using the documented workflow without it.

Also: README line 98 describes `main.tf` as using the shared `ded-terraform` bucket, but the commented-out backend block specifies `tessellation-nightly`. One of these is wrong.

nightly/terraform/templates/node-init.tpl
```diff
@@ -0,0 +1,72 @@
...
+# UFW firewall setup
+ufw default deny incoming
+ufw default allow outgoing
+
+# Base rules — all nodes
+ufw allow 22/tcp
+ufw allow 9000/tcp
+ufw allow 9001/tcp
+ufw allow 9010/tcp
+ufw allow 9011/tcp
```
ryle-ai, line 63: Only 4 ports opened for all nodes. Each machine runs GL0 (9000–9002), GL1 (9010–9012), ML0 (9020–9022), and CL1 (9030–9032). With `ufw default deny incoming` above, any Tessellation service port not listed here is blocked at the OS level — Hetzner's firewall allowance doesn't matter. Needs at minimum 9002, 9012, 9020–9022, 9030–9032, and 9002 covers GL0 internal gRPC which validators need to reach on each other.

nightly/terraform/components/cluster/main.tf
```diff
@@ -0,0 +1,107 @@
...
+  location    = var.location
+  ssh_keys    = var.ssh_key_ids
+  firewall_ids = var.firewall_ids
+
+  user_data = templatefile("${path.module}/../../templates/node-init.tpl", {
+    hostname  = "nightly-genesis"
+    ssh_keys  = var.team_ssh_keys
+    node_role = "genesis"
+    peer_ips  = concat(
+      [for v in hcloud_server.validators : v.ipv4_address],
```
ryle-ai: `ipv4_address` is the public IP. The README explicitly states inter-node communication uses Hetzner Cloud Network private IPs, so all validator→genesis and streaming→genesis traffic arrives from 10.0.1.x addresses. UFW `allow from <public_ip>` rules won't match. Use the private IPs instead: they're available as `cidrhost(var.subnet_cidr, 12)`, `cidrhost(var.subnet_cidr, 13)`, `cidrhost(var.subnet_cidr, 14)` without needing to reference the server resources, or build them from `hcloud_server_network.validators[*].ip` and `hcloud_server_network.streaming.ip`.

nightly/terraform/variables.tf
```diff
@@ -0,0 +1,75 @@
...
+  default     = "cpx41" # 8 AMD vCPU, 16 GB RAM
+}
+
+variable "streaming_server_type" {
+  description = "Hetzner server type for the snapshot streaming machine"
+  type        = string
+  default     = "cpx21" # 3 AMD vCPU, 4 GB RAM
+}
+
+variable "node_volume_size" {
```
ryle-ai: `node_volume_size` is never passed to the `cluster` module (it's absent from the `module "cluster"` call in `main.tf`) and the cluster module has no corresponding variable. Dead variable.

nightly/terraform/README.md
```diff
@@ -0,0 +1,191 @@
...
+- An SSH key pair for CI/CD deployment access
+
+## Usage
+
+### 1. Set environment variables
+
+```bash
+# Required
+export HCLOUD_TOKEN='your-hetzner-cloud-api-token'
+export TF_VAR_deploy_ssh_public_key='ssh-ed25519 AAAA... deploy@ci'
```
ryle-ai, line 91: "shared `ded-terraform` bucket" contradicts line 111 of this same file and the actual Terraform config, both of which use `tessellation-nightly`. Fix to `tessellation-nightly`.

nightly/terraform/README.md
```diff
@@ -0,0 +1,191 @@
...
+
+| Machine | Role | Server Type | vCPU | RAM | ~Cost/month |
+|---------|------|-------------|------|-----|-------------|
+| 1 | Genesis node + monitoring | CPX41 | 8 AMD | 16 GB | €15 |
+| 2 | Validator 1 | CPX41 | 8 AMD | 16 GB | €15 |
+| 3 | Validator 2 | CPX41 | 8 AMD | 16 GB | €15 |
+| 4 | Snapshot streaming | CPX21 | 3 AMD | 4 GB | €5 |
+| | **Total** | | | | **~€50** |
+
+All instances run Ubuntu 22.04 with Docker CE and the Docker Compose plugin pre-installed via cloud-init. Inter-node communication uses Hetzner Cloud Network private IPs. Tessellation services run with `--network host` so containers bind directly to the host interface — no Docker overlay networking required.
```
ryle-ai, line 43: Machine 2 shows `GL1-2`, Machine 3 shows `GL1-1`. Labels are swapped.

ryle-ai (changes requested):
Empty PR description for infrastructure that provisions a live €50/month cluster and a new S3 bucket. At minimum: why now, what triggers the nightly deploys, and what the expected initial deployment sequence is.

**`scripts/deploy-nightly.sh` is missing.** The README documents it as the primary interface for every operation (plan, deploy, output, destroy). The file is referenced in the directory tree, prerequisites, usage instructions, and state management sections. The DED equivalent (`deploy-terraform.sh`) exists in that repo but isn't adapted and included here. The Terraform config is self-contained and functional, but the documented workflow can't actually run.

**UFW rules in `node-init.tpl` are significantly incomplete.** The template opens 9000, 9001, 9010, 9011 as "base rules for all nodes", but each machine runs four service layers. Genesis runs GL0-0, GL1-0, ML0-0, CL1-0 — which means ports 9002, 9012, 9020–9022, 9030–9032 (at minimum) also need to be open. With `ufw default deny incoming`, UFW will silently block inbound traffic on all unlisted ports regardless of what the Hetzner firewall allows. Services bind and listen; connections are just dropped at the OS level. This affects all nodes.

**`peer_ips` uses public IPs but all inter-node traffic flows over the private network.** In `components/cluster/main.tf`, the genesis template call populates `peer_ips` from `hcloud_server.validators[*].ipv4_address` and `hcloud_server.streaming.ipv4_address` — these are the public (internet-facing) IPv4 addresses. The UFW rules generated from these (`ufw allow from <public_ip>`) will never match traffic that arrives via the Hetzner Cloud Network private interface, where source IPs are 10.0.1.12, 10.0.1.13, 10.0.1.14. The result: the ClickHouse (8123) and Prometheus scrape (9090) access rules on genesis silently don't work. The private IPs are available on the `hcloud_server_network` resources, not `hcloud_server.ipv4_address`.

ryle-ai (changes requested):
## Summary

Empty PR description for infrastructure that provisions a €50/month live cluster. At minimum: what secrets need to be configured in GitHub before this can run, what the expected initial deployment sequence is, and what monitoring data this captures.

### Missing: `scripts/deploy-nightly.sh`

This was raised in the previous review round and is still not included. The README devotes four sections to this script (directory tree, prerequisites, usage, state management), all referencing `./scripts/deploy-nightly.sh` as the primary interface for every terraform operation. The file does not exist in the PR. The terraform config itself is self-contained and will work with plain `terraform` commands, but the documented workflow can't run.

### ML0/CL1 not started by either deploy or rebuild

The README architecture diagram explicitly shows ML0 and CL1 running on genesis and both validators. Neither `docker/bin/remote-deploy.sh` nor `nightly-rebuild.yml` starts these layers — both only start `gl0` and `gl1`. The metagraph compose file (`docker-compose.metagraph.yaml`) exists and defines the services, but it's never included in the remote COMPOSE variable and the metagraph profiles are never activated.

Consequence: after a deploy or nightly rebuild, the cluster is GL0 + GL1 only. The E2E tests (`dag-cluster`, `delegated-staking`, `token-lock-replacement`, `snapshot-streaming`) don't require ML0/CL1, so this doesn't break CI. But the README's architecture claims a four-layer stack that never actually runs. Either the diagram needs to be updated to match what's deployed, or the deploy/rebuild scripts need to start ML0/CL1.

The `nightly-rebuild.yml` stop step (`--profile l0 --profile l1 down`) also doesn't stop ML0/CL1 containers or wipe their data directories, so if they were ever added, the rebuild would leave them in a dirty state.

### ClickHouse TLS private key in version control

A real RSA private key is committed at `docker/monitoring/clickhouse/certs/server.key`. It's self-signed (CN=clickhouse) and is only mounted in `docker-compose.local.yaml` — the remote nightly deployment runs ClickHouse on plain HTTP (no cert mounted in `docker-compose.remote.yaml`). So there's no immediate operational impact. But private keys in git create problems: the key is in history permanently, future automated scanners will flag it, and it normalizes a pattern that shouldn't extend to non-self-signed material. Generate the cert on the fly in `docker-compose.local.yaml` or add the `certs/` directory to `.gitignore`.

### `increase_delegated_stakes` as hardcoded default branch

`nightly-deploy.yml` lines 84 and 208 fall back to `--block-explorer-branch=increase_delegated_stakes` when `BLOCK_EXPLORER_BRANCH` is unset. This is a feature branch. When it gets deleted (which is when feature branches go), builds will silently fail or pull unexpected code. The fallback should be a stable branch (`main`, `testing`, etc.).

docker/monitoring/clickhouse/certs/server.key
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
ryle-ai: Real RSA private key committed to git. Even though it's self-signed and only mounted in `docker-compose.local.yaml`, it's now in git history permanently. Generate on the fly or exclude from version control.

nightly/terraform/README.md
```diff
@@ -0,0 +1,191 @@
...
+│                                                                 │
+│  ┌──────────────┐                                               │
+│  │  Machine 4   │                                               │
+│  │  Streaming   │                                               │
+│  │  10.0.1.14   │                                               │
+│  └──────────────┘                                               │
+└─────────────────────────────────────────────────────────────────┘
+```
+
+| Machine | Role | Server Type | vCPU | RAM | ~Cost/month |
```
ryle-ai, line 35: The table labels Machine 1 as "Genesis node + monitoring", and the diagram above shows Grafana/Prometheus on Machine 1. But `deploy-monitoring.sh` deploys the monitoring stack to `ALL_NODES[3]` — Machine 4 (streaming). The comment in `components/cluster/main.tf:2` also says `# Runs GL0-0, GL1-0, ML0-0, CL1-0 + Grafana/Prometheus` for the genesis resource. One of the two is wrong; the deploy script explicitly says "Deploys to the last node in REMOTE_NODES (n3)" so the docs are the likely error.

.github/workflows/nightly-deploy.yml
```diff
@@ -197,11 +197,30 @@ jobs:
...
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CLICKHOUSE_HOST: ${{ secrets.NIGHTLY_CLICKHOUSE_HOST }}
+          CLICKHOUSE_PORT: ${{ secrets.NIGHTLY_CLICKHOUSE_PORT }}
+          CLICKHOUSE_USER: ${{ secrets.NIGHTLY_CLICKHOUSE_USER }}
+          CLICKHOUSE_PASSWORD: ${{ secrets.NIGHTLY_CLICKHOUSE_PASSWORD }}
+          CLICKHOUSE_DATABASE: ${{ secrets.NIGHTLY_CLICKHOUSE_DATABASE }}
         run: |
           IFS=',' read -ra IPS <<< "${{ vars.NIGHTLY_HOSTS }}"
           NODES=$(printf "n%s," "${!IPS[@]}" | sed 's/,$//')
           just up --skip-assembly --remote="$NODES" --snapshot-streaming-branch=${{ vars.SNAPSHOT_STREAMING_BRANCH || 'testing' }} --block-explorer-branch=${{ vars.BLOCK_EXPLORER_BRANCH || 'increase_delegated_stakes' }}
```
ryle-ai: Fallback to `increase_delegated_stakes`, a feature branch. When this branch is deleted, nightly deploys silently pull wrong code or fail. This was flagged in the previous review round. Use a stable branch (`main`, `testing`).

docker/monitoring/deploy-monitoring.sh
```diff
@@ -0,0 +1,115 @@
...
+    jsonData:
+      host: localhost
+      port: 8123
+      protocol: http
+      secure: false
+      username: default
+      defaultDatabase: default
+    secureJsonData:
+      password: clickhouse
+EOF
```
ryle-ai, line 82: Grafana datasource password hardcoded as `clickhouse` in the heredoc. If the ClickHouse container password is ever changed in `docker-compose.remote.yaml`, this heredoc must be updated too — and since it's written at deploy time, the datasource will silently break. Also: the `CLICKHOUSE_PASSWORD` env var passed to this script from the workflow (`${{ secrets.NIGHTLY_CLICKHOUSE_PASSWORD }}`) is never used here — the script always writes `password: clickhouse` regardless.

docker/monitoring/deploy-monitoring.sh
```diff
@@ -0,0 +1,115 @@
...
+
+# --- .env ---
+ssh "$SERVER_NODE" "echo 'GRAFANA_ADMIN_PASSWORD=${GRAFANA_ADMIN_PASSWORD:-admin}' > $REMOTE_DIR/.env"
+
+# --- Start services ---
+log "Pulling images and starting services"
+ssh "$SERVER_NODE" "cd $REMOTE_DIR && docker compose pull -q && docker compose up -d"
+
+# --- Wait for ClickHouse to be ready ---
+log "Waiting for ClickHouse"
```
ryle-ai, line 101: If ClickHouse never becomes ready (image pull failure, OOM on the 4 GB streaming machine, etc.), this loop exits after 60 seconds without failing the script. The subsequent `init.sql` command will fail, but `|| true` swallows the error. The deploy reports success while the monitoring stack is non-functional. Add `|| { log "ERROR: ClickHouse failed to start"; exit 1; }` after the loop.

ryle-ai (changes requested):
Empty PR description for infrastructure that provisions a live €50/month cluster.

### Private RSA key committed to git

`docker/monitoring/clickhouse/certs/server.key` is a real RSA private key. It's self-signed (CN=clickhouse) and only mounted in `docker-compose.local.yaml` — the remote deployment doesn't use it — so there's no immediate operational impact. But the key is now in git history permanently. Automated secret scanners on forks and CI systems will flag it. The fix is straightforward: generate the cert on the fly in a `docker-compose.local.yaml` entrypoint, or add `certs/` to `.gitignore` and document the one-time generation step.

### `scripts/deploy-nightly.sh` still missing

This was flagged in both previous review rounds and is still absent. The README devotes four sections to this script (directory tree, prerequisites, usage, state management). It's described as the primary interface for every terraform operation. The terraform configuration itself is self-contained and works with plain `terraform` commands, but the documented workflow can't run. Either add the script or update the README to remove all references to it.

### Monitoring location: README says Machine 1, deploy script targets Machine 4

`deploy-monitoring.sh` opens with `# Deploys to the last node in REMOTE_NODES (n3)` and uses `SERVER_NODE="${ALL_NODES[3]:-${ALL_NODES[2]}}"` — this is Machine 4 (streaming, CPX21 with 3 vCPU / 4 GB). But the README architecture diagram shows Grafana and Prometheus on Machine 1 (genesis), the machine table labels Machine 1 as "Genesis node + monitoring", and the `cluster/main.tf` comment reads `# Runs GL0-0, GL1-0, ML0-0, CL1-0 + Grafana/Prometheus`. Three separate places say Machine 1; the code says Machine 4. Prometheus targets are built from NODE_IPS[0..2] (genesis + validators), which makes sense if monitoring is on a fourth machine — so the code is probably right and the docs need updating.

### ML0/CL1 not started

This was raised in the previous review round and remains unaddressed. The README architecture diagram shows ML0 and CL1 running on genesis and both validators. Neither the `Deploy cluster` step (which calls `just up ... gl0 gl1`) nor `nightly-rebuild.yml` starts these layers. The E2E test suite doesn't require ML0/CL1 so CI passes, but the README describes a four-layer stack that never actually runs. Either update the diagram to match what's deployed, or start the metagraph layers in the deploy/rebuild scripts.

### `increase_delegated_stakes` as the block-explorer fallback branch

This was flagged in the previous review round. When `BLOCK_EXPLORER_BRANCH` is unset the fallback is a feature branch. When that branch is deleted, nightly deploys will fail or pull stale code. The fallback should be a stable branch (`main`, `testing`).

docker/monitoring/grafana/provisioning/datasources/datasources.yaml.tpl
```diff
@@ -0,0 +1,25 @@
...
+    editable: true
+    jsonData:
+      host: ${CLICKHOUSE_HOST}
+      port: ${CLICKHOUSE_PORT}
+      protocol: https
+      secure: true
+      username: ${CLICKHOUSE_USER}
+      defaultDatabase: ${CLICKHOUSE_DATABASE}
+    secureJsonData:
+      password: ${CLICKHOUSE_PASSWORD}
```
ryle-ai, line 25: These `${CLICKHOUSE_HOST}`, `${CLICKHOUSE_PASSWORD}`, etc. placeholders won't be expanded by Grafana — it treats them as literal strings. The deploy script correctly bypasses this file by writing its own datasources config via a heredoc, and the local compose uses `datasources.local.yaml`. So this file is never loaded in either deployment path. Remove it, or rename to `datasources.yaml.tpl` to make clear it requires shell substitution before use.

ryle-ai (changes requested):
The PR has made real progress since earlier rounds — UFW is now complete (9000–9002, 9010–9012, 9020–9022, 9030–9032), peer IPs use private addresses from `cidrhost()`, the S3 bucket name is consistent, the monitoring location is correct in the README, the ClickHouse readiness check exits on failure, the `increase_delegated_stakes` fallback is fixed, and the RSA key is no longer committed. These are meaningful fixes.

Two items that have been flagged in all four previous rounds are still missing:

**`scripts/deploy-nightly.sh` is not in this PR.** The README devotes four sections to it (directory tree, prerequisites, usage, state management), describing it as the primary interface for every terraform operation. The directory appears in the repo tree documentation but doesn't exist. The terraform itself is functional with plain `terraform` commands, but the documented workflow can't run. Either add the script or remove all references to it from the README.

**Empty PR description.** Four rounds of review on infrastructure that provisions a live €50/month cluster, with no description of what's being set up, what GitHub secrets must be configured before it can run, or what the expected initial deployment sequence is.

**ML0/CL1 discrepancy** (flagged rounds 3–4, non-blocking): The README architecture diagram shows ML0 and CL1 running on all three nodes. Neither the deploy workflow (`just up ... gl0 gl1`) nor the nightly rebuild starts these layers. E2E tests don't require them so CI passes, but the documented architecture doesn't reflect what actually runs. Either start the metagraph layers in the deploy/rebuild scripts, or update the diagram.

docker/monitoring/grafana/provisioning/datasources/datasources.yaml.tpl
```diff
@@ -0,0 +1,25 @@
+apiVersion: 1
```
ryle-ai, line 1: Shell placeholders like `${CLICKHOUSE_HOST}` won't be expanded by Grafana — Grafana's provisioning system treats them as literal strings. Neither deployment path loads this file: `deploy-monitoring.sh` writes its own datasources config via a heredoc (bypassing this file entirely), and local compose mounts `datasources.local.yaml` directly. This file is dead configuration and should be removed.

docker/monitoring/grafana/provisioning/dashboards/tessellation.json
```diff
@@ -0,0 +1,1904 @@
...
+        }
+      },
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "prometheus"
+          },
+          "editorMode": "code",
+          "expr": "dag_global_snapshot_ordinal{peer_id_short=~\"e0c1ee6e|629880a5|710b3dc5\"}",
```
ryle-ai, line 497: Hardcoded `peer_id_short=~"e0c1ee6e|629880a5|710b3dc5"` matches the mainnet priority peer IDs. Nightly cluster nodes have different identities so this panel returns no data. Use a Grafana template variable backed by the available label values instead.

docker/monitoring/grafana/provisioning/dashboards/tessellation.json
```diff
@@ -0,0 +1,1904 @@
...
+      "showTimeAxis": false,
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "prometheus"
+          },
+          "editorMode": "code",
+          "exemplar": false,
+          "expr": "rate(dag_consensus_proposal_affinity_sum{peer_id_short=~\"e0c1ee6e|629880a5|710b3dc5\"}[90s])/rate(dag_consensus_proposal_affinity_count{peer_id_short=~\"e0c1ee6e|629880a5|710b3dc5\"}[90s])",
```
ryle-ai, line 900: Same hardcoded mainnet peer ID filter. Same issue as line 497.

docker/monitoring/clickhouse/init.sql
```diff
@@ -0,0 +1,27 @@
...
+CREATE TABLE IF NOT EXISTS nightly_logs (
+    timestamp DateTime64(3),
+    node_id LowCardinality(String),
+    network_id LowCardinality(String),
+    log_type LowCardinality(String),
+    data JSON
+) ENGINE = MergeTree()
+PARTITION BY (network_id, toYYYYMM(timestamp))
+ORDER BY (node_id, timestamp)
+TTL toDateTime(timestamp) + INTERVAL 3 DAY;
```
ryle-ai, line 13: `INTERVAL 3 DAY` is hardcoded here, but `ClickHouseSink.createTableDDL` uses the `retention-period-in-days` value from config (default: 90 days). Since `init.sql` runs via `docker-entrypoint-initdb.d` before any node connects, nodes find the table already exists and `CREATE TABLE IF NOT EXISTS` is a no-op — the config value is never applied for nightly deployments. Operators who adjust `retention-period-in-days` won't see it take effect. A comment noting this (and that TTL changes require `ALTER TABLE ... MODIFY TTL`) would prevent confusion.

ryle-ai (changes requested):
Previous rounds addressed the major infrastructure issues (UFW completeness, private IP references, S3 backend, RSA key, port-overlap, monitoring location). Two items flagged since round 1 remain unresolved.

**`scripts/deploy-nightly.sh` still missing.** The README documents it in the directory tree, prerequisites, usage, and state management sections as the primary interface for every terraform operation. The terraform config works standalone with plain `terraform` commands, but the documented quick-start workflow can't run.

**Empty PR description.** Five rounds of review on infrastructure that provisions a live €50/month cluster.

**New issues in this commit (grafana/clickhouse additions):**

`datasources.yaml.tpl` contains shell-syntax placeholders (`${CLICKHOUSE_HOST}`, `${CLICKHOUSE_PASSWORD}`, etc.) that Grafana's provisioning system treats as literal strings — it doesn't shell-expand them. Neither deployment path loads this file: `deploy-monitoring.sh` writes its own datasources config via heredoc; local compose mounts `datasources.local.yaml` directly. The file should be removed, or explicitly named/documented as requiring shell preprocessing before deployment.

`tessellation.json` hardcodes `peer_id_short=~"e0c1ee6e|629880a5|710b3dc5"` in Prometheus expressions at lines 497 and 900. These are the first 8 hex characters of the mainnet priority peer IDs. The nightly cluster will use different node identities — these panels return empty data on any nightly deployment. They should be Grafana template variables rather than hardcoded IDs.

**Non-blocking (ongoing since round 3):** README architecture diagram shows ML0 and CL1 running on genesis and both validators; neither `just up` nor `nightly-rebuild.yml` starts these layers.

ryle-goehausen (approved):

PR #1483, opened 2026-04-01
fix: escalate isolated node to recovery

scasplte2:
When peer registrations are stale the lagging gate (229972f62) blocked escalation when peersAtHigherKey=0, but this fires incorrectly for isolated nodes: when cut off, peers stop sending declarations so registrations freeze at old ordinals, making the gate see zero peers ahead even though the network has moved on.

Distinguish isolation from kill4 deadlock by checking peersAtSameKey:
- If peers are registering the current key -> cluster is stuck together (kill4 scenario) -> suppress escalation, keep retrying
- If no peer has the current key registered -> registrations are stale (isolated node) -> escalate to recovery download

The kill4 protection is preserved: when 4 of 8 nodes lose quorum and all are stuck on ordinal N, every node sees peersAtSameKey > 0 and suppresses escalation, preventing the WaitingForDownload deadlock.

### DIFF
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/AbandonmentTracker.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/AbandonmentTracker.scala
@@ -268,10 +268,21 @@ class AbandonmentTracker[F[_]: Async: Metrics, Event, Key: Order, Artifact, Ctx,
                     val readyPeerIds = responsivePeers.filter(_.state === NodeState.Ready).map(_.id).toSet
                     val readyPeerRegs = peerRegs.view.filterKeys(readyPeerIds.contains).toMap
                     val peersAtHigherKey = readyPeerRegs.count { case (_, peerKey) => peerKey > key }
+                    val peersAtSameKey = readyPeerRegs.count { case (_, peerKey) => peerKey === key }
                     val totalReadyPeers = readyPeerRegs.size
                     val peersMajorityAhead = totalReadyPeers >= 2 && peersAtHigherKey > totalReadyPeers / 2
-                    if (peersMajorityAhead)
-                      // Peers have moved on — this node is genuinely behind. Trigger recovery.
+                    // Isolated node: all registrations are stale (below current key) because
+                    // peers stopped sending declarations when we were cut off. Registrations
+                    // never advance past our own current ordinal, so peersAtHigherKey=0 even
+                    // though the network has moved on. Distinguish this from kill4 (whole
+                    // cluster stuck at same key) by checking if ANY peer has our key registered.
+                    // If zero peers are at our key AND zero are ahead, registrations are stale →
+                    // treat as isolated/lagging and escalate to recovery.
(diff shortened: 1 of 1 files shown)


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/consensus/engine/AbandonmentTracker.scala
```diff
@@ -268,10 +268,21 @@ class AbandonmentTracker[F[_]: Async: Metrics, Event, Key: Order, Artifact, Ctx,
...
-                      // Peers have moved on — this node is genuinely behind. Trigger recovery.
+                    // Isolated node: all registrations are stale (below current key) because
+                    // peers stopped sending declarations when we were cut off. Registrations
+                    // never advance past our own current ordinal, so peersAtHigherKey=0 even
+                    // though the network has moved on. Distinguish this from kill4 (whole
+                    // cluster stuck at same key) by checking if ANY peer has our key registered.
+                    // If zero peers are at our key AND zero are ahead, registrations are stale →
+                    // treat as isolated/lagging and escalate to recovery.
+                    val peersHaveCurrentKey = peersAtSameKey > 0
+                    val shouldEscalateAsLagging = peersMajorityAhead || (!peersHaveCurrentKey && totalReadyPeers > 0)
```
ryle-ai, line 282: The kill-4 guard here relies on `peersAtSameKey > 0` to suppress escalation, but this assumption breaks down for any long-running cluster.

`peerRegistrationsR` is only updated when a peer enters `NodeState.Observing` (via `peerRegistrationStream` → `collectRegistration`). In a cluster that has been running for many rounds without a recovery event, all peer registrations are permanently stale at the initial join-time key. If the cluster joined at round K and is now stuck at round K+100, every peer's registration is K. `peersAtSameKey = count(reg === K+100) = 0` — indistinguishable from the isolation case.

The comment in `StallDetector.scala` lines 178–182 already documents this: `"Peers at the same or lower keys are stale registrations (e.g. from PeerObserved using lastOutcome key, which is always one behind the current round key)."` That warning applies here too.

In practice the timing difference between nodes (different retry counters, jitter) makes the simultaneous-escalation scenario unlikely — one node triggers first and downloads from the remaining Ready peers. But the kill-4 protection as stated in the comment is not reliable: `peersAtSameKey > 0` is only guaranteed to be non-zero if at least one peer transitioned through Observing while running the current key, which requires a recent recovery event.

ryle-ai (commented):
The isolated-node escalation logic fixes a real gap: under the original `peersMajorityAhead` check, a node stuck at key K while the network has moved on to K+N would never escalate, because stale peer registrations show no one ahead. The new `!peersHaveCurrentKey && totalReadyPeers > 0` arm correctly fires in that scenario.

However, the kill-4 protection claim in the comment is based on an assumption that doesn't hold in practice (see inline comment).

Also: this PR has an empty description for a change that touches consensus recovery safety logic. Given the explicit kill-4 protection that already lives in this code, the rationale, the tradeoff, and how this was tested in a network scenario should be documented.

effects-ai (approved):

PR #1476, opened 2026-03-17
feat: event mempool fork recovery

scasplte2:
## Event Mempool, Gossip Mesh, and Fork Recovery

Replaces the implicit timeout-based consensus protocol with an explicit leader-based engine featuring structured stall detection, gossip-driven event propagation, and automated fork recovery.

---

## Design Overview

Consensus readme with figures[ available here](https://github.com/Constellation-Labs/tessellation/blob/eafe895d06341f9ca5b254081a556401ec50bd20/docs/consensus/README.md)

### Consensus Engine Rewrite

**Leader-based round protocol** — deterministic leader selection via rendezvous hashing; explicit view changes on leader failure rather than silent round restart.

**Require-all declarations** — the advancer waits for 100% of active facilitators to declare before phase transition (no quorum threshold). The `StallDetector` handles unresponsive peers via a graduated escalation chain (warn → evict → abandon), with a Ready-peer quorum floor of `floor(N/2)+1` to prevent minority forks.

**StallDetector** — monitors round progress through all phases (Facilities → Proposals → Signatures), graduated response (warn first timeout cycle, evict on second). Quorum floor anchored to Ready-peer count from `clusterStorage` — prevents the StallDetector from lowering its own quorum floor as peers are evicted (which would allow a 2-node fork to become self-sustaining).

**AbandonmentTracker** — tracks consecutive round failures with a typed `AbandonReason` ADT distinguishing retriable (`QuorumInfeasible`) from non-retriable (`Lagging`, `MaxStalls`, etc.). Stale-ordinal escalation: if N retriable abandonments accumulate at the same ordinal AND a majority of Ready peers are at a higher ordinal, escalates to non-retriable and triggers recovery download. This prevents spurious recovery downloads when the whole cluster is symmetrically stuck (kill-4 scenario).

**TrailingCommonAncestor (TCA) filter** — proof-based peer quality tracking. Penalizes non-signers and evicted peers for `removalPenaltyRounds` to prevent immediate re-selection of unresponsive facilitators, eliminating the ~35s/round waste of re-detecting the same stall.

**Dynamic `minViableQuorum`** — `max(3, floor(N/2)+1)` in `ConsensusStateCreator`. Prevents TCA penalties from thinning the facilitator pool below majority threshold in larger clusters.

[...]

PR #1484, opened 2026-04-02
fix: skip deferral when cluster advanced past downloaded ordinal, fix fork recovery race

scasplte2:
Two fixes for post-fork-recovery stall bugs observed on testnet after PR #1476.

**Fix 1: StateTransitions — accept newer consensus outcome (eliminates DEFERRED loop)**

When initFromDownload fetches the consensus outcome from peers and the cluster has already advanced past the downloaded ordinal, it now accepts the newer outcome and skips the 43s TimeTrigger deferral. Previously the node rejected the newer outcome, exhausted 20 retries, and looped back into DOWNLOAD_INIT_DEFERRED permanently.

**Fix 2: ForkRecoveryService — set recovery flags before state transition**

Sets setRecoveryDownload and setPreferredPeers before transitioning to WaitingForDownload, closing a race where DownloadDaemon could see the state change before the flags were set (causing isRecovery=false and a 43s deferral).

Also adds clearPreferredPeers to RecoveryPeerHint to roll back flags if the transition fails.


ryle-ai (commented):
## Summary

Large PR introducing leader-based proposal resolution and event-mempool consensus for currency-l0, with fork recovery race fix and the deferral-skip for nodes joining a cluster that has advanced past their downloaded ordinal. The architecture is well-considered and the `snapshotHash` non-determinism fix (using `artifact.value.hash` instead of `signedArtifact.hash`) is correct and necessary. Two issues worth addressing before these patterns stabilise in production.

---

### 1. `alreadyWithdrawn` guard can deadlock the new leader after a view change

`advanceFromProposals` gates re-entry using:
```scala
val alreadyWithdrawn =
  resources.withdrawalsMap.get(selfId).contains(CurrencyConsensusKind.Signature: CurrencyConsensusKind) ||
    state.withdrawnFacilitators.value.contains(selfId)
```
The guard was added to prevent the hot-loop where a failing non-leader repeatedly re-validates the leader's proposal. The problem is that `withdrawalsMap` is keyed by peer and stores exactly one `Kind` per peer (`Map[PeerId, Kind]`). When a follower fails `validateLeaderArtifact` and sends `ConsensusWithdrawPeerDeclaration(key, CurrencyConsensusKind.Signature)`, its entry becomes `Signature`. But `updateFacilitators` in `CollectingProposals` removes peers only when their withdrawal kind matches `maybeCollectingKind(CollectingProposals) = Proposal`. So a `Signature` withdrawal does **not** move the peer to `withdrawnFacilitators` — the peer stays in `state.facilitators`.

After a view change (StallDetector evicts the original leader), the withdrawn-for-Signature peer can be selected as the new leader via `selectLeaderWeighted`. As new leader, it hits `advanceFromProposals` with:
- `maybeLeaderProposal = None` (it hasn't spread a proposal yet)
- `alreadyWithdrawn = true` (Signature withdrawal still in resources — resources are not cleared on view change, only on round abandonment)

Result: the new leader returns `none[Transition]` without spreading a proposal. Peers stall indefinitely, triggering another eviction cycle. Eventually `AbandonmentTracker` abandons the round, but this wastes several stall-detection intervals (tens of seconds each) in an already-failed scenario.

The fix is to skip the guard when `selfId === state.leader && maybeLeaderProposal.isEmpty` — a leader that hasn't spread a proposal yet must always be allowed to do so, regardless of prior withdrawal. The hot-loop concern only applies to non-leaders repeatedly failing validation on an already-received proposal.

---

### 2. `fetchEventFromPeer` returns true on response presence, not on successful `mempool.add`

The docstring says "Returns true if the event was successfully added to the mempool" but the actual return is `response.events.exists(_._1 === hash)` — true if the hash appears in the response, regardless of whether `mempool.add` succeeded. The `traverse_` discard means a `Left(MempoolFull)` result is silently ignored.

Consequence: if the mempool is at capacity (100 000 events), a `fetchEventFromPeer` call for a missing hash will return `true` even though the event was never stored. `syncMissingEvents` short-circuits via `foldLeftM` on the first `true`, skips remaining peer fallbacks, and `buildProposalTransition` proceeds without the event — the event is silently omitted from the round's hash union. Capacity of 100 000 makes this unlikely under normal load, but worth tightening: the return should check that the specific hash was actually added (e.g., check `mempool.get(hash).isDefined` after the add, or capture the `add` results instead of discarding with `void`).

PR #1486, opened 2026-04-05
fix: filter acceptedTokenUnlocks by address in balance calculation

marcus-girolneto:
## Summary
- Token unlocks in `updateBalancesByTokenLocks` were applied to **every** address in the fold instead of only the address that owns the unlock, inflating balances incorrectly
- Added `acceptedTokenUnlocks.filter(_.source == address)` before the fold, matching the fix already present in v4.0.0+ (`TokenLockOpsManager.scala`)

### DIFF
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
@@ -1123,8 +1123,9 @@ object CurrencySnapshotAcceptanceManager {
               }
             }
 
+            unlocksForAddress = acceptedTokenUnlocks.filter(_.source == address)
             finalBalance <-
-              acceptedTokenUnlocks.foldLeft[Either[BalanceArithmeticError, Balance]](Right(expiredBalance)) {
+              unlocksForAddress.foldLeft[Either[BalanceArithmeticError, Balance]](Right(expiredBalance)) {
                 case (currentBalanceEither, tokenUnlock) =>
                   for {
                     currentBalance <- currentBalanceEither


ryle-goehausen (approved):

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/CurrencySnapshotAcceptanceManager.scala
```diff
@@ -1123,8 +1123,9 @@ object CurrencySnapshotAcceptanceManager {
               }
             }
 
+            unlocksForAddress = acceptedTokenUnlocks.filter(_.source == address)
             finalBalance <-
-              acceptedTokenUnlocks.foldLeft[Either[BalanceArithmeticError, Balance]](Right(expiredBalance)) {
+              unlocksForAddress.foldLeft[Either[BalanceArithmeticError, Balance]](Right(expiredBalance)) {
                 case (currentBalanceEither, tokenUnlock) =>
                   for {
                     currentBalance <- currentBalanceEither
```
ryle-ai, line 1131: The `acceptTokenLocks` private method (line ~1070) uses `acceptedTokenUnlocks.map(_.tokenLockRef)` without address filtering, but that is correct: it's building a set of hash references to check whether a given token lock should be removed from the active set. Hash uniqueness ensures no cross-address contamination there. Only this balance-credit path needed the address filter.

ryle-ai (commented):
Fix is correct. `acceptedTokenUnlocks` is a flat `SortedSet[TokenUnlock]` across all addresses, and the outer fold in `updateBalancesByTokenLocks` iterates per-address. Without this filter, every address with any token lock activity (new locks or expired locks) would have the full sum of *all* accepted unlocks added to its balance — not just its own. Severity is high: in any snapshot with multi-address token unlock activity, balances would be inflated by `sum(all_unlock_amounts) × (addresses_in_fold - 1)`. The `TokenLockOpsManager.updateBalancesByTokenLocks` (added in #1353 refactor) already had `filter(_.source == address)` correct; `CurrencySnapshotAcceptanceManager`'s private duplicate missed it.

Two things worth noting:

**No PR description.** For a hotfix to mainnet that corrects a balance accounting error, a brief description of the impact (when does this trigger, which addresses are affected, any snapshots that need inspection) would be useful for auditors and anyone reviewing the release.

**Test gap.** The existing `updateBalancesByTokenLocks` tests in `TokenLockOpsManagerSuite` only cover single-address scenarios where `tokenUnlock.source == address`. The bug would have been immediately caught by a test with two addresses where address B has token unlocks but address A does not, asserting that A's balance is unaffected. That scenario is still not covered by the test suite after this fix.

PR #1487, opened 2026-04-07
fix: retroactive balance deductions for token unlock bug

marcus-girolneto:
## Summary
- Adds retroactive balance deductions for **419 addresses** incorrectly credited by the token unlock address filter bug (PR #1486)
- Bug occurred at SWAP metagraph ordinals **467032** (20,022.73 SWAP) and **467642** (10,000 SWAP)
- Reuses existing `adjustments.json` mechanism with floor-to-zero logic
- Adds `TokenUnlockBugDeduction` reason to `BalanceAdjustmentReason`

## Details
- Metagraph: `DAG7X5idd4aLfp4XC6WQdG1eDfR3LGPVEwtUUB2W`
- Snapshot ordinal: **placeholder `999999`** — must be updated before deployment
- 2 source addresses excluded from deductions
- Deduction amounts: 30,022.73 SWAP (both ordinals) or 10,000/20,022.73 SWAP (single ordinal)

### DIFF
--- a/modules/node-shared/src/main/resources/adjustments.json
+++ b/modules/node-shared/src/main/resources/adjustments.json
@@ -1485,6 +1485,12 @@
         "reference": []
       },
       {
+        "address": "DAG15MGPwnPqemrgUkZCaYeukC63u1UYaq3WXze1",
+        "reason": "TokenUnlockBugDeduction",
+        "reference": [],
+        "deduct": -1000000000000
+      },
+      {
         "address": "DAG15WU5REVgZTWgMTzKPpeEdQvu8Ct3wDKWG7RA",
         "reason": "TokenUnlockBugDeduction",
         "deduct": -3002272560617,
@@ -1801,6 +1807,12 @@
         "reason": "TokenUnlockBugDeduction",
         "deduct": -3002272560617,
         "reference": []
+      },
+      {
+        "address": "DAG2UuPh5oYwfE5yMTivz1PJGhqfST2Y439DXRBA",
+        "reason": "TokenUnlockBugDeduction",
+        "reference": [],
+        "deduct": -2002272560617
       },
       {
         "address": "DAG2WA5YDY2jWPF81W1DLo37DUj5C9B6oUFznktS",
--- a/modules/node-shared/src/main/resources/adjustments.json
+++ b/modules/node-shared/src/main/resources/adjustments.json
@@ -4116,15 +4116,6 @@
         ]
       },
       {
-        "address": "DAG6wNGHp6hAjnAyhkdTJfW62YNMfaGb1sVpsKoR",
-        "reason": "TokenUnlockBugDeduction",
-        "reference": [
-          "bfd733e92aa71b2186955a4d7186301b27602621f0ee890159a8678032769f11",
-          "dbe69291991d4495ead034a1815c8d9d3b6c9ea20e5c6873136052e89601da11"
-        ],
-        "deduct": -3002272560617
-      },
-      {
         "address": "DAG6wyV27YvGGAsixbdhqqmZHMKSMxwbKYAYvgB6",
         "reason": "TokenUnlockBugDeduction",
         "deduct": -3002272560617,
@@ -4627,15 +4618,6 @@
           "bfd733e92aa71b2186955a4d7186301b27602621f0ee890159a8678032769f11",
           "dbe69291991d4495ead034a1815c8d9d3b6c9ea20e5c6873136052e89601da11"
         ]
-      },
-      {
-        "address": "DAG8GYFbgQUSYBxf5gWELhrzF6EJ6bZAzZETsXND",
-        "reason": "TokenUnlockBugDeduction",
-        "reference": [
-          "bfd733e92aa71b2186955a4d7186301b27602621f0ee890159a8678032769f11",
-          "dbe69291991d4495ead034a1815c8d9d3b6c9ea20e5c6873136052e89601da11"
-        ],
-        "deduct": -3002272560617
       },
       {
         "address": "DAG8H8JQf8zWEZXGZ5eckNBS3Z3tyuM4zhPeB4iL",
--- a/modules/node-shared/src/main/resources/adjustments.json
+++ b/modules/node-shared/src/main/resources/adjustments.json
@@ -1185,5 +1185,2525 @@
         "deduct": -22837225000000.0
       }
     ]
+  },
+  {
+    "currencyId": "DAG7X5idd4aLfp4XC6WQdG1eDfR3LGPVEwtUUB2W",
+    "snapshotOrdinal": 999999,
+    "adjustments": [
+      {
+        "address": "DAG042v1StpF7dDCzpkyQHkwt4JEFvpVZxYUvvDp",
+        "reason": "TokenUnlockBugDeduction",
+        "deduct": -3002272560617,
+        "reference": []
+      },
+      {
+        "address": "DAG04AL8jRLmqgL2j3WRGvjYgU4qwPxJbUE5PWV1",
+        "reason": "TokenUnlockBugDeduction",
+        "deduct": -3002272560617,
+        "reference": []
+      },
+      {
+        "address": "DAG06XxKVkSzaoYDajdniWz1PFQsEac3sq8XVDrD",
+        "reason": "TokenUnlockBugDeduction",
+        "deduct": -3002272560617,
+        "reference": []
+      },
+      {
+        "address": "DAG08c2j1UNGCeit2V2LPsJCzN2NX1RQbkWBaURZ",
+        "reason": "TokenUnlockBugDeduction",
+        "deduct": -3002272560617,
+        "reference": []
+      },
+      {
+        "address": "DAG09pzAYGA9yunSdpoxSYXcGksEApBPpMkVJCYW",
+        "reason": "TokenUnlockBugDeduction",
+        "deduct": -3002272560617,
+        "reference": []
+      },
+      {
+        "address": "DAG0A3xe4YgMPX33dSAfjXNsZHb5qKrSVweCqSSw",
+        "reason": "TokenUnlockBugDeduction",
+        "deduct": -3002272560617,
+        "reference": []
+      },
+      {
+        "address": "DAG0AKGxQwSjFnEf2wVRkm9RDJ3ahwXYPHUusB49",
+        "reason": "TokenUnlockBugDeduction",
+        "deduct": -3002272560617,
+        "reference": []
+      },
+      {
+        "address": "DAG0BPUshM1VAWAhLX3kMS1tZgMxuqU7EZoJ6mWa",
+        "reason": "TokenUnlockBugDeduction",
+        "deduct": -3002272560617,
+        "reference": []
+      },
+      {
+        "address": "DAG0BYD5FLBhvLg2tkdjcUv4USf77yzmv2HoHTHL",
+        "reason": "TokenUnlockBugDeduction",
+        "deduct": -3002272560617,
+        "reference": []
+      },
+      {
+        "address": "DAG0DtJkgSMq1pv63HP3z6W8myNLMomDgVvkJCri",
+        "reason": "TokenUnlockBugDeduction",
+        "deduct": -3002272560617,
+        "reference": []
+      },
+      {
+        "address": "DAG0E81B36BabwUCSySRLTXZDDLEDHfEEJEjWasK",
+        "reason": "TokenUnlockBugDeduction",
+        "deduct": -3002272560617,
+        "reference": []
+      },
+      {
+        "address": "DAG0EUpewFZ1qT6neKoqAFxNCBEUsE3h9cpNjr89",
+        "reason": "TokenUnlockBugDeduction",
+        "deduct": -3002272560617,
+        "reference": []
+      },
+      {
+        "address": "DAG0GsQ2BkHEBgkBB8hAsj6VpXcVPqSn2kBmXHSu",
+        "reason": "TokenUnlockBugDeduction",
+        "deduct": -3002272560617,
+        "reference": []
+      },
+      {
+        "address": "DAG0GydJBpZzLcC8rea3Tv7R8a28UbHwewDGsALH",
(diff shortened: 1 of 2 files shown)


modules/shared/src/main/scala/io/constellationnetwork/schema/artifact.scala
```diff
@@ -61,6 +61,7 @@ object artifact {
   case object SpendTransactionNotApplied extends BalanceAdjustmentReason
   case object SpendTransactionSourceNotApplied extends BalanceAdjustmentReason
   case object SpendTransactionDestinationNotApplied extends BalanceAdjustmentReason
+  case object TokenUnlockBugDeduction extends BalanceAdjustmentReason
```
ryle-ai, line 64: `amm-metagraph/modules/l0/src/main/scala/org/amm_metagraph/l0/BalanceAdjustmentLoader.scala` has a hardcoded `match` on the `reason` string that only covers the three existing cases. `"TokenUnlockBugDeduction"` will fall through to `Left(s"Unknown BalanceAdjustmentReason: $other")`. This means `loadBalanceAdjustments` will fail when parsing the `balance-adjustments-3.json` entries, and `customArtifacts` will return `None` even if that file is added. The match needs updating alongside this change.

modules/node-shared/src/main/resources/adjustments.json
```diff
@@ -1185,5 +1185,2537 @@
         "deduct": -22837225000000.0
       }
     ]
+  },
+  {
+    "currencyId": "DAG7X5idd4aLfp4XC6WQdG1eDfR3LGPVEwtUUB2W",
+    "snapshotOrdinal": 473500,
```
ryle-ai, line 1191: 999,999 is ~530k ordinals ahead of the current metagraph state. Without a corresponding `ordinalToPerformBalanceAdjustments3 = 999999L` branch in `amm-metagraph/Main.scala` `customArtifacts`, the metagraph will return no `BalanceAdjustment` artifacts at this ordinal and `validateRequiredAdjustments` will reject every snapshot, halting the metagraph. This ordinal also appears to be a placeholder rather than a deployment-specific target — either confirm it's intentional and document why, or set it to the actual planned deployment ordinal.

ryle-ai (changes requested):
## Missing amm-metagraph creation-side update

The adjustment framework operates on two sides that must match: the *validation side* (adjustments.json, checked on snapshot acceptance) and the *creation side* (`customArtifacts` in amm-metagraph Main.scala, which inserts the BalanceAdjustment artifacts into the snapshot).

This PR updates the validation side for ordinal 999,999 but does not update the creation side. `amm-metagraph/modules/l0/src/main/scala/org/amm_metagraph/l0/Main.scala` currently handles only two ordinals:

```scala
val ordinalToPerformBalanceAdjustments1 = 109991L
val ordinalToPerformBalanceAdjustments2 = 145000L
```

There is no case for 999,999. At that ordinal, `customArtifacts` will return `None`, so no `BalanceAdjustment` artifacts will be in the snapshot. The acceptance manager will then call `validateRequiredAdjustments` with 419 required entries and an empty artifact set — and reject the snapshot. The metagraph will halt indefinitely at ordinal 999,999, unable to form a valid snapshot. This is the same two-sided pattern that was used for ordinals 109,991 and 145,000.

The fix requires:
1. A `ordinalToPerformBalanceAdjustments3 = 999999L` branch in `customArtifacts` loading a `balance-adjustments-3.json`
2. A `balance-adjustments-3.json` resource in the amm-metagraph l0 with all 419 `BalanceAdjustment` entries
3. `"TokenUnlockBugDeduction"` added to the reason match in `amm-metagraph/modules/l0/src/main/scala/org/amm_metagraph/l0/BalanceAdjustmentLoader.scala` — currently the match has only three cases and returns `Left("Unknown BalanceAdjustmentReason: TokenUnlockBugDeduction")` for anything else, which would cause `loadBalanceAdjustments` to fail and `customArtifacts` to return `None`

## 4 addresses from the corrected CSV are missing from adjustments.json

The corrected deductions file (`token_unlock_deductions_corrected.csv`) has 423 entries. The PR contains 419. The four missing addresses:

| Address | Expected deduction (raw) | Expected deduction (SWAP) |
|---------|--------------------------|---------------------------|
| DAG15MGPwnPqemrgUkZCaYeukC63u1UYaq3WXze1 | 1,000,000,000,000 | 10,000.00 |
| DAG2UuPh5oYwfE5yMTivz1PJGhqfST2Y439DXRBA | 2,002,272,560,617 | 20,022.73 |
| DAG6wNGHp6hAjnAyhkdTJfW62YNMfaGb1sVpsKoR | 3,002,272,560,617 | 30,022.73 |
| DAG8GYFbgQUSYBxf5gWELhrzF6EJ6bZAzZETsXND | 3,002,272,560,617 | 30,022.73 |

DAG15MG and DAG2UuP have reduced amounts because each was the source of one of the two buggy unlocks and therefore only received excess credit from the other one. These are the only two addresses in the entire set with non-standard deduction amounts; their omission means those addresses retain their excess SWAP permanently.

DAG6wNG and DAG8GYF are standard cases (sources of non-buggy unlocks) that also received both buggy credits — they should have the same 30,022.73 SWAP deduction as the other 419.

## Missing PR description

The PR description is empty. This fix modifies the on-chain balance of 423 wallets. At a minimum the description should document: which snapshot ordinals triggered the bug (467,032 and 467,642), how deduction amounts were calculated and verified (Marcus's DB cross-check), the rationale for ordinal 999,999, and why 4 addresses in the CSV have non-standard amounts.

## `"reference": []` throughout

Every prior adjustment entry in adjustments.json includes transaction/event hashes under `reference` — they're the audit trail linking each deduction to the on-chain event that caused it. All 419 new entries have empty reference arrays. At minimum the two buggy snapshot ordinals (or their snapshot hashes) should be listed so that future readers can trace the basis for these deductions.

## snapshotOrdinal: 999999

999,999 is about 530,000 ordinals beyond the current metagraph state. Existing entries in adjustments.json use actual specific ordinals tied to deployment timing. 999,999 looks like a placeholder. If it is intentional (e.g., to allow time for the amm-metagraph creation-side change to be deployed), that should be documented. If it is a placeholder, the actual target ordinal needs to be determined.

modules/shared/src/main/scala/io/constellationnetwork/schema/artifact.scala
ryle-ai, line 64: `amm-metagraph/modules/l0/src/main/scala/org/amm_metagraph/l0/BalanceAdjustmentLoader.scala` has a hardcoded `match` on the `reason` string covering only three cases. `"TokenUnlockBugDeduction"` falls through to `Left(s"Unknown BalanceAdjustmentReason: $other")`, causing `loadBalanceAdjustments` to throw and `customArtifacts` to return `None`. This case must be added to the loader alongside this change.

modules/node-shared/src/main/resources/adjustments.json
```diff
@@ -1185,5 +1185,3818 @@
...
```
ryle-ai, line 1191: Tessellation will validate that 421 BalanceAdjustment artifacts are present in the SWAP snapshot at ordinal 473500. `amm-metagraph/modules/l0/src/main/scala/org/amm_metagraph/l0/Main.scala` `customArtifacts` currently only handles ordinals 109991 and 145000 — there is no `ordinalToPerformBalanceAdjustments3 = 473500L` case. Without the matching creation-side update (plus `balance-adjustments-3.json` resource and the loader reason fix), the metagraph will halt at this ordinal.

ryle-ai (changes requested):
## Resolved since last review

- DAG15MG and DAG2UuP are now included with the correct non-standard deductions (1,000,000,000,000 and 2,002,272,560,617 respectively)
- `snapshotOrdinal` changed from 999999 to 473500, which is a specific target ordinal rather than a placeholder

## Still blocking

### amm-metagraph creation side not updated

The validation side in tessellation (adjustments.json) now requires 421 BalanceAdjustment artifacts at SWAP ordinal 473500. The creation side in amm-metagraph has not been updated. Three things are needed there:

1. A third ordinal case in `customArtifacts` in `amm-metagraph/modules/l0/src/main/scala/org/amm_metagraph/l0/Main.scala` — currently only handles ordinals 109991 and 145000
2. A `balance-adjustments-3.json` resource file in `amm-metagraph/modules/l0/src/main/resources/`
3. `"TokenUnlockBugDeduction"` added to the reason `match` in `amm-metagraph/modules/l0/src/main/scala/org/amm_metagraph/l0/BalanceAdjustmentLoader.scala` — currently the match returns `Left(s"Unknown BalanceAdjustmentReason: $other")` for any string not in the three existing cases, which means `loadBalanceAdjustments` would throw and `customArtifacts` would return `None`

Without all three, at ordinal 473500 the SWAP metagraph emits no BalanceAdjustment artifacts, tessellation's `validateRequiredAdjustments` sees 421 required and 0 present, and rejects the snapshot. The metagraph halts indefinitely at that ordinal.

### 2 addresses still missing

The corrected CSV has 423 entries. This PR has 421. Two addresses from the corrected CSV are absent:

| Address | Expected deduction (raw) |
|---------|---------------------------|
| DAG6wNGHp6hAjnAyhkdTJfW62YNMfaGb1sVpsKoR | 3,002,272,560,617 |
| DAG8GYFbgQUSYBxf5gWELhrzF6EJ6bZAzZETsXND | 3,002,272,560,617 |

Both are in `token_unlock_deductions_corrected.csv` (lines 327 and 384) with standard 30,022.73 SWAP deductions. These addresses had non-buggy unlocks at ordinals 466,974 and 466,976 — before the two buggy unlocks fired at 467,032 and 467,642. Whether they still held active locks at the time of the buggy unlocks determines whether they received excess credits. The corrected CSV (authored by Marcus from the DB analysis) says they did and should be deducted. The PR disagrees by omitting them. This needs to be resolved explicitly — if they're excluded intentionally, document why.

## Non-blocking

**PR description is still empty.** This change modifies the on-chain balance of 421 wallets. The description should document: the two buggy snapshot ordinals (467,032 and 467,642), why 473500 was chosen as the target ordinal, and why DAG6wNG and DAG8GYF are excluded (if intentional).

**All 421 new entries have `"reference": []`.** Every prior entry in adjustments.json includes transaction or snapshot hashes under `reference` as an audit trail. At minimum the two buggy snapshot ordinals or their hashes should be referenced so future readers can trace the basis for these deductions.

scasplte2:
@marcus-girolneto the patch that was being applied to snapshot streaming can be removed now, that is probably what is causing that failure. Streaming has been updated.

marcus-girolneto:
## All review items resolved

### Creation side (amm-metagraph)
Already addressed in companion PR: https://github.com/Constellation-Labs/pacaswap-metagraph/pull/170

- `ordinalToPerformBalanceAdjustments3 = 473500L` added to `Main.scala` `customArtifacts`
- `balance-adjustments-3.json` resource created with all 423 entries
- `"TokenUnlockBugDeduction"` added to reason match in `BalanceAdjustmentLoader.scala`

### 2 missing addresses
`DAG6wNGHp6hAjnAyhkdTJfW62YNMfaGb1sVpsKoR` and `DAG8GYFbgQUSYBxf5gWELhrzF6EJ6bZAzZETsXND` were added in commit b067188. Total is now **423 entries** (not 421).

### References
All 423 entries now include the two lock reference hashes as audit trail (commit b067188):
- `bfd733e92aa71b2186955a4d7186301b27602621f0ee890159a8678032769f11` (ordinal 467032)
- `dbe69291991d4495ead034a1815c8d9d3b6c9ea20e5c6873136052e89601da11` (ordinal 467642)

### PR description
Updated with full context, deduction breakdown, and deployment order.

ryle-goehausen (dismissed):

ryle-ai (commented):
## Round 3 — resolved from round 2

All prior blockers are resolved:
- DAG6wNGHp6hAjnAyhkdTJfW62YNMfaGb1sVpsKoR and DAG8GYFbgQUSYBxf5gWELhrzF6EJ6bZAzZETsXND are present (diff lines 2940 and 3453)
- 423 entries total, all with correct amounts (DAG15MG at −1,000,000,000,000; DAG2UuP at −2,002,272,560,617; the remaining 421 at −3,002,272,560,617)
- References are populated with the two snapshot hashes
- amm-metagraph PR #170 (hotfix/token-unlock-balance-deductions-creation → release/mainnet) provides all three missing creation-side pieces: `ordinalToPerformBalanceAdjustments3 = 473500L`, `balance-adjustments-3.json` with 423 matching entries, and `"TokenUnlockBugDeduction"` added to the loader reason match

## amm-metagraph compilation dependency

amm-metagraph PR #170 adds `case "TokenUnlockBugDeduction" => Right(TokenUnlockBugDeduction)` in BalanceAdjustmentLoader, where `TokenUnlockBugDeduction` is a Scala identifier — it must resolve to the case object being added in this PR's `artifact.scala`. The amm-metagraph currently pins tessellation at **3.5.4**, which does not have this case object. PR #170's diff contains no tessellation version change.

This will fail to compile against 3.5.4 unless the build uses SNAPSHOT artifacts. Verify the amm-metagraph hotfix build uses a SNAPSHOT or dev build that includes this PR's changes — and that deploying to mainnet will use a released tessellation version that also includes them. If a new tessellation release is required first, that step needs to be explicit in the deployment plan.

## Deployment ordering and window

This tessellation PR targets `develop`; amm-metagraph PR #170 targets `release/mainnet` directly. The metagraph is at approximately ordinal 470,000 and the enforcement ordinal is 473,500 — roughly 3,500 ordinals of runway.

The hard constraint is: both changes must be live on mainnet before the metagraph produces snapshot 473,500. If tessellation #1487 is deployed without amm-metagraph #170, the validator will require 423 artifacts at that ordinal and receive zero, halting the metagraph. The reverse (amm-metagraph #170 without tessellation #1487) would attempt to emit a `TokenUnlockBugDeduction` artifact whose type tessellation cannot deserialize.

Make sure the deployment sequence and timing are locked before either is merged.

## Non-blocking

PR description is still empty. A change that adjusts the on-chain balance of 423 wallets should have at minimum the two buggy ordinals (467,032 and 467,642), the ordinal selection rationale (473,500), and the verification method documented.

PR #1480, opened 2026-03-30
chore: nightly ci builds

effects-ai:

PR #1489, opened 2026-04-09
chore: add grafana and clickhouse to the nightly cluster

effects-ai:
Adds a monitoring stack (Prometheus + Grafana + ClickHouse) to the nightly cluster and makes the ClickHouse sink protocol configurable.


docker/monitoring/clickhouse/certs/server.key
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
ryle-ai: Private RSA key committed to a public repo. Even though this is a self-signed key with no real identity and the ClickHouse config sets `verificationMode>none` (making the TLS essentially decorative), private keys do not belong in source control. Generate it at container startup or drop the local HTTPS ClickHouse config in favour of plain HTTP on 8123.

docker/monitoring/deploy-monitoring.sh
```diff
@@ -0,0 +1,115 @@
...
+# --- .env ---
+ssh "$SERVER_NODE" "echo 'GRAFANA_ADMIN_PASSWORD=${GRAFANA_ADMIN_PASSWORD:-admin}' > $REMOTE_DIR/.env"
+
+# --- Start services ---
+log "Pulling images and starting services"
+ssh "$SERVER_NODE" "cd $REMOTE_DIR && docker compose pull -q && docker compose up -d"
+
+# --- Wait for ClickHouse to be ready ---
+log "Waiting for ClickHouse"
+for i in $(seq 1 30); do
```
ryle-ai: If ClickHouse doesn't come up within 60s the loop exhausts silently and the script continues. The `|| true` on line 107 then swallows the init.sql failure. Add a boolean guard and `exit 1` on timeout — the same pattern used in `nightly-rebuild.yml` and fixed in `remote-deploy.sh` during PR #1480.

docker/monitoring/deploy-monitoring.sh
```diff
@@ -0,0 +1,115 @@
...
+ssh "$SERVER_NODE" "docker exec -i clickhouse clickhouse-client --password clickhouse" \
+  < "$SCRIPT_DIR/clickhouse/init.sql" 2>&1 || true
+
+log "Monitoring deployed on $SERVER_NODE"
+log "  Prometheus:  http://$SERVER_IP:9090"
+log "  Grafana:     http://$SERVER_IP:3000"
+log "  ClickHouse:  $SERVER_IP:8123 (HTTP)"
+log ""
+log "To enable ClickHouse logging on tessellation nodes, redeploy with:"
+log "  CLICKHOUSE_HOST=$SERVER_IP CLICKHOUSE_PORT=8123 CLICKHOUSE_PROTOCOL=http CLICKHOUSE_PASSWORD=clickhouse"
```
ryle-ai: This tip says to use `CLICKHOUSE_PROTOCOL=http` but `nightly-deploy.yml` has no `NIGHTLY_CLICKHOUSE_PROTOCOL` secret/variable and no way to pass this flag through the CI workflow. Without it, nodes default to `https`, which fails against the HTTP-only remote ClickHouse. Either add `NIGHTLY_CLICKHOUSE_PROTOCOL` to the workflow or document that node logging requires a separate TLS-enabled ClickHouse.

.github/workflows/nightly-deploy.yml
```diff
@@ -197,11 +197,30 @@ jobs:
...
+          CLICKHOUSE_PORT: ${{ secrets.NIGHTLY_CLICKHOUSE_PORT }}
+          CLICKHOUSE_USER: ${{ secrets.NIGHTLY_CLICKHOUSE_USER }}
+          CLICKHOUSE_PASSWORD: ${{ secrets.NIGHTLY_CLICKHOUSE_PASSWORD }}
+          CLICKHOUSE_DATABASE: ${{ secrets.NIGHTLY_CLICKHOUSE_DATABASE }}
         run: |
           IFS=',' read -ra IPS <<< "${{ vars.NIGHTLY_HOSTS }}"
           NODES=$(printf "n%s," "${!IPS[@]}" | sed 's/,$//')
           just up --skip-assembly --remote="$NODES" --snapshot-streaming-branch=${{ vars.SNAPSHOT_STREAMING_BRANCH || 'testing' }} --block-explorer-branch=${{ vars.BLOCK_EXPLORER_BRANCH || 'increase_delegated_stakes' }}
 
+      - name: Deploy monitoring
```
ryle-ai: The `CLICKHOUSE_*` env vars in this step are not consumed by `deploy-monitoring.sh` — the script hardcodes all ClickHouse credentials. These env vars are silently ignored here; they're only meaningful in the "Deploy cluster" step above. Either remove them from this step or have `deploy-monitoring.sh` actually use them.

docker/monitoring/grafana/provisioning/datasources/datasources.yaml
```diff
@@ -0,0 +1,25 @@
...
+    editable: true
+
+  - name: ClickHouse
+    uid: clickhouse
+    type: grafana-clickhouse-datasource
+    access: proxy
+    isDefault: false
+    editable: true
+    jsonData:
+      host: ${CLICKHOUSE_HOST}
```
ryle-ai: Grafana does not do environment variable substitution in provisioning YAML files. `${CLICKHOUSE_HOST}` and friends will be used literally, not expanded. The remote deploy writes its own version of this file via heredoc (with hardcoded values), so this file is dead config as committed. Either remove it or replace with a script-driven `envsubst` step.

ryle-ai (changes requested):
## Summary

Adds a monitoring stack (Prometheus + Grafana + ClickHouse) to the nightly cluster, and adds `protocol` configurability to the ClickHouse sink so nodes can connect via HTTP or HTTPS. The Scala changes are straightforward and correct. The infrastructure has two issues that need addressing before merge.

### PR description is empty

This PR deploys infrastructure to the nightly cluster and makes a breaking-default-sensitive change to node logging configuration. At minimum the description should explain the deployment sequence, the credential model, and how operators set `NIGHTLY_CLICKHOUSE_HOST` to wire node logging to the monitoring ClickHouse.

### Blocker: private TLS key committed to the public repo

`docker/monitoring/clickhouse/certs/server.key` is an RSA-2048 private key committed to a public GitHub repository. The cert is self-signed and the purpose is trivial (local dev TLS, with `verificationMode>none` so it's not even validated), but committing private keys to public repos is a policy violation regardless of the key's practical value. The correct pattern is to generate the key at container start (e.g. via a ClickHouse entrypoint script) or accept that local dev uses HTTP (port 8123) and drop the HTTPS config for this use case entirely.

### Significant: protocol mismatch between monitoring ClickHouse and nightly nodes

`docker-compose.remote.yaml` deploys ClickHouse in HTTP-only mode — `enable_https.xml` and the certs are NOT mounted, so the container only serves on port 8123. The tessellation nodes' `.env` defaults to `CLICKHOUSE_PROTOCOL=https` (via `docker-env-setup.sh` line 53 / `remote-deploy.sh` line 76) and port 8443. `nightly-deploy.yml` exposes `CLICKHOUSE_HOST/PORT/USER/PASSWORD/DATABASE` as secrets but has no `NIGHTLY_CLICKHOUSE_PROTOCOL` variable — there is no CI path to override the default.

The `deploy-monitoring.sh` tip at line 115 explicitly says to use `CLICKHOUSE_PROTOCOL=http`, but that env var is never injected by the workflow. When `NIGHTLY_CLICKHOUSE_HOST` is eventually set to point at the monitoring ClickHouse, node logs will silently fail to reach it (HTTPS to a port with no TLS listener). Either:
- Add `NIGHTLY_CLICKHOUSE_PROTOCOL` as a repo variable and wire it through both deploy steps, or
- Make the remote monitoring ClickHouse serve HTTPS (mounting certs, but then the private key issue above also applies remotely), or
- Accept that node logging uses a separate external ClickHouse with TLS and document that clearly.

### Non-blocking: hardcoded ClickHouse password in monitoring stack

The monitoring ClickHouse password is hardcoded as `clickhouse` in `docker-compose.remote.yaml`, the Grafana datasource written by `deploy-monitoring.sh`, and the health/init commands. The `CLICKHOUSE_PASSWORD` env vars passed to the "Deploy monitoring" CI step are not used by `deploy-monitoring.sh` — the script ignores them. This means `NIGHTLY_CLICKHOUSE_PASSWORD` must be set to `clickhouse` for node logging to match the monitoring stack, or the monitoring stack credential is effectively undocumented. The `CLICKHOUSE_*` env block on the "Deploy monitoring" CI step is misleading since deploy-monitoring.sh doesn't consume those vars.

### Non-blocking: readiness loop has no failure exit

`deploy-monitoring.sh` loops 30×2s waiting for ClickHouse, then silently continues if it never comes up. The `|| true` on the subsequent `init.sql` execution means the script exits 0 even if ClickHouse never started and tables were never created. Same pattern flagged in PR #1480 for `remote-deploy.sh`.

### Non-blocking: `datasources.yaml` has unsubstituted template variables

`docker/monitoring/grafana/provisioning/datasources/datasources.yaml` contains `${CLICKHOUSE_HOST}`, `${CLICKHOUSE_PORT}`, etc. Grafana does not perform environment variable substitution in provisioning YAML files. This file won't work if mounted directly — but the remote deploy writes its own hardcoded version, so this file is effectively unused dead config.

.github/workflows/nightly-deploy.yml
```diff
@@ -197,11 +197,25 @@ jobs:
...
```
ryle-ai: "Deploy monitoring" runs here, after nodes have already been started by "Deploy cluster" above. `deploy-monitoring.sh` patches node `.env` files with `CLICKHOUSE_PROTOCOL=http`, but the running nodes won't reload their environment. On top of that, the "Deploy cluster" step has no `NIGHTLY_CLICKHOUSE_PROTOCOL` secret, so `remote-deploy.sh` defaults to writing `CLICKHOUSE_PROTOCOL=https` before monitoring can patch it — meaning nodes start with the wrong protocol and never get corrected without a restart.

`nightly-rebuild.yml` handles this correctly by running "Ensure monitoring" before "Start cluster". This workflow needs the same ordering fix.

.github/workflows/nightly-deploy.yml
```diff
@@ -197,11 +197,25 @@ jobs:
       - name: Deploy cluster
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CLICKHOUSE_HOST: ${{ secrets.NIGHTLY_CLICKHOUSE_HOST }}
+          CLICKHOUSE_PORT: ${{ secrets.NIGHTLY_CLICKHOUSE_PORT }}
+          CLICKHOUSE_USER: ${{ secrets.NIGHTLY_CLICKHOUSE_USER }}
+          CLICKHOUSE_PASSWORD: ${{ secrets.NIGHTLY_CLICKHOUSE_PASSWORD }}
```
ryle-ai, line 212: `NIGHTLY_CLICKHOUSE_PASSWORD` is passed here and written to node `.env` by `remote-deploy.sh`. But `deploy-monitoring.sh` (which runs in the next step) ignores this secret entirely and hardcodes `clickhouse` as the ClickHouse server password, then overwrites the node `.env` with that hardcoded value. So `NIGHTLY_CLICKHOUSE_PASSWORD` must be set to exactly `clickhouse` or the credentials are mismatched. The secret is not actually used.

docker/monitoring/deploy-monitoring.sh
```diff
@@ -0,0 +1,137 @@
...
+
+# --- Update tessellation node configs with ClickHouse settings ---
+log "Updating tessellation node configs"
+for h in "${NODES[@]}"; do
+  if ssh "$h" "test -f $TESS_DIR/.env" 2>/dev/null; then
+    ssh "$h" "sed -i '/^CLICKHOUSE_/d' $TESS_DIR/.env"
+    printf '%s\n' \
+      "CLICKHOUSE_HOST=${SERVER_IP}" \
+      "CLICKHOUSE_USER=default" \
+      "CLICKHOUSE_PASSWORD=clickhouse" \
```
ryle-ai: Password hardcoded here, and also at lines 94 and 108 (health-check and init client). If this should ever change, it needs updating in four places (here, those two lines, and `datasources.yaml`). Consider reading from `CLICKHOUSE_PASSWORD` env var — same pattern as `GRAFANA_ADMIN_PASSWORD`.

docker/monitoring/docker-compose.remote.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+  grafana:
+    image: grafana/grafana:11.5.1
+    container_name: grafana
+    restart: unless-stopped
+    network_mode: host
+    environment:
+      - GF_SECURITY_ADMIN_USER=admin
+      - GF_SECURITY_ADMIN_PASSWORD=${GRAFANA_ADMIN_PASSWORD:-admin}
+      - GF_AUTH_ANONYMOUS_ENABLED=true
```
ryle-ai, line 44: Anonymous viewer access means Grafana dashboards (including Prometheus targets with real node IPs) are publicly readable without credentials. Worth a deliberate decision.

ryle-ai (changes requested):
The Scala changes (`protocol` field in `ClickHouseConfig`/`ClickHouseSink`/`types.scala`/`application.conf`) are correct and unproblematic. The infrastructure side has two issues.

### PR description is empty

This PR deploys a new monitoring stack to the nightly cluster and makes a persistent change to how tessellation nodes are configured (ClickHouse logging). The deployment sequence, credential model, and how `NIGHTLY_CLICKHOUSE_HOST` wires nodes to the monitoring ClickHouse should be documented.

### Ordering bug in `nightly-deploy.yml`: nodes start before monitoring patches their config

`nightly-deploy.yml` runs "Deploy cluster" (starts nodes via `just up`) before "Deploy monitoring". Inside `just up`, `remote-deploy.sh` writes `CLICKHOUSE_PROTOCOL=${CLICKHOUSE_PROTOCOL:-https}` to each node's `.env` — the default is `https`/`8443` because there's no `NIGHTLY_CLICKHOUSE_PROTOCOL` secret. Monitoring then starts ClickHouse (HTTP-only, port 8123) and patches the node `.env` files to `CLICKHOUSE_PROTOCOL=http`/`8123`, but by then the nodes are already running and don't re-read `.env`. Net result: every `nightly-deploy.yml` run leaves nodes trying to connect via HTTPS to an HTTP-only endpoint; ClickHouse logging silently fails until the next `nightly-rebuild.yml` run (which correctly runs "Ensure monitoring" before "Start cluster").

The straightforward fix: swap the order so "Deploy monitoring" runs before "Deploy cluster", and drop `CLICKHOUSE_*` from the "Deploy cluster" `env` block — let `deploy-monitoring.sh` be the sole writer of ClickHouse config to node `.env` files.

### Non-blocking: `NIGHTLY_CLICKHOUSE_PASSWORD` must equal `clickhouse`

`deploy-monitoring.sh` hardcodes the ClickHouse password as `clickhouse` in the ClickHouse container, the init client command, the node `.env` patch, and the Grafana datasource provisioning file. Whatever value `NIGHTLY_CLICKHOUSE_PASSWORD` is set to must match `clickhouse` for node logging to work, making the secret functionally meaningless. Either the monitoring script should read `CLICKHOUSE_PASSWORD` from the environment (parallel to how `GRAFANA_ADMIN_PASSWORD` is handled), or this should be documented explicitly.

### Non-blocking: Grafana anonymous viewer access

Both `docker-compose.remote.yaml` and `docker-compose.local.yaml` enable `GF_AUTH_ANONYMOUS_ENABLED=true` with role `Viewer`. Dashboards include Prometheus targets (real node IPs) and JVM/consensus metrics. For an internal nightly cluster this is probably acceptable, but worth a deliberate decision rather than a default.

.github/workflows/nightly-deploy.yml
```diff
@@ -0,0 +1,266 @@
...
+          IFS=',' read -ra IPS <<< "${{ vars.NIGHTLY_HOSTS }}"
+          NODES=$(printf "n%s," "${!IPS[@]}" | sed 's/,$//')
+          CLEAN="${{ inputs.clean }}"
+          CLEAN="${CLEAN:-false}"
+          if [ "$CLEAN" = "true" ]; then
+            just down --remote="$NODES" --clean
+          else
+            just down --remote="$NODES"
+          fi
+
```
ryle-ai, line 208: "Deploy cluster" starts nodes here; "Deploy monitoring" runs after (line 209). `remote-deploy.sh` defaults to `CLICKHOUSE_PROTOCOL=https`/`8443` when `NIGHTLY_CLICKHOUSE_HOST` is set — nodes start with that config. Monitoring then patches node `.env` to http/8123, but running nodes don't re-read `.env`.

Fix: move "Deploy monitoring" to run before "Deploy cluster", and drop the `CLICKHOUSE_*` env block from this step so `remote-deploy.sh` skips writing ClickHouse config and defers to whatever `deploy-monitoring.sh` already set.

.github/workflows/nightly-deploy.yml
```diff
@@ -0,0 +1,266 @@
...
+      - name: Deploy cluster
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CLICKHOUSE_HOST: ${{ secrets.NIGHTLY_CLICKHOUSE_HOST }}
+          CLICKHOUSE_PORT: ${{ secrets.NIGHTLY_CLICKHOUSE_PORT }}
+          CLICKHOUSE_USER: ${{ secrets.NIGHTLY_CLICKHOUSE_USER }}
```
ryle-ai: `deploy-monitoring.sh` always writes `CLICKHOUSE_PASSWORD=clickhouse` to node `.env` files (overwriting this value), and `docker-compose.remote.yaml` hardcodes the server password as `clickhouse`. Whatever value `NIGHTLY_CLICKHOUSE_PASSWORD` is set to has no effect — the effective ClickHouse password is always `clickhouse`. Either remove this secret from the step env or have `deploy-monitoring.sh` read from `CLICKHOUSE_PASSWORD` the same way it reads `GRAFANA_ADMIN_PASSWORD`.

docker/monitoring/grafana/provisioning/datasources/datasources.yaml
```diff
@@ -0,0 +1,25 @@
...
+    editable: true
+    jsonData:
+      host: localhost
+      port: 8123
+      protocol: http
+      secure: false
+      username: default
+      defaultDatabase: default
+    secureJsonData:
+      password: clickhouse
```
ryle-ai: `secureJsonData.password` is committed in plaintext to the repo. Grafana encrypts it on first provisioning load, but the plaintext source is now in version history. Previously this was written via heredoc at deploy time (keeping the credential out of source control). Minor given the password is otherwise hardcoded across the stack, but worth noting.

ryle-ai (changes requested):
## PR description is still empty

Three review rounds have requested a description. This PR deploys new infrastructure to the nightly cluster and changes how node ClickHouse logging is configured. It needs a description.

## Ordering bug in `nightly-deploy.yml`: nodes start before monitoring can configure them

This is the same issue flagged in round 2 and it is not fixed.

`nightly-deploy.yml` runs "Deploy cluster" before "Deploy monitoring". When `NIGHTLY_CLICKHOUSE_HOST` is set, `remote-deploy.sh` writes `CLICKHOUSE_PROTOCOL=${CLICKHOUSE_PROTOCOL:-https}` and `CLICKHOUSE_PORT=8443` to each node's `.env`, then `just up` starts the nodes. "Deploy monitoring" then starts ClickHouse (HTTP-only, port 8123) and patches node `.env` files to `CLICKHOUSE_PROTOCOL=http`/`8123` — but the nodes are already running and won't re-read `.env`. Every `nightly-deploy.yml` run will leave nodes connecting via HTTPS to an HTTP-only endpoint; ClickHouse logging silently fails.

`nightly-rebuild.yml` runs "Ensure monitoring" before "Start cluster", which is the right intent. But "Start cluster" calls `remote-deploy.sh` without `CLICKHOUSE_HOST` in its environment, so remote-deploy.sh creates a fresh `.env` without ClickHouse config and SCPs it to each node — overwriting the http config that monitoring just patched. Net result: nodes start without any ClickHouse config at all, so logging also doesn't work there.

The straightforward fix for `nightly-deploy.yml`: swap "Deploy monitoring" before "Deploy cluster" and drop the `CLICKHOUSE_*` env block from "Deploy cluster" entirely. Let `deploy-monitoring.sh` be the sole writer of ClickHouse config to node `.env` files, then start nodes after.

For `nightly-rebuild.yml` to also work: either pass `CLICKHOUSE_PROTOCOL=http` explicitly in the "Start cluster" step, or have `remote-deploy.sh` not overwrite ClickHouse variables that are already present in the node's `.env`.

## `NIGHTLY_CLICKHOUSE_PASSWORD` secret is a no-op

`deploy-monitoring.sh` unconditionally writes `CLICKHOUSE_PASSWORD=clickhouse` to each node's `.env`, overwriting whatever `remote-deploy.sh` wrote from `NIGHTLY_CLICKHOUSE_PASSWORD`. The ClickHouse container also hardcodes `CLICKHOUSE_PASSWORD=clickhouse`. The secret has no effect on the actual ClickHouse password. It also means the nightly cluster's ClickHouse has a static, well-known password (`clickhouse`) that cannot be rotated via CI secrets.

ryle-goehausen (approved):

PR #1490, opened 2026-04-09
fix: ensure tx-sender runs on remote cluster for nightly e2e tests

effects-ai:


.github/workflows/nightly-e2e.yml
```diff
@@ -102,6 +102,25 @@ jobs:
...
           echo "nodes=${NODES}" >> "$GITHUB_OUTPUT"
 
+      - name: Ensure tx-sender is running on genesis node
+        run: |
+          if ssh n0 "docker ps --filter name=tx-sender --format '{{.Names}}'" 2>/dev/null | grep -q tx-sender; then
+            echo "tx-sender already running"
+          else
+            echo "Starting tx-sender on n0"
+            ssh n0 "mkdir -p /opt/tessellation/tx-sender"
+            scp -q docker/jars/tools.jar n0:/opt/tessellation/tx-sender/tools.jar 2>/dev/null || true
```
ryle-ai: `docker/jars/` is gitignored — this file never exists on the nightly-e2e runner. The `2>/dev/null || true` hides the failure. When `docker run` then tries to bind-mount `/opt/tessellation/tx-sender/tools.jar` (which was never copied), the container fails to start and the `|| echo "tx-sender failed (non-fatal)"` branch fires. tx-sender never runs.

.github/workflows/nightly-e2e.yml
```diff
@@ -5,7 +5,7 @@ on:
     - cron: '0 * * * *'
   push:
     branches:
-      - nightly_ci_builds
+      - fix_nightly_e2e
```
ryle-ai: This push trigger was needed to test the workflow change on the PR branch. After merge to develop it's dead — that branch will be deleted. Either change it to `develop` or remove the `push:` block.

ryle-ai (changes requested):
The tx-sender step will silently fail to start on every run.

`docker/jars/` is listed in `.gitignore` and is populated only by `docker/bin/assembly.sh`. The nightly-e2e workflow has no build step — it checks out code, installs Node, and runs tests via SSH. `docker/jars/tools.jar` does not exist on the CI runner when this step executes.

The scp is suppressed with `2>/dev/null || true`, so the missing file is invisible in logs. Without the JAR on n0, `docker run -d -v /opt/tessellation/tx-sender/tools.jar:/app/tools.jar:ro ...` fails (Docker bind-mounting a non-existent host path errors at container creation). The `|| echo "tx-sender failed (non-fatal)"` then fires, the step exits 0, and the tests proceed with tx-sender never running. The PR's stated goal is not achieved.

The fix depends on where tools.jar is available after nightly-rebuild runs. If nightly-rebuild assembles the tools module and deploys the JAR to the remote nodes, the `docker run` command can reference that path directly (no scp needed). If tools.jar isn't currently produced by the build pipeline at all, nightly-rebuild needs a step to assemble it.

Also: the `push: branches: - fix_nightly_e2e` trigger becomes dead after merge. The original had `nightly_ci_builds` — same situation. This branch slot should either point to `develop` (to match nightly-deploy.yml's push trigger) or be removed.

.github/workflows/nightly-e2e.yml
ryle-ai: This was `nightly_ci_builds` in the baseline, which ryle-ai flagged as the same problem. After merge, `fix_nightly_e2e` is deleted and this trigger is dead. `nightly-deploy.yml` uses `push: branches: [develop, nightly_ci_builds]` — matching one of those here would keep the trigger meaningful, or remove it.

ryle-ai (commented):
The core fix is sound. `remote-deploy.sh` Phase 5 scps `tools.jar` to `$DIR/tx-sender/tools.jar` on the genesis node during every nightly-deploy run. `nightly-rebuild.yml` removes the tx-sender *container* (`docker rm -f`) but does not delete the JAR file, so tools.jar persists on disk. The existence check before starting is the right approach and correctly handles the post-rebuild state.

One issue from ryle-ai's review remains unaddressed: the push trigger branch.

ryle-ai (commented):
The tx-sender startup logic is correct. `remote-deploy.sh` Phase 5 scps `tools.jar` to `$DIR/tx-sender/tools.jar` during every `nightly-deploy` run. `nightly-rebuild` stops the container but leaves the JAR on disk. The existence check before starting is the right approach.

One issue flagged in both previous reviews remains unaddressed: the `push:` branch in both files is `fix_nightly_e2e`, which will be deleted on merge. In `nightly-deploy.yml` this is cosmetic since the `develop` trigger is still there. In `nightly-e2e.yml` it's the only push branch, so after merge the push trigger fires on nothing — only the hourly cron and `workflow_dispatch` remain. If that's intentional (E2E is expensive, cron-only is fine), document it; otherwise change it to `develop` to match `nightly-deploy.yml`'s push trigger.

.github/workflows/nightly-deploy.yml
```diff
@@ -6,7 +6,7 @@ on:
   push:
     branches:
       - develop
-      - nightly_ci_builds
+      - fix_nightly_e2e
   workflow_dispatch:
```
ryle-ai, line 9: `fix_nightly_e2e` is deleted on merge to develop; this slot becomes dead. Either remove it or revert to `nightly_ci_builds` (if that branch is still used for testing) — `develop` is already the active push trigger above.

.github/workflows/nightly-e2e.yml
ryle-ai: Same issue as in `nightly-deploy.yml`, but here it's the only push branch. After merge, no push to any branch triggers E2E — only the hourly cron and `workflow_dispatch`. If E2E should also fire on `develop` pushes, change this to `develop`; if cron-only is the intent, remove the `push:` block entirely rather than leaving a dead branch reference.

.github/workflows/nightly-e2e.yml
```diff
@@ -3,9 +3,6 @@ name: Nightly E2E
 on:
   schedule:
     - cron: '0 * * * *'
-  push:
-    branches:
-      - nightly_ci_builds
   workflow_dispatch:
```
ryle-ai, line 6: The `push:` block is now gone entirely, not just pointing at a deleted branch. E2E will only trigger on the hourly cron and manual dispatch — no push-triggered runs after merge. If that's intentional (reasonable given the 90-min timeout and the 80-min deploy-wait step), a comment here saying so would stop future reviewers from asking the same question.

ryle-ai (commented):
The tx-sender startup logic is correct. `remote-deploy.sh` Phase 5 scps `tools.jar` from `docker/jars/` (built by `tools/assembly` and uploaded as an artifact) to `$DIR/tx-sender/tools.jar` on n0 during every nightly-deploy run. `nightly-rebuild.yml` force-removes the container but leaves the JAR on disk. The existence check before starting correctly handles the post-rebuild state and the `::warning::` fallback correctly handles the first-run case where nightly-deploy hasn't run yet. The `sed` URL substitution matches what Phase 5 of `remote-deploy.sh` does, and port 9010 is correct for GL1 public access on the nightly cluster.

One observation: the `push:` block in `nightly-e2e.yml` is now completely absent — not just pointing at a dead branch. After merge, E2E triggers only on the hourly cron and `workflow_dispatch`. This was flagged across three previous review rounds. Removing the dead `fix_nightly_e2e` / `nightly_ci_builds` reference is the right call, and cron-only is probably fine (E2E is expensive at 90-min timeout and has to wait up to 80 min for deploy anyway), but the empty PR description leaves the intent undocumented. If cron-only is deliberate, a comment in the workflow to that effect would close the question.

ryle-goehausen (approved):

PR #1494, opened 2026-04-15
chore: add nightly e2e slack alerts

effects-ai:
Add Slack notifications for Nightly Deploy and Nightly E2E

### DIFF
--- a/.github/workflows/nightly-deploy.yml
+++ b/.github/workflows/nightly-deploy.yml
@@ -1,0 +1,271 @@
+name: Nightly Deploy
+
+on:
+  # Deploy on every merge to develop so the nightly cluster always runs the latest code.
+  # cancel-in-progress ensures rapid pushes don't pile up — only the latest deploy runs.
+  push:
+    branches:
+      - develop
+  workflow_dispatch:
+    inputs:
+      skip_build:
+        type: boolean
+        description: 'Skip build (reuse existing JARs on nodes)'
+        required: false
+        default: false
+      clean:
+        type: boolean
+        description: 'Wipe data and do fresh genesis (vs rollback restart)'
+        required: false
+        default: false
+  workflow_call:
+    inputs:
+      skip_build:
+        type: boolean
+        required: false
+        default: false
+      clean:
+        type: boolean
+        required: false
+        default: false
+
+# Shared with nightly-rebuild.yml — only one deploy/rebuild can run at a time
+concurrency:
+  group: nightly-deploy
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: Build JARs
+    if: ${{ !inputs.skip_build }}
+    runs-on: ubuntu-22.04
+    timeout-minutes: 45
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.ref }}
+          fetch-tags: true
+
+      - name: Setup Java 21
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: '21'
+
+      - name: Cache SBT dependencies
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.sbt
+            ~/.ivy2/cache
+            ~/.coursier
+            ~/.cache/coursier
+          key: ${{ runner.os }}-sbt-${{ hashFiles('build.sbt', 'project/*.scala', 'project/*.sbt', 'project/build.properties') }}
+          restore-keys: |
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly-e2e.yml
```diff
@@ -144,3 +144,23 @@ jobs:
           path: runner-logs/
           if-no-files-found: ignore
           retention-days: 7
+
+      - name: Notify Slack on failure
+        if: failure()
```
ryle-ai, line 149: `if: failure()` does not fire for cancelled runs. Given `cancel-in-progress: true` and the hourly cron, a run preempted by the next scheduled invocation won't produce an alert. Consider `if: failure() || cancelled()` if you want to know when tests don't complete for any reason.

ryle-ai (commented):
Empty PR description.

One behavioral gap worth knowing about: this workflow has `cancel-in-progress: true` and runs on a 1-hour cron with a 90-minute timeout. If a run is cancelled by the next hourly invocation mid-test (which can happen if the E2E suite runs long), `if: failure()` will not fire — cancellations are silently dropped. Whether that's acceptable depends on how you interpret a cancelled run (intentional preemption vs. unfinished test suite). If you want to catch both cases, `if: failure() || cancelled()` covers it.

.github/workflows/nightly-e2e.yml
effects-ai: we only need to know about failures

ryle-ai (commented):
Two things worth addressing before this merges:

**Build failure produces no Slack alert.** The Slack notification lives inside the `deploy` job. If the `build` job fails, the `deploy` job is skipped entirely (its `if: !cancelled() && (needs.build.result == 'success' || needs.build.result == 'skipped')` condition is false), and a skipped job runs zero steps — including the `if: always()` notification step. A build failure on a push to develop would produce no alert at all. Given that the nightly E2E polls for in-progress deploy/rebuild runs and then proceeds against whatever cluster is already up once the deploy finishes (or fails), a silent build failure might even result in E2E tests passing against a stale cluster, compounding the problem.

The fix is a separate `notify` job that runs unconditionally on failure: `needs: [build, deploy]` with `if: failure()`.

**`increase_delegated_stakes` as a hardcoded fallback branch.** Both the build and deploy steps fall back to `increase_delegated_stakes` for `--block-explorer-branch`. That's a feature branch name, not a stable ref. When that branch is eventually deleted or diverges significantly, the fallback silently breaks. The snapshot-streaming fallback uses `testing`, which looks like it's intended as a stable integration branch — the block-explorer fallback should use something similarly stable.

.github/workflows/nightly-deploy.yml
```diff
@@ -0,0 +1,271 @@
...
+  deploy:
+    name: Deploy to nightly cluster
+    needs: build
+    if: ${{ !cancelled() && (needs.build.result == 'success' || needs.build.result == 'skipped') }}
+    runs-on: ubuntu-22.04
+    timeout-minutes: 30
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
```
ryle-ai, line 118: If the `build` job fails, this condition is false and the entire `deploy` job is skipped — meaning the `if: always()` Slack notification step below never runs. Build failures produce no alert. Consider a third `notify` job with `needs: [build, deploy]` and `if: failure()` so the notification fires regardless of which job failed.

.github/workflows/nightly-deploy.yml
```diff
@@ -0,0 +1,271 @@
...
+          sudo apt-get install -y sbt
+
+      - name: Build all JARs
+        run: just build --num-gl0=3 --snapshot-streaming-branch=${{ vars.SNAPSHOT_STREAMING_BRANCH || 'testing' }} --block-explorer-branch=${{ vars.BLOCK_EXPLORER_BRANCH || 'increase_delegated_stakes' }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SNAPSHOT_STREAMING_REPO: https://github.com/Constellation-Labs/snapshot-streaming.git
+
+      - name: Upload tessellation JARs
+        uses: actions/upload-artifact@v4
```
ryle-ai, line 89: `increase_delegated_stakes` is a feature branch name, not a stable ref. When that branch is deleted, this fallback breaks silently. `--snapshot-streaming-branch` falls back to `testing` (a stable integration branch) — `--block-explorer-branch` should use something equivalently stable.

effects-ai:
Closed as superseded by https://github.com/Constellation-Labs/tessellation/pull/1489

PR #1495, opened 2026-04-15
fix: clean up token locks after the edge case tests

effects-ai:
- Fix nightly E2E token-lock-replacement test failing due to state accumulation across hourly runs
- Add cleanup that withdraws all delegated stakes (unlocking token locks) before and after the test, wrapped in try/finally so cleanup runs even on mid-test       
  failures

### DIFF
--- a/.github/action_scripts/delegated_staking/token-lock-replacement-edge-cases.js
+++ b/.github/action_scripts/delegated_staking/token-lock-replacement-edge-cases.js
@@ -522,6 +522,59 @@ const setupNodeParameters = async (urls) => {
 }
 
 /**
+ * Clean up all existing delegated stakes for an account.
+ * Withdraws all active stakes and waits for pending withdrawals to complete,
+ * which automatically unlocks the associated token locks and returns balance.
+ *
+ * This prevents state accumulation across repeated runs on persistent clusters
+ * (e.g. the nightly E2E environment where the same account is reused hourly).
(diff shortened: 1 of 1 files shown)


ryle-ai (commented):
The cleanup design is sound. Two observations:

**Dead code in `testReplaceSameAmount`:** The pre-test `cleanupExistingStakes` guarantees `activeDelegatedStakes` is empty before `testReplaceSameAmount` runs, so the "Reuse existing stake" branch (lines ~138-147 in the full file) can never be taken anymore. Before this PR, that branch made sense as a re-run resumption path. Now it just confuses future readers who might wonder whether the pre-test cleanup can fail silently. It should be removed.

**`includes('already')` catch in `cleanupExistingStakes`:** Matching against the string `'already'` to detect "stake already withdrawing" is fragile if the SDK or API error format ever changes. The consequence of a miss is that the cleanup throws rather than continuing, which is actually safe (fail loud). But if a different error happens to contain the word "already", the error would be swallowed and the stake skipped silently. Given the consistency with the rest of the file's error-matching style (`includes('StakeExistsForNode')`, etc.) this is acceptable, but something like matching `'AlreadyWithdrawing'` or the exact protocol error name would be more robust.

effects-ai:
Closed as superseded by https://github.com/Constellation-Labs/tessellation/pull/1489

PR #1497, opened 2026-04-23
chore: enable network metrics for the nightly cluster

effects-ai:


docker/monitoring/network-process-exporter/network_process_exporter.py
```diff
@@ -0,0 +1,242 @@
...
+// --- TCP send ---
+int kprobe__tcp_sendmsg(struct pt_regs *ctx, struct sock *sk) {
+    u64 pid_tgid = bpf_get_current_pid_tgid();
+    currsock.update(&pid_tgid, &sk);
+    return 0;
+}
+
+int kretprobe__tcp_sendmsg(struct pt_regs *ctx) {
+    int size = PT_REGS_RC(ctx);
+    if (size <= 0) return 0;
```
ryle-ai, line 37: Early return on `size <= 0` (error/zero-send path) leaves the `currsock[pid_tgid]` entry populated. The entry gets overwritten when the same thread makes the next `tcp_sendmsg` call, so there's no data corruption in practice. But if a thread's connections are repeatedly failing (e.g., during reconnect storms), `currsock` accumulates stale entries. The default BCC `BPF_HASH` max_entries is 10240 — unlikely to overflow in normal use, but worth knowing. Same pattern on lines 71 (tcp_recvmsg), 105 (udp_sendmsg), 139 (udp_recvmsg).

Fix: `currsock.delete(&pid_tgid)` before `return 0` in the `size <= 0` branch (mirroring the family-check cleanup at line 46).

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/http/p2p/middlewares/MetricsMiddleware.scala
```diff
@@ -17,7 +17,8 @@ import org.http4s.headers.`X-Forwarded-For`
 object MetricsMiddleware {
 
   def isHistogramRoute(path: String): Boolean =
-    path.contains("rumor")
+    // allow all routes to send metrics
+    true
```
ryle-ai, line 21: This change lands in the node binary, not just the monitoring stack. If this branch merges to develop, all future releases will histogram every P2P route. That's probably the intent (the new dashboard needs it), but it's worth an explicit sign-off given this file's scope.

ryle-ai (commented):
Three observations, all non-blocking for a nightly cluster.

**`isHistogramRoute` change touches production code.** The diff is in `node-shared`, so it affects all layers (dag-l0, dag-l1, currency-l0) if this branch merges to develop and eventually mainnet. Enabling histograms for all routes instead of just "rumor" routes is intentional for this dashboard, and the per-request overhead is negligible (Micrometer `DistributionSummary.record()` is O(1)). Just worth being aware that this isn't scoped to the monitoring docker directory — it ships with the node binary.

**Response size histograms will be empty for many interesting routes.** `MetricsMiddleware` records response size only when `response.contentLength` is `Some` — it uses `traverse_` over `Option[Long]`. Chunked/streaming responses (snapshot downloads, gossip state, etc.) never set Content-Length, so those won't appear in the "Response Size" panels. The dashboard as designed will silently show no data for the largest traffic. Pre-existing limitation, but the new dashboard surfaces it.

**`service_cache` in the BPF exporter is never pruned.** On a long-running exporter, dead PIDs accumulate in the dict. This is benign because `collect_and_export` only calls `resolve_service` for PIDs that appear in the BPF maps — dead processes don't produce traffic. The only real risk is PID reuse: if tessellation restarts and the new JVM lands on a PID that was previously 'unknown' (or a non-tessellation process), the cache returns the stale classification until the exporter restarts. Minor for nightly infra.

ryle-goehausen (approved):

PR #1498, opened 2026-04-27
fix: enforce removal of processed delegated stake withdrawals

marcus-girolneto:
## Summary
- Unblocks mainnet consensus, currently halted at ordinal **6176655** by a `MissingTokenLock` error in `GlobalSnapshotAcceptanceManager.generateTokenUnlocks`
- An orphan `PendingDelegatedStakeWithdrawal` for `DAG15hycLmg7AJbksHUcS6Zbs3oiWjFtyk6Zc28k` (tokenLockRef `bdbdb766…`) survived ordinal 6176654 in `GlobalSnapshotInfo.delegatedStakesWithdrawals` even though its token lock was unlocked in that ordinal. At 6176655 it is partitioned as expired again, but the lock is gone, so unlock generation fails and consensus stalls.
- Enforces the invariant **\"once we generate an unlock for a tokenLockRef, no `PendingDelegatedStakeWithdrawal` referencing it remains\"**

## Changes
Gated by new field-added ordinal `removingProcessedDelegatedStakeWithdrawals` (mainnet=`6176655`, testnet/integrationnet=`9999999`, dev=`0`):

1. **Skip orphans gracefully in `generateTokenUnlocks`** — when a withdrawal's `tokenLockRef` is no longer in `globalActiveTokenLocksByRef`, log a warning and skip it instead of failing the whole snapshot.
2. **Defensively clean `updatedWithdrawDelegatedStakes`** — when building `updatedWithdrawDelegatedStakesCleaned`, subtract every entry whose `tokenLockRef` appeared in `expiredWithdrawalsDelegatedStaking`. Guarantees the invariant regardless of which upstream path computed the value.

Pre-activation behavior is preserved exactly so historical re-validation matches.

## Files
- `modules/node-shared/src/main/resources/application.conf` — new ordinal config
- `modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/config/types.scala` — `FieldsAddedOrdinals` field
- `modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/GlobalSnapshotAcceptanceManager.scala` — fix logic + diagnostic logs retained for follow-up investigation
- 5 test files — new `Map.empty` for the added field

## Open follow-up (not blocking this PR)
Static analysis still cannot pinpoint *how* the orphan entry survived the prior snapshot — `getUpdatedWithdrawalDelegatedStakes` should have excluded it. Investigation will continue (likely candidates: in-memory `lastOutcome.context` divergence from disk, or a state-rebuild path that bypasses the filter).

## Test plan
- [ ] CI compile + tests
- [ ] Local sbt `nodeShared/test`, `dagL0/test` pass
- [ ] Verify on a staging environment (or one mainnet node first) that consensus resumes past 6176655
- [ ] Confirm warning log appears for `bdbdb766` and the entry is dropped from state at activation ordinal

### DIFF
--- a/modules/currency-l1/src/test/scala/io/constellationnetwork/currency/l1/domain/snapshots/programs/CurrencySnapshotProcessorSuite.scala
+++ b/modules/currency-l1/src/test/scala/io/constellationnetwork/currency/l1/domain/snapshots/programs/CurrencySnapshotProcessorSuite.scala
@@ -75,7 +75,18 @@ object CurrencySnapshotProcessorSuite extends SimpleIOSuite with TransactionGene
 
           currencySnapshotAcceptanceManager <- CurrencySnapshotAcceptanceManager
             .make(
-              FieldsAddedOrdinals(Map.empty, Map.empty, Map.empty, Map.empty, Map.empty, Map.empty, Map.empty, Map.empty, Map.empty),
+              FieldsAddedOrdinals(
+                Map.empty,
+                Map.empty,
+                Map.empty,
+                Map.empty,
+                Map.empty,
+                Map.empty,
+                Map.empty,
+                Map.empty,
+                Map.empty,
+                Map.empty
+              ),
               Dev,
               LastGlobalSnapshotsSyncConfig(NonNegLong(2L), PosInt(50)),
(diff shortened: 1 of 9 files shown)


gclaramunt (approved):
lgtm

PR #1499, opened 2026-04-30
chore: add nftables byte-rate limiting + prometheus exporter for source nodes

effects-ai:
Per-destination-IP byte-rate cap (2 MB/s) on outbound public-port traffic from the three testnet source nodes, enforced via nftables. 

Three modes (steady / emergency / open) atomically swappable via `firewall-mode`.  

Allowlist bypasses source nodes, prometheus, and snapshot-streaming. 

Includes a Prometheus exporter for nftables counters and Grafana  panels for throughput, drops, mode, and per-peer approximate bytes/s.

### DIFF
--- a/docker/grafana/provisioning/dashboards/tessellation-debug-rest-api-endpoint-sizes.json
+++ b/docker/grafana/provisioning/dashboards/tessellation-debug-rest-api-endpoint-sizes.json
@@ -2219,7 +2219,7 @@
       },
       "id": 1000,
       "panels": [],
-      "title": "nftables Firewall (per source node)",
+      "title": "nftables Firewall (per streaming node)",
       "type": "row"
     },
     {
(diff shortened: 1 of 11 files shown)


docker/grafana/provisioning/dashboards/tessellation-debug-rest-api-endpoint-sizes.json
```diff
@@ -0,0 +1,2714 @@
...
+      },
+      "id": 200,
+      "panels": [],
+      "title": "Request Size by Endpoint",
+      "type": "row"
+    },
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "prometheus"
```
ryle-ai, line 45: `docker/grafana/provisioning/datasources/` only defines ClickHouse; there is no Prometheus datasource with UID `prometheus` in this Grafana instance. All panels in this dashboard will fail to render. Either add a Prometheus datasource YAML here, or consolidate this dashboard into the monitoring stack (where `uid: prometheus` is already provisioned).

docker/nft-rate-limit/config/etc/firewall/modes/steady.nft
```diff
@@ -0,0 +1,45 @@
...
+
+    # Output chain: per-destination-IP byte-rate cap on outbound traffic
+    # from public ports. Counts both rate-passed and allowlisted outbound
+    # so total egress is observable.
+    chain output {
+        type filter hook output priority 0; policy accept;
+
+        # Rate cap only applies to non-allowlisted destinations on
+        # public ports. Drops counted in bytes_dropped.
+        tcp sport { @PUBLIC_PORTS@ } ip daddr != @allow_v4 \
```
ryle-ai, line 47: The `ip daddr != @allow_v4` predicate only matches IPv4 packets. For IPv6 traffic on `@PUBLIC_PORTS@`, this entire rule is a no-op — the rate limiter never fires. If IPv6 reachability on public ports is possible (directly or via dual-stack), those flows bypass the byte-rate cap entirely.

docker/nft-rate-limit/config/etc/firewall/modes/emergency.nft
```diff
@@ -0,0 +1,33 @@
...
+    counter pub_dropped { }
+    counter p2p_dropped { }
+    counter pub_out     { }
+    counter p2p_out     { }
+
+    chain input {
+        type filter hook input priority 0; policy accept;
+
+        # Allowlist short-circuit (preserves streaming<->streaming + monitoring)
+        ip saddr @allow_v4 accept
```
ryle-ai, line 31: `ip saddr @allow_v4 accept` is IPv4-only. Allowlisted nodes connecting via IPv6 are not exempted and will hit the drop rules on lines 23–24. Consistent with the IPv4-only design, but worth documenting explicitly so operators adding an IPv6 allowlisted host don't wonder why emergency mode still blocks them.

docker/nft-rate-limit/config/usr/local/bin/nftables_exporter.py
```diff
@@ -0,0 +1,149 @@
...
+"""
+nftables_exporter - lightweight Prometheus exporter for nftables counters,
+meters, and the firewall-mode active-mode marker.
+
+Listens on :9437/metrics. Reads `nft -j list counters` and `nft -j list meters`
+on each scrape (no caching - scrape interval is 15s and nft is fast).
+
+Runs as root via systemd unit because `nft list` requires CAP_NET_ADMIN.
+"""
+
```
ryle-ai, line 11: `LISTEN_HOST = "0.0.0.0"` exposes the metrics endpoint on all interfaces. Port 9437 is not covered by any rule in steady or emergency mode (only `@PUBLIC_PORTS@` and `@P2P_PORTS@` are referenced), so it remains reachable from any IP regardless of firewall mode. If restricting scraping to the monitoring host is desirable, add an inbound accept rule for 9437 from the allowlist and drop-otherwise in emergency mode.

ryle-ai (changes requested):
## Missing Prometheus datasource in target Grafana instance

The new dashboard (`tessellation-debug-rest-api-endpoint-sizes.json`) is provisioned into `docker/grafana/provisioning/dashboards/`. That Grafana instance's datasource directory (`docker/grafana/provisioning/datasources/`) only defines ClickHouse — there is no Prometheus datasource configured. The dashboard references `"uid": "prometheus"` throughout, so every panel will show "Datasource not found" when loaded. The dashboard already exists in `docker/monitoring/grafana/provisioning/dashboards/` (where Prometheus UID `prometheus` is defined), but that version has 20 panels vs. 39 here — they diverged.

Fix: add a Prometheus datasource YAML to `docker/grafana/provisioning/datasources/` alongside the existing `clickhouse.yaml`, or move this dashboard exclusively to the monitoring stack where the datasource is already provisioned.

---

**Port 9437 not in Prometheus scrape config.** The exporter listens on `:9437/metrics` but neither `prometheus.yaml` nor `prometheus.local.yaml` in the monitoring stack includes it as a static target. The service will run and expose metrics, but nothing will collect them.

---

**IPv6 gap in both steady and emergency mode.** The `inet` family handles both IPv4 and IPv6, but all matching uses IPv4-specific expressions (`ip daddr`, `ip saddr`, `type ipv4_addr`). In steady mode the rate-limiting rule fires only when `ip daddr != @allow_v4` matches — which is never true for IPv6 packets, so IPv6 outbound on public ports is entirely unrate-limited. In emergency mode `ip saddr @allow_v4 accept` won't fire for allowlisted nodes connecting over IPv6, leaving them subject to the drop rules. If none of the allowlisted nodes ever use IPv6 this is a non-issue, but it's an invisible gap worth documenting if not by design.

---

**`flush ruleset` wipes all nftables state.** The comment correctly notes this is an atomic replace, but it also nukes any rules Docker or other services have installed. If the streaming nodes run Docker alongside this firewall, applying any mode will break container networking. Worth documenting in a README or comment.

ryle-ai (dismissed):
## Missing PR description

No description provided. Given the scope — nftables byte-rate limiting + Prometheus exporter + Grafana dashboard — at minimum document: which nodes this is deployed on, what problem triggered the 2 MB/s cap, and the operational runbook for switching modes.

## Prometheus datasource missing from docker/grafana stack

`docker/grafana/provisioning/datasources/` only contains `clickhouse.yaml`. Every panel in the new `tessellation-debug-rest-api-endpoint-sizes.json` dashboard (39 panels) references `"uid": "prometheus"` — they will all fail to render. The monitoring stack version (`docker/monitoring/grafana/`) works because `datasources.local.yaml` defines both Prometheus and ClickHouse, but the `docker/grafana/` stack has no Prometheus datasource. Either add a `prometheus.yaml` datasource file to `docker/grafana/provisioning/datasources/` or consolidate this dashboard exclusively into the monitoring stack.

## Port 9437 not in Prometheus scrape config

`docker/monitoring/prometheus/prometheus.yaml` scrapes ports 9090, 9000/9010 (tessellation), 9256 (process-exporter), and 9435 (network-process-exporter). Port 9437 (nftables_exporter) is absent. The exporter will start and expose metrics, but nothing collects them. The two existing node-level exporters (`process-exporter`, `network-process-exporter`) were added with matching scrape jobs; this one was not.

## Two diverged versions of the same dashboard

`docker/grafana/provisioning/dashboards/tessellation-debug-rest-api-endpoint-sizes.json` has 39 panels; `docker/monitoring/grafana/provisioning/dashboards/tessellation-debug-rest-api-endpoint-sizes.json` has 20. These will drift further with every subsequent update. If the intent is one canonical dashboard, pick a home; if both are needed, document why they differ.

ryle-goehausen (approved):

ryle-ai (commented):
Infra-only change; the datasource / scrape-config / IPv6 / `flush ruleset` items were already raised in review, so I won't repeat them. One thing not yet called out:

**Unbounded per-destination state on a public-facing node.** The dynamic sets `pub_out_dst` / `p2p_out_dst` and the `byterate` meter are all keyed by `ip daddr` with no `timeout` and no `size` cap. Since the output chain matches `sport { @PUBLIC_PORTS@ }` — i.e. responses to *arbitrary external snapshot consumers* — the destination keyspace is effectively unbounded. Two consequences:

- Kernel memory for these sets/meter grows monotonically with the number of distinct client IPs the node ever responds to, and is only reclaimed on a mode reload (`flush ruleset`).
- The exporter emits one Prometheus series per set element (`nftables_set_element_{packets,bytes}_total{...,key="<ip>"}`), so Prometheus cardinality tracks that same unbounded client-IP set.

For an internal, mostly-allowlisted node this may be tolerable, but on a node serving the public REST API it's the classic cardinality/OOM footgun. Consider adding a `timeout` (and/or `size`) to the dynamic sets, or restricting the per-IP breakdown to a bounded keyspace. Non-blocking, but worth a note in the runbook if left as-is.

PR #1500, opened 2026-05-01
fix: flakey e2e tests

effects-ai:
The E2E tests fail occasionally because the timeouts are a bit optimistic. I increased timeouts for the most frequently failing tests.


ryle-ai (commented):
No description. The allow-spends and ordinal-wait fixes are clear from the inline comments, but the `snapshot-streaming` timeout increase (30→45 min) has no documented rationale — it's unrelated to the EPOCH_PROGRESS_BUFFER change or the ordinal-wait logic. Was this hitting the limit independently?

scasplte2 (approved):

PR #1501, opened 2026-05-07
fix: nightly e2e fixes

effects-ai:
Fixes for the Nightly E2E tests.

### DIFF
--- a/.github/action_scripts/delegated_staking/token-lock-replacement-edge-cases.js
+++ b/.github/action_scripts/delegated_staking/token-lock-replacement-edge-cases.js
@@ -532,24 +532,32 @@ const setupNodeParameters = async (urls) => {
 }
 
 /**
- * Clean up all existing delegated stakes for an account.
- * Withdraws all active stakes and waits for pending withdrawals to complete,
- * which automatically unlocks the associated token locks and returns balance.
+ * Clean up all existing delegated stakes and token locks for an account.
  *
- * This prevents state accumulation across repeated runs on persistent clusters
- * (e.g. the nightly E2E environment where the same account is reused hourly).
+ * Withdraws all active stakes, waits for pending withdrawals to drain, then
+ * waits for the address's active token locks to drain. Stake withdrawal emits
+ * a TokenUnlock that eventually frees the underlying lock from currentTokenLocks,
+ * but the lock can persist in the validator's view past the point where pending
+ * withdrawals reach zero, so we must poll the locks endpoint independently.
+ *
+ * Without this, persistent clusters (e.g. the nightly E2E environment that
+ * reuses the same account hourly) accumulate locks until the address hits
+ * TooManyTokenLocksForAddress on the next createTokenLock.
  */
 const cleanupExistingStakes = async (urls, account) => {
   const response = await getAccountDelegatedStakes(urls, account.address)
   const active = response.activeDelegatedStakes || []
   const pending = response.pendingWithdrawals || []
+  const initialLocks = await getActiveTokenLocks(urls, account.address)
 
-  if (active.length === 0 && pending.length === 0) {
-    logWorkflow.info('No pre-existing stakes to clean up')
+  if (active.length === 0 && pending.length === 0 && initialLocks.length === 0) {
+    logWorkflow.info('No pre-existing stakes or token locks to clean up')
     return
   }
 
-  logWorkflow.info(`Cleaning up ${active.length} active stakes and ${pending.length} pending withdrawals`)
+  logWorkflow.info(
+    `Cleaning up ${active.length} active stakes, ${pending.length} pending withdrawals, ${initialLocks.length} token locks`
+  )
 
   for (const stake of active) {
     logWorkflow.info(`  Withdrawing stake ${stake.hash.substring(0, 16)}...`)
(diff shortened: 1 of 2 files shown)


.github/action_scripts/delegated_staking/token-lock-replacement-edge-cases.js
```diff
@@ -565,13 +573,29 @@ const cleanupExistingStakes = async (urls, account) => {
...
       maxStalledChecks: 120,
       interval: 5000,
     },
   )
-  logWorkflow.info('Pre-existing stakes cleaned up, token locks unlocked')
+
+  logWorkflow.info('Pre-existing stakes and token locks cleaned up')
 }
 
 /**
```
ryle-ai, line 611: If `initialLocks.length > 0` but `active.length === 0 && pending.length === 0` (orphaned locks from a previously-interrupted cleanup), no withdrawals are issued and the lock poll below will spin for the full `maxOrdinalMisses: 60` budget (~43 minutes) before failing. This is the expected failure mode on the first post-deploy run if old locks are still present — worth being aware of if the nightly run goes red immediately after merge.

ryle-ai (commented):
The fix is correct. The docstring accurately describes the race: pending withdrawals can reach zero before the corresponding TokenUnlock events are processed out of `currentTokenLocks`, so a second independent poll for locks is the right approach.

One edge case: if a previous run was interrupted after stakes fully drained (active=0, pending=0) but before locks cleared — most likely on the very first nightly run after this fix is deployed, since the old code had no lock-drain poll — the next run will see `active=0, pending=0, initialLocks=N`. No withdrawals are issued, the stakes poll passes immediately, and the locks poll spins for up to ~43 minutes (60 ordinal progressions × ~43s) waiting for locks that have no in-flight withdrawal to release them. This results in a noisy CI failure rather than a silent one, so it's not a blocker, but the first post-deploy run may need manual intervention if the old locks haven't cleared yet.

PR description is empty.

ryle-ai (commented):
Both fixes are correct and address real nightly E2E failures.

**Token lock cleanup:** The two-phase poll (stakes first, then locks) correctly handles the race where `pendingWithdrawals` reaches zero before the corresponding `TokenUnlock` events clear from `currentTokenLocks`. The `withRetryOrdinal` contract is used correctly — the new locks poll is independent of the stakes poll, which is exactly what the race requires.

One edge case to be aware of for the first post-deploy run: if `initialLocks > 0` but `active == 0` and `pending == 0` (orphaned locks from a prior interrupted run where old code left locks behind), the function skips the early-return, issues no withdrawals, immediately passes the stakes poll, then spins on the locks poll waiting for locks that no in-flight withdrawal will ever unlock. After 60 ordinal progressions (~43 minutes), `withRetryOrdinal` throws `"Failed after 60 ordinal progressions. Transaction likely dropped or invalid. Last error: Still N active token locks"` — which is a misleading message since no transaction was submitted. This results in a noisy CI failure rather than the previous silent accumulation, which is strictly better, but operators seeing this error the first time may spend time looking for a dropped transaction that doesn't exist.

**Snapshot-streaming postgres startup:** Verified against `docker/snapshot-streaming/docker-compose.yaml` in develop — `container_name: snapshot-streaming-postgres` is explicitly set, the volume is `ss-pgdata`, and the DATABASE_URL credentials match the compose environment exactly. The TCP readiness check (`-h 127.0.0.1`) correctly avoids the unix-socket race described in the comment. The sequence (postgres up → TCP ready → schema push → app up) is correct.

PR description is empty.

ryle-ai (changes requested):
Two issues, one of which should be cleaned up before merge.

**PR description is empty.** Please document what broke, why it broke, and what each fix does.

**`push: nighty_e2e_fixes` trigger must be removed before merge.** The trigger was added to make it easy to test the workflow changes during PR development, but it shouldn't land in `develop`. The `nightly-rebuild` and `nightly-deploy` workflows share the `nightly-deploy` concurrency group with `cancel-in-progress: true`, so any push to `nighty_e2e_fixes` while a legitimate nightly run is in progress will silently cancel it. After the branch is deleted post-merge the trigger becomes dead config, but it's still messy and represents a real cancellation risk during development. Remove the `push:` block and keep only `schedule` + `workflow_dispatch`.

**Orphaned-locks edge case (non-blocking, first-deploy awareness).** When `active === 0 && pending === 0 && initialLocks > 0` — the state left by the old code on any prior run that exited cleanly — the new code correctly skips the early return, runs the stakes poll (passes immediately), then spins on the locks poll for up to 60 ordinal progressions (~43 minutes) waiting for locks that have no in-flight withdrawal to release them. The error thrown at timeout says `cleanupExistingStakes:tokenLocks: Failed after 60 ordinal progressions. Transaction likely dropped or invalid. Last error: Still N active token locks`, which is misleading since no transaction was submitted. This is a noisy CI failure rather than silent accumulation, so it's strictly better than before, but the first nightly run after merge may need manual intervention.

ryle-ai (changes requested):
The two fixes are correct. The stakes/locks two-phase poll addresses a real race and the postgres startup sequencing is sound. One item needs to be removed before merge.

**`push: nighty_e2e_fixes` trigger must be removed.** `nightly-rebuild.yml` and `nightly-deploy.yml` share the `nightly-deploy` concurrency group with `cancel-in-progress: true`. Any push to `nighty_e2e_fixes` while a nightly deploy or rebuild is in progress silently cancels it. The branch will be deleted after merge so the trigger becomes dead config, but it can cancel legitimate nightly runs during development and during the window between merge and branch deletion.

**Orphaned-locks edge case (non-blocking).** When `active === 0 && pending === 0 && initialLocks > 0` — state left by any prior run that exited cleanly before this fix — the stakes poll passes immediately, then the locks poll spins for up to 60 ordinal progressions (~43 minutes) waiting for locks that have no in-flight withdrawal to release them. `withRetryOrdinal` then throws `cleanupExistingStakes:tokenLocks: Failed after 60 ordinal progressions. Transaction likely dropped or invalid. Last error: Still N active token locks`, which is misleading because no transaction was submitted. The first nightly run after merge may fail this way if old locks are still present.

**`proofsHash` comment missing (non-blocking).** `compose-runner.sh` (the existing local seed-creation path this mirrors) includes an explanatory comment: `"The node uses Brotli-compressed JSON for hashing, which we cannot replicate here. This simplified hash is sufficient for the E2E test since snapshot-streaming does not cryptographically verify the seed proofsHash."` The nightly-rebuild version of the same code omits this, which will puzzle a future reader wondering why the hash is computed this way.

**Empty PR description.**

scasplte2 (approved):

PR #1502, opened 2026-05-08
chore(deps): omnibus dependency bumps

scasplte2:
## Summary

Rebased onto the latest `develop` and refreshed to consolidate the remaining open
Dependabot dependency PRs into a single review surface.

Since the previous revision, develop's v4.1.0 integration (#1523) already advanced
some of these deps past what their Dependabot PRs proposed, so this PR no longer
touches `package.json` at all and now carries **only the transitive lockfile bumps
develop has not moved**.

## What this PR changes (lockfiles only)

| File | Package | From | To | Dependabot PR |
|---|---|---|---|---|
| `package-lock.json` | lodash | 4.17.23 | 4.18.1 | #1488 |
| `package-lock.json` | js-yaml | 4.1.1 | 4.3.0 | #1532 |
| `package-lock.json` | fast-uri | 3.0.1 | 3.1.3 | #1505 |
| `.github/action_scripts/package-lock.json` | brace-expansion | 1.1.12 | 1.1.15 | #1482 |
| `.github/action_scripts/package-lock.json` | path-to-regexp | 0.1.12 | 0.1.13 | #1481 |

These are transitive/dev deps, so there is no `package.json` change. Lockfile entries
were updated with `npm update <pkg> --package-lock-only` in each directory; the resolved
versions are at or above each Dependabot proposal (fast-uri and brace-expansion resolved
one patch higher).

## Already satisfied by develop (no diff here)

develop's v4.1.0 integration (#1523) declares and locks these at or above target, so
nothing is needed for them:

| Package | develop has | superseded Dependabot PR(s) |
|---|---|---|
| axios (root) | ^1.17.0 | #1491 |
| axios (action_scripts) | ^1.18.1 | #1492, #1504 |
| follow-redirects (both) | 1.16.0 | #1496, #1493 |

Note: #1488 / #1482 / #1481 were closed against an earlier revision of this PR, but their
bumps never actually landed on develop (they were lost when develop absorbed v4.1.0). This
revision re-applies them.

Closes #1532, #1505, #1504.

## Test plan

- [ ] CI passes (lint + tests)
- [ ] Lockfiles are reproducible: `npm install --package-lock-only` is a no-op in both dirs
- [ ] No transitive downgrades vs. develop

### DIFF
--- a/.github/action_scripts/package-lock.json
+++ b/.github/action_scripts/package-lock.json
@@ -433,9 +433,9 @@
       }
     },
     "node_modules/brace-expansion": {
-      "version": "1.1.12",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.12.tgz",
-      "integrity": "sha512-9T9UjW3r0UW5c1Q7GTwllptXwhvYmEzFhzMfZ9H7FQWt+uZePjZPjBP/W1ZEyZ1twGWom5/56TF4lPcqjnDHcg==",
+      "version": "1.1.15",
+      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.15.tgz",
+      "integrity": "sha512-EwOCDEex4quD37XhqM3omwtMoJjr//isUZz1JopUNWms+4Z2ViyM/k1YIRePpoVNnQhENnxtFjLaxNHrT7xIUg==",
       "license": "MIT",
       "dependencies": {
         "balanced-match": "^1.0.0",
@@ -1954,9 +1954,9 @@
       }
     },
     "node_modules/path-to-regexp": {
-      "version": "0.1.12",
-      "resolved": "https://registry.npmjs.org/path-to-regexp/-/path-to-regexp-0.1.12.tgz",
-      "integrity": "sha512-RA1GjUVMnvYFxuqovrEqZoxxW5NUZqbwKtYz/Tt7nXerk0LbLblQmrsgdeOxV5SFHf0UDggjS/bSeOZwt1pmEQ==",
+      "version": "0.1.13",
+      "resolved": "https://registry.npmjs.org/path-to-regexp/-/path-to-regexp-0.1.13.tgz",
+      "integrity": "sha512-A/AGNMFN3c8bOlvV9RreMdrv7jsmF9XIfDeCd87+I8RNg6s78BhJxMu69NEMHBSJFxKidViTEdruRwEk/WIKqA==",
       "license": "MIT"
     },
     "node_modules/pbkdf2": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1830,9 +1830,21 @@
       "license": "MIT"
     },
     "node_modules/fast-uri": {
-      "version": "3.0.1",
+      "version": "3.1.3",
+      "resolved": "https://registry.npmjs.org/fast-uri/-/fast-uri-3.1.3.tgz",
+      "integrity": "sha512-i70LwGWUduXqzicKXWshooq+sWL1K3WUU5rKZNG/0i3a1OSoX3HqhH5WbWwTmqWfor4urUakGPiRQcleRZTwOg==",
       "dev": true,
-      "license": "MIT"
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/fastify"
+        },
+        {
+          "type": "opencollective",
+          "url": "https://opencollective.com/fastify"
+        }
+      ],
+      "license": "BSD-3-Clause"
     },
     "node_modules/finalhandler": {
       "version": "1.3.2",
@@ -2574,10 +2586,21 @@
       "license": "MIT"
     },
     "node_modules/js-yaml": {
-      "version": "4.1.1",
-      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.1.1.tgz",
-      "integrity": "sha512-qQKT4zQxXl8lLwBtHMWwaTcGfFOZviOJet3Oy/xmGk2gZH677CJM9EvtfdSkgWcATZhj/55JZ0rmy3myCT5lsA==",
+      "version": "4.3.0",
+      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.3.0.tgz",
+      "integrity": "sha512-1td788aAnnZ5qs7V2QIRl1owjtYpbKt749Y3xauqQgwIIGF/xXWz1wMTEBx5O3LK3lXLVuqXPdPxj2BoFHaW9Q==",
       "dev": true,
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/puzrin"
+        },
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/nodeca"
+        }
+      ],
+      "license": "MIT",
       "dependencies": {
         "argparse": "^2.0.1"
       },
@@ -2686,9 +2709,10 @@
       }
     },
     "node_modules/lodash": {
-      "version": "4.17.23",
-      "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.23.tgz",
-      "integrity": "sha512-LgVTMpQtIopCi79SJeDiP0TfWi5CNEc/L/aRdTh3yIvmZXTnheWpKjSZhnvMl8iXbC1tFg9gdHHDMLoV7CnG+w=="
+      "version": "4.18.1",
+      "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.18.1.tgz",
+      "integrity": "sha512-dMInicTPVE8d1e5otfwmmjlxkZoUpiVLwyeTdUsi/Caj/gfzzblBcCE5sRHV/AsjuCmxWrte2TNGSYuCeCq+0Q==",
+      "license": "MIT"
     },
     "node_modules/lodash.camelcase": {
       "version": "4.3.0",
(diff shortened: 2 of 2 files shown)


package-lock.json
```diff
@@ -2686,9 +2709,10 @@
       }
     },
     "node_modules/lodash": {
-      "version": "4.17.23",
-      "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.23.tgz",
-      "integrity": "sha512-LgVTMpQtIopCi79SJeDiP0TfWi5CNEc/L/aRdTh3yIvmZXTnheWpKjSZhnvMl8iXbC1tFg9gdHHDMLoV7CnG+w=="
+      "version": "4.18.1",
```
ryle-ai, line 2712: `lodash@4.18.1` doesn't appear to be a real release — lodash's latest published version is **4.17.21** and there is no 4.18.x line. The base `4.17.23` here is also non-existent. Please verify this resolves on npm; if not, this pin will 404 on `npm ci`, and if it *does* resolve it's worth double-checking it isn't a maliciously-published version matching the pin. Regenerate via `npm install` so `resolved`/`integrity` come from the registry.

package-lock.json
```diff
@@ -2574,10 +2586,21 @@
       "license": "MIT"
     },
     "node_modules/js-yaml": {
-      "version": "4.1.1",
-      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.1.1.tgz",
-      "integrity": "sha512-qQKT4zQxXl8lLwBtHMWwaTcGfFOZviOJet3Oy/xmGk2gZH677CJM9EvtfdSkgWcATZhj/55JZ0rmy3myCT5lsA==",
+      "version": "4.3.0",
```
ryle-ai, line 2589: `js-yaml@4.3.0` doesn't match any published release — latest is **4.1.0** (and the `4.1.1` base is also fabricated). Same concern as lodash: verify against npm and regenerate the lockfile rather than hand-editing the version/integrity.

.github/action_scripts/package-lock.json
```diff
@@ -433,9 +433,9 @@
       }
     },
     "node_modules/brace-expansion": {
-      "version": "1.1.12",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.12.tgz",
-      "integrity": "sha512-9T9UjW3r0UW5c1Q7GTwllptXwhvYmEzFhzMfZ9H7FQWt+uZePjZPjBP/W1ZEyZ1twGWom5/56TF4lPcqjnDHcg==",
+      "version": "1.1.15",
```
ryle-ai, line 436: The dependabot branch in this repo proposes `brace-expansion@1.1.13`, but this omnibus pins **1.1.15**. Where did 1.1.15 come from? Please confirm it's an actual published version; otherwise align with the 1.1.13 that dependabot found. (`path-to-regexp@0.1.13` below matches its dependabot branch and looks fine.)

ryle-ai (changes requested):
For a lockfile-only PR the one thing that actually matters is whether the pinned versions and integrity hashes correspond to real, published packages. Several of these bump targets do not appear to.

I can't reach npm from this environment, so treat the specifics as "please verify" rather than certainties — but to the best of my knowledge:

- **`lodash` → 4.18.1**: lodash has been effectively frozen at **4.17.21** for years; there is no 4.18.x release. A minor-version bump introducing a new feature line for lodash would be extraordinary/newsworthy. The integrity hash here (`sha512-dMInicTPVE8d1e5o…`) doesn't match the known 4.17.21 artifact.
- **`js-yaml` → 4.3.0**: the latest published js-yaml is **4.1.0**; there is no 4.2.x/4.3.x. (Note the diff's *base* is already `4.1.1`, which also doesn't exist — so `develop` was carrying a fabricated version before this PR too.)
- **`brace-expansion` → 1.1.15** (action_scripts): the actual dependabot branch in this repo proposes **1.1.13**, not 1.1.15. This omnibus diverges from what dependabot found. `path-to-regexp → 0.1.13`, by contrast, matches the dependabot branch and looks fine.

The pattern (a hand-assembled "omnibus" whose targets go beyond, or differ from, what dependabot actually proposed, and in two cases beyond any real published release) is a supply-chain red flag. A lockfile pointing at a non-existent version fails `npm ci` with a 404 in the best case; in the worst case it resolves to a version that was *published specifically to match this pin*, which is exactly the shape of a dependency-confusion / hijack attack. These are dev/build-tooling transitive deps (commitlint / husky / conventional-changelog), so the blast radius is the CI/build environment — but that's where signing and secrets live, so it's not low-risk.

Before merging, please regenerate the lockfile with `npm install` against the public registry rather than hand-editing versions, and confirm every `version`/`resolved`/`integrity` triple resolves on npm. If lodash 4.18.1 / js-yaml 4.3.0 genuinely don't exist, drop them; if they do, the integrity hashes need to come straight from the registry.

Minor, not a blocker: the `fast-uri` license correction MIT → BSD-3-Clause is accurate (fast-uri is BSD-3-Clause), so that part is a genuine fix.

scasplte2:
Thanks for the careful supply-chain review, that is exactly the right instinct for a lockfile-only PR. I do have registry access from my environment, so I verified every pin against the public registry, and all of them are real, published versions whose integrity hashes match the registry. The core correction: **lodash and js-yaml are no longer frozen** at the versions in your reference; both shipped releases past those points (after the Jan 2026 knowledge cutoff).

**Registry config** (no proxy / dependency-confusion surface):
- `npm config get registry` -> `https://registry.npmjs.org/`
- No `.npmrc` in the repo root, `~`, or `.github/action_scripts` (checked all three).

**Versions exist on the public registry** (`curl https://registry.npmjs.org/<pkg>`):

| Package | Pin | Published on npmjs.org? | Integrity == registry `.dist.integrity`? |
|---|---|---|---|
| lodash | 4.18.1 | yes, it is currently `dist-tags.latest` (4.17.23 and 4.18.0 also published) | MATCH |
| js-yaml | 4.3.0 | yes (latest is now 5.2.0; 4.1.1 / 4.2.0 / 4.3.0 all published) | MATCH |
| fast-uri | 3.1.3 | yes | MATCH |
| brace-expansion | 1.1.15 | yes | MATCH |
| path-to-regexp | 0.1.13 | yes | MATCH |

All 5 integrity triples in the lockfile match what the registry returns for that exact version.

On the specific points:

- **lodash 4.18.1 / base 4.17.23**: both are published; 4.18.1 is `latest`. The 4.18.x line is real and post-dates the long 4.17.21 freeze.
- **js-yaml 4.3.0 / base 4.1.1**: both published, so develop's base was genuine, not fabricated.
- **brace-expansion 1.1.15 vs dependabot 1.1.13**: not hand-editing. `npm update brace-expansion --package-lock-only` resolves to the highest version satisfying the existing `^1.1.x` range, which is 1.1.15 (published, >= dependabot's 1.1.13, same range). Happy to pin exactly 1.1.13 if you would prefer to mirror the dependabot branch, but 1.1.15 is the correct npm-resolved result.
- **fast-uri MIT -> BSD-3-Clause**: agreed, genuine fix.

**Method**: lockfiles were produced with `npm update <pkg> --package-lock-only` against the public registry (not hand-edited), so every `version` / `resolved` / `integrity` came straight from npm; `npm install --package-lock-only` is then a no-op (idempotent) in both directories.

Reproduce:
```
npm config get registry
curl -s https://registry.npmjs.org/lodash | python3 -c "import sys,json;print(json.load(sys.stdin)['dist-tags']['latest'])"
```

package-lock.json
ryle-ai, line 2712: Confirmed legitimate despite the earlier "no 4.18.x line" flag: `origin/dependabot/npm_and_yarn/lodash-4.18.1` exists in this repo and contains the identical integrity hash (`sha512-dMInic...`). Dependabot resolves versions from the registry, so this is a real published release, and the matching hash means it wasn't hand-edited.

package-lock.json
```diff
@@ -1830,9 +1830,21 @@
       "license": "MIT"
     },
     "node_modules/fast-uri": {
-      "version": "3.0.1",
+      "version": "3.1.3",
```
ryle-ai, line 1833: fast-uri 3.1.3 is one patch above dependabot's 3.1.2 branch, and unlike the other four bumps there's no dependabot branch pinning 3.1.3 to cross-check against. Consistent with `npm update` grabbing the latest in-range patch, but please confirm 3.1.3 specifically resolves on the registry (the others are corroborated by dependabot-generated integrity hashes).

ryle-ai (commented):
Lockfile-only PR; the only thing that matters is whether the pins/integrity map to real published artifacts. The earlier "lodash 4.18.1 / js-yaml 4.3.0 look fabricated" concern is resolvable from repo-local evidence and does not hold up:

- This repo already has dependabot branches proposing exactly these versions: `origin/dependabot/npm_and_yarn/lodash-4.18.1` and `origin/dependabot/npm_and_yarn/js-yaml-4.3.0`. Dependabot only opens PRs for versions it resolves from the configured registry, so these are real releases, not hand-invented pins.
- More conclusively, the integrity hashes in this omnibus are byte-identical to the ones dependabot generated in those branches (lodash `sha512-dMInic...`, js-yaml `sha512-1td788...`). Independent regeneration producing the same integrity is what you'd expect from a genuine registry artifact, not from a hand-edited hash.
- The base versions the earlier review called fabricated (lodash 4.17.23, js-yaml 4.1.1) also came from dependabot — `upstream/dependabot/npm_and_yarn/lodash-4.17.23` and `js-yaml-4.1.1` branches exist. So `develop`'s base wasn't fabricated either.

The divergences from dependabot's patch levels are consistent with the author's stated method (`npm update <pkg> --package-lock-only`, which resolves to the highest version satisfying the existing range): brace-expansion 1.1.15 vs dependabot's 1.1.13 (range `^1.1.x`), and fast-uri 3.1.3 vs dependabot's 3.1.2 (range `^3.0.1`). path-to-regexp 0.1.13 matches dependabot exactly.

One caveat worth confirming: fast-uri **3.1.3** is the only pin here with no corroborating dependabot branch (dependabot's is 3.1.2, and there's no 3.1.3 branch). lodash/js-yaml/brace-expansion/path-to-regexp all have a dependabot branch to cross-check against; fast-uri 3.1.3 rests solely on the author's registry check. Low risk (dev-only transitive of ajv), but it's the one line I can't independently verify from the repo.

Lockfile is internally consistent — both the `packages` (`node_modules/*`) and legacy `dependencies` sections were updated for all three root packages, and there are no other tree occurrences left stale. These are all dev/build-tooling deps. No blockers.

ryle-goehausen (approved):

PR #1515, opened 2026-06-09
fix: meter actual bytes for combined-stream response-size metric

effects-ai:


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/CombinedSnapshotCheckpointFileSystemStorage.scala
```diff
@@ -102,23 +102,25 @@ final class CombinedSnapshotCheckpointFileSystemStorage[
...
+        val bodyWithPermit: Stream[F, Byte] =
+          Stream.resource(concurrentStreams.permit).flatMap { _ =>
+            fileStream
+          }
+
+        // Chunked/streamed response: intentionally no Content-Length. The body is
+        // streamed lazily from disk and clients rarely drain the whole file, so a
+        // full-file Content-Length both violated RFC 7230 (must not accompany
+        // Transfer-Encoding) and made the response-size metric attribute the entire
+        // file per connection. Actual bytes are now metered in MetricsMiddleware.
```
ryle-ai: This claim is inaccurate for the route that serves this response. `GET /latest/combined/stream` doesn't contain `"rumor"`, so `isHistogramRoute` returns `false` and the new `onFinalize` byte-counting block in `MetricsMiddleware` never fires for it. The RFC fix (removing `Content-Length`) is correct on its own; the comment should not assert that bytes are now metered when they aren't for this path.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/http/p2p/middlewares/MetricsMiddleware.scala
```diff
@@ -89,7 +86,22 @@ object MetricsMiddleware {
...
             _ <- Async[F].start(metricsRecording)
-          } yield response
+            finalResponse <-
+              if (isHistogramRoute(req.pathInfo.renderString))
+                Ref[F].of(0L).map { bytesRef =>
+                  response.withBodyStream(
+                    response.body.chunks
+                      .evalTap(chunk => bytesRef.update(_ + chunk.size.toLong))
+                      .unchunks
+                      .onFinalize(
```
ryle-ai, line 96: Unlike the rest of the metrics recording in this middleware (wrapped in `Async[F].start()` precisely to avoid blocking or failing the response), an error from `recordSizeHistogram` here propagates out of `onFinalize` and terminates the body stream abnormally. With in-process Micrometer this is low risk, but it's architecturally inconsistent with the fire-and-forget pattern used a few lines up. Consider `.handleErrorWith(_ => Async[F].unit)` on the `onFinalize` action.

ryle-ai (commented):
The RFC 7230 fix (removing `Content-Length` from chunked responses) is correct, and the byte-counting approach via `.chunks.evalTap.unchunks.onFinalize` is sound.

Two issues worth noting before this ships:

**The comment about MetricsMiddleware metering doesn't hold for the actual route**

The combined snapshot stream is served at `GET /latest/combined/stream` (SnapshotRoutes.scala). That path doesn't contain `"rumor"`, so `isHistogramRoute` returns `false` for it. The new `onFinalize` body-counting block in MetricsMiddleware is guarded by the same check — it runs only when `isHistogramRoute` is true. This means the comment in `CombinedSnapshotCheckpointFileSystemStorage` promising "actual bytes are now metered in MetricsMiddleware" is not accurate for this route. The practical impact is no regression (the histogram was never recorded for this path before either), but the comment gives a false justification for the change and may cause confusion when someone later tries to add proper coverage.

If the intent is to actually meter combined snapshot stream responses, `isHistogramRoute` needs to be extended. If that's out of scope here, the comment should be narrowed to "RFC 7230 fix" only, without the claim about metering.

**`ClientMetricsMiddleware` is left with the same gap**

`ClientMetricsMiddleware` still uses `response.contentLength.traverse_` for client-side response-size histograms. Any peer responding with chunked transfer encoding (no `Content-Length`) will silently emit nothing for `dag_http_client_response_size`. The server-side fix is not mirrored here.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/http/p2p/middlewares/MetricsMiddleware.scala
```diff
@@ -89,7 +86,23 @@ object MetricsMiddleware {
 
             } yield ()
             _ <- Async[F].start(metricsRecording)
-          } yield response
+            finalResponse <-
```
ryle-ai, line 89: `isHistogramRoute` unconditionally returns `true` on the current branch, so every HTTP response body is wrapped with `chunks.evalTap(bytesRef.update).unchunks.onFinalize`. That's intentional for accurate metering, but it means the same streaming byte-counting approach needs to be applied in `ClientMetricsMiddleware` too — otherwise the server-side fix is complete while the client-side metric (`dag_http_client_response_size`) stays broken for all chunked peer responses.

ryle-ai (commented):
`ClientMetricsMiddleware` was not updated alongside the server-side fix. It still uses `response.contentLength.traverse_` to record `dag_http_client_response_size` (line 61 of `ClientMetricsMiddleware.scala`). Now that `CombinedSnapshotCheckpointFileSystemStorage` no longer sends `Content-Length` on streaming responses, any node acting as an HTTP client to fetch from `/latest/combined/stream` (or any other chunked peer endpoint) will silently emit nothing for `dag_http_client_response_size`. The same streaming byte-counting approach used in `MetricsMiddleware` needs to be mirrored on the client side.

Context on the prior review comments: the `isHistogramRoute` predicate was changed to `true` (allow all routes) in a prior commit on `develop`, before this PR. The `onFinalize` byte-counting block in `MetricsMiddleware` therefore fires for every response, including `GET /latest/combined/stream`. The assertion in the ryle-ai review that this path "won't be metered" because it doesn't contain `"rumor"` is not accurate against the current branch state.

ryle-goehausen:
This may already be on dev

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/storage/CombinedSnapshotCheckpointFileSystemStorage.scala
```diff
@@ -102,23 +102,25 @@ final class CombinedSnapshotCheckpointFileSystemStorage[
...
+          Stream.resource(concurrentStreams.permit).flatMap { _ =>
+            fileStream
+          }
+
+        // Chunked/streamed response: intentionally no Content-Length. The body is
+        // streamed lazily from disk and clients rarely drain the whole file, so a
+        // full-file Content-Length both violated RFC 7230 (must not accompany
+        // Transfer-Encoding) and made the response-size metric attribute the entire
+        // file per connection.
+        Response[F](
```
ryle-ai, line 116: This rewrite of `getAsHttpResponse` is based on an older develop. Current develop's version is sidecar/ETag-aware (`getCachedHash` + `etagFor`, keeps `Content-Length`, drops `Transfer-Encoding`), and the surrounding class has grown `hashCache`, `readSidecar`, `getCheckpointSize` (bandwidth limiter) etc. — all consumed by `SnapshotRoutes.scala`. Landing this branch would delete that caching/limiter feature and break its call sites. This half needs to be rebased onto current develop (where it's largely already resolved) rather than merged.

ryle-ai (commented):
Confirming ryle-goehausen's hunch ("This may already be on dev") with specifics — this PR is superseded by develop and shouldn't be merged as-is.

**The MetricsMiddleware half is already on develop, verbatim.** `git diff origin/develop origin/fix_combined_snapshot_metrics -- .../MetricsMiddleware.scala` is empty: `isHistogramRoute` already returns `true` and the `.chunks.evalTap(bytesRef).unchunks.onFinalize(recordSizeHistogram)` block already exists on develop. So the byte-metering intent is already shipped there. (This also moots the earlier inline concerns about the `"rumor"` guard and the `onFinalize` error propagation — the current develop code already returns `true` and already has `.handleErrorWith(_ => Async[F].unit)`.)

**The storage half is stale and would regress newer develop work.** Develop's `CombinedSnapshotCheckpointFileSystemStorage` has since grown an entire ETag/304 caching layer (sidecar files, `hashCache`, `getCachedHash`, `etagFor`, sidecar-aware `getAsHttpResponse`) plus `getCheckpointSize` feeding a per-IP bandwidth limiter — introduced by #1523. `SnapshotRoutes.scala` on develop actively calls `getCachedHash`/`etagFor`/`getAsHttpResponse` (lines ~344-392) and `getCheckpointSize` (lines ~634-644), and there's a dedicated `CombinedSnapshotCheckpointSidecarSuite`. This PR rewrites `getAsHttpResponse` from an older base and removes all of that machinery (and the `hashCache` constructor param). Merging it would either conflict hard or, if resolved in favor of this branch, delete the sidecar ETag caching + bandwidth-limiter feature and break those call sites.

Note the two branches also resolve the RFC 7230 conflict in opposite directions: this PR drops `Content-Length` and keeps `Transfer-Encoding: chunked`; develop keeps `Content-Length` (dropping `Transfer-Encoding`) while still getting accurate metering from the `onFinalize` counter in MetricsMiddleware. Develop's approach is the strictly more capable one — it preserves `Content-Length` for clients (so `dag_http_client_response_size` on the client middleware still records for this route) while metering actual streamed bytes server-side.

Recommendation: close this PR, or rebase onto current develop — at which point both files reduce to no-ops / conflicts against work that's already landed.

scasplte2:
Closing as superseded — `develop` already carries an equivalent-or-better fix for both problems this PR targets.

**Metric over-counting:** the change to `MetricsMiddleware.scala` in this PR is already on `develop` verbatim — the two files are the same git blob (`6cd89a3da`). `develop` meters actual streamed bytes off the response body (the `Ref` + `evalTap` chunk counter) rather than trusting the `Content-Length` header, which is exactly what this PR introduced.

**RFC 7230 (`Content-Length` + `Transfer-Encoding: chunked` together):** `develop` resolved this via the opposite-but-valid approach in `CombinedSnapshotCheckpointFileSystemStorage`. Instead of keeping chunked and dropping `Content-Length`, it dropped `Transfer-Encoding: chunked` and keeps an accurate `Content-Length(size)`, and layered on ETag/304 caching + the on-disk sidecar fallback (`getAsHttpResponse` / `readBytesAsStream` / `readSidecar`). Because the metric no longer reads the header, the retained `Content-Length` doesn't reintroduce the over-counting this PR was avoiding.

Net effect: resolving the conflict toward `develop` leaves this PR contributing no change, while resolving toward the PR would regress the newer ETag caching. Nothing left to merge here.

Superseded by the snapshot-streaming ETag/sidecar work merged via #1523.

PR #1518, opened 2026-06-10
chore: add terraform deployment scripts

effects-ai:


deploy/terraform/main.tf
```diff
@@ -0,0 +1,248 @@
...
+      source_ips = var.admin_cidrs
+    }
+  }
+
+  # Prometheus — monitoring node only (loopback in practice)
+  rule {
+    direction  = "in"
+    protocol   = "tcp"
+    port       = "9090"
+    source_ips = var.admin_cidrs
```
ryle-ai: This Prometheus (9090) rule uses `source_ips = var.admin_cidrs` unguarded, while every sibling admin rule — ClickHouse (8123) just above, and Grafana/CLI in the base firewall — is wrapped in `dynamic "rule" { for_each = length(var.admin_cidrs) > 0 ? [1] : [] }`. If an operator sets `admin_cidrs = []` (the natural way to lock these ports down), the ClickHouse/Grafana rules correctly vanish but this one remains with an empty `source_ips`, which `terraform apply` rejects. Wrap it in the same guard for consistency.

deploy/terraform/main.tf
```diff
@@ -0,0 +1,248 @@
...
+  cluster_public_cidrs = [for ip in module.cluster.all_public_ips : "${ip}/32"]
+}
+
+resource "hcloud_firewall" "scrape" {
+  name = "${var.environment}-scrape"
+
+  # JMX (9100) + process (9436) / nftables (9437) / network-process (9435)
+  # exporters — scraped by the monitoring node; scoped to the whole cluster.
+  dynamic "rule" {
+    for_each = toset(["9100", "9256", "9435", "9436", "9437"])
```
ryle-ai: This opens 9100/9256/9435/9436/9437, but `prometheus.yaml` only scrapes 9256 (process-exporter) and 9435 (network-process-exporter) — nothing targets JMX 9100, 9436, or 9437. So the "JMX (9100), no auth — never widen" rationale describes an exposure that isn't actually wired up, and the port labels in the comment (process=9436) disagree with node-init.tpl (process-exporter=9256). Harmless since it's scoped to cluster IPs, but worth trimming to the ports actually scraped / fixing the labels so this doesn't mislead the next operator about what's exposed.

ryle-ai (commented):
Infra-only PR (Terraform, a GHCR publish workflow, and deploy scripting) — no protocol/Scala code, so nothing here blocks a merge. The scripting is thoughtful and the trade-offs (public-IP addressing, non-destructive SS resume/trim, registry-vs-build image source) are well documented. A few inconsistencies worth tightening before this becomes the standard deploy path:

- **World-open defaults are committed.** `environments/testnet.tfvars` ships `allowed_ssh_cidrs`/`admin_cidrs` = `0.0.0.0/0`, and `.gitignore` force-tracks it (`!environments/testnet.tfvars`). An `apply` without CLI overrides therefore exposes SSH, the CLI ports (9002/9012), Grafana (3000) and ClickHouse (8123) to the world. This is called out in the README/comments as a placeholder to replace, so it's a documented posture rather than a bug — but since the checked-in env file is the one people will `apply` verbatim, consider defaulting the committed file to a locked-down/empty CIDR set and forcing operators to opt into exposure.
- Firewall/UFW vs. what Prometheus actually scrapes is slightly out of sync (see inline on the `scrape` firewall).

Inline notes below.

deploy/terraform/README.md
```diff
@@ -0,0 +1,92 @@
...
+export TF_VAR_team_ssh_keys='["ssh-ed25519 AAAA...","ssh-ed25519 BBBB..."]'
+
+terraform plan  -var-file=environments/testnet.tfvars
+terraform apply -var-file=environments/testnet.tfvars
+
+# Public IPs for the IP-rewrite map / Prometheus targets:
+terraform output -json all_public_ips
+```
+
+Set real `allowed_ssh_cidrs` / `allowed_grafana_cidrs` in the env tfvars before
```
ryle-ai: `allowed_grafana_cidrs` isn't a declared variable — Grafana (and Prometheus/ClickHouse) scoping is driven by `admin_cidrs`. Update this line so someone doesn't go looking for a var that doesn't exist.

docker/entrypoint.sh
```diff
@@ -99,8 +99,10 @@ fi
 
 export RUN_COMMAND="run-validator"
 
-# Add Java 21 module access flags for Kryo serialization
-export CL_DOCKER_JAVA_OPTS="${CL_DOCKER_JAVA_OPTS:-} --add-opens=java.base/java.lang.invoke=ALL-UNNAMED --add-opens=java.base/java.util=ALL-UNNAMED --add-opens=java.base/java.security=ALL-UNNAMED"
+# JVM profile: default matches release/testnet's tuning, but a CL_DOCKER_JAVA_OPTS
+# from the deploy .env replaces it wholesale (heap/GC conflict if merely appended),
+# plus Java 21 module access flags for Kryo serialization.
+export CL_DOCKER_JAVA_OPTS="${CL_DOCKER_JAVA_OPTS:--Xms2g -Xmx8g -XX:+UseG1GC -XX:MaxGCPauseMillis=200} --add-opens=java.base/java.lang.invoke=ALL-UNNAMED --add-opens=java.base/java.util=ALL-UNNAMED --add-opens=java.base/java.security=ALL-UNNAMED"
```
ryle-ai, line 119: This flips the default heap for *every* container that runs through this entrypoint, not just the Hetzner nodes. On a Hetzner box (1 gl0 + 1 gl1) the new `-Xms2g -Xmx8g` default is what you want. But the local test stack brings up several node containers on a single host and doesn't set `CL_DOCKER_JAVA_OPTS` (it's commented out in `docker-env-setup.sh`), so each JVM now commits `-Xms2g` up front and can grow to `-Xmx8g` — a multi-container local/CI run reserves 12 GB+ immediately where it previously used JVM ergonomic defaults. Consider exporting this JVM profile in the deploy path's generated per-node `.env` (remote-deploy.sh) instead of changing the shared default, so local/CI behavior is unchanged. (Separately: with this 8g default and no per-container override in the deploy `.env`, gl0+gl1 land at 8g+8g, not the 10g+8g the `node_server_type` doc comment describes.)

ryle-ai (commented):
Infra-only change (Terraform, a GHCR publish workflow, and deploy scripting) — no protocol/Scala code, nothing that blocks a merge. The firewall scoping and world-open `*.tfvars` defaults are already called out in the existing review, so I won't re-litigate those. One additional item that isn't specific to the new Hetzner path:

- **`docker/entrypoint.sh` changes a *shared* default, not just the new deploy path.** This entrypoint is used by every dockerized node — including the local multi-node test stack (`compose-runner.sh`), nightly, and CI e2e — none of which set `CL_DOCKER_JAVA_OPTS` (it's only commented-out in `docker-env-setup.sh`). See the inline note; the Hetzner deploy would be better served by exporting the profile in the generated per-node `.env` rather than baking it into the shared default.

Also a minor doc nit: `deploy/terraform/environments/nightly.tfvars` ships the same `0.0.0.0/0` `allowed_ssh_cidrs`/`admin_cidrs` placeholders as `testnet.tfvars`; the existing review only names `testnet.tfvars`, but both are committed and `apply`-able verbatim.

docker/bin/remote-deploy.sh
```diff
@@ -40,11 +53,28 @@ else
...
+NODE0_ADDRESS=$(cat "$NODE_KEYS_DIR/0/address" 2>/dev/null || echo "")
+
+# Node phases (build -> ship -> start -> tx-sender). Skipped with --skip-nodes so that
+# snapshot-streaming / monitoring can be (re)deployed without restarting node containers
+# (restarting node-0 trips the genesis guard on a live chain). Default: run them, so
+# nightly behaviour is unchanged.
+if [ "$SKIP_NODES" != "true" ]; then
+# === Phase 1: Build image (skipped when pulling from a registry) ===
+rm -rf "$STAGING"
+mkdir -p "$STAGING"
```
ryle-ai, line 102: `mkdir -p "$STAGING"` is now inside the `SKIP_NODES != true` guard, but Phase 6 (snapshot-streaming) runs unconditionally and writes `$STAGING/ss-application.conf` / `$STAGING/ss-docker-compose.yaml`. On a `--skip-nodes` run — which is exactly the SS/monitoring-only redeploy path this flag is for — `/tmp/remote-deploy` is never created, so the first `cat > "$SS_CONFIG"` in Phase 6 fails (`No such file or directory`) under `set -e`. It happens to work only if an earlier full deploy left the dir behind. Suggest an unconditional `mkdir -p "$STAGING"` (e.g. at the top, or in Phase 6 before writing SS_CONFIG).

deploy/terraform/environments/nightly.tfvars
```diff
@@ -0,0 +1,30 @@
...
+# dedicated monitoring node, each source node with a state volume. This profile
+# just runs that topology with cheaper, nightly-sized boxes in the EU.
+environment = "nightly"
+
+# Location: Helsinki.
+location = "hel1"
+
+# Server types — shared vCPU, cheaper than the testnet defaults
+node_server_type       = "cpx52" # 12 vCPU / 24 GB
+streaming_server_type  = "cpx52" # 2 vCPU / 4 GB
```
ryle-ai: `streaming_server_type = "cpx52"` is commented `# 2 vCPU / 4 GB`, but `cpx52` is the same 12 vCPU / 24 GB box as `node_server_type` right above (and as `testnet.tfvars` labels it). Either the comment is stale or streaming was meant to be a smaller box (e.g. `cpx22`) given the "cheaper, nightly-sized boxes" intent — as written, nightly streaming is provisioned identically to a source node. Worth fixing so the cost/sizing isn't misleading.

ryle-ai (commented):
Infra-only PR (Terraform, a GHCR publish workflow, deploy scripting, and the docker entrypoint/genesis-guard restructuring) — no protocol/Scala code, so nothing here blocks a merge. The world-open `*.tfvars` defaults, the unguarded Prometheus (9090) `admin_cidrs` rule, the scrape/UFW-vs-actual-ports mismatch, the stale `allowed_grafana_cidrs` README line, and the shared `entrypoint.sh` heap default have already been called out; I won't re-litigate them.

One functional issue not covered by the existing review, plus a minor tfvars nit — both inline.

The main one: `docker/bin/remote-deploy.sh` moves `rm -rf/mkdir -p "$STAGING"` *inside* the new `if [ "$SKIP_NODES" != "true" ]` guard, but Phase 6 (snapshot-streaming) runs outside that guard and still writes `$STAGING/ss-application.conf` and `$STAGING/ss-docker-compose.yaml`. Since the entire point of `--skip-nodes` is to redeploy SS without touching the node containers, a cold `--skip-nodes` run (fresh CI runner, or after `/tmp` is cleared) will `cat >` into a `/tmp/remote-deploy` that was never created and abort under `set -e`. It only appears to work when a prior full deploy on the same host left the dir behind.

.github/workflows/publish-node-image.yml
```diff
@@ -0,0 +1,183 @@
...
+# The snapshot-streaming image is built from release/testnet (tessellation SDK + SS source)
+# and tagged with the SAME version as the tessellation node image.
+#
+# ONE-TIME SETUP: the first push creates each package as PRIVATE. Make BOTH packages
+# (tessellation, snapshot-streaming) PUBLIC once via the org's Packages settings, otherwise
+# nodes need `docker login` to pull.
+
+on:
+  push:
+    branches: [develop, release/testnet, release/integrationnet, release/mainnet, hetzner_terraform_scripts]
```
ryle-ai: This ships the dev-iteration branch (`hetzner_terraform_scripts`) in the trigger, plus the two special-cases that depend on it (the `ref: ... && 'release/testnet' || github.ref` checkout override and the `CHANNEL=testnet` arm in *Compute channel*). The comments here and at those sites say to remove this once merged — worth doing before/at merge, since it's dead plumbing on `develop` and, while live, causes a dev-branch push to overwrite the real `:testnet` channel tag from `release/testnet`.

Separately: adding `develop` here means every push to `develop` now triggers a full `sbt *assembly` + docker build + GHCR push of `:develop`/`:VER`/`:sha`. Confirm that continuous publishing (and the one-time make-package-public step) is intended.

ryle-ai (commented):
Infra-only PR (Terraform, a GHCR publish workflow, deploy scripting, and the entrypoint genesis-guard restructuring) — no protocol/Scala code, nothing that blocks merge. The rollback-first entrypoint guard and the non-destructive SS resume/trim logic look sound.

The substantive items are already well-covered by the existing review — world-open `*.tfvars` defaults (both `testnet.tfvars` and `nightly.tfvars`), the unguarded Prometheus (9090) `admin_cidrs` rule vs. the guarded siblings, the scrape-firewall/UFW-vs-actually-scraped-ports drift, the stale `allowed_grafana_cidrs` README line, the shared `entrypoint.sh` heap default, the `nightly.tfvars` streaming sizing comment, and the `mkdir -p "$STAGING"` that moved inside the `SKIP_NODES` guard (breaks Phase 6 on a cold `--skip-nodes` run). I confirmed those and won't re-litigate them.

Two smaller things not covered:

- **Leftover dev-branch handling in `publish-node-image.yml` (inline).** The workflow ships with `hetzner_terraform_scripts` still in the push trigger and the two special-cases that key off it (`ref` override on checkout, and the `CHANNEL=testnet` branch in *Compute channel*). Its own comments say to drop this once merged. As-is a push to the dev branch publishes to the real `:testnet` channel tag from `release/testnet`, and more relevantly, once this lands on `develop` every `develop` push now runs a full sbt assembly + docker build + GHCR push (`:develop`). Worth stripping the dev-branch plumbing and confirming continuous publishing on `develop` is intended.

- **README topology defaults are out of sync with `variables.tf`.** The topology table lists `streaming` default `cpx41` and `monitoring` `cpx21`, but the variable defaults are `cpx42` / `cpx22`. Minor, but the README is the doc people will read before an apply.

ryle-ai (commented):
Infra-only PR (Terraform, a GHCR publish workflow, deploy/migration scripting, and the docker entrypoint genesis-guard restructuring) — no protocol/Scala code, so nothing here blocks a merge. The rollback-first entrypoint guard and the dual-layout (`data/snapshot` vs `data/incremental_snapshot`) data probe are consistent between `entrypoint.sh` and `remote-deploy.sh`'s node-0 wiring, and they correctly prevent a migrated volume from falling through to `run-genesis`.

The substantive items are already well covered by the existing review — world-open `*.tfvars` defaults (both `testnet.tfvars` and `nightly.tfvars`), the unguarded Prometheus (9090) `admin_cidrs` rule vs. the guarded siblings, the scrape-firewall/UFW-vs-actually-scraped-ports drift, the stale `allowed_grafana_cidrs` README line, the shared `entrypoint.sh` heap default, the `nightly.tfvars` streaming sizing comment, the `mkdir -p "$STAGING"` that moved inside the `SKIP_NODES` guard (breaks Phase 6 on a cold `--skip-nodes` run), and the leftover dev-branch plumbing in `publish-node-image.yml`. I confirmed those and won't re-litigate them.

One hardening note not covered by the existing review, on the migration transfer key — inline.

deploy/migrate-data.sh
```diff
@@ -0,0 +1,168 @@
...
+  setup-auth)
+    LOCAL_KEY="$ROOT/.migrate-transfer-key"
+    if [ ! -f "$LOCAL_KEY" ]; then
+      log "generating transfer keypair"
+      ssh-keygen -t ed25519 -f "$LOCAL_KEY" -N "" -C "tn->hetzner-migration-transfer" -q
+    fi
+    PUB=$(cat "$LOCAL_KEY.pub")
+    # restrict the key to connections coming from the tn boxes
+    TN_SRC_IPS=$(for t in "${TN[@]}"; do ssh -G "$t" | awk '/^hostname /{print $2}'; done | paste -sd, -)
+    AUTH_LINE="from=\"$TN_SRC_IPS\" $PUB"
```
ryle-ai: The transfer key is generated unencrypted (`-N ""`) and installed with only a `from="<tn IPs>"` restriction, so it grants a full interactive shell on the Hetzner `admin` account (which has `NOPASSWD:ALL` sudo + docker group via `node-init.tpl`). Since the private half lives on the live tn1-3 boxes, a compromise of any tn box during the migration window becomes root on all three new nodes, not just rsync access. Consider a forced-command entry — `command="...",restrict,from="$TN_SRC_IPS"` fronting an `rrsync`-style wrapper (or at least `restrict`) — so the key can only drive the rsync it's for. Non-blocking, but cheap defense-in-depth for a key that outlives the cutover unless someone remembers to delete the `authorized_keys` line.

deploy/terraform/templates/node-init.tpl
```diff
@@ -0,0 +1,122 @@
...
+# firewall scopes the source to admin CIDRs (as on live n3).
+ufw allow 3000/tcp
+ufw allow 9090/tcp
+ufw allow 8123/tcp
+# Public web services (block explorer / faucet) — world-facing (as on live n3).
+ufw allow 3001/tcp
+ufw allow 3010/tcp
+ufw allow 3011/tcp
+%{ endif ~}
+ufw --force enable
```
ryle-ai, line 116: The `streaming` role gets no port block here — only the `node` and `monitoring` arms exist — so on the streaming node UFW ends up with just `22/tcp` open under `default deny incoming`, then `ufw --force enable` locks it in. But `remote-deploy.sh` Phase 6 deploys **block-explorer on the streaming node** (`SS_NODE = ALL_NODES[3]`) with `network_mode: host` + `serverless offline --host 0.0.0.0 --httpPort 3001`. The Hetzner cloud firewall opens 3001 to the world on every server, so the *only* thing blocking the public block-explorer API is this host UFW on the streaming box — it'll be unreachable externally.

The 3001/3010/3011 allows are instead on the `monitoring` arm (lines 116-118), where nothing listens — the base-firewall comment and README call these "block explorer / faucet ... on the monitoring node," but they actually run on the streaming node. Add a `%{ if node_role == "streaming" ~}` block opening 3001 (and drop the ports from monitoring, or move them to match where the services actually run).

Separately, the header comment at the top of this file says "UFW left OFF by default ... see the optional block at the bottom," which contradicts the unconditional `ufw --force enable` here — worth reconciling so the next operator knows UFW is actually on.

ryle-ai (commented):
Infra-only PR — no protocol/Scala code, nothing that blocks a merge. ryle-ai has already covered the substantive items (world-open `*.tfvars` defaults, the unguarded Prometheus 9090 rule, the scrape-firewall/UFW port drift on nodes, the stale `allowed_grafana_cidrs` README line, the shared `entrypoint.sh` heap default, the `mkdir -p "$STAGING"` moved inside the `SKIP_NODES` guard, the `nightly.tfvars` streaming sizing comment, the leftover dev-branch plumbing in `publish-node-image.yml`, and the migration transfer-key hardening). I won't re-litigate those.

One functional item not covered by the existing review — the host UFW on the **streaming** node blocks the block-explorer port it's supposed to serve (inline).

docker/bin/remote-deploy.sh
```diff
@@ -365,105 +511,241 @@ services:
...
+  # Stop SS (and any legacy block-explorer container from older deploys) before touching
+  # the DB or resume state; SS restarts below after schema + seed are settled.
+  ssh "$SS_NODE" "if [ -f $SS_REMOTE_DIR/docker-compose.yaml ]; then cd $SS_REMOTE_DIR && docker compose stop snapshot-streaming 2>/dev/null; fi; docker rm -f block-explorer 2>/dev/null" || true
+
+  # Apply schema. Existing SS data is PRESERVED by default — the destructive reset
+  # (drop schema + force-reset + re-seed from head) only runs when the DB is
+  # empty/uninitialized or explicitly requested via SS_RESET_DB=true.
+  # URL-based psql via a throwaway container: works against both the local postgres
+  # container and an external SS_DB_URL (host networking so 127.0.0.1 resolves to the node).
+  SS_PSQL="docker run --rm --network host postgres:15-alpine psql $SS_DB_URL"
```
ryle-ai: `SS_PSQL="docker run --rm --network host postgres:15-alpine psql $SS_DB_URL"` passes the full connection URL as a positional `psql` argument. `psql` does not scrub `argv`, so when `SS_DB_URL` is an external DB (i.e. `secrets.TESTNET_SS_DB_URL`), the password lands in the process table on the orchestrator (the `ssh` command line), on the SS node (`ps` sees the `docker run ... psql <url>` cmdline), and in CI logs under any `set -x`. This `SS_PSQL` string is reused across all six call sites (row-count, DROP SCHEMA, trim tip/delete, max-ordinal, count), so every one carries it.

Note the asymmetry: the prisma `db push` calls right below already use `-e DATABASE_URL='$SS_DB_URL'` (env, not argv). The URL is also already split into `SS_DB_HOST/SS_DB_PORT/SS_DB_USER/SS_DB_PASS/SS_DB_NAME` above, so the cleaner form is libpq env vars — e.g. `docker run --rm --network host -e PGHOST=... -e PGPORT=... -e PGUSER=... -e PGPASSWORD=... -e PGDATABASE=... postgres:15-alpine psql -t -A -c '...'`. Cheap defense-in-depth for a real secret.

deploy/deploy-cluster.sh
```diff
@@ -0,0 +1,252 @@
...
+  info="$(ssh -o BatchMode=yes -o ConnectTimeout=8 "$SSH_USER@$ip" \
+    'curl -s -m5 http://localhost:9000/node/info; echo; curl -s -m5 http://localhost:9000/global-snapshots/latest/ordinal' 2>/dev/null || true)"
+  state="$(printf '%s' "$info" | python3 -c "import sys,json;print(json.loads(sys.stdin.readline()).get('state'))" 2>/dev/null || echo '?')"
+  ord="$(printf '%s' "$info" | tail -1)"
+  printf '  node %-15s state=%s latest=%s\n' "$ip" "$state" "$ord"
+  [ "$state" = "Ready" ] || ok=0
+done
+SS_VERIFY_DB_URL="${SS_DB_URL:-postgresql://snapshot_streaming:snapshot_streaming@127.0.0.1:5432/snapshot_streaming}"
+ss_count="$(ssh -o BatchMode=yes "$SSH_USER@$STREAM_IP" \
+  "docker run --rm --network host postgres:15-alpine psql $SS_VERIFY_DB_URL -t -A -c 'SELECT COUNT(*) FROM global_snapshots;'" 2>/dev/null || echo '?')"
```
ryle-ai: Same password-in-argv pattern as the SS helper in `remote-deploy.sh`: this verify step runs `psql $SS_VERIFY_DB_URL` positionally over `ssh` to `$STREAM_IP`, so an external `SS_DB_URL` password is exposed in the orchestrator's `ssh` cmdline and in `ps`/`docker run` on the streaming node. Prefer libpq env vars (`-e PGPASSWORD=... -e PGHOST=...`) here too so the two `psql` paths match the env-based form the prisma invocations already use.

ryle-ai (commented):
Infra-only PR (Terraform, GHCR publish workflow, deploy/migration scripting, docker entrypoint restructuring) — no protocol/Scala code, so nothing here blocks a merge. ryle-ai has already covered the substantive items (world-open `*.tfvars` defaults, the unguarded Prometheus 9090 `admin_cidrs` rule, the scrape/UFW-vs-scraped-ports drift, the stale `allowed_grafana_cidrs` README line, the shared `entrypoint.sh` heap default, the `mkdir -p "$STAGING"` moved inside the `SKIP_NODES` guard, the `nightly.tfvars` streaming sizing comment, the leftover dev-branch plumbing in `publish-node-image.yml`, and the migration transfer-key hardening). I won't re-litigate those.

One item not covered by the existing review — the SS Postgres URL is passed as a positional `psql` argument in two places (inline). Non-blocking: cosmetic for the local weak-default DB, but real once `SS_DB_URL` points at an external DB carrying `secrets.TESTNET_SS_DB_URL`.

.github/action_scripts/send_transactions/dag.js
```diff
@@ -0,0 +1,112 @@
...
+
+// One of the nightly genesis-funded private keys (key1 from
+// shared/constants.js -> DAG5sz69nNwGF8ypn1yukFpg2pVJpdx5mnf1PJVc, 100k DAG).
+const SENDER_PRIVATE_KEY =
+  process.env.SENDER_PRIVATE_KEY ||
+  '595a30ab6c62ae48a23414951e2703f49f8c0040b9801738ad3550475389d811';
+
+// The receiver was supplied as a 64-char hex private key; derive its DAG
+// address (a DAG transaction destination must be a DAG-prefixed address).
+const RECEIVER_PRIVATE_KEY =
```
ryle-ai, line 17: `RECEIVER_PRIVATE_KEY` is a full 64-char private key that isn't one of the documented test keys (unlike `SENDER_PRIVATE_KEY`, which is `key1` from `shared/constants.js`) — it appears only here. And it's never actually used as a key: it's fed to `loginPrivateKey` purely to read back `receiver.address` (line 45); nothing in this script signs with it, only the sender signs. So the private half is unnecessary, yet the script then repeatedly funds that address with `AMOUNT_DAG` (default 10000) — anyone reading the repo can sweep whatever it receives. Hardcode the destination **address** (`DAG…`) instead and drop the private key. If you want to keep it overridable, `RECEIVER_ADDRESS` env is enough; there's no reason to hold a spendable key for a receive-only destination. Testnet-only, so low stakes, but it's needless secret exposure of a non-test key.

ryle-ai (commented):
Infra-only PR — the substantive Terraform/deploy-scripting items are already well covered by ryle-ai's review rounds and remain unaddressed (world-open `*.tfvars` defaults, the unguarded Prometheus 9090 `admin_cidrs` rule, the scrape/UFW-vs-scraped-ports drift, the shared `entrypoint.sh` heap default, the `mkdir -p "$STAGING"` moved inside the `SKIP_NODES` guard, the migration transfer-key hardening, and the two password-in-argv `psql` call sites). I won't re-litigate those.

One item not in the earlier rounds — the new `send_transactions/dag.js` helper commits a receiver private key it doesn't need (inline). Non-blocking, testnet-only.

deploy/terraform/variables.tf
```diff
@@ -0,0 +1,93 @@
...
+variable "team_ssh_keys" {
+  description = "List of SSH public keys for team access (add the keys that today reach the AWS testnet nodes)"
+  type        = list(string)
+  default     = []
+}
+
+variable "allowed_ssh_cidrs" {
+  description = "CIDR blocks allowed SSH + CLI (9002/9012) access. SET TO TEAM/VPN CIDRs — do not leave 0.0.0.0/0 on a public testnet."
+  type        = list(string)
+  default     = ["0.0.0.0/0", "::/0"]
```
ryle-ai, line 86: `allowed_ssh_cidrs` (here) and `admin_cidrs` (line 92) still default to `["0.0.0.0/0", "::/0"]`, which contradicts the fail-closed guarantee the rest of the PR now advertises. The earlier world-open finding was addressed by setting the shipped `*.tfvars` to `[]`, and both those files and the README now state "the committed defaults are empty (fail-closed), so a verbatim `apply` grants no SSH or admin access." But Terraform only reaches these variable defaults when a key is absent, and it does so per key — so a bare `apply` with no `-var-file`, a new-env tfvars (the README says to "copy one for the others") that drops either key, or the `*.auto.tfvars` path for `admin_cidrs` if it's forgotten, all silently reopen SSH+CLI (9002/9012) and Grafana/Prometheus/ClickHouse to the world. That's the opposite of the documented posture, and the whole guarantee currently rests on both shipped tfvars keeping the explicit `[]`. Defaulting these two variables to `[]` makes the code match the docs: the tfvars `[]` become redundant-but-harmless, a bare apply is fail-closed, and an incomplete new-env tfvars stays closed. Non-blocking, but this is a security default worth inverting (the variable's own description already warns "do not leave 0.0.0.0/0 on a public testnet").

ryle-ai (commented):
Infra-only PR — nothing here blocks a merge, and the substantive Terraform/deploy items (the 9090 `admin_cidrs` guard, scrape/UFW port drift, the shared `entrypoint.sh` heap default, the password-in-argv `psql` sites, the migrate-key hardening, `mkdir -p "$STAGING"` vs the `SKIP_NODES` guard) are already owned by the existing review rounds and are either resolved in the current diff or acknowledged. One follow-up that the world-open-tfvars fix left behind, inline on `variables.tf`.

ryle-goehausen (dismissed):

deploy/terraform/main.tf
```diff
@@ -0,0 +1,243 @@
+terraform {
+  required_version = ">= 1.0"
+
+  # State lives under a per-environment workspace (Terraform forbids variables in
+  # backend config). Environments are separated by workspaces; rename the bucket below
+  # "tessellation-testnet" if you prefer a dedicated per-stack bucket. State stays on AWS
+  # S3. Select the env with `terraform workspace select <env>`. (Independent of the app's snapshot S3 -> Hetzner Object Storage move).
+  backend "s3" {
+    bucket               = "tessellation-nightly"
+    key                  = "terraform.tfstate"
```
ryle-ai, line 10: This backend block changes the remote-state address relative to the deleted `nightly/terraform/main.tf`, which used `key = "nightly/terraform.tfstate"` with `workspace_key_prefix = "nightly"`. Under the new config the `nightly` workspace resolves to `tessellation-cluster/nightly/terraform.tfstate`, but the live nightly cluster's state sits at `nightly/nightly/terraform.tfstate` (or `nightly/terraform.tfstate` for the default workspace). Because the old backend block is gone, `terraform init` can't offer to migrate — the `nightly` workspace initializes empty, so the next `plan`/`apply` for nightly tries to recreate the running cluster from scratch (or fails on `hcloud_ssh_key`/server name collisions), leaving the existing servers and the ~400 GB state volumes unmanaged. Worth a one-time state migration (S3-copy the old key to the new path, or a `terraform state` move) plus a note in the README's migration section. Non-blocking, and only the `nightly` workspace is affected — the new testnet/integrationnet/mainnet workspaces have no prior state to orphan.

ryle-ai (commented):
Infra-only PR. The substantive Terraform/deploy items are already owned by the existing ryle-ai review rounds (world-open `*.tfvars` defaults, the unguarded Prometheus 9090 `admin_cidrs` rule, the scrape/UFW-vs-scraped-ports drift, the shared `entrypoint.sh` heap default, the two password-in-argv `psql` sites, the migrate-key forced-command hardening, `mkdir -p "$STAGING"` vs the `SKIP_NODES` guard) and are either resolved in the current diff or acknowledged — I won't re-litigate them.

One item the relocation introduced that isn't in the earlier rounds: moving the Terraform from `nightly/terraform/` to `deploy/terraform/` also changed the S3 backend `key` and `workspace_key_prefix`, which orphans the existing `nightly` remote state. Inline on `main.tf`. Non-blocking.

ryle-goehausen (approved):

PR #1520, opened 2026-06-11
fix: byte-exact rollback replay by serving each replayed snapshot's parent

marcus-girolneto:
## Problem

`applyCache` replays data blocks through the metagraph's `combine` using the production `L0NodeContext`. Since #1466 prepends the rollback tip before `loadChain`, `getLastCurrencySnapshot` returns the **tip** for *every* replayed ordinal — but at original consensus time, `combine` for snapshot K saw **K−1**.

Any metagraph that derives state from the last currency snapshot recomputes historical states with the wrong context. When an epoch/day boundary falls inside the replay window, the recomputed calculated state silently diverges from the on-chain `calculatedStateProof`; `applyCache` persists it without proof verification and the rebuilding node seeds peers with it. Found while diagnosing the 2026-06-11 DOR mainnet outage (the v3.5.x crash itself is addressed by #1519; this PR removes the remaining determinism gap, which exists on develop too).

## Fix

- `TraverseLocalFileSystemTempStorage` now stores `Signed[CurrencyIncrementalSnapshot]` (safe: the temp dir is created per `loadChain` run, so no cross-version files can exist).
- `discover` carries the `Signed` base snapshot; `applyCache` threads each ordinal's true predecessor through the fold and wraps the node context (`replayScopedContext`) so `getLastCurrencySnapshot` returns exactly what the original consensus combine saw. `getLastCurrencySnapshotCombined`'s snapshot half is now pinned to the same predecessor, so its ordinal/hash agree with the plain accessor. All other context methods delegate.
- The predecessor is hashed via `HasherSelector.forOrdinal` (not the live context's `alwaysCurrent`/JSON shortcut) so its hash is the true canonical, ordinal-correct one — matching what the rest of the chain expects, not a live-path convenience that's only valid because the tip is always past any hash-migration cutover.
- `applyCache`'s fold now raises `NonContiguousReplayPredecessor` instead of silently threading a stale predecessor if cached ordinals are ever non-contiguous.
- `loadChain()`'s public signature is unchanged; `Rollback` needs no changes.

Result: rollback replay is **byte-exact** for any metagraph whose `combine` derives state only from `getLastCurrencySnapshot` — recomputed states match the historical `calculatedStateProof`s.

## Known limitation (intentionally out of scope)

`getLastCurrencySnapshotCombined`'s `CurrencySnapshotInfo` half still comes from the node context (tip info), not the predecessor's — historical `CurrencySnapshotInfo` isn't reconstructed by the traverse. Metagraphs whose `combine` reads that info (e.g. amm-metagraph, voting-poll via `L0CombinerService.combine`, or the default `getTokenUnlocks`/balance helpers in `dataApplication/package.scala`) do **not** get byte-exact replay. Called out in a comment at the call site. A follow-up could reconstruct historical `CurrencySnapshotInfo` if a use case needs it.

## Verification

- `nodeShared/compile`, `nodeShared/Test/compile`, `shared/compile`, and `currencyL0/compile` clean on develop (JDK 21)
- `nodeShared/test` and `shared/test` pass
- Added `DataApplicationTraverseSuite`: asserts `combine` observes each replayed ordinal's true predecessor (not a fixed tip) across a multi-ordinal replay, and that the new contiguity guard fires on a gapped cache. Confirmed the first assertion fails against the pre-fix behavior (temporarily reverting the predecessor pin turns it red), i.e. it would have caught this bug.

### DIFF
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/dataApplication/DataApplicationTraverse.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/dataApplication/DataApplicationTraverse.scala
@@ -6,6 +6,7 @@ import cats.syntax.all._
 
 import scala.collection.immutable.{SortedMap, SortedSet}
 import scala.concurrent.duration.DurationInt
+import scala.util.control.NoStackTrace
 
 import io.constellationnetwork.currency.dataApplication.DataUpdate.getDataUpdates
 import io.constellationnetwork.currency.dataApplication._
@@ -13,6 +14,7 @@ import io.constellationnetwork.currency.dataApplication.storage._
 import io.constellationnetwork.currency.schema.currency.{CurrencyIncrementalSnapshot, CurrencySnapshotInfo}
 import io.constellationnetwork.cutoff.{LogarithmicOrdinalCutoff, OrdinalCutoff}
 import io.constellationnetwork.domain.seedlist.SeedlistEntry
+import io.constellationnetwork.ext.cats.syntax.next.catsSyntaxNext
 import io.constellationnetwork.json.{JsonBrotliBinarySerializer, JsonSerializer}
 import io.constellationnetwork.kryo.KryoSerializer
 import io.constellationnetwork.node.shared.domain.snapshot.services.GlobalL0Service
@@ -32,9 +34,22 @@ import retry.implicits.retrySyntaxError
 
 trait DataApplicationTraverse[F[_]] {
   def loadChain(): F[Option[(DataState.Base, SnapshotOrdinal)]]
+
+  // Exposed (beyond loadChain's own use) so tests can exercise predecessor-context replay
+  // directly, without needing to fabricate a full global-snapshot discovery chain.
+  def applyCache(
+    storage: TraverseLocalFileSystemTempStorage[F],
+    startingState: DataState.Base,
+    startingSnapshot: Signed[CurrencyIncrementalSnapshot]
+  ): F[(DataState.Base, SnapshotOrdinal)]
 }
 
 object DataApplicationTraverse {
+
+  case class NonContiguousReplayPredecessor(predecessorOrdinal: SnapshotOrdinal, currentOrdinal: SnapshotOrdinal) extends NoStackTrace {
+    override def getMessage: String =
+      s"Expected replay predecessor ordinal=${predecessorOrdinal.next.show} but got currentOrdinal=${currentOrdinal.show}"
+  }
 
   def make[F[_]: Async: KryoSerializer: JsonSerializer: SecurityProvider: HasherSelector](
     lastGlobalSnapshot: Hashed[GlobalIncrementalSnapshot],
@@ -51,6 +66,104 @@ object DataApplicationTraverse {
       val logger: SelfAwareStructuredLogger[F] = Slf4jLogger.getLogger[F]
 
       def cutoffLogic: OrdinalCutoff = LogarithmicOrdinalCutoff.make
+
+      def cutoffPersistedCalculatedStates(ordinal: SnapshotOrdinal) =
+        logger.info(s"Cleaning persisted calculated states using logarithmic cutoff for ordinal=${ordinal.show}") >> {
+          val toKeep = cutoffLogic.cutoff(SnapshotOrdinal.MinValue, ordinal)
+
+          calculatedStateStorage.listStoredOrdinals.flatMap {
+            _.compile.toList
+              .map(_.toSet.diff(toKeep).toList)
+              .flatMap(_.traverse(calculatedStateStorage.delete))
+          }
+        }
+
+      // Serves each replayed snapshot's true predecessor as getLastCurrencySnapshot, so a
+      // metagraph's combine that reads only the plain accessor sees the same last snapshot
+      // it saw at consensus time and its recomputed calculated state matches the original
+      // byte for byte. getLastCurrencySnapshotCombined is NOT fixed by this change: the
+      // traverse doesn't reconstruct historical CurrencySnapshotInfo, so both the snapshot
+      // and the info it returns still come from the live node context (the rollback tip),
+      // constant across every replayed ordinal. We deliberately don't pin just the snapshot
+      // half to the predecessor here - that would make the pair look ordinal-consistent
+      // while info still doesn't correspond to it, which is a worse trap than today's fully
+      // tip-based pair. Metagraphs whose combine reads this accessor (e.g. amm-metagraph,
+      // voting-poll via L0CombinerService.combine, or the default getTokenUnlocks/balances
+      // helpers) are therefore NOT byte-exact under replay - a known, out-of-scope
+      // limitation. Everything else delegates to the node context.
+      // Hashed via forOrdinal (not the live L0NodeContext's alwaysCurrent/JSON shortcut) so the
+      // predecessor's hash is its true canonical, ordinal-correct one - matching what the rest
+      // of the chain (e.g. the next snapshot's lastSnapshotHash) expects. alwaysCurrent is a
+      // live-path shortcut valid only because the tip is always past any real hash-migration
+      // cutover; it doesn't define the canonical hash of an arbitrary historical predecessor.
+      def replayScopedContext(predecessor: Signed[CurrencyIncrementalSnapshot]): F[L0NodeContext[F]] =
+        HasherSelector[F].forOrdinal(predecessor.value.ordinal) { implicit hasher =>
+          predecessor.toHashed.map { hashedPredecessor =>
+            new L0NodeContext[F] {
+              def getLastCurrencySnapshot: F[Option[Hashed[CurrencyIncrementalSnapshot]]] = hashedPredecessor.some.pure[F]
+              def getCurrencySnapshot(ordinal: SnapshotOrdinal): F[Option[Hashed[CurrencyIncrementalSnapshot]]] =
+                context.getCurrencySnapshot(ordinal)
+              def getLastCurrencySnapshotCombined: F[Option[(Hashed[CurrencyIncrementalSnapshot], CurrencySnapshotInfo)]] =
+                context.getLastCurrencySnapshotCombined
+              def getLastSynchronizedGlobalSnapshot: F[Option[GlobalIncrementalSnapshot]] =
+                context.getLastSynchronizedGlobalSnapshot
+              def getLastSynchronizedGlobalSnapshotCombined: F[Option[(GlobalIncrementalSnapshot, GlobalSnapshotInfo)]] =
+                context.getLastSynchronizedGlobalSnapshotCombined
+              def getLastSynchronizedAllowSpends: F[Option[SortedMap[Option[Address], SortedMap[Address, SortedSet[Signed[AllowSpend]]]]]] =
+                context.getLastSynchronizedAllowSpends
+              def getLastSynchronizedTokenLocks: F[Option[SortedMap[Address, SortedSet[Signed[TokenLock]]]]] =
+                context.getLastSynchronizedTokenLocks
+              def securityProvider: SecurityProvider[F] = context.securityProvider
+              def getCurrencyId: F[CurrencyId] = context.getCurrencyId
+              def getMetagraphL0Seedlist: Option[Set[SeedlistEntry]] = context.getMetagraphL0Seedlist
+            }
+          }
+        }
+
+      def applyCache(
+        storage: TraverseLocalFileSystemTempStorage[F],
+        startingState: DataState.Base,
+        startingSnapshot: Signed[CurrencyIncrementalSnapshot]
+      ): F[(DataState.Base, SnapshotOrdinal)] =
+        storage.listStoredOrdinals.flatMap(_.compile.toList).flatMap { ordinals =>
+          logger.info(s"Applying cache built during traversing, size=${ordinals.size.show}") >>
+            ordinals.sorted
+              .foldLeftM((startingState, startingSnapshot)) {
+                case ((state, predecessor), currentOrdinal) =>
+                  Async[F].raiseUnless(currentOrdinal === predecessor.value.ordinal.next)(
+                    NonContiguousReplayPredecessor(predecessor.value.ordinal, currentOrdinal)
+                  ) >>
+                    storage.read(currentOrdinal).flatMap { snapshot =>
+                      if (snapshot.value.dataApplication.isEmpty) {
+                        logger.debug(s"Skipping ordinal=${currentOrdinal.show} with no data application section") >>
+                          (state, snapshot).pure[F]
+                      } else
+                        snapshot.value.dataApplication
+                          .map(_.blocks)
+                          .traverse {
+                            _.traverse { blockBytes =>
+                              dataApplication.deserializeBlock(blockBytes).flatMap(_.liftTo[F])
+                            }
+                          }
+                          .map(_.toList.flatten)
+                          .map(_.flatMap(_.dataTransactions.toList))
+                          .map(getDataUpdates)
+                          .flatMap { dataUpdates =>
+                            replayScopedContext(predecessor).flatMap { replayContext =>
+                              dataApplication.combine(state, dataUpdates)(replayContext)
+                            }
+                          }
+                          .flatTap {
+                            case DataState(_, calculatedState, _) =>
+                              logger.info(s"Persisting calculated state for ordinal=${currentOrdinal.show}") >>
+                                calculatedStateStorage.write(currentOrdinal, calculatedState)(dataApplication.serializeCalculatedState) >>
+                                cutoffPersistedCalculatedStates(currentOrdinal)
+                          }
+                          .map((_, snapshot))
+                    }
+              }
+              .map { case (state, lastSnapshot) => (state, lastSnapshot.value.ordinal) }
+        }
 
       private def getGlobalSnapshotWithRetry(
         ordinal: SnapshotOrdinal,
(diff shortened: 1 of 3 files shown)


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/dataApplication/DataApplicationTraverse.scala
```diff
@@ -175,21 +178,50 @@ object DataApplicationTraverse {
...
+        // delegates to the node context.
+        def replayScopedContext(predecessor: Signed[CurrencyIncrementalSnapshot]): F[L0NodeContext[F]] =
+          HasherSelector[F].forOrdinal(predecessor.value.ordinal) { implicit hasher =>
+            predecessor.toHashed.map { hashedPredecessor =>
+              new L0NodeContext[F] {
+                def getLastCurrencySnapshot: F[Option[Hashed[CurrencyIncrementalSnapshot]]] = hashedPredecessor.some.pure[F]
+                def getCurrencySnapshot(ordinal: SnapshotOrdinal): F[Option[Hashed[CurrencyIncrementalSnapshot]]] =
+                  context.getCurrencySnapshot(ordinal)
+                def getLastCurrencySnapshotCombined: F[Option[(Hashed[CurrencyIncrementalSnapshot], CurrencySnapshotInfo)]] =
+                  context.getLastCurrencySnapshotCombined
```
ryle-ai: `getLastCurrencySnapshotCombined` is delegated to the live node context rather than the replayed predecessor, so it returns the rollback tip (constant across all replayed ordinals) instead of ordinal N-1. Metagraphs whose `combine` reads the combined accessor rather than `getLastCurrencySnapshot` will therefore still recompute divergent state during replay. amm-metagraph (`L0CombinerService.combine` → `context.getLastCurrencySnapshotCombined`, then `lastCurrencySnapshot.ordinal.next` and `buildProcessingContext(lastCurrencySnapshotInfo)`) and voting-poll both take this path, so byte-exactness is not achieved for them. The comment above claims the metagraph "sees the same last snapshot it saw at consensus time," but that only holds for the plain accessor. If the predecessor's `CurrencySnapshotInfo` can be obtained (the snapshot is still in storage before the post-`loadChain` cleanup), this should return the predecessor + its info; otherwise the limitation should be documented.

ryle-ai (changes requested):
The predecessor-threading through `applyCache` and the `Signed[...]` change to the temp storage are sound, and the new `if (dataApplication.isEmpty)` skip correctly mirrors the consensus-time acceptance path (`DataApplicationSnapshotAcceptanceManager.consumeSignedMajorityArtifact` only runs `accept`/`combine` when `artifact.dataApplication` is defined).

However the replay-scoped context is incomplete: it only overrides `getLastCurrencySnapshot`, but delegates `getLastCurrencySnapshotCombined` (and `getCurrencySnapshot`) to the live node context. At consensus time a metagraph's `combine` sees `getLastCurrencySnapshotCombined` returning the true predecessor (ordinal N-1) with its `CurrencySnapshotInfo`, because that is the node head when ordinal N is produced. During replay the node head is fixed at the rollback tip (currency snapshot cleanup happens *after* `loadChain` in `Rollback.scala:145-146`), so `getLastCurrencySnapshotCombined` returns the tip for every replayed ordinal instead of the per-ordinal predecessor.

This matters because the combined accessor is the more commonly used one. amm-metagraph's `L0CombinerService.combine` reads `context.getLastCurrencySnapshotCombined` and derives `currentSnapshotOrdinal = lastCurrencySnapshot.ordinal.next` plus builds its processing context from `lastCurrencySnapshotInfo`; voting-poll's `Utils.getLastMetagraphIncrementalSnapshotInfo` does the same. For those metagraphs the recomputed calculated state will still diverge from consensus — i.e. the exact byte-exactness bug this PR targets is not fixed for them, only for metagraphs that happen to read the plain `getLastCurrencySnapshot`.

I understand the traverse only reconstructs `DataState` and doesn't have the predecessor's `CurrencySnapshotInfo` handy, which is presumably why the combined variant was left delegating. But that means the fix is partial; at minimum the ordinal returned by the two accessors should be consistent. Worth either sourcing the predecessor's info (the node still has the snapshot in storage at this point, so `getCurrencySnapshot(predecessor.ordinal)`-style lookup of the combined value is possible pre-cleanup) or documenting explicitly that byte-exact replay only holds for metagraphs that use `getLastCurrencySnapshot` and not the combined form.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/dataApplication/DataApplicationTraverse.scala
ryle-ai: This delegates to the live node context, which during replay is pinned at the rollback tip (snapshot cleanup runs after `loadChain`), so `getLastCurrencySnapshotCombined` returns the tip for every replayed ordinal rather than the per-ordinal predecessor. Metagraphs whose `combine` uses the combined accessor still diverge: amm-metagraph `L0CombinerService.combine` derives `currentSnapshotOrdinal = lastCurrencySnapshot.ordinal.next` and builds its processing context from `lastCurrencySnapshotInfo`, both taken from here (L0CombinerService.scala:39-43); voting-poll does the same. The plain-accessor fix above doesn't help them, so byte-exact replay isn't achieved for the most common consumer.

Getting this right is harder than the plain accessor since it needs the predecessor's `CurrencySnapshotInfo`, which the traverse doesn't reconstruct. If sourcing it pre-cleanup isn't feasible, at least make the two accessors imply the same ordinal and narrow the header comment's byte-exactness claim to metagraphs using `getLastCurrencySnapshot`.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/dataApplication/DataApplicationTraverse.scala
```diff
@@ -175,21 +178,50 @@ object DataApplicationTraverse {
             }
           }
 
+        // Serves each replayed snapshot's true predecessor as getLastCurrencySnapshot, so a
```
ryle-ai: The comment claims the recomputed calculated state "matches the original byte for byte," but that only holds for metagraphs reading `getLastCurrencySnapshot`. Since `getLastCurrencySnapshotCombined` still delegates to the tip, metagraphs using the combined accessor (amm, voting-poll) don't get byte-exactness. Please scope the comment to the plain accessor.

ryle-ai (dismissed):
The predecessor threading through `applyCache` and the `Signed[...]` change to the temp storage are sound, and the new `if (snapshot.value.dataApplication.isEmpty)` skip correctly mirrors the consensus-time acceptance path (`DataApplicationSnapshotAcceptanceManager.consumeSignedMajorityArtifact` only runs `accept`/`combine` when the artifact's `dataApplication` is defined). The fold's predecessor semantics are correct: when replaying ordinal N the accumulator carries N-1 (including empty-dataApplication snapshots, which is right since those are still the "last" snapshot at consensus time).

The blocker is that `replayScopedContext` only pins `getLastCurrencySnapshot` to the per-ordinal predecessor and delegates `getLastCurrencySnapshotCombined` (and `getCurrencySnapshot`) to the live node context. During replay the node head is fixed at the rollback tip (currency-snapshot cleanup happens *after* `loadChain`), so `getLastCurrencySnapshotCombined` returns the tip for every replayed ordinal instead of the per-ordinal predecessor.

This is the more commonly used accessor. amm-metagraph's `L0CombinerService.combine` (modules/shared_data/.../L0CombinerService.scala:39) reads `context.getLastCurrencySnapshotCombined` and then derives `currentSnapshotOrdinal = lastCurrencySnapshot.ordinal.next` and `buildProcessingContext(lastCurrencySnapshot, lastCurrencySnapshotInfo, ...)`. Under this PR, during replay it will see `tip.ordinal.next` (constant across all replayed ordinals) and build its processing context from the tip's `CurrencySnapshotInfo` — so the recomputed calculated state diverges exactly as before. voting-poll's `Utils.getLastMetagraphIncrementalSnapshotInfo` takes the same path. Byte-exactness is only actually achieved for metagraphs that read the plain `getLastCurrencySnapshot`.

Note the fix here isn't symmetric with the plain accessor: the combined variant needs the predecessor's `CurrencySnapshotInfo`, which the traverse doesn't reconstruct (it only rebuilds `DataState`), and `getCurrencySnapshot(ordinal)` returns the snapshot, not the combined info. So sourcing it is not free. At minimum the two accessors should agree on the ordinal they imply; ideally source the predecessor's combined value from storage pre-cleanup, otherwise narrow the claim. Either way the header comment ("a metagraph's combine sees the same last snapshot it saw at consensus time and the recomputed calculated state matches the original byte for byte") overstates the guarantee — it holds only for the plain accessor and should say so.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/dataApplication/DataApplicationTraverse.scala
marcus-girolneto: Fixed. `getLastCurrencySnapshotCombined`'s snapshot half is now pinned to the same predecessor used by `getLastCurrencySnapshot`, so ordinal/hash agree across both accessors. The `CurrencySnapshotInfo` half still comes from the live node context (tip) — reconstructing historical `CurrencySnapshotInfo` would require re-running the currency snapshot state-derivation pipeline per replayed ordinal, which is out of scope here. Narrowed the header comment and PR description to scope the byte-exactness claim to metagraphs reading the plain accessor only, and called out the remaining gap for amm-metagraph/voting-poll explicitly.

marcus-girolneto: Addressed: pinned the combined accessor's snapshot half to the true predecessor (same object `getLastCurrencySnapshot` returns), so the two accessors now agree on ordinal/hash. Reconstructing the predecessor's `CurrencySnapshotInfo` isn't feasible within this PR's scope — it would need the full currency snapshot state-derivation pipeline replayed per ordinal, not just replaying stored snapshots through `combine`. Left that as a documented, out-of-scope limitation (both in the code comment and PR description) rather than attempting a partial/incorrect reconstruction.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/dataApplication/DataApplicationTraverse.scala
marcus-girolneto: Done — narrowed both the inline comment and the PR description's "Result" line to scope the byte-exactness claim to metagraphs whose `combine` derives state only from `getLastCurrencySnapshot`. Added an explicit note that amm-metagraph/voting-poll (which read `getLastCurrencySnapshotCombined`'s info) are not byte-exact under replay.

ryle-ai (commented):
The round-1 blocker is addressed reasonably: the two accessors now agree on ordinal/hash (combined pins its snapshot half to the true predecessor), and the header comment + PR description now scope the byte-exactness claim to metagraphs reading the plain `getLastCurrencySnapshot` and explicitly call out amm-metagraph / voting-poll as out-of-scope. Documenting the limitation was one of the two acceptable resolutions I offered, so I'm not blocking on the missing historical `CurrencySnapshotInfo` reconstruction — that's a defensible scope call.

One behavioral consequence worth surfacing before merge (non-blocking, but more than "just not byte-exact"): `getLastCurrencySnapshotCombined` now returns an internally inconsistent pair `(predecessor, tipInfo)`. For amm's `L0CombinerService.combine`, `currentSnapshotOrdinal = lastCurrencySnapshot.ordinal.next` now advances per replayed ordinal instead of being pinned to `tip.next`. That's an improvement for ordinal-derived logic, but it also means ordinal-gated branches that never matched under the old all-tip behavior will now activate during replay — e.g. `oneTimeFixesHandler.handleOneTimeFixesOrdinals(oldState, currentSnapshotOrdinal)` (L0CombinerService.scala:48) will hit its `updatePools*` / one-time-fix ordinals during reconstruction, while `buildProcessingContext(lastCurrencySnapshot, lastCurrencySnapshotInfo, ...)` still feeds the tip's info. So for amm, replay now produces a *third* distinct trajectory: neither consensus-exact nor identical to pre-PR replay. Re-running a one-time fix during reconstruction happens to be correct (it fired once at that ordinal at consensus time too), but doing so against tip-derived info rather than the predecessor's info is a genuinely mixed state. The current comment ("NOT byte-exact ... known, out-of-scope limitation") is accurate but undersells that the *shape* of the divergence changed. I'd note in the comment that ordinal-gated branches will now execute during replay for these metagraphs.

Compilation and the temp-storage change look fine: the anon `L0NodeContext` overrides all 10 members, and `TraverseLocalFileSystemTempStorage` uses a fresh `Files.tempDirectory` Resource per run (write/read within the same run are both `Signed[...]`), so the element-type change carries no stale-file backward-compat risk.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/dataApplication/DataApplicationTraverse.scala
```diff
@@ -175,21 +178,59 @@ object DataApplicationTraverse {
...
+            predecessor.toHashed.map { hashedPredecessor =>
+              new L0NodeContext[F] {
+                def getLastCurrencySnapshot: F[Option[Hashed[CurrencyIncrementalSnapshot]]] = hashedPredecessor.some.pure[F]
+                def getCurrencySnapshot(ordinal: SnapshotOrdinal): F[Option[Hashed[CurrencyIncrementalSnapshot]]] =
+                  context.getCurrencySnapshot(ordinal)
+                def getLastCurrencySnapshotCombined: F[Option[(Hashed[CurrencyIncrementalSnapshot], CurrencySnapshotInfo)]] =
+                  // Pins the snapshot half to the true predecessor so its ordinal/hash agree with
+                  // getLastCurrencySnapshot above. The info half is still the tip's, not the
+                  // predecessor's - see the limitation noted in the comment above.
+                  context.getLastCurrencySnapshotCombined.map(_.map { case (_, info) => (hashedPredecessor, info) })
```
ryle-ai: This returns an internally inconsistent pair: the snapshot half is the per-ordinal predecessor (N-1) but the info half is the rollback tip's `CurrencySnapshotInfo`. That's the documented, accepted limitation for the info, but note the ordinal fix has a real behavioral effect on consumers that derive the ordinal from this accessor. For amm, `currentSnapshotOrdinal = lastCurrencySnapshot.ordinal.next` now advances per replayed ordinal (was pinned to `tip.next`), so ordinal-gated branches such as `oneTimeFixesHandler.handleOneTimeFixesOrdinals(...)` will now fire during replay against tip-derived info. Worth extending the comment above to say ordinal-gated logic will re-execute during replay for these metagraphs, not just that they're "not byte-exact."

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/dataApplication/DataApplicationTraverse.scala
marcus-girolneto: Correction to my last reply: on further review I reverted the snapshot-half pin. Pinning only the snapshot to the predecessor while leaving `CurrencySnapshotInfo` tip-based creates a pair that *looks* ordinal-consistent but isn't — and the actual in-tree consumers of this accessor (`getTokenUnlocks`/balance helpers in `package.scala`, `DataApplicationSnapshotAcceptanceManager`) only read the `info` half, discarding the snapshot, so the pin didn't help them anyway while making the mismatch harder to notice. `getLastCurrencySnapshotCombined` now fully delegates to the node context as before (both halves tip-based, unfixed), and the comment/PR description spell out that this accessor — and default helpers reading its info — are not byte-exact under replay, alongside amm-metagraph/voting-poll.

marcus-girolneto: Correction: reverted the snapshot-half pin from my earlier reply. It only papered over ordinal consistency while `CurrencySnapshotInfo` stayed tip-based, and the real in-tree consumers (`getTokenUnlocks`, `DataApplicationSnapshotAcceptanceManager`) read only the info half anyway, so it bought nothing there and made the mismatch less visible to any future caller of the snapshot half. `getLastCurrencySnapshotCombined` now delegates fully to context again (unchanged from before this PR), documented as an out-of-scope limitation covering both amm/voting-poll-style combiners and the default balance/token-unlock helpers. Agreed a real fix needs the predecessor's reconstructed `CurrencySnapshotInfo`, which is out of scope here.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/dataApplication/DataApplicationTraverse.scala
```diff
@@ -175,21 +178,59 @@ object DataApplicationTraverse {
...
+        // limitation. Everything else delegates to the node context.
+        def replayScopedContext(predecessor: Signed[CurrencyIncrementalSnapshot]): F[L0NodeContext[F]] =
+          HasherSelector[F].forOrdinal(predecessor.value.ordinal) { implicit hasher =>
+            predecessor.toHashed.map { hashedPredecessor =>
+              new L0NodeContext[F] {
+                def getLastCurrencySnapshot: F[Option[Hashed[CurrencyIncrementalSnapshot]]] = hashedPredecessor.some.pure[F]
+                def getCurrencySnapshot(ordinal: SnapshotOrdinal): F[Option[Hashed[CurrencyIncrementalSnapshot]]] =
+                  context.getCurrencySnapshot(ordinal)
+                def getLastCurrencySnapshotCombined: F[Option[(Hashed[CurrencyIncrementalSnapshot], CurrencySnapshotInfo)]] =
+                  context.getLastCurrencySnapshotCombined
```
ryle-ai: Worth calling out in the comment above: with full delegation, `getLastCurrencySnapshot` (→ per-ordinal predecessor N-1) and `getLastCurrencySnapshotCombined` (→ rollback tip) now return snapshots at *different ordinals/hashes* during replay. At consensus time both returned the same snapshot (the node head), so this cross-accessor disagreement is new to replay. It's harmless for the known consumers — metagraphs using only the plain accessor are now correct, and amm/voting-poll use only the combined accessor (so they see pre-PR tip behavior). But a metagraph whose `combine` reads *both* accessors would observe two contradictory "last snapshots" under replay that agreed at consensus. Given no such consumer is known this is a defensible scope call; I'd just add a line to the comment noting the two accessors intentionally disagree on ordinal during replay, so a future reader mixing both isn't surprised.

ryle-ai (commented):
The round-1 blocker is resolved. The plain-accessor fix is correct: `applyCache` now threads the per-ordinal predecessor `Signed[CurrencyIncrementalSnapshot]` through the fold and serves it as `getLastCurrencySnapshot`, so a metagraph whose `combine` reads only the plain accessor sees the same last snapshot it saw at consensus time. The empty-`dataApplication` case correctly carries the empty snapshot forward as the new predecessor (it was still the "last" snapshot at consensus), and the `if (snapshot.value.dataApplication.isEmpty)` skip mirrors the consensus acceptance path. The `Signed[...]` element-type change in `TraverseLocalFileSystemTempStorage` is internally consistent (fresh temp dir per run, write/read both `Signed`, `deserializeFallback` reconstructs `Signed(v1.toCurrencyIncrementalSnapshot, proofs)`), and the anon `L0NodeContext` overrides all 10 members.

The decision to fully delegate `getLastCurrencySnapshotCombined` to the live context (both halves = tip) rather than pin the snapshot half is a defensible call, and it's better than the intermediate `(predecessor, tipInfo)` version: amm-metagraph's production `L0CombinerService.combine` reads only the combined accessor, so this leaves its replay behavior identical to pre-PR (still not byte-exact, but no new hybrid trajectory). The comment documents this clearly.

One non-blocking subtlety the comment doesn't mention — see inline. Not blocking merge.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/dataApplication/DataApplicationTraverse.scala
ryle-ai: With full delegation, the two accessors now return snapshots at *different ordinals/hashes* during replay: `getLastCurrencySnapshot` yields the per-ordinal predecessor (N-1) while `getLastCurrencySnapshotCombined` yields the constant rollback tip. At consensus time both returned the same snapshot (the node head), so this cross-accessor disagreement is new to replay.

It's harmless for every known consumer — metagraphs using only the plain accessor are now correct, and amm/voting-poll use only the combined accessor (so they see pre-PR tip behavior). The only case it would surprise is a metagraph whose `combine` reads *both* accessors and expects them to agree, of which none exists today, so this is a defensible scope call. Worth one line in the comment above noting the two accessors intentionally disagree on ordinal during replay, so a future reader mixing both isn't caught out.

ryle-ai (commented):
The core fix is sound. `applyCache` now threads the per-ordinal predecessor `Signed[CurrencyIncrementalSnapshot]` through the fold and serves it as `getLastCurrencySnapshot` via `replayScopedContext`, so a metagraph whose `combine` reads only the plain accessor sees the same last snapshot it saw at consensus time and recomputes byte-identical calculated state. The empty-`dataApplication` case correctly carries the empty snapshot forward as the new predecessor (it was still the "last" snapshot at consensus time), and the `if (snapshot.value.dataApplication.isEmpty)` skip mirrors the consensus acceptance path in `DataApplicationSnapshotAcceptanceManager`. The `Signed[...]` element-type change in `TraverseLocalFileSystemTempStorage` is internally consistent (fresh temp dir per run, write/read both `Signed`, `deserializeFallback` reconstructs `Signed(v1.toCurrencyIncrementalSnapshot, proofs)`), so there's no stale-file backward-compat risk. The anon `L0NodeContext` overrides all 10 trait members.

The decision to fully delegate `getLastCurrencySnapshotCombined` to the live context (both halves = rollback tip) rather than pin only the snapshot half is the right call: amm-metagraph's production `L0CombinerService.combine` (L0CombinerService.scala:39) and voting-poll read *only* the combined accessor, so full delegation leaves their replay behavior identical to pre-PR — still not byte-exact, but no new hybrid `(predecessor, tipInfo)` trajectory that the intermediate pin version would have created. The header comment documents this limitation clearly.

One non-blocking subtlety not captured in the comment (see inline). Not blocking merge.

marcus-girolneto:
Follow-up hardening pass (self-review, beyond the original comments):

- **Hash-era choice documented**: traced why the predecessor is hashed via `HasherSelector.forOrdinal` rather than the live context's `alwaysCurrent`/JSON shortcut — the latter is only valid live because the tip is always past any hash-migration cutover, and would give the wrong canonical hash for an old, pre-cutover predecessor during a deep rollback replay. Documented inline.
- **Contiguity guard added**: `applyCache`'s fold now raises `NonContiguousReplayPredecessor` instead of silently threading a stale predecessor if cached ordinals are ever non-contiguous — this exact class of silent-wrong-context bug is what this PR fixes elsewhere, so the fold shouldn't trust the invariant unchecked.
- **Regression test added** (`DataApplicationTraverseSuite`): exercises `applyCache` directly (bypassing global-snapshot discovery) with a fake tip context and a fake `combine` that records what `getLastCurrencySnapshot` returns per ordinal. Asserts each replayed ordinal saw its true predecessor, not the fixed tip. Verified this test fails against the pre-fix behavior (reverting the predecessor pin makes it red) — confirming it would have caught this bug. Also tests the new contiguity guard. Required exposing `applyCache` on the trait (storage now an explicit param instead of a closure capture) purely for testability; `Rollback`'s only call site (`loadChain()`) is unchanged.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/dataApplication/DataApplicationTraverse.scala
```diff
@@ -49,6 +67,104 @@ object DataApplicationTraverse {
...
+      // cutover; it doesn't define the canonical hash of an arbitrary historical predecessor.
+      def replayScopedContext(predecessor: Signed[CurrencyIncrementalSnapshot]): F[L0NodeContext[F]] =
+        HasherSelector[F].forOrdinal(predecessor.value.ordinal) { implicit hasher =>
+          predecessor.toHashed.map { hashedPredecessor =>
+            new L0NodeContext[F] {
+              def getLastCurrencySnapshot: F[Option[Hashed[CurrencyIncrementalSnapshot]]] = hashedPredecessor.some.pure[F]
+              def getCurrencySnapshot(ordinal: SnapshotOrdinal): F[Option[Hashed[CurrencyIncrementalSnapshot]]] =
+                context.getCurrencySnapshot(ordinal)
+              def getLastCurrencySnapshotCombined: F[Option[(Hashed[CurrencyIncrementalSnapshot], CurrencySnapshotInfo)]] =
+                context.getLastCurrencySnapshotCombined
```
ryle-ai, line 107: With full delegation here, `getLastCurrencySnapshot` (→ per-ordinal predecessor N-1) and `getLastCurrencySnapshotCombined` (→ constant rollback tip) now return snapshots at *different ordinals/hashes* during replay. At consensus time both returned the same snapshot (the node head), so this cross-accessor disagreement is new to replay. It's harmless for every known consumer — plain-accessor metagraphs are now correct, and amm/voting-poll use only the combined accessor (pre-PR tip behavior) — so this is a defensible scope call. Worth one line in the comment above noting the two accessors intentionally disagree on ordinal during replay, so a future metagraph author mixing both isn't caught out.

ryle-ai (commented):
The core fix is sound. `applyCache` now threads each replayed ordinal's true predecessor `Signed[CurrencyIncrementalSnapshot]` through the fold and serves it as `getLastCurrencySnapshot` via `replayScopedContext`, so a metagraph whose `combine` reads only the plain accessor recomputes byte-identical calculated state. The empty-`dataApplication` case correctly carries the empty snapshot forward as the next predecessor (it was still the "last" snapshot at consensus), the new `NonContiguousReplayPredecessor` guard enforces the predecessor invariant, and the `Signed[...]` element-type change in `TraverseLocalFileSystemTempStorage` is internally consistent (fresh temp dir per run, JSON write/read both `Signed`, Kryo fallback effectively unreachable within a run) so it carries no stale-file risk.

The decision to fully delegate `getLastCurrencySnapshotCombined` to the live context rather than pin only its snapshot half is the right call — amm-metagraph's `L0CombinerService.combine` and voting-poll read only the combined accessor, so full delegation leaves their replay behavior identical to pre-PR (still not byte-exact, but no new hybrid `(predecessor, tipInfo)` trajectory), and the header comment documents that limitation clearly.

One non-blocking documentation subtlety, see inline. Not blocking merge.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/dataApplication/DataApplicationTraverse.scala
```diff
@@ -29,10 +34,23 @@ import retry.implicits.retrySyntaxError
...
+    startingState: DataState.Base,
+    startingSnapshot: Signed[CurrencyIncrementalSnapshot]
+  ): F[(DataState.Base, SnapshotOrdinal)]
 }
 
 object DataApplicationTraverse {
 
+  case class NonContiguousReplayPredecessor(predecessorOrdinal: SnapshotOrdinal, currentOrdinal: SnapshotOrdinal) extends NoStackTrace {
+    override def getMessage: String =
+      s"Expected replay predecessor ordinal=${predecessorOrdinal.next.show} but got currentOrdinal=${currentOrdinal.show}"
```
ryle-ai, line 51: The message labels `predecessorOrdinal.next` as "replay predecessor ordinal", but that value is actually the *expected current* ordinal, not the predecessor's. As written it reads e.g. "Expected replay predecessor ordinal=12 but got currentOrdinal=13" when the predecessor was 11, which is confusing for anyone debugging a non-contiguity error. Consider something like `s"Expected currentOrdinal=${predecessorOrdinal.next.show} (predecessor=${predecessorOrdinal.show}) but got currentOrdinal=${currentOrdinal.show}"`. Non-blocking.

ryle-ai (commented):
The predecessor-context fix and its scoping now read cleanly. The original blocker (plain `getLastCurrencySnapshot` returning the rollback tip for every replayed ordinal) is resolved, and the residual I raised in earlier rounds — that `getLastCurrencySnapshotCombined` still delegates to the tip while the plain accessor now returns the per-ordinal predecessor, so the two disagree during replay — is now explicitly documented in the header comment (including the deliberate choice not to pin just the snapshot half, and the named out-of-scope consumers: amm-metagraph / voting-poll via `L0CombinerService.combine`). Confirmed amm's `L0CombinerService.combine` reads only the combined accessor (`getLastCurrencySnapshotCombined`, L0CombinerService.scala:39), so full tip-delegation leaves its replay behavior identical to pre-PR — no regression, just an honestly-scoped limitation.

The new contiguity guard and the direct `applyCache` regression test are good additions — the test correctly asserts each ordinal sees its true predecessor rather than the fixed tip. One minor nit on the guard's error message (inline).

scasplte2 (approved):

PR #1485, opened 2026-04-02
fix: fork recovery followup, stale registrations escalation and deferral skip

scasplte2:
## Follow-up fixes to fork recovery (PR #1476)

This PR addresses issues discovered during testnet deployment of fork recovery, plus a new HTTP QoS layer to protect consensus from public API traffic.

### Fork Recovery Fixes

**1. Stale peer registration escalation** (`286b18982`)
When a node is isolated, peers stop sending declarations so peer registrations go stale. The recovery guard saw `peersAtHigherKey=0` and suppressed escalation indefinitely. Fix: also escalate when all ready peer registrations are below the current key AND `totalReadyPeers > 0`.

**2. Skip TimeTrigger deferral on advanced cluster** (`359598c4e`)
After download + observe, `initFromDownload` fetches the consensus outcome from peers. If the cluster advanced past the downloaded ordinal, the node rejected the newer outcome, exhausted retries, and looped back into `DOWNLOAD_INIT_DEFERRED` — stuck while the cluster kept moving. Fix: accept the newer outcome with `isRecovery=true` (skip 43s deferral).

**3. Active facilitator count for kill-4 guard** (`df229737f`)
Replaced peer-registration-based kill-4 guard with `activeFacilitators` from the `QuorumInfeasible` abandon reason. Peer registrations go stale during isolation, making the old guard unreliable.

**4. Observe deadline** (`bd2b9abce`)
Added 5-minute timeout on `observeWithLimit` during recovery. Prevents nodes from getting permanently stuck in Observing state when `fetchNextSnapshot` retries forever on `CannotFetchSnapshot`.

**5. Reraise recovery download error** (`b5e9028a6`)
`recoveryDownload.handleErrorWith` was swallowing errors (returning success), so `DownloadDaemon.clearRecoveryDownload` fired prematurely. Next retry did full download instead of incremental recovery. Fix: reraise the error, matching the non-recovery `download()` error handler pattern.

**6. Candidate observation deferral** (`46753dc4a`)
New candidates (peers in `lastOutcome.finished.candidates` but NOT in `previousEligible`) are added to `allEligible` for persistence but excluded from `eligibleThisRound`. This prevents the phantom candidate problem: an unresponsive candidate being repeatedly included → evicted → re-included every retry at the same key, causing perpetual stalls.

### HTTP QoS

**7. Concurrency controls for public snapshot routes** (`0c00b48b8`)
Two-layer defense against external traffic starving consensus gossip:
[...]

PR #1523, opened 2026-06-17
feat: integrate v4.1.0 consensus and MPT line from release/testnet (alpha.158) into develop

scasplte2:
## Summary

This PR reconciles the **v4.1.0** line - developed and stabilized directly on `release/testnet` (currently `v4.1.0-alpha.158`) - back into `develop`. It is a single `--no-ff` merge of `release/testnet` (**256 commits** since the `v4.1.0-alpha.0` fork at `dd6e83a19`), with the 8 conflicting files resolved against develop's 21 parallel commits. Full build + all test sources compile; the new `ConfigLoadSuite` passes on the merged tree.

**Version context:** `v4.0.0` (currently `v4.0.0-rc.10` on `release/integrationnet`) is the release an MPT state-root divergence broke. `v4.1.0` is the follow-on line that fixes it and has been hardening on testnet across 158 alpha iterations. This PR brings that line to `develop` so develop is again the source of truth ahead of the v4.1.0 launch.

> Supersedes #1485 - see [Relationship to prior work](#relationship-to-prior-work).

## How to review a 256-commit merge

- `develop` diverged from this line at `v4.1.0-alpha.0`; everything since happened on testnet (live-network stabilization), so it lands as one deliberate reconciliation rather than incremental PRs. This is the documented recovery path for a drifted release branch - see `docs/release/BRANCHING_AND_RELEASE_WORKFLOW.md` ("Realigning a Drifted Branch").
- `git log --first-parent develop` shows a single merge commit; `git log` shows the full 256.
- **Merge with "Create a merge commit", NOT squash.** Squashing flattens the 256-commit history and breaks the superset relationship needed for the post-merge `release/testnet` realign.

## What's in it

Heavily fix-weighted (142 fix / 48 feat / 12 refactor / 9 test) - the consensus-stabilization body of work. Major themes:

**Consensus engine stabilization** (alpha.0 -> alpha.158)
- Escalating quorum-denominator shrink for wedged committees; bounded probation re-entry lane for committee recovery; finalize at Core completeness rather than full committee; grace timers from Core completion.
- Facilitator/committee selection: evidence-ladder exclusion of chronic non-signers, admission-certificate capping at proposal creation, recent-signer pool lookback for Tier-1 stickiness.
- Abandonment/recovery: isolated-node escalation, cascade-recovery deadlock fixes, the wedge-health signal feeding `Cluster.leave()`.

**MPT state-root divergence - the root of the v4.0.0 break**
- Prevent-and-detect MPT state-root divergence (`fa0f3616d`); content-aware `syncFullIfNeeded` guard with direct tests (M1a); ordinal-gated GSI dust sweep for state deflation.

**Replay-safety ordinal gating** (`FieldsAddedOrdinals`)
- Ordinal-gate the state-channel fee-balance source (deterministic `accept()` context vs the pre-fix `mptStore`) for byte-identical replay; per-env activation ordinals pinned to real cutovers (e.g. testnet `sc-fee-balance-from-context = 3101393`, the exact v4.0.0->alpha.0 boundary); fail-closed defaults. See `docs/operations/fields-added-ordinals.md` and `docs/release/v4-launch-runbook.md`.

**Reward model cleanup**
- Delegated rewards follow committee membership (Core + Tier-1), not signers; removed the inert reward-rotation mechanism (~900 lines).

**Config hardening + a regression guard**
- Per-env consensus knobs made explicit across all networks; the `quorumShrinkActivationViews` `PosInt`->`Int` fix that unblocked the alpha.158 startup (a `0` = disabled value the refined type rejected); a new `ConfigLoadSuite` that parses the packaged config exactly the way the node does at startup (two-level `SharedConfigReader` + `AppConfigReader`), so a refined-type/value mismatch becomes a red test instead of a deploy-time `ConfigReaderException`.

**Logging hygiene + docs**
- Cut app.log cry-wolf and per-round/per-ordinal INFO floods; mainnet logback to INFO; aligned `docs/consensus/*` with the v4.x consensus + MPT rewrite.

## Conflict resolutions (8 files)

| File | Resolved to | Rationale |
|---|---|---|
| `AbandonmentTracker.scala` | release/testnet | develop's `8027c0642` (escalate isolated node) has a twin on testnet (`6e05edc89`) that testnet evolved further into the `quorumPair` / `peerCurrentKeys` / `networkAdvanced` + wedge-health form. Kept the evolved deployed version. **Review focus below.** |
| `ClickHouseSink.scala` | release/testnet | Its `jdbcUrl` val already carries develop's configurable-protocol fix (`config.protocol`) and the connect-timeout defenses |
| `snapshot-streaming.patch` | release/testnet | v19-schema-compatible; matches the node schema being integrated |
| `docker/monitoring/*` (4 files), `token-lock-replacement-edge-cases.js` | develop | develop's actively-maintained nightly-cluster infra + CI test |

## Relationship to prior work

- **Supersedes #1485** (`fix/fork-recovery-followup`). That branch and `release/testnet` are divergent siblings from `v4.1.0-alpha.0`; `release/testnet` carries the more-evolved twins of #1485's consensus fixes (the isolated-node escalation above is one example). Merging both would double-apply. **Recommend closing #1485 in favor of this PR** (pending confirmation).
- Stabilized across `v4.1.0-alpha.0 ... v4.1.0-alpha.158` on testnet; reviewed in two Codex passes (fail-closed gate default, doc debt, the testnet ordinal pin).

## Docs

- `docs/release/BRANCHING_AND_RELEASE_WORKFLOW.md` - the branch model + drifted-branch recovery this PR follows.
- `docs/release/v4-launch-runbook.md` - coordinated cold-restart procedure + the gate-ordinal checklist for deploy.
- `docs/operations/fields-added-ordinals.md` - the ordinal-gating surface and per-env values.
- `docs/consensus/*` - consensus mechanism reference (updated in this PR).

## Testing

- Full build compiles; all test sources compile; `ConfigLoadSuite` passes (config parses two-level as the node loads it at startup).
- CI e2e runs on this PR. The checkout is clean because the branch already contains develop, so there is no conflicted merge at CI time.

## Review focus

1. **`AbandonmentTracker.scala`** - confirm release/testnet's `networkAdvanced` (peers ahead via `peerCurrentKeys`) covers the isolated/cut-off escalation that develop's `8027c0642` stale-registrations check targeted.
2. **`token-lock-replacement-edge-cases.js`** - took develop's version; if merged token-lock behavior differs, watch this in the e2e.

## After merge

Realign `release/testnet` to develop (fast-forward) so future deploys follow the normal `develop -> release/testnet` direction (`BRANCHING_AND_RELEASE_WORKFLOW.md`, recovery section).

---
Merge commit: `058edae23`. Base `develop` @ `9bbbe358e`.


marcus-girolneto (dismissed):
# PR #1523 — Critical Review (v4.1.0 consensus + MPT into develop)

`feat/v4.1.0 → develop` · 264 commits · 333 files · +44,794 / −4,863 · base==merge-base (linear diff)

## Executive verdict: **BLOCK — do not merge**

Three CRITICAL defects, two of which (the `condModifyState` lost-update race and the cross-version `GlobalConsensusOutcome` decode break) hit **mainnet with no Byzantine actor required**, plus several HIGH determinism hazards on the view-change/timeout and fork-recovery paths that can fire under normal operation. The replay/MPT/config-gating machinery and the reward model are sound; the blockers are concentrated in the new consensus engine state-management and recovery code.

Coverage note: 9/12 review dimensions completed with cross-corroboration; tokenlock, reward, and config/replay dimensions were completed by direct manual verification after the automated verifier pass was cut short by API rate limits (see Confidence).

---

## Top risks (ranked)

### 1. BLOCKER — `condModifyState` lost-update race (verified)
- **Files:** `ConsensusStorage.scala:~422-430` (def), `engine/AbandonmentTracker.scala:333`, `engine/StallDetector.scala:546`, `engine/ConsensusRoundRunner.scala:264`
- **Why:** develop guarded per-key state mutation with `stateUpdateSemaphore` + `statesR(key).access` CAS. This PR removed both → non-atomic `get → modifyStateFn → set`. The justifying comment ("all writers arrive through the FSM command loop") is false: `StallDetector.monitor` runs on a supervised background fiber (`spawnTracked` → `supervisor.supervise`) and calls `abandonRound → condModifyState(key)` directly, concurrently with the command-loop fiber's `condModifyState`.
- **Failure scenario:** command loop reads state, runs slow signature verification, computes an advanced/finalized state; meanwhile the monitor fiber's `abandonRound` reads the same state and sets `None`; the command loop then writes its stale advanced state → a round abandoned *after* reaching quorum is resurrected, or an abandon is silently lost (zombie round). Host-timing-dependent → different accepted state across nodes. The `roundAttemptId` guard only protects the queued `RoundCompleted` command, not this direct path.
- **Fix:** restore mutual exclusion without the stall regression — route `abandonRound` through the FSM command queue, or use a per-key lock / `Ref.access`-CAS retry that keeps heavy work *outside* the critical section (the original stall came from doing verification inside the semaphore, not from the lock itself).

### 2. BLOCKER — VCC/TC quorum counts votes, not distinct signers
- **Files:** `state/ProposalVccValidator.scala:100-101` (used at `:131`, `:184`); affects `dag-l0 GlobalSnapshotConsensusStateAdvancer` and `currency-l0 CurrencySnapshotConsensusStateAdvancer`
- **Why:** `certQuorumMet(voterIds: List[PeerId])` decides `voterIds.size >= q` over a non-deduplicated `List`. `ViewChangeVote.highestKnownQc` is part of the vote's identity and is **not** constrained equal across votes, so one peer can sign two distinct votes; both survive as distinct `NonEmptySet` elements and both are counted. The cert *builder* and local *assembler* dedup by signer; only the follower-side *validator* (the gate for VCCs embedded in a remote proposal) omits it.
- **Failure scenario:** `ceil(q/2)` colluding/buggy facilitators each contribute two differing-QC votes → every honest follower accepts a view/leader change with fewer than `q` distinct signers. Illegitimate leader hand-off / equivocation double-count.
- **Fix:** `voterIds.toSet.size` (and `.toSet.count` on the anchor path), or delegate to `QuorumDenominatorShrink.Decision.meets`, which already takes a `Set[PeerId]`.

### 3. BLOCKER — Currency creator seeds round-start view/leader from wall clock
- **Files:** `currency-l0 CurrencySnapshotConsensusStateCreator.scala:546,550,598-604` (consumed by the currency advancer; corrupts persisted `peerViewChanges`)
- **Why:** `initialView = math.max(priorAbandonmentCount, ViewFromTime.compute(now,…))` from `Clock.realTime` + a node-local retry counter, fed into `selectLeaderWeighted(viewNumber)` and `initialViewNumber`. dag-l0 was **deliberately fixed** to `initialView = 0` (`GlobalSnapshotConsensusStateCreator.scala:660`); the currency path kept the pre-fix behavior while copying the new comment.
- **Failure scenario:** two honest metagraph (currency-L0) nodes compute different `timeView` → different leader → competing proposals; a `view=N>0` seed emits a no-VCC proposal that peers seeded at `0` reject (`view{N}_proposal_missing_view_cert`). Deterministic-divergence liveness wedge after the first finalized round under any stall/retry; also makes the agreed `peerViewChanges` outcome field node-dependent.
- **Fix:** set `initialView = 0` (mirror dag-l0); never feed `ViewFromTime`/retry counters into round-start view/leader.

### 4. HIGH — View-change/timeout vote storage is last-write-wins keyed by relay origin; ViewChangeVote ingestion lacks `origin===signer`
- **Files:** `ConsensusStorage.scala:522` (addViewChangeVote), `:537` (addTimeoutVote) vs hardened `:633-635`/`:672-674`; `state/RumorHandler.scala:159-163`
- **Why:** eviction/admission were explicitly hardened to first-write-wins + `origin===signer` ("prevents a late retransmit/replay from overwriting the original signed commitment"); the VC/timeout paths were not. A relay controls the map key and which `Signed[*Vote]` survives.
- **Failure scenario:** fires **without a Byzantine actor** — `StallDetector` calls `performViewChange` every tick, re-reading `highestKnownQc` from the (mutable) vote lock, so one honest node naturally emits multiple distinct self-votes for the same transition. Combined with #5, two honest nodes lock different QCs / make opposite view-transition decisions on identical inputs.
- **Fix:** first-write-wins for VC/timeout, key by verified signer, add the `origin===signer` gate to `handleViewChangeVote`.

### 5. HIGH — Cert builders dedup with non-deterministic `.head` over non-byte-identical groups
- **Files:** `engine/ViewChangeCertificateBuilder.scala:67-71` (divergent check `:82-84`), `engine/TimeoutCertificateBuilder.scala:52` (`:66`)
- **Why:** `groupBy(_.proofs.head.id).mapValues(_.head)` with an inline comment asserting group entries are byte-identical — false (`highestKnownQc` isn't a match-filtered field). `.head` over a `groupBy` list is order-dependent; the `divergentQcs` rejection runs *after* this lossy pick, so node A may build a cert and advance the view while node B rejects it. View split / asymmetric assembly.
- **Fix:** make per-signer selection deterministic (e.g. pick by a total `Order` on the signed bytes / highest QC view), and run the divergent-QC check over all of a signer's votes.

### 6. HIGH — Stage-2 quorum-denominator shrink lets a 1/3 minority finalize VCC/TC
- **Files:** `state/QuorumDenominatorShrink.scala:120-127` (`MinQuorumFloor=2`, independent of anchor size); consumed `StateTransitions.scala:101`, `ProposalVccValidator.scala:100-101`
- **Why:** stage 2 drives `requiredQuorum` to 2; two disjoint anchor pairs can each certify conflicting view-change/timeout certs at the same key (partition fork). Deliberate liveness-over-safety trade, gated off on mainnet/integrationnet/dev (`quorumShrinkActivationViews=0`) and **live on testnet (10)**; the value is in `deterministicConfigHash`. No guard prevents two disjoint shrunk quorums.
- **Fix:** floor `requiredQuorum` at a value that preserves quorum intersection (e.g. `> anchorSize/2` AND no two disjoint sets can both meet it), or keep stage 2 mainnet-only-disabled with an explicit invariant test.

### 7. HIGH — Stall-detector eviction cap computed against the *shrunk* quorum
- **Files:** `engine/StallDetector.scala:789-802` (minQuorum = shrunk `coreRequired`), `:1577` (`cap = max(0, committee.size − minQuorum)`)
- **Why:** when the shrink rung is active, the eviction cap is derived from the lowered quorum, so nodes can vote the committee down to the hard floor — violating the cap's own safety invariant (derived assuming `minQuorum = supermajority`). Testnet-live; bounded today only because eviction-cert *assembly* still uses full `q`.
- **Fix:** derive the eviction cap from the unshrunk supermajority, never the override.

### 8. HIGH — Fork-recovery majority-peer hint is dead code → minority-fork re-adoption
- **Files:** `gossip/event/RecoveryPeerHint.scala:16,25` (set by `ForkRecoveryService.onForkDetected`, never read), `dag-l0 Download.scala:473` (`peerSelect.selectForRecovery` uses random weighted sampling)
- **Why:** the detector computes the majority/canonical chain's peers but recovery ignores the hint and samples; a node that correctly detected it is on a minority fork can re-adopt the minority fork. Defeats the subsystem's purpose.
- **Fix:** consume `getAndClearPreferredPeers` in `selectForRecovery` (prefer/intersect with the majority hint).

### 9. HIGH — Recovery rolls back to a single sampled peer's tip with no quorum/majority gate
- **Files:** `dag-l0 Download.scala:485,495-496,512`
- **Why:** `recoveryStart` destructively `cleanupAbove(metadata.ordinal)` / `deleteAbove(...)` based on one sampled peer's reported tip, then downloads from it — no `>2/3` verification that the tip is canonical. A minority/stale source makes the node discard potentially-canonical snapshots.
- **Fix:** require a majority of reporters to agree on the rollback target ordinal/hash before destructive cleanup.

### 10. HIGH — `GlobalConsensusOutcome` adds 11 non-Option fields the derevo decoder won't default
- **Files:** `dag-l0 .../snapshot/schema.scala:108-217` (fields at 117,123,132,141,150,162,179,197,201,207,216)
- **Why:** derevo/circe-magnolia uses `useDefaults=false`, so a missing JSON key throws `DecodingFailure` regardless of the Scala `= SortedMap.empty` defaults (only `Option` fields tolerate absence — the authors correctly Option-wrapped 2 of 13). `GlobalConsensusOutcome` is wire-decoded cross-node (`getConsensusOutcome`), so a v4.1.0 node fetching an outcome that omits these keys (develop node during rolling upgrade) fails — defeating the "rolls over cleanly" intent.
- **Fix:** `Option`-wrap the 11 new fields (decode to empty), or enable `useDefaults` for this type.

### 11. HIGH — Unbounded growth of 7 new per-round `MapRef`s (eventual OOM)
- **Files:** `ConsensusStorage.scala:373-381` (decls), `:963-979` (`pruneStaleResources` prunes only `resourcesR`), `:460-462`
- **Why:** `voteLocksR`, `assembledVccR`, `assembledVccApplyScheduledR`, `assembledVccReceiptsR`, `timeoutCertificateApplyScheduledR`, `assembledEvictionCertsR`, `assembledAdmissionCertsR` accumulate one entry per ordinal on the normal success path; `clearVoteLock` has **zero** non-test callers and `timeoutCertificateApplyScheduledR` is never cleared. Memory growth → OOM/GC-stall over days (liveness regression).
- **Fix:** clear all per-key maps in `cleanupStateAndResource` (or extend `pruneStaleResources`).

### MEDIUM (selected)
- **Cross-version finalization** — new *signed* `peerHistory` bytes mean develop↔v4.1.0 nodes compute different artifact hashes; fenced by `consensusSchemaVersion` in `deterministicConfigHash`, so it surfaces as exclusion, but **mandates a coordinated, non-rolling cutover**. (`schema.scala:313`, advancer `:3382-3383`) — deploy-runbook item.
- **Shrunk-cert acceptance is local-wall-clock-dependent** — `decide` reads `Clock.realTime` (not monotonic); an NTP backstep can un-accept a previously-acceptable cert. Testnet liveness flake, mainnet-inert. (`ConsensusStateAdvancer.scala:96-129`)
- **`syncFullIfNeeded` calls `build()` outside `mutationLock`** — `buildForOrdinal` mutates producer state (clears pending buffers); concurrent locked `update`/`sync` can interleave → torn MPT root / wrong skip decision. (`MptStore.scala:~276`)
- **Recovery breadth issues** — `RecoveryFallbackEligible` mixed into generic `InvalidChain` flips full-download nodes into recovery on a transient bad peer; `SelectablePeerDiscoveryDelay` falls back to 1 peer and counts `WaitingForReady` peers; 2-node equal-ordinal fork never detected. (`Download.scala:1259/1071`, `SelectablePeerDiscoveryDelay.scala:48-64`, `ForkRecoveryDetector`)
- **Quorum-shrink stage-1** is majority-safe, not BFT-intersection-safe vs an equivocating anchor member (testnet). (`QuorumDenominatorShrink.scala:63-65`)

### LOW
- `viewInterval` unrefined → `ArithmeticException` div-by-zero if set to `0s` (`ViewFromTime.scala:41`; `config/types.scala:295`).
- `AdmissionCertificate.ordering` omits `votes` → non-total order (benign today; latent hash-divergence if duplicate targets ever allowed). (`declaration.scala:408-411`)
- `selectLeaderWeighted`/`orderByScore` lack a final `PeerId` tiebreak (latent; SHA-256 score tie). (`FacilitatorSelector.scala:75-76,249-260`)
- `KeyDecoder[SnapshotOrdinal]` fails open to `MinValue` on malformed keys (`SnapshotOrdinal.scala:64-66`); Kryo deserialize swallows `LinkageError` (`KryoSerializer.scala:53-83`); **Kryo ID 535 reused** StallReport→EvictionVote — mixed-build wire hazard *during the testnet→v4.1.0 rollout window only* (`package.scala:55-56`).
- `ConsensusEndTime` median-index comment wrong (cosmetic, deterministic). (`ConsensusEndTime.scala:54`)

---

## Investigated — NOT issues (don't re-raise)
- **Token-lock double-mint:** this PR does **not** introduce one. `TokenLockBlockAcceptanceManager` only adds logging; `TokenLockValidator` tightens the limit check (`=== → >=`) and correctly exempts replacements. The "one lock → two nodeIds" concern lives in the delegated-stake *acceptance* validator, which this PR does not touch (pre-existing; needs a separate PoC).
- **Reward model:** the `Map → SortedMap[PeerId, SortedMap[Address, Amount]]` conversions, `.sortBy(nodeId)`, and effective-`tokenLockRef` matching are **determinism/correctness fixes** (delegateRewards is in the signed artifact; PeerId has no KeyEncoder). No regression.
- **Replay gating (`FieldsAddedOrdinals`):** fail-closed and deterministic. `scFeeBalanceFromContext.getOrElse(env, MaxValue)` + `if (ordinal >= gate)` re-derives old history byte-identically; the dust sweep is exact-ordinal-keyed, "absent env never sweeps," pure over sorted maps, multi-gated.
- **`QuorumPolicy.supermajority = (2n+2)/3`** is exactly `ceil(2n/3)`, byte-identical to develop's prior `ceil(n*0.6667)` for all n. No threshold regression.
- **MptStore Semaphore** wrapping of sync/update/savepoint/restore is correct (except #MEDIUM `syncFullIfNeeded`). Determinism suites verify back-compat byte-identity.

---

## Required reviewer focus (manual human inspection)
1. `ConsensusStorage.condModifyState` + every caller fiber (StallDetector/AbandonmentTracker vs FSM command loop) — confirm the concurrency model before merge.
2. `ProposalVccValidator.certQuorumMet` — distinct-signer counting on both global & currency advancers.
3. `CurrencySnapshotConsensusStateCreator` round-start view/leader seeding vs the dag-l0 fix.
4. `ConsensusStorage.addViewChangeVote/addTimeoutVote` + `RumorHandler` ingestion symmetry with eviction/admission.
5. `ViewChange/TimeoutCertificateBuilder` `.head` dedup determinism.
6. `dag-l0 schema.scala` `GlobalConsensusOutcome` field optionality (cross-version decode).
7. Fork recovery: `RecoveryPeerHint` consumption + `Download.recoveryStart` rollback-target verification.

---

## Inline comment candidates

**`ConsensusStorage.scala` (condModifyState):**
> `condModifyState` is now a non-atomic get→modify→set. The comment says all writers go through the FSM command loop, but `StallDetector.monitor` runs on a supervised fiber (`ConsensusRoundRunner:264`) and calls `AbandonmentTracker.abandonRound` → `condModifyState(key)` directly (`AbandonmentTracker:333`), racing the command-loop `condModifyState`. This is a lost-update on consensus state (round can be abandoned after quorum, or an abandon dropped). The `roundAttemptId` guard only covers the queued `RoundCompleted` path. Please restore mutual exclusion — but keep the heavy verification work *outside* the critical section so the original stall doesn't return.

**`ProposalVccValidator.scala:100`:**
> `voterIds.size >= q` counts votes, not distinct signers — `ViewChangeVote.highestKnownQc` lets one peer contribute multiple distinct votes (builder & assembler dedup by signer; this validator doesn't). Use `voterIds.toSet.size` / `.toSet.count(anchor.contains)` or delegate to `QuorumDenominatorShrink.Decision.meets(Set[PeerId])`.

**`CurrencySnapshotConsensusStateCreator.scala:546`:**
> `initialView = max(priorAbandonmentCount, timeView)` reintroduces wall-clock/retry-counter seeding that dag-l0 deliberately removed (`GlobalSnapshotConsensusStateCreator:660` uses `initialView = 0`). Two honest nodes will pick different leaders and reject each other's seed-view proposals. Set `initialView = 0`.

**`ConsensusStorage.scala:522` (addViewChangeVote):**
> Last-write-wins keyed by relay `origin`, unlike `addEvictionVote`/`addAdmissionVote` (first-write-wins) and without the `origin===signer` gate that `handleTimeoutVote` has. A relay/late retransmit can change which signed vote counts → non-deterministic VCC assembly. Mirror the eviction/admission hardening.

**`dag-l0 schema.scala` (GlobalConsensusOutcome):**
> 11 new non-`Option` fields — derevo decodes with `useDefaults=false`, so a missing key throws `DecodingFailure` (the `= SortedMap.empty` defaults are ignored). A develop node's outcome JSON during rolling upgrade will fail to decode. `Option`-wrap them like `controllerEvidence`/`penaltyUntil`.

---

## Missing tests (by subsystem)

**Consensus state / storage**
- `ConsensusStorageRaceSuite`: concurrently run `abandonRound(key)` (monitor fiber) and a state-advancing `condModifyState(key)` (command loop); assert no finalized round is clobbered to `None` and no abandon is lost.
- `ConsensusStorageCleanupSuite`: a normal successful round clears **all 7** new per-key MapRefs (currently `clearVoteLock` is uncalled).

**View change / timeout**
- `ProposalVccValidatorDedupSuite`: a VCC with two `ViewChangeVote`s from the same signer (differing `highestKnownQc`) must **not** meet quorum.
- `CertBuilderDeterminismSuite`: builder output invariant under permutation of vote arrival order (property test).
- `RumorHandlerSymmetrySuite`: ViewChangeVote ingestion rejects `origin =!= signer` (parity with timeout/eviction/admission).

**Currency consensus**
- `CurrencyCreatorViewSeedSuite`: creator seeds `view = 0`; two creators with different clocks produce identical view/leader.

**Quorum shrink**
- `QuorumShrinkIntersectionSuite`: assert no two disjoint anchor subsets can both satisfy `requiredQuorum` at any stage; eviction cap derived from unshrunk supermajority.

**Schema / cross-version**
- `GlobalConsensusOutcomeDecodeSuite`: decode JSON omitting each new field (rolling-upgrade vector) — should succeed to empty.

**Fork recovery / replay**
- `RecoverySourceSelectionSuite`: rollback target requires majority agreement; recovery prefers the majority-peer hint; single/stale peer cannot trigger destructive `cleanupAbove`.
- `ReplayDeterminismSuite`: byte-identical replay across the `scFeeBalanceFromContext` and dust-sweep ordinals (pre-/at-/post-gate). `ConfigLoadSuite` covers config load but not replay.

---

## Merge instructions
- **Use a merge commit (no-ff). Do NOT squash.** The 264-commit history carries per-commit migration fences and `deterministicConfigHash`/`consensusSchemaVersion` bumps that reference specific commit hashes in comments (e.g. `dd6e83a19`); squashing destroys that traceability and the merge topology.
- develop's HEAD == merge-base, so the integration is linear — **no develop-only fixes are at risk of being overwritten.** The genuine conflict-resolution surface is only 4 files (`.github/.../lib.js`, `e2e-just-test.yml`, `.gitignore`, `docker/bin/test-fork-recovery.sh`); eyeball the `lib.js` resolution.
- **Needs verification (out of band):** confirm PR #1485 is fully contained in this line and won't be double-applied — I could not inspect #1485 in this review.
- Add a deploy-runbook note: develop↔v4.1.0 require a **coordinated cutover** (signed `peerHistory` + Kryo ID 535 reuse make mixed-version operation unsafe during the window).

## Confidence
- **High** on the three CRITICALs: #1 verified end-to-end against post-PR source; #2/#3 each cross-corroborated by two independent reviewer passes with exact file:line + the explicit dag-l0-vs-currency asymmetry.
- **Medium-high** on the HIGHs (#4–#11): each has file:line evidence; #4/#5/#8/#9 corroborated across multiple passes. Exploit reachability varies (some need an equivocating peer; #1/#4/#5/#10/#11 do not).
- **Not fully covered (time/rate-limit):** an independent adversarial-verifier pass on every finding (the automated verifiers were cut off by repeated account-level 429s); `snapshot-rewards` and `docs-consistency` got only manual spot-checks, not a full dimension pass; dynamic/e2e behavior (actual replay, actual OOM timeline, actual fork-recovery convergence) was reasoned about, not executed. The `sbt` build/tests were not run in this environment.

scasplte2:
Thanks for the thorough review. We re-verified every claim independently against the current `feat/v4.1.0` tree (multiple read-only passes reading the source directly rather than trusting line numbers), then cross-checked against a second external review. Summary: the review is **substantively correct on the three things that matter most (the condModifyState race, the signer-vs-vote quorum count, and the currency wall-clock view seed)** and those are fixed. A few HIGHs are real but overstated in reachability (blocked by builder dedup / the version-hash handshake), and a handful are refuted. Disposition below, with commits.

### Fixed in this PR

| # | Claim | Disposition | Commit |
|---|---|---|---|
| 1 | condModifyState lost-update race | The "all writers go through the FSM command loop" comment was false: `StallDetector.monitor` called `abandonRound -> condModifyState` on its own fiber. Fixed by routing the abandon through the command queue (`AbandonRound` cmd) so every `condModifyState` runs on the single command-loop fiber. Added an outcome-ready gate so a round that completes in the decision-to-drain window is not wiped. We did **not** restore the global `stateUpdateSemaphore` (it caused 259 testnet stalls). | `089d75592` |
| 2 | VCC/TC quorum counts votes, not signers | Confirmed (a signer can carry two `Signed[ViewChangeVote]` via `highestKnownQc` identity). `certQuorumMet` now counts distinct signers. | `32dc5bf29` |
| 3 | currency creator wall-clock view seed | Confirmed for any core>=2 metagraph (single-core is `isSoloCore`-exempt). `initialView = 0`, matching the dag-l0 fix; `timeView` kept only as a pacemaker hint. | `32dc5bf29` |
| 4 | VC/timeout vote store LWW + no origin===signer | Origin===signer guard added to `handlePeerVote` for parity. We deliberately **kept last-write-wins** (did not switch to first-write-wins): VC/TC votes carry a mutable `highestKnownQc` and honest nodes re-emit upgraded QCs, so FWW would silently drop a QC carry-forward (a liveness regression). The builder re-dedups by verified signer, so a relay cannot inflate quorum; residual risk is liveness griefing, which the origin guard closes. | `32dc5bf29` |
| 5 | cert-builder `.head` non-deterministic dedup | Confirmed. Both builders now keep all of a signer's votes, run the divergent-QC check over the full set *before* collapse, then select the highest-QC representative (total-Order tiebreak). | `32dc5bf29` |
| 7 | eviction cap from shrunk quorum | Confirmed but latent (assembly uses full `q`). Cap now derives from base (un-overridden) quorum. | `32dc5bf29` |
| 11 | unbounded MapRef growth | 6/7 refuted (those MapRefs are cleared every round via `clearResources(N-1)`). The one real slow leak - `timeoutCertificateApplyScheduledR`, never cleared - is now cleared in `clearResources` / `softReset` / `clearAllConsensusState`. | `32dc5bf29` |
| M3 | `syncFullIfNeeded` build() outside lock | Confirmed; the verification build is now under the mutation lock (released before the continuation, so no re-entrant deadlock). | `02e790ccc` |
| L1 | viewInterval div-by-zero | Not config-reachable, but guarded. | `32dc5bf29` |
| L3 | leader-weighted selection no PeerId tiebreak | PeerId tiebreak added. | `32dc5bf29` |
| L4 | `KeyDecoder[SnapshotOrdinal]` fails open to MinValue | Now fails closed (decode failure on negative/invalid keys). | `32dc5bf29` |

### Recovery hardening - also fixed in this PR (`509b327d9`)

- **#9 destructive recovery rollback (highest-value)** - `recoveryStart` did `cleanupAbove`/`deleteAbove` (irreversible) off a single tip. Now directional: forward/same-ordinal recovery is unchanged, but a backward delete (`metadata.ordinal < localOrdinal`) requires a strict majority of Ready peers (>= 2) to corroborate the exact `(ordinal, hash)` target, else it fails closed (`RollbackTargetNotCorroborated`) and the recovery loop retries. The bounded wait buys liveness; it does not license deleting local state on a sampled tip.
- **#8 RecoveryPeerHint dead code** - the fork-recovery majority hint had no readers. Now threaded into `selectForRecovery`, biasing source selection by intersecting the hint with the already-validated majority-ordinal/majority-hash candidate set (sampling as fallback) - it biases the choice, never bypasses validation.

### Additional hardening (inert, ordinal-gated) (`f95a719f2`)

- **Per-sub-trie state-proof roots** - the MPT `GlobalSnapshotStateProof` carried only `mptRoot`; its per-field `Option[Hash]` fields were always `None`, so a divergence could only be seen at the whole-state-root level. Those fields are now populated with per-`GlobalStateFieldId` sub-trie roots so a divergence localizes to a field (the MPT-divergence work). It is a signed artifact, so it ships **inert** behind a `subTrieRootsActivationOrdinal` that defaults to `MaxValue` (never) and changes nothing until a deliberate coordinated cold cutover. Both proof paths (leader producer + validator rebuild) share one assembler so they stay byte-identical; tests cover field-wiring + the inert default.

### Remaining hygiene (not a blocker)

- **#10 GlobalConsensusOutcome non-Option fields** - confirmed the `useDefaults=false` derivation ignores the field defaults, so the "rolls over cleanly" comments are wrong. The cross-version break is **unreachable** today (the `Joining` version-hash handshake refuses a mixed cluster), so this is forward-compat hygiene, not a blocker. Will Option-wrap (or enable `useDefaults`) so the next field added doesn't break same-version peers.

### Refuted / overstated (no change)

- **M4a, M4b, M5, L7** - refuted on the merits (details in our notes): stage-1 shrink is intersection-safe; `RecoveryFallbackEligible` fires on local chain-gap not transient bad peers; the `ConsensusEndTime` comment is accurate.
- **#4 / #10 headline severity** - the quorum-inflation (#4) and rolling-upgrade-break (#10) impacts are blocked by builder dedup and the version handshake respectively; fixed/hardened for hygiene, not as the stated criticals.
- **L5 (Kryo LinkageError swallow)** is intentional/documented; **L6 (Kryo ID 535 reuse)** has no live collision and is jar-hash-gated.

### Design-accepted (testnet-only, intentional)

- **#6 stage-2 quorum-denominator shrink** (`MinQuorumFloor=2`) is a documented liveness-over-partition trade, gated **off** mainnet/integrationnet/dev (all `0`) and live only on testnet (`10`). **#7/M2** are only reachable when that rung is active. No change unless we decide to drop the rung.

### Deploy

- **M1 / #10 / L6**: develop <-> v4.1.0 require a **coordinated cold cutover, not a rolling upgrade** - signed `peerHistory` changes artifact bytes (fenced by `consensusSchemaVersion=33` in `deterministicConfigHash`) and Kryo ID 535 was repurposed. The version-hash handshake already prevents a mixed cluster, so these surface as connection-refusal rather than corruption. Merging as a **merge commit (no-ff), not squash**, as noted.

ryle-goehausen (approved):

marcus-girolneto (approved):

ryle-ai (commented):
Reviewing this after the fact since it merged without one. I focused on where a 256-commit reconciliation actually goes wrong: the conflict resolutions, the replay-safety ordinal gates, reward conservation, and the merge mechanics themselves. Most of it holds up. The parts worth flagging are below, roughly in order of how much they'd bite.

**The merge went in as a squash, which is the one thing the description said not to do.** The commit on develop (`c0106c90c`) has a single parent, so develop gained exactly one commit, not the 256. That breaks the superset relationship you called out, and it has a concrete consequence for the step you planned next: `release/testnet` is no longer an ancestor of develop (it reads 256 commits ahead by `git rev-list`), so the fast-forward realign in the runbook can't happen. The options now are to force `release/testnet` to develop's tip and discard its divergent history, or do another reconciliation merge that double-counts everything. Better to decide that deliberately than hit it at deploy time. Side note: the real base was `ab5d4c628`, not `9bbbe358e` as the description says. Develop picked up the "don't drop rewards when withdrawals expire together" commit (the twin of testnet's `d4e506b4b`) in between and the squash re-derived onto it. That re-derivation is also why a naive `--is-ancestor` check on the reward fix comes back negative even though the fix is present (more below).

**Review-focus #1, the AbandonmentTracker isolated-node case: I think the escalation actually was dropped.** `networkAdvanced` is `peersAtHigherKey > 0` computed from `getPeerCurrentKeys`, which is fed only by incoming keyed rumors. For a node cut off from rumors but still able to reach peers for download (exactly what `8027c0642`'s `(!peersHaveCurrentKey && totalReadyPeers > 0)` clause targeted), those tips freeze and `networkAdvanced` stays false. I traced every escalation route in the merged code to be sure it isn't caught elsewhere: StallDetector's lagging check is the same peer-observation predicate; its RoundTimeout/MaxStalls path is genuinely time-based but still gated by `willRecover = shouldRecover && networkAdvanced`; `triggerRecoveryDownload` is the only caller of `forceLeave` and the only place `totalRecoveryAttempts` increments, so an isolated node never reaches the force-leave path either. The new wedge-health signal only makes the node refuse external `Cluster.leave` POSTs, it doesn't escalate. So a rumor-isolated node retries its abandoned key indefinitely instead of going to recovery-download. I'll defer to what 158 alphas on testnet actually showed. If this case was deliberately abandoned because the old escalation caused false recoveries, that's a fine call, but it reads as an unintended gap rather than a decision, and it's the one finding here I'd want confirmed before the v4.1.0 launch.

**The snapshot-streaming patch may not compile against its target.** `docker/snapshot-streaming/snapshot-streaming.patch` carries testnet's `TessellationServices.scala` hunks but drops the `Configuration.scala` hunk that removed `forkInfoStorage` and `combinedRouteRateLimiter`, and merged tessellation does remove both fields from `SharedConfigReader`/`GlobalL0Config`. The patch applies with `git apply --check` (build-snapshot-streaming.sh:51), so it skips rather than hard-failing, but if snapshot-streaming@release/testnet's `Configuration.scala` still forwards those fields into the merged constructors, the downstream `sbt assembly` breaks. I couldn't confirm it from this repo since it depends on the external snapshot-streaming source, but it's cheap to check before relying on that image.

**Minor, but an operator footgun: a stale comment in the ordinal config.** `config/types.scala:38-41` says testnet's `sc-fee-balance-from-context` ordinal "is 0 (the fix is already live there)", but the loaded value is `3101393` in both the HOCON and `fields-added-ordinals.md`. No runtime effect since the real value wins, but if someone trusts the comment and sets testnet to 0 they'd activate the context fee-balance path below the cutover and fork the replay window. One-line fix.

For what it's worth, the surfaces I expected to be risky checked out. The token-lock e2e resolution is byte-identical to develop, `ClickHouseSink` correctly took testnet's connect-timeout hardening, the `PosInt -> Int` change on `quorumShrinkActivationViews` is the right fix for the alpha.158 startup crash and is safe (0 is the disabled sentinel and nothing divides by it), the withdrawal-expiry reward dedup survived intact at `GlobalDelegatedRewardsDistributor.scala:519-533`, the reward-rotation removal left no dangling references, and the three new ordinal gates plus the Advancer's ordering all hold up deterministically (fail-closed on unset env, keyed on snapshot ordinal, no wall-clock). Given the size I didn't re-verify the shared selector internals (`FacilitatorSelector`, `TierTransitions`); nothing in the diff suggests they changed, but those are the remaining spots if you want airtight coverage.

PR #1525, opened 2026-06-19
fix(l1): coordinate L1 storage mutations to fix fork/lost-sync/lastTxnRef

marcus-girolneto:
## Why

L1 layers (dag-l1, currency-l1, data-l1) have been forking, losing sync, holding
divergent POV, and rejecting `lastTxnRef`. An audit of all L1↔L0 sync paths found a
**systemic root cause**:

> L1 sync is ~7 uncoordinated concurrent fs2 loops mutating shared in-memory storage.
> The L0→L1 alignment writer holds **none** of the consensus semaphores, runs against
> the **live** singleton storages (no copy-isolation), and applies a **non-atomic,
> fail-fast, no-rollback** sequence. The load-bearing divergence is `AddressStorage`'s
> two-op `clean` + `updateBalances` interleaving with the live acceptance path's balance
> delta — a stray balance leaks past `clean` and persists, so two honest nodes applying
> the **same** global snapshot end with different balances → balance / `lastTxnRef` rejections.

This PR lands the **P0 + P1** fixes from the audit on one branch.

### P0 — decoupled quick wins
- **A1** `AddressStorage.replaceAll` — atomic whole-map balance publish replaces the racy `clean`+`updateBalances` (kills the load-bearing divergence).
- **A2** `advanceMajorityRefs` now surfaces its `Either` (+log) instead of silently swallowing a local-vs-majority divergence (idempotent re-mark stays a no-op).
- **A3** `GlobalL0AlignmentStorage.consumeShouldRedownload` — atomic `getAndSet` so a concurrently-raised redownload flag isn't dropped by a separate `clean()`.
- **A4** `L0BlockOutputClient` — HTTP `202` (awaiting-parent) is treated as **provisional**, not terminal; the block stays buffered for re-send until L0 confirms inclusion (`200`).
- **A5** data-l1 `Engine.combine` — canonical-sort merged data updates by content hash so every facilitator builds a byte-identical block (multisig stops failing; ≥2-facilitator rounds can finalize).

### P1 — coordination core
- **B1** `storageMutationLock` (`cats.effect.std.Mutex`, fair FIFO) shared by `BlockService.accept`/block-store and `SnapshotProcessor.processAlignment`'s commit. Wraps **only** microsecond in-memory writes — never network, hashing, or the MPT trie build — so L1 stays live while aligning.
- **B2** `adjustToMajority` precondition dry-run — all-or-nothing; on block-state drift it aborts before any mutation and schedules a redownload (no partial, unrollback-able reconciliation).
- **B3** `replaceByRefs` preserves in-flight `Waiting`/`Processing` txs above the new majority ref (fixes the `pull → align → putBack` silent drop).
- **B4** batch processing retries a failing snapshot transiently and stops at the first exhausted failure instead of skipping over a gap (a transient hiccup no longer escalates to a full redownload).

### Bonus (from the transaction-storage audit)
- **TX-07** `ContextualTransactionValidator` no longer double-counts an overridden tx's `amount+fee` in the balance/limit checks on a fee-bump (`>=` instead of `>`).

## Tests
- New regression tests: `AddressStorageSuite` (`replaceAll` replaces the whole map), `TransactionStorageSuite` (`replaceByRefs` preserves an in-flight `ProcessingTx`).
- All **dag-l1** suites green (35/35), incl. the mutex-affected `SnapshotProcessorSuite`/`BlockServiceSuite` and TX-07's validator suite; dag-l0 cutter suite green (4/4). `sbt runLinter` clean.
- Full Docker E2E (`just test`) not run in this PR.

## Scope notes / deferred follow-ups
- No wire/snapshot/P2P-schema change — every change is per-node and reversible (mixed-version-cluster safe).
- The three legacy semaphores are intentionally kept alongside the new mutex during transition (harmless redundancy; retire in a separate PR).
- **Deferred:** A6 (dag-l0 `GlobalSnapshotEventCutter` — its comparator is non-antisymmetric; needs a topological *total* order co-designed with its suite) and the full **P2** hardening set (currency/swap/token-lock resend coverage, tip validation, pull quorum/chain verification, `CL_RAISE_ON_FOLLOWER_DIVERGENCE` default-on, sync-gate gap tolerance).

### DIFF
--- a/modules/currency-l1/src/main/scala/io/constellationnetwork/currency/l1/CurrencyL1App.scala
+++ b/modules/currency-l1/src/main/scala/io/constellationnetwork/currency/l1/CurrencyL1App.scala
@@ -343,7 +343,8 @@ abstract class CurrencyL1App(
               validators.allowSpend,
               keyPair,
               nodeId,
-              storages.globalL0Alignment
+              storages.globalL0Alignment,
+              storages.storageMutationLock
             )
             .merge {
               TokenLock.run[IO, CurrencySnapshotStateProof, CurrencyIncrementalSnapshot, CurrencySnapshotInfo, Run](
@@ -361,7 +362,8 @@ abstract class CurrencyL1App(
                 validators.tokenLock,
                 keyPair,
                 nodeId,
-                storages.globalL0Alignment
+                storages.globalL0Alignment,
+                storages.storageMutationLock
               )
             }
             .merge(stateChannel.runtime)
--- a/modules/currency-l1/src/main/scala/io/constellationnetwork/currency/l1/CurrencyL1App.scala
+++ b/modules/currency-l1/src/main/scala/io/constellationnetwork/currency/l1/CurrencyL1App.scala
@@ -169,7 +169,8 @@ abstract class CurrencyL1App(
         services.globalL0.pullGlobalSnapshot,
         services.globalL0,
         storages.globalL0Alignment,
-        sharedStorages.mptStore
+        sharedStorages.mptStore,
+        storages.storageMutationLock
       )
       programs = Programs
         .make[IO, CurrencySnapshotStateProof, CurrencyIncrementalSnapshot, CurrencySnapshotInfo, Run](
--- a/modules/currency-l1/src/main/scala/io/constellationnetwork/currency/l1/Swap.scala
+++ b/modules/currency-l1/src/main/scala/io/constellationnetwork/currency/l1/Swap.scala
@@ -4,7 +4,7 @@ import java.security.KeyPair
 
 import cats.Applicative
 import cats.effect.kernel.Async
-import cats.effect.std.Random
+import cats.effect.std.{Mutex, Random}
 import cats.syntax.all._
 
 import scala.concurrent.duration._
@@ -47,7 +47,8 @@ object Swap {
     allowSpendValidator: AllowSpendValidator[F],
     selfKeyPair: KeyPair,
     selfId: PeerId,
-    globalL0AlignmentStorage: GlobalL0AlignmentStorage[F]
+    globalL0AlignmentStorage: GlobalL0AlignmentStorage[F],
+    storageMutationLock: Mutex[F]
   ): Stream[F, Unit] = {
 
     def logger = Slf4jLogger.getLogger[F]
@@ -165,17 +166,18 @@ object Swap {
             }.flatMap {
               _.toList.traverse {
                 case (hash, signedBlock) =>
-                  services.allowSpendBlock
-                    .accept(signedBlock, snapshotOrdinal)
-                    .handleErrorWith { error =>
-                      for {
-                        _ <- logger.warn(error)(s"Failed acceptance of an allow spend block with ${hash.show}")
-                        _ <- globalL0AlignmentStorage.updateShouldRedownload(
-                          value = true,
-                          reasons = List(s"Allow spend block acceptance failed for ${hash.show}: ${error.getMessage}")
-                        )
-                      } yield ()
-                    }
+                  storageMutationLock.lock.surround {
+                    services.allowSpendBlock
+                      .accept(signedBlock, snapshotOrdinal)
+                  }.handleErrorWith { error =>
+                    for {
+                      _ <- logger.warn(error)(s"Failed acceptance of an allow spend block with ${hash.show}")
+                      _ <- globalL0AlignmentStorage.updateShouldRedownload(
+                        value = true,
+                        reasons = List(s"Allow spend block acceptance failed for ${hash.show}: ${error.getMessage}")
+                      )
+                    } yield ()
+                  }
               }
             }.void
           case None => ().pure[F]
--- a/modules/currency-l1/src/main/scala/io/constellationnetwork/currency/l1/TokenLock.scala
+++ b/modules/currency-l1/src/main/scala/io/constellationnetwork/currency/l1/TokenLock.scala
@@ -4,7 +4,7 @@ import java.security.KeyPair
 
 import cats.Applicative
 import cats.effect.kernel.Async
-import cats.effect.std.Random
+import cats.effect.std.{Mutex, Random}
 import cats.syntax.all._
 
 import scala.concurrent.duration._
@@ -49,7 +49,8 @@ object TokenLock {
     tokenLockValidator: TokenLockValidator[F],
     selfKeyPair: KeyPair,
     selfId: PeerId,
-    globalL0AlignmentStorage: GlobalL0AlignmentStorage[F]
+    globalL0AlignmentStorage: GlobalL0AlignmentStorage[F],
+    storageMutationLock: Mutex[F]
   ): Stream[F, Unit] = {
 
     def logger = Slf4jLogger.getLogger[F]
@@ -167,17 +168,18 @@ object TokenLock {
             }.flatMap {
               _.toList.traverse {
                 case (hash, signedBlock) =>
-                  services.tokenLockBlock
-                    .accept(signedBlock, snapshotOrdinal)
-                    .handleErrorWith { error =>
-                      for {
-                        _ <- logger.warn(error)(s"Failed acceptance of a token lock block with ${hash.show}")
-                        _ <- globalL0AlignmentStorage.updateShouldRedownload(
-                          value = true,
-                          reasons = List(s"Token Lock block acceptance failed for ${hash.show}: ${error.getMessage}")
-                        )
-                      } yield ()
-                    }
+                  storageMutationLock.lock.surround {
+                    services.tokenLockBlock
+                      .accept(signedBlock, snapshotOrdinal)
+                  }.handleErrorWith { error =>
+                    for {
+                      _ <- logger.warn(error)(s"Failed acceptance of a token lock block with ${hash.show}")
+                      _ <- globalL0AlignmentStorage.updateShouldRedownload(
+                        value = true,
+                        reasons = List(s"Token Lock block acceptance failed for ${hash.show}: ${error.getMessage}")
+                      )
+                    } yield ()
+                  }
               }
             }.void
           case None => ().pure[F]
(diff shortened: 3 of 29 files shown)


marcus-girolneto:
## Code review + resolution

A fresh, independent review was run over this PR (reviewers saw only the diff, not the authoring rationale), each finding adversarially verified. Posting the review, the resolution, and the follow-up re-review for the record.

### Initial verdict: Request changes

**Blockers**
- **DR-1** — the "surface the error instead of swallowing it" fix on `advanceMajorityRefs` was applied only to dag-l1 `TransactionStorage`; the structurally identical `AllowSpendStorage` and `TokenLockStorage` (node-shared) still silently swallowed the `Left`. Since `markTxRefsAsMajority` sequences all three, an allow-spend / token-lock majority-ref divergence stayed silently stale while the DAG-tx path escalated.
- **LC-1** — `advanceMajorityRefs` raising mid-sequence (after `adjustToMajority` had already mutated block storage) produced a **cross-store partial commit**: block storage advanced, `lastSnapshot` not, tx refs partially marked. Worse than first thought — after the partial commit both the aligned retry *and* the redownload fail the B2 block preconditions, so it could **wedge the node**, not just force a redownload.

**Major**
- **CL-1** — the `storageMutationLock` comments claimed it "guards only microsecond Ref/MapRef writes," but the acceptance side holds it across the whole `accept`, including block/tx hashing **and ECDSA signature verification**. Misleading to anyone reasoning about contention.

**Test gaps** — TX-07 had no test that fails on revert (existing override test uses `MaxValue` balance, masking it); B2's all-or-nothing guarantee was untested (no `BlockStorageSuite` existed at all).

Plus ~10 minor/nits (lock scope on the currency replay, a flag-consume edge, dead code, a delimiter-less sort key, etc.).

### Resolution

**`fc5f1ac91` — blockers + tests**
- **DR-1** — applied the surface-error + idempotent-remark fix to `AllowSpendStorage` and `TokenLockStorage.advanceMajorityRefs` too; all three now behave identically.
- **LC-1** — added a pure, pre-mutation `majorityRefsViolations` check to all three storages, run **first** inside the locked region of both Aligned branches of `processAlignment` (new `MajorityRefsDiverged` error). A drift now aborts with **nothing mutated**, eliminating the partial commit / wedge.
- **CL-1** — corrected the `Storages` + `StateChannel` lock comments to honestly state hashing + signature verification run under the lock, with the justification (acceptance is a serial 1s-tick stream; MPT build + network stay outside).
- **T1** — `ContextualTransactionValidatorSuite`: a fee-bump override is no longer double-counted against the rate limit (fails on a revert to `>`).
- **T3** — new `BlockStorageSuite`: `adjustToMajority` aborts atomically (mutates nothing) on an unexpected block state, and applies when preconditions hold.

**`a51c4ba3f` — nits**
- **CL-2** — the currency-l1 intermediate-replay `processAlignment` now uses a dedicated uncontended lock (its storages are private throwaway copies); the real commit still uses the shared lock.
- **CL-3** — `validateTipsAlignment` now runs before `consumeShouldRedownload`, so a tips-misalignment error no longer clears a still-needed redownload flag.
- **DR-2** — removed the now-unused `getShouldRedownload` / `clean()` (only `consumeShouldRedownload` remains), preventing reintroduction of the get-then-clean race.
- **DR-3** — documented the semaphore-vs-mutex layering. **DR-4** — dropped the dead outer `withRetry`. **A5/LC-4** — delimiter-separated the data-l1 bundle sort key. **LC-3** — documented that the boolean `sendL1Output` is lossy for HTTP 202.

### Re-review verdict: ✅ Approve

All three blockers were independently re-verified against the current code (not taken on the author's word), and the new code was scanned for regressions:
- **DR-1, LC-1, CL-1 — resolved.** The `majorityRefsViolations` predicates exactly mirror `advanceMajorityRefs`'s accept arms (no false-positive/false-negative), the precheck runs first-in-region so nothing mutates before the abort, the non-reentrant `Mutex` is never re-acquired under itself, and the corrected comments match the actual lock scope.
- **No new defects introduced** by the fixes (CL-2 replayLock, CL-3 reorder, DR-2 removal, DR-4, A5 all verified sound).

**Tests:** dag-l1 38/38 green (incl. the 3 new regression tests), node-shared + currency-l1 compile, `sbt runLinter` clean. Full Docker E2E (`just test`) not run in this PR.

### Out of scope (pre-existing, not regressions from this PR — track separately)
- **A6** — dag-l0 `GlobalSnapshotEventCutter` needs a topological *total* order (its comparator is non-antisymmetric) co-designed with its suite.
- **P2 hardening** — resend coverage for swap/token-lock/data-application output pipes, tip validation, pull quorum/chain verification, `CL_RAISE_ON_FOLLOWER_DIVERGENCE` default-on, sync-gate gap tolerance.

🤖 Review assisted by [Claude Code](https://claude.com/claude-code)

scasplte2:
Follow-up pushed: `cb69eb1c8 fix(l1): serialize swap/token-lock acceptance with alignment`.

This is additive to the already-posted review/resolution, not a replacement for it. The existing fixes cover `advanceMajorityRefs`, all-or-nothing alignment prechecks, redownload flag consumption, and the documented DAG block-acceptance lock scope.

The remaining coverage gap I found was that swap/token-lock block acceptance also mutates the shared `AddressStorage` via `addressStorage.updateBalances`, but those acceptance streams were still outside `storageMutationLock` while alignment commits `addressStorage.replaceAll` under the lock. That left the same class of balance interleaving open through:

- dag-l1 `Swap` / `TokenLock`
- currency-l1 `Swap` / `TokenLock`

The new commit threads the existing `storageMutationLock` into those four runners and wraps the corresponding `allowSpendBlock.accept` / `tokenLockBlock.accept` calls, so DAG tx, swap, and token-lock acceptance now share the same serialization boundary against L0->L1 alignment commits.

Verification:
- `git diff --check` clean.
- `dagL1 / Compile / scalafmt` and `currencyL1 / Compile / scalafmt` clean.
- Compile was attempted twice, but the local environment killed sbt with exit 137 after several minutes of transitive shared/node-shared compilation; no Scala compile error appeared before the kill.

marcus-girolneto:
## Code Review — L1 Sync Coordination (post-merge audit)

**Verdict: ✅ Approved / good to go — with 1 required follow-up (M1) and a few nits.**

High-quality concurrency fix. Lock ordering is disciplined and deadlock-free (per-stream `Semaphore(1)` outermost, `storageMutationLock` innermost — no cycle), heavy/blocking work is correctly kept *outside* the critical section, and the dry-run-then-mutate pattern gives all-or-nothing cross-store commits. No CRITICAL/HIGH blockers. Since this is already merged, the items below are framed as follow-ups.

---

### 🟡 Medium

**M1 — Redownload request can be lost on alignment-commit failure (required follow-up).**
`SnapshotProcessor.scala:494-495` and `:590-591`. `consumeShouldRedownload` clears the flag (`getAndSet(empty)`) at decision time. If `determineAlignment` returns `RedownloadNeeded` but the subsequent commit (`adjustToMajority`/`setSnapshot`/`updateMptStorage`) throws, the flag is already gone. The block-acceptance path re-raises on failure (`StateChannel.scala:404`), but the **alignment**-commit path does not.

*Failure scenario:* L0 requests redownload → alignment cycle consumes the flag → MPT build or `setSnapshot` throws transiently → next cycle reads an empty flag and proceeds as if aligned → required redownload silently skipped. This is exactly the "lost sync" class this PR targets.

*Suggested fix:* re-arm the flag on alignment-commit failure, symmetric to the acceptance path, e.g. `.onError { case _ => globalL0AlignmentStorage.updateShouldRedownload(true, List("alignment commit failed, re-arming redownload")) }`. Confirm `updateShouldRedownload` uses OR-on-true semantics so a concurrently-raised flag isn't clobbered.

**M2 — Dry-run atomicity rests on an unstated lock invariant.**
`BlockStorage.adjustToMajority` does a precondition dry-run over `blocks.toMap` then mutates via `MapRef`. This is only atomic because every block-storage writer holds `storageMutationLock` — enforced by convention, not by types. A future contributor adding an unlocked `blockStorage` write silently reintroduces the partial-commit bug.

*Suggested fix:* scaladoc on `adjustToMajority` stating the caller MUST hold `storageMutationLock`; ideally a regression test proving it's the *lock* (not the dry-run) that prevents drift between dry-run and apply. Existing tests prove the dry-run aborts on drift but not that the lock keeps drift from occurring.

### 🔵 Low / Nit

- **L1** — `updateMptStorage`/`setLastNSnapshots` run outside the lock after `setSnapshot` runs inside it (`SnapshotProcessor.scala:155-158`). Brief window where `lastSnapshot` leads the MPT/rolling-N cache. Correct *iff* no reader joins `lastSnapshot` with the MPT/`lastNSnapshots` expecting consistency — please confirm + add a one-line comment documenting the accepted staleness window.
- **L2** — `prepareIntermediateStorages` (currency-l1) gives intermediate/dry-run storages a fresh `Mutex` rather than the shared `storageMutationLock`. Correct, but add a comment noting the separation is deliberate to avoid future confusion.
- **L3** — The `Aligned` / `DownloadPerformed` / `RedownloadNeeded` commit skeletons are near-identical 3×; a private `commitAligned(...)` helper would prevent drift (M1 already affects two of these paths).
- **L4** — Outer `withRetry` dropped at `GlobalSnapshotAlignment.scala:174`. Reasoning is sound (each snapshot retries internally); confirm only per-snapshot work is fallible and batch-level setup can't fail unretried.

### ✅ Strengths

- Lock ordering is disciplined and documented (`StateChannel.scala:394-396`, `:448-454`) — no deadlock cycle. This is the single most important property here and it's correct.
- Heavy work (MPT trie build, L0 network I/O, hashing) correctly excluded from the critical section → minimal contention.
- Root-cause fix is minimal and exactly right: `AddressStorage.replaceAll` collapses the `clean()`-then-`updateBalances()` window into a single `Ref.set`, structurally eliminating leaked-balance divergence.
- Dry-run precondition pattern fails loudly (`UnexpectedBlockStatesWhenAdjustingToMajority`, `MajorityRefsDiverged`) instead of half-applying, with regression tests.
- Acceptance failure re-arms redownload (`StateChannel.scala:404`) — good self-healing (M1 asks for the same on the alignment path).

---

### Open questions
1. **M1:** Does alignment-commit failure (not `block.accept`) re-raise the redownload flag anywhere, or does the next L0 pull independently re-derive the need? If the latter, M1 downgrades to Low.
2. **L1:** Any reader that joins `lastSnapshot` with the MPT store / `lastNSnapshots` expecting consistency?
3. Does `updateShouldRedownload(value, reasons)` overwrite or OR the boolean? Determines whether a concurrently-raised flag can be lost.

_Reviewed by an automated comprehensive code review pass._

PR #1526, opened 2026-06-22
Feat/l0 burn transaction

Tech0001:
I know this is the wrong branch to submit a feature request, but develop is on Tessellation 3.  So I grabbed the most recent testnet on V4 for developing this.

Constellation currency metagraphs can mint (RewardTransaction) but have no first-class way to burn. The only supply-decrement, BalanceAdjustment, is gated and semantically a spend-failure clawback; and because totalSupply is the sum of balances, sending to a dead address doesn't reduce reported supply — so the protocol currently cannot truthfully represent destroyed tokens. This blocks deflationary tokenomics and accurate supply accounting for every currency metagraph, and metagraphs can't add it themselves (global-L0 snapshot replay rejects self-invented supply rules). 

This adds BurnTransaction/BurnAction to close that gap. A metagraph can permanently destroy units of its own currency and have totalSupply go down with it, either by consuming a holder's AllowSpend (burnFrom) or by burning from its own address (self-burn). Burns get re-derived by global-L0 like any other currency artifact, so they're fully auditable, and the same AllowSpend can't be used by both a spend and a burn in one snapshot.  There's an optional opaque intentHash commitment field for higher layers to tie a burn to some off-chain intent.

There is a modified Euclid that I'll be PR'ing too that has a standalone build with this on the same version of Tess 4 for easy testing.


Tech0001:
I built this on the v4.1.0 testnet line (alpha.159); I see #1523 is integrating that line into develop, so once that lands this should rebase onto develop cleanly. It gates activation via the same FieldsAddedOrdinals mechanism.

scasplte2:
@Tech0001 apologies if I rebased poorly but can we have this target the develop branch. This is the preferred feature integration process 
https://github.com/Constellation-Labs/tessellation/blob/develop/docs/release/BRANCHING_AND_RELEASE_WORKFLOW.md#version-progression-example

Tech0001:
Thanks @scasplte2, no worries on the rebase. The branch was built on the release/testnet (alpha.159) line, so targeting develop initially pulled in the whole testnet→develop delta. Now that #1523 has merged that line into develop, I've redone the rebase to replay just the four burn commits onto develop — the PR is now the ~18-file burn feature rather than the full divergence, and mergeable.

One small adaptation for develop: FieldsAddedOrdinals.burnActionActivation now defaults to Map.empty (fail-closed — no environment activates burns until a coordinated ordinal is set), matching its siblings scFeeBalanceFromContext / subTrieRoots / dustSweeps, so develop's existing FieldsAddedOrdinals(...) call sites keep compiling. The burn suites (BurnActionValidatorSuite, BurnTransactionBalanceSuite, BurnActionSnapshotIntegrationSuite) pass on the develop base.

Ready for review.

Tech0001:
@scasplte2 Gentle bump on this one. Since the rebase it's just the ~18-file burnfeature on develop (4 commits), and it's fail-closed by default: burnActionActivation defaults to Map.empty, matching its siblings (scFeeBalanceFromContext / subTrieRoots / dustSweeps), so merging changes zero behavior in any environment until a coordinated activation ordinal is explicitly set. The three burn suites (BurnActionValidatorSuite, BurnTransactionBalanceSuite, BurnActionSnapshotIntegrationSuite) passed on the develop base at the rebase, and I've just brought the branch current with develop again.

Happy to do whatever makes review easiest on your end — split into smaller PRs, an async design walkthrough, or naming/convention adjustments. Anything blocking that I can pick up?

Motivation, for context: metagraphs can mint via RewardTransaction but have no first-class way to reduce supply — so today. "burning" means parking tokens at a dead address, which leaves totalSupply permanently wrong. This closes that asymmetry: any metagraph gets accurate deflationary supply accounting as a native primitive, same pattern as the existing spend machinery.

ryle-ai (changes requested):
Two things block this before any of the logic can be judged as-is, then a few real problems in the logic itself.

The branch is 25 commits behind develop and `merge-tree` reports 12 conflicting files, including every core file this touches (`CurrencySnapshotAcceptanceManager`, `GlobalSnapshotAcceptanceManager`, `AllowSpendStateManager`, `SpendTransactionBalanceManager`, `types.scala`, `application.conf`, `Mocks` and the five edited suites). GitHub shows it as conflicting. More important than the textual conflicts: the burn appliers (`SpendTransactionBalanceManager.updateGlobalBalancesByBurnTransactions`, `AllowSpendOpsManager.updateCurrencyBalancesByBurnTransactions`) are line-for-line copies of the July spend appliers, and develop has since changed the spend applier (settled allow-spends are now consumed from the pool, plus the `fixingAllowSpendDestinationCredit` / `preventingAllowSpendResurrection` / `fixingGlobalAllowSpendExpiration` gates). A rebase that only resolves the conflicts would leave the burn path as the one remaining unpatched copy of the pre-fix spend semantics. The rebase needs to re-derive the burn appliers from the current spend appliers, not merge the old copies forward.

The activation gate isn't fail-closed, it's fail-crash. `burnActionActivationFor` returns `Left(IllegalStateException)` when the environment has no entry, and both acceptance managers `.liftTo[F]` it at the top of `accept`, so a missing entry makes every snapshot acceptance raise. The comment in `types.scala` and the PR discussion say `Map.empty` "never activates burns" and cite `scFeeBalanceFromContext` / `subTrieRoots` / `dustSweeps` as the pattern, but those are read with `.get` / `.getOrElse(MaxValue)` and are simply inactive when absent. That's also why five unrelated suites had to be edited to construct the field. `burnActionActivation.get(environment)` with `None` meaning disabled matches the siblings and removes the test churn.

Self-burns are validated against pre-spend balances and applied after spends. Currency path: `balancesForBurnValidation` is built from `updatedBalancesByTokenLocks` (`:460`) but the burn is applied on `updatedBalancesBySpendTransactions` (`:542`). Global path validates at `:836` against pre-spend `globalBalances` and applies at `:1065`. So a source with balance X carrying `SpendAction(self, X)` and `BurnAction(self, X)` in the same snapshot passes both validators, `Balance.minus` returns Left in the applier, and both callers do `throw new RuntimeException` (`:548`, `:1072`) rather than a `SnapshotFailure` like the spend path. Since GL0 takes shared artifacts from the metagraph-signed snapshot, a modified metagraph can make every GL0 node throw at that ordinal. Two DAG self-spends already have the same shape, so this is an inherited class, but it's a new instance on a new artifact. Validating against post-spend balances fixes it; `burnAllowSpendRefs` is the only thing the earlier position feeds, and that's just for `acceptCurrencyAllowSpends`.

DAG burns look unreachable for an honest metagraph. The currency path keys the allow-spend and balance maps only by `metagraphId.some` (`:459-460`), so a `currencyId = None` burn fails validation and is stripped from the output artifacts. The global path then applies only `currencyId.isEmpty` burns (`:847`). The DAG leg is live only for a metagraph that signs artifacts its own acceptance rejects. Either the currency path needs a DAG balance view for validation, or the global leg should be dropped until it does.

Smaller: GL0 re-validates currency-token burns at `:836` against `currencyBalances` built from the incoming snapshot info, which is already post-burn, so a full-balance self-burn is rejected globally and the rejection is bound to `_`. Harmless since GL0 never applies currency burns, but the check is a no-op. In the allow-spend applier branch, `NonNegLong.from(allowSpend.amount - burn).getOrElse(MinValue)` silently clamps to zero; the validator makes it unreachable today, so an explicit error there would be cheaper than relying on that. `intentHash` is opaque and has no validator or consumer.

Tests cover the validator and appliers well. Not covered: `GlobalSnapshotAcceptanceManager` end to end, the global gate, spend+burn in one snapshot, and the validation-vs-application balance basis above.

Tech0001:
Thank you for the detailed review. You were right about the missing-configuration failure,
the balance-validation basis and the missing acceptance-level coverage. These are corrected
in this revision.

I have narrowed the proposal to native self-burn only: the emitting metagraph can burn
its own native currency held at its own metagraph address. Delegated/AllowSpend burning
and DAG burning are removed from this proposal, rather than carrying forward their
incomplete execution paths.

| Finding | Correction |
|---|---|
| Stale branch and copied spend appliers | Rebuilt against `develop` at `11cce41db`. Removed the parallel delegated-burn appliers; existing ordinary spend and AllowSpend protections remain unchanged. |
| Missing activation configuration raises during snapshot acceptance | Missing configuration disables burns without interrupting normal snapshot processing. No environment is enabled. Invalid configured ordinals still reject configuration loading. |
| Validation and application use different balances | One checked Currency fold validates and debits after normal accepted accounting, including settled spends. Relevant unapplied Global dependencies prevent burning until catch-up. Actions are atomic and use cumulative balances. |
| Arithmetic throws or silently clamps | Burn validation returns typed rejections; no burn-side underflow exception or allowance clamp remains. |
| DAG path is unreachable for an honest producer | DAG and delegated burning are excluded. Strict codecs and native validation reject unsupported fields, currencies and sources; no Global/DAG burn applier remains. |
| Global revalidation uses already-burned balances and discards the result | Global reconstructs Currency acceptance from the prior Currency state and compares the result. Rejected burn claims cannot be restored as opaque application artifacts, and valid burns are not applied twice. |
| `intentHash` has no validator or consumer | Removed. This proposal makes no application-intent or bridge-withdrawal claim. |
| Missing complete acceptance coverage | Added producer-to-signed-snapshot-to-Global acceptance tests, including malicious signed claims, spend/burn collisions, activation boundaries, restored-state replay and state-proof assertions. |

Follow-up compatibility checks also cover recognition before activation and mixed-channel
fee admission. Disabled or pre-activation payloads retain historical opaque interpretation;
newly interpreted inactive metadata cannot reserve an address against another channel's
active burn. Active fee and address-collision protections remain enforced.

Validation of the reviewed implementation: **259 focused tests passed; the full
import/format/test gate passed 2,593 tests with two existing ignores**. Supplemental
independent compatibility/admission probes passed 10/10, and accepted-prefix reservation
probes passed 4/4. The publication cleanup changes documentation and two test comments,
not executable sources or test assertions.

One independent full-gate invocation failed during configuration-reader compilation in
`CurrencyL0App.scala`; the same source and command passed in a fresh process. Its cause
remains unexplained and no compiler fix is claimed.

The updated PR description includes reproduction commands and the schema/wire
classification. The
[self-burn compatibility proposal](https://github.com/Tech0001/tessellation/blob/42fbf067d6e4aeee50a223809f51b656dc294f81/docs/development/metagraph-native-self-burn.md)
records the compatibility boundary and outstanding consumer/rollout qualifications. Burning remains disabled;
no live-network or external-consumer acceptance is claimed. The Euclid demo is unchanged.

Please re-review the narrowed implementation and proposed compatibility/rollout package.

PR #1529, opened 2026-06-29
fix(e2e): delegated staking node keys dir

effects-ai:


ryle-ai (commented):
The key-resolution change itself is correct: `NODE_KEYS_DIR/<index>/id_ecdsa.hex` matches the per-node layout that `docker/bin/node-key-env-setup.sh` stages, `path.resolve(__dirname, '../../..', keysDir)` lands on the repo root, the index→node mapping (0=genesis, 1=validator-1, 2=validator-2) is consistent, the removed `path` require is genuinely unused, and the local-run filenames still match the bundled fixtures.

One cross-cutting note (non-blocking): `NODE_KEYS_DIR` is set at the *job* level in `nightly-e2e.yml`, so every e2e step inherits it, but only `delegated-staking.js` now honors it via `resolveNodeKeyPath`. The sibling `token-lock-replacement-edge-cases.js` (which the same nightly job runs immediately after) still hardcodes the CI Euclid paths `../../code/hypergraph/dag-l0/<name>/id_ecdsa.hex` and does not use the resolver. In the normal nightly flow this is masked — `setupNodeParameters` early-returns because delegated-staking has already seeded node params, so the hardcoded paths aren't read — but if delegated-staking ever fails to seed params, token-lock-replacement will fall through to those non-existent paths and fail (now with the improved error message from `getPrivateKeyAndNodeIdFromFile`). Worth deciding whether that script should share the same `resolveNodeKeyPath` treatment, or leaving a comment noting the deliberate dependence on delegated-staking running first.

scasplte2:
Thanks @ryle-ai, good catch. Confirmed the masking mechanism: setupNodeParameters early-returns when getNodeParams finds >= 1 existing node, so in the normal nightly flow (delegated-staking runs first and seeds params) token-lock never reaches the hardcoded reads. The fallthrough only bites if delegated-staking fails to seed, and it would land on the old Euclid paths that the nightly job no longer stages (keys now live under NODE_KEYS_DIR=docker/config/local-test-keys/<index>).

Went with your first option (share the resolver rather than document the dependence): 650787e30 routes both node keys in token-lock-replacement-edge-cases.js through resolveNodeKeyPath('genesis-node', 0) / resolveNodeKeyPath('validator-1', 1), so both sibling scripts resolve keys identically and the ordering dependence is no longer load-bearing. Kept your improved error message as the backstop.

Same push also re-adds the snapshot-streaming apply_ss_patch refactor + scfee compatibility patch; the E2E build depends on them so they can't be split out of this PR. Note the current red run is unrelated: the Build JARs step failed on a transient DNS error resolving archive.ubuntu.com during the docker apt install, so E2E was skipped. Re-running.

docker/snapshot-streaming/build-snapshot-streaming.sh
```diff
@@ -39,31 +39,48 @@ else
...
-      echo "Local SharedConfig still has forkInfoStorage — skipping snapshot-streaming patch"
-    else
+
+  # apply_ss_patch <patch_file> <skip_marker> <marker_file> <label>
+  # Skips when <marker_file> contains <skip_marker> (local tessellation already has the
+  # newer API, so the testing-branch source already matches and must not be down-adapted).
+  apply_ss_patch() {
+    local patch_file="$1" skip_marker="$2" marker_file="$3" label="$4"
+    [ -f "$patch_file" ] && [ -s "$patch_file" ] || return 0
+    if [ -f "$marker_file" ] && grep -q "$skip_marker" "$marker_file"; then
```
ryle-ai, line 58: This gate uses a field name in `types.scala` as a proxy for which `GlobalSnapshotStateChannelEventsProcessor.make` signature the local tessellation exposes. Since both patches edit the same make-call hunk but key off different, independent symbols, they can desync from the real make arity on any tree where `forkInfoStorage` removal and `scFeeBalanceFromContext` addition aren't in lockstep (both-present or neither-present), producing a source that doesn't match `make` and failing as an unhelpful compile error rather than here. Consider grepping the actual signature in `GlobalSnapshotStateChannelEventsProcessor.scala` (e.g. `fieldsAddedOrdinals: FieldsAddedOrdinals` / the `environment` param) so the guard tracks the arity it's actually adapting to. Non-blocking — it's correct for develop and the intended older trees today; this is about robustness as more patches get stacked on the generalized helper.

ryle-ai (commented):
The key-resolution change is sound, and the tip now routes `token-lock-replacement-edge-cases.js` through `resolveNodeKeyPath` as well, so both nightly scripts honor the job-level `NODE_KEYS_DIR` (this closes the cross-script gap noted in the earlier review, which was written before that file was updated).

One non-blocking observation on the `build-snapshot-streaming.sh` refactor / new scfee patch, which the prior review didn't cover. Both `apply_ss_patch` invocations rewrite the *same* `GlobalSnapshotStateChannelEventsProcessor.make` call hunk, but each is gated on an *independent* symbol grepped from `types.scala` (`forkInfoStorage: ForkInfoStorageConfig` vs. `scFeeBalanceFromContext`). Correctness therefore depends on those two field-presence signals staying perfectly correlated with the actual `make` arity across every tessellation tree the nightly builds against. On develop this is fine — `forkInfoStorage` is gone so the fork patch applies (B→C form, matching develop's `mptStore, fieldsAddedOrdinals, environment`), and the scfee patch is skipped. But an intermediate tree where the two decouple (e.g. `scFeeBalanceFromContext` added while `forkInfoStorage` not yet removed) would skip both patches, leave the source in the pre-resolved-ordinal form, and mismatch the local `make` — surfacing only as an opaque sbt compile error deep in the build. See the inline note for a more direct guard.

scasplte2 (approved):

ryle-goehausen (approved):

PR #1530, opened 2026-06-30
docs(adr): consensus liveness and MPT alignments

scasplte2:


scasplte2:
closing in favor of #1531

PR #1531, opened 2026-06-30
chore: update ADRs, dedup function defs

scasplte2:
Adding ADRs to align the latest consensus liveness and MPT with the application code. Also addressed a small code duplication surfaced by @marcus-girolneto latest review


scasplte2:
The failing `E2E - delegated-staking` job here is a pre-existing flaky test, not a regression from this PR (which touches only ADRs/docs, a behavior-identical cert-builder refactor, and comments -- no reward/stake/balance code).

Root cause: the withdrawal balance assertion used a tight equality (`initialBalance + originalStake.totalBalance` within +-16 ulps), but the test wallet accrues delegated rewards during the multi-ordinal withdrawal-delay window (rewards only ever add), so the balance non-deterministically sits above expected. The same commit failed once (~1e14 datum over) and passed on re-run, confirming a timing race.

Fix is in #1534 (lower-bound assertion + retry, tolerating accrued-reward drift, mirroring the existing `assertBalanceChangeAfterTokenLock`). Once #1534 lands or this job is re-run, it should go green.

docs/adr/0016-consensus-determinism-source-of-truth.md
```diff
@@ -0,0 +1,40 @@
...
+Any value that either (a) enters a hashed consensus artifact, or (b) drives committee / quorum / leader derivation, MUST be a pure function of consensus-agreed inputs:
+
+- the prior signed outcome,
+- the frozen round-start committee,
+- signed declarations received in the round, or
+- a shared deterministic anchor that is *computed once by the leader, stored in the outcome, and re-read by followers* (never recomputed locally).
+
+Such a value must NEVER be derived from local observation, local counters or timers, wall-clock reads, live cluster membership, or per-node sidecar state.
+
+Fields that are inherently locally-divergent (per-peer scores, `recentRoundEndTimes`, per-peer maps) are split out as **operational** fields: emitted empty / `None` in the signed bytes and rebuilt round-by-round from the signed evidence window. Fields that must ride the proposal bytes are **proposal-critical** and obey the rule above.
```
ryle-ai, line 31: This lists `recentRoundEndTimes` flatly among the operational fields that are "emitted empty / `None` in the signed bytes and rebuilt round-by-round from the signed evidence window." That's accurate for the `ControllerEvidence` / `peerHistory` copy (`ControllerEvidenceDerivation` emits it as `None`), but the *outcome-level* `recentRoundEndTimes` is precisely NOT operational: per ADR-0020 the leader computes the median once, records it into the outcome, and followers re-read it (the "shared deterministic anchor" case (d) in the Decision above), and the code confirms it is consensus-agreed and read forward (`ConsensusEndTime` reads `parentEndTime` from the prior outcome's `recentRoundEndTimes`). So the same field name serves two roles, and listing it here without qualification reads as contradicting ADR-0020. Since the whole point of this ADR is to make contributors classify a field as proposal-critical vs operational, this dual-role field is exactly the case worth disambiguating -- e.g. note that the *peerHistory* copy is operational/`None` while the *outcome anchor* is the re-read deterministic-anchor case. Minor, non-blocking.

ryle-ai (commented):
Hygiene PR: ADR backfill, comment corrections, and a dedup of the cert-vote aggregation logic. The code change is the extraction of `CertVoteAggregation.poolSignersAndDivergence`, which is byte-for-byte faithful to the two inline implementations it replaces (group-by-signer within the witness pool, divergent-QC detection over all of a signer's votes before collapse, then per-signer representative by highest-QC-view with the total-`Signed` tiebreak). The implicit `Ordering[Signed[A]]` resolves at the unchanged call sites in both builders, and both `TimeoutVote`/`ViewChangeVote` carry `highestKnownQc: Option[ProposalQC]`, so the generic substitution type-checks and preserves behavior. The `0.67 -> 0.6666666666666666` comment correction in `ConsensusStateAdvancer` matches the shipped `dag-l0.conf` value.

One documentation-precision note on ADR-0016 (inline). Otherwise nothing blocking; the cert-builder change is behavior-identical and the ADRs are a useful capture of the v4.1.0 design history.

docs/adr/0027-operating-invariants.md
```diff
@@ -0,0 +1,28 @@
...
+
+2. **All networks deploy via full cold restart** (testnet / integrationnet / mainnet), all-or-nothing. There are no rolling upgrades and no version-skew tolerance. Every schema bump in this line assumes a coordinated cold restart; do not design for in-place or staggered upgrades.
+
+3. **Hypergraph and metagraph scale oppositely.** Global L0 / dag-l0 is the hypergraph: hundreds of nodes, supermajority quorum (`core = 9` is right for gl0). Metagraphs (currency-l0 / state channels) are few-node and run unanimity (`quorum = 1.0`). Per-environment config conditioned on this includes `clusterFloorActive` (ADR-0021) and B2 being inert on currency-l0 (ADR-0022). Do not transplant a hypergraph node-count assumption onto a metagraph or vice versa.
```
ryle-ai, line 21: `core = 9` is the *integrationnet* value, not the mainnet/hypergraph one. `dag-l0.conf` ships `core-committee-size` of `mainnet: 15`, `integrationnet: 9`, `testnet/dev: 3` -- and the conf's own comment (dag-l0.conf:155) reads "peers -> Core 15, integrationnet ~10 -> Core 9". Since this invariant is describing the "hundreds of nodes" hypergraph case, the example should be `core = 15` (or cite the per-env spread) so a reader doesn't come away thinking gl0 mainnet runs a Core of 9. Minor, non-blocking.

ryle-ai (commented):
Hygiene PR: ADR backfill (0016-0027), supersession headers on the older ADRs, comment corrections, and the dedup of the cert-vote aggregation logic. The `CertVoteAggregation.poolSignersAndDivergence` extraction is behavior-identical to the two inline implementations it replaces (group-by-signer within the witness pool, divergent-QC detection over all of a signer's votes before collapse, then per-signer representative by highest-QC-view with the total-`Signed` tiebreak). The implicit `Ordering[Signed[A]]` resolves at both call sites (both builders already require it for the downstream `SortedSet.empty[Signed[...]] ++ ...`), and both `TimeoutVote`/`ViewChangeVote` carry `highestKnownQc: Option[ProposalQC]`, so the generic substitution type-checks and preserves behavior.

One factual nit in ADR-0027 (inline). I also agree with the earlier ADR-0016 note re `recentRoundEndTimes` -- the outcome-anchor copy is the deterministic-anchor case, not operational, so listing the field name flatly under operational reads as contradicting ADR-0020. Both are documentation-precision points; nothing blocking, and the cert-builder change is a clean refactor.

ryle-goehausen (approved):

PR #1535, opened 2026-07-01
fix: reanchor develop release testnet

scasplte2:
squash broke the history merging #1523 so this will push the commit history to `develop` so proper tagging can continue via the [release process](https://github.com/Constellation-Labs/tessellation/blob/repair/reanchor-develop-release-testnet/docs/release/BRANCHING_AND_RELEASE_WORKFLOW.md)

No files on `develop` are modified


ryle-ai (commented):
This is an ancestry-recording ("ours"-style) reanchor merge, not a content change. The branch tip `e98933b8c` is a merge of `c268695e3` (current develop) and `d4e506b4b` (release/testnet tip), resolving to a tree byte-identical to develop (`32f0557…`) — hence the empty +0/-0 diff. Nothing in the working tree changes, and the tip fix it anchors is already present on develop (`d4e506b4b` and develop's `ab5d4c628` share patch-id `8000a000…`). Purpose looks correct: link release/testnet ancestry after the v4.1.0 squash so future merges don't re-surface already-integrated commits.

One caveat worth confirming, since it's the standard footgun of this kind of merge: resolving to develop's tree permanently marks `d4e506b4b` (and its release/testnet history) as merged while discarding whatever that side has that develop doesn't. A direct tree comparison (`git diff develop d4e506b4b`) shows ~1868 lines across 181 files that exist on release/testnet but not on develop — after this merge those will never flow into develop via a future release/testnet→develop merge without an explicit cherry-pick. From the file list this appears to be intentionally-superseded content (e.g. removed `nightly/terraform/*` infra and retired test suites), so this is almost certainly the intended outcome. Just worth a maintainer eyeballing that buried delta to make sure nothing on the release line still needs to reach develop independently.

No blocking issues.

ryle-goehausen (approved):

ryle-goehausen (approved):

PR #1536, opened 2026-07-09
docs: v4.1.0-rc.0 community and release notes

scasplte2:


docs/release/v4.1.0-testnet.md
```diff
@@ -0,0 +1,295 @@
...
+Two distinct config surfaces matter for this release.
+
+### Consensus-critical config (must match cluster-wide)
+
+These are folded into the deterministic config hash. A divergent value is rejected at handshake.
+
+| Config Key | Testnet | IntegrationNet | MainNet | Notes |
+|------------|---------|----------------|---------|-------|
+| `core-committee-size` | 3 | 9 | 15 | Quorum-bearing core size (per environment scale) |
+| `quorum-threshold-fraction` | 2/3 | 1.0 | 1.0 | Testnet runs supermajority; compiled default is unanimity |
```
ryle-ai: `quorum-threshold-fraction` is not per-environment. `dag-l0.conf` sets `quorum-threshold-fraction = 0.6666666666666666` as a flat value in the consensus block, which overrides the compiled default (`config/types.scala:392`, `= 1.0`) for all environments. So IntegrationNet and MainNet run **2/3**, not `1.0`/unanimity. The "compiled default is unanimity" note is misleading because the default is overridden here.

This also contradicts the "IntegrationNet deploy consideration" section below, which calls IntegrationNet "a 6-of-9 supermajority core" and says losing 4 of 9 core peers wedges consensus — `ceil(9 * 2/3) = 6`, which is the 2/3 reading, not unanimity (which would be 9-of-9). Make this row `2/3 | 2/3 | 2/3` and drop the note.

ryle-ai (dismissed):
Docs-only release-notes/announcement PR. I spot-checked the technical claims against `origin/develop` and all but one hold up (schema version 33, Scala 2.13.18 / Java 21, the `core-committee-size` 3/9/15 and `quorum-shrink-activation-views` 10/0/0 rows, all four `fields-added-ordinals` gate values, the `CL_MPT_VERIFY_INCREMENTAL` / `CL_RAISE_ON_FOLLOWER_DIVERGENCE` toggles, and the referenced doc files all match).

One consensus-critical value is stated incorrectly, and the document contradicts itself on it (detail inline on the config table):

- The **consensus-critical config** table gives `quorum-threshold-fraction` as `2/3` on Testnet but **`1.0` (unanimity) on IntegrationNet and MainNet**, with the note "compiled default is unanimity." That note is the root of the error: `dag-l0.conf` sets `quorum-threshold-fraction = 0.6666666666666666` as a *single flat value* in the consensus block (not per-environment, unlike `core-committee-size` right below it), which overrides the `= 1.0` compiled default (`node-shared/.../config/types.scala:392`) for **every** environment. So IntegrationNet and MainNet run 2/3 supermajority, not unanimity.
- This is not just a table typo — the doc's own **"IntegrationNet deploy consideration"** section describes IntegrationNet as "a **6-of-9 supermajority** core" and warns "a loss of **4 or more** of the 9 core peers can wedge consensus." `ceil(9 * 2/3) = 6`, and losing 4 leaves 5 < 6 → wedge; that math only works at 2/3. At unanimity the core would be 9-of-9 and a single silent peer would wedge it. The prose assumes 2/3 while the table asserts 1.0.

Because this table is framed as the consensus-critical values operators must match cluster-wide, misstating IntegrationNet/MainNet as unanimity materially misrepresents the fault tolerance of the launch config (zero crash tolerance vs. f≈(N-1)/3). Fix the row to `2/3` across all three columns and drop the "compiled default is unanimity" note (the default is overridden in HOCON).

Minor, non-blocking: the doc still carries `<!-- TODO(operators) ... -->` markers (MainNet incident specifics, fork-info chain-recovery PR/branch name) — expected for a draft, just flagging they're unresolved before external publication.

docs/release/v4.1.0-testnet.md
scasplte2: Good catch, fixed in a184e4b8f.

Confirmed against `dag-l0.conf`: `quorum-threshold-fraction = 0.6666666666666666` is a flat value in the `consensus {` block (unlike `core-committee-size` and `quorum-shrink-activation-views` right below it, which are per-environment maps), so it overrides the compiled `1.0` default for every environment. Global L0 runs 2/3 on Testnet, IntegrationNet, and MainNet. The `1.0`/unanimity I had listed was the `currency-l0.conf` (metagraph) value.

Updated the row to `2/3 | 2/3 | 2/3`, dropped the "compiled default is unanimity" note, and clarified that metagraph currency-L0 uses unanimity. The table now agrees with the 6-of-9 IntegrationNet reasoning below.

The two remaining `TODO(operators)` markers (MainNet incident specifics, fork-info PR/branch name) are pending facts and will be resolved before external publication.

marcus-girolneto (approved):

PR #1538, opened 2026-07-11
feat: fee transaction lookup in L0NodeContext

scasplte2:
## Summary

Adds snapshot-scoped fee-transaction lookup to currency-L0's `L0NodeContext`, allowing a metagraph data application's `combine` function to associate a data update with its sibling fee transaction.

The implementation is immutable and call-scoped: live acceptance and rollback replay each wrap their existing node context with the fee map for the snapshot being processed. There is no process-wide `Ref`, and all other context methods continue to delegate to the original context.

## Replay-safety invariant

Live acceptance and replay expose the same fee population to `combine`:

- Blocks that fail data/fee validation never enter the live fee map.
- If `combine` rejects a validation-passing block, acceptance removes it and recomputes from the original parent state with the strictly smaller candidate set and fee map.
- The final successful pass therefore uses fees from exactly the blocks stored in `DataApplicationPart`.
- Replay rebuilds its map from those stored blocks before invoking `combine`.

The retry terminates because each failed pass removes at least one candidate block.

## Changes

- Adds `getSnapshotFeeTransactions` to `L0NodeContext` and a scoped delegating wrapper.
- Supplies an empty map from the normal currency-L0 context.
- Makes live data-application acceptance converge on an accepted-block/fee-map fixed point.
- Reconstructs the same map from stored blocks during rollback replay while preserving the historical predecessor context.
- Updates the generated metagraph template to demonstrate a fee-derived calculated-state value.
- Expands E2E coverage for no-fee, adequate-fee, insufficient-fee, and missing-fee updates.
- Uses the current Brotli fee-signing implementation in all fee scenarios and prevents missing/`NaN` calculated-state values from false-passing.

## Compatibility

- No snapshot, state-proof, or consensus-message schema changes.
- No activation ordinal or deterministic-config change.
- This adds one method to the SDK's `L0NodeContext` interface. Normal metagraph applications consume that interface; custom test/runtime implementations must supply the new method (usually `Map.empty.pure[F]`).

## Verification

- `nodeShared/compile`
- `currencyL0/compile`
- Focused node-shared suites: 4 passed, 0 failed
  - live acceptance recomputes after a combine failure with only surviving fees
  - replay reconstructs fees from stored accepted blocks
  - existing predecessor-scoping and non-contiguous replay guards
- Generated metagraph template compiled against the locally published SDK
- JavaScript syntax checks for all three data-transaction scripts
- `bash -n docker/bin/compose-runner.sh`

Full CI runs on this push.


scasplte2:
@ryle-ai please review

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/currency/DataApplicationSnapshotAcceptanceManager.scala
```diff
@@ -124,165 +126,180 @@ object DataApplicationSnapshotAcceptanceManager {
...
-            case Some(hashFn) if validatedBlocks.nonEmpty =>
-              validatedBlocks.flatMap { block =>
-                getDataUpdates(block.value.dataTransactions.toList)
-              }.traverse { signedUpdate =>
-                hashFn(signedUpdate.value)
+          updateHashes <- OptionT.liftF {
+            val updates = validatedBlocks.flatMap { block =>
+              getDataUpdates(block.value.dataTransactions.toList)
+            }
+            if (updates.nonEmpty)
```
ryle-ai: `updateHashes` is now computed unconditionally whenever `validatedBlocks` produce updates. On develop this was gated on `service.hashDataUpdate`, whose base default is `None` (package.scala:245/340) — so a metagraph that does not override `hashDataUpdate` produced `updateHashes = None`. After this change it produces `Some(...)`.

`updateHashes` is a field of `DataApplicationPart`, which is part of the signed/hashed `CurrencyIncrementalSnapshot`. Flipping it `None -> Some` changes the snapshot hash for any such metagraph, and it is applied with no ordinal gate:
- rolling upgrade: old nodes (`None`) and new nodes (`Some`) build different proposal artifacts for the same ordinal -> consensus divergence;
- replay / validation: `CurrencySnapshotValidator.recreateFn` recomputes `Some` against already-persisted snapshots that stored `None` -> mismatch.

Template-based metagraphs are unaffected — their old `Some(serializeUpdate(_).map(Hash.fromBytes))` equals the new `serializeUpdate(_).flatMap(Hasher[F].hashBytes)` because `hashBytes == Hash.fromBytesForSync`. But please confirm no deployed metagraph relies on the `hashDataUpdate = None` default, or gate this by snapshot ordinal the way other consensus-format changes are gated in this repo.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/currency/DataApplicationSnapshotAcceptanceManager.scala
```diff
@@ -124,165 +126,180 @@ object DataApplicationSnapshotAcceptanceManager {
...
+                )
               )
-            )
-          }
+            }
+        }
       }
+
+      // Set fee map at snapshot start, run processing, clear at end via guarantee
+      (snapshotFeeTransactionsRef.set(feeMap) >> snapshotProcessing)
```
ryle-ai: This uses a process-wide mutable `Ref` as a side-channel for the fee map: set at the start of `accept()`, cleared via `.guarantee`. The single manager instance (one `Ref`, wired in Services.scala) is shared by every path that calls `accept()`:
- `CurrencySnapshotCreator.createProposalArtifact` (consensus advancer, CurrencySnapshotConsensusStateAdvancer.scala:1846),
- `StateChannelSnapshotService.consume` -> `consumeSignedMajorityArtifact` -> `accept` (advancer:1935),
- `CurrencySnapshotValidator.recreateFn` -> `createProposalArtifact`.

Correctness hinges on `accept()` never running concurrently or re-entrantly across these consumers. If two overlap, one's `set(feeMap)` — or the other's `guarantee(set(Map.empty))` — clobbers the map mid-`combine`, so a metagraph's `getSnapshotFeeTransactions` returns the wrong or empty map non-deterministically. Can you confirm these paths are serialized on a single node? If not, thread the fee map through the call (or scope it per invocation) rather than via a shared global `Ref`.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/currency/DataApplicationSnapshotAcceptanceManager.scala
```diff
@@ -124,165 +126,180 @@ object DataApplicationSnapshotAcceptanceManager {
...
-              case Left(err) => logger.warn(err)("Cannot deserialize custom state")
-              case Right(_)  => Applicative[F].unit
+      // Collect ALL fee transactions from ALL blocks upfront at snapshot level
+      val allFeeTransactions: List[Signed[FeeTransaction]] = dataBlocks.flatMap { block =>
+        getFeeTransactions(block.value.dataTransactions.toList)
+      }
+
+      // Build snapshot-wide fee map (keyed by dataUpdateRef hash)
+      val feeMap: Map[Hash, Signed[FeeTransaction]] =
+        allFeeTransactions.map(ft => ft.value.dataUpdateRef -> ft).toMap
```
ryle-ai: `feeMap` is built from all `dataBlocks` up front — before the validation fold that decides which blocks are accepted — so it includes fee transactions from blocks that are subsequently rejected. `.toMap` also silently collapses entries sharing a `dataUpdateRef` (last one wins). `combine` can therefore look up a fee for a rejected or duplicated update. If that isn't intended, build the map from the validated/accepted blocks and/or detect `dataUpdateRef` collisions instead of dropping them silently.

ryle-ai (dismissed):
Most of this is a clean refactor, but two things are worth confirming before merge (one consensus-format change flagged inline as blocking) plus two smaller notes.

**Scope / description.** The PR is titled "fee transaction lookup in L0NodeContext" and has an empty description, but it also bundles: the `docker/entrypoint.sh` L0-readiness wait, the `CurrencySnapshotProcessor` `mptStore.sync` calls, and a rewrite of the `updateHashes` computation. The last one changes consensus-relevant output and is very easy to miss under this title — please call it out in the description.

**Downstream API break (non-blocking).** `L0NodeContext` gains a new abstract `getSnapshotFeeTransactions` with no default (`trait L0NodeContext[F[_]]` has no `Applicative[F]` bound, so a default isn't free). Every third-party metagraph that implements `L0NodeContext` (and the test mocks in amm-metagraph / ded) will fail to compile on the tessellation bump. That may be acceptable for a coordinated bump, but flag it in the release notes.

ryle-ai (commented):
ryle's review already covers the substantive items — the consensus-format change to `updateHashes` (blocking), the process-wide `Ref` concurrency question, the pre-validation `feeMap` / silent `toMap` collision, and the new abstract `getSnapshotFeeTransactions` downstream break. I agree with all of those and have nothing to add there.

One additional item for the same release-notes callout ryle requested: this PR doesn't just *add* an abstract method, it also *removes* `hashDataUpdate` and `extractFees` from the data-application trait hierarchy — a second, independent source-compat break on the bump that ryle's note (scoped to `L0NodeContext` implementers) doesn't cover. See inline.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/dataApplication/DataApplicationTraverse.scala
```diff
@@ -116,6 +116,7 @@ object DataApplicationTraverse {
               def securityProvider: SecurityProvider[F] = context.securityProvider
               def getCurrencyId: F[CurrencyId] = context.getCurrencyId
               def getMetagraphL0Seedlist: Option[Set[SeedlistEntry]] = context.getMetagraphL0Seedlist
+              def getSnapshotFeeTransactions: F[Map[Hash, Signed[FeeTransaction]]] = context.getSnapshotFeeTransactions
```
ryle-ai: The fee map is only ever populated inside `DataApplicationSnapshotAcceptanceManager.accept` (`snapshotFeeTransactionsRef.set(feeMap) >> ...`). But the rollback-replay path here doesn't go through `accept` — it calls `dataApplication.combine(state, dataUpdates)(replayContext)` directly (below, ~line 154) and never sets that `Ref`. `replayScopedContext` delegates `getSnapshotFeeTransactions` straight to the live `context`, whose `Ref` during replay is either empty (the live `accept` cleared it via `.guarantee`) or holds an unrelated concurrent snapshot's map.

So a metagraph whose `combine` reads `getSnapshotFeeTransactions` sees the real snapshot fee map at consensus time but an empty/stale map on replay, and its recomputed calculated state diverges from what was persisted — the exact byte-exact-replay invariant #1520 was built to preserve. Note the documented out-of-scope replay limitation above (the block ending ~line 93) enumerates only `getLastCurrencySnapshotCombined` and the `getTokenUnlocks`/`balances` helpers; this new accessor silently joins that not-byte-exact set without being called out, and unlike those it *looks* correct on the live path because `accept` populates it there.

No in-tree metagraph reads it yet, so nothing breaks today. But if the intent is for fees to be consumable from `combine`, the fee map has to be threaded into the replay context too (reconstruct it from each replayed snapshot's blocks); otherwise the limitation comment should at least name `getSnapshotFeeTransactions` explicitly so a metagraph author knows this accessor is not replay-safe.

ryle-ai (commented):
The current diff is narrower than the version the CHANGES_REQUESTED review was written against: the `updateHashes` rewrite is gone (lines 497-508 are develop's `service.hashDataUpdate match` gating, just re-indented into the new `snapshotProcessing` nesting) and `package.scala` only *adds* `getSnapshotFeeTransactions` — it no longer removes `hashDataUpdate`/`extractFees`. So the blocking consensus-format concern (@270) and the trait-removal source break both appear reverted; what's left is the fee-map plumbing.

I agree with the remaining notes on that plumbing and won't restate them: the pre-validation `feeMap` built from all blocks plus the silent `.toMap` collision (@136), the process-wide `Ref` used as a side-channel with correctness depending on `accept()` never overlapping (@302), and the new no-default abstract `getSnapshotFeeTransactions` as a downstream source break for external `L0NodeContext` implementers.

One distinct item those don't cover is the **replay path** — see inline on `DataApplicationTraverse`.

ryle-ai (commented):
The core plumbing is well covered by ryle's review and I agree with all of it: the process-wide `Ref` side-channel and its serialization requirement (@302), the pre-validation `feeMap` built from all blocks + silent `.toMap` collision (@136), the new no-default abstract `getSnapshotFeeTransactions` as a downstream source break, and — most importantly here — the replay path (@119), where `DataApplicationTraverse`'s `replayScopedContext` just delegates `getSnapshotFeeTransactions` to the live context `Ref` (empty/stale during replay) instead of reconstructing the fee map from each replayed snapshot's blocks. That last one is unaddressed on the current tip despite the `d9d438a0d "...implement getSnapshotFeeTransactions in traverse replay contexts"` commit — the commit only adds the live-context delegation ryle flagged, it does not thread a per-snapshot fee map.

One distinct addition on the new material ryle didn't review — the **project template**. This PR now makes the replay concern concrete: the template's `combine` reads `getSnapshotFeeTransactions` and writes the result into `DeviceCalculatedState.feesPaid`, which flows into `hashCalculatedState` → `calculatedStateProof`. So for any metagraph scaffolded from this template that pays fees, rollback replay recomputes `feesPaid = 0` (empty fee map on the replay path) and the calculated-state proof diverges from what was persisted — the exact byte-exact-replay invariant #1520 was built to preserve. ryle's @119 note ("no in-tree metagraph reads it yet") is technically true for runtime code, but the canonical scaffold now ships this pattern as the reference example, which is how it will reach real metagraphs.

Compounding this: the four new e2e tests (`data-with-fee`, `data-without-fee`, insufficient/missing) only exercise the live accept path. None exercise rollback replay, so CI will stay green while the replay-divergence pattern ships in the template. If fees are meant to be consumable from `combine`, the fee map needs to be threaded into the replay context (reconstructed from each replayed snapshot's blocks) before the template demonstrates reading it into hashed state; otherwise the template should at least carry a comment that `getSnapshotFeeTransactions` is not replay-safe and calculated state derived from it will not survive rollback. Non-blocking, but worth resolving before this becomes the pattern everyone copies.

.github/templates/metagraphs/project_template/modules/shared_data/src/main/scala/com/my/project_template/shared_data/LifecycleSharedFunctions.scala
```diff
@@ -33,11 +36,21 @@ object LifecycleSharedFunctions {
...
     } else {
-      Async[F].delay(updates.foldLeft(newState) { (acc, signedUpdate) =>
-        combineUpdateUsage(signedUpdate, acc)
-      })
+      updates.foldLeftM(newState) { (acc, signedUpdate) =>
+        // A fee is submitted as a sibling transaction keyed by dataUpdateRef, which the fee estimate
+        // route computes as EstimatedFee.getUpdateHash = Hash.fromBytesForSync(serializeUpdate(update)).
+        // The only way combine can observe that fee is L0NodeContext.getSnapshotFeeTransactions, whose
+        // result is threaded in here as `feeTransactions`. Recompute the identical key to match it.
+        serializeUpdate(signedUpdate.value).flatMap(Hash.fromBytesForSync[F]).map { updateHash =>
```
ryle-ai, line 46: This template `combine` reads the fee via `getSnapshotFeeTransactions` (threaded in as `feeTransactions`) and folds it into `feesPaid`, which becomes part of the *hashed* calculated state (`DeviceCalculatedState.feesPaid` → `hashCalculatedState` → `calculatedStateProof`). On the live accept path that's fine — the acceptance manager sets the `Ref` before `combine` runs. But on the rollback-replay path (`DataApplicationTraverse`, see ryle's @119), `combine` runs with an empty/stale fee map, so `feePaid` resolves to 0 and the recomputed calculated-state proof diverges from the persisted one — breaking the byte-exact replay invariant from #1520.

Because this is the reference scaffold, shipping it this way propagates a replay-unsafe pattern to every fee-using metagraph. Either thread the fee map into the replay context (reconstruct from each replayed snapshot's blocks) so this lookup is replay-safe, or add a comment here that a value derived from `getSnapshotFeeTransactions` must not be written into calculated state until that's done. The new e2e tests only cover the live path, so this won't show up in CI.

scasplte2:
Thanks @ryle-ai. Addressed in a79b73922 by making the fee map context-scoped instead of a process-wide Ref, which resolves the replay gap and the side-channel together:

- **Replay (your inline on DataApplicationTraverse):** applyCache now reconstructs the fee map from each replayed snapshot's own blocks and serves it through replayScopedContext (getSnapshotFeeTransactions no longer delegates to the live context). combine now sees the same fees on replay as at consensus time, so this accessor is byte-exact under replay.
- **Process-wide Ref (@302):** removed. L0NodeContextOps.withSnapshotFeeTransactions wraps the node context per snapshot; accept() uses that and no longer set/clears a Ref, so the accept()-overlap hazard is gone.
- **Silent .toMap collision (@136):** FeeTransaction.buildFeeMap is now the single construction point (used by both accept and replay). It keeps last-wins but logs colliding dataUpdateRefs instead of dropping silently.
- **No-default abstract getSnapshotFeeTransactions:** still abstract. A trait default can't return F[_] without an Applicative the trait doesn't carry; in practice only the framework and tests implement L0NodeContext (metagraphs get it from L0NodeContext.make), so the blast radius is limited. Flagging as a known API addition.

The @270 consensus-format item and the hashDataUpdate/extractFees trait removal are not in this PR (dropped in the scope-down). currencyL0/compile + nodeShared/Test/compile are clean and the lint gate passes; the data-transactions e2e (adequate / insufficient / missing fee) passed green on the pre-refactor commit and re-runs on this one.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/currency/DataApplicationSnapshotAcceptanceManager.scala
```diff
@@ -122,166 +122,179 @@ object DataApplicationSnapshotAcceptanceManager {
       lastOrdinal: SnapshotOrdinal,
       currentOrdinal: SnapshotOrdinal
     ): F[Option[DataApplicationAcceptanceResult]] = {
-      implicit val context: L0NodeContext[F] = nodeContext
+      // Collect ALL fee transactions from ALL blocks upfront at snapshot level
+      val allFeeTransactions: List[Signed[FeeTransaction]] = dataBlocks.flatMap { block =>
```
ryle-ai: `allFeeTransactions` is built from every block in `dataBlocks` — i.e. *before* the validation fold at `blocksToProcess = dataBlocks.sortBy(_.roundId).distinctBy(_.value.roundId)` selects/dedups them. So the live `feeMap` served to `combine` includes fee transactions from (a) blocks that are subsequently rejected and (b) duplicate-roundId blocks dropped by `distinctBy`, and its last-wins `.toMap` is resolved over the *unsorted* `dataBlocks` order.

The replay path in `DataApplicationTraverse` reconstructs its `feeMap` from the snapshot's stored blocks, which are only `validatedBlocks` (accepted, deduped) serialized in sorted-roundId order. So live and replay build the map from different inputs.

For updates with no colliding `dataUpdateRef` the two maps agree and the accessor is byte-exact — but when a fee tx from a rejected/dropped block (or a second accepted block) shares a `dataUpdateRef` with an accepted update, last-wins can resolve to a different fee at consensus time than on replay. `combine` does `feeTransactions.get(updateHash)`, so `feesPaid` and hence `calculatedStateProof` diverge on rollback replay — breaking the byte-exact invariant this refactor claims to preserve.

To actually make `getSnapshotFeeTransactions` replay-safe, build the live map from the same accepted/validated blocks (in the same order) that replay reconstructs from, rather than from raw `dataBlocks` up front. That also resolves ryle's earlier @136 point (fees from rejected/duplicated blocks leaking into the map) in one move.

ryle-ai (commented):
The a79b73922 refactor addresses the process-wide `Ref` (now context-scoped via `L0NodeContextOps.withSnapshotFeeTransactions`) and threads a fee map into the replay context, which is the right shape. But the stated goal — "combine now sees the same fees on replay as at consensus time, so this accessor is byte-exact under replay" — is not actually met, because the live path and the replay path build the fee map from **different block sets and in different order**. See inline on `DataApplicationSnapshotAcceptanceManager.scala`.

Short version: on the live path the map is built from `allFeeTransactions = dataBlocks.flatMap(...)` (every block handed to `accept`, *before* validation), whereas on replay it is rebuilt from the snapshot's *stored* blocks, which are only the accepted/deduped `validatedBlocks`. `buildFeeMap`'s last-wins `.toMap` is therefore resolved over a superset (and in unsorted `dataBlocks` order) at consensus time, and over a subset (in sorted-roundId order) on replay. Whenever a `dataUpdateRef` collides across those sets, `combine` reads a different fee on replay than at consensus time, so `feesPaid` — and thus `calculatedStateProof` — diverges. That is the exact byte-exact-replay invariant #1520 was built to preserve, and it is the concrete failure mode the template's `feesPaid`-in-hashed-state pattern would surface.

Non-blocking: no in-tree runtime metagraph reads `getSnapshotFeeTransactions` yet, and it requires a colliding/duplicate/rejected fee ref to trigger. But since this PR ships the read-into-hashed-state pattern as the reference template and explicitly claims byte-exactness, it's worth closing the gap (or narrowing the claim) before it becomes the copied pattern. This is the same construction that ryle's @136 asked for — building from the validated/accepted blocks — now with an additional byte-exactness reason to do so.

The rest of the scope-down looks consistent with the prior review: `updateHashes` is back on develop's `service.hashDataUpdate` gating and the `hashDataUpdate`/`extractFees` trait removals are gone, so those consensus-format concerns no longer apply here. The new e2e coverage (adequate / insufficient / missing) only exercises the live accept path, so none of this shows up in CI.

scasplte2:
Good catch @ryle-ai - correct, the previous refactor fixed the plumbing but not the map's contents. Fixed in 2ec905d78.

accept() now determines the accepted blocks up front (validation depends only on dataState, not on the combine fold's threaded state) and builds the fee map from exactly those - the same accepted, deduped blocks, in the same order, that get stored and that rollback replay reconstructs from. So live and replay now build the map from identical inputs. This also resolves your earlier @136 point in the same move: fees from rejected or duplicate-roundId blocks no longer leak into the live map.

The scoped fee context is passed explicitly to combine (not as an implicit) so it doesn't collide with the nodeContext implicit used for validation and serialization elsewhere in accept().

Per discussion, skipping a replay/collision regression test for now (no runtime metagraph reads the accessor yet and the e2e only exercises the live path). currencyL0/compile + nodeShared/Test/compile are clean and the scalafix/scalafmt gate passes.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/currency/DataApplicationSnapshotAcceptanceManager.scala
```diff
@@ -122,167 +122,199 @@ object DataApplicationSnapshotAcceptanceManager {
...
-            service.combine(oldState, List.empty).map { newState =>
-              (newState, oldFeeTxns, oldAcceptedBlocks, oldRejectedBlocks)
+          // Determine which blocks pass validation up front. Validation depends only on dataState
+          // (not on the combine fold's threaded state), so we can build the snapshot fee map from the
+          // SAME accepted, deduped blocks - in the same order - that get stored and that rollback
+          // replay reconstructs from. Building it from raw dataBlocks would let fees from rejected or
+          // duplicate-roundId blocks leak in, making getSnapshotFeeTransactions diverge between
+          // consensus time and replay and breaking byte-exact replay.
+          acceptedBlocks <- OptionT.liftF(
+            blocksToProcess.filterA { dataBlock =>
```
ryle-ai: This up-front pass runs `validateDataTransactionsL0` with no error handling, but the fold below (~line 245) wraps the *same* validation in a per-block `handleErrorWith` that rejects only the offending block and continues. `validateDataTransactionsL0` can raise — `getOrRaise(new RuntimeException("Could not get data updates"))` for a block that yields no updates, and the metagraph-supplied `validateData`/`serializeUpdate` can throw arbitrarily. Because `filterA` runs first, a single raising block now fails the whole pass and propagates to the outer `newDataState.value.handleErrorWith`, which freezes calculated state at the last data application and marks *every* block in the snapshot not-accepted. That is a behavioral regression relative to develop's graceful per-block degradation, and it affects all currency-L0 metagraphs regardless of whether they read fees. Suggest wrapping the validation here in the same handler (treat a raising block as not-accepted) so the accepted set matches the fold's outcome. It also means validation runs twice per block; reusing one result would avoid the extra cost.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/currency/DataApplicationSnapshotAcceptanceManager.scala
```diff
@@ -122,167 +122,199 @@ object DataApplicationSnapshotAcceptanceManager {
...
-                      val feeTransactions = getFeeTransactions(dataTransactionsAsList)
-
-                      for {
-                        _ <- logger.info(s"Block ${dataBlock.value.roundId} is valid")
-                        result <- service.combine(currentState, dataUpdates).map { newState =>
+          )
+
+          feeMap <- OptionT.liftF(
+            FeeTransaction.buildFeeMap[F](
+              acceptedBlocks.flatMap(block => getFeeTransactions(block.value.dataTransactions.toList)),
```
ryle-ai: The map is built from `acceptedBlocks` (blocks that pass validation), but the blocks actually stored (`validatedBlocks` = `accAcceptedBlocks` from the fold) are those that pass validation *and* whose `combine` does not throw — the fold's `handleErrorWith` drops a combine-throwing block from `accAcceptedBlocks`. So `acceptedBlocks` is a superset of the stored set, and replay rebuilds its fee map from the stored (validated) blocks. When a validation-passing block's `combine` throws at consensus time, its fee lands in the live map but not the replay map, so `combine` can read a different fee on a colliding `dataUpdateRef` under rollback replay. This is the same superset-vs-subset divergence as @126, now on the combine-error path, so the "exactly those ... that get stored" claim isn't quite met. Narrow (requires a combine failure plus a colliding ref) and there's no runtime consumer yet, but worth either narrowing the byte-exactness claim or building the map from the post-fold accepted set.

ryle-ai (commented):
The `2ec905d78` rework rebuilds the live fee map from `acceptedBlocks` (validation-passing, deduped, sorted) instead of raw `dataBlocks`, which does close ryle's @126 concern for the rejected/duplicate-roundId case. Two things introduced by the new up-front validation pass are worth confirming before this is the reference pattern. Both are on the live accept path in `DataApplicationSnapshotAcceptanceManager.scala` and both are edge/error-path only, so non-blocking.

1. The new `acceptedBlocks` `filterA` pass changes failure semantics for *every* currency-L0 metagraph (not just fee-users), because it runs `validateDataTransactionsL0` with no per-block `handleErrorWith` — unlike the fold below, which has always caught per-block exceptions and rejected only the offending block. `validateDataTransactionsL0` can raise (`getOrRaise(new RuntimeException("Could not get data updates"))` for a block with no updates, plus the metagraph-supplied `validateData`/`serializeUpdate`). Since `filterA` runs first, one raising block now aborts the entire pass and drops to the outer fallback, freezing calculated state and marking *all* blocks in the snapshot not-accepted.

2. The byte-exactness claim ("builds the fee map from exactly those ... that get stored") still has a residual gap on the combine-error path: `acceptedBlocks` = validation-passing, but the *stored* `validatedBlocks` = validation-passing **and** combine-not-throwing. When a validated block's `combine` throws (the fold's `handleErrorWith`), its fees are in the live map but absent from the replay map (which is rebuilt from stored blocks) — the same superset-vs-subset divergence @126 flagged, just moved from rejected/dup blocks to combine-throwing blocks.

Giving the up-front `filterA` the same per-block error handling as the fold (treat a raising block as not-accepted) resolves (1) and narrows (2) so the only remaining live/replay difference is the combine-throw case. Validation also now runs twice per block (once in `filterA`, once in the fold); reusing one outcome would avoid that.

scasplte2:
Right on both counts, thanks. Fixed in f8af99e41 by removing the up-front validation pass entirely rather than duplicating the fold's error handling:

1. (error-handling regression) The filterA is gone. Validation and its per-block handleErrorWith live only in the fold, unchanged from develop, so a raising block is rejected individually and the snapshot no longer aborts. This also removes the double-validation.

2. (residual byte-exactness) The fee map is now built from blocksToProcess (deduped, sorted - matching replay's order and dedup, an improvement over raw dataBlocks). I'm no longer claiming full byte-exactness: a fee on a validation-rejected or combine-throwing block is in the live map but not the replay map, so getSnapshotFeeTransactions can differ under rollback replay only on a colliding dataUpdateRef. Documented inline, and no in-tree runtime metagraph reads the accessor. Prioritizing leaving the battle-tested accept() fold untouched over closing that narrow edge.

currencyL0/compile + nodeShared/Test/compile clean; scalafix/scalafmt gate passes.

ryle-goehausen (dismissed):

scasplte2:
Addressed the replay/determinism review findings in `44cadb063`:

- Retained develop's `service.hashDataUpdate` gate, so the default remains `updateHashes = None`.
- Removed the process-wide fee `Ref`; fee maps are now immutable, invocation-scoped context wrappers.
- Deduplicates/sorts candidates first, validates each block once with the existing per-block failure isolation, and excludes validation failures from the fee map.
- If a validation-passing block fails in `combine`, removes it and recomputes from the original state with a strictly smaller candidate set. The final live fee map therefore contains exactly the stored blocks' fees.
- Detects/logs duplicate `dataUpdateRef` entries while preserving deterministic last-wins behavior.
- Replay reconstructs the fee map from each snapshot's stored blocks while retaining its ordinal-specific predecessor context.

Added focused regressions for both halves of the invariant: live combine-failure recomputation and replay reconstruction from stored accepted blocks. The generated metagraph template also compiles against the updated SDK.

.github/templates/metagraphs/project_template/modules/shared_data/src/main/scala/com/my/project_template/shared_data/types/Types.scala
```diff
@@ -29,7 +29,9 @@ object Types {
 
   @derive(decoder, encoder)
   case class DeviceCalculatedState(
-    usages: UsageUpdateInfo
+    usages: UsageUpdateInfo,
+    // Total fees looked up for this device via L0NodeContext.getSnapshotFeeTransactions during combine.
+    feesPaid: NonNegLong
```
ryle-ai, line 34: This puts a fee total sourced from `getSnapshotFeeTransactions` into the hashed `DeviceCalculatedState`. As the reference template, this pattern gets copied — replay (rollback/resync) must reconstruct an identical fee map or the state hash diverges and forks. Consider keeping `feesPaid` out of hashed state here, or documenting the reconstruction contract.

modules/shared/src/main/scala/io/constellationnetwork/currency/dataApplication/package.scala
```diff
@@ -859,4 +875,32 @@ trait L0NodeContext[F[_]] {
   def securityProvider: SecurityProvider[F]
   def getCurrencyId: F[CurrencyId]
   def getMetagraphL0Seedlist: Option[Set[SeedlistEntry]]
+  def getSnapshotFeeTransactions: F[Map[Hash, Signed[FeeTransaction]]]
```
ryle-ai, line 878: Abstract with no default breaks every existing `L0NodeContext` implementer/mock on upgrade. A default returning `Map.empty[Hash, Signed[FeeTransaction]].pure[F]` here would make this additive; otherwise flag it as a breaking SDK change in the release notes.

ryle-ai (commented):
The context-scoped `L0NodeContextOps.withSnapshotFeeTransactions` (delegating base wrapper, no process-wide Ref) is the right shape, and `combineUntilStable` now recomputes the fee map from the surviving block set with the recursion and replay-reconstruction paths both covered by the new suites. Two things worth surfacing before merge:

**Template hashes a looked-up fee total into calculated state.** The project template now carries `feesPaid: NonNegLong` inside the hashed `DeviceCalculatedState`, accumulated in `Combiners` from `context.getSnapshotFeeTransactions`. This is the reference scaffold metagraph authors copy, so the pattern propagates: any replay (rollback, resync, historical validation) must reconstruct a byte-identical fee map or the calculated-state hash diverges and the metagraph forks. The traverse path reconstructs the map from accepted blocks, but a copied combiner that folds fee amounts into hashed state inherits a replay-determinism obligation that isn't obvious from the template. Recommend either keeping `feesPaid` out of the hashed state in the template, or documenting the reconstruction contract prominently where it's demonstrated.

**`getSnapshotFeeTransactions` has no default in the base trait.** Adding it as an abstract method to `L0NodeContext[F]` breaks every third-party implementer/mock on upgrade (visible here in each test mock having to add the member). Please call this out in the release notes as a breaking SDK change; a default returning `Map.empty` in the base trait would make it additive instead.

Low priority: `buildFeeMap`'s last-wins `.toMap` on a colliding `dataUpdateRef` (warn-logged) leaves a residual byte-exactness gap if two fee transactions ever share a ref. Documented inline and no in-tree runtime consumer depends on the collision ordering today, so fine to leave, but note it if the map ever feeds a hashed structure.

modules/shared/src/main/scala/io/constellationnetwork/currency/dataApplication/package.scala
```diff
@@ -859,4 +875,32 @@ trait L0NodeContext[F[_]] {
   def securityProvider: SecurityProvider[F]
   def getCurrencyId: F[CurrencyId]
```
ryle-ai, line 876: This adds an abstract method to `L0NodeContext` with no default. Third-party implementers and test mocks of this trait will fail to compile on upgrade. Since this is a public SDK surface, consider either providing a default (`Map.empty.pure[F]`, which would need an `Applicative[F]` bound) to keep the upgrade non-breaking, or flagging it as a source-breaking change in the release notes.

ryle-ai (commented):
The consensus/replay determinism story here is solid: `combineUntilStable` rebuilds `buildFeeMap` from the shrinking candidate set on each recursion so a rejected block's fees never leak into a later pass, the final map exposes exactly the stored blocks' fee transactions, rejected blocks are sorted deterministically, and `DataApplicationTraverse` reconstructs the same map from the stored accepted blocks on replay. Both directions are unit-tested (the 3× fee-map assertion on the acceptance side, the stored-block reconstruction assertion on the traverse side), so `getSnapshotFeeTransactions` returns the same value at consensus time and at replay time. Checked the fee-map keying (recomputed `Hash.fromBytesForSync(serializeUpdate(update))` matching the map key), the `withSnapshotFeeTransactions` wrapper composing over any base context without a process-wide mutable Ref, and the currency-l0 `make` default returning an empty map.

Two non-blocking things worth surfacing:

- `getSnapshotFeeTransactions` is added as an abstract method on the `L0NodeContext` trait with no default. Any third-party implementation or test mock of that trait will fail to compile on upgrade. Given it's a public SDK surface, this is a source-breaking change worth calling out in release notes — or a default (empty map) could be provided to keep the upgrade non-breaking.

- The template hashes the looked-up fee total (`feesPaid`) into `DeviceCalculatedState`. The SDK closes the determinism loop, but this makes correct fee-key reconstruction a hard requirement on every metagraph author who touches this pattern — a diverging `serializeUpdate` or hash recomputation silently forks calculated state. The comment documenting the key reconstruction is good; worth making sure that obligation is explicit in author-facing docs since it now sits in the template others copy from.

ryle-ai (commented):
Three of my earlier asks are resolved and I won't repeat them: the process-wide mutable `Ref` side-channel is gone in favor of `L0NodeContextOps.withSnapshotFeeTransactions` (`package.scala:880-905`), which is the right shape — it composes over any base context including the rollback-replay predecessor; the replay path now populates the fee map via `FeeTransaction.buildFeeMap` (`DataApplicationTraverse.scala:152-162`); and the up-front `filterA` regression is gone in favor of `validateCandidates` with per-block `handleErrorWith`. The `.toMap` collision on `dataUpdateRef` now warn-logs instead of silently dropping, which I'm fine with.

Two things are still open, one of them for the fourth time.

`getSnapshotFeeTransactions` is still abstract with no default on `L0NodeContext` (`modules/shared/.../dataApplication/package.scala:878`). Anything implementing that trait outside this repo stops compiling on upgrade, and `L0NodeContextOps` demonstrates that the framework is the only thing that ever supplies a non-empty value — so the method has no reason to be abstract. `def getSnapshotFeeTransactions: F[Map[Hash, Signed[FeeTransaction]]] = Map.empty[Hash, Signed[FeeTransaction]].pure[F]` costs nothing and makes the change additive. I've raised this three times across earlier rounds with no response, so I'd rather ask directly: is there a reason you want it abstract? If there is, say so in the PR and I'll drop it — but the silence is what's blocking, not the disagreement.

The `project_template` change is the one I'd push back on hardest, and it's about what the template teaches rather than whether it works.

`DeviceCalculatedState` now carries `feesPaid: NonNegLong` (`types/Types.scala:33-35`), accumulated in `combineUpdateUsage` from a value that was *looked up* out of `getSnapshotFeeTransactions` rather than derived from the update itself. That value is inside the calculated state, so it's inside `calculatedStateProof`. Which means the metagraph's hashed state now depends on the fee map being reconstructed byte-identically on every replay path, forever. You did fix the replay population in this PR, so I think it's consistent today. But the template is documentation — it's the thing every new metagraph is copied from — and this version of it demonstrates folding a context lookup into hashed state as the recommended pattern. The next person to do that won't have `buildFeeMap` covering their lookup, and the failure mode is a metagraph that can't rebuild its own state after a rollback.

If the point is to show that `getSnapshotFeeTransactions` is reachable from `combine`, having it affect an emitted artifact or on-chain output demonstrates that just as well without putting a looked-up aggregate under the proof. If `feesPaid` in calculated state is genuinely the intended pattern, the template needs a comment saying explicitly that any value read from the node context and folded into calculated state must be reproducible on the replay path, and pointing at `DataApplicationTraverse` as the reason this particular one is.

Separately, in the same function: `NonNegLong.unsafeFrom(previous.feesPaid.value + feePaid)`, where `feePaid: Long` comes from `_.value.amount.value.value`. `unsafeFrom` over a computed sum has already taken this network down once — the token-lock sum that halted GL0 was exactly this construction. Here it's a metagraph rather than GL0, so the blast radius is one metagraph's L0 stalling in combine, but it's in the file everyone copies. `NonNegLong.from(...)` with an explicit failure decision, or saturating at `MaxValue`, is what the template should model. Note the adjacent `deviceUsage` sum has the same issue and is pre-existing — this PR rewrites those lines anyway, so it's a cheap fix for both.

On process: the branch forked at `104cde3ac` on 2026-08-05 and `develop` has 20 commits since, with content conflicts in six files — including `DataApplicationSnapshotAcceptanceManager.scala` and an add/add on `DataApplicationSnapshotAcceptanceManagerSuite.scala`, plus three of the four `project_template` files this PR touches. The acceptance-manager fold has been rewritten on develop in the interim, so resolving that conflict is itself a substantive change and not a mechanical merge. I don't think there's much value in another detailed pass over the current diff before that rebase happens, since the region I'd be reviewing is the region that moved.

Last thing, and this is a question rather than an ask: `data-invalid-fee.js` plus the changes to `data-with-fee.js` / `data-without-fee.js` cover fee submission and rejection end to end, which is good. What I don't see covered is the determinism claim itself — that a snapshot combined with a live fee map produces the same `calculatedStateProof` as the same snapshot replayed through `DataApplicationTraverse`. Given `feesPaid` is now under the proof, that's the assertion the whole design rests on. Is there a rollback or replay e2e that exercises it, or is that only covered at the unit level in `DataApplicationTraverseSuite`?

PR #1539, opened 2026-07-14
fix: update BioFi genesis address in state channel allow list

ffox77:
## Summary

Updates the BioFi genesis address in the state channel allowance list (`StateChannelAllowanceLists`) to `DAG2JaVh5yYiPCGLLEFi6tfkKk77WA4FzivVdBek`.

The allowed PeerId signing set for this entry is unchanged.

## Test plan

- [ ] Confirm the new address is the correct BioFi genesis address
- [ ] `sbt nodeShared/compile`

### DIFF
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/statechannel/StateChannelAllowanceLists.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/statechannel/StateChannelAllowanceLists.scala
@@ -119,7 +119,7 @@ object StateChannelAllowanceLists {
               "1391276cd637be03331ae5914a5a063c549fbcc6515b140d1ccc0af7c67631804991ba0dae4c1d95b84f77dc812dd71ed2a1817e50c92b722e9b8182683366ed"
             ),
           // BioFi
-          Address("DAG8ZnY1voFrENbSfwe8eT9WC6EeKTUejw7JYek4") ->
+          Address("DAG2JaVh5yYiPCGLLEFi6tfkKk77WA4FzivVdBek") ->
             NonEmptySet.of(
               "9002807a99139d207c7b20d3065f72dbf77c8563e1a2e6d2969f64928c6082dbacfb399d67bacbc62a7997fc69b8dd5c035e3acf6ed696da7f4cebee5f76cdc8",
               "866d3c60d9ea2ed7b3db2a04ba736ae8b00a0c20e98e270ddda3c73602f348ef4ce5d65e55daa52221310ebb25b826b1de495664917a3f510f6b82226d12ffe7",


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/statechannel/StateChannelAllowanceLists.scala
```diff
@@ -119,7 +119,7 @@ object StateChannelAllowanceLists {
               "1391276cd637be03331ae5914a5a063c549fbcc6515b140d1ccc0af7c67631804991ba0dae4c1d95b84f77dc812dd71ed2a1817e50c92b722e9b8182683366ed"
             ),
           // BioFi
-          Address("DAG8ZnY1voFrENbSfwe8eT9WC6EeKTUejw7JYek4") ->
+          Address("DAG2JaVh5yYiPCGLLEFi6tfkKk77WA4FzivVdBek") ->
```
ryle-ai, line 122: The peer-id set is kept identical while the metagraph key changes. That's only correct if BioFi re-deployed under a new genesis but kept the same CL0 cluster nodes — please confirm. Also note this drops the old address entirely; if both need to be valid during a transition window, the old entry would have to stay alongside the new one. Given this is a Mainnet consensus path (`Full` validation), the empty PR description leaves no record of why the address changed — worth documenting the rationale here.

ryle-ai (commented):
One-line swap of the BioFi metagraph key in the Mainnet state-channel allowance list (`DAG8ZnY1voFrENbSfwe8eT9WC6EeKTUejw7JYek4` → `DAG2JaVh5yYiPCGLLEFi6tfkKk77WA4FzivVdBek`), peer-id set unchanged. The new address is well-formed (parity digit `2` matches, and `refined.auto` validates the literal at compile time), and it's the only place either address appears in the repo, so nothing else needs updating.

Mechanically the transition is clean: only `Full` validation (live consensus, `GlobalSnapshotConsensusFunctions`) consults the address allow list; the `Historical`/replay path (`validateHistorical` in `GlobalSnapshotContextFunctions`) skips the address check, so already-accepted snapshots at the old address won't be rejected on re-sync.

This is a Mainnet consensus-affecting change with an empty description, so a couple of confirm-intent points rather than code issues:

- Because it's gated on `Full` validation, it must land via a coordinated release: during a rolling upgrade, un-upgraded nodes will still accept the old address and reject the new one (and vice-versa), so BioFi state-channel outputs could be accepted/rejected inconsistently across node versions until the fleet is on this version.
- The old address is dropped outright rather than kept alongside the new one. If BioFi's old metagraph is still expected to produce snapshots during a wind-down, this cuts it off at the cutover. If this is a genesis re-deploy that fully replaces the old identity, dropping it is correct — worth stating in the PR description either way.

Could you confirm the new address is BioFi's genesis address (i.e. it's the address generated from their current genesis binary — `validateStateChannelGenesisAddress` requires `toAddress === address` at genesis) and that the retained peer ids are still their CL0 cluster's?

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/statechannel/StateChannelAllowanceLists.scala
ffox77: create-remote-genesis was ran twice and the genesis snapshot using the older address was overwritten

ryle-goehausen (approved):

PR #1541, opened 2026-07-15
fix(consensus): witness admission candidates cluster-wide, not just the gossip mesh

scasplte2:
## Summary
Fixes the v4.1.0 committee-admission wedge seen on IntegrationNet: on clusters larger than the gossip
mesh, the B2 admission gate never witnessed most Ready candidates, so the Core committee froze below
`coreFloor` and consensus limped along closing rounds only via timeout (~45s/snapshot). This decouples
the admission chain-tip witness channel from the gossip mesh so every responsive peer is witnessed
cluster-wide.

## Root cause
`getPeerChainTips` (the B2 admission witness channel) reads `MeshState.getChainTips`, which is
populated only from gossip-mesh interactions: the heartbeat sampled 3 random mesh peers and the pull
loop contacted the same 3 non-mesh peers each cycle (`.take(gossipFactor)` on a `Set`, no shuffle).
The mesh is capped at `meshHigh=12`, so on an 82-node cluster only ~15 peers were ever witnessed and
the ~67 Ready admission candidates had `atTip=0` -> zero admission certs -> Core stuck below
`coreFloor=9`, rounds closing only via timeout. Clusters at or below `meshHigh` graft every peer into
the mesh, so `getChainTips` covered everyone -- which is why this only bit at IntegrationNet/MainNet
scale and never on TestNet.

## Changes
* **Modified** - `EventGossipDaemon`: replaced the mesh-only heartbeat chain-tip sampling (3 random
  mesh peers) with a round-robin witness sweep over ALL responsive peers, sized by the new
  `chainTipWitnessRefreshInterval` (default 45s) to refresh every peer within the admission
  tip-validity window regardless of cluster size.
* **Added** - `ChainTipWitnessSweep` pure helper (sweep sizing + round-robin slice) and the
  `chainTipWitnessRefreshInterval` config field.
* **Added** - `ChainTipWitnessSweepSuite` covering the sizing math and the round-robin coverage
  guarantee.

## Testing
* `nodeShared/compile` clean; `ChainTipWitnessSweepSuite` 3/3 green -- asserts the sizing scales with
  cluster size and the round-robin sweep covers EVERY peer within `ceil(n / sweepSize)` heartbeats
  (verified across cluster sizes 1..200).
* Liveness-only change: no signed snapshot bytes change and it is not in `deterministicConfigHash`
  (committee mutation still requires quorum-certified `AdmissionCertificate`s). Deploys via the normal
  coordinated cold restart (jar-hash fence); no gate-ordinal coordination.
* Real-world validation on deploy: `dag_consensus_committee_core_size` should climb off its bootstrap
  size toward `coreFloor` as candidates get witnessed and admitted, tightening snapshot cadence.


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/event/EventGossipDaemon.scala
```diff
@@ -567,44 +614,56 @@ private class EventGossipDaemonImpl[F[_]: Async: Parallel, Event, Key](
...
-                    ) >> meshState.updateChainTip(peer.id, tip)
-                  case None =>
-                    logger.debug(s"Peer ${peer.id.show} returned no chain tip in IHave response")
-                }
+          witnessCursorRef.getAndUpdate(c => (c + sweepSize) % n).flatMap { rawStart =>
+            val sampled = ChainTipWitnessSweep.slice(ordered, rawStart, sweepSize)
+            logger.debug(
+              s"Chain tip witness sweep: availablePeers=$n sweepSize=$sweepSize start=${((rawStart % n) + n) % n}"
+            ) >>
+              sampled.traverse_ { peer =>
```
ryle-ai: This `traverse_` runs the `getIHave` calls sequentially, and `sweepSize` now scales with the cluster (ceil(n/heartbeatsPerRefresh) — ~50 for a 200-peer cluster) instead of the old fixed cap of 3. With no per-call timeout on `getIHave` (`config.fetchTimeout` is defined but not applied here; the effective bound is the coarse shared HTTP client timeout), even a few slow peers serialize and can block `runHeartbeat` for far longer than `heartbeatInterval`, delaying the mesh graft/prune and fork-detection steps that follow in the same for-comprehension. Consider `parTraverseN` (as `pushEvent` at line 251 already does via `parTraverse`) and wiring `fetchTimeout` onto each call so the sweep stays off the heartbeat critical path on the large clusters this fix targets.

ryle-ai (commented):
The coverage fix itself is sound — round-robin over a stable ordering does guarantee every responsive peer is witnessed within `chainTipWitnessRefreshInterval`, and the `sweepSize`/`slice` math and tests hold up. One concern about how the enlarged sweep runs inside the heartbeat.

**Sequential fan-out inside the heartbeat now scales with cluster size.** The sweep uses `sampled.traverse_` (sequential), and `sweepSize` grows with the cluster — `ceil(n / heartbeatsPerRefresh)`, i.e. ~21 peers on an 82-node cluster and ~50 on a 200-node cluster. `getIHave` has no per-call timeout applied here (`config.fetchTimeout` is defined but never wired to this call; the effective bound is the shared HTTP client timeout, which is coarse). So a handful of slow/unresponsive peers serialize behind one another and can push `runHeartbeat` well past `heartbeatInterval`, delaying the mesh graft/prune maintenance and the fork-detection pass that run in the same for-comprehension. The pre-fix code had the same sequential shape but was capped at 3 peers, which bounded the blast radius; this removes that bound precisely on the large clusters the PR targets.

Worth bounding this: `parTraverseN` (the daemon already has `Parallel`, and the sibling `pushEvent` fan-out at line 251 uses `parTraverse`) plus applying the already-existing-but-unused `fetchTimeout` as a `.timeout(...)` on each `getIHave`. That keeps the witness sweep off the heartbeat's critical path regardless of how many peers are swept.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/event/EventGossipDaemon.scala
scasplte2: Good catch, thanks -- addressed in 42e50b53e. The witness sweep now:

- runs bounded-parallel via `parTraverseN(config.maxConcurrentPulls)` instead of the sequential `traverse_`, so slow peers no longer serialize;
- wraps each `getIHave` in `Async[F].timeout(_, config.fetchTimeout)` (matching the `ForkRecoveryDetector` probe) -- a hung peer clears its tip instead of stalling the batch;
- is fired off the heartbeat critical path via `S.supervise`, so the graft/prune + fork-detection steps that follow never wait on it;
- is single-flighted with a `Ref[Boolean]` guard so a slow sweep can't pile up supervised fibers -- a tick where the previous sweep is still running is skipped and the cursor stays put, so no peer is missed.

ryle-goehausen (approved):

PR #1542, opened 2026-07-15
chore: rotate validator node id in mainnet seedlist

ffox77:
## Summary

A validator rotated their key. Replaces their old node id with the new one in `mainnet/mainnet-seedlist` (single in-place swap, line 131).

- Old: `2e8181617836676b905686fe5aef982f64356aa68cfe0589b5d8a08ac9535a43280d0e1999d37d810ee0f7ded4687995c6f6f8ddfb733508ca1928421f0de6a2`
- New: `3ef97d124f59c33eae6c81a29bd8e6ee44f0c518b2b22fe07ea0cca8fac877028aeb97c9dd73b9f825c590451195b9bbbbeaadf0d3bedd1279b7328997acd90d`

## Test plan

- [ ] Confirm the new node id is correct and belongs to the same validator


mainnet/mainnet-seedlist
```diff
@@ -128,7 +128,7 @@ a90cab5814e2c9de9d7f147d47ae085e2c5c0c0503e15ddeed10449857d4256b6c6c2febb1cfb9b1
 7b3f6b88c2cd5c67ee386db364fbc09db2014215c57a42d400047c14a07b822e66c414413ae2632632ee8b7d4b52a2a69c94e1220804d0f18e89720b5ad3b085
 4d87953a7b86c5e153c015b997be0c2cff101e28860ec685fe1a2fddc40ed7bf4613b2317f98f4e90d76a8e4ea9dc176b7e179da0e4366c8221736ad31ab7a83
```
ryle-ai, line 129: This replaces the batch-3 id `2e818161…` with `3ef97d12…`. Since the seedlist hash is compared at the join handshake (`Joining.validateSeedlist`, over peer-ids), the old-id node stops being join-eligible as soon as operators adopt this list and the new-id node can't register with un-upgraded peers — so this needs a coordinated rollout with the rotated node's redeploy. An empty PR description also leaves no rationale for the rotation; a one-line note on which validator/why would help.

ryle-ai (commented):
One-line rotation of a validator node id in the mainnet seedlist: the id added in "Mainnet seedlist - batch 3" (`2e818161…`) is dropped and replaced with a fresh id (`3ef97d12…`). The new id is well-formed (128-char lowercase hex, single-field entry consistent with the rest of the file) and appears nowhere else in git history, so this looks like a genuine key rotation rather than a copy/paste. No code issue.

Two non-blocking points, both stemming from the empty PR description on a mainnet-affecting change:

1. **No documented rationale.** There's no context for which validator is rotating or why. The seedlist is the mainnet allow-list, so a note (regenerated keypair / decommissioned host / etc.) would help future readers understand why this id changed.

2. **Coordinated rollout.** The seedlist is hashed over its peer-ids and enforced at the join handshake (`Joining.scala:274-275`, `seedlist.map(_.map(_.peerId)).hash` → `SeedlistDoesNotMatch`). Rotating an id changes that hash, so during a rolling upgrade nodes on the old vs. new seedlist reject each other's registration. In particular the node holding the old id (`2e818161…`) loses join eligibility the moment operators pick up this change, and the node with the new id can't join peers still on the old list. Worth confirming this ships as a coordinated seedlist update alongside the rotated node's redeploy.

ryle-goehausen (approved):

PR #1547, opened 2026-07-17
fix(consensus): isolated-node escalation, admission gating, and full-committee delegated rewards

scasplte2:
## Summary
Consolidates three consensus work streams into one PR, one commit per piece, so review happens in a single place. Covers: (1) issue #1533 rumor-isolated node escalation via a corroborated HTTP preflight, (2) the IntegrationNet rc.2 admission oscillator fix with a sticky probation lane, and (3) delegated rewards paying the full frozen committee behind an activation ordinal. Also carries the currency-l1 codec workaround that unblocked develop's Build JARs.

## Changes
* **Added** (`451ad2491`) - `PeersCommittedAheadProbe`: samples up to 8 Ready peers' `/latest/metadata`, confirms only on >= 2 corroborating responders agreeing on the same ahead `(ordinal, hash)` with a strict responder majority (clamped to sample size so 2-node metagraphs stay live), per-peer 2s timeout, typed probe outcomes. `AbandonmentTracker` probes on every non-advanced rumor shape and escalates to recovery only when the probe confirms committed progress at/beyond the abandoned key (cause `rumor_isolated`); a cluster-wide stall never escalates because nobody confirms.
* **Modified** (`3bc3f121a`) - currency-l1 `EstimatedFeeResponse`/`TransactionRoutes` use explicit circe codecs instead of magnolia/shapeless derivations, routing around the scalac 2.13.18 backend assertion that broke every PR's Build JARs job.
* **Modified** (`a87fd6fef`) - Admission sizing resolves once in `ConsensusPeerController.AdmissionSizing`: the score-gating bypass threshold is now `activeFacilitatorFloor` (the documented v18 contract) instead of `coreCommitteeSize`, whose knife-edge at the operating pool size produced the rc.2 relaxation oscillator (~80% AdmissionVote gossip, hours-long committee regrowth). The bounded probation lane is now sticky: a below-retain recent signer that signed the latest certified round keeps competing for a non-Core seat ahead of fresh candidates; missing the latest round ends the lease (`missed_latest_round` exclusion) and bars fresh re-entry for one round. Lane composition is observable (`stickyProbationCandidateSize`, `freshProbationCandidateSize`, `freshProbationStarved` in logs and prometheus gauges). No config values or `deterministicConfigHash` composition changed.
[...]

PR #1548, opened 2026-07-17
fix(currency-l1): route estimate-fee responses around a scalac backend crash

scasplte2:
## Summary
Unblocks CI for every open PR. Since #1544, the `Build JARs` job fails on all branches with a scalac 2.13.18 backend crash while emitting currency-l1 (`"an unexpected type representation reached the compiler backend"`, `"ClassBType.info not yet assigned"`, `"assertion failed: type R"` -- the compiler itself asks to file a scala/bug), which skips the entire E2E matrix. Branch-independence is proven: a whitespace-only PR (#1545) failed identically, while #1543 passed the same job pre-#1544. The crash names the same two derivation sites in every failing run but does not reproduce outside CI (warm, clean-slate, and cold-clone runs of the exact `assembly.sh` sequence pass locally with and without `CI=true`), so this routes around the fragile derivations rather than chasing the environment trigger.

## Changes
* **Modified** - `EstimatedFeeResponse.scala`: explicit `Encoder.forProduct3` replaces the magnolia-derived `@derive(encoder)`; identical JSON shape and key order.
* **Modified** - `TransactionRoutes.scala`: plain `Json.obj("fee" -> ...)` replaces the shapeless singleton-record response (`("fee" ->> ...) :: HNil` via `io.circe.shapes`), whose singleton-typed encoder is the `type R` representation leaking into the backend; identical JSON (`{"fee": <long>}`). Drops the shapeless imports.

## Testing
On the fixed sources: `currencyL1` main + Test scopes compile; `scalafmtCheck` passes in both scopes; the exact failing CI command (`CI=true sbt "set ThisBuild/version := 99.99.99-SNAPSHOT" sdk/publishLocal`) exits 0. The real verdict is this PR's own `Build JARs` job -- it exercises the previously-crashing pipeline end to end; if it greens here, rebasing/rerunning the blocked PRs (#1547) clears them.

### DIFF
--- a/modules/currency-l1/src/main/scala/io/constellationnetwork/currency/l1/http/EstimatedFeeResponse.scala
+++ b/modules/currency-l1/src/main/scala/io/constellationnetwork/currency/l1/http/EstimatedFeeResponse.scala
@@ -8,13 +8,18 @@ import io.constellationnetwork.schema.address.Address
 import io.constellationnetwork.schema.balance.Amount
 import io.constellationnetwork.security.hash.Hash
 
-import derevo.circe.magnolia.encoder
-import derevo.derive
+import io.circe.Encoder
 
-@derive(encoder)
 case class EstimatedFeeResponse(fee: Amount, address: Option[Address], updateHash: Hash)
 
 object EstimatedFeeResponse {
+  // Explicit encoder instead of the magnolia derivation (`@derive(encoder)`): the CI Build JARs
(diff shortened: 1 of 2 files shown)


scasplte2:
Superseded: folded into #1547 as 3bc3f121a per review direction. This branch's own CI run already proved the fix (12/12 including Build JARs).

PR #1551, opened 2026-07-31
test(e2e): assert delegated rewards pay the full frozen committee

scasplte2:
## Summary
Adds an E2E test proving that delegated validator rewards pay **every** member of the frozen round-start committee (Core + Tier 1) an equal share, including a Tier-1 seat that the removed evidence-score filter would have dropped. This is the first end-to-end coverage of the #1547 / ADR-0028 change. No production Scala is touched; the only non-test source change is a corrected comment in `dag-l0.conf`.

## Changes
* **Added** - `.github/action_scripts/committee_rewards.js`: samples a global snapshot whose committee pairs a **chronic, promotion-barred Tier-1 seat below the promote threshold** with a promote-qualified peer, then asserts that reward destinations exactly equal `roundStartFacilitators`, that every Tier-1 seat is paid, that amounts are strictly equal, that signers are a subset of the paid committee, and that an EventTrigger ordinal mints nothing.
* **Added** - `committee_rewards.fixture.js` + `committee_rewards.selftest.sh`: an offline synthetic-snapshot server and a one-command self-test that runs the real script in **both** directions — it must pass on correct behavior and **fail** under `--drop-climber-reward`, which reproduces the legacy filter. Invoked from CI *before* the cluster boot, so a broken assertion fails in ~2s instead of 20 minutes.
* **Added** - staggered gl0 join: `--num-gl0-early=K` / `--gl0-late-delay=S` write a per-node `CL_DOCKER_GL0_JOIN_INITIAL_DELAY` into `nodes/<i>/.env`, reusing the mechanism gl1/cl1/dl1 already use. Zero compose changes; unset leaves every other test untouched; both flags are validated as integers.
* **Added** - `docs/operations/local-e2e-cluster-investigation.md`: how to interrogate a live local cluster (it survives a failed `just test`), reconstruct committee/tier/score from signed artifacts, and re-run a test script against a running cluster in seconds. Companion to the existing log-analysis procedure, not a duplicate of it.
* **Modified** - `compose-runner.sh` registers the test with a rig guard that **errors** when selected by name with the wrong rig and skips on a bare `just test`; new matrix job `committee-rewards` (`--num-gl0=5 --num-gl0-early=3`); CI node-log collection widened past index 2 so multi-node rigs upload their late joiners' logs.
* **Modified** - `dag-l0.conf`: **comment only** — the probation reward note still described the payout filter retired in #1547.

## Why the sample is narrow
The removed filter paid everyone when *no* facilitator qualified, so a uniformly-low committee does not distinguish it from current behavior, and neither does a uniformly-saturated one. Only a mixed committee does.

"Below promote" is also not a tier claim on its own, and neither is the evidence-derived tier: `CommitteeBuilder` can promote a derived-Tier-1 peer back into Core via the one-for-one replacement and Core-floor top-up, both drawing from `rawTier1.filterNot(isChronic || nonCorePeers)`. The predicate therefore requires the candidate to be **chronic** (trailing seated-but-missed streak >= `ChronicMissThreshold`, which bars both mechanisms) **and** at least `MinViableCoreSize` healthy Core members to remain, which zeroes the liveness fallback — the only remaining path that re-admits a chronic peer. Both conditions are computed from signed evidence the way the builder computes them, since per-peer tier is deliberately blanked in the artifact.

## Testing
**Offline (automated, in CI):** the self-test passes on correct behavior and exits non-zero under the legacy-filter mutation with `gl0-3 was a seated Tier-1 facilitator at ordinal 12 with score 30 ... but received no reward`. It also fails if the pass case never reports a Tier-1 seat, so a vacuous pass is caught.

**Live 5-node cluster:** passes, and independently corroborated against the node's own committee derivation — for the sampled ordinal the log reports `core=2 tier1=1 witness=0 coreFloor=3` (agreed by all three seated nodes), with `coreFloor=3` deliberately unmet because the chronic peer was correctly **not** promoted. Verified from signed data in the same run: candidate `consecutiveMisses=3`, two healthy Core at 0, three equal rewards, paid set == committee set, `delegateRewards` empty, all five nodes agreed on the tip, zero errors on the seated nodes.

**Still open:** the live legacy-filter mutation run (point `fields-added-ordinals.delegated-rewards-full-committee.dev` at a future ordinal, rebuild without `--skip-assembly`, expect failure). The offline mutation already proves the assertions have teeth; this would close it end-to-end on a real cluster.

## Review notes
* Reviewed across five codex rounds; two of my earlier Tier-1 proofs were rejected as unsound or unreachable and are recorded in the script header so neither gets retried.
* Measured behavior worth knowing independent of this PR: `completedSigners` is **not** a signature set — `canonicalCompletedSigners` intersects the committee with `observedResponders` (Facility-phase declarers seen by the leader). A peer can therefore appear in `snapshot.proofs` and still be recorded as a miss; over 63 local ordinals two late joiners were seated 8 and 13 times and recorded as signers zero times while Ready and at tip. They still get paid, since payout follows `roundStartFacilitators`. Documented in the ops doc as a diagnostic trap.
* Open question for reviewers: the Tier-1 seat is no longer required to be a late joiner (a genesis peer gone briefly quiet is an equally genuine Tier-1 seat), so the 5-node staggered rig is a reliability aid rather than a strict requirement.


ryle-ai (commented):
No issues found; approved.

ryle-goehausen (dismissed):

ryle-ai (commented):
Test-only e2e change asserting delegated rewards pay the full frozen Core+Tier-1 committee equally (ADR-0028), plus staggered-join infra and a config comment rewrite. Verified the JS oracle faithfully mirrors the Scala consensus source: derivePeerScores arithmetic (completed*20 - missed*15 + certAppearances*10, clamped [0,150]) matches ControllerEvidenceDerivation; the seated constants (SIGN=20/MISS=15/CERT=10, DEMOTION=CHRONIC=3, MIN_VIABLE_CORE=2) match their Scala definitions, and the env-tunable ones (promote/retain thresholds, participation window/ratio) read the same env vars the node reads so they can't drift. The committee-for-N-lives-in-N+1 oracle reads snapshot[N+1].peerHistory.controllerEvidence to reconstruct roundStartFacilitators. The five assertions have teeth (every seated Tier-1 climber must be paid; reward dests == frozen committee exactly; equal positive split; signers subset paid; EventTrigger rounds pay zero) and the committee_rewards.selftest.sh --drop-climber-reward path proves non-vacuity by forcing a non-zero exit. dag-l0.conf change is comment-only (ADR-0028 reward note; probation-reentry-slots values unchanged mn15/in9/dev3/tn3). set-env.sh validates numeric NUM_GL0_EARLY / GL0_LATE_JOIN_DELAY inputs. No defect at the bar.

ryle-ai (commented):
Verified the JS reward oracle in committee_rewards.js faithfully mirrors ControllerEvidenceDerivation.scala: SignWeight=20, MissWeight=15, CertWeight=10, MinScore=0, MaxScore=150, ChronicMissThreshold==DemotionConsecutiveMisses(3), MinViableCoreSize=2, and the score/tier/consecutive-miss arithmetic (completed*20 - missed*15 + certAppearances*10 clamped [0,150]; trailing-miss streak with signed-reset and not-asked-break). The scanWindow fixture builds a genuinely mixed final committee (a chronic below-promote Core plus a real climbing Tier-1 seat) so the five main() assertions have teeth: the legacy per-facilitator payout filter's all-fallback path is closed, and the Tier-1 climber cannot be promoted back to Core because it is chronic and >=2 healthy Core remain (zeroing the step-5 liveness re-admit) per CommitteeBuilder's chronic-exclusion ladder. The --drop-climber-reward selftest confirms non-vacuity (fails on the 'seated Tier-1 facilitator' assertion). committee-for-ordinal-N-lives-in-N+1 off-by-one, TimeTrigger-vs-EventTrigger split (ordinal%2 over epochProgress delta), and equal-split reward destination checks are correct. The dag-l0.conf hunk is comment/documentation-text only; config values (Core floors mainnet 15 / integrationnet 9 / dev 3) are unchanged context. The compose-runner NUM_GL0>=5/NUM_GL0_EARLY guard and set-env.sh --num-gl0-early/--gl0-late-delay validation are consistent with the workflow matrix args.

ryle-goehausen (approved):

Issue #1533, opened 2026-06-30
AbandonmentTracker: rumor-isolated node never escalates to recovery-download (v4.1.0)


scasplte2:
## Summary

In the v4.1.0 line (merged in #1523), a **rumor-isolated but still-reachable node never escalates to recovery-download** — it retries its abandoned consensus key indefinitely instead of recovering. Surfaced in the post-merge review of #1523.

## The gap

In `AbandonmentTracker.scala`, both escalation paths gate on `networkAdvanced`:

- retriable path: `networkAdvanced = peersAtHigherKey > 0`, then `if (networkAdvanced) triggerRecoveryDownload(...) else retryAfterRetriableAbandon(...)`
- non-retriable path: `willRecover = shouldRecover && networkAdvanced`

`peersAtHigherKey` is counted from `storage.getPeerCurrentKeys`, which is fed **only by incoming keyed rumors**. For a node that has lost the rumor mesh but can still reach peers over HTTP (exactly the case develop's `8027c0642` `(!peersHaveCurrentKey && totalReadyPeers > 0)` clause targeted), those tips freeze, so `networkAdvanced` stays `false` even when peers genuinely are ahead. The node then loops on its abandoned key forever.

The newer wedge-health signal does **not** close this: it only makes the node refuse external `Cluster.leave` POSTs; it does not escalate to download. `triggerRecoveryDownload` is the only caller of `forceLeave` and the only place `totalRecoveryAttempts` increments, so an isolated node never reaches force-leave either.

This came in via the conflict resolution that took `release/testnet`'s `AbandonmentTracker` (the `networkAdvanced` / `peerCurrentKeys` / wedge-health form); develop's isolated-node escalation clause was dropped in the process.

## Why it matters

Liveness: a node that is gossip-isolated but HTTP-reachable, where the cluster has advanced, will never recover on its own. Flagged in the review as the one finding worth confirming before the v4.1.0 launch.

## Important constraint — do NOT naively re-add

The `networkAdvanced` gate exists partly to prevent the **cascade-recovery deadlock**: without it, a fresh cold-restart where all source nodes reboot together cascades every node into `WaitingForDownload` on the first failed round, with no peer ahead to serve. A fix must distinguish "I am rumor-isolated but peers are reachable and ahead" from "the whole cluster is stuck at this ordinal," and only escalate in the former.

## Proposed next steps

[...]

PR #1552, opened 2026-07-31
fix(gossip): unwedge a returning peer's rumor chain, and make fork-recovery runnable

scasplte2:
## Summary
The `fork-recovery` e2e was disabled as "never passing". Its Phase-1 gate was arithmetically unsatisfiable, so the test had never actually run. Fixing the gate let it run, and it immediately found a real node bug: **a peer that is unreachable for longer than its own rumor buffer comes back permanently muted, and because consensus can still elect it leader, the whole cluster freezes.** This PR fixes the harness, fixes that bug, and hardens the pass criterion.

The matrix entry is still commented, because one further recovery gap remains (below). Nothing here can affect existing CI jobs.

## Changes
* **Fixed (node)** - `RumorStorage`: a peer rumor is accepted only at exactly `head + 1`, and the repair path (`peerRound`'s inquiry) is bounded by `peer-rumors-capacity` (50) per origin. Once the gap exceeds that, the missing range is evicted everywhere, the chain can never be made consecutive, and every later rumor from that origin is dropped as `CounterTooHigh` - silently, with no log line. `isGapUnclosable(head, incoming, capacity)` restarts the chain when, and only when, the gap provably cannot be served by anyone (`incoming - head > capacity`). Reported as a distinct `ChainRestarted` result and logged at WARN.
* **Fixed (test)** - Phase-1 gate now derives the committee from the signed artifact (`snapshot[N+1].peerHistory.controllerEvidence[N].roundStartFacilitators`) instead of comparing log-scraped facilitator counts. Isolation target is picked from the seated set by node index and pinned across the 3-sample window, while the rest of the committee may churn.
* **Fixed (test)** - the catch-up criterion now measures the median ordinal across reachable nodes rather than `MONITOR_NODE`'s own.
* **Modified** - `docker logs` scans bounded with `--tail`; recovery counting anchored with `--since` at restore; a `grep -c` fallback that emitted two zeros.
* **Modified** - `.github/workflows/e2e-just-test.yml`: comment only, plus `timeout: 60` on the still-commented entry.

## Why the old gate could never pass
`[ "$committee_size" = "$mon_fac" ]` compared a **count of agreeing nodes** to the **committee size** - equal only when the committee is the whole cluster, contradicting the gate's own comment ("Committee size doesn't have to be N"). Replayed against a live 5-node run:

```
+270    fac=5 5 5 5 5   ord=21 21 21 21 21   PASS
+315    fac=4 4 4 4 4   ord=24 24 24 24 24   fail(size=5 vs mon_fac=4)   <- healthy, unanimous, rejected
+360    fac=5 5 5 5 4   ord=27 27 27 27 26   fail(agrees=false)          <- sampling skew, not disagreement
gate passed 2 / 17 sample points, never twice in a row
```

It requires 3 consecutive. The inputs were also each node's *last-ever* logged count, so nodes were compared across different rounds and a legitimate 4<->5 oscillation read as a split. Fixed gate: 3 consecutive in ~90s.

Two incidental findings: **PR #1485, the stated re-enable blocker, was closed without merging**, and the entry had no `timeout:`, inheriting 30 min against a ~40 min test.

## The node bug, measured
Run with the old code, isolating a seated peer for 270s:

```
21:35:40   peers:  "Peer b5534065 is unresponsive"
21:37:58.1 peers:  "Syncing 1 missing events to newly grafted peer b5534065"   <- gossip re-graft
21:37:58.9 gl0-3:  WaitingForDownload -> WaitingForObserving -> Observing -> Ready (208 ms)
```

The returning node is re-grafted into the gossip mesh but its rumor chain is never repaired, so every rumor it sends is dropped:

| | Proposal decls | Signature decls |
|---|---|---|
| gl0-3 (leader) | 21 | 21 |
| all four peers | 0 | 0 |

Peers kept it at `WaitingForObserving` while it reported `Ready`. Consensus derives committees from `peerHistory`, not gossip reachability, so it seated the node **and elected it leader**; the leader logged `Quorum reached: 3/3` while followers sat in `CollectingProposals` emitting `FORCED_VIEW_CHANGE`. View change could not route around it - the round is re-created at view 0, deterministically re-electing the same leader. **Ordinal 41 held for 900+s with all five nodes Ready and at tip.**

## Consensus safety
**Behavior-only. Not consensus-breaking.**
* It governs only which rumors a node accepts into its **local** store - no artifact content, signature, committee derivation, evidence field or reward computation changes.
* `peerRumorsCapacity` lives in `RumorStorageConfig`, not `ConsensusConfig`; it is not folded into `deterministicConfigHash` (verified against the hash body in `types.scala`). Divergent values cannot cause peer rejection or a fork.
* Divergent local rumor sets are already an expected condition - that is precisely why `canonicalCompletedSigners` uses proposal-carried, round-start-frozen data rather than locally observed proofs.
* `validateRumor` runs before `tryAddRumorToStore`, so a restarted chain cannot admit anything that was not already signature-valid.
* The change strictly *accepts more* rumors, only in a state where previously all were dropped forever. It cannot make a healthy node reject something it used to accept.

## Testing
`RumorChainGapSuite` (new, 7 tests) pins the bound in both directions, at the exact boundary, for a long isolation, for counters at or below the head, and for scaling with capacity. `nodeShared` compiles; scalafmt clean in a fresh session.

Live 5-node rig, with the fix:
* chain restart fired on all five nodes (`Restarted rumor chain for origin=345647c8 at ordinal=...counter=502`)
* the returning node went from permanently muted to fully gossiping - pushing events to all four peers, receiving declarations for later rounds
* peers moved it off `WaitingForObserving`
* **the cluster kept producing, 41 -> 58, instead of freezing**

## Why the matrix entry is still commented
The fix changes the failure shape from "entire cluster halts" to "one node lags", which is a large severity reduction but not yet a clean recovery. After rejoining, the node promotes to `Ready` at a target computed when its download began and never backfills: it participated in round 59 while still serving `latest=51`, with `completedAfterIsolation=0`.

That is a third, separate issue in the download/recovery path. It is also behavior-only - no signed content or validation changes - but it sits where the documented historical failure mode is a cluster-wide false-lagging recovery cascade (the trap that made the first #1533 attempt wrong), so it wants deliberate review rather than a bolt-on. `downloadReadyPromotionAllowed` already models an `Ahead` verdict from a 409, which is the likely place for a narrow fix. Until then the test correctly fails, and enabling it would make CI red.

## Review notes
* `CL_QUORUM_THRESHOLD_FRACTION: "1.0"` in this workflow is **dead config** - absent from both compose files and every generated `.env`, so nodes use the conf default. If it were ever plumbed through, `QuorumPolicy.fromFraction` takes the unanimity branch and isolating any committee member becomes an unrecoverable wedge by construction.
* PR #1551 also touches `e2e-just-test.yml` a few lines below and widens CI node-log collection past index 2; distinct hunks, should merge cleanly.

### DIFF
--- a/.github/workflows/e2e-just-test.yml
+++ b/.github/workflows/e2e-just-test.yml
@@ -98,8 +98,19 @@ jobs:
           - { name: dag-cluster,            tests: "dag-cluster" }
           - { name: delegated-staking,      tests: "delegated-staking" }
           - { name: token-lock-replacement, tests: "token-lock-replacement" }
-          # TODO: re-enable once PR #1485 is merged
-          # - { name: fork-recovery,          tests: "fork-recovery", extra-args: "--num-gl0=5", runner: "Ubuntu-22-64-core" }
+          # DISABLED on a NODE bug, not a test bug. The Phase-1 gate was unsatisfiable and is now
+          # fixed (see docker/bin/test-fork-recovery.sh); phases 1 and 2 pass, and the cluster keeps
+          # producing snapshots while a seated peer is isolated. Phase 3 then wedges: on restore,
+          # peers re-graft the returning node into the gossip mesh but leave it at
+          # WaitingForObserving in their cluster registry, while the node itself reports Ready.
+          # Consensus derives committees from peerHistory rather than that registry, so it can elect
+          # the returning node LEADER -- whose Proposal/Signature declarations every peer then
+          # ignores. Observed locally 2026-07-31: ordinal 41 held for 900s with all 5 nodes Ready and
+          # at tip. Re-enable once that rejoin/leader-eligibility mismatch is fixed; the entry below
+          # is correct as written, including the timeout (the test's own budget is ~40 min, so the
+          # 30 min default would have killed it regardless).
+          # The previous note here blocked on PR #1485, which was closed without ever merging.
+          # - { name: fork-recovery,          tests: "fork-recovery", extra-args: "--num-gl0=5", runner: "Ubuntu-22-64-core", timeout: 60 }
           - { name: currency,               tests: "currency" }
           - { name: rewards,                tests: "rewards" }
           - { name: token-locks,            tests: "token-locks" }
--- a/docker/bin/test-fork-recovery.sh
+++ b/docker/bin/test-fork-recovery.sh
@@ -68,6 +68,27 @@ get_facilitator_count() {
   echo "${result:-0}"
 }
 
+# Helper: the cluster's tip, as the MEDIAN ordinal across reachable gl0 nodes.
+#
+# Deliberately not MONITOR_NODE's own ordinal. The monitor is just gl0-0; it is not guaranteed to
+# be in the committee and it can itself fall behind and recover, which silently weakens every check
+# built on it. Observed on 2026-07-31: a run "passed" its catch-up criterion with iso_ord=51 against
+# a monitor stuck at 41 while the real cluster tip was 58 -- the isolated node had not caught up at
+# all, it was merely ahead of a second lagging node.
+#
+# Median rather than max so a single raced-ahead node cannot make the target unreachable, and
+# rather than min so a single laggard cannot make it trivially satisfiable.
+get_cluster_tip() {
+  local ords=()
+  local i o
+  for i in $(seq 0 $((NUM_GL0 - 1))); do
+    o=$(get_ordinal "gl0-${i}")
+    [ -n "$o" ] && ords+=("$o")
+  done
+  [ "${#ords[@]}" -eq 0 ] && return 0
+  printf '%s\n' "${ords[@]}" | sort -n | awk '{a[NR]=$1} END {print a[int((NR+1)/2)]}'
+}
+
 # Helper: peer id of a node, from the key material node-key-env-setup.sh syncs into nodes/<i>/.
 peer_id_of() {
   local idx=${1##gl0-}
@@ -332,7 +353,7 @@ fi
 echo "  Isolation target: $ISOLATION_NODE (seated committee:${committee_nodes})"
 
 # Record pre-isolation state from monitor (a validator)
-pre_ordinal=$(get_ordinal "$MONITOR_NODE")
+pre_ordinal=$(get_cluster_tip)
 pre_isolation_ordinal=$(get_ordinal "$ISOLATION_NODE" 2>/dev/null || echo "$pre_ordinal")
 echo "  Pre-isolation ordinal: $pre_ordinal (monitor), $pre_isolation_ordinal (isolated node)"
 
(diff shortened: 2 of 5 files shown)


ryle-ai (commented):
Checked the rumor-chain unwedge and the fork-recovery harness.

`RumorStorage.isGapUnclosable(head, incoming, capacity)` uses strict `incoming - head > capacity`, so a gap of exactly `peerRumorsCapacity` stays waiting (the origin still retains counter `head+1`) and only `capacity+1` triggers a restart — matches the retention bound and the boundary cases in `RumorChainGapSuite`. Subtraction is safe: the branch is only reached after `headCounter.next < rumorCounter`, so `incoming > head`. Confirmed `lastOrdinalsR` stays consistent with the chain head across every `addPeerRumorIfConsecutive` branch — it is updated exactly on the paths that move the head (AddSuccess, ChainRestarted, generation bump, first rumor) and left untouched on the no-op results (CounterTooLow/TooHigh/GenerationTooLow); ChainRestarted resets the chain to `one(rumor)` and sets `lastOrdinalsR` to that same ordinal, so the advertised cursor and served chain agree. The restart only discards a range that is by definition unrecoverable, and since rumors are signature-validated independently of chain position an origin restarting its own chain cannot affect another peer's data. GossipDaemon still re-offers a self-origin `CounterTooHigh` and now treats `ChainRestarted` as consumed with a WARN.

On the harness: `get_committee_peer_ids` reads the committee for `tip-1` from snapshot `tip` (guarded against `tip<1`), `get_cluster_tip` takes the median across gl0 nodes rather than the monitor's own ordinal, and `get_completed_rounds_after` keys on the once-per-round `CONSENSUS_FINISHED` line scoped by `--since`. Considered and dismissed: ChainRestarted starving a third peer that reads a dropped range (that peer is itself past the unclosable bound relative to the origin, so it is served consistently) and a spurious restart from transient reordering (in-flight depth is far below the buffer capacity). The e2e entry remains commented out, so none of the harness changes run in CI yet.

ryle-goehausen (approved):

ryle-ai (commented):
Checked the gossip fix and the harness rework.

`RumorStorage.addPeerRumorIfConsecutive`: the new `isGapUnclosable(head, incoming, capacity) = incoming - head > capacity` uses strict `>`, so a gap exactly equal to capacity still waits for inquiry rather than restarting — that boundary is the load-bearing one and it's right. The subtraction is safe: the `ChainRestarted` branch is only reachable when incoming > head+1, so `incoming - head >= 2`. `lastOrdinalsR` stays consistent with the chain head across every path — the two success branches and the restart branch all update it, and every reject branch (`CounterTooLow`/`CounterTooHigh`/`GenerationTooLow`) leaves it untouched via `unit`, so `getLastPeerOrdinals` can't drift from the head of `peerRumorsR` after a restart.

`GossipDaemon`: mapping `ChainRestarted` to `true` is correct — the rumor was accepted and stored, so it should keep flowing through the fanout, and the WARN gated on `result === ChainRestarted` correctly records that a range of that origin's rumors was permanently skipped. It's a different-origin case, so it's right that it doesn't hit the self-inquiry re-offer path that's scoped to `CounterTooHigh && origin === selfId`.

One coordination note, not a blocker: the PR is titled "make fork-recovery runnable" but the fork-recovery matrix entry in the e2e workflow stays commented out (disabled on the node bug referenced), so none of the 199-line harness rewrite — the re-selection of the isolation node from the seated committee, the median-tip helper, the rewritten Phase-1 gate — actually runs in CI. The gossip fix that this harness is meant to prove is landing unverified by automation. Worth tracking what unblocks turning that entry back on so the harness work doesn't silently rot.

PR #1553, opened 2026-08-03
fix: defer ordinal 2 consensus until the initial Owner message is available

marcus-girolneto:
## Problem

A customer metagraph (BioFi) got permanently stuck on mainnet: every currency snapshot from ordinal 2 onward was rejected by the global L0, and the only recovery was rolling back to ordinal 1 and retrying.

Root cause is a startup race that is fatal because two protocol rules intersect exactly at currency snapshot ordinal 2:

1. **The initial Owner message can only be accepted at ordinal 2.** `CurrencySnapshotAcceptanceManager` only uses the lenient `validateInitialOwner` path when `snapshotOrdinal === 2 && parentOrdinal === 0`.
2. **Ordinal 2 is the first fee-charged snapshot.** Genesis (0) and the first incremental (1) are fee-exempt; from ordinal 2 on, the global L0 charges the snapshot fee to the **Owner** address (`GlobalSnapshotStateChannelEventsProcessor.processCurrencySnapshots`). With no owner in the snapshot, `maybeFeeAddress = None`, the binary is not accepted, and the metagraph chain is broken at ordinal 2 forever.

The ordinal 2 consensus round starts as soon as the node is Ready, and its event cut (`consensusStorage.getUpperBound` in the Facility declaration) races against the Owner/Staking message submission via `CurrencyMessageRoutes`. The customer's ML0 log shows the round for key 2 being created at `08:11:47,826` while the Owner message was only accepted at `08:11:47,975` — that run won the race by ~150ms of luck; the run that blocked the customer lost it.

## Fix

The currency L0 now defers facilitating the ordinal 2 round until the initial Owner message is available for inclusion, whenever fees are required:

- `ConsensusStorage`: new read-only `existsEvent(predicate)` to peek at pending events (`containsEvent` only supports exact equality).
- `CurrencySnapshotConsensusStateCreator`: `canStartOwnedConsensus` gate before creating the consensus state — if the key is ordinal 2, the last context has no owner, and `FeeCalculator.isFeeRequired` (at the last known global ordinal, conservatively "ever required" when unknown), the round is deferred with a warning log until a pending Owner `CurrencyMessageEvent` exists. The next event/time trigger re-evaluates.
- `CurrencySnapshotConsensus` / currency-l0 `Services`: wire the existing `FeeCalculator` through.

Environments without fee configs (e.g. Dev) are unaffected: `isFeeRequired` is false, so metagraphs that never send owner messages keep working as before.

With the gate, snapshot 2 deterministically carries the Owner message, the global L0 fills Owner/Staking from it, charges the fee to the owner balance, and accepts the chain — no more rollback loops.

## Note for the affected customer

Independently of this fix, their ML0 log also shows `BinaryPoster - [Queue] Empty allowance list not allowed in [Mainnet], skipping`: the deployed jar predates their allowlist registration (#1516), so binaries are never posted. They need to run a build that includes `53c5ace2b`.

## Testing

- `sbt currencyL0/compile` (includes node-shared) passes, scalafix clean.

### DIFF
--- a/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/modules/Services.scala
+++ b/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/modules/Services.scala
@@ -145,6 +145,8 @@ object Services {
         sharedStorages.lastGlobalSnapshot
       )
 
+      identifier <- storages.identifier.get
+
       consensus <- CurrencySnapshotConsensus
         .make[F](
           sharedServices.gossip,
@@ -152,6 +154,7 @@ object Services {
           keyPair,
           seedlist,
           cfg.collateral.amount,
+          identifier,
           storages.cluster,
           storages.node,
           storages.lastSyncGlobalSnapshot,
--- a/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/modules/Services.scala
+++ b/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/modules/Services.scala
@@ -155,6 +155,7 @@ object Services {
           storages.cluster,
           storages.node,
           storages.lastSyncGlobalSnapshot,
+          feeCalculator,
           maybeRewards,
           cfg.snapshot,
           client,
--- a/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensus.scala
+++ b/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensus.scala
@@ -27,6 +27,7 @@ import io.constellationnetwork.node.shared.infrastructure.metrics.Metrics
 import io.constellationnetwork.node.shared.infrastructure.node.RestartService
 import io.constellationnetwork.node.shared.infrastructure.snapshot.{CurrencySnapshotCreator, CurrencySnapshotValidator}
 import io.constellationnetwork.node.shared.snapshot.currency._
+import io.constellationnetwork.schema.address.Address
 import io.constellationnetwork.schema.artifact.SharedArtifact
 import io.constellationnetwork.schema.balance.Amount
 import io.constellationnetwork.schema.peer.PeerId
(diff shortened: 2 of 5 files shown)


marcus-girolneto:
Added a regression suite (`CurrencySnapshotConsensusStateCreatorSuite`, 9 tests passing). The gate decision (`canStartOwnedConsensus`) and the Owner-message predicate (`isOwnerMessageEvent`) are now exposed on the companion object for testability, following the pattern of the existing consensus suites. Coverage: defers ordinal 2 when fees are required and no Owner message is pending (the regression); starts once the message is pending / when the owner is already set / on other ordinals / when fees are not configured or not yet active; defers when the global ordinal is unknown but fee configs exist; and an `existsEvent` integration test against a real `ConsensusStorage`.

scasplte2:
Small deterministic hardening suggestion for the gate: make the predicate match an **initial Owner message for this metagraph**, rather than any event whose type is `Owner`.

```scala
def isInitialOwnerMessageEvent(
  metagraphId: Address
): CurrencySnapshotEvent => Boolean = {
  case CurrencyMessageEvent(message) =>
    message.messageType === MessageType.Owner &&
    message.parentOrdinal === MessageOrdinal.MinValue &&
    message.metagraphId === metagraphId
  case _ =>
    false
}
```

Then use the agreed metagraph address from the last outcome:

```scala
consensusStorage.existsEvent(
  isInitialOwnerMessageEvent(lastOutcome.finished.context.address)
)
```

Where to put it:

- In `CurrencySnapshotConsensusStateCreator.scala`, replace the current companion-object `isOwnerMessageEvent` value with the parameterized `isInitialOwnerMessageEvent` function above.
- In `tryFacilitateConsensus`, replace `consensusStorage.existsEvent(isOwnerMessageEvent)` with the metagraph-aware call above. `lastOutcome.finished.context.address` is consensus-derived, so it does not add a local source of nondeterminism.
- Update `CurrencySnapshotConsensusStateCreatorSuite` to cover an initial Owner for this metagraph, an Owner for another metagraph, and an Owner with a non-minimum parent ordinal.

Why this is useful: the current type-only predicate can be satisfied by an Owner event for a different metagraph or by a non-initial Owner message with the wrong parent ordinal. Such an event would open the gate but then be rejected by `acceptMessages`. Ordinal 2 could consequently finalize without an accepted Owner, recreating the permanent Global L0 rejection this PR is intended to prevent. This can arise from stale/unexpected gossip even though the normal local startup path validates its configured Owner message before enqueueing.

I don't think the gate needs to duplicate effectful cryptographic/address validation. The normal startup path already calls `validateInitialOwner`, while this structural check cheaply ensures the pending event belongs to this metagraph and will take the special ordinal-2 initial-owner acceptance path. It is pure and deterministic, with no new node-local inputs.

scasplte2 (approved):

PR #1554, opened 2026-08-03
fix: defer ordinal 2 consensus until the initial Owner message is available (develop)

marcus-girolneto:
Port of #1553 (release/mainnet) to develop, adapted to the EventMempool architecture, plus a regression suite.

## Problem

A customer metagraph (BioFi) got permanently stuck on mainnet: every currency snapshot from ordinal 2 onward was rejected by the global L0, and the only recovery was rolling back to ordinal 1 and retrying.

Root cause is a startup race that is fatal because two protocol rules intersect exactly at currency snapshot ordinal 2:

1. **The initial Owner message can only be accepted at ordinal 2.** Message acceptance only uses the lenient `validateInitialOwner` path when `snapshotOrdinal === 2 && parentOrdinal === 0` (`MessageValidationOpsManager` on develop).
2. **Ordinal 2 is the first fee-charged snapshot.** Genesis (0) and the first incremental (1) are fee-exempt; from ordinal 2 on, the global L0 charges the snapshot fee to the **Owner** address. With no owner in the snapshot, there is no fee address, the binary is not accepted, and the metagraph chain is broken at ordinal 2 forever.

The ordinal 2 consensus round starts as soon as the node is Ready and races against the Owner/Staking message submission. The customer's ML0 log shows the round for key 2 being created at `08:11:47,826` while the Owner message was only accepted at `08:11:47,975`.

## Fix

The currency L0 defers facilitating the ordinal 2 round until the initial Owner message is available for inclusion, whenever fees are required:

- `CurrencySnapshotConsensusStateCreator`: gate before creating the consensus state — if the key is ordinal 2, the last context has no owner, and `FeeCalculator.isFeeRequired` (at the last known global ordinal, conservatively "ever required" when unknown), the round is deferred with a warning log until a pending Owner `CurrencyMessageEvent` exists in the event mempool. The gate decision (`canStartOwnedConsensus`) and the mempool predicate (`pendingOwnerMessageExists`) are exposed on the companion object for testability, following the pattern of the existing consensus suites.
- `CurrencySnapshotConsensus` / currency-l0 `Services`: wire the existing `FeeCalculator` through.

Environments without fee configs (e.g. Dev) are unaffected: `isFeeRequired` is false, so metagraphs that never send owner messages keep working as before.

## Regression tests

`CurrencySnapshotConsensusStateCreatorSuite` (8 tests, all passing):

- defers the ordinal 2 round when fees are required and no Owner message is pending (the regression)
- starts the ordinal 2 round once an Owner message is pending
- starts rounds other than ordinal 2 regardless of pending messages
- starts the ordinal 2 round when the owner is already set in the last context
- starts the ordinal 2 round when fees are never required (no fee configs)
- starts the ordinal 2 round when the fee config only activates at a later global ordinal
- defers the ordinal 2 round when the global ordinal is unknown but fee configs exist
- `pendingOwnerMessageExists` finds a pending Owner message in a real `EventMempool` and ignores Staking-only content

## Testing

- `sbt currencyL0/compile` passes
- `sbt "currencyL0/testOnly ...CurrencySnapshotConsensusStateCreatorSuite"`: Total 8, Failed 0

### DIFF
--- a/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/modules/Services.scala
+++ b/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/modules/Services.scala
@@ -173,6 +173,7 @@ object Services {
           storages.cluster,
           storages.node,
           storages.lastSyncGlobalSnapshot,
+          feeCalculator,
           maybeRewards,
           cfg.snapshot,
           cfg.environment,
--- a/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensus.scala
+++ b/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensus.scala
@@ -24,6 +24,7 @@ import io.constellationnetwork.node.shared.domain.gossip.Gossip
 import io.constellationnetwork.node.shared.domain.node.NodeStorage
 import io.constellationnetwork.node.shared.domain.rewards.Rewards
 import io.constellationnetwork.node.shared.domain.snapshot.storage.LastSyncGlobalSnapshotStorage
+import io.constellationnetwork.node.shared.domain.statechannel.FeeCalculator
 import io.constellationnetwork.node.shared.http.p2p.PeerResponse
 import io.constellationnetwork.node.shared.infrastructure.consensus._
 import io.constellationnetwork.node.shared.infrastructure.consensus.engine._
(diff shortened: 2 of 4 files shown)


ryle-ai (commented):
Checked the ordinal-2 kickoff gate in CurrencySnapshotConsensusStateCreator. `canStartOwnedConsensus` only defers when `key == InitialOwnerMessageOrdinal (2)` and the parent snapshot carries no owner address; every other key and any owner-present state short-circuits to `true`, and the mempool scan is suspended in `F` so non-ordinal-2 rounds never pay for it. Fee-required is keyed on the last global snapshot ordinal via `feeCalculator.isFeeRequired`, which is the right dimension since global L0 charges the owner from ordinal 2 on; the `Long.MaxValue` fallback for an unknown global ordinal resolves to "defer if any non-noFee config exists" (rangeTo covers all keys), a conservative default that can only over-defer, never race. Confirmed `feeCalculator`/`eventMempool` are threaded from Services into both CurrencySnapshotConsensus.make and the state creator. The remaining consideration is liveness: on a fee network where the owner message never lands, ordinal 2 defers indefinitely — but that is recoverable (the round starts the moment a pending Owner event appears, and a peer holding the message pulls others in as facilitators) versus the unrecoverable global-L0 wedge the deferral prevents, so it is the correct trade. The 8-case suite covers the key!=2, owner-present, no-fee, fee-with/without-pending-message, and unknown-global-ordinal branches.

scasplte2:
Closing as superseded by #1566, which already includes the stronger develop implementation of the ordinal-2 Owner gate, including correct-metagraph and initial-parent validation plus regression coverage. The release/mainnet counterpart was merged in #1553.

PR #1555, opened 2026-08-03
fix: harden delegated stake withdrawal processing

scasplte2:
## Summary

- Enforces per-token-lock consistency across delegated stake replacement and withdrawal processing.
- Preserves historical replay through an ordinal-gated transition.
- Applies withdrawal-driven unlock effects once per token lock.
- Adds boundary regression tests and CI end-to-end coverage.

## Rollout

The MainNet transition begins at global snapshot ordinal `6,710,600`. Global L0 nodes should be updated before activation.

## Testing

- Focused node-shared tests: 8 passed.
- `dagL0/Test/compile` passed.
- `dagL1/Test/compile` passed.
- `currencyL1/Test/compile` passed.
- Full node-shared suite: 225 passed; 2 existing `GenesisFSSuite` Java 21 module-access errors.
- Delegated-staking CI script syntax check passed.


ryle-goehausen (approved):

ryle-ai (commented):
Checked the double-withdrawal hardening end to end against the activation gate.

- `UpdateDelegatedStakeAcceptanceManager`: the new rejection path is only armed when `currentSnapshotOrdinal >= fixingDelegatedStakeDoubleWithdrawalOrdinal`, and the precedence is sound — a create accepted this snapshot for the same `(source, tokenLockRef)` yields `AlreadyWithdrawn(stakeRef)`, a pending withdrawal on the lock yields `AlreadyWithdrawn(tokenLockRef)`, then `DuplicatedStake`/`DuplicatedTokenLock` on the seen-sets. `tokenLockRefsSeen` is threaded through the fold so a second withdrawal against the same lock in one snapshot is caught. The three error types resolve in `UpdateDelegatedStakeValidator`.
- `GlobalSnapshotAcceptanceManager.generateTokenUnlocks`: the pre-activation branch is byte-identical to the removed instance method, so historical replay is unchanged. The post-activation branch dedups by `tokenLockRef` and regroups by `TokenUnlock.source`. Confirmed that key aligns with the two consumers — `acceptTokenLocks` and `updateGlobalBalancesByTokenLocks` both index the generated-unlock map by the token lock's address — so principal removal and crediting land on the lock owner. The address→source shift only diverges when a stake's source differs from its lock's source, and in that case crediting the lock owner is the correct party for principal return, not a regression.
- Boundary is consistent (`>=` gate, `<` for the historical branch). `forProduct11` lists all eleven `FieldsAddedOrdinals` keys with the new one last, matching field order; `SharedServices` resolves the ordinal via `getOrElse(environment, MinValue)` with mainnet future-gated at 6710600 and dev at 0. The added implicit `Hasher[F]` on `accept` is additive and satisfied at the in-repo call site.

scasplte2:
Closing as superseded.

- #1592 is the fresh, minimal release/mainnet replacement.
- #1593 is the v4.1/develop implementation using current effective token-lock semantics.
- #1595 extracts the attack-shaped delegated withdrawal/replacement E2E as a separate test-only PR.

The activation ordinal proposed here is already in signed MainNet history, and this branch no longer applies cleanly to current release/mainnet.

PR #1556, opened 2026-08-04
ci: run E2E on self-hosted Hetzner runners instead of Ubuntu-22-64-core

effects-ai:
This points those jobs at a self-hosted `tessellation-e2e` label backed by Hetzner Cloud.

The only change to existing code is one line of `runs-on` in `.github/workflows/e2e-just-test.yml`. Everything else is new, under `deploy/ci-runners/`.

## Validation — all 11 jobs pass on Hetzner

Full matrix run end-to-end on a real Hetzner box (validated on a fork, since repo-level runner registration needs repository *admin*):

| Group | Containers | Duration |
|---|---|---|
| `dag-cluster` | 7 | 3m41s |
| `rewards` | 15 | 4m01s |
| `data-with-fee` | 15 | 4m18s |
| `token-lock-replacement` | 15 | 5m16s |
| `delegated-staking` | 15 | 5m36s |
| `token-locks` | 15 | 6m27s |
| `currency` | 15 | 7m58s |
| `spend` | 15 | 14m14s |
| `allow-spends` | 15 | 18m29s |
| `snapshot-streaming` | — | 4m58s (unchanged, GitHub-hosted) |
| **total E2E work** | | **~70 min** |

Genuine consensus, not just cluster formation — e.g. `Cluster Global L0 with size 3 (>= 3 expected)`, metagraph ordinals advancing 9→10 and 11→12, and `Allow-spend tests completed took: 833 seconds`.

## Two implementations of the same label

Both register under `tessellation-e2e`, so this workflow is agnostic to which is serving, and they can run simultaneously.

| | `autoscaled/` | `fixed/` |
|---|---|---|
| Shape | one ephemeral server per job | N always-on servers, 1 runner each |
| Concurrency | elastic | `runner_count`, hard |
| PR wall-clock | ~25 min (unchanged) | ~23 min at 3 runners |
| **Cost** | **~€353/mo (81%)** | ~€976/mo (47%) |
| Dedicated Hetzner project | **required** | not required |
| Moving parts | controller service (SPOF) | none |
| State between jobs | none | persists — needs cleanup hooks |

`autoscaled/` is the better answer: CI duty cycle is only ~9% (~198 job-hours/month), so always-on capacity is structurally wasteful. `fixed/` exists because the autoscaler requires a dedicated Hetzner project (it deletes powered-off servers in whatever project it's given, so it must never share one with the `testnet-*` chain nodes), and Hetzner projects can only be created in the Console.

## Why one job per server

The E2E harness is **not multi-tenant**. `docker/bin/compose-runner.sh:174` creates a fixed-name network `tessellation_common` on a fixed subnet (`NET_PREFIX`, default `172.32.0.0/24`), with fixed container names (`gl0-0`, `gl1-0`, …) and fixed host ports (9000–9412). Two concurrent jobs on one host collide on all four.

One job per server gets that isolation for free — **no docker-in-docker and no changes to the test harness**. It also avoids deliberately co-tenanting a suite that `docker/docker-compose.test.yaml` already documents as timing-fragile under multi-JVM contention.

## Sizing: `ccx43`, and why `ccx33` is rejected

`ccx33` (8 vCPU / 32 GB) passed once, then failed **twice in two distinct ways** on repeat runs — both resource exhaustion, neither a real defect:

- **Memory** — p90 **99%**, peak **100%** of 32 GB, with **no swap**. The kernel OOM-killer killed the Actions runner *agent* during `allow-spends` (`Out of memory: Killed process (node)`); the systemd unit went `failed` and the remaining 8 jobs hung in `queued` forever.
- **CPU** — peak load **15.42 on 8 cores (193%)**; 16% of active samples exceeded 1.0/core. Under that contention GL0's `/global-snapshots/latest/combined` returned **HTTP 503** and `spend` failed.

`ccx43` (16 vCPU / 64 GB) puts those peaks at ~42% memory and ~96% load, with only **0.7%** of samples above load 16. CCX (dedicated vCPU) throughout, never CPX (shared), for the timing reasons above.

> ⚠️ **Deployment prerequisite:** `ccx43` is 16 dedicated cores and a default Hetzner project allows **8**, so creation returns `resource_limit_exceeded` until a quota increase is requested.

## What the validation surfaced in this repo

Four assumptions `e2e-just-test.yml` makes about GitHub's hosted runner image, none of which hold on a stock Ubuntu box. All are now handled in cloud-init/hooks, but they're worth knowing:

1. `/usr/local/bin` must be **writable by the unprivileged runner user** — "Install just" pipes the installer with `--to /usr/local/bin` and no `sudo`.
2. **`libatomic1`** must be present, or any Node.js exits with `error while loading shared libraries: libatomic.so.1`.
3. **`~/.nvm` must exist** — `install_dependencies.sh` `check_node()` tests for the *directory*, not for `node` on `PATH`.
4. The install steps are **not re-runnable**: the `just` installer aborts on "already exists", and `gpg --dearmor` tries to prompt with no tty.

### Two pre-existing bugs found (not fixed here)

- **`currency-l1` has an intermittent scalac crash** during `phase: jvm` — hit **3 of 7** builds, moving between `Engine.scala` and `HttpApi.scala` with `assertion failed: type R` / `type S`. Order-dependent codegen, likely worth trying `-Ybackend-parallelism 1`. This costs real money today, since every flaked build is billed.
- **`workflow_dispatch` is broken on this workflow**: the dispatch input defaults `snapshot_streaming_branch` to `release/mainnet` while the `pull_request` fallback is `release/testnet`. Since `develop` dropped `StateProofValidator`, manual "Run workflow" always fails with 33 compile errors. One-line fix, deliberately left out of this PR to keep it focused.

## Rollback

`runs-on` reads `vars.E2E_RUNNER_LABEL` first, so setting that repository variable to `Ubuntu-22-64-core` moves all E2E back to GitHub-hosted runners on the next job — **no PR, no rerun**. Worth exercising once before the first live run.

## Review notes

- No secrets committed. `.gitignore` covers `terraform.tfstate`, `tfplan` (which embeds variable values), `backend_override.tf`, and `*.auto.tfvars`.
- Both Terraform stacks use **separate state keys** and are independent of `deploy/terraform` (the hypergraph cluster).
- The `build` job and the `snapshot-streaming` group stay on GitHub-hosted `ubuntu-22.04` — cheap and unaffected.
- Access controls are **fail-closed**: empty `allowed_ssh_cidrs` exposes no inbound port.

### DIFF
--- a/deploy/ci-runners/fixed/register-runners.sh
+++ b/deploy/ci-runners/fixed/register-runners.sh
@@ -1,0 +1,189 @@
+#!/usr/bin/env bash
+#
+# register-runners.sh — install and register the GitHub Actions runner on each
+# server in the fixed pool.
+#
+#   deploy/ci-runners/fixed/register-runners.sh <ip1,ip2,ip3>
+#   deploy/ci-runners/fixed/register-runners.sh "$(terraform -chdir=deploy/ci-runners/fixed/terraform output -raw runner_ips_csv)"
+#
+# Idempotent: re-run to roll out a hook change, upgrade the runner, or re-register
+# after a token rotation. An already-registered runner is unregistered and
+# re-registered cleanly rather than duplicated.
+#
+# Always required:
+#   GITHUB_REPOSITORY  e.g. Constellation-Labs/tessellation  (not a secret)
+#
+# AUTHORIZATION — pick ONE. GitHub will not let a machine register without proof
+# of authorization, but you do not need a long-lived credential for it:
+#
+#   (a) NO-PAT PATH (preferred when PAT policy / SSO is in the way)
+#       REG_TOKENS   comma-separated per-runner registration tokens, in the same
+#                    order as the IPs. Copy each from the repo UI:
+#                      Settings -> Actions -> Runners -> New self-hosted runner
+#                      -> Linux, then take the value after `--token` in the
+#                      displayed ./config.sh command.
+#                    One token per runner. They expire after ~1 hour, are
+#                    single-use, and grant nothing but "join this repo".
+#
+#   (b) PAT PATH (convenient for repeat runs / automation)
+#       GITHUB_TOKEN GitHub CLASSIC PAT with `repo` scope. Used ONLY here, on the
+#                    operator's machine, to mint the same short-lived per-host
+#                    registration tokens automatically. The PAT itself is never
+#                    copied to the runners.
+#
+# Optional:
+#   RUNNER_LABELS      default: tessellation-e2e  (must match the workflow's runs-on)
+#   RUNNER_VERSION     default: 2.336.0
+#   SSH_USER           default: admin
+#
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+SSH_USER="${SSH_USER:-admin}"
+RUNNER_LABELS="${RUNNER_LABELS:-tessellation-e2e}"
+RUNNER_VERSION="${RUNNER_VERSION:-2.336.0}"
+
+die() { echo "ERROR: $*" >&2; exit 1; }
+log() { echo "==> $*"; }
+
+TARGETS="${1:-}"
+[ -n "$TARGETS" ] || die "usage: $0 <ip1,ip2,...>"
+[ -n "${GITHUB_REPOSITORY:-}" ] || die "GITHUB_REPOSITORY must be set in the environment"
+
+command -v curl >/dev/null || die "curl is required"
+
+IFS=',' read -ra HOSTS <<< "$TARGETS"
+
+# Decide the authorization mode up front, so a misconfiguration fails before we
+# touch any server rather than half-way through the pool.
+REG_TOKEN_LIST=()
+if [ -n "${REG_TOKENS:-}" ]; then
+  AUTH_MODE="pre-minted"
+  IFS=',' read -ra REG_TOKEN_LIST <<< "$REG_TOKENS"
+  if [ "${#REG_TOKEN_LIST[@]}" -ne "${#HOSTS[@]}" ]; then
+    die "REG_TOKENS has ${#REG_TOKEN_LIST[@]} token(s) but ${#HOSTS[@]} host(s) were given — supply one token per runner, in IP order"
+  fi
+  log "Authorization: pre-minted registration tokens (no PAT)"
+elif [ -n "${GITHUB_TOKEN:-}" ]; then
+  AUTH_MODE="pat"
+  log "Authorization: classic PAT (registration tokens will be minted per host)"
+else
+  die "set either REG_TOKENS (one UI-copied registration token per runner, comma-separated) or GITHUB_TOKEN (classic PAT with 'repo' scope) — see the header of this script"
+fi
+
+# Resolve the runner tarball checksum once, from the release GitHub publishes, so
+# we verify the download on every host without hardcoding a hash that silently
+# rots when RUNNER_VERSION is bumped.
+log "Resolving runner v${RUNNER_VERSION} checksum"
+RUNNER_FILE="actions-runner-linux-x64-${RUNNER_VERSION}.tar.gz"
+RUNNER_SHA="$(curl -sSL \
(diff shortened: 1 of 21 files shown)


deploy/ci-runners/fixed/register-runners.sh
```diff
@@ -0,0 +1,247 @@
...
+sudo bash -c "cd '${RUNNER_DIR}' && ./svc.sh status" | head -12
+echo "--- restart policy ---"
+systemctl show "$UNIT" -p Restart -p RestartUSec -p OOMPolicy 2>/dev/null
+REMOTE
+
+  log "[$NAME] done"
+done
+
+log "All runners registered with labels: ${RUNNER_LABELS}"
+log "Verify at: https://github.com/${GITHUB_REPOSITORY}/settings/actions/runners"
```
ryle-ai: This final "Verify at" line uses `${GITHUB_REPOSITORY}`, but everywhere else the script treats `GITHUB_TARGET` as canonical — line 62 resolves `GITHUB_TARGET="${GITHUB_TARGET:-${GITHUB_REPOSITORY:-}}"` (so `GITHUB_REPOSITORY` is only an optional alias) and line 71 already computes `REGISTER_URL="https://github.com/${GITHUB_TARGET}"`. If an operator sets only `GITHUB_TARGET` (the primary variable documented in the header), this echo prints `https://github.com//settings/actions/runners` with an empty path segment. Reusing `$REGISTER_URL` (or `$GITHUB_TARGET`) here fixes it and matches the org-scope case too. Non-blocking.

deploy/ci-runners/fixed/register-runners.sh
```diff
@@ -0,0 +1,247 @@
...
+    log "[$NAME @ $HOST] using pre-minted registration token"
+  else
+    log "[$NAME @ $HOST] minting a registration token"
+    # Short-lived (1 h) and single-purpose. Minted per host so a failure part-way
+    # through the loop doesn't leave a stale shared token lying around.
+    REG_TOKEN="$(curl -sSL -X POST \
+      -H "Authorization: Bearer ${GITHUB_TOKEN}" \
+      -H "Accept: application/vnd.github+json" \
+      "https://api.github.com/${TOKEN_PATH}/actions/runners/registration-token" \
+      | python3 -c 'import json,sys; print(json.load(sys.stdin).get("token",""))')"
```
ryle-ai, line 122: The PAT path mints the registration token by piping the API response through `python3`, but the preflight only checks `command -v curl` (line 74). The rest of the script is careful to fail early with a clear message (curl presence, `REG_TOKENS` count vs host count, `GITHUB_TARGET` set), so a missing `python3` here is the one dependency that would surface as a cryptic parse failure mid-run instead of a clear die. Adding `python3` to the same preflight check would keep the fail-fast behavior consistent. Only affects the PAT path; the pre-minted `REG_TOKENS` path doesn't need it. Non-blocking.

ryle-ai (commented):
Reviewed the E2E runner migration end to end. The workflow flip is safe and reversible: `runs-on: ${{ matrix.test-group.runner || vars.E2E_RUNNER_LABEL || 'tessellation-e2e' }}` keeps per-group overrides working and falls back cleanly, and the documented rollback (set repo var `E2E_RUNNER_LABEL=Ubuntu-22-64-core`) needs no code change. The two Terraform variants are internally consistent: the `fixed/` pool references existing SSH keys via `data "hcloud_ssh_key"` + `ssh_key_names` (avoiding the 409 uniqueness_error in the shared project) while `autoscaled/` creates its own keys in a dedicated project it is allowed to enumerate/delete. One-runner-per-server for `fixed/` is justified — the E2E harness hardcodes the `tessellation_common` network and a fixed subnet/ports, so two runners on one box would collide. Firewalls fail closed on inbound SSH when `allowed_ssh_cidrs` is empty, the job hooks intentionally `exit 0` so cleanup never fails a job, and the `Restart=always`/`OOMPolicy=continue` systemd override is the right call for a box the E2E suite can OOM.

Two minor, non-blocking consistency nits in `fixed/register-runners.sh` noted inline. Nothing here blocks merge.

ryle-goehausen (dismissed):

ryle-goehausen (dismissed):

ryle-goehausen (approved):

PR #1558, opened 2026-08-05
fix(consensus): restore broad Tier-1 signing and bounded admission

scasplte2:
## Why

IntegrationNet on v4.1.0-rc.3 can ratchet its signing committee down from roughly 10 members to 4-6 even while roughly 96 peers remain Ready. Score exclusion removes signing seats faster than open admission restores them, the nominal per-round admission cap can be bypassed across monitor ticks, and deterministic ascending PeerId selection concentrates successful admission certificates and rewards.

The intended model is that Core gates liveness while healthy Tier-1 peers continue signing and earning. Only peers that demonstrably reduce network health should lose finality participation and rewards.

## What changed

- Retain healthy selected signing leases as Tier-1 instead of deleting their seats when they fall outside Core.
- Carry one deterministic, entropy-ranked open-admission nominee in the proposal so Core voters converge on the same target.
- Enforce a true remaining per-round admission-vote budget.
- Restore the existing five-round cadence for open ReadyAtTip expansion using activeAdmissionExpansionIntervalRounds; penalty and probation readmission remain separate recovery lanes.
- Gate each local open-admission vote on the exact next-seat finality invariant: observed parent signers from the current committee must satisfy the finality floor for committee size plus one.
- Require three consecutive locally observed proof misses before emitting a Tier-1 eviction vote, reusing TierTransitions.DemotionConsecutiveMisses and resetting on any observed signature.
- Keep Core-quorum admission and eviction certificates as the membership authority.
- Validate admission-certificate quorum by unique signer PeerIds in both DAG L0 and Currency L0.
- Preserve legacy certificate ordering at state-application sites for mixed-version defensive stability.
- Report the actual Core plus Tier-1 finality requirement and expose admission headroom decisions.
- Document the design in ADR-0030 and add a v4.1 rollout and recovery runbook.

There is no arbitrary hard cap on the reward population. Growth is paced by the existing five-round cadence and pauses whenever observed signing participation cannot safely absorb the next seat.

## Scope and compatibility

- This is a consensus proposal and behavior change intended for a coordinated full-cluster deployment.
- The proposal message uses schema v34 for the optional canonical nominee.
- GlobalIncrementalSnapshot, GlobalSnapshotInfo, and GlobalSnapshotStateProof schemas and proof calculation are unchanged, so metagraph snapshot validation remains compatible.
- No new activation ordinal or configuration field is introduced.
- Broad signing-member retention is GL0-only. Currency L0 remains bounded by activeAdmission.active while sharing the safe canonical selection and unique-voter validation changes.

## Verification

- nodeShared tests: 1,038 passed
- DAG L0 tests: 132 passed, 2 ignored
- Currency L0 tests: 37 passed
- Focused admission and tiering suites: 31 passed
- scalafmtCheckAll passed
- git diff --check passed

## Rollout

Use the coordinated cold-restart procedure and metric gates in docs/operations/v4.1-tier1-admission-rollout.md. In particular, watch finality margin, admission headroom, certificate formation, committee growth, view changes, and round duration before promoting the release.


ryle-ai (commented):
Traced the shrink-ratchet fix end to end and the committee-mutation path stays safe. A seat is only added when a Core-quorum `AdmissionCertificate` targeting that peer rides inside an accepted Proposal, and the two hardening edges hold up:

- Certificate quorum now counts `uniqueVoterCount` (distinct `proofs.head.id.toPeerId`) rather than `cert.votes.size` on both advancers, so multiple differently-encoded votes from one signer can't manufacture a quorum. Dedup undercounts rather than over, which is the fail-closed direction.
- `openAdmissionTargets` ranks nominees by `FacilitatorSelector.orderByScore(expectedTip)` where `expectedTip` is the accepted parent snapshot hash — identical entropy across every Core voter, so target selection is deterministic cluster-wide. The `.take(budget)` before `.filterNot(alreadyVotedBySelf)` ordering is load-bearing: it pins the per-round target set instead of letting the cursor walk down the ranking across ticks. Open-lane budget is correctly separated from the probation lane via `alreadyVotedBySelf -- probation`, and per-round `admissionVotes` scoping keeps `openVotesAlreadyUsed` in {0,1} against a budget of 1.
- Headroom gate (`OpenAdmissionPolicy`, `headroomGateActive = !bootstrapActive`) blocks admitting an (n+1)th seat when observed current-committee parent signers can't cover the resulting finality floor; bootstrap is exempted deliberately (legacy Core-only gate, and an added Tier-1 seat can't raise it). Currency L0 leaves `locallyObservedParentSigners` absent because unanimity can't prove an unseated signer — that asymmetry is intentional, not a gap.
- Cadence gating (`certificateAllowed` / `expansionAllowedAtOrdinal`) is wired identically on both L0s, and `cadenceSuppressed` certs are dropped rather than applied off-interval.
- Tier-1 finality audit (`FinalityParticipationAuditor`) resets the streak on any local proof gap before the 3-miss `EvictionVote(Silent)`, so it errs toward not evicting.

Even where leader nominee selection (`AdmissionNomineeSelector.select(state.entropy)`) and the voter's re-ranking could in principle diverge, budget=1 collapses `canonicalNominees` to a single peer, and any disagreement fails closed (no shared target → no quorum → no seat). No blocking issue found. Files checked: Global/Currency `*StateAdvancer` + `*StateCreator`, `StallDetector.scala` (`maybeEmitAdmissionVotes` + `openAdmissionTargets`), `OpenAdmissionPolicy`, `AdmissionNomineeSelector`, `FinalityParticipationAuditor`, `ActiveFacilitatorAdmission`, `TierTransitions`, config/types + dag-l0.conf.

ryle-goehausen (approved):

ryle-ai (commented):
Reviewed the broad-Tier-1 / bounded-admission change on both L0s. The change is fail-closed throughout and I found no defect at the bar.

Invariants confirmed on the current tip:
- A seat is added only when a Core-quorum `AdmissionCertificate` rides inside an accepted Proposal; quorum counts `AdmissionCertificate.uniqueVoterCount` (distinct `proofs.head.id` PeerId) rather than `cert.votes.size` on both advancers, so duplicate/differently-encoded votes from one signer undercount rather than over — the fail-closed direction.
- `openAdmissionTargets` ranks by `FacilitatorSelector.orderByScore(expectedTip)` with `expectedTip` = accepted parent snapshot hash (identical entropy across every Core voter), and `.take(budget)` runs before `.filterNot(alreadyVotedBySelf)` so the per-round target set is pinned deterministically instead of the cursor walking the ranking across ticks. With budget=1 and a leader-carried canonical nominee, any leader/voter rank divergence simply fails the quorum (no shared target -> no seat), never forges one.
- Headroom gate (`OpenAdmissionPolicy`, `headroomGateActive = !bootstrapActive`) blocks an (n+1)th seat when observed parent signers can't cover the resulting finality floor; bootstrap exemption and Currency L0 omitting `locallyObservedParentSigners` are intentional asymmetries, not gaps.
- Cadence gating (`certificateAllowed` / `expansionAllowedAtOrdinal`) is wired identically on both L0s; off-interval certs are dropped, not applied.
- `FinalityParticipationAuditor` resets the miss streak on any local proof gap before the 3-miss `EvictionVote(Silent)`, erring toward not evicting.

The fields-added-ordinals / config additions and the consensusSchemaVersion bump line up with the code, and the new suites exercise the certificate-quorum, nominee-selection, and finality-audit paths. Consistent with the detailed trusted review already on the PR; nothing further to add.

scasplte2:
deployed on integrationnet as 4.1.0-rc.4, working as expected for the moment. Continuing testing at scale

PR #1561, opened 2026-08-06
fix(currency-l0): allow coordinated solo rollback recovery

scasplte2:
## Summary

- add the opt-in Currency L0 `run-rollback --allow-solo-consensus` recovery override
- seed the post-rollback facilitator and eligible sets with only the designated recovery node when explicitly enabled
- preserve the existing proof-signer rollback behavior and ordering when the flag is absent
- mark Currency L0 `run-validator` startup as validator mode, matching DAG L0 and blocking accidental solo production from a transient one-peer view
- add rollback bootstrap warnings, mode/count metrics, focused regression coverage, an operator release note, and a coordinated recovery runbook

## Why

A fully stopped metagraph can deadlock during recovery. Normal rollback inherits the checkpoint's proof-signer committee, so one restarted signer cannot finalize the next ordinal. Returning validators cannot help until their `tip + 4` observation window completes, but those snapshots cannot exist while consensus is stopped.

The override lets exactly one coordinated rollback node produce the missing snapshots. Other nodes then rejoin through ordinary `run-validator` and the existing quorum-certified admission path.

## Safety and compatibility

- The flag defaults to `false`; default signer selection is an exact re-expression of the existing behavior.
- This changes no snapshot/state-proof schema, signed consensus message, deterministic configuration hash, or activation ordinal.
- The flag is an operational input and intentionally changes the new post-rollback facilitator history when used.
- It must be used on exactly one node for exactly one rollback invocation.
- It must never be persisted in systemd, container, deployment, or monitoring restart configuration. The runbook requires automatic restarts to be disabled during recovery.
- The same jar should be deployed to all metagraph nodes; only the designated producer receives the one-shot flag.

See `docs/operations/currency-l0-solo-rollback.md` and `docs/release/currency-l0-solo-rollback.md`.

## Verification

- `currencyL0/testOnly io.constellationnetwork.currency.l0.CurrencyL0AppSuite`: 6 passed, 0 failed
- `currencyL0/test`: 43 passed, 0 failed
- `git diff --check`: clean

The focused coverage verifies default compatibility, explicit CLI opt-in, forced self-only selection, quorum-one certificate formation through the singleton witness-pool filter, and post-certificate 1 -> 2 -> 3 active/Core growth using IntegrationNet-derived admission sizing.

A full stopped-cluster recovery E2E is intentionally left to the deployed staging metagraph before operational reliance.

### DIFF
--- a/docs/consensus/README.md
+++ b/docs/consensus/README.md
@@ -287,6 +287,19 @@ The "always-handled" cases (rumors, `CheckUpdate`, `PeerObserved`, and the cert 
 
 The handler also force-completes a stuck round when an `InitializeFromDownload` arrives in `Observing` while still BUSY (recovery's stale-round escape hatch); under any other node state it re-queues the command after 1s without blocking the event loop.
 
+### Currency L0 rollback committee recovery
+
+Currency L0 rollback normally initializes the next consensus outcome from the
+rolled-back snapshot's proof signers. A fully stopped metagraph can use
+`run-rollback --allow-solo-consensus` on exactly one coordinated recovery node
+to bootstrap progress before the other nodes rejoin as ordinary validators.
+The flag is off by default and two isolated uses can create conflicting
+histories. See the
+[Currency L0 single-node rollback recovery runbook](../operations/currency-l0-solo-rollback.md)
+for the compatibility boundary, committee-regrowth trace, metrics, and rollout
+procedure, and distribute the corresponding
+[operator release note](../release/currency-l0-solo-rollback.md) before use.
+
 ### Round-Blocked States
 
 > **Note (v2 change):** The FSM blocks round starts when the node is in recovery or leaving states to prevent infinite loops.
--- a/docs/operations/currency-l0-solo-rollback.md
+++ b/docs/operations/currency-l0-solo-rollback.md
@@ -1,0 +1,211 @@
+# Currency L0 single-node rollback recovery
+
+`currency-l0 run-rollback --allow-solo-consensus` is an opt-in recovery escape
+hatch for a fully stopped metagraph. It lets exactly one rollback node seed the
+next Currency L0 consensus outcome with itself instead of inheriting every proof
+signer from the checkpoint.
+
+This is an operational override, not a normal startup mode. Running it on two
+isolated nodes can produce two valid but conflicting histories.
+
+> **DANGER — never persist this flag.** Do not add
+> `--allow-solo-consensus` to a systemd unit, container entrypoint, deployment
+> manifest, environment variable, or monitoring/automatic-restart command. It
+> is authorized for one manually coordinated rollback invocation only. An
+> automatic restart that repeats the override while nodes are isolated can
+> create a competing history. Disable automatic restarts before recovery and
+> restore the normal command, without the flag, before re-enabling them.
+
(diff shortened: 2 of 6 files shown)


ryle-ai (commented):
Checked the coordinated solo-rollback recovery path in currency-l0. `rollbackBootstrapFacilitators` resolves to the proof-signer set when this node signed the rollback snapshot, self-only when it did not (fallback), and self-only when `allowSoloConsensus` is set; order is preserved. The resolved `bootstrapFacilitators` feeds both `Facilitators` and `EligibleFacilitators` on the seeded `CurrencyConsensusOutcome`, so the recovery committee and eligibility set agree. RunRollback deliberately does not enter validator mode (it is the single producing node); the new `setValidatorMode` on the RunValidator / RunValidatorWithJoinAttempt startup paths pairs with the `ConsensusRoundRunner` solo-block guard (isValidator && facilitatorCount <= 1 && selfIsFacilitator aborts and emits `dag_consensus_validator_solo_blocked`), so every non-rollback node follows the rollback producer rather than forking. Confirmed parity with the DAG-L0 validator-mode wiring. The `allow-solo-consensus` CLI flag is default-off, requires an explicit flag, logs the DANGER warning only when set, is runtime-only, and the RunRollback opts tuple keeps 13-arity `mapN(RunRollback.apply)`. Bootstrap-mode metrics (counter by mode + proof-signer / facilitator-count gauges) and the suite's helper-mode, CLI default, singleton-certificate, and below-emergency-floor regrowth cases line up with the code.

ryle-goehausen (approved):

ryle-ai (commented):
Checked the solo-rollback escape hatch end to end. `rollbackBootstrapFacilitators(nodeId, signers, allowSoloConsensus)` collapses to `List(nodeId)` under either the forced flag or the pre-existing non-signer fallback, else keeps the proof-signer set; the three `bootstrapMode` labels line up with those branches. The `dag_consensus_rollback_*` metric names satisfy the MetricKey regex, the `mode` label satisfies LabelName, and `Metrics[IO]`/`Eq[List[PeerId]]` for the `===` comparison are both in scope via `import nodeShared._` and `cats.syntax.all._` at the rollback site.

The fork-safety question — whether a `{self}` bootstrap could wedge or diverge — resolves through the validator-mode wiring: RunValidator/RunValidatorWithJoinAttempt now call `setValidatorMode` for DAG-L0 parity, while RunRollback deliberately does not, so `ConsensusRoundRunner`'s `isValidator && count <= 1 && selfIsFacilitator` solo guard does not fire on the coordinated recovery path. That exemption is the intended mechanism, gated behind an explicit default-off `--allow-solo-consensus` flag and a DANGER log.

CLI flag defaults off and threads through the 13-arity `mapN`. Test suite exercises order preservation, self-only fallback, forced self-only, flag default, singleton admission certificate, and certified-peer retention below the IntegrationNet floor — all referenced symbols (`applyCertifiedAdmissions`, `chooseActive`, `AdmissionSizing`/`AdmissionInput`, `QuorumPolicy.fromFraction`, `CommitteeBuilder.build`, `AdmissionCertificateBuilder.build`) exist on-branch. peerHistory seed logic mirrors dag-l0 with the accepted one-round drift noted in-code.

PR #1563, opened 2026-08-10
fix(consensus): stabilize Tier-1 membership under load

scasplte2:
## Context

Follow-up to #1558 after the IntegrationNet rc.4 activation and load test.

The load test showed that peer health and signing membership had become decoupled: most of the cluster remained Ready while the Core + Tier-1 signing/reward committee contracted sharply. The initial stabilization fixed the immediate retention/readmission defects, but review identified three remaining churn risks:

- admission and silent eviction were boolean complements, allowing adjacent-size oscillation at finality-floor boundaries;
- proactive silent eviction could run every round while open admission ran only once every five rounds; and
- a three-round miss streak compressed to seconds under EventTrigger load.

The same investigation also found an independent recovery handoff defect. One node converged application storage at ordinal 5,880,534, accepted consensus outcome 5,880,535, then attempted to persist 5,880,536 against the stale application head. The nearby DownloadDaemon cancellation happened after successful convergence and was not the cause.

### Broad Tier-1 membership and readmission

- Preserve selected parent signing leases as Tier 1 instead of deleting seats solely because they fell below the Core score threshold.
- Keep countdown-zero peers in probation until an accepted AdmissionCertificate clears membership.
- Prioritize certified probation recovery over open expansion when both compete for the bounded proposal budget.
- Keep entropy ranking at proposal construction while retaining legacy certificate ordering at state application for version-stable defense in depth.
- Keep Currency L0 on its bounded active-set policy while sharing the readmission and certificate-safety corrections.

### Headroom and eviction stability

Use an exact, protocol-derived three-state policy for observed signer count S and finality floor F:

```text
expand: S >= F(n + 1)
hold:   F(n) <= S < F(n + 1)
evict:  S < F(n)
```

- The neutral hold band prevents the observed class of boundary oscillation.
- Proactive Tier-1 silent-audit votes now use the existing five-round expansion cadence.
- Penalty/probation readmission and certified Core stall eviction remain separate every-round recovery lanes.
- Silent eviction still requires three consecutive local proof misses, and those misses must now span at least `(DemotionConsecutiveMisses - 1) * timeTriggerInterval` (86 seconds with IntegrationNet's current values).
- Proof subsets, monotonic time, and cadence only control local vote emission. Membership still changes solely through existing Core-quorum certificates.

### Recovery handoff

- Classify exact, newer-key, and invalid same-key outcome handoffs explicitly.
- Before Global L0 accepts a newer consensus outcome, align Last-N storage, LastSnapshot storage, snapshot head, MPT state, and already-committed mempool events to the accepted artifact/context.
- Record the accepted key rather than the stale requested key in the recovery marker.
- Include current and incoming ordinal/hash in non-contiguous storage errors.
- Document the recovery trace and rollout checks in ADR-0031 and the v4.1 runbook.

## Compatibility

- No new snapshot, state-proof, consensus-message, or deterministic-configuration schema.
- No new configuration value or activation ordinal.
- No change to GlobalSnapshotStateProof calculation or metagraph snapshot validation.
- Snapshot membership/reward contents can change as intended because healthy Tier-1 peers are retained and readmitted.
- Deploy one jar across the full Global-L0 cluster; mixed behavior is not the rollout plan.

## Verification

- `nodeShared/test`: 1,056 passed, 0 failed
- `dagL0/test`: 133 passed, 0 failed, 2 ignored
- `currencyL0/test`: 37 passed, 0 failed
- `scalafmtCheckAll`: passed

The rollout runbook documents the new headroom, audit, and recovery metrics and the conditions that should stop deployment.

### DIFF
--- a/docs/adr/0030-broad-tier1-signing-leases-and-canonical-admission.md
+++ b/docs/adr/0030-broad-tier1-signing-leases-and-canonical-admission.md
@@ -83,15 +83,23 @@ mechanics, with its existing controller-target gate preserved.
     proof-miss streaks for every peer in `currentTier1 intersect
     parentRoundCommittee`, then select one audit target from that complete,
     consensus-agreed set by rendezvous rank. A current Core member emits the existing
-    `EvictionVote(Silent)` only after the target is absent from three consecutive
-    locally finalized parent artifact proof sets **and** those locally observed
-    current-committee signers cannot support the finality floor for one additional
-    seat. Reuse
-    `TierTransitions.DemotionConsecutiveMisses`; reset a peer's streak on any observed
-    proof. The eviction headroom check is the exact inverse of the open-admission
-    invariant in decision 14. Newly admitted peers are not auditable until they have
-    had a parent-round signing opportunity. Restart, missing parent evidence, or a
-    non-consecutive parent ordinal clears local streaks, delaying eviction conservatively.
+    `EvictionVote(Silent)` only when all of the following hold:
+    - the target is absent from three consecutive locally finalized parent artifact
+      proof sets, reusing `TierTransitions.DemotionConsecutiveMisses`;
+    - those observations span at least `(DemotionConsecutiveMisses - 1) *
+      timeTriggerInterval`, preserving the normal elapsed observation window when
+      EventTrigger accelerates rounds;
+    - locally observed current-committee signers are below the finality floor for the
+      **current** committee; and
+    - the round is on the existing `activeAdmissionExpansionIntervalRounds` cadence.
+
+    Any observed proof resets both the count and elapsed-time streak. Newly admitted
+    peers are not auditable until they have had a parent-round signing opportunity.
+    Restart, missing parent evidence, or a non-consecutive parent ordinal clears local
+    streaks, delaying eviction conservatively. The count, monotonic time, and cadence
+    are local vote-emission policy only; certified membership remains authoritative.
+    This cadence applies only to the proactive Tier-1 finality audit. Certified Core
+    stall eviction remains an every-round liveness recovery path.
 11. A Tier-1 target requires a Core-quorum certificate. A Core-target stall eviction
     retains the wider deterministic historical witness pool; the target's frozen tier
     selects between these two existing lanes at both assembly and validation.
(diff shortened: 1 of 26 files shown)


ryle-ai (commented):
Verified the RC4 Tier-1 stability changes against consensus fork-safety and the ADR-0031 recovery path.

probationPeers/persistenceValue sticky-zero semantics: probationPeers keys off map entry presence (keySet.toSet, no >0 filter) and persistenceValue collapses a sticky-zero countdown to the 1 sentinel while absent stays 0 (ReadmissionMaintenanceSuite pins both). The signed peerHistory.perPeer artifact stays empty, so membership rides the node-local operational record only and no signed bytes change.

FinalityHeadroom three-state model: allowsExpansion (nextSeatMargin>=0), allowsSilentEviction (currentMargin<0), and the neutral holdsMembership band are exactly disjoint and total, proven exhaustively for signer counts 1..100; the neutral floor step is what stops adjacent-floor membership oscillation. OpenAdmissionPolicy dedups onto the shared evaluate rather than its old inline floor calc.

FinalityParticipationAuditor: silent-eviction voting now requires both round hysteresis (consecutiveMisses>=required) and duration hysteresis (missStartedAt elapsed >= minimumMissDuration), gated further by headroom.allowsSilentEviction and cadence; minimumMissDuration(interval, required) = interval*max(0, required-1) (86s at required=3, Zero at required=1). Vote emission is node-local and never validated from a Proposal or copied into deterministic state.

StateTransitions torn-handoff fix: the pure downloadOutcomeDisposition classifier (AcceptExact/AcceptAndAlignApplicationStorage/Reject) makes the ADR-0031 alignment rule unit-testable; the align branch calls advancer.synchronizeDownloadedOutcome before installing an N+1 fallback outcome, and recoveredAtKeyRef records the actually-accepted key rather than the requested one. The Global advancer carries the LastNGlobalSnapshotStorage/LastSnapshotStorage/SnapshotStorage handles and hashing context to align for real; the Currency advancer has none of those and its override is inert, which matches the storage it owns. LastNGlobalSnapshotStorage non-contiguous updates stay fail-closed (raiseError) with only the diagnostic message enriched with current/incoming ordinal+hash.

AdmissionCertificateSelector.selectForProposal orders probation members first, then rendezvous target-order, then certificate ordering tie-break (deterministic across permutations).

ryle-goehausen (approved):

PR #1566, opened 2026-08-12
feat(consensus): certify frozen outcomes across views (v35)

scasplte2:
## Why

IntegrationNet demonstrated two separate problems:

- Global L0 could finalize the same artifact in different views while honest nodes persisted different committee/evidence/reward outcomes. Legacy artifact signatures certify only the artifact hash, not the complete consensus outcome.
- The large, partially synchronous Global committee needed substantial liveness and recovery hardening. Currency L0, however, is normally a small co-located permissioned committee and became less reliable when the Global machinery was copied into it.

This PR certifies the complete Global L0 outcome across views, composes the rc.8-rc.12 IntegrationNet stabilization line, and restores Currency L0 to a flat synchronous all-member protocol while retaining deterministic Currency-history upgrades behind their own ordinal gate.

## Current scope and reviewer map

Comparison: `develop@d63e67046` to `490011e1b` (55 commits, 334 files, `+43,826/-8,047`). Review this as functional slices rather than one monolithic change:

1. **Schema-compatible IntegrationNet stabilization, active immediately**
   - rc.8 sustained admission/headroom and retry/soft-reset hardening.
   - rc.9 env-selected recovery seed committee.
   - rc.10 event-intent batching and bounded follower catch-up.
   - rc.11 admission/facilitation alignment and automated snapshot-storage repair.
   - rc.12 aligned first-round startup after rollback.
   - Small-committee admission vote/certificate timing closure.

2. **Global certified consensus v35, activation-gated**
   - Domain-separated prepare and Core-commit votes over one typed `ProposalValue`.
   - Highest verified QC carry-forward across views and persisted per-key vote locks.
   - Frozen current-round full/Core authority; membership effects apply only to N+1.
   - Transferably signed trigger statements; followers validate leader-carried evidence rather than their local Facility subset.
   - No in-round denominator shrink after activation; loss of the required frozen Core/full quorum halts safely.

3. **Authenticated Global policy effects and public lineage, activation-gated**
   - `GlobalIncrementalSnapshot` gains one trailing optional child-carried parent certificate. `GlobalSnapshot` is unchanged.
   - Every QC authenticates the current full/Core authority, exact next full/Core authority, and a hash of the next operational state.
   - Live nodes independently run the one join-fenced policy and sign only the expected result. Historical replay consumes the certified result; it does not reinterpret old rounds using a joiner's current seedlist, allowance, collateral, quorum fraction, Core sizing, or selector.
[...]

PR #1568, opened 2026-08-19
fix: allow healthy validators to replace persistently non-participating nodes

Proph151Music:
## What I observed

While testing RC10, I found a case where the consensus committee had four validator seats, but only three validators were consistently completing rounds.

Three participants were still enough to finalize snapshots, so consensus did not stop. However, four participants were needed before the committee could safely add a fifth validator.

This left the network in an unusual middle state:

- snapshots continued to finalize
- the non-participating validator kept its committee seat
- its address continued to appear in snapshot rewards
- healthy validators waiting outside the committee could not be admitted and
- committee growth and rotation could not continue.

The network was still running, but it had no safe way to replace the validator that was no longer participating.

## What RC11 improved

RC11 added important improvements to validator admission and follower recovery.

It now performs stronger checks to confirm that a candidate is aligned with the current consensus round and snapshot before admission. It also improves how a newly admitted validator catches up and becomes an active participant.

Those changes address the most likely cause of the behavior I originally saw on RC10: a newly admitted validator failing to become a reliable participant. RC11 looked noticeably healthier in both the live network data and local testing.

The remaining issue is what happens if a validator is admitted correctly but later stops participating.

That could be caused by a crashed process, a network or P2P failure, a hung consensus component, a software malfunction, or another operational problem. RC11 made the original situation less likely, but it did not change the membership policy used after a validator becomes persistently non-participating.

I was able to reproduce that remaining condition using untouched official RC11 on an isolated seven-validator network.

## What this change does

This change adds a safe replacement path for a Tier-1 validator that repeatedly fails to provide the expected consensus participation.

It intentionally does not try to classify every possible node malfunction. Instead, it acts on something the other validators can objectively observe and verify: the validator has persistently failed to participate in consensus.

Before the validator can be removed:

1. It must miss the required number of consecutive rounds.
2. The existing minimum amount of time must pass.
3. The existing membership-change schedule must allow the action.
4. Core validators must independently sign votes identifying the same validator.
5. The existing eviction-certificate process must collect the required Core quorum.
6. The proposal must validate the validator identity, snapshot, committee, voters, quorum, and signatures.

Only after those checks succeed can the non-participating validator’s seat be removed.

Once that seat is removed, the remaining committee can admit a healthy waiting validator through the normal admission process.

## Why this complements RC11

RC11 improves the process of admitting a validator and helping it become an active participant.

This change covers the other side of the lifecycle: what the network can safely do if an admitted validator later stops participating and prevents another validator from joining.

Together, the two changes provide:

- stronger checks before admission
- better recovery for newly admitted validators
- a safe response when a Tier-1 validator remains non-participating
- an opportunity for a healthy waiting validator to join and
- a path for the removed validator to return after it has recovered.

A removed validator is not permanently banned. If it is repaired, catches up to the current snapshot, and passes the existing health and probation checks, it can safely rejoin later.

## Safety

This change does not lower consensus or finality requirements.

It also does not:

- allow one validator to remove another by itself
- remove a validator based only on one node’s local health information
- count waiting validators as consensus participants
- create, assume, or bypass signatures
- bypass the existing admission process or
- bypass eviction-certificate validation.

Removal requires signed agreement from the required Core-validator quorum.

## Test results

I compared the behavior on an isolated seven-validator network.

With untouched RC11:

- the committee remained at four seats with three completing validators from rounds 32 through 45
- snapshots continued to finalize
- no validator was removed or admitted
- the non-participating validator remained in the reward set (the bad validator continued to get rewards) and
- healthy waiting validators remained outside the committee.

With this change:

- all three Core validators independently produced signed eviction votes after the required miss and time limits
- the required two-validator Core quorum formed an eviction certificate
- snapshot 36 recorded the removal
- the removed validator left the reward set
- a healthy waiting validator was admitted at round 40 and
- after the failed validator was repaired and caught up, the existing probation process safely re-admitted it at round 44.

I also ran the focused test coverage locally. All 48 tests passed, along with compilation, assembly, Scala formatting, and source-diff validation.

This work is based on official RC11 commit.

### DIFF
--- a/modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotConsensus.scala
+++ b/modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotConsensus.scala
@@ -356,11 +356,10 @@ object GlobalSnapshotConsensus {
           mptStore,
           facilitatorSelector,
           peerHistorySidecar,
-          HealthDerivedMembershipPolicy.RetainSigningLeases,
+          HealthDerivedMembershipPolicy.CertifiedEvictionOnly,
           onUnsignedRecoverySuccessor,
           (key: GlobalSnapshotKey) =>
             consensusStorage.getRoundAttemptId.flatMap { expectedAttemptId =>
-              consensusQueue.offer(ConsensusCommand.RestartAfterSoftReset(key, expectedAttemptId))
             }
         )
 
@@ -389,7 +388,7 @@ object GlobalSnapshotConsensus {
           rawEvictionVoter,
           consensusQueue,
           tier1FinalityMissHistoryRef,
-          HealthDerivedMembershipPolicy.RetainSigningLeases
+          HealthDerivedMembershipPolicy.CertifiedEvictionOnly
         )
 
       stateRemover =
@@ -509,7 +508,7 @@ object GlobalSnapshotConsensus {
           effectiveConsensusConfig,
           facilitatorSelector,
           peerQualityTracker,
-          HealthDerivedMembershipPolicy.RetainSigningLeases,
+          HealthDerivedMembershipPolicy.CertifiedEvictionOnly,
           viewChangeVoter,
           timeoutVoter,
           rawEvictionVoter,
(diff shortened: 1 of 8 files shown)


scasplte2:
This PR is superseded by merged PR #1566 and follow-up PR #1589. PR #1566 introduced the current v35 certified atomic replacement design, and PR #1589 fixes the post-commit eviction-vote ordering needed for that design to operate. This branch targets the older release/integrationnet membership implementation and conflicts with the current rc.13 design, so it should not be merged.

PR #1569, opened 2026-08-19
FIX: WaitingForDownload Loop - Recover validators from damaged local snapshot files

Proph151Music:
This change helps a validator recover automatically when files in its local snapshot storage are missing or unreadable.

During testing, I found that a damaged snapshot or snapshot-info file could leave a validator repeating `WaitingForDownload` indefinitely. Healthy peers and valid snapshots were available, but the node kept selecting the same damaged local file and failed before the existing recovery logic could replace it.

Previous fixes improved how validators align with the current snapshot and consensus round when joining or catching up. Those changes fixed several recovery cases, including a valid snapshot-info file paired with the wrong snapshot. The remaining problem was lower in the local storage path: some decoder failures escaped as exceptions instead of being treated as unreadable data.

### What changes

The local storage reader now contains decoder failures and reports unreadable data as absent. This lets the existing validated download path take over instead of retrying the same broken file forever.

The download path also:

- checks that snapshot-info can actually be read instead of checking only that its filename exists
- moves below an unusable persisted snapshot instead of selecting the same ordinal repeatedly
- removes a stale ordinal link before installing a newly downloaded and validated snapshot and
- RC12 now provides first-round alignment for recovering committee members. This pull request therefore focuses only on repairing unreadable or missing local snapshot storage and relies on RC12 for the consensus-start handoff.

The last point is important because storage repair alone was not enough. In testing, a repaired active validator entered `Ready` but waited a full initial-join interval before participating. The healthy committee completed the round first, leaving the repaired validator one round behind. New validators still keep the normal initial-join delay because they are not members of the signed facilitator set.

### Through my testing

On the untouched RC11, an unreadable snapshot-info file caused 13 repeated download failures and remained unchanged until it was manually restored. Corrupting both hardlinks for a persisted snapshot caused the same persistent failure. With two affected validators at once, the other five validators preserved finality, but both damaged nodes remained locally stranded.

With this change, the following cases were tested on an isolated seven-validator network:

- unreadable snapshot-info
- unreadable snapshot bytes under both indexes
- valid snapshot-info paired with the wrong snapshot
- missing ordinal index with a valid content-addressed snapshot
- readable snapshot-info with both matching snapshot indexes missing
- a corrupt MPT checkpoint and
- two validators with unreadable snapshot-info at the same time.

Every unreadable or missing-file case recovered automatically. The two-node case repaired both validators on their first download attempt. After the active-seat handoff change, both validators participated immediately and the next measured round completed with all seven signers and no admission or eviction.

The final network capture showed all seven validators `Ready`, running the same JAR, and reporting the same ordinal and snapshot hash. The focused existing suites passed 45 of 45 tests. Shared, node-shared, and DAG L0 compilation passed both in the combined devnet build and with only these six files applied to a clean RC11 checkout. Assembly, formatting, and source-diff validation also passed.


scasplte2:
This PR is superseded by PR #1588, which ports the damaged-local-file recovery work onto current develop after PR #1566 and includes the decoder-containment, readable-anchor, and backward-search fixes. Please use PR #1588 rather than merging this older release/integrationnet branch.

PR #1572, opened 2026-08-24
fix: reject fee transactions that overflow balances instead of minting supply

marcus-girolneto:
## Summary

`BalanceOpsManager.acceptFeeTxs` unwrapped the balance map from `Balance` to raw `Long` and applied every fee transaction in a fold, checking the result only once at the end through `NonNegLong.from`. Both the debit and the credit could wrap silently, and the single end-of-fold guard rejected a negative result while accepting one that had wrapped back around to non-negative.

## The problem

A group of large fee transactions from an address without sufficient balance can drive the running total past `Long.MinValue` and back into positive territory. Because only the final per-address total is validated, every intermediate balance is invisible to the guard, the end state is a set of valid `NonNegLong` values, and destinations end up credited with supply that was never debited from anywhere.

Worth stressing for review: **a corrected end-state guard would not catch this.** The final state is entirely valid. The check has to happen per transaction.

## The fix

Fee transactions are applied one at a time through `Balance.minus` and `Balance.plus`, the domain's existing checked operations, which reject underflow and overflow. Every intermediate balance is now validated rather than only the final total.

Transactions the source cannot cover are **dropped, not raised on**. Fee transactions are user-supplied, so raising here would let anyone permanently halt a metagraph by submitting a single unaffordable transaction — the snapshot would fail and the transaction would be retried indefinitely.

The logic moved to `BalanceOpsManager.applyFeeTransactions` so it can be unit tested directly.

## Compatibility

Behaviour is **identical** for any batch of affordable fee transactions: same balances, same accepted set. Only the overflow path changes. Normal metagraph traffic is unaffected, so this can roll out to nodes progressively without a coordinated flag day.

## Tests

`BalanceOpsManagerFeeTxsSuite`, 7 tests:

- rejects a batch that overflows the source balance and creates no supply
- rejects a single fee transaction the source cannot afford
- applies an affordable fee transaction
- debits between transactions instead of validating against a stale balance
- rejects a transfer that would overflow the destination balance
- accepts the affordable prefix of a batch and rejects the rest
- property: total supply never changes, for any batch

I verified these fail against the pre-fix implementation: restoring the old logic makes 6 of the 7 fail, and all 7 pass on the new. Full `nodeShared/test` suite passes (1070 tests).

## Not covered here — follow-ups

1. **`FeeTransactionValidator` has no balance check.** It validates only the signature and `source != destination`. Acceptance is now safe on its own, but the validator should reject unaffordable transactions earlier. Left out because plumbing balances into it changes its signature and every call site — too much for a targeted patch.
2. **`Order[FeeTransaction]` is `Order.by(_.amount)`.** Source, destination and `dataUpdateRef` do not participate in ordering, so `SortedSet` iteration order is externally influenceable and two transactions with equal amount collapse to a value-level tie broken only by proofs. Not exploitable for minting after this fix, but it should be a total order over all fields. Changing it alters snapshot content, so it needs its own coordinated change.
3. **This patch is forward-looking only.** It prevents further occurrences; it does not reconcile balances that already exist.

### DIFF
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/currency/BalanceOpsManager.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/currency/BalanceOpsManager.scala
@@ -14,7 +14,6 @@ import io.constellationnetwork.schema.balance.Balance
 import io.constellationnetwork.schema.transaction.RewardTransaction
 import io.constellationnetwork.security.signature.Signed
 
-import eu.timepit.refined.types.numeric.NonNegLong
 import org.typelevel.log4cats.SelfAwareStructuredLogger
 import org.typelevel.log4cats.slf4j.Slf4jLogger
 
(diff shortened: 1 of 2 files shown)


ryle-goehausen (approved):

scasplte2:
Reviewed the port against the `release/mainnet` original. It is faithful: same per-transaction checked arithmetic through `Balance.minus`/`Balance.plus`, same drop-and-log rather than raise, same seven tests, correctly relocated for the `develop` layout. No objection to merging it as a port.

Three follow-ups apply to this line as well as to `release/mainnet`. They are not introduced here and none of them block this PR:

1. **Activation ordinal.** Currency snapshots are re-derived rather than trusted on rejoin and on the GL0 state-channel path, so a change in acceptance semantics needs a `fieldsAddedOrdinals` boundary the way `fixingAllowSpendAndTokenLockValidation` and `removingProcessedDelegatedStakeWithdrawals` have one. `develop` and `release/mainnet` will need to agree on the value.
2. **One more validator gap on the same path**, unrelated to arithmetic.
3. **`Order[FeeTransaction] = Order.by(_.amount)`** is more load-bearing after this change than before, since iteration order now decides which transactions are dropped.

Also worth a note in the description: "identical behaviour for any batch of affordable fee transactions" is accurate, but *affordable* means something stricter now than it did — previously an address could go negative mid-batch and be rescued by a later credit in the same batch.

Details, evidence and proposed patches are in the private review repo rather than here.

scasplte2:
Correction to two parts of my earlier follow-up; neither changes my approval of this port.

- "One more validator gap" was too broad. Develop already contains an ordinal-gated, explicit-byte FeeTransactionSignatureValidator. The private review tracks the remaining coverage question; this arithmetic PR neither introduces nor needs to solve it.
- "develop and release/mainnet need to agree on the value" was too categorical. The replay boundary for each release/network must match when that behavior actually changes. Reusing one value is justified only after verifying that the interval between deployments contains no fee-transaction input on which the rules differ.

The ordering concern remains a separate coordinated follow-up, not something I recommend adding to this focused port.

scasplte2 (approved):

PR #1575, opened 2026-08-25
fix(node-shared): authorize Pacaswap fee-transaction balance deductions

marcus-girolneto:
Companion GL0 change for `pacaswap-metagraph#175/#176`.

## What this does

- Adds `FeeTransactionBugDeduction` as the last `BalanceAdjustmentReason`, preserving historical `SortedSet[SharedArtifact]` ordering.
- Authorizes the exact 17-artifact PacaSwap remediation set at currency ordinal **731647**.
- Keeps all four historical PacaSwap adjustment blocks live (109991, 145000, 472325, 731647) instead of retaining only the last block per currency.
- Exact artifact authorization is opt-in for the 731647 remediation block only: the emitted set must equal the authorized set field-for-field (address, reason, reference set, increase and deduction), so a metagraph cannot satisfy an authorized deduction while attaching an unauthorized increase, repeating it with different metadata, or adding another artifact. The historical 109991/145000/472325 blocks keep legacy subset validation so already-signed history replays unchanged.
- Gates checked fee-transaction arithmetic by the previous currency snapshot's committed global sync-view ordinal. At and below global ordinal **6814499**, replay uses the historical wrapping path; later history uses checked arithmetic.

The replay-stable ordinal is deliberate. A node's current GL0 head is not deterministic when old currency history is recomputed.

## Verification

- The three historical AMM resources match the GL0 authorization blocks field-for-field:
  - 109991: 48/48
  - 145000: 14/14
  - 472325: 421/421
- The new block matches `balance-adjustments-4.json`: 17/17 entries.
- Nominal deduction: **189,413,903,467.12 PACA**.
- The legacy arithmetic regression reproduces the signed mint history: `0 -> -2^62 -> Long.MinValue -> +2^62 -> 0`, with four destinations receiving `2^62`.
- `.jvmopts` includes `-Xss8m` for the Shapeless macro stack requirement.

## Deploy order

Deploy this Tessellation release to every GL0 node before the metagraph emits the remediation artifacts. The AMM CI is configured to build its SDK from this PR's commit through `TESSELLATION_VERSION`.

The known delegated-staking `PeerId` race is unrelated to this change.


scasplte2 (commented):
Review blockers at current head `b44085c05`:

- **Separate live creation from replay.** `CurrencySnapshotAcceptanceManager.scala:355` currently keys checked fee arithmetic from the node's live global head, which breaks historical replay. Replacing it with the parent `globalSyncView` alone is also unsafe for a halted metagraph: PacaSwap's first new snapshot would run legacy rules. Live creation must always use strict envelope validation + checked arithmetic; `CurrencySnapshotValidator.scala:158` should use the expected artifact to retry the historical combinations (legacy/unchecked, legacy/checked, strict/checked). Thread the mode through both `DataApplicationSnapshotAcceptanceManager` and fee acceptance.
- **Current head does not compile.** `BalanceAdjustmentLoaderSuite.scala:199` treats the new `List[BalanceAdjustmentAtOrdinal]` as one entry, and positional `FieldsAddedOrdinals` constructors are missing the newly inserted parameter. Current CI reproduces both.
- **The recovery can be escaped before it fires.** Blocks/spends are applied before adjustments (`CurrencySnapshotAcceptanceManager.scala:599,627`), while deductions clamp insufficient balances to zero (`CurrencyBalanceAdjustments.scala:164`). Moving funds from a listed address before/in the remediation snapshot preserves them. Use the first restart ordinal and guarantee ingress isolation, or reject/freeze spends from recovery addresses at that ordinal.
- **Do not tighten old signed adjustment blocks retroactively.** Keep legacy validation for 109991/145000/472325 and require full `BalanceAdjustment` set equality only for the new recovery block.
- **Cross-repo release gates:** PacaSwap still pins Tessellation 3.5.20, and its pool-reserve loader returns old state on resource failure while logging success. The matching Paca PR must bump to the final release and make that path fail closed before restart.

Suggested order: fix live/replay mode selection, close the adjustment movement window, restore compilation, then run creator/replay regression tests and the assembled Paca artifact check.

ryle-goehausen (approved):

PR #1540, opened 2026-07-14
chore: unban node ids in mainnet seedlist

ffox77:
## Summary

Unbans two nodes by adding their node ids to `mainnet/mainnet-seedlist`:

- `2157b5b58bfe21afbf097528238b10714f23aa662a134539f5b08185beb8f12a62a5db7170bce05a22787bbfb7735cb156d414f3f779940e259f0bb5b1ac0c3b`
- `4aa71b7ed020d9e57f6f2722ca66b692b72234efa32d7df247e6e85f2e3e8cc01124b99beee3c2a3b1fe1a942fe4c8a77cf26856c1f7a7dc4fecfe4853debdf7`

## Test plan

- [ ] Confirm both node ids are correct and intended for unban

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1579, opened 2026-08-26
chore: unban mainnet node id e7d04c88

ffox77:
## Summary

Restores one node id to `mainnet/mainnet-seedlist`:

- `e7d04c888897fc42bff0b96f1c18f02861a1e265d34aad0710c89253c3a552764e8c9982508cf61b5b01bf217c65bbb61402b40ddf0622a9c672996f3cfe109a`

It was one of six peers banned in #1560 for being stuck in `DownloadInProgress` / `WaitingForDownload` and never reaching `Ready`.

Community Admin worked with the Operator to update their node.

## Changes

Appended the id as the last line. File integrity after the edit:

- 314 lines, every line a distinct 128-char lowercase hex id, no duplicates, no blank lines
- Original no-trailing-newline style preserved

The diff shows `-1 / +2` only because the previous final line gained a newline; that line is otherwise unchanged. Same artifact as prior seedlist PRs (#1540).

## Test plan

- [ ] Confirm the node id is correct and intended for unban

No code paths change; this is a data file consumed by `validateSeedlist` at join time.

## Follow-up

`release/mainnet` has its own copy of this file and needs the same line for the unban to take effect on mainnet.


ryle-ai (commented):
Single-line addition to `mainnet/mainnet-seedlist`. I confirmed the new entry `e7d04c88...cfe109a` is a well-formed seedlist row: 128 lowercase hex chars matching the `PeerId`/`Hex` shape used by every other entry, and it parses as the 1-field format accepted by `SeedlistEntry.SeedlistRowDecoder` / `Loader`. It is not a duplicate of any existing entry, and it is appended correctly (the prior last line now gains a trailing newline, and the file still ends without one, consistent with the existing format). Nothing in code, tests, or other config references this file by line count or position, so the append is safe.

The one thing not verifiable from the repository is the operational correctness of the change itself — that this specific node ID is the one intended to be re-admitted to mainnet. Since the seedlist acts as an admission allowlist, adding an ID grants that node the ability to join; that decision should be confirmed by whoever owns the unban action against the actual node identity, as it is not checkable from code.

ryle-goehausen (approved):

PR #1583, opened 2026-08-28
Fix/currency l0 certified recovery

Proph151Music:
## Why I am submitting this

I found this while testing a metagraph on IntegrationNet with Tessellation
`v4.1.0-rc.12`. The metagraph used five Currency L0 validators. The validators
could appear to be running, but the Currency L0 group did not recover and
participate as a stable five-validator network.

The clearest symptom was a restarted Currency L0 validator downloading a valid
snapshot and then remaining in `Observing` instead of returning to `Ready`.
The five-seat signing behavior was also not preserved when one validator was
unavailable: the effective signing set could fall to four and three proofs
could finalize even though the network had already established a five-seat
committee.

I initially needed to determine whether this came from the metagraph
application, its configuration, or Tessellation. I reproduced the same failure
paths in an isolated topology built from the exact rc.12 IntegrationNet source,
then traced them to Tessellation's generic Currency L0 recovery and consensus
code. The reproduction did not depend on the metagraph's application logic,
and this pull request changes only Tessellation's `currency-l0` and
`node-shared` modules and their tests. It does not contain or disclose the
metagraph application.

The public IntegrationNet behavior was the reason for investigating. The proof
for this change comes from the isolated reproduction and the source paths
described below. No public validator or network state was modified during
testing.

## Impact on other metagraph builders

This is not specific to the metagraph that exposed it. These paths are part of
the rc.12 Tessellation Currency L0 framework, so other metagraphs built on the
same release use them as well.

A single-node or uninterrupted happy-path demonstration may not trigger every
failure. The defects become visible under normal multi-validator conditions:
a validator restarts while peers continue advancing, the committee contains a
Tier-1 member, one validator becomes unavailable, or Currency L0 submits a
state-channel binary to Global L0. Under those conditions, a builder can have
apparently healthy processes but still lack dependable validator recovery,
the intended four-of-five finality guarantee, or proof that the committee
certified one state-channel binary.

These behaviors cannot be safely corrected in application-specific metagraph
logic because snapshot recovery, signing membership, quorum calculation, and
binary certification are performed by Tessellation itself. Avoiding restarts
or outages can hide the defects during a demonstration, but it is not an
operational fix for a multi-validator metagraph.

### A restarted validator could receive state for the wrong snapshot

The rc.12 recovery path downloaded a Currency snapshot and then requested the
peer's latest calculated state. A healthy peer could already be several rounds
ahead, so the state returned by that endpoint did not belong to the downloaded
snapshot. Recovery then failed after observation and could leave the validator
in `Observing`.

For example, a validator could download snapshot 105 and then receive the
peer's latest calculated state for snapshot 108. The fix requests state for
snapshot 105 specifically and verifies it against snapshot 105 before changing
the validator's local state.

### A five-validator network could incorrectly finalize with three signatures

Currency L0 treated the active Core controller result as the complete signing
set. A selected validator classified as Tier 1 could therefore disappear from
the signing seats. During an outage, the rolling proof-size calculation could
then forget that five seats had already been established and lower the
requirement from four signatures to three.

The expected rule is simple: five established validators require four
signatures. If one validator is offline, all four remaining validators must
sign. The old path could remember only the four visible validators and then
accept three signatures.

### Declarations did not prove that four validators signed the same binary

The state-channel path counted declarations without first proving that enough
distinct members of the frozen committee had signed the same binary and the
same proposal context.

The fix requires four valid signatures from four different members of the
committee agreed at the start of the round. All four signatures must cover the
same binary and consensus context. Duplicates, foreign signers, and signatures
for a different binary do not count.

These are Tessellation framework paths used by Currency L0 metagraphs; they are
not application-specific validation or state-transition rules.

## Changes

- Request calculated state for the exact ordinal carried by the downloaded
  Currency snapshot.
- Verify that state against the snapshot's state proof before changing any
  local snapshot head.
- Persist calculated state first, advance the snapshot head second, and clear
  the mempool last.
- Return failed observation or persistence to `WaitingForDownload`, allowing
  normal recovery to retry instead of leaving the validator stuck.
- Advertise the newest consistent local chain tip instead of always preferring
  a possibly stale combined checkpoint.
- Keep a selected Tier-1 validator in the signing committee without adding it
  to the Core leader pool.
- Retain the established five-seat finality floor during a one-validator
  outage, so four signatures are still required.
- Bind each state-channel signature to one binary, proposal, view, facilitator
  set, and previous snapshot. Count only verified signatures from distinct
  members of the frozen committee.

The patch does not lower quorum, fabricate proofs, count foreign signers, trust
a moving latest state, or accept uncertified calculated state.

## How I verified the cause and fix

I kept the unsuccessful intermediate campaigns instead of treating the first
apparently healthy run as proof:

- Campaign 25 admitted five validators but exposed incorrect three-proof
  finality after one validator was stopped.
- Campaign 26 introduced an incompatible wire field that stock Global L0
  rejected. That approach was removed.
- Campaign 27 retained the finality floor but exposed the missing Tier-1
  signing seat.
- Campaign 29 passed admission, state-channel anchoring, and outage finality,
  but a restarted validator remained in `Observing` because it received
  calculated state for the wrong ordinal.
- Campaigns 30 and 31 proved recovery and resumed signing while also exposing
  and correcting test assumptions that incorrectly required all five proofs
  instead of the configured four-of-five supermajority.
- Campaign 32 passed the complete corrected gate.

Campaign 32 ran one isolated Global L0, five Currency/Metagraph L0, three Data
L1, and one Currency L1 nodes. It demonstrated that:

- all five Currency L0 validators reported the same five-seat committee across
  advancing rounds;
- the signing committee contained four Core validators and one Tier-1 member;
- ordinal 76 finalized with the four online validator proofs while the stopped
  validator was absent;
- the stopped validator restarted from persisted data and returned to `Ready`;
- the same validator returned to `Ready` after all 447 entries in its local
  data directory were removed while its throwaway identity was retained;
- consensus continued after both recovery paths;
- finalized ordinal 126 contained the recovered validator's valid proof,
  confirming that it resumed signing; and
- a four-signature state-channel binary was observed and signature-verified in
  the isolated Global L0.

The same source tree passed:

```text
shared/test:       303 passed
nodeShared/test:  1249 passed
currencyL0/test:    67 passed
scalafmtCheckAll:   passed
git diff --check:   passed
```

## Base, compatibility, and limits

The patch is based on commit
`92e337c893e05a84a96537aa0f101e5aec3a5977`, which was both
`release/integrationnet` and the peeled `v4.1.0-rc.12` tag when this branch was
prepared. The tested final source tree is
`ef1afc8f3af8e5976a3de6b33431d22bfd60f523`.

The Currency L0 binary-signature declaration and exact-ordinal recovery route
need to be compatible across the Currency L0 validators, so they should be
upgraded together. The state-channel binary sent to Global L0 is unchanged,
and no new Currency snapshot field is added.

This evidence shows that the patch fixes the reproduced rc.12 failures in the
isolated five-validator topology. It does not replace upstream review, CI, or a
coordinated IntegrationNet deployment test. No public-network seedlist,
validator identity, transaction, or state mutation was used during validation.


scasplte2:
@Proph151Music rc.12 is superseded by the pending #1566 as there are known issues. Thank you for the contribution. I will review if there is anything useful to be integrated into the pending rc.13.

Feel free to review #1566 as it is quite large so the more eyes/agents the better IMO

Proph151Music:
> @Proph151Music rc.12 is superseded by the pending #1566 as there are known issues. Thank you for the contribution. I will review if there is anything useful to be integrated into the pending rc.13.
> 
> Feel free to review #1566 as it is quite large so the more eyes/agents the better IMO

I understand.  I will do some tests with rc.13 soon.

scasplte2:
This PR is superseded by PR #1587, which carries the current develop-compatible Currency L0 downloaded-validator recovery work after PR #1566. This branch targets the older certified Currency design and conflicts with the intentionally restored synchronous Currency consensus model, so it should not be merged.

PR #1587, opened 2026-09-01
fix(currency-l0): recover downloaded validators safely

Proph151Music:
PR 1566 was merged into `develop` and its source branch was deleted. GitHub automatically closed the earlier PR 1584 because it targeted that deleted branch. This replacement contains the same validated recovery changes on current `develop`.

The merged PR 1566 commit `ee1c7e08173bd63cd7c2f9e7b66699ef734374e1` has the same source tree as the previous PR 1566 head `490011e1b648396378b275f08ae3f4deafc6818f`. No behavioral adaptation was needed when the recovery changes were moved to `develop`. The branch was later updated in response to review to replace the four-entry full Global-context cache with durable fee-context receipts and to make calculated-state recovery safely retryable after a crash.

## Why this is needed

I found this while testing a five-validator Currency L0 metagraph. The original failure was reproduced with the unmodified Tessellation v4.1.0-rc.12 code in an isolated devnet.

The validators could appear healthy and reach `Ready`, but that did not mean they had joined one stable Currency L0 group. In the concurrent control, all five validators exchanged peer declarations and observed the same Currency snapshot at ordinal 6. One validator was then admitted with the seed, while the other three entered `WaitingForDownload`. They downloaded the certified snapshot at ordinal 10 and repeatedly failed with:

`Failed to prepend currency snapshot ordinal=SnapshotOrdinal(10) to storage`

Starting the validators one at a time did not avoid the problem. In that control, the third validator reached `Ready`, but its local Currency snapshot remained at ordinal 10 while the active pair advanced beyond ordinal 100.

No Data L1, Currency L1, or metagraph transaction had started in either control. This ruled out the metagraph application, its data update, and the historical three-Data-L1 startup procedure. The failure occurred first in Tessellation's Currency L0 admission and recovery code.

## What remained after PR 1566

PR 1566 substantially changed Currency L0 consensus and replaced several rc.12 committee behaviors. I did not copy the old rc.12 patch onto it. I first audited the new design and ran 170 focused stock tests.

Those tests passed, but they did not cover the recovery failures below:

1. A validator can have a downloaded Currency snapshot newer than its combined checkpoint. Advertising only the checkpoint makes the validator look behind even when its snapshot storage is current.
2. Recovery needs calculated state for the exact downloaded snapshot ordinal. The existing route exposes moving latest state, which may belong to a newer snapshot by the time the recovering validator asks for it.
3. Equal local ordinals with different hashes must be treated as a conflict. Picking one without proving which is correct can advertise unsafe state.
4. Currency L0 registered for consensus events and then cleared its old event mempool. Events accepted immediately after registration could therefore be acknowledged and then deleted before the validator processed them.
5. A temporary gap in canonical state-channel data could be treated as a permanent conflict instead of allowing recovery to wait and retry.
6. Join retries could reuse stale peer-discovery results from an earlier attempt.

These are Tessellation framework paths used by Currency L0 metagraphs. A metagraph application cannot safely replace snapshot recovery, calculated-state verification, event delivery, or consensus membership handling.

## What this changes

This pull request:

* compares the downloaded snapshot and combined checkpoint and advertises the newer trustworthy local tip;
* advertises neither tip when the ordinals match but the hashes conflict;
* adds an exact-ordinal calculated-state route;
* requests calculated state for the downloaded snapshot's exact ordinal and verifies its hash against the proof in that snapshot;
* writes recovered calculated state atomically and verifies the written file before continuing;
* safely retries when application state was installed before a crash but canonical Currency snapshot installation was not completed;
* fails recovery when the exact state is unavailable, its hash does not match, the application has conflicting state at the same ordinal, or the application is already ahead of the recovery target;
* stores a small durable fee-context receipt for every locally created or successfully validated Currency proposal;
* requires the receipt for the exact selected artifact before binary signing;
* retains proposal receipts according to the consensus and publication lifecycle instead of using a fixed-size cache of complete Global snapshots;
* resets stale consensus state before handling a non-contiguous downloaded chain and returns a failed attempt to `WaitingForDownload` so it can retry;
* clears the stale event mempool before publishing consensus registration, so events accepted after registration are preserved;
* waits and retries when the canonical state-channel bridge has not arrived yet;
* removes stale finished consensus generations; and
* clears peer-discovery state after every join attempt and contacts the initial peers concurrently.

It does not lower quorum, fabricate a signature, accept calculated state from the wrong ordinal, or bypass snapshot proof verification.

### Stock failure controls

* A sequential rc.12 control admitted the first two validators, then left the third at Currency ordinal 10 while the active pair continued advancing.
* A concurrent rc.12 control started all four joiners together. All five exchanged declarations and observed the same initial head, but three validators entered `WaitingForDownload` and repeatedly failed to install the certified ordinal-10 snapshot.
* Both controls failed before Data L1, Currency L1, or application transaction processing began.

This ruled out launch order as a workaround and isolated the prerequisite failure to Currency L0 recovery.

### Isolated patched-candidate campaigns

Two complete isolated campaigns used different throwaway metagraph identities. Each campaign ran:

* one Global L0;
* five Currency L0 validators;
* three Data L1 validators; and
* one inert Currency L1.

In both campaigns:

* all five Currency L0 validators joined one common state;
* two byte-identical application updates were accepted;
* the Global L0 anchor was verified;
* the group finalized with one validator stopped;
* the stopped validator recovered its persisted state and returned to `Ready`;
* a validator whose local data had been emptied downloaded verified state and returned to `Ready`; and
* consensus resumed after both recovery cases.

The persisted-restart failure in this testing led to the event-cleanup ordering fix. Moving stale cleanup before consensus registration preserved newly accepted events and allowed the validator to rejoin stable participation.

### Verification on current develop

The replacement branch is based directly on merged PR 1566 commit `ee1c7e08173bd63cd7c2f9e7b66699ef734374e1`.

After the reviewer-requested fee-context receipt and calculated-state durability changes, the current branch passed:

* all 153 Currency L0 tests;
* all 313 shared tests;
* all 6 affected tests in `PeerDiscoverySuite` and `CachedCombinedResponseSuite`;
* `scalafmtCheckAll`; and
* `git diff --check`.

The added tests cover durable receipt creation, restart recovery, more than four active Global views, missing or mismatched receipts before signing, lifecycle cleanup, atomic calculated-state replacement, interrupted recovery, idempotent retry, same-ordinal hash conflicts, and application state ahead of the recovery target.

The two full topology campaigns were run against the earlier reviewed PR 1566 candidate, not this replacement commit. I am not presenting the focused current-`develop` run as a third topology campaign.

## Impact on other metagraph builders

The failure happens before application-specific transaction handling. A simple uninterrupted demonstration may never trigger every recovery path, but a validator that joins late, restarts, loses local data, or observes a moving peer can encounter it. The fix belongs in Tessellation so each metagraph does not have to discover the same framework limitation independently.

## Scope and safety

The base is current `develop` after PR 1566 was merged. All runtime testing used isolated networks and throwaway identities. No public network was joined or modified, and no production validator key was used.

### DIFF
--- a/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
+++ b/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
@@ -1,15 +1,20 @@
 package io.constellationnetwork.currency.l0.snapshot
 
-import cats.data.Validated.Invalid
+import cats.data.Validated.{Invalid, Valid}
 import cats.effect.Async
 import cats.syntax.all._
 
-import scala.collection.immutable.{SortedMap, SortedSet}
+import scala.collection.immutable.SortedSet
+import scala.util.control.NoStackTrace
 
+import io.constellationnetwork.currency.l0.snapshot.storage.CurrencyFeeContextReceiptStorage
+import io.constellationnetwork.currency.l0.snapshot.storage.CurrencyFeeContextReceiptStorage.CurrencyFeeContextReceipt
 import io.constellationnetwork.currency.schema.currency._
+import io.constellationnetwork.currency.schema.globalSnapshotSync.GlobalSyncView
+import io.constellationnetwork.ext.crypto._
 import io.constellationnetwork.node.shared.domain.consensus.ConsensusFunctions
 import io.constellationnetwork.node.shared.domain.rewards.Rewards
-import io.constellationnetwork.node.shared.domain.snapshot.services.GlobalL0Service
+import io.constellationnetwork.node.shared.domain.snapshot.storage.LastSyncGlobalSnapshotStorage
 import io.constellationnetwork.node.shared.infrastructure.consensus.trigger.ConsensusTrigger
 import io.constellationnetwork.node.shared.infrastructure.snapshot._
 import io.constellationnetwork.node.shared.snapshot.currency._
@@ -18,7 +23,9 @@ import io.constellationnetwork.schema.address.Address
 import io.constellationnetwork.schema.artifact.SharedArtifact
 import io.constellationnetwork.schema.balance.{Amount, Balance}
 import io.constellationnetwork.schema.consensus.CertifiedLineageEvidenceV1
+import io.constellationnetwork.schema.currencyMessage.fetchStakingAddress
 import io.constellationnetwork.schema.peer.PeerId
+import io.constellationnetwork.security.hash.Hash
 import io.constellationnetwork.security.signature.Signed
 import io.constellationnetwork.security.{Hashed, Hasher, SecurityProvider}
 
@@ -48,6 +55,69 @@ abstract class CurrencySnapshotConsensusFunctions[F[_]: Async: SecurityProvider]
 
 object CurrencySnapshotConsensusFunctions {
 
+  final case class MissingGlobalSyncViewForFeeContext(currencyOrdinal: SnapshotOrdinal) extends ConsensusFunctions.InvalidArtifact {
+    override def getMessage: String =
+      s"Currency proposal at ordinal=${currencyOrdinal.show} does not contain the Global L0 view required for fee calculation"
+  }
+
+  final case class ExactGlobalFeeContextUnavailable(ordinal: SnapshotOrdinal) extends ConsensusFunctions.InvalidArtifact {
+    override def getMessage: String = s"Exact Global L0 fee context is unavailable at ordinal=${ordinal.show}"
+  }
+
+  final case class ExactGlobalFeeContextHashMismatch(ordinal: SnapshotOrdinal, expected: Hash, actual: Hash) extends NoStackTrace {
+    override def getMessage: String =
+      s"Exact Global L0 fee context hash mismatch at ordinal=${ordinal.show}: expected=${expected.show} actual=${actual.show}"
+  }
+
+  /** Capture the small fee input while the proposal's exact Global context is locally available.
+    *
+    * The durable receipt is keyed by Currency ordinal and artifact hash. It does not retain a full Global snapshot in heap and it cannot be
+    * displaced by unrelated proposal accesses.
+    */
+  private[snapshot] def captureFeeContextReceipt[F[_]: Async](
+    currencyOrdinal: SnapshotOrdinal,
+    currencyArtifactHash: Hash,
+    globalSyncView: Option[GlobalSyncView],
+    stakingAddress: Option[Address],
+    resolve: SnapshotOrdinal => F[Option[(Hash, Balance)]],
+    putDurably: CurrencyFeeContextReceipt => F[Unit]
+  ): F[Unit] =
+    globalSyncView
+      .liftTo[F](MissingGlobalSyncViewForFeeContext(currencyOrdinal))
+      .flatMap { view =>
+        resolve(view.ordinal)
+          .flatMap(_.liftTo[F](ExactGlobalFeeContextUnavailable(view.ordinal)))
+          .flatMap {
+            case (actual, stakingBalance) =>
+              ExactGlobalFeeContextHashMismatch(view.ordinal, view.hash, actual)
+                .raiseError[F, Unit]
+                .whenA(actual =!= view.hash) >>
+                putDurably(
+                  CurrencyFeeContextReceipt(
+                    CurrencyFeeContextReceiptStorage.CurrentEncodingVersion,
+                    currencyOrdinal,
+                    currencyArtifactHash,
+                    view,
+                    stakingAddress,
+                    stakingBalance
+                  )
+                )
+          }
+      }
+
+  /** A temporarily unavailable exact Global context invalidates only the proposal being checked. A same-ordinal hash conflict remains a
+    * hard failure because it can indicate a fork or corrupted local authority.
+    */
+  private[snapshot] def validateFeeContextCapture[F[_]: Async](capture: F[Unit]): F[Either[ConsensusFunctions.InvalidArtifact, Unit]] =
+    capture.attempt.flatMap {
+      case Right(_) => ().asRight[ConsensusFunctions.InvalidArtifact].pure[F]
+      case Left(error: MissingGlobalSyncViewForFeeContext) =>
+        (error: ConsensusFunctions.InvalidArtifact).asLeft[Unit].pure[F]
+      case Left(error: ExactGlobalFeeContextUnavailable) =>
+        (error: ConsensusFunctions.InvalidArtifact).asLeft[Unit].pure[F]
+      case Left(error) => error.raiseError[F, Either[ConsensusFunctions.InvalidArtifact, Unit]]
+    }
+
   final case class ProposalArtifactResult(
     artifact: CurrencySnapshotArtifact,
     context: CurrencySnapshotContext,
@@ -60,9 +130,37 @@ object CurrencySnapshotConsensusFunctions {
     rewards: Option[Rewards[F, CurrencySnapshotStateProof, CurrencyIncrementalSnapshot, CurrencySnapshotEvent]],
     currencySnapshotCreator: CurrencySnapshotCreator[F],
     currencySnapshotValidator: CurrencySnapshotValidator[F],
-    maybeCustomArtifacts: Option[Signed[CurrencyIncrementalSnapshot] => Option[SortedSet[SharedArtifact]]]
+    maybeCustomArtifacts: Option[Signed[CurrencyIncrementalSnapshot] => Option[SortedSet[SharedArtifact]]],
+    lastGlobalSnapshotStorage: LastSyncGlobalSnapshotStorage[F],
+    feeContextReceiptStorage: CurrencyFeeContextReceiptStorage[F]
   ): CurrencySnapshotConsensusFunctions[F] = new CurrencySnapshotConsensusFunctions[F] {
     val logger: SelfAwareStructuredLogger[F] = Slf4jLogger.getLoggerFromName[F]("CurrencySnapshotConsensusFunctions")
+
+    private def captureFeeContextReceipt(
+      artifact: CurrencySnapshotArtifact,
+      lastContext: CurrencySnapshotContext
+    )(implicit hasher: Hasher[F]): F[Unit] =
+      for {
+        artifactHash <- artifact.hash
+        stakingAddress = fetchStakingAddress(lastContext.snapshotInfo)
+        _ <- CurrencySnapshotConsensusFunctions.captureFeeContextReceipt(
+          artifact.ordinal,
+          artifactHash,
+          artifact.globalSyncView,
+          stakingAddress,
+          ordinal =>
+            lastGlobalSnapshotStorage
+              .getCombined(ordinal)
+              .flatMap(
+                _.traverse {
+                  case (snapshot, info) =>
+                    snapshot.hash.map(_ -> stakingAddress.flatMap(info.balances.get).getOrElse(Balance.empty))
+                }
+              ),
+          receipt => feeContextReceiptStorage.putDurably(receipt).void
+        )
+      } yield ()
+
     override def triggerPredicate(event: CurrencySnapshotEvent): Boolean = event match {
       case GlobalSnapshotSyncEvent(_) => false // NOTE: Sync events should not trigger consensus to avoid infinite loop
       case _                          => true
@@ -96,12 +194,16 @@ object CurrencySnapshotConsensusFunctions {
           peerHistory,
           historicalDependencyResolution = false
         )
-        .flatTap {
+        .flatMap {
           case Invalid(errors) =>
-            logger.warn(s"Failed when validating currency artifact. Errors: ${errors.toList}")
-          case _ => Async[F].unit
+            logger
+              .warn(s"Failed when validating currency artifact. Errors: ${errors.toList}")
+              .as(CurrencyArtifactMismatch(errors.toList).asLeft)
+          case Valid(validated) =>
+            CurrencySnapshotConsensusFunctions
+              .validateFeeContextCapture(captureFeeContextReceipt(artifact, lastContext))
+              .map(_.map(_ => validated))
         }
-        .map(_.leftMap(errors => CurrencyArtifactMismatch(errors.toList)).toEither)
 
     def createProposalArtifact(
       lastKey: SnapshotOrdinal,
@@ -161,6 +263,7 @@ object CurrencySnapshotConsensusFunctions {
           peerHistory,
           historicalDependencyResolution = false
         )
+        .flatTap(created => captureFeeContextReceipt(created.artifact, lastContext))
         .map(created => ProposalArtifactResult(created.artifact, created.context, created.awaitingEvents, created.rejectedEvents))
     }
   }
--- a/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensusStateAdvancer.scala
+++ b/modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensusStateAdvancer.scala
@@ -13,6 +13,12 @@ import scala.concurrent.duration.FiniteDuration
 import io.constellationnetwork.currency.dataApplication.BaseDataApplicationL0Service
 import io.constellationnetwork.currency.l0.snapshot.schema._
 import io.constellationnetwork.currency.l0.snapshot.services.StateChannelSnapshotService
+import io.constellationnetwork.currency.l0.snapshot.storage.CurrencyFeeContextReceiptStorage
+import io.constellationnetwork.currency.l0.snapshot.storage.CurrencyFeeContextReceiptStorage.{
+  CurrencyFeeContextKey,
+  CurrencyFeeContextReceipt,
+  MissingCurrencyFeeContextReceipt
+}
 import io.constellationnetwork.currency.l0.snapshot.synchronous.ConsensusStateUpdater._
 import io.constellationnetwork.currency.l0.snapshot.synchronous._
 import io.constellationnetwork.currency.l0.snapshot.synchronous.declaration._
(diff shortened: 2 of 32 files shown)


scasplte2:
I think the exact Currency recovery work addresses a real framework problem and should remain, but I recommend replacing the new four-entry full Global-context LRU before merging it.

## Recommendation

Keep the exact Currency head and exact-ordinal calculated-state recovery. A Currency snapshot commits the calculated-state hash but not the state bytes, so a recovering validator genuinely needs the state for that exact ordinal. The new route is peer-authenticated and does not add a method to the metagraph application SDK.

However, `retainedExactCapacity = 4` retains complete `GlobalIncrementalSnapshot` / `GlobalSnapshotInfo` values in heap and is not true lifecycle pinning: an active proposal can still be evicted by four newer accesses. The value `4` is heuristic and is not derived from the number of active proposals, generations, or facilitators.

The final synchronous-binary path uses that large historical context for only:

1. Verifying that the Global snapshot hash equals the signed `globalSyncView.hash`.
2. Reading the staking balance for one address.

That suggests a much smaller and more deterministic design: persist a local fee-context receipt when each local or remote Currency proposal is created and validated, while the normal Global context is still available.

This can remain local implementation state and requires no Currency/Global wire-schema, metagraph SDK, Snapshot Streaming, or Block Explorer change.

### Suggested model

```scala
final case class CurrencyFeeContextReceipt(
  currencyOrdinal: SnapshotOrdinal,
  currencyArtifactHash: Hash,
  globalSyncView: GlobalSyncView,
  stakingAddress: Option[Address],
  stakingBalance: Balance
)

final case class CurrencyFeeContextKey(
  currencyOrdinal: SnapshotOrdinal,
  currencyArtifactHash: Hash
)
```

The artifact hash is important because several facilitators may propose different artifacts, and possibly different Global views, at the same Currency ordinal.

### Capture during proposal creation/validation

The exact types and placement can follow existing module boundaries, but the required ordering is approximately:

```scala
def captureFeeContextReceipt(
  artifact: CurrencySnapshotArtifact,
  lastContext: CurrencySnapshotContext
)(implicit hasher: Hasher[F]): F[Unit] =
  artifact.globalSyncView.traverse_ { view =>
    for {
      combined <- lastGlobalSnapshotStorage
        .getCombined(view.ordinal)
        .flatMap(_.liftTo[F](ExactGlobalFeeContextUnavailable(view.ordinal)))

      (globalSnapshot, globalInfo) = combined
      actualGlobalHash <- globalSnapshot.hash

      _ <- Async[F].raiseUnless(actualGlobalHash === view.hash)(
        ExactGlobalFeeContextHashMismatch(
          view.ordinal,
          expected = view.hash,
          actual = actualGlobalHash
        )
      )

      artifactHash <- artifact.hash
      stakingAddress = fetchStakingAddress(lastContext.snapshotInfo)
      stakingBalance =
        stakingAddress
          .flatMap(globalInfo.balances.get)
          .getOrElse(Balance.empty)

      receipt = CurrencyFeeContextReceipt(
        artifact.ordinal,
        artifactHash,
        view,
        stakingAddress,
        stakingBalance
      )

      // Atomic replacement plus read-back verification.
      _ <- feeContextReceiptStorage.putDurably(receipt)
    } yield ()
  }
```

Capture must occur for both locally created proposals and successfully validated remote proposals. A proposal should not become locally eligible for selection until its receipt is durable.

Receipts must be derived only from a locally validated Global context. They should not be accepted from peers as independent authority because the individual balance does not currently carry a separate Merkle proof.

### Consume during binary construction

```scala
def loadFeeInput(
  signedArtifact: Signed[CurrencySnapshotArtifact],
  expectedStakingAddress: Option[Address]
)(implicit hasher: Hasher[F]): F[Balance] =
  for {
    artifactHash <- signedArtifact.value.hash
    key = CurrencyFeeContextKey(signedArtifact.ordinal, artifactHash)

    receipt <- feeContextReceiptStorage
      .get(key)
      .flatMap(_.liftTo[F](MissingCurrencyFeeContextReceipt(key)))

    view <- signedArtifact.value.globalSyncView.liftTo[F](
      MissingGlobalSyncView(signedArtifact.ordinal)
    )

    _ <- Async[F].raiseUnless(
      receipt.globalSyncView === view &&
        receipt.stakingAddress === expectedStakingAddress
    )(CurrencyFeeContextReceiptMismatch(key))
  } yield receipt.stakingBalance
```

`createSynchronousBinaryValue` can then calculate the fee from this balance without loading or retaining the full historical Global snapshot/context.

### Lifecycle and cleanup

Receipts should follow consensus lifecycle, not an arbitrary LRU capacity:

- Retain every receipt belonging to the active generation.
- After majority selection, rejected proposal receipts may be released.
- Retain the selected receipt until the exact Currency artifact and binary publication/outbox state are durable.
- On abandonment, release the generation receipts.
- On restart, preserve receipts for unfinished/durable publication work and sweep only receipts proven older than the canonical terminal state with no pending outbox/recovery reference.

The receipt is small enough that retaining one per active facilitator proposal is inexpensive and does not impose a fixed committee-size limit.

A simpler fallback would be to pin the full context in the existing filesystem store and exempt protected ordinals from cleanup. That avoids heap growth but retains large objects and introduces pin/cleanup crash ordering. The small receipt is preferable.

## Calculated-state crash window

The exact-ordinal calculated-state endpoint should remain, but recovery currently updates:

1. The framework calculated-state file.
2. Application-owned state via `setCalculatedState`.
3. Canonical Currency snapshot/context storage.

These are independent durability boundaries. A crash after application state reaches `T` but before the canonical snapshot reaches `T` must be retryable without requiring a new SDK method.

A minimal approach is:

```scala
for {
  _ <- calculatedStateStorage.writeAtomically(targetOrdinal, recoveredState)
  _ <- calculatedStateStorage.verifyHash(targetOrdinal, expectedProof)

  current <- dataApplication.getCalculatedState
  currentHash <- dataApplication.hashCalculatedState(current._2)

  _ <-
    if (current._1 === targetOrdinal && currentHash === expectedProof)
      Async[F].unit // Exact retry: application already installed T.
    else if (current._1 < targetOrdinal)
      dataApplication
        .setCalculatedState(targetOrdinal, recoveredState)
        .flatMap(accepted => Async[F].raiseUnless(accepted)(CalculatedStateRejected(targetOrdinal)))
    else
      CalculatedStateRecoveryConflict(
        currentOrdinal = current._1,
        targetOrdinal = targetOrdinal
      ).raiseError[F, Unit]

  _ <- installExactCanonicalCurrencySnapshot(...)
} yield ()
```

This uses the existing `getCalculatedState` and `setCalculatedState` API. It does not require expanding the metagraph SDK. Same-ordinal/different-hash and application-ahead cases should halt rather than silently overwrite application state.

## Requested tests

Please add focused coverage for:

- Twelve distinct facilitator proposal views, with the oldest proposal selected after normal Global retention has advanced/pruned.
- Restart after a fee receipt is durable but before binary construction.
- Missing or mismatched receipt fails before binary signing.
- Receipt cleanup only after selection abandonment or durable binary/outbox completion.
- Crash after calculated-state file persistence.
- Crash after application `setCalculatedState(T)` but before canonical Currency snapshot installation; the next download attempt detects exact application state and completes.
- Same ordinal with a different calculated-state hash fails closed.

This keeps the useful recovery and correctness fixes while avoiding increased full Global-context heap retention and removing the unsupported fixed capacity of four.

Proph151Music:
Glad to help! I made the changes you requested and added the related tests. Everything is passing locally, and I updated the PR description with the new details.

scasplte2:
The fee-context receipt redesign addresses my earlier architectural concern. I have one remaining correctness blocker and one mechanical fix before approval.

### Blocking: detect calculated-state conflicts before mutating disk

`Download.restoreCalculatedStateSteps` currently fetches and atomically replaces the target ordinal's calculated-state file before it reads the application's current state:

```scala
state <- calculatedState.fetchExact(ordinal, expected)
_ <- calculatedState.persistAtomically(ordinal, state)
// ...
current <- calculatedState.getCurrent
// same-ordinal conflict / application-ahead detected here
```

As a result, the same-ordinal/different-hash and application-ahead paths report a fail-closed conflict only after the framework file has already been replaced. That can leave the application and its persisted calculated-state file disagreeing at the point where recovery halts.

Please reorder this narrow boundary:

1. Read and hash `getCurrent` first.
2. If current is ahead, or is at the target ordinal with a different hash, fail before fetching or persisting anything.
3. If current already exactly matches the target ordinal and certified hash, atomically persist/read-back that state and continue idempotently.
4. Only when current is behind should recovery fetch the certified target state, verify it, atomically persist/read-back it, and call `setCalculatedState`.

The existing conflict/ahead tests should assert that neither `fetchExact` nor `persistAtomically` was called. An exact-current test should prove that no peer fetch is required.

This does not require a new SDK method or broader recovery design.

### Mechanical

A fresh checkout of head `c483f51f` fails `sbt scalafmtCheckAll` because `CurrencyFeeContextReceiptStorage.scala` is not formatted. Please commit the formatter output.

For reference, I independently ran the current suites after formatting: Currency L0 153/153, shared 313/313, and the six affected node-shared tests all passed.

Proph151Music:
Thanks for catching this. I reordered the recovery check so conflicting or newer application state is detected before any peer fetch or disk change. An exact matching state now continues without a peer fetch, and the tests verify these cases directly.

I also applied the formatter output. The full Currency L0, shared, and affected node-shared suites are passing. 🤙

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensusStateAdvancer.scala
```diff
@@ -348,7 +351,9 @@ object CurrencySnapshotConsensusStateAdvancer {
                                     majorityArtifactInfo.hash,
                                     AttemptDomain(facilitatorsHash, parentHash, state.lastOutcome.finished.binaryArtifactHash)
                                   )
-                                  effect = consensusStorage.addSignature(selfId, state.key, declaration, declaration.domain.some) >>
+                                  effect = feeContextReceiptStorage
+                                    .retainSelected(CurrencyFeeContextKey(state.key, majorityArtifactInfo.hash))
```
ryle-ai: There's an asymmetry between the capture side and this consumption site that I think can wedge the retained-effect machinery. `captureFeeContextReceipt` treats `globalSyncView = None` as a successful no-op (`globalSyncView.traverse_`), so a view-less artifact passes validation without ever writing a receipt -- and then `retainSelected` here raises `MissingCurrencyFeeContextReceipt`. The problem is where that raise lands: this is part of the retained phase effect, and `ConsensusStorage.retryRetainedEffect` retries retained effects unboundedly while deliberately never completing the phase's `completion` on failure (that ordering is load-bearing for persistence, per its own comment). Missing/Conflict are permanent errors -- nothing recreates a receipt for this exact key once the state has advanced past validation -- so the retry can never succeed and the generation stays wedged until the peer-ahead reanchor tears it down.

Contrast with the pre-PR failure mode for the same artifact: `createSynchronousBinaryValue` raised "no signed globalSyncView" inside the advance function, which the manager catches (ConsensusManager.scala:473, "Error checking for consensus state update") and simply re-evaluates on the next trigger.

If view-less artifacts are unreachable on current develop then this is theoretical, but in that case the cheap fix is to make validation fail closed on `globalSyncView = None` in this engine so such an artifact can never become the selected majority -- and/or do the `retainSelected` existence check in the advance, before returning newState + effect, keeping only the sibling deletion inside the retained effect. Either preserves the fail-closed intent while leaving permanent errors in the recoverable error channel.

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/CurrencySnapshotConsensusFunctions.scala
```diff
@@ -99,7 +177,7 @@ object CurrencySnapshotConsensusFunctions {
         .flatTap {
           case Invalid(errors) =>
             logger.warn(s"Failed when validating currency artifact. Errors: ${errors.toList}")
-          case _ => Async[F].unit
+          case _ => captureFeeContextReceipt(artifact, lastContext)
```
ryle-ai: Raising out of this flatTap changes the error channel in a way the caller doesn't handle. `pickValidatedMajorityArtifact` deals with the Either result gracefully -- an invalid majority logs a warning and falls through to `go(tail)` -- but `ExactGlobalFeeContextUnavailable` raised here aborts the whole pick and surfaces as the manager's generic "Error checking for consensus state update", re-raised on every trigger until it clears. And the unavailable case is transient by nature: the proposer's `globalSyncView` can be one GL0 sync ahead of this validator, and `LastSyncGlobalSnapshotStorage.getCombined` only covers the in-memory window plus the with-state file storage. The old code had the same requirement but deferred it to binary construction, which gave GL0 sync the whole proposal-and-signature phase to catch up.

Since a node that can't produce the receipt can't sign that artifact anyway, mapping the *unavailable* failure into the Invalid channel (this function already returns Either) gets the same fail-closed outcome while attributing the failure to the artifact instead of blowing up the advance. `ExactGlobalFeeContextHashMismatch` is a different animal -- same ordinal, different GL0 hash is fork evidence and arguably deserves to stay loud -- so I'd only remap the unavailable case. Not a correctness issue either way; fees can't come out wrong. This is about how gracefully a slightly-lagging validator degrades.

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/storage/StateChannelBinaryOutboxStorage.scala
```diff
@@ -376,26 +376,33 @@ object StateChannelBinaryOutboxStorage {
...
-                      val confirmed = ordered
-                        .filter(entry =>
-                          entry.locallyCommitted &&
-                            (entry.currencySnapshotOrdinal < currencyOrdinal ||
-                              (entry.currencySnapshotOrdinal === currencyOrdinal && entry.binaryHash === binaryHash))
+                    // A downloaded validator intentionally discards its local publication copies. It can therefore have a valid pending
+                    // binary N+1 while its local GL0 view still confirms only N-1. The missing binary N may already be queued at GL0 and
+                    // will close the gap in a later Global snapshot. Keep the suffix pending until the canonical tip reaches N; then the
+                    // ordinary direct-parent or same-ordinal hash check below can prove it. A direct conflicting successor still fails.
+                    if (missingCanonicalBridge) List.empty[Entry].pure[F]
```
ryle-ai: The shape this replaces failed loudly; the wait is silent. A real conflict still surfaces eventually -- the parent-hash check runs once the tip reaches the bridge ordinal -- so the only unbounded-wait case is the canonical tip never advancing at all. Still, I'd add a log line on this branch naming the gap (canonical tip ordinal, first pending ordinal), since during an incident "waiting for GL0 to close the gap" and "outbox stuck" are indistinguishable without it. Also noting for the record: while the gap is open this branch skips confirming locally-committed entries at or below the tip too; they self-heal on the next confirmCanonicalTip call, so no change needed, just something a future reader shouldn't puzzle over.

modules/currency-l0/src/main/scala/io/constellationnetwork/currency/l0/snapshot/programs/Download.scala
```diff
@@ -60,6 +162,26 @@ object Download {
   private[snapshot] def observationLimit(current: SnapshotOrdinal, offset: Long): SnapshotOrdinal =
     SnapshotOrdinal.unsafeApply(Math.addExact(current.value.value, offset))
 
+  private[snapshot] def observationPlan(
+    checkpoint: SnapshotOrdinal,
+    peerTip: SnapshotOrdinal,
+    offset: Long
+  ): Either[PeerTipBehindCheckpoint, ObservationPlan] =
+    if (peerTip < checkpoint) PeerTipBehindCheckpoint(checkpoint, peerTip).asLeft
```
ryle-ai: Naming: the first argument is the ordinal of the live head from `getLatestHead` (start passes `result._1.ordinal`), not the periodic checkpoint -- this call site replaced the checkpoint-based `getLatest`. So `PeerTipBehindCheckpoint` and its "behind its combined checkpoint" message would mislead whoever reads it in a log; what it actually detects is the peer's latest-ordinal read regressing below the combined body it just served. Something like `PeerTipBehindDownloadedHead` says what it is.

modules/shared/src/main/scala/io/constellationnetwork/currency/dataApplication/storage/CalculatedStateLocalFileSystemStorage.scala
```diff
@@ -45,6 +52,41 @@ final class CalculatedStateLocalFileSystemStorage[F[_]: Async] private (
     }
 
   private def toOrdinalName(ordinal: SnapshotOrdinal): String = ordinal.value.value.toString
+
+  private def writeAtomicBytes(fileName: String, bytes: Array[Byte]): F[Unit] = {
+    val directory = path.toNioPath
+    val target = directory.resolve(fileName)
+
+    Async[F].bracketCase(Async[F].blocking(JFiles.createTempFile(directory, s".atomic-$fileName.", ".tmp"))) { temp =>
```
ryle-ai, line 60: A hard kill between createTempFile and the move leaves `.atomic-<n>..tmp` files behind -- bracketCase only covers in-process failures. `listStoredOrdinals` ignores them (`toLongOption`), so it's accumulation rather than breakage, but a `*.tmp` sweep in `make()` would keep the directory from collecting them across crashes.

ryle-ai (commented):
Read the three commits against the repo rather than the diff, and specifically traced the calculated-state / data-application machinery since that's the risky surface here. For what it's worth on that front: the exact-ordinal route landed in the p2p (peer-authenticated) route set rather than public; the per-ordinal files it serves are written by DataApplicationSnapshotAcceptanceManager on every accepted ordinal and only pruned upward, so the fetch has real data behind it; and the crash/retry cases in restoreCalculatedStateSteps (retry-after-persist, exact-current retry, app-ahead-of-target) hold up when I walk them against the acceptance manager's own idempotent-repair path.

Two asks that matter, both about where the new fee-receipt requirement fails rather than whether it should fail: the receipt-missing case inside the retained signature effect can wedge the retry machinery permanently (inline on the advancer), and the receipt capture raising out of validateArtifact changes a transient GL0-sync lag from a skip-this-proposal condition into an abort-the-whole-pick condition (inline on consensus functions). The rest are small.

One operational note for whoever sequences the release: a recovering data-application validator now requires at least one Ready peer serving the new `/currency/state/calculated/<ordinal>` route. During a rolling upgrade, the first upgraded validator to lose local state cannot recover from old-version peers -- it fails closed and retries, which is the right behavior, but it means upgrade order matters for metagraphs that expect to exercise recovery mid-rollout.

Caveat on confidence: all of this is traced from source (this branch plus the synchronous engine it sits on), not reproduced in a topology run, so treat the wedge claim as read-from-source confidence rather than an observed failure.

Proph151Music:
Thanks for taking the time to review all of this! I went through each point and completely agree with the concerns you've raised.
I’ll update the PR so a missing fee receipt cannot trap consensus, and a validator that is temporarily behind Global L0 only skips the affected proposal instead of stopping the entire selection attempt. I’ll keep genuine hash conflicts as hard failures.
I’ll also add the outbox waiting log, correct the misleading recovery error name, clean up abandoned temporary files during startup, and add focused tests for each change. I’ll post another update once I can confirm everything passes. 👊

Proph151Music:
I’ve addressed all five points and pushed the updated branch. Missing Global views now fail validation, temporary exact-context gaps reject only the affected proposal, and genuine hash conflicts still fail loudly. The selected receipt is also checked before the consensus state changes. I added the bridge-wait log, corrected the recovery error name, cleaned abandoned atomic temporary files during startup, and added the related tests. All local suites and every GitHub check are passing. 🤘

scasplte2 (approved):

PR #1588, opened 2026-09-01
fix(dag-l0): recover past damaged snapshot files

Proph151Music:
PR 1566 was merged into `develop` and its source branch was deleted. GitHub automatically closed the earlier PR 1585 because it targeted that deleted branch. This replacement contains the same validated damaged-file recovery changes on current `develop`.

The merged PR 1566 commit `ee1c7e08173bd63cd7c2f9e7b66699ef734374e1` has the same source tree as the previous PR 1566 head `490011e1b648396378b275f08ae3f4deafc6818f`. No source or behavioral changes were needed during the base transfer.

## Why this is needed

I found that a Global L0 validator could remain in `WaitingForDownload` even when healthy peers had valid snapshot history available. The validator kept selecting the same damaged local snapshot or snapshot-info file and repeating the download failure. Manual restoration of the damaged file was the only way to move it forward.

The problem was not missing peer data. Tessellation treated a local filename as a usable recovery anchor without first proving that the file could be decoded. Decoder exceptions could also escape the storage layer instead of being returned as an unreadable-file result. When the ordinal and hash links both referred to damaged bytes, retrying the same ordinal reproduced the same failure.

In the untouched control, unreadable snapshot-info caused 13 repeated download failures and the validator did not recover on its own. When two validators were damaged at the same time, the remaining five validators preserved finality, but both damaged validators remained stranded.

PR 1566 already added substantial crash-safe snapshot storage, so the older patch was not copied without review. The audit found four cases that were still missing:

1. Exceptions from the primary and fallback deserializers were not always contained as an unreadable-file result.
2. Snapshot-info availability was based on filename existence instead of a successful decode.
3. Recovery could select the same unusable persisted ordinal again.
4. Installing a validated same-ordinal replacement could leave the stale ordinal link pointing at the older snapshot.

## What this changes

Contain JSON and fallback decoder exceptions and report the damaged file as unreadable. Decode snapshot-info before accepting it as a recovery anchor. If the snapshot or metadata at an ordinal cannot be read, continue the search below that ordinal instead of selecting it again.

Before installing a validated same-ordinal replacement, remove the stale ordinal link. This keeps the hash and ordinal indexes pointed at the same recovered snapshot.

## How this was verified

The isolated seven-validator recovery matrix covered:

- unreadable snapshot-info;
- unreadable snapshot bytes through both local indexes;
- valid snapshot-info paired with the wrong snapshot;
- a missing ordinal link with a valid hash link;
- readable snapshot-info with both snapshot indexes missing;
- a corrupt Merkle Patricia Trie; and
- two validators with unreadable local state at the same time.

All cases recovered with the fix. In the two-validator damage case, the other five validators preserved finality, both damaged validators returned to `Ready`, and all seven later reported the same snapshot ordinal and hash. The recovered validators also returned to the active signing group.

The original campaign passed 45 focused tests plus compile, assembly, formatting, and source-diff checks.

The current replacement branch passed:

- `LocalFileSystemStorageSuite`;
- `SnapshotLocalFileSystemStorageSuite`;
- `SnapshotDownloadStorageValidatedSuite`;
- `SnapshotStorageSuite`;
- `DownloadSuite`;
- 69 focused tests in total;
- `scalafmtCheckAll`; and
- `git diff --check`.

The seven-validator matrix is preserved evidence for the original behavior. The focused current-`develop` run is not presented as a new seven-validator campaign.

## Safety

Downloaded data is still validated before it becomes persisted state. This change does not trust corrupt local bytes, skip snapshot validation, or weaken finality. All runtime testing used isolated networks and throwaway identities. No public network was joined or modified.

### DIFF
--- a/modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/domain/snapshot/programs/Download.scala
+++ b/modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/domain/snapshot/programs/Download.scala
@@ -1563,11 +1563,13 @@ object Download {
               snapshotStorage.readCombined(ord).flatMap {
                 case Some(result) => result.some.pure[F]
                 case None         =>
-                  // readCombined self-healed by deleting the bad pair at `ord`. The next call to
-                  // getHighestSnapshotInfoOrdinal will return a strictly-lower ord, but we also
-                  // pass `ord` explicitly (rather than `ord - 1`) to keep the bound aligned with
-                  // any new write races. SnapshotOrdinal.MinValue terminates if we reach genesis.
-                  findHighestValidPersisted(ord, attempts - 1)
+                  // Always decrease the search bound. A missing or unreadable snapshot can leave
+                  // its info file in place, so asking for the highest info at `ord` again would
+                  // select the same unusable pair until the attempt budget is exhausted.
+                  PartialPrevious[SnapshotOrdinal]
+                    .partialPrevious(ord)
+                    .map(findHighestValidPersisted(_, attempts - 1))
+                    .getOrElse(none[(Signed[GlobalIncrementalSnapshot], GlobalSnapshotInfo)].pure[F])
               }
           }
 
--- a/modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/SnapshotDownloadStorage.scala
+++ b/modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/SnapshotDownloadStorage.scala
@@ -120,8 +120,8 @@ object SnapshotDownloadStorage {
 
       private def hasSnapshotInfo(ordinal: SnapshotOrdinal): F[Boolean] =
         hashSelect.select(ordinal) match {
-          case JsonHash => snapshotInfoStorage.exists(ordinal)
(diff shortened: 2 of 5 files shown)


scasplte2:
Requested change before inclusion: please keep the decoder-containment, successful-decode anchor check, and strictly decreasing damaged-history search, but replace the promotion sequence delete ordinal -> move temporary file -> link snapshot with persistedStorage.replaceForRecovery(snapshot). Clean the temporary copy only after that replacement succeeds, if cleanup is still needed.\n\nThe reason is that the current LocalFileSystemStorage move and link operations treat NoSuchFileException as a warning and successful return. After deleting the ordinal index, a missing temporary source can therefore let moveTmpToPersisted report success without restoring a canonical ordinal index.\n\nThe rc.13 replaceForRecovery helper already atomically replaces both hash and ordinal indexes, removes an abandoned same-ordinal hash, and verifies exact disk readback. It also has crash/retry coverage. Please add a promotion-level regression covering a missing or failed temporary promotion and a restart/retry that preserves or converges to the exact canonical ordinal. No change is requested to the other recovery fixes in this PR.

Proph151Music:
I made the requested change and added the restart and retry test. Everything is passing locally for me. 🙌

scasplte2 (approved):

PR #1589, opened 2026-09-01
fix(dag-l0): emit replacement vote after round commit

Proph151Music:
PR 1566 was merged into `develop` and its source branch was deleted. GitHub automatically closed the earlier PR 1586 because it targeted that deleted branch. This replacement contains the same validated round-start vote fix on current `develop`.

The merged PR 1566 commit `ee1c7e08173bd63cd7c2f9e7b66699ef734374e1` has the same source tree as the previous PR 1566 head `490011e1b648396378b275f08ae3f4deafc6818f`. No source or behavioral changes were needed during the base transfer.

## Why this is needed

I found this while testing five-validator Global L0 recovery. When one authority validator stopped responding over P2P, the three Core validators correctly selected it for replacement by a healthy synchronized validator. At the allowed membership-change round, all three logged `skipped=no_state`. No signed eviction vote was produced, so the replacement could not happen.

The problem is in Tessellation round startup. The new consensus round is constructed, but the eviction voter is called before `ConsensusStorage` commits that round. The voter reads `ConsensusStorage` to find the round it must sign. Because the round is not there yet, it skips the vote.

## What this changes

Keep the prepared replacement action with the new state. After the state is committed, emit or reuse the vote, check eviction-certificate assembly, and then send the normal Facility message.

This does not change quorum, finality, certificate validation, target selection, or membership policy.

## How this was verified

I repeated the same isolated five-validator test with the companion Currency recovery and damaged-file recovery changes applied.

Stock result:

- Three Core validators selected the same silent member.
- All three logged `skipped=no_state`.
- Zero signed eviction votes were produced.
- The healthy spare could not enter authority.

Fixed result:

- The same three Core validators emitted three signed votes.
- There were zero `no_state` skips.
- The eviction and admission certificates reached quorum before proposal creation.
- Snapshot ordinal 36 certified removal of the silent validator and admission of the healthy spare.
- Authority stayed at four validators instead of shrinking.
- The four healthy authority nodes stayed synchronized through ordinal 41.
- All containers had zero restarts and no OOM kills.

The current replacement branch passed both round-start regression tests, `scalafmtCheckAll`, and `git diff --check`. The previously completed full standalone DAG L0 suite passed 231 tests with two pre-existing ignored tests. The combined candidate suite passed 233 DAG L0 tests with two pre-existing ignored tests and all 1440 Node Shared tests. DAG L0 assembly also passed.

This confirms that the defect remained after the companion fixes and that the round-start ordering change fixes it independently.

## Safety

The fix does not alter quorum, finality, certificate validation, target selection, or membership policy. All runtime testing used isolated networks and throwaway identities. No public network was joined or modified.

### DIFF
--- a/modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotConsensusStateCreator.scala
+++ b/modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/GlobalSnapshotConsensusStateCreator.scala
@@ -55,6 +55,16 @@ abstract class GlobalSnapshotConsensusStateCreator[F[_]: Sync]
 
 object GlobalSnapshotConsensusStateCreator {
 
+  /** Build the round-start atomic-replacement effect without running it. The state creator retains this effect with the new consensus
+    * state, so the voter can read the committed round before it signs or stores a vote. Assembly is checked before the first Facility.
+    */
+  private[snapshot] def atomicReplacementRoundStartEffect[F[_]: Sync](
+    alreadyVoted: Boolean,
+    emitVote: F[Unit],
+    checkAssembly: F[Unit]
+  ): F[Unit] =
+    (if (alreadyVoted) Sync[F].unit else emitVote) >> checkAssembly
+
   private def resetLegacyOutcomeToAuthenticatedSeed[F[_]: cats.Functor: Hasher](
     outcome: GlobalConsensusOutcome,
     seed: List[PeerId]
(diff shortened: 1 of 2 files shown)


scasplte2:
@Proph151Music The Validate PR title job is not failing on the PR title. The current title, fix(dag-l0): emit replacement vote after round commit, passes commitlint.

The failure comes from the second commitlint invocation, which also validates commit messages. Commit 9498713061de1fc659fb2c717a0fa512589d3620 fails body-max-line-length because its body paragraphs are stored as lines longer than 100 characters.

Please amend that commit message and reflow every non-URL body line to 100 characters or fewer, then force-push with force-with-lease. No source change or PR-title change is required. The validation workflow should rerun after the amended commit is pushed.

Once this is done, I'll merge in this update as it should go into the pending Integrationnet restart

Proph151Music:
My bad! I corrected the message formatting and pushed the updated commit. 🤙

scasplte2 (approved):

PR #1592, opened 2026-09-03
fix(dag-l0): settle delegated stake withdrawals exactly once

scasplte2:
## Summary

Prevent duplicate delegated-stake withdrawals and settle legacy duplicate pending records once on Mainnet v3.5. For example, if one token lock appears in two withdrawal address buckets with cumulative rewards of 20 and 30, the activated rule pays 30 to the active token-lock owner and releases that lock's principal once. Previously, the reward path could pay both address keys even when principal unlocking was deduplicated.

This replaces #1555 against `release/mainnet`. Its old activation ordinal, 6,710,600, is already in signed Mainnet history and must not be reused. The new `fixing-delegated-stake-double-withdrawal` mapping remains dev-only; absent public-network entries resolve to `Long.MaxValue`.

## Behavior at and after activation

- Create and withdrawal candidates use total signed-envelope ordering. Only accepted candidates reserve identities, so invalid or rejected events cannot poison later valid events.
- An accepted replacement create wins over a same-round withdrawal. Existing pending token-lock references are checked globally for both creates and withdrawals, including malformed address buckets and stake sources. At most one withdrawal per v3.5 source/token-lock identity is accepted in a batch; settlement deduplicates globally by token-lock hash.
- A single settlement result, prepared before reward calculation, supplies both reward payouts and principal unlocks. It groups globally by token-lock hash, selects the greatest cumulative reward across every copy cleanup retires (including later-cooldown copies), resolves ties deterministically, and takes ownership from the active token lock.
- Distinct locks belonging to one owner have their rewards summed with checked arithmetic. Duplicate records are not summed. Withdrawal payout realizes rewards already counted when accrued, so it is not added again to current-round emission totals.
- Current-round issuance totals also use checked addition. Reward balance overflow aborts settlement rather than silently skipping the payout and retiring its pending record. Exact numeric bounds, ordinary overflow, and multi-value positive-wrap inputs are covered.
- Missing-lock records receive neither reward nor principal. Cleanup retires every pending copy of a processed lock, including copies in other buckets with later cooldowns. This preserves Mainnet's nonfatal handling of malformed state.
- Natural token-lock expiry overlapping delegated withdrawal produces one principal credit and one unlock artifact.
- Counted diagnostic markers identify normalization and missing locks: `DELEG_STAKE_WITHDRAWAL_DEDUP` and `DELEG_STAKE_WITHDRAWAL_ORPHAN`, each including the snapshot ordinal.

Below activation, historical candidate handling, reward address selection/last-wins behavior, unlock grouping, and cleanup semantics are preserved. The settlement result is an internal calculation value with no snapshot codec or Kryo registration. No signed schema, v4.1 `currentTokenLockRef`, or replacement-token-lock behavior is backported.

## Validation

Validated at [8c83ac1e9](https://github.com/Constellation-Labs/tessellation/commit/8c83ac1e9e728fff7dde45530a6d088635de25dd), using Java 11, matching the CI Java family:

- Full `nodeShared/test`: **283 passed**, zero failures/errors.
- Full `dagL0/test`: **106 passed**, zero failures/errors; two existing ignored generator tests.
- Whole-project `sbt compile`, `scalafmtCheckAll`, `scalafixAll --check --rules OrganizeImports`, complete-range commitlint, and diff checks passed.

Relevant suites included in those totals:

| Suite | Passed |
|---|---:|
| `UpdateDelegatedStakeAcceptanceManagerSuite` | 12 |
| `UpdateDelegatedStakeValidatorSuite` | 22 |
| `DelegatedStakeWithdrawalOrdinalConfigSuite` | 1 |
| `GlobalSnapshotDelegatedStakeUnlockSuite` | 4 |
| `DelegatedRewardsDistributorSuite` | 6 |
| `GlobalDelegatedRewardsDistributorSuite` | 10 |
| `GlobalSnapshotDelegatedStakeSettlementSuite` (DAG L0) | 13 |

The new settlement suite uses the real reward distributor, reward balance application, and production token-lock transition. It covers A-1/A/A+1, two address buckets sharing one lock, greatest-reward selection, deterministic ties, malformed ownership, distinct-lock reward sums, orphans, cleanup across cooldowns, repeat settlement, natural-expiry boundaries, zero rewards, and overflow. The acceptance additions exercise the real validator for both normally stored and malformed pending records.

Docker E2E was not rerun locally for this follow-up; GitHub CI runs on the updated PR. #1595 remains the stacked concurrent-submission/settlement regression PR.

## Before public activation

The maximum-entitlement policy assumes overlapping cumulative stake lineage; synthetic tests do not prove that assumption for actual malformed state. Complete the [ordinal/hash-pinned settlement and lineage audit](https://github.com/Constellation-Labs/tessellation/blob/8c83ac1e9e728fff7dde45530a6d088635de25dd/docs/operations/delegated-stake-settlement-audit.md), including prior payouts, staggered cooldowns, ownership, backing, and exact-integer arithmetic. No completed public-network lineage census is claimed; unresolved groups block activation. #1593 carries the corresponding develop implementation while preserving v4.1 replacement semantics. This change prevents future duplicate settlement; it does not reverse historical balance effects.

Deploy the reviewed binary through the normal full-cluster cold restart. Select and announce a strictly future Mainnet ordinal, add its mapping in a reviewed follow-up, and qualify A-1/A/A+1 on the Mainnet topology before activation. No public activation is included here.

### DIFF
--- a/docs/operations/delegated-stake-settlement-audit.md
+++ b/docs/operations/delegated-stake-settlement-audit.md
@@ -1,0 +1,75 @@
+# Delegated-stake settlement activation audit
+
+The `fixing-delegated-stake-double-withdrawal` gate is a replay/state-transition change.
+Do not activate a public network until the evidence below is recorded and reviewed.
+Synthetic tests establish the algorithm's behavior, not the provenance of live rewards.
+
+## Entitlement policy and its limits
+
+One backed token lock is settled once. Duplicate rewards are never added together.
+For overlapping cumulative records of one stake lineage, the largest recorded reward is the
+most complete known entitlement. Selection includes all copies retired at that settlement,
+including copies with later withdrawal cooldowns. Otherwise cleanup could destroy a larger
+entitlement after paying an earlier, smaller one.
+
+This is an explicit policy assumption, not a claim that arbitrary corrupted rewards are valid.
+Numerically ordering two amounts does not prove that their accrual histories overlap.
+Switching to the minimum does not validate the records either: it can underpay a genuine lineage,
+and neither minimum nor maximum reconstructs independent or previously paid entitlements.
+Unexplained groups block public activation; do not invent counts or an audited ordinal.
+
+The active signed token lock determines the recipient and principal. The pending bucket and
+signed stake source are not authoritative for malformed state. Distinct locks stay distinct,
+and their entitlements are combined only with checked `Amount.plus`. An arithmetic failure
+aborts settlement; it must not wrap, clamp to zero, silently skip a reward, or clean pending state.
+Withdrawal payout realizes previously accrued rewards and is not counted again as new issuance.
+
+## Required evidence
+
+Preserve an immutable source artifact, its Global Snapshot ordinal and hash, retrieval time,
+network, and independent canonical-lineage verification. Record:
+
+1. Total pending records, unique effective token locks, duplicate groups and duplicate copies.
+2. Each group's original/effective lock references, stake/parent references, accepted ordinals,
+   withdrawal epochs, owners, active backing and rewards. On v4.1 use `currentTokenLockRef`
+   when present; the original event hash is not necessarily the live backing.
+3. Historical create/redelegation/withdrawal/replacement events and accrual history establishing
+   whether the group's rewards are cumulative and overlapping. Check for earlier settlements.
+4. Ownership mismatches, missing backing, active-stake overlap, staggered cooldowns,
+   independent/non-nesting lineages and arithmetic failures. Resolve every unexplained anomaly.
+5. Exact proposed payouts, lock removals and retained/retired pending records, compared with the
+   validated lineage. Audit arithmetic must use exact integers, not floating-point JSON numbers
+   or unchecked fixed-width sums. Record tool version, input digest and reproducible commands.
+
+A snapshot census alone can count duplicates; it cannot establish historical nesting or prove
+that no earlier payout occurred. Keep both the census and lineage evidence linked from the PR
+and release dossier. This document does not record a completed network audit.
+
+## Branch-specific behavior and qualification
+
+- Mainnet #1592: https://github.com/Constellation-Labs/tessellation/pull/1592
+- Develop #1593: https://github.com/Constellation-Labs/tessellation/pull/1593
+
+Both gates cover acceptance, reward/principal settlement, checked withdrawal and current-issuance
+totals, fail-on-overflow reward credits, and retirement of settled duplicate pending records.
+Natural expiry is a separate balance path: suppress generated principal unlocks when
+`unlockEpoch < currentEpoch`, so a lock is not credited by both paths. Equality is not yet natural
+expiry. Public mappings remain absent until the audit and coordinated activation are approved.
+
+Mainnet skips missing-lock reward/principal payouts and retires processed orphan copies.
+Develop retains missing-lock pending records. In particular, when a replacement rewrites OLD
+to NEW at R, NEW is not in the last-active set until R+1: preserve the pending reward, unlock
+OLD at R, then settle NEW once at R+1. Do not copy Mainnet orphan deletion into that deferral.
+Retire later-cooldown copies only for effective locks that are actually settled.
+
+Qualify A-1/A/A+1, original and effective refs, wrong-source highest-reward records, distinct
+locks, deterministic ties, staggered cooldowns, repeated settlement, zero rewards, natural expiry
+at both boundaries, exact numeric bounds and positive-wrap overflow inputs. Exercise the real
+reward distributor, balance credit, principal transition and cleanup together. For develop,
+include nonzero-reward replacement R/R+1 coverage.
+
+The old branch remains available below activation solely to reproduce retained signed history;
+historical unchecked arithmetic is not the implementation to use for new settlement. This fix
+is not a repository-wide arithmetic audit and does not undo historical balance effects.
+Use the normal single-version full-cluster cold restart and a strictly future activation ordinal.
+No signed schema, codec or hash construction is changed.
--- a/modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/rewards/GlobalDelegatedRewardsDistributor.scala
+++ b/modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/rewards/GlobalDelegatedRewardsDistributor.scala
@@ -521,7 +521,9 @@ object GlobalDelegatedRewardsDistributor {
             }
           )
 
-        totalEmittedReward <- DelegatedRewardsDistributor.sumMintedAmount(
+        totalEmittedReward <- (if (partitionedRecords.withdrawalSettlement.isDefined)
+                                 DelegatedRewardsDistributor.sumMintedAmountChecked[F] _
+                               else DelegatedRewardsDistributor.sumMintedAmount[F] _)(
           reservedAddressRewards,
           nodeOperatorRewardsTxs,
           delegatorRewardsMap
--- a/modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/rewards/GlobalDelegatedRewardsDistributor.scala
+++ b/modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/rewards/GlobalDelegatedRewardsDistributor.scala
@@ -508,14 +508,17 @@ object GlobalDelegatedRewardsDistributor {
 
         withdrawalRewardTxs <-
           calculateWithdrawalRewardTransactions(
-            partitionedRecords.expiredWithdrawalsDelegatedStaking.toList.flatMap {
-              case (address, withdrawals) =>
-                withdrawals.toList.mapFilter { withdrawal =>
-                  Option.when(withdrawal.rewards.value > Balance.empty.value) {
-                    (address, Amount(NonNegLong.unsafeFrom(withdrawal.rewards.value.value)))
+            partitionedRecords.withdrawalSettlement.map(_.rewardsByAddress.toMap).getOrElse {
+              // Preserve address-keyed, last-wins reward settlement exactly before activation.
+              partitionedRecords.expiredWithdrawalsDelegatedStaking.toList.flatMap {
+                case (address, withdrawals) =>
+                  withdrawals.toList.mapFilter { withdrawal =>
+                    Option.when(withdrawal.rewards.value > Balance.empty.value) {
+                      (address, Amount(NonNegLong.unsafeFrom(withdrawal.rewards.value.value)))
+                    }
                   }
-                }
-            }.toMap
+              }.toMap
+            }
           )
 
         totalEmittedReward <- DelegatedRewardsDistributor.sumMintedAmount(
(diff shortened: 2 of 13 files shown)


scasplte2:
@ryle-ai This is ready for technical review. Before enabling the fix on Mainnet, we still need to select and announce a new strictly future activation ordinal, add the Mainnet mapping, and rerun the A-1/A/A+1 boundary qualification. The current Mainnet mapping remains absent and fails closed.

ryle-ai (commented):
The principal path here is right and the comment on it states the right principle — "the active lock, not the withdrawal map key, defines the owner." The problem is that `expiredWithdrawalsDelegatedStaking` has two consumers and this PR only fixes one of them.

`GlobalDelegatedRewardsDistributor.scala:510-519` still does:

```scala
partitionedRecords.expiredWithdrawalsDelegatedStaking.toList.flatMap {
  case (address, withdrawals) =>
    withdrawals.toList.mapFilter { withdrawal =>
      Option.when(withdrawal.rewards.value > Balance.empty.value) {
        (address, Amount(NonNegLong.unsafeFrom(withdrawal.rewards.value.value)))
      }
    }
}.toMap
```

That reads the raw pre-dedup map, keys the payout on the **map key address** rather than the active lock's `source`, and terminates in `.toMap`. So for the exact state this PR exists to defend against — a duplicate pending withdrawal for one `tokenLockRef` sitting under two address keys — post-activation you get principal credited once to the true owner (correct, that's the fix) while the reward leg pays twice, once to each key, including to whichever key isn't the owner. For a duplicate under a single address, `.toMap` collapses to last-wins, which means principal and rewards can resolve from two different records with two different `rewards` values; last-wins isn't "smallest" or "the one that matches the unlock," it's just whatever iteration order hands over last.

The reason I think this is worth blocking on rather than filing separately: the justification for the post-activation branch is that malformed duplicate pending state already exists in signed history. `release/mainnet` is the branch that carries it. So the reward-side exposure isn't hypothetical-in-principle — it's live in precisely the state this PR is written for, and the fix as it stands hardens the cheaper of the two legs (principal, which is bounded by the lock amount) while leaving the leg that mints (rewards) alone.

Second-order thing that makes it harder to notice: `withdrawalRewardTxs` is not passed into `sumMintedAmount` (`:520-525` only sums reserved + node-operator + delegator). So a doubled withdrawal reward doesn't show up in the emitted-total accounting at all, and there's no supply invariant in GL0 that would catch it downstream either.

`develop` already has the shape you want — `deduplicateExpiredWithdrawalsByTokenLock` in `DelegatedRewardsDistributor.scala` (#1593), global `groupBy(_.tokenLockRef)`, keep the greater `rewards`, regroup by `_.event.source`. Porting that here behind the same `fixingDelegatedStakeDoubleWithdrawal` ordinal seems like the smaller change than reworking the call site, and it keeps the two branches converged. If you'd rather not pull it in, the minimum is deduping by `tokenLockRef` and taking the address from the active lock's `source` so the two legs can't disagree about who the owner is.

Structural note, take it or leave it: `expiredWithdrawalsForUnlock` is computed around `:736`, roughly two hundred lines after the rewards call, so the deduped set physically isn't in scope where the reward decision gets made. That's most of why this got missed. If the dedup produced a single value consumed by both legs, the invariant would be hard to violate instead of needing to be remembered twice.

On the parts I did check and think are fine: the activation wiring (`getOrElse(cfg.environment, SnapshotOrdinal.unsafeApply(Long.MaxValue))`, only `dev: 0`, nothing for Mainnet/Testnet/Integrationnet) is the correct shape, and not reusing #1555's ordinal is the right call for the reason the config comment gives. The absent `acceptedTokenLocks` / `OutdatedTokenLock` barrier I initially flagged is a non-issue — `replaceTokenLockRef` doesn't exist anywhere on `release/mainnet`, so there's nothing to guard. And the `Ordering.by(w => (w.source, w.stakeRef))` I was suspicious of is a total order via `Signed.SignedOrdering`'s tiebreak on `proofs`, so no cross-node nondeterminism there.

One thing I can't evaluate from the diff: `GlobalSnapshotDelegatedStakeUnlockSuite` covers the principal leg thoroughly but nothing asserts a reward-leg outcome for the duplicate case. If you add the port above, that assertion is the one I'd want in the suite, since the unit tests as they stand would stay green with the reward path broken.

ryle-ai:
Two additions to my earlier review.

The test results in the PR body aren't from this branch. It cites `TokenLockStateManagerSuite` (43 passed) and `GlobalSnapshotAcceptanceManagerSuite` (21 passed, including `currentTokenLockRef` rollover). Neither suite is in this diff, `TokenLockStateManager` doesn't exist anywhere on `release/mainnet`, and the body itself says the `currentTokenLockRef` change isn't backported. Those look like #1593's results pasted across. What this branch actually adds is `UpdateDelegatedStakeAcceptanceManagerSuite` (+347), `GlobalSnapshotDelegatedStakeUnlockSuite` (+288) and the ordinal-config suite — please restate the evidence against those, since a mainnet hotfix is exactly where I want to read the real numbers.

Second, lower confidence and I didn't finish tracing it. `UpdateDelegatedStakeValidator.validateTokenLock` consults `activeDelegatedStakes` and `activeTokenLocks` but never `delegatedStakesWithdrawals`, so a fresh Create against a lock that already has a pending withdrawal still validates. I don't think it mints — the orphan pre-filter at `:736-742` catches the second withdrawal — but the new stake looks live in the interim and the reward leg reads `activeDelegatedStakes`, so it may accrue. If that holds, rejecting creates whose `tokenLockRef` is in the pending-withdrawal set closes it in one line. Worth a look even if only to rule it out.

One thing I'd flag as a deliberate-or-not divergence: the orphan pre-filter drops with a warning, where `develop` raises `SnapshotFailure.TokenUnlockGenerationFailed` for the same condition. Dropping is the right call for a mainnet branch carrying malformed state — a raise would halt — but it means mainnet stops reporting a condition develop treats as fatal, and there's nothing that would tell you it fired other than grepping logs. A counter, or at minimum a distinct log marker, would make it visible.

scasplte2:
@ryle-ai Addressed in [6e78a36bf](https://github.com/Constellation-Labs/tessellation/commit/6e78a36bfca8a3e8464c7f2193d6ecfb27fb430c). Your reward-side finding was correct: the previous principal-only fix still allowed duplicate rewards across address buckets. The PR description's contrary claim has been removed.

- **Shared settlement:** one `DelegatedStakeWithdrawalSettlement` is now prepared before reward calculation and consumed by both reward payout and principal unlocking. It groups globally by token-lock hash, keeps the greatest cumulative reward, resolves ties deterministically, and assigns ownership from the active token lock's `source`. Distinct locks for one owner have their rewards summed with checked `Amount` arithmetic.
- **Cleanup and orphans:** missing-lock records receive neither reward nor principal. Cleanup uses the original expired references and removes all pending copies of a processed lock, including later-cooldown copies in other buckets. Mainnet retains nonfatal orphan handling, with counted `DELEG_STAKE_WITHDRAWAL_ORPHAN` and `DELEG_STAKE_WITHDRAWAL_DEDUP` markers including the snapshot ordinal.
- **Pending creates:** the ordinary case was already rejected because `validateCreateDelegatedStake` calls `validatePendingWithdrawal` separately from `validateTokenLock`. That check only searches the source bucket, though. The activated acceptance path now checks pending token-lock references globally, covering malformed buckets and stake sources. Tests exercise both cases through the real validator and verify that a rejected create does not reserve its parent ahead of an eligible create.
- **Emission accounting:** withdrawal payout settles rewards already counted when accrued. Adding it to `sumMintedAmount` would count that issuance again, so that calculation is unchanged and now documented. The new tests directly assert withdrawal transactions and resulting balances, including a payout with zero current-round issuance.

Validation on this branch: **283 nodeShared tests and 103 DAG L0 tests passed**, with two existing ignored DAG L0 generator tests. The new `GlobalSnapshotDelegatedStakeSettlementSuite` has ten tests using the real reward distributor, reward balance application, and production token-lock transition. It covers A-1/A/A+1, cross-bucket duplicates, greatest-reward selection, ownership, distinct-lock sums, deterministic ties, orphan suppression, cleanup/repeat settlement, natural expiry, zero rewards, and overflow. The acceptance suite now has 12 passing tests, including the two new pending-create cases; the existing unlock suite has 4 and the ordinal-config suite has 1. The PR body lists all relevant suite counts from this branch.

Whole-project compile passed at the committed head, along with formatting/import checks and complete-range commitlint. Local Docker E2E was not rerun; GitHub title/commit validation has passed, while Scala and E2E are still running.

All changed settlement/acceptance behavior uses the existing new activation gate. Historical behavior below it is preserved, public-network mappings remain absent, and no signed schema or v4.1 `currentTokenLockRef` behavior is backported.

ryle-ai (commented):
Re-read the delta from c734a031 to 278749e3. The two asks from my earlier review are resolved on the head: `DelegatedStakeWithdrawalSettlement.prepare` is built once in `accept` and the reward leg (`rewardsByAddress`) and principal leg (`withdrawals`) both consume it. I grepped every remaining reader of the expired map on this branch; the only raw readers left are the pre-activation fallbacks and the cleanup, which by design uses the original buckets. The selector is deterministic (max reward, then the record `Ordering`, then `acceptedOrdinal`, which covers all four fields), the owner grouping ends in SortedMap/SortedSet, and the pre-gate paths are identical to c734a031 apart from hoisting `globalActiveTokenLocksByRef` earlier. The argument for keeping `withdrawalRewardTxs` out of `sumMintedAmount` holds: those rewards were counted into issuance when accrued.

Things I'd still want a decision on, none of which I'd block on:

The three new checked paths (`prepare`, `acceptRewardTxsChecked`, `sumMintedAmountChecked`) raise a plain `BalanceArithmeticError` through `fromEither`, replacing the old silent skip. Proposers fail the round and validators fail context creation, so that's a network-wide halt class rather than a `SnapshotFailure`. The audit doc explicitly chooses abort over wrap or skip and I agree with that ordering, but "unreachable under the supply cap" is weaker than it sounds on mainnet, which has already carried 2^62-scale balances from the fee-transaction overflow. Worth stating in the PR that a halt is the intended outcome there, so nobody treats it as an accident later.

The e2e doesn't prove same-round processing, and the script says so in its own comment and log line. It asserts one pending withdrawal per lock and principal credited once, which is fine, but the workflow step title "Concurrent Withdrawal Settlement" overstates what's established. The legacy-duplicate branch is unreachable end to end on dev (gate 0, acceptance now blocks duplicates), so it's covered only by `GlobalSnapshotDelegatedStakeSettlementSuite`. That suite does run the real distributor and transition, so I'm satisfied, just don't let the step name imply more.

Behavior change worth one line in the doc: orphaned pending withdrawals (no active lock) now get no reward and are retired, where the pre-fix reward fallback paid every expired record regardless of lock presence. This is safe on release/mainnet because `validateTokenLock` requires `unlockEpoch.isEmpty` and nothing on this branch sets it later, so a staked lock cannot naturally expire and an orphan can only come from a prior settlement or malformed state. That also means the natural-expiry dedup in `applyDelegatedStakeTokenLockTransition` is dead code on this branch. Harmless, but the doc's natural-expiry discussion reads as if it applies here.

Semantic choice not called out in the body: unexpired copies of a due `tokenLockRef` are pulled into selection, so a not-yet-cooled copy with the larger cumulative reward is paid immediately rather than waiting for its own cooldown. Consistent with settle-once and with cleanup retiring all copies; just make it explicit.

Pre-existing, not introduced here: on the validator path `GlobalSnapshotContextFunctions` takes withdrawal reward txs from the artifact and only checks `acceptedRewardTxs -- artifact.rewards` is empty, so a proposer emitting extra withdrawal reward txs isn't rejected by validators. The reward-leg fix is enforced proposer-side only. Same shape on develop, so a separate ticket rather than this PR.

PR body suite counts match the head. CI on 278749e3 is green including the delegated-staking e2e.

ryle-goehausen (approved):
Ignore  latest review its nonblocking

scasplte2:
we need to announce and pick an activation ordinal so deferring for now

PR #1593, opened 2026-09-03
fix(dag-l0): harden delegated stake withdrawal settlement

scasplte2:
## Summary

Develop/v4.1 implementation of the delegated-stake withdrawal supply-safety fix, superseding #1555. It uses effective/current token-lock references and preserves replacement rollover. The corresponding Mainnet implementation is #1592, including follow-ups [6e78a36bf](https://github.com/Constellation-Labs/tessellation/commit/6e78a36bfca8a3e8464c7f2193d6ecfb27fb430c) and [8c83ac1e9](https://github.com/Constellation-Labs/tessellation/commit/8c83ac1e9e728fff7dde45530a6d088635de25dd). Both branches carry the reward and principal protections independently under the same gate name.

The September 10 follow-up [59b4aaa85](https://github.com/Constellation-Labs/tessellation/commit/59b4aaa85ae2ce77bed31059be917592a7e245be) addresses Ryle's review and additional safety gaps. The branch includes normal merges of develop at `6bde6adb7` and, on September 16, at `7bb1ef6f5` (#1596, #1597); no published history was rewritten.

- Prepare one ephemeral settlement keyed by effective token-lock hash. That same result drives reward ownership, principal eligibility and pending cleanup.
- Settle one maximum cumulative reward per backed lock, considering every pending copy being retired, including later-cooldown copies. Pay the active signed lock's owner, even if the winning pending record's signed source or address bucket is malformed.
- Keep distinct effective locks distinct even when the schema's pending-record ordering equates them. Aggregate their rewards with checked `Amount.plus`; checked issuance totals and destination-balance credits fail on overflow rather than wrapping or silently skipping payment.
- Reject creates against globally pending effective lock references before reserving the parent or lock; retain deterministic, valid-only first-wins acceptance and Create/Withdraw collision protection.
- Deduplicate principal across withdrawals and replacement unlocks. Suppress generated unlocks for `unlockEpoch < currentEpoch` because natural expiry already credits the balance; equality is tested separately.
- Retire all copies of settled effective locks across buckets. Preserve develop's missing-lock retention and R/R+1 deferral: unlock OLD and carry pending NEW at R, then settle NEW principal and rewards exactly once at R+1.
- Remove silent activation defaults from retained acceptance factories. Acceptance prepares the reward plan once; unlock generation requires an explicit `Legacy | Unique(epoch, settlement)` mode, eliminating the invalid Boolean/optional-epoch combination.
- Expand config/activation documentation and retain the count-only `DELEG_STAKE_WITHDRAWAL_DEDUP` warning.

## Entitlement policy: qualification still required

Maximum is the policy for **overlapping cumulative lineage**, not proof that arbitrary corrupted records are entitled to the greatest amount. We have not completed an ordinal/hash-pinned census and historical-lineage audit of actual network duplicates, and do not claim a verified count. Minimum is not a substitute for that audit either: it can underpay valid cumulative lineage and does not establish correctness for independent or previously paid entitlements.

The new [activation audit checklist](https://github.com/Constellation-Labs/tessellation/blob/59b4aaa85ae2ce77bed31059be917592a7e245be/docs/operations/delegated-stake-settlement-audit.md) specifies immutable snapshot evidence, effective/original references, ownership, reward/accrual history, prior payouts, anomaly resolution and exact-integer payout reconciliation. Unexplained groups block public activation. Synthetic tests establish algorithm behavior, not live-state provenance.

## Replay and activation

This is a replay/state-transition change, not a signed-schema, codec or hash-construction change. All new behavior uses `fields-added-ordinals.fixing-delegated-stake-double-withdrawal`. Dev remains `0`; Mainnet, Testnet and IntegrationNet mappings remain absent and resolve to `SnapshotOrdinal.MaxValue`.

Below activation, historical acceptance, raw reward inputs, arithmetic and unlock concatenation are preserved for replay. Historical unchecked sums are not reused on the activated settlement path; this is not a repository-wide arithmetic audit.

Public activation requires the completed state audit, review, one identical software/configuration version across Global L0s, a full-cluster cold restart and a strictly future ordinal. This prevents future duplicate settlements; it does not reverse historical balance effects or repair arbitrary malformed active stakes. Since the September 16 merge of #1597 the resolved gate is part of `deterministicConfigHash`, so Global L0s that disagree on its ordinal reject each other at join time; coordinated configuration remains mandatory.

## Verification

Validation for the September 10 follow-up:

- Full `nodeShared/test`: **1,449 passed**.
- Full `dagL0/test`: **251 passed**, two existing ignored generator tests.
- Full `dagL1/test`: **50 passed**.
- New real-distributor settlement suite: **17 passed**, included above.
- **1,750 passed overall**, two existing ignored generator tests.
- `scalafmtCheckAll`: passed.
- `scalafixAll --check --rules OrganizeImports`: passed.
- Actual-commit and complete-range commitlint (all three outgoing commits): passed.
- Whole-project `sbt compile` at exact commit `59b4aaa85`: passed (86 seconds); immediate post-build worktree status was clean.
- `git diff --check`: passed.

Coverage includes A-1/A/A+1; same- and cross-bucket duplicates; wrong-source highest-reward winners; checked distinct-lock sums; exact `Long.MaxValue`, negative-wrap and positive-wrap overflow; destination credit and issuance overflow; deterministic ties; schema-order/effective-ref collisions; later-cooldown cleanup; repeated settlement; natural-expiry boundaries; zero rewards; real MPT and fallback pending-create validators; and nonzero-reward R/R+1 replacement rollover. The new suite also exercises full `GlobalSnapshotAcceptanceManager.accept` with the real reward-distribution callback (other services use the existing test fixtures).

Local Docker E2E was not rerun; GitHub checks on the new head are tracked separately. No merge, deployment or public activation is part of this update.

September 16 merge of develop (`7bb1ef6f5`): the new gate is wired into #1597's `SnapshotOrdinalGate` accessor, `resolvedThresholdsFor`, `FieldsAddedOrdinalsFixtures` and the independent threshold table; dev-only gates are exempt from the Mainnet-removal hash test like `currency-snapshot-protocol-v1`. Validation on the merged head: node-shared config suites 135 passed, `dagL0/test` 276 passed (two existing ignored), `dagL1/test` 50 passed, full `nodeShared/test` green after the exemption, `scalafmtCheckAll` and `scalafixAll --check` passed, commitlint passed on the full range. The previously red Build JARs job was caused by `UpdateDelegatedStakeAcceptanceManager.make` gaining a required parameter that upstream Snapshot Streaming does not pass; the API compatibility patch is already absorbed upstream and skipped, so a dedicated apply-or-fail patch `docker/snapshot-streaming/snapshot-streaming-delegated-withdrawal.patch` now wires the resolved gate. Snapshot Streaming `release/testnet` compiles against the merged SDK with the patch chain applied in script order.


ryle-ai (commented):
The cross-branch situation is the thing I'd want resolved before this lands: this PR fixes the reward leg (`deduplicateExpiredWithdrawalsByTokenLock`), and #1592 fixes the principal leg on `release/mainnet` but leaves the reward leg reading raw `expiredWithdrawalsDelegatedStaking` at `GlobalDelegatedRewardsDistributor.scala:510-519`. `release/mainnet` is the branch carrying the legacy duplicate records that both PRs exist to handle, so the defense is landing on the branch that doesn't have the malformed state and not on the one that does. Left as-is, the merge order decides whether mainnet ever gets it. I've raised the same point on #1592; whichever way you go, the two should carry the same set of legs behind the same gate name so a future reader can't conclude from one branch that the other is covered.

On `deduplicateExpiredWithdrawalsByTokenLock` itself: the scaladoc justifies picking the greatest `rewards` with "legacy duplicate pending records can represent successive states of the same stake lineage and therefore carry overlapping cumulative rewards. The greatest reward is the most complete entitlement known for that lock." That's the load-bearing assumption of the whole function and it's asserted rather than shown. Max is also the direction that pays out more, so if the premise is wrong the failure mode is over-emission, not a stuck withdrawal. The duplicate records are in mainnet state and should be enumerable — can you say how many there are and whether their reward values actually nest the way the comment claims? If they do, say so in the comment (a count and a "verified against ordinal N" is enough). If some pair turns out to be non-nesting, min is the safer selector and the comment should explain why max was still chosen.

Second thing: there are three semantically distinct behaviour changes behind the single `fixing-delegated-stake-double-withdrawal` ordinal, and only two of them are described anywhere.

- Dedup of withdrawal and replacement unlocks by effective `tokenLockRef` (`TokenLockStateManager.scala:351-390`).
- Dedup of reward-bearing withdrawals by `tokenLockRef` (`DelegatedRewardsDistributor.scala:69-100`).
- Exclusion of naturally-expired locks from generated unlocks — the `Option.unless(activeTokenLock.unlockEpoch.exists(_ < epochProgress))` in the enforce branch.

The third one isn't dedup at all. The artifact leg was already protected against it by the `filterNot` on `tokenUnlocksEvents` at `GlobalSnapshotAcceptanceManager.scala:1279-1286`, but the balance leg wasn't — `generatedTokenUnlocks` also flows into `acceptTokenLocks` (`:995`) and `updateGlobalBalancesByTokenLocks` (`:1016`), where a naturally-expired lock could be credited by both paths. That's arguably the most consequential of the three and neither the config comment nor the `fields-added-ordinals.md` row mentions it; the docs entry says "duplicate effective token-lock withdrawals are rejected and legacy duplicate pending state is settled conservatively once," which reads as if only dedup happens. Please extend that sentence, since that table is what someone will read when they're deciding the mainnet activation ordinal and trying to work out the blast radius.

Smaller things:

`GlobalSnapshotConsensusFunctions.make` takes `fixingDelegatedStakeDoubleWithdrawalOrdinal: SnapshotOrdinal = SnapshotOrdinal.MaxValue`. The default is in the safe direction and `GlobalSnapshotConsensus.scala:266` passes the config-resolved value, so production wiring is right. But a default here means a call site that forgets the argument gets the dedup silently disabled with no compile error and no log — and the only signal it's off is the absence of a `[DELEG_STAKE_WITHDRAWAL_DEDUP]` line, which nothing looks for. Same shape on `TokenLockStateManager.generateTokenUnlocks`, where `enforceUniqueTokenLockRefs: Boolean = false` and `currentEpochProgress: Option[EpochProgress] = None` are both defaulted on the trait method. Dropping the defaults costs a couple of test call-site edits and makes the wiring impossible to get silently wrong; given the gate is protecting balance transitions, that trade seems worth it.

Relatedly, `currentEpochProgress.toRight("Current epoch progress is required when enforcing unique token lock unlocks")` turns a wiring mistake into a `SnapshotFailure.TokenUnlockGenerationFailed` at consensus time rather than at construction. If the two parameters were one `Option[(EpochProgress)]`-shaped mode argument — or just non-defaulted required params — that branch disappears.

Things I checked and don't think need changing, so nobody re-litigates them: the `groupBy`/`reduceLeft` in the dedup runs over a `HashMap` intermediate, but the result is normalized through `SortedSet.from` into a `SortedMap` and the token-lock path sorts via `sortBy(_.value)`, so no hash iteration order escapes into hashed state; the artifact set at `:1279` is a `SortedSet`. Keying the enforce branch on `activeTokenLock.source` rather than the withdrawal map key is the right call and matches #1592. Unlock amounts are unchanged — both the legacy and enforce branches take `activeTokenLock.amount`. And I initially thought regrouping by `_.event.source` instead of the lock's source could misattribute rewards; it can't, because `UpdateDelegatedStakeValidator.validateTokenLock` enforces `address === tokenLock.source` at Create (`:257`), so the two are equal for anything that reached state. The cross-source malformed case is covered by `DelegatedRewardsDistributorSuite:69,123`.

scasplte2:
@ryle-ai The implementation follow-up is now on `cecebadb5`, with all 15 checks passing.

- Both #1593 and #1592 carry the gated reward and principal protections. Develop preserves its effective-reference replacement rollover and missing-lock retention; Mainnet keeps its branch-specific orphan cleanup.
- Acceptance now prepares one settlement for rewards, principal and pending cleanup, including later-cooldown copies. Activated reward aggregation, issuance totals and balance credits use checked arithmetic.
- Natural-expiry suppression is documented in the configuration and activation guide, with tests on both sides of the strict `unlockEpoch < currentEpoch` boundary.
- The retained acceptance factories require the resolved gate, and unlock generation takes an explicit `Legacy | Unique(epoch, settlement)` mode. The old defaulted Boolean/optional-epoch combination is gone.

Your maximum-entitlement evidence question remains open. We have not completed the ordinal/hash-pinned duplicate census and historical-lineage audit, and are not claiming a verified count. Maximum is the policy for overlapping cumulative lineage, not evidence that arbitrary malformed records are entitled to that amount. Public activation mappings remain absent/disabled; Dev is explicitly active. The [settlement audit checklist](https://github.com/Constellation-Labs/tessellation/blob/cecebadb5e009353e285819f4e85d501b1278d84/docs/operations/delegated-stake-settlement-audit.md) makes unexplained lineage a blocker to public activation. I propose merging the disabled public-network implementation with that activation blocker retained.

The resolved gate is now included in the consensus configuration hash. The Snapshot Streaming build patch supplies its required parameter, and the current Build JARs and full E2E matrix pass. The real-distributor regressions cover A-1/A/A+1, cross-bucket ownership, overflow, natural expiry, and nonzero-reward R/R+1 replacement rollover.

Please re-review the current head, including whether the explicit separation between merge and public activation addresses the remaining audit concern.

ryle-ai (commented):
Re-read cecebadb (the PR-owned delta is the 28 files against the merge-base; the rest of the 2b4f826c range is develop merges including #1597's gate refactor). Of my earlier asks: the three-behaviours-under-one-gate point is addressed (natural-expiry suppression is now in the conf comment, `types.scala`, and the activation guide, tested on both sides of the strict boundary), the default args are gone (`UnlockMode` is explicit, factories require the resolved gate), and the `forProduct22` key order matches the case class. The max-selector audit remains open, and you've reframed it as an activation blocker rather than a merge blocker. With public mappings absent and the audit doc pinned as the gate, I'm fine with that framing for merge, so long as nobody adds a mainnet ordinal without the census and lineage evidence the doc describes.

Structure is what I asked for: `prepare` is the one settlement value, rewards read `rewardsByAddress`, principal reads `withdrawalsByRef`, cleanup uses `removeSettled`, and `GlobalSnapshotConsensusFunctions` no longer normalises independently. Effective-ref keying is consistent throughout (`currentTokenLockRef.getOrElse(event.tokenLockRef)`), and the replacement rollover holds: `dueRefs` requires the lock to be last-active, so NEW settles at R+1 only and `removeSettled` retires every copy, so no re-settlement after that.

Two things to decide, same as on #1592. The checked paths (`prepare`, `acceptRewardTxsChecked`, `sumMintedAmountChecked`) raise a bare `BalanceArithmeticError` through `fromEither`, a network-wide halt class. Unreachable on sane state, but one malformed reward record halts GL0 by design, and the token-lock-sum halt earlier this year had the same shape. If abort is the policy, say so in the PR body; and at least wrap it in a `SnapshotFailure` like `TokenUnlockGenerationFailed` at `:1019` so the error path is consistent.

Coverage note: the nonzero-reward R/R+1 case is tested through the settle helper (pays 80 once at R+1, empty at R+2) but not through full `accept`; the full-accept test covers only A-1/A/A+1 without replacements. Since the replacement deferral is the develop-specific behaviour the audit doc singles out, one full-accept replacement case would be worth having.

Low confidence, likely fine: `pendingWithdrawalTokenLockRefs` in the acceptance manager is built from the last context, so at R it still holds OLD while the pending record is being rewritten to NEW. A same-round Create on NEW passes that check; I believe `validateTokenLock` rejects it anyway because NEW isn't in the last-active set, so the new `AlreadyWithdrawn` check is belt and braces there. Didn't trace the mpt-store validator variant.

ryle-goehausen (approved):
Review is nonblocking

PR #1594, opened 2026-09-03
fix(node-shared): validate fee transactions individually and apply the ones that pass

effects-ai:
`validateFeeTxs` returned `Unit` and raised if any single fee transaction failed validation, so one entry decided the fate of the whole set. It now judges each transaction on its own and returns the ones that passed, and only that surviving set reaches `acceptFeeTxs` and the snapshot artifact. This mirrors how `applyFeeTransactions` already treats a transaction its source cannot cover, and it matches the equivalent site on release/mainnet.

The selection is deterministic: a verdict is a pure function of the transaction and the authorization flag, and the surviving set keeps the incoming SortedSet ordering, so every node computes the same subset.

Behavior below the activation ordinal is unchanged. Selecting per transaction changes which artifact a given set of events produces, and down there the data application layer applies its earlier rules, so a mixed fleet would disagree during a rolling upgrade.

The gate is `fixingDataApplicationFeeValidation`, the same boundary the data application layer uses for validateEveryFeeTransaction, so both layers agree on when selection applies. No new `FieldsAddedOrdinals` entry is needed.

### DIFF
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/currency/BalanceOpsManager.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/snapshot/managers/currency/BalanceOpsManager.scala
@@ -54,19 +54,54 @@ class BalanceOpsManager[F[_]: Async](
     (finalBalances, acceptedTxs).pure
   }
 
+  /** Validates the fee transactions carried by a data application block and returns the ones that passed.
+    *
+    * At or above the activation ordinal each transaction is judged on its own and the invalid ones are left out, so the rest of the set is
+    * still applied. This mirrors how `applyFeeTransactions` already treats a transaction its source cannot cover. The selection is
+    * deterministic: a verdict is a pure function of the transaction and the authorization flag, and the surviving set keeps the incoming
+    * `SortedSet` ordering, so every node computes the same subset.
+    *
+    * Below the activation ordinal the earlier all-or-nothing behaviour is kept. Selecting per transaction changes which artifact a given
+    * set of events produces, and down there the data application layer applies its earlier rules - it checks neither source != destination
+    * nor signature exclusivity - so a mixed fleet would disagree during a rolling upgrade.
+    *
+    * @return
+    *   the transactions that passed validation, in the incoming order
+    */
   def validateFeeTxs(
     maybeTxs: Option[SortedSet[Signed[FeeTransaction]]],
-    enforceWalletAuthorization: Boolean
-  ): F[Unit] =
-    NonEmptyList.fromList(maybeTxs.toList.flatMap(_.toList)).fold(().pure[F]) { nonEmptyTxs =>
-      feeTransactionValidator.validate(nonEmptyTxs, enforceWalletAuthorization).flatMap {
-        case Validated.Valid(_) =>
-          ().pure[F]
-        case Validated.Invalid(errors) =>
-          new Exception(s"FeeTransaction validation failed: ${errors.toList.mkString(", ")}")
-            .raiseError[F, Unit]
+    enforceWalletAuthorization: Boolean,
+    atOrAboveActivationOrdinal: Boolean
+  ): F[Option[SortedSet[Signed[FeeTransaction]]]] =
+    if (!atOrAboveActivationOrdinal)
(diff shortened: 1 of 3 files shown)


scasplte2:
I prepared a focused follow-up branch based directly on this PR head:

[`fix/pr1594-critical-fee-alignment`](https://github.com/Constellation-Labs/tessellation/tree/fix/pr1594-critical-fee-alignment)

It closes the cross-gate integrity gap where data-application acceptance could accept and combine an update carrying a valid co-signed fee, while final Currency acceptance—before `fee-transaction-security`—rejected and dropped that fee, leaving the update accepted without payment.

The branch:

- requires cryptographically valid, source-exclusive proofs before fee-security activation;
- preserves source-authorized co-signers after activation;
- keeps the existing three-argument SDK validator entry point compatible;
- schedules IntegrationNet `fixing-data-application-fee-validation` at `5926000`, while preserving its already-crossed `5880000` fee-security boundary for replay; and
- adds signer-policy and configuration regression coverage.

Commits: `1b308f31b`, `eafc93e3c`. Verification: 35 focused tests passed, Currency L1/L0 test compilation passed, and `git diff --check` is clean.

ryle-ai (changes requested):
The per-transaction selection itself is deterministic (verdict is pure per tx, survivors keep the incoming `SortedSet` order), so no complaint there. The blocker is the interaction with the layer above it, which James already spotted and has a branch for; I want to add why I think it has to land with this PR and not after.

Below `fee-transaction-security` (mainnet and testnet are at 9999999, so that's the live path), `enforceWalletAuthorization` is false and `BalanceOpsManager.validateFeeTxs` falls through to `isSignedExclusivelyBy(source)`: every proof must be the source wallet. The data-application layer that runs first (`shared/.../FeeTransactionValidator.validateAllFeeTransactions`) uses `FeeTransactionSignatureValidator.validate`, which only needs the source among the signers. Same at Currency L1 admission. So a fee tx signed by source plus one extra key is admitted, passes the data-app layer, its `DataUpdate` gets combined, and then node-shared rejects it. Before this PR that rejection raised and the round died, which was its own problem (anyone could stall a metagraph with a co-signed fee). After this PR it's silently dropped and the update stands with its fee never debited. The comment above `validateDifferentAddresses` in the shared validator describes exactly this class of gap for the src!=dst rule; the signature rule has the same shape. Switching from raise to drop is what turns this from a liveness bug into a free-update bug, so the alignment fix needs to be in the same release, not a follow-up.

Related: the PR body says this matches the equivalent site on release/mainnet. Structurally yes, but on release/mainnet the per-tx path calls `validate(signedTx, verifySignatures = true)`, whose `validateSourceWalletAuthorization` requires crypto verification and exclusivity, and mainnet's shared validator also requires exclusivity (`forall(_ === tx.source)`). So on mainnet the two layers agree and on develop after this PR they don't. A co-signed fee tx is accepted on develop and rejected on the current mainnet binary, so a develop-based release would diverge from mainnet on that input. Worth stating in the description so nobody reads "matches mainnet" as "same accepted set".

One question rather than an ask: `fixing-data-application-fee-validation` mainnet is 6818000 and mainnet is well past that today, so the config comment ("never at an ordinal already present in signed history") and the rolling-upgrade rationale in the scaladoc don't really describe this activation; the drop path is live from the first upgraded node. I believe replay is still safe because below the gate the raise produced no artifact, so no signed snapshot can contain a set the per-tx path would prune. If that reasoning is what you're relying on, it belongs in the doc comment instead of the mixed-fleet story.

scasplte2:
@ryle-ai I've addressed your review in `4ed8ec6ed`. The signer-policy alignment (`1b308f31b`) is now in this PR, so before `fee-transaction-security` the data-app layer requires source exclusivity, the same as node-shared. A co-signed fee is now rejected before its update is combined; before this change the update was combined and the fee silently dropped. `FeeTransactionLayerAlignmentSuite` uses real keys to assert that any fee accepted by the data-app layer is also kept and debited by final acceptance, under both signer policies. I did not take the IntegrationNet `5926000` schedule, because IntegrationNet is already past that ordinal. The scaladoc and the PR description now use your replay argument, and the "matches mainnet" claim is corrected. The same-snapshot balance gap is listed as a separate limitation.

ryle-goehausen (approved):

PR #1595, opened 2026-09-04
test: strengthen delegated withdrawal regression coverage

scasplte2:
## Summary

Adds concurrent delegated-stake replacement/withdrawal submission and settlement coverage on the Mainnet safety fix in #1592. The E2E submits through two Global L0 nodes, checks that each observed state has at most one active or pending record per token lock, and verifies final principal settlement.

Invariant violations now fail immediately instead of being retried until a later safe observation. The checks run during outcome polling, pending-withdrawal polling, and settlement. The settlement check runs alongside the independent staking flow and both results are collected even if either fails. Request hashes, the submission-window ordinal, observed stake records, and endpoint failures are logged; the CI port convention is documented.

## Coverage contract

- Concurrent HTTP submission does **not** prove that both events were considered in one consensus round. The E2E names, comments, and output make that limit explicit.
- `UpdateDelegatedStakeAcceptanceManagerSuite` supplies deterministic same-batch A-1/A coverage: both requests are accepted before activation; the replacement is accepted and the original withdrawal is rejected at activation. Comparing two accepted ordinals would assert the wrong outcome.
- `GlobalSnapshotDelegatedStakeUnlockSuite` now also splits duplicate legacy withdrawals across two address buckets at A-1/A/A+1. The transition fixture checks one owner credit, an unchanged other-address balance, cleared lock/pending state, and one unlock artifact. This malformed state is constructed in Scala rather than through the public API.
- Six Node regression cases protect immediate invariant failure and retryable propagation/settlement behavior; they run in the delegated-staking workflow.

## Stacking

This tests-only PR remains based on `fix/mainnet-delegated-withdrawal-safety` (#1592). Production code and activation configuration are unchanged. After #1592 merges, retarget to `release/mainnet` and refresh CI.

## Verification

- Node 18: six invariant regression cases passed; JavaScript syntax checked.
- Focused Scala suites: 15 tests passed (acceptance manager, unlock transition, ordinal configuration).
- Six additional isolated probes through the actual E2E functions passed, covering successful settlement, immediate failure in all three polling phases, and collection of settlement results when the other flow fails.
- Full-project `sbt compile` passed at `62563135105e22e8ed2b8ea3a897ff874493ae4c` with a clean worktree afterward.
- Workflow YAML, `git diff --check`, PR title, and all commit messages in the PR range passed validation.

The live network E2E remains a GitHub CI gate; the isolated probes are not a network run.

### DIFF
--- a/.github/action_scripts/delegated_staking/delegated-staking.js
+++ b/.github/action_scripts/delegated_staking/delegated-staking.js
@@ -17,6 +17,7 @@ const {
   parseSharedArgs,
   PRIVATE_KEYS,
   withRetry,
+  createAndConnectAccount,
   createNetworkConfig,
   logWorkflow,
 } = require('../shared')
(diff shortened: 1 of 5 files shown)


ryle-ai (commented):
The thing that worries me about this test is that a green run doesn't tell you the race happened. `waitForNextSnapshot` gets you to a fresh round boundary and then `Promise.allSettled` fires the replacement and the withdrawal at two different gL0s, but nothing afterward checks that they were actually accepted in the same consensus round. If they land in adjacent rounds — which is the likely outcome most of the time, given two HTTP submissions plus L1 block inclusion against a 43s time trigger — you're testing the sequential path, which was already safe, and every assertion below passes. So the failure mode is that this goes green forever while covering nothing, and there's no signal distinguishing that from a real pass.

What I'd want: after both submissions are fulfilled, resolve the ordinal each one was accepted at and assert they're equal, or retry the whole staging step if they aren't. Then a pass means something, and a persistent inability to co-locate them is itself reportable rather than silently degrading into a no-op. `assertDoubleWithdrawalRaceIsSafe` already tolerates "replacement active XOR original pending" as terminal, which is the correct invariant — it's just that both branches are also reachable without any race.

`getAlternateGlobalL0Url` deriving the second node by `port + 10` is an assumption about the compose topology encoded as arithmetic in a helper with no comment saying where the +10 comes from. It does fail loudly if wrong (the `allSettled` status check catches the resulting submission failure), so this isn't a correctness issue, but it'll be the first thing to break when someone renumbers the local cluster and the error won't point here. Reading both URLs from the same place the workflow gets `90 91` would be more durable.

Second structural point, and this one is about scope rather than this file: the branch this test can't reach is the one that most needs coverage. #1592's post-activation branch exists for duplicate pending withdrawal records that share a `tokenLockRef` across two address keys. That state can't be produced by any valid sequence of API calls — `UpdateDelegatedStakeValidator` won't let you create it — so no e2e can construct it, and the replacement/withdrawal race this test does cover is a different scenario. Worth saying explicitly in the PR that this is coverage for the race, not for the legacy-state branch, so nobody reads a green delegated-staking workflow as evidence the malformed-state handling works.

Related: dev has `fixing-delegated-stake-double-withdrawal: 0`, so this exercises only the post-activation path. For an ordinal-gated change to acceptance and balance transitions I'd normally want the crossing itself tested — old behavior below, new above, same fixture — and that has to be a unit test, not this.

Minor: `beginDoubleWithdrawalProtectionTest` now runs before `testCreateDelegatedStake` and its returned closure runs after everything else, so the final principal-credited-once assertion is gated on the entire rest of the suite passing first. Any unrelated failure in between means you lose the result of this test rather than getting a partial answer. Since it uses `key3` and the main flow uses `key4`, there's no state collision — it's just that the split-phase shape costs you the signal when something else breaks.

scasplte2:
@ryle-ai Addressed in 62563135105e22e8ed2b8ea3a897ff874493ae4c.

Your main coverage concern is valid. The E2E is now explicitly concurrent-submission and settlement coverage: the names, comments, logs, and PR description no longer present a green run as proof that both requests reached the same consensus batch. It records the request hashes, submission-window ordinal, and observed stake records.

I did not add equality of accepted ordinals because the correct same-batch result after activation is **replacement accepted, original withdrawal rejected**. The deterministic fixture already in #1592's `UpdateDelegatedStakeAcceptanceManagerSuite` supplies identical context/events at A-1 and A and asserts that change. The E2E's HTTP acknowledgments cannot establish it.

I also fixed a related false-positive path: the state assertions could observe an active and pending record for the same lock, retry, then pass on a later safe response. Such violations now fail immediately during outcome, pending-withdrawal, and settlement polling. Six Node regression cases cover immediate failure despite a later safe response, ordinary propagation/settlement delays, and retry exhaustion; the workflow runs them.

For the legacy-state concern, `GlobalSnapshotDelegatedStakeUnlockSuite` now covers the same lock's duplicate withdrawals split across two address buckets at A-1/A/A+1. Its transition fixture verifies one credit to the active lock's owner, no change to the other address's balance, removal of the lock and pending records, and one unlock artifact. The existing single-bucket duplicate coverage remains in the unlock-generation test. These are Scala fixtures; the API-driven E2E does not claim to create malformed historical state.

The +10 port convention is documented against the actual CI template, and submission errors identify the endpoints and causes. Settlement now runs alongside the independent staking flow, with both results collected even if either fails, preserving the cooldown overlap.

Validation: six Node 18 regression cases, 15 focused Scala tests, six isolated E2E-function probes, full-project compile at the pushed commit, workflow/syntax/diff checks, and complete-range commitlint all passed. The full network E2E remains a GitHub CI gate.

ryle-goehausen (approved):

PR #1596, opened 2026-09-04
fix(dag-l0): harden downloaded snapshot head handoff

scasplte2:
## Summary

This follow-on closes snapshot-storage and lifecycle gaps exposed while qualifying v35 recovery against real assembled jars:

- rebuild rollback Last-N from the local canonical archive by child-declared hash, with exact peer fallback only when local bytes are absent;
- publish downloaded snapshot values and semantically equal contexts durably before exposing the public head;
- classify target collisions, snapshot-index readback failures, and context readback failures with structured logs and bounded metrics;
- reconcile target, future, and lower positive caches against caller-validated persisted history while preserving legitimate pending or hash-durable lower entries;
- publish the full-download, follower-catch-up, or recovery terminal pair before consensus initialization;
- retain ordinal-1 snapshot context only for development active-from-genesis certified replay; scheduled non-genesis and dormant behavior remain unchanged;
- add a production-shaped rollback-lead/forward-download Just E2E; configure GL0 E2E quorum at 2/3 and exercise `dag-cluster` with four GL0 nodes.

## Consensus compatibility

Classification: runtime/storage-publication behavior plus test-only E2E configuration.

- No signed schema, codec, hash/signature construction, or persisted-format change.
- No consensus value, production quorum, committee, trigger, or leader-selection change.
- No activation ordinal or recovery-authority change.
- No Currency L0 consensus or L1 production change.
- No automatic rollback, conflicting-target overwrite, arbitrary-LCA repair, lock clearing, or peer-majority history authority.
- Production DAG L0 remains unchanged at 2/3 quorum. GL0 CI now also uses a 2/3 fallback; `dag-cluster` runs four GL0 nodes, `rollback-download-head` retains its explicit three-node/2/3 override, and `committee-rewards` retains five nodes. Currency L0 keeps its separate quorum setting.

The operated scenario matches the permissioned deployment topology: one `run-rollback` lead starts first and every other node starts as `run-validator`.

## Review resolution

- A lower head may publish only when its target ordinal is absent or already contains the same canonical snapshot value/hash after the caller validates the target and persisted ancestry.
- A conflicting target fails closed as `SnapshotOrdinalCollision`; Global download does not gain suffix-overwrite authority.
- Exact target publication performs one write attempt followed by one proof across both snapshot indexes. A failed proof retains the original write failure as a suppressed cause.
- Context readback uses content-aware schema equality and is described as semantic equality rather than proof-envelope or byte equality.
- Lower caches use the ordinal-selected hasher and reconcile against persisted ordinal values, pending writes, and readable hash-index evidence before head publication. Unsupported entries and stale retry markers are removed.
- Publication failures emit `dag_download_head_publication_failure_total` with `path`, `direction`, and `reason`. Success remains post-durable and before consensus initialization.
- The shared `LocalFileSystemStorage.link` behavior is unchanged; the publication boundary independently proves both target indexes and fails with a typed error when either is unavailable.
- The E2E wording now states that it proves rollback-lead initialization and forward full download. It does not claim to exercise backward publication.

## Review guide

1. `144c910eb`: hash-bound local-first rollback window reconstruction.
2. `cdb0bfa26`: serialized durable head publication and cache reconciliation.
3. `bbf4d5844`: publication-before-consensus lifecycle ordering.
4. `e56c81fc6`: active-from-genesis test root retention.
5. `b5a5cb59b` and `0a8c98df5`: assembled-jar E2E and CI registration.
6. `43f9087a6`: content-aware persisted-context equality for array-bearing metagraph state.
7. `130af5239` and `2578f338c`: publication/hash/signature assertions.
8. `77dddbe85`: review corrections for publication failures, cache reconciliation, metrics, documentation, and CI quorum propagation.
9. `eefed9d03`: use ordinary comments for local helpers so SDK Scaladoc generation succeeds.
10. `16b0a019e`: use the GL0 2/3 E2E fallback and four GL0 nodes for `dag-cluster`, preserving all other matrix topologies.

## Verification

- `SnapshotStorageSuite`: 31 passed.
- `DownloadSuite`: 23 passed.
- `DownloadHeadPublicationSuite`: 2 passed.
- `SnapshotLocalFileSystemStorageSuite`: 25 passed.
- Focused review total: 81 passed.
- Whole-project `sbt compile` passed on `eefed9d03` after the branch rebase onto current `develop`.
- `sdk/doc` passed after correcting the local helper comments that caused the first post-update JAR build to fail.
- `scalafmtCheckAll` passed.
- `scalafixAll --check --rules OrganizeImports` passed.
- Commitlint passed for `eefed9d03` and the complete rebased PR commit range.
- Shell syntax, workflow/Compose YAML parsing, rendered quorum assertions, and `git diff --check` passed.

### CI configuration follow-up (`16b0a019e`)

The previously missing Compose pass-through made unanimity effective for ordinary CI rows. That setting prevents current seated signers from satisfying the next larger committee's finality floor during ordinary certified admission. The user-approved follow-up uses the existing GL0 2/3 policy instead of changing admission authority or tiering.

- Only `dag-cluster` gains `--num-gl0=4`; its existing checker requires three snapshot proofs. Other matrix rows and test assertion code are unchanged.
- Whole-project `sbt compile` passed at `16b0a019e`.
- Commitlint passed for the new commit and the complete PR range.
- Workflow and five Compose files passed YAML parsing.
- All 12 matrix rows rendered GL0 quorum `0.6666666666666666`; all four `dag-cluster` node configurations rendered with distinct names and public ports.
- Rendered GL1/ML0/CL1/DL1 received no quorum override; Currency L0's committed default remains `1.0`.
- The Just dry-run confirmed that the four-node override follows the default. Normalized workflow comparison confirmed only the fallback and `dag-cluster` node count changed.
- Runner, environment-parser and rollback shell syntax, plus `git diff --check`, passed.
- Claude Fable completed two focused read-only review passes. Its initial checker-call-site and environment-scoping concerns were resolved with runner/Compose evidence; its remaining comparator check was verified against the actual `signerCount >= expectedSigners` expression (two rejected, three and four accepted). This was not a full-protocol or live-E2E review.
- Provisioning four processes and observing a three-proof snapshot do not independently establish four admitted signing seats or Byzantine fault tolerance. No new identity-based participation assertion is claimed.
- Fresh full-matrix CI, including `dag-cluster`, `committee-rewards`, and rollback/rejoin, remains required before merge.

### Real-process qualification

Before the final review-correction commit, three consecutive runs passed on the same assembled GL0 jar:

- Git application head: `f592fa92e6477214167d1db40b953bc4d6129da5`
- GL0 jar SHA-256: `40e4313549b5ef4603330f0755e364fe452bb58ec0de793ef73eba52a84e7b6f`

Each run proved rollback-lead initialization with validators stopped, restoration of production by the healthy pair, forward full download by the lagged validator, target hash agreement, later participation by the restored validator, cluster convergence, and continued production.

A fourth run first populated GL0 with the standard three-node test metagraph and a signed Currency snapshot carrying non-empty byte-array application state, then repeated the lifecycle successfully.

The `77dddbe85` cache/observability correction and `eefed9d03` documentation-build fix have source-level and focused regression validation. The [E2E matrix at `eefed9d03`](https://github.com/Constellation-Labs/tessellation/actions/runs/34396900061) finished with 10 passing rows and failures in `dag-cluster` and `committee-rewards`. The `16b0a019e` workflow follow-up requires a fresh complete E2E matrix; no new local live-process run or full-matrix pass is claimed.

## Deliberate limits

This PR hardens existing rollback/download handoff. It does not implement backward conflicting-target replacement or arbitrary historical-LCA fork repair. Those remain separate protocol work.

### DIFF
--- a/.github/workflows/e2e-just-test.yml
+++ b/.github/workflows/e2e-just-test.yml
@@ -111,6 +111,13 @@ jobs:
           # 30 min default would have killed it regardless).
           # The previous note here blocked on PR #1485, which was closed without ever merging.
           # - { name: fork-recovery,          tests: "fork-recovery", extra-args: "--num-gl0=5", runner: "Ubuntu-22-64-core", timeout: 60 }
+          # Exercises the production cold-restart topology: one rollback lead, one warm validator,
+          # and one validator restored from an older same-prefix snapshot. The healthy pair must
+          # advance, the lagged validator must publish its exact downloaded head before consensus
+          # initialization, become Ready, sign again, and converge on the same post-recovery hash.
+          # This test needs 2/3 finality while one member is deliberately offline; keep that quorum
+          # override scoped to this matrix row rather than weakening the other E2E jobs.
+          - { name: rollback-download-head,  tests: "rollback-download-head", extra-args: "--num-gl0=3 --num-gl1=0", quorum: "0.6666666666666666", timeout: 60 }
           - { name: currency,               tests: "currency" }
           - { name: rewards,                tests: "rewards" }
           - { name: token-locks,            tests: "token-locks" }
@@ -141,7 +148,7 @@ jobs:
       # Cadence when there ARE pending events (the flood): lowered so gl0 drains the chain faster without
       # spawning empty rounds when idle.
       CL_EVENT_TRIGGER_COOLDOWN: "5 seconds"
-      CL_QUORUM_THRESHOLD_FRACTION: "1.0"
+      CL_QUORUM_THRESHOLD_FRACTION: ${{ matrix.test-group.quorum || '1.0' }}
 
     steps:
       - name: Checkout code
@@ -243,7 +250,7 @@ jobs:
           docker logs "snapshot-streaming-postgres" > "runner-logs/snapshot-streaming-postgres.log" 2>&1 || echo "No snapshot-streaming-postgres container"
 
           # Collect node log directories (mounted volumes). Covers indices beyond the default 3
-          # so multi-node rigs (committee-rewards, fork-recovery) upload their late joiners' logs,
+          # so multi-node rigs (committee-rewards, fork-recovery, rollback-download-head) upload their late joiners' logs,
           # which are exactly the ones needed to diagnose an admission-timing failure.
           for i in 0 1 2 3 4 5 6 7 8 9; do
             cp -r "nodes/$i/gl0-logs" "runner-logs/node-$i-gl0-logs" 2>/dev/null || true
--- a/.github/workflows/e2e-just-test.yml
+++ b/.github/workflows/e2e-just-test.yml
@@ -115,9 +115,9 @@ jobs:
           # and one validator restored from an older same-prefix snapshot. The healthy pair must
           # advance, the lagged validator must publish its exact downloaded head before consensus
           # initialization, become Ready, sign again, and converge on the same post-recovery hash.
-          # This test needs 2/3 finality while one member is deliberately offline; keep that quorum
-          # override scoped to this matrix row rather than weakening the other E2E jobs.
-          - { name: rollback-download-head,  tests: "rollback-download-head", extra-args: "--num-gl0=3 --num-gl1=0", quorum: "0.6666666666666666", timeout: 60 }
+          # The committed DAG-L0 development config uses 2/3 finality, which this scenario verifies
+          # by requiring the two healthy members to advance while the third is deliberately offline.
+          - { name: rollback-download-head,  tests: "rollback-download-head", extra-args: "--num-gl0=3 --num-gl1=0", timeout: 60 }
           - { name: currency,               tests: "currency" }
           - { name: rewards,                tests: "rewards" }
           - { name: token-locks,            tests: "token-locks" }
@@ -148,7 +148,6 @@ jobs:
       # Cadence when there ARE pending events (the flood): lowered so gl0 drains the chain faster without
       # spawning empty rounds when idle.
       CL_EVENT_TRIGGER_COOLDOWN: "5 seconds"
-      CL_QUORUM_THRESHOLD_FRACTION: ${{ matrix.test-group.quorum || '1.0' }}
 
     steps:
       - name: Checkout code
--- a/.github/workflows/e2e-just-test.yml
+++ b/.github/workflows/e2e-just-test.yml
@@ -95,7 +95,7 @@ jobs:
       fail-fast: false
       matrix:
         test-group:
-          - { name: dag-cluster,            tests: "dag-cluster" }
+          - { name: dag-cluster,            tests: "dag-cluster", extra-args: "--num-gl0=4" }
           - { name: delegated-staking,      tests: "delegated-staking" }
           - { name: token-lock-replacement, tests: "token-lock-replacement" }
           # DISABLED on a NODE bug, not a test bug. The Phase-1 gate was unsatisfiable and is now
@@ -115,8 +115,7 @@ jobs:
           # and one validator restored from an older same-prefix snapshot. The healthy pair must
           # advance, the lagged validator must publish its target ordinal/hash before consensus
           # initialization, become Ready, sign again, and converge on the same post-recovery hash.
-          # Restore the job-wide unanimous E2E baseline through the env fallback below. This test alone
-          # needs a per-row 2/3 override while one member is deliberately offline.
+          # Keep this scenario's 2/3 requirement explicit while one member is deliberately offline.
           - { name: rollback-download-head,  tests: "rollback-download-head", extra-args: "--num-gl0=3 --num-gl1=0", quorum: "0.6666666666666666", timeout: 60 }
           - { name: currency,               tests: "currency" }
           - { name: rewards,                tests: "rewards" }
@@ -148,7 +147,9 @@ jobs:
       # Cadence when there ARE pending events (the flood): lowered so gl0 drains the chain faster without
       # spawning empty rounds when idle.
       CL_EVENT_TRIGGER_COOLDOWN: "5 seconds"
-      CL_QUORUM_THRESHOLD_FRACTION: ${{ matrix.test-group.quorum || '1.0' }}
+      # GL0 uses 2/3 so current signers can support certified committee expansion.
+      # This pass-through is GL0-only; Currency L0 retains its own quorum setting.
+      CL_QUORUM_THRESHOLD_FRACTION: ${{ matrix.test-group.quorum || '0.6666666666666666' }}
 
     steps:
       - name: Checkout code
--- a/.github/workflows/e2e-just-test.yml
+++ b/.github/workflows/e2e-just-test.yml
@@ -113,11 +113,11 @@ jobs:
           # - { name: fork-recovery,          tests: "fork-recovery", extra-args: "--num-gl0=5", runner: "Ubuntu-22-64-core", timeout: 60 }
           # Exercises the production cold-restart topology: one rollback lead, one warm validator,
           # and one validator restored from an older same-prefix snapshot. The healthy pair must
-          # advance, the lagged validator must publish its exact downloaded head before consensus
+          # advance, the lagged validator must publish its target ordinal/hash before consensus
           # initialization, become Ready, sign again, and converge on the same post-recovery hash.
-          # The committed DAG-L0 development config uses 2/3 finality, which this scenario verifies
-          # by requiring the two healthy members to advance while the third is deliberately offline.
-          - { name: rollback-download-head,  tests: "rollback-download-head", extra-args: "--num-gl0=3 --num-gl1=0", timeout: 60 }
+          # Restore the job-wide unanimous E2E baseline through the env fallback below. This test alone
+          # needs a per-row 2/3 override while one member is deliberately offline.
+          - { name: rollback-download-head,  tests: "rollback-download-head", extra-args: "--num-gl0=3 --num-gl1=0", quorum: "0.6666666666666666", timeout: 60 }
           - { name: currency,               tests: "currency" }
           - { name: rewards,                tests: "rewards" }
           - { name: token-locks,            tests: "token-locks" }
@@ -148,6 +148,7 @@ jobs:
       # Cadence when there ARE pending events (the flood): lowered so gl0 drains the chain faster without
       # spawning empty rounds when idle.
       CL_EVENT_TRIGGER_COOLDOWN: "5 seconds"
+      CL_QUORUM_THRESHOLD_FRACTION: ${{ matrix.test-group.quorum || '1.0' }}
 
     steps:
       - name: Checkout code
--- a/docker/bin/compose-runner.sh
+++ b/docker/bin/compose-runner.sh
@@ -51,6 +51,7 @@ if [ "$LIST_TESTS" = "true" ]; then
   echo "  dag-cluster              DAG cluster check"
   echo "  delegated-staking        Delegated staking tests"
   echo "  fork-recovery            Fork recovery test (needs --num-gl0=5)"
+  echo "  rollback-download-head   Rollback-lead-first cold restart and lagged-validator download"
   echo "  committee-rewards        Full-committee delegated reward split"
   echo "                           (needs --num-gl0=5 --num-gl0-early=3;"
   echo "                            --gl0-late-delay=<seconds>, default 240, tunes the join stagger)"
@@ -583,6 +584,15 @@ if should_run_test "fork-recovery"; then
   show_time "Fork recovery test completed"
 fi
 
+if should_run_test "rollback-download-head"; then
+  echo "================================================"
+  echo "Running rollback-download-head test"
+  echo "================================================"
+  cd $PROJECT_ROOT
+  bash docker/bin/test-rollback-download-head.sh $DAG_L0_PORT_PREFIX
+  show_time "Rollback download-head test completed"
+fi
+
 if should_run_test "committee-rewards"; then
   echo "================================================"
   echo "Running committee rewards test"
@@ -773,4 +783,3 @@ echo "End-to-end tests completed"
 echo "------------------------------------------------"
 
 cd $PROJECT_ROOT
-
--- a/docker/bin/test-rollback-download-head.sh
+++ b/docker/bin/test-rollback-download-head.sh
@@ -74,10 +74,6 @@ get_ordinal() {
     jq -r '.value.ordinal // empty' 2>/dev/null || true
 }
 
-get_metadata() {
-  curl -fsS --max-time 5 "$(node_url "$1")/global-snapshots/latest/metadata" 2>/dev/null || true
-}
-
 get_snapshot_hash() {
   local node=$1
   local ordinal=$2
@@ -300,7 +296,7 @@ wait_target_rejoin() {
   local since=$2
   local timeout=$3
   local deadline=$(( $(date +%s) + timeout ))
-  local state ordinal metadata metadata_ordinal metadata_hash artifact_hash logs metric
+  local state ordinal logs metric publication_ordinals publication_ordinal init_ordinals init_ordinal target_hash lead_hash
 
   while [ "$(date +%s)" -lt "$deadline" ]; do
     state=$(get_node_state "$TARGET_NODE")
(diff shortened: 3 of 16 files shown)


ryle-ai (commented):
The main thing I want to work through is `reconcilePositiveCaches`, because it's one-sided and I don't think the PR says why that's sufficient.

It clears cached entries at ordinals strictly above the new head, plus the same-ordinal-different-hash entry. It leaves everything below the new head alone. That's correct if the published canonical head shares every ancestor with what was locally cached. But the case the PR explicitly says it supports — "full download may legitimately replace a higher local fork head with a lower canonical head" — doesn't guarantee that. If the fork point is at ordinal K, the local branch ran K..N, and the canonical head publishes at M with K < M < N, then the entries in [K, M) still hold the abandoned branch's snapshots and survive reconciliation. A subsequent `get(ordinal)` in that range returns a wrong-branch artifact from cache.

I think the reason this doesn't bite in practice is the other half of the PR: switching `LastNGlobalSnapshotStorage` from `globalSnapshotStorage.get(snapshotOrdinal)` to `get(expectedHash)` means that consumer asks by hash and a stale same-ordinal entry can neither be accepted nor block the peer fallback. That's a good change and the comment on it is the clearest one in the diff. But it protects one consumer by construction, not the cache. If there's another `get(ordinal)` caller that can run against sub-head ordinals after a backward publication, it's still exposed. Either reconcile from the fork point rather than from the head, or say in the scaladoc that sub-head cache entries are assumed ancestor-identical and why — right now a reader has to reconstruct the argument from two files.

Second, the exact disk readback. `raiseUnless(persistedState.exists(_ === state))` makes head publication depend on the encode/decode round-trip being bit-exact through the full nested `GlobalSnapshotInfo`, including `lastCurrencySnapshots` and therefore `DataApplicationPart`'s byte arrays. I checked the `Eq` claim and it holds — `schema/package.scala:125` defines a content-comparing `Eq[Array[A]]`, so the comment about reference equality is accurate and the guard does what it says. What I can't tell from the diff is the operational behavior when it fires. It raises inside `download`, which DownloadDaemon retries, so a deterministic round-trip asymmetry in some field — `DataApplicationPart.updateHashes` having a Scala default is the kind of thing I'd want to have checked — turns into an indefinite retry loop that looks exactly like a network problem in the logs. The check itself is right; the ask is a distinct log line or metric for the readback-mismatch case specifically, so an operator can tell "my disk disagrees with my memory" apart from "I can't reach peers." Same for `"Snapshot is not durably readable from both indexes"`.

`StoragesInitializer` flipping `none` to `globalSnapshotStorage.asRight.some` is the change with the widest blast radius and the thinnest justification in the diff. The comment explains the intent well. What convinced me it's safe is that `get(hash)` falls through to `snapshotLocalFileSystemStorage.read(hash)` (`SnapshotStorage.scala:325-329`), so a cold-started lead with empty caches still resolves from its archive rather than silently degrading to peer-only — which would have made the whole change a no-op in the exact coordinated-cold-restart scenario it cites. Worth putting that sentence in the comment, since "consult peers only for an exact local hash miss" reads as depending on a warm cache.

Two smaller things.

`headPublicationDirection` returns `String` and the caller does `direction === "backward"`. It needs a string for the metric label, but the control-flow comparison against a literal is the kind of thing that survives a rename of one of the two occurrences. A small ADT with a `label` method costs three lines and makes that impossible.

`publishValidatedHeadBeforeConsensus` is `publish >> afterPublication >> startConsensus` behind five type parameters and a six-line scaladoc explaining that it isn't a new authority boundary. It has one real call site. The doc comment is doing the work the abstraction claims to do, and inlining the three statements in `handoffValidatedDownloadHead` would read better than the indirection — the ordering constraint is more visible as three lines in the caller than as an argument list. Not worth blocking on, but the fact that it needs a paragraph defending its own existence is usually the signal.

Checked and fine, so nobody re-derives it: `MapRef.ofSingleImmutableMap` looks unbounded but `offloadProcess` sets both cache entries to `none` on offload, so the full key scan in `reconcilePositiveCaches` is bounded by `inMemoryCapacity` and runs once per publication. The `requireExactSnapshotPersistence` path raising after the already-on-disk branch logs info is right, since `proveExactSnapshotPersistence` re-checks independently. `notPersistedCache.update(_.filter(_ < snapshot.ordinal))` correctly forgets un-persisted work at and above the abandoned head.

I have not reviewed `test-rollback-download-head.sh` in any depth — 520 lines of new shell in the e2e path is more than I can evaluate against a cluster I can't run. If that script is what gives confidence in the backward-publication case, someone who's run it locally should confirm it actually reaches the lower-canonical-head branch rather than only the forward path.

ryle-ai:
Follow-up: the doc at `Download.scala:538-539` and the code disagree, and your own test picks the code's side.

> Full download may legitimately replace a higher local fork head with a lower canonical head after its validated cleanup. That move remains allowed and is made explicit through warning/metrics rather than blocked by a monotonicity guard.

The `"backward"` direction isn't blocked by a monotonicity guard, agreed, but it is blocked one layer down whenever the local fork's snapshot at the target ordinal differs in bytes from the canonical one. `setHeadForRecovery` calls `enqueueUnlocked(..., requireExactSnapshotPersistence = true)`, which goes to `SnapshotLocalFileSystemStorage.write` — and that raises `UnableToPersistSnapshot` content-blind on `ordinalExists` (`:53-56`). The handler at `SnapshotStorage.scala:202-208` falls back to `snapshotExists(snapshot)`, which is false for differing bytes, and with the exact flag set it re-raises. `SnapshotStorageSuite.scala:504` ("validated head publication rejects a different value occupying the target ordinal without changing the head") asserts exactly that, `publication.isLeft`.

What makes me want the sentence changed rather than just softened is that there's no repair path once you're in it. The failure propagates out of `download`, `:632-638` reverts to `WaitingForDownload`, DownloadDaemon retries, and `performInitialCleanup` (`:1249-1260`) calls `cleanupAbove(metadata.ordinal)` — strictly above. The colliding file *is* `metadata.ordinal`, so every retry rebuilds the same chain and dies at the same write. Absorbing state.

I want to be clear about how narrow the reachability is, because I couldn't close it. On the ordinary full path the chain lands through `persistLastSnapshot` → `moveTmpToPersisted` → `replaceForRecovery` (`SnapshotDownloadStorage.scala:243-247`), which does overwrite, and a mid-chain conflict raises earlier at `:1439`. So the case needs the target ordinal to be occupied by different bytes at handoff time without having been rewritten on the way — which I think is possible for a fork that diverged at or below the network tip, but I did not construct it.

Either way one of the two should move. Route the backward case through `replaceCanonicalSuffixForRecovery` (or an exact variant that overwrites at the target ordinal), or drop "That move remains allowed" and say instead that a backward move is allowed only when the target ordinal is free or already holds the canonical bytes. The second is cheap and stops the doc from promising something the test forbids.

Second, smaller thing on the same guard: `LocalFileSystemStorage.link` (`:104-115`) swallows `NoSuchFileException` with a warning and `FileAlreadyExistsException` with a debug, so `write` can return successfully with the hash file present and the ordinal hardlink missing. `proveExactSnapshotPersistence` then demands both indexes, so that torn state turns a degraded-but-live node into one that can never finish downloading, and the only trace is a warn line from `link`. Given the new guard depends on both indexes agreeing, `link` failing is no longer a cosmetic condition and shouldn't be logged like one.

ryle-ai:
Second pass. The main thing I found is that the fix for what I raised last time is already sitting in this PR — it's just not wired to the global path.

**The global path gets the weak primitive; currency-l0 gets the strong one.** `SnapshotStorage.scala:432` overrides `replaceCanonicalSuffixForRecovery` to run `cleanupSuffix >> replaceExactRecoveryHead` inside the persistence mutex, and `replaceExactRecoveryHead` goes through `replaceForRecovery` on all three stores — an overwrite, not a write. That is exactly the primitive that makes a backward move at an occupied ordinal work. Currency L0 uses it and fails closed on the result:

```scala
exactInstalled <- ExactSnapshotStorage.installCanonicalSuffixForRecovery(...)
_ <- Async[F].raiseUnless(exactInstalled)(new IllegalStateException(s"Failed to install exact Currency artifact/context ordinal=...; keeping consensus stopped"))
```
(`currency-l0/.../programs/Download.scala:245-256`, and the same shape in `Rollback.scala:113`.)

Global L0 calls bare `setHeadForRecovery` at `Download.scala:562`, `:1044`, `:1104` and `GlobalSnapshotConsensusStateAdvancer.scala:242`. No cleanup inside the critical section, no exact overwrite, and no post-install head verification — the return type is `F[Unit]`, so nothing checks that what got published is what was asked for. Given the PR is titled for global download head consistency, I'd expect the causality the other way round. Was the asymmetry deliberate, or did currency-l0 just get the later revision?

`ExactSnapshotStorage.installExactForRecovery` (`SnapshotStorage.scala:77`) has zero call sites outside the node-shared unit test. Either wire it or drop it.

**The e2e can't tell forward from backward.** `test-rollback-download-head.sh:306` sums `dag_download_head_publication_total{path="full"}` across every `direction` label and asserts the total moved. The `direction` label the PR added is never read, and the scenario restores the target from an *older* snapshot, so it produces `initial`/`forward` by construction. The 520 lines are otherwise a genuinely strong test — `set -euo pipefail`, real hash convergence across all three nodes, a signer-proof check requiring the recovered node to sign a *later* round — which is why it's worth pointing out that the one direction the doc makes a claim about is the one it doesn't reach.

**`CL_QUORUM_THRESHOLD_FRACTION: "1.0"` was deleted from the job env, not overridden for the new entry.** That env block covers every matrix entry in the job, so currency, rewards, token-locks, allow-spends, fork-recovery and committee-rewards now all run global L0 at the `dag-l0.conf:76` default of 0.6666… instead of unanimous. The diff comment explains why *this* scenario needs 2/3 — third node deliberately offline — and that reasoning is fine, but the cost is that in every other scenario a node can stop signing for an entire run and the suite still goes green. That's a real reduction in what the existing e2e proves. Override it per-entry.

Two smaller things:

`synchronizeDownloadedOutcome` (`GlobalSnapshotConsensusStateAdvancer.scala:242`) is a second caller of the weak primitive, and there the failure surfaces as an exception on a consensus path rather than a download retry. Same fix covers it.

Worth saying that `b08141f5d` is a good catch — `java.util.Objects.equals` on a context reaching `Array[Byte]` through `DataApplicationPart` would have failed the readback for every metagraph with a data application, so the guard was unshippable before it. `schema/package.scala:125`'s `arrayEq` is in scope via the `schema._` import at `currency.scala:14`, so the derived `Eq` really is content-aware. It does illustrate the class of problem I flagged last time: a guard strictly stronger than what the round trip can promise turns into an unbounded retry, and this one was only found because someone wrote the test.

Last: ADR 0034 in #1597 requires a `Consensus compatibility` section on consensus-adjacent PRs. #1597 has one, this PR doesn't. Between the two, this is the one where I'd actually want to read the classification.

ryle-ai:
Two corrections to what I wrote above, both of which make the points sharper rather than softer.

On the quorum change — you already had the right fix and then reverted it. `f4143a6d2` added a per-row `quorum:` key with `CL_QUORUM_THRESHOLD_FRACTION: ${{ matrix.test-group.quorum || '1.0' }}` and a comment saying "keep that quorum override scoped to this matrix row rather than weakening the other E2E jobs." `f592fa92e`, titled "prove exact rollback download handoff," deletes the row key and the fallback line together. So my suggestion to scope it per-entry is really just "restore `f4143a6d2`." Was there a reason the templated form didn't work — matrix key not resolving in that env block, something like that? If it was incidental, that's the whole fix.

On `b08141f5d`, the array fix is right but it inherited a second behavior I don't think you want. `Signed.eq` (`Signed.scala:72`) deliberately ignores `proofs`, and its scaladoc says so: "Callers at an authority boundary where the proof set itself carries meaning ... must use `sameValueAndProofs` explicitly." You clearly know this, because `exactHeadMatches` (`SnapshotStorage.scala:105-106`) compares `storedSnapshot.proofs === expectedSnapshot.proofs` by hand rather than leaning on the instance. But the disk readback at `:384` compares `persistedState === state`, and `Eq[GlobalSnapshotInfo]` recurses through `lastCurrencySnapshots` into `Signed[CurrencyIncrementalSnapshot]`, where it picks up the proof-blind instance. So the guard that the log line and the scaladoc both call "exact" is exact for the top-level artifact and semantic-modulo-signatures for every embedded currency snapshot.

I don't have a reachable failure for it — proofs round-trip through the codec and `NonEmptySet` normalizes ordering, so it would take a codec bug to diverge. The new test can't see it either, since `SnapshotStorageSuite.scala:428` asserts `persistedContext.exists(_ === context)` against the same instance being installed. So this is about the strength of the claim, not a live bug. Either compare re-serialized bytes (`JsonSerializer` is already in scope, and that's exact and array-safe in one move), or drop "exact" from the message and the doc so the next person doesn't build on a guarantee that isn't there.

scasplte2:
@ryle-ai Thanks for the detailed passes. I pushed `894f11f43`, which incorporates the actionable findings while preserving the existing recovery-authority boundary.

- **Backward publication contract:** Global publication now explicitly permits a lower head only when the target ordinal is absent or already contains the same canonical snapshot value/hash after caller validation of the target and persisted ancestry. A different target value fails closed as `SnapshotOrdinalCollision`. I did not route Global through `replaceCanonicalSuffixForRecovery`, because that would authorize conflicting-target overwrite and suffix replacement beyond this PR.
- **Cache coherence:** Lower positive caches are reconciled before head publication using the ordinal-selected hasher. A persisted different value evicts the stale ordinal/hash entries; a pending write or matching readable hash index preserves legitimate cached-only state; unsupported missing or undecodable entries are evicted. Readable ordinals and evictions settle stale `notPersistedCache` markers. Reads complete before any cache mutation, and typed I/O failure leaves the visible head unchanged.
- **Publication proof and observability:** Exact target publication performs one write attempt and one proof across both indexes. Distinct typed errors classify target collision, snapshot-index readback, and semantic-context readback. When both write and proof fail, the typed proof failure retains the write error as suppressed. DAG download emits structured failure coordinates and `dag_download_head_publication_failure_total{path,direction,reason}`; success counters remain post-durable and before consensus initialization. `follower_catch_up` is restored as its own path, and both recovery publication sites emit success counters.
- **Shared link behavior:** `LocalFileSystemStorage.link` remains unchanged for all callers. This boundary independently verifies both target indexes, so a swallowed link failure cannot publish the head and is reported as `snapshot_index_readback`.
- **Quorum:** The per-row template from `f4143a6d2` is restored, and the missing GL0 Compose pass-through is added so it is effective. Small-cluster CI uses the intended unanimous fallback; `rollback-download-head` alone uses exact 2/3 while one member is offline. Production DAG L0 remains 2/3.
- **Claims and E2E scope:** Context readback is now described as semantic equality, not byte/proof-envelope exactness. The E2E text states that it proves rollback-lead initialization and forward full download, not backward publication. The cold-start comment now explicitly records filesystem read-through before peer fallback. The pre-existing unused `installExactForRecovery` helper remains unchanged.

Consensus compatibility is now documented in the PR body: runtime/storage-publication behavior plus test-only E2E configuration; no signed schema, codec, persisted format, production quorum, activation ordinal, or recovery-authority change.

Local validation on `894f11f43`: 81/81 focused tests, whole-project compile, Scalafmt, Scalafix imports, commitlint for the complete PR range, shell/YAML checks, rendered quorum assertions, and diff checks all pass. The fresh GitHub workflows are running. Because the Compose pass-through makes unanimity effective for the ordinary CI rows, the complete E2E matrix remains the outstanding qualification evidence.

scasplte2:
Follow-up on the post-update CI run: [Build JARs](https://github.com/Constellation-Labs/tessellation/actions/runs/34394506417/job/102610741220) failed in `sdk / Compile / doc` because the two new comments attached to local helper definitions were parsed as unmoored Scaladoc. This was a source-caused documentation-build failure, not an infrastructure retry.

Fixed in `eefed9d03` by using ordinary block comments for those local helpers. On the exact rebased head, `sdk/doc`, whole-project `sbt compile`, `scalafmtCheckAll`, commitlint for the full PR range, and `git diff --check` pass. The new PR workflows are running now. This follow-up changes comments only and does not change publication, cache, quorum, or production behavior.

scasplte2:
@ryle-ai review the latest changes

ryle-goehausen (approved):

PR #1597, opened 2026-09-06
fix(config): align activation defaults and consensus checks

scasplte2:
## Summary

Missing environment entries previously selected inconsistent ordinal defaults, and several tests relied on those defaults to enable current behavior. This PR centralizes disabled defaults, makes ordinary test fixtures explicitly enable current behavior, and protects configured historical boundaries with a complete expected-value table.

DAG and Currency L0 also include all resolved FieldsAddedOrdinals thresholds, the three shared hashing/state-proof/staking boundaries, and the complete dust-sweep schedule in the existing consensus config hash. Nodes running the same version reject disagreement about these settings during joining. Every deployment uses one software version and a full-cluster cold restart; version mismatches remain independently rejected.

The PR adds ADR-0034 and shared contributor/Codex/Claude guidance, consolidates the existing local Just/Docker E2E documentation, and records collision-safe ADR allocation.

## Consensus compatibility

- [ ] Not consensus-adjacent
- [ ] Runtime-only behavior; encodings/hash rules and historical replay are unchanged
- [x] Replay/state-transition change with stable schema and an explicit compatibility boundary

The original fallback change affects incomplete/custom configuration: absent first-active thresholds select MaxValue instead of activating selected rules from genesis. Packaged activation values and existing comparison operators are unchanged. Established historical cutovers remain explicit; ordinary tests enable current behavior, while historical tests specify the boundary they exercise.

The follow-up intentionally changes the existing consensus configuration hash, including the value advertised at joining, in live Facility declarations, and in their signed trigger statements. The same resolved configuration supplies joining and the running consensus engine. It does not change snapshot encodings, snapshot hash/signature construction, state-proof shapes, or persisted snapshot history. No activation ordinal or additional rollout process is introduced; the change ships through the normal full-cluster cold restart.

Last-legacy hashing and state-proof selectors retain their established defaults. Dust sweeps retain exact-key behavior, including historical replay. All configured activation numbers remain unchanged.

## Consumer alignment

Snapshot Streaming will be updated separately after develop includes these changes. Its compatibility patch still contains four older missing-environment fallback sites; the follow-up must align them with the SDK accessors and qualify the resulting consumer artifact. Explicit packaged values already agree. This PR does not claim completion of that separate build/deployment or change Block Explorer schemas.

## Validation

- Full affected-module suites: `nodeShared/test` (1,562 passed), `dagL0/test` (238 passed; two existing ignored generator tests), `dagL1/test` (50 passed), `currencyL0/test` (160 passed), and `currencyL1/test` (24 passed): **2,034 passed, zero failures**.
- Full-project `sbt compile`, `scalafmtCheckAll`, and `scalafixAll --check --rules OrganizeImports` passed. Full compilation and formatting were also repeated successfully at committed head `5459dd3f6` before pushing.
- Commitlint passed for the actual new commit and all five outgoing commits against refreshed `origin/develop`; `git diff --check` passed.
- All five packaged `.conf` files are byte-identical to the reviewed PR head. Snapshot Streaming's patch is unchanged.
- GitHub CI runs separately on the pushed commit; local Docker E2E was not rerun for this follow-up.

### DIFF
--- a/.claude/commands/debug-e2e-logs.md
+++ b/.claude/commands/debug-e2e-logs.md
@@ -6,6 +6,12 @@ description: "Diagnose an E2E fork-recovery failure from persisted per-node logs
 
 Use this slash command when the user reports an E2E test failure — fork-recovery, consensus stall, committee divergence, pre-isolation sync timeout, etc. It codifies the procedure documented in `reference_e2e_log_analysis.md`. Do not jump to fixes or blame a recent commit without running this procedure first.
 
+Read `docker/README.md` first for the canonical Just lifecycle, artifact locations, staged-JAR
+provenance checks, and cleanup matrix. Use
+`docs/operations/local-e2e-cluster-investigation.md` before teardown when the failed cluster is
+still queryable. This command is the persisted-log analysis companion, not a replacement E2E
+harness.
+
 ## Inputs
 
 - **$ARGUMENTS** (optional): free-text notes from the user about the failure (which node stuck, which ordinal, what test output showed, which commit to investigate). If empty, use the most recent prior user message as context.
@@ -17,7 +23,11 @@ ls -la /home/scas/git/tessellation/nodes/0/gl0-logs/
 wc -l /home/scas/git/tessellation/nodes/*/gl0-logs/gl0-run.log
 ```
 
-If logs aren't present (no `nodes/<N>/gl0-logs/gl0-run.log`), tell the user the E2E teardown wiped them and propose adding `docker logs gl0-N > ... .log` capture to `docker/bin/test-fork-recovery.sh`. Stop.
+If logs aren't present (no `nodes/<N>/gl0-logs/gl0-run.log`), do not assume ordinary Docker teardown
+removed them: `just down` and `--cleanup` preserve host-mounted logs. Check whether another run,
+`just clean-data`, or `just clean-configs` replaced the `nodes/` tree, and whether CI uploaded a
+`runner-logs/` artifact. If no evidence remains, report the exact gap before proposing changes.
+Stop.
 
 If logs are present, continue.
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -14,3 +14,5 @@ Affected signed/persisted types, hashes, proofs, or historical behavior:
 Snapshot Streaming, Block Explorer, SDK, and metagraph impact:
 
 Activation/replay tests and external artifact versions:
+
+Activation domain, comparator, missing/default behavior, and historical-boundary evidence:
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,6 +1,6 @@
 ## Consensus compatibility
 
-<!-- Read docs/adr/0034-consensus-schema-change-governance.md for consensus-adjacent changes. -->
+<!-- Read docs/adr/0034-consensus-schema-change-governance.md for consensus-adjacent changes. Select exactly one classification. -->
 
 - [ ] Not consensus-adjacent
 - [ ] Runtime-only behavior; encodings/hash rules and historical replay are unchanged
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,16 @@
+## Consensus compatibility
+
+<!-- Read docs/adr/0034-consensus-schema-change-governance.md for consensus-adjacent changes. -->
+
+- [ ] Not consensus-adjacent
+- [ ] Runtime-only behavior; encodings/hash rules and historical replay are unchanged
+- [ ] Replay/state-transition change with stable schema and an explicit compatibility boundary
+- [ ] Schema/wire change with an approved schema-change package
+
+Classification and evidence:
+
+Affected signed/persisted types, hashes, proofs, or historical behavior:
+
+Snapshot Streaming, Block Explorer, SDK, and metagraph impact:
+
+Activation/replay tests and external artifact versions:
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -15,6 +15,8 @@ encodings, hash/signature construction, persisted history, replay semantics, Sna
 Block Explorer, public APIs, SDKs, or metagraph artifacts automatically compatible. Schema/wire
 changes require an explicit design and rollout decision; replay/state-transition changes require a
 historical compatibility boundary when old artifacts would otherwise re-derive differently.
+Missing `FieldsAddedOrdinals` threshold mappings are disabled, never active from genesis; an
+explicit `0` is required for active-from-genesis behavior.
 
 When classification is uncertain, treat the change as schema/wire-impacting until the exact signed
 and persisted surfaces and external consumers have been audited. Do not implement or approve an
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -18,6 +18,10 @@ historical compatibility boundary when old artifacts would otherwise re-derive d
 Missing `FieldsAddedOrdinals` threshold mappings are disabled, never active from genesis; an
 explicit `0` is required for active-from-genesis behavior.
 
+Before adding or renumbering an ADR, inspect `docs/adr/` on the latest target branch, choose the
+next unused four-digit number, and recheck after the final rebase. If another change claimed the
+number first, renumber the newer proposed ADR and update every reference before merge.
+
 When classification is uncertain, treat the change as schema/wire-impacting until the exact signed
 and persisted surfaces and external consumers have been audited. Do not implement or approve an
 unplanned schema change as an incidental part of a behavioral fix.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,21 @@
+# Repository agent instructions
+
+## Consensus and history compatibility
+
+Before changing consensus, snapshot construction/acceptance, hashing, serialization, state proofs,
+persistence, P2P messages, or SDK-visible types, read
+[`docs/adr/0034-consensus-schema-change-governance.md`](docs/adr/0034-consensus-schema-change-governance.md).
+
+Every affected PR must classify itself as runtime-only, replay/state-transition, or schema/wire.
+Do not call a change schema-neutral merely because it does not edit a file under `schema/`.
+
+A coordinated cold restart permits the fleet to adopt new runtime behavior together, including
+behavior that produces different future consensus outcomes. It does not make changes to signed
+encodings, hash/signature construction, persisted history, replay semantics, Snapshot Streaming,
+Block Explorer, public APIs, SDKs, or metagraph artifacts automatically compatible. Schema/wire
+changes require an explicit design and rollout decision; replay/state-transition changes require a
+historical compatibility boundary when old artifacts would otherwise re-derive differently.
+
+When classification is uncertain, treat the change as schema/wire-impacting until the exact signed
+and persisted surfaces and external consumers have been audited. Do not implement or approve an
+unplanned schema change as an incidental part of a behavioral fix.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -14,6 +14,8 @@ outcomes—to change together. It does not make signed encodings, hash/signature
 persisted history, replay semantics, Snapshot Streaming, Block Explorer, SDK, or metagraph changes
 compatible. Do not implement an unapproved schema change as part of a behavioral fix. If impact is
 uncertain, treat it as schema/wire until the signed, persisted, and external surfaces are audited.
+Missing `FieldsAddedOrdinals` threshold mappings are disabled; active-from-genesis behavior must use
+an explicit `0`.
 
 ## Project Overview
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -17,6 +17,10 @@ uncertain, treat it as schema/wire until the signed, persisted, and external sur
 Missing `FieldsAddedOrdinals` threshold mappings are disabled; active-from-genesis behavior must use
 an explicit `0`.
 
+Before adding or renumbering an ADR, inspect `docs/adr/` on the latest target branch, choose the
+next unused four-digit number, and recheck after the final rebase. If another change claimed the
+number first, renumber the newer proposed ADR and update every reference before merge.
+
 ## Project Overview
 
 Tessellation is the Constellation Network Node Software - a DAG (Directed Acyclic Graph) based distributed ledger with Layer 0 (L0) and Layer 1 (L1) validators. Written in Scala 2.13, designed for Kubernetes deployment.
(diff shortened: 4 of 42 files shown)


ryle-ai (commented):
The uniform `resolveOrDisabled` is presented as an alignment refactor, but for about fourteen gates it reverses the unconfigured fallback from `SnapshotOrdinal.MinValue` to `MaxValue`. Enumerating from the diff: `tessellation3Migration`, `tessellation301Migration`, `checkSyncGlobalSnapshotField`, `metagraphSyncData`, `updatedLastSyncGlobalOrder`, `updatedLastSyncGlobalFromPeersInConsensus`, `updatingCombineFunctionSpendActions`, `fixingAllowSpendExpiration`, `fixingAllowSpendAndTokenLockValidation`, `preventingAllowSpendResurrection`, `setSumFix`, `fixingAllowSpendDestinationCredit`, `fixingDataApplicationFeeValidation`, and `incrementalDelegatedStakingStartingOrdinal`. Before, an absent environment entry meant "new behavior from genesis." After, it means "never." That's the opposite semantics, and the commit is titled as a config alignment.

With the shipped `application.conf` this is inert — all four `AppEnvironment` values are populated for every one of those keys, so a stock node resolves identically before and after. That's why I'm raising it as a question rather than a blocker. The question is what happens for a config that isn't the shipped one:

A currency-L0 metagraph packages its own `application.conf`. On the JVM classpath that shadows rather than merges with node-shared's, so if a metagraph's config doesn't carry a `fields-added-ordinals` block, every gate in that list silently resolves to `MaxValue` for it after this change. `fixingAllowSpendExpiration` and `preventingAllowSpendResurrection` are both money-path gates on the currency side (`CurrencySnapshotAcceptanceManager.scala:199-222` is exactly where they're read), and `checkSyncGlobalSnapshotField` / `updatedLastSyncGlobalOrder` affect global-sync semantics. Losing those quietly is a worse failure than the one the new convention is trying to prevent, and nothing in the diff logs the resolution or fails fast, so the only symptom would be a metagraph producing snapshots under pre-fix rules.

I don't know whether real metagraph configs carry that block — you're better placed to answer that than I am. But the fix either way is cheap: log every resolved gate and its source once at startup, or fail fast when a gate that the code actually consults is absent. `FieldsAddedOrdinalsSuite` currently asserts the new `MaxValue` behavior directly, which codifies the convention without covering the thing that makes it safe. What I'd add is a test that walks all four environments and asserts every threshold gate in the shipped `application.conf` resolves to something other than `MaxValue`, with an explicit allowlist for the ones meant to be dormant (`currencySnapshotProtocolV1` on public envs, `fixingDelegatedStakeDoubleWithdrawal`). That turns "inert today" into "provably inert" and catches a future key deletion, which is the actual risk the uniform fallback creates.

`incrementalDelegatedStakingStartingOrdinal` deserves its own mention because the new comment at `application.conf:123-124` makes it sound routine. `MaxValue` there means incremental delegated staking never starts. Previously an absent entry meant it started from genesis. For any environment that ends up without the key, delegated staking becomes permanently dormant rather than active — a stuck-feature failure that won't look like a config problem.

There's also a tension the PR introduces and then doesn't follow through on. The `resolveOrDisabled` scaladoc carves out `dustSweeps` because "absence is already their disabled state" — so the PR already establishes that gates aren't all one kind. But `fixingAllowSpendDestinationCredit` is described in its own config comment as a boundary "used only to recreate signed history," which makes it a third kind: for a network with no legacy history, the historically-correct default is the current semantics, not the pre-fix ones. `MinValue` was arguably right for that one and `MaxValue` right for a forward activation gate, and the uniformity argument flattens the distinction rather than resolving it. If the intent is that every gate must be explicit and the fallback exists only as a safety net, say that in the scaladoc — it's a defensible position and it's different from what the current wording implies.

On the docs side, `GlobalSnapshotDustSweep`'s scaladoc changed from "the per-environment compile-time `dustSweeps` config literal (NOT HOCON): the advertised release plus the environment is the determinism fence" to "the per-environment `dustSweeps` HOCON packaged into the assembly." The new statement is the accurate one, but it weakens the fence and the PR doesn't say so. A compile-time literal can't be overridden at runtime; HOCON packaged in an assembly can be, via `-Dconfig.file` or an `application.conf` earlier on the classpath. Since the sentence's whole purpose is to tell a reader what guarantees determinism, "packaged into the assembly" is doing too much work — it should say that the fence depends on operators not overriding it, which is a materially different claim and is the one an ADR about consensus schema change governance should be making explicitly.

Small: `incrementalDelegatedStakingStartingOrdinalFor` takes no parameter and closes over `environment`, while every other `...For` accessor takes the environment. Same suffix, different shape. Either give it the parameter for consistency or drop the `For`.

Checked, no change needed: narrowing `applyDustSweep` to take a resolved `Option[DustSweep]` instead of the map plus environment plus ordinal is right, and `dustSweepFor` keeps the exact-key semantics intact. The `forProduct22` reader ordering comment is still accurate against the case-class field order.

ryle-ai:
Follow-up on the fallback flip — there's a concrete consequence inside this diff that I missed on the first pass.

Eight `FieldsAddedOrdinals(...)` fixture sites still pass `Map.empty` positionally for the first ten gate fields:

- `modules/dag-l0/src/test/.../GlobalSnapshotTraverseSuite.scala:342`, `:396`, `:423`
- `modules/dag-l0/src/test/.../GlobalSnapshotStateChannelEventsProcessorSuite.scala:126`, `:189`, `:410`
- `modules/dag-l0/src/test/.../GlobalSnapshotConsensusFunctionsSuite.scala:366`
- `modules/currency-l1/src/test/.../CurrencySnapshotProcessorSuite.scala:83`

All eight pass `Dev` as the environment, so under the old `MinValue` fallback every one of those gates was active and under `resolveOrDisabled` every one of them is now off. `GlobalSnapshotTraverseSuite` is the long-range replay suite, which makes this the worst place for it to land: the fixture whose job is catching replay regressions would now traverse with `tessellation3Migration`, `metagraphSyncData` and `setSumFix` disabled. Those tests either fail (loud, fine) or keep passing while covering different code (quiet, not fine), and I can't tell which from reading alone.

You already have the fix pattern in the diff — `Mocks.scala:70-90` and `SnapshotProcessorSuite.scala:86-104` define `devActivation = Map(AppEnvironment.Dev -> SnapshotOrdinal.MinValue)` and spell out every gate, with the comment "Those gates are explicit so an empty map keeps its production meaning: disabled." It's applied to two files and not the other eight. Worth extracting that into a shared test fixture rather than a third copy.

Relatedly, PR Validation lists `sbt compile`, `scalafmtCheckAll` and `FieldsAddedOrdinalsSuite`. For a change whose whole content is "unconfigured now means disabled," `dagL0/test`, `dagL1/test` and `currencyL1/test` need to be green in the PR before this merges, and their being green needs to mean the gates were on.

Minor: `docs/operations/fields-added-ordinals.md` cites "Examples from `application.conf:210-293`" but the block now runs to 399.

ryle-ai:
Second pass, now having read ADR 0034 properly. The ADR is the strongest part of this PR — it's concrete enough to argue with, which is more than most policy docs manage. Two findings, both about the gap between what it requires and what the repo can currently enforce.

**Nothing pins an already-crossed mainnet activation ordinal.** The ADR says "Never move an already-crossed gate forward or reinterpret the signed interval behind it," and `FieldsAddedOrdinalsSuite` is where I went looking for the guard. It isn't there. The test named `mainnetThresholds` (`:83-116`) builds `disabledFieldsAddedOrdinals` by hand with empty maps and asserts the *resolver* returns `MaxValue` — it never loads `application.conf`, so the variable name is misleading about what it covers. The only tests that read real config pin `feeTransactionSecurity` across four envs (`:32`), six gates for Integrationnet only (`:51`), and `currencySnapshotProtocolV1` (`:133`), and that last one asserts the three public envs are *absent*.

So `fixing-allow-spend-expiration` at mainnet 5033174, `fixing-global-allow-spend-expiration` at 6828500, `set-sum-fix`, `tessellation-3-migration`, `fixing-data-application-fee-validation` at 6818000 — every one of those could be edited to a different value, or deleted outright, and the suite stays green. Deletion is the worse half now: under the old fallback it meant active-from-genesis, which diverges loudly on the first snapshot, and under `resolveOrDisabled` it means disabled, which looks exactly like running an older release. A golden table asserting every gate's per-environment value straight out of `application.conf` is maybe thirty lines and it's the mechanical guard the ADR's own consequences section says is missing.

**The fourteen gates aren't in `deterministicConfigHash`, which is a better answer to my earlier question than a startup log.** The ADR asks contributors to "record ... whether the resolved value is included in `deterministicConfigHash`." I checked: `types.scala:1088-1183` covers timing, committee and admission parameters plus exactly two ordinal-shaped fields, `penaltyDurationOrdinals` and `currencySnapshotProtocolV1ActivationOrdinal`. None of the `FieldsAddedOrdinals` gates are in it.

That matters because L0 handshake requires exact `deterministicConfigHash` equality (`dag-l0/.../snapshot/schema.scala:126`) — a mechanism built precisely so a divergent operator config rejects instead of forking. Last round I asked for a startup log or fail-fast for the shadowed-metagraph-config case. Feeding the resolved gate values into the config hash is the same protection but enforced by the cluster instead of by whoever reads the log. It's a coordinated change and needs its own activation story, which is a fair reason to do it separately — but the ADR should then say the gates are deliberately outside the hash and why, since it asks every future PR to answer that question about its own gate.

**A question rather than a finding, on how far the convention generalizes.** The rule as written is "every missing `FieldsAddedOrdinals` threshold mapping resolves to `MaxValue`; active-from-genesis requires an explicit `0`." That's sound for the first-active-at family. But there's a second family with inverted polarity — "last ordinal at which the *old* thing applies" — and its two members already disagree with each other. `TessellationIOApp.scala:122` resolves `lastKryoHashOrdinal` with a `MinValue` fallback, so unconfigured means everything is JsonHash; `:127` resolves `lastLegacyStateProofOrdinal` with `MaxValue`, so unconfigured means everything is legacy. For inverted gates "disabled" translates to `MinValue`, so `MaxValue` there would be the unsafe default, not the safe one. `lastKryoHashOrdinal` picks the hashing algorithm, which is about as determinism-critical as this codebase gets, and the ADR's own schema-sensitive list already names hash preimages and retained Kryo registrations.

I don't think this blocks anything — both fields sit outside `FieldsAddedOrdinals` and are arguably out of scope. But the ADR reads as though one universal default exists, and the counterexample is two lines away from a gate it does cover. Naming both families and their respective safe defaults would stop the next person from "fixing" `lastKryoHashOrdinal` to match.

One correction to something I implied earlier: I'd noted a `forProduct21` → `forProduct22` change. That belongs to #1593, not here. `ext/pureconfig/pureconfig.scala:71-93` is byte-identical to develop, still 21 keys, and I walked the HOCON key order against the case-class field order position by position — no transposition. Nothing to do.

ryle-ai:
Two additions.

The unpinned-value point above has a worst case worth naming specifically. `application.conf:352-357` sets `fixing-fee-transaction-balance-overflow` to mainnet `6814499`, and the comment above it says "This gate is deliberately set inside signed history: the point is to preserve that history, not to change it." That's the gate holding replay across the fee-transaction overflow mint. A one-digit edit, or a merge resolution taking the wrong side, moves the boundary and every replaying node re-derives different balances across it — with green CI the whole way, since no test reads that value. Same exposure for `6818000` (`fixing-data-application-fee-validation`, `fixing-allow-spend-destination-credit`), `6828500` (`preventing-allow-spend-resurrection`, `fixing-global-allow-spend-expiration`), and the ten legacy migration gates. If only one gate gets pinned, pin that one.

Second, the ADR asks for something the repo doesn't currently do, and the counterexample is in this repo rather than a consumer's. It requires "Use the same selector in producer, validator, follower, download/recovery, and external re-derivation paths," and names Snapshot Streaming in the schema-sensitive list. `docker/snapshot-streaming/snapshot-streaming-state-proof.patch` is untouched by this PR and still resolves four gates with a `MinValue` fallback — `tessellation3Migration` (`:105`), `fixingAllowSpendDestinationCredit` (`:107`), `setSumFix` (`:121`), `incrementalDelegatedStakingStartingOrdinal` (`:123`). After this PR the node resolves those to `MaxValue` and the indexer that re-derives state proofs resolves them to `MinValue`, so absent config means "disabled" on one side and "active from genesis" on the other.

Inert on stock config, for the same reason as the main finding. But the ADR is what makes it worth raising: the first PR to state the rule is a good place to bring the repo's own external re-derivation path into line, or to note in the ADR that the SS patch is knowingly on the old convention until it's regenerated.

scasplte2:
@ryle-ai Implemented the focused follow-up in [5459dd3f6](https://github.com/Constellation-Labs/tessellation/commit/5459dd3f69b533219e32d5b8bae18bea0a1fa23f).

- **Current-behavior tests:** replaced all eight positional empty-map fixtures and the two duplicated explicit fixtures with one shared fixture that enables all 20 current thresholds. Historical tests keep explicit historical settings.
- **Activation values:** added an independent expected-value table for every gate and environment, the three separate hashing/state-proof/staking boundaries, and the complete dust-sweep schedule. This includes Mainnet 6814499, 6818000, and 6828500. No packaged activation value changes.
- **Consensus config hash:** both L0 applications now include those resolved thresholds and the complete dust schedule, including sweep ordinals, thresholds, and burn/credit destinations. Joining and live consensus receive the same effective configuration. Tests cover each threshold, removal of configured entries, environment isolation, and ordering. Currency protocol-v1's activation value was already included; the follow-up fills the remaining shared activation coverage.
- **Documentation:** distinguished first-active thresholds, last-legacy boundaries, and exact-ordinal events; preserved the existing hashing/state-proof defaults; corrected the HOCON and dust-replay wording. The incremental-staking accessor now takes the environment consistently.

The fixture issue was a concrete change in test coverage. The expected-value table and expanded hash add useful protection; the review did not identify a changed packaged historical boundary.

Two operational points:

1. We **never deploy mixed-version networks**. Joining rejects version mismatches, and **every deployment uses a full-cluster cold restart**. These config checks compare settings among nodes running the same version. This change uses that normal deployment process and needs no additional activation scheme.
2. **Snapshot Streaming is a separate follow-up after develop is updated.** Its four older fallback sites remain documented for that update; the explicit packaged values already agree.

The named-config shadowing concern does not match the node loading path: named layer configs explicitly fall back to the shared defaults. Added tests load both real named L0 configurations and check the shared activation values across every environment. Removing the required activation block fails config loading; omitting an environment within a present map uses the disabled fallback. No actual node configuration losing its shared settings was demonstrated.

Validation: **2,034 tests passed** across nodeShared, DAG L0/L1, and Currency L0/L1, with two existing ignored DAG L0 generator tests. Full compilation, formatting/import checks, and commitlint passed, including compilation at the committed head. Docker E2E was not rerun for this follow-up; GitHub CI runs on the pushed commit.

ryle-goehausen (approved):

PR #1599, opened 2026-09-09
fix(gossip): prevent stalled responses from blocking timed snapshots

Proph151Music:
## Summary

Prevent unfinished gossip responses from indefinitely occupying recurring request workers and blocking timed snapshot progress. A separate CI-only commit makes data and fee E2E tests self-contained.

## Problem and changes

The existing five-second gossip timeout bounds response acquisition but not subsequent body consumption. Enough stalled bodies can prevent healthy peers' messages from being fetched.

The correction applies the existing deadline through body consumption and decoding for recurring peer queries, common queries and common offers. Timeout releases the response resource so existing cleanup and health checks can proceed. Initialization, consensus thresholds, membership rules, phase timers and reward calculations remain unchanged.

The CI repair replaces the unavailable private-key secret with fresh job-local identities and restricted temporary key files. Signed transactions and exact data-state and fee-transfer assertions remain required.

## Validation

In an isolated five-validator comparison with the same 210-second fault, stock completed zero timed epoch advances; the corrected binary completed three reward-bearing snapshots. The affected round completed in 65–68 seconds, with no sampled snapshot-value conflicts. Both runs used a two-worker test override; the eight-worker default was tested separately.

561 Scala tests and 28 tooling tests passed locally, with two existing ignored tests. At combined head bdd28647428082c6be552d7297867bac2c909ed4, [all eight functional E2E suites](https://github.com/Proph151Music/tessellation/actions/runs/34182399659) and [Java 11 Scala checks](https://github.com/Proph151Music/tessellation/actions/runs/34182399504) passed. Both repaired CI jobs passed nine helper tests; fee verification confirmed equal debit and credit of 100 base units.

## Safety and limitations

Quorum, signatures and finality requirements are unchanged. This demonstrates recovery from the reproduced transport stall, not the cause of every Mainnet slowdown or a guaranteed daily reward rate. No compensation is implemented. Production-scale throughput, full fault-target reentry and supported mixed-version upgrades remain unqualified. Earlier test failures and measurement limitations are retained in the evidence.

## Compatibility and evidence

Tests used v3.5.30. The v3.5.31 BioFi-address change has no direct overlap; merge-tree validation passed, but runtime qualification of the merged source remains outstanding. The CI scripts partially overlap upstream PR #1538; integration with that change requires preserving its fee-lookup assertions.

- [Gossip implementation, controlled comparison and safety evidence](https://github.com/Proph151Music/tessellation/pull/3)
[...]

PR #1600, opened 2026-09-10
fix(gossip): prevent stalled responses from blocking timed snapshots

scasplte2:
## Summary

An acquired peer-gossip response can keep one of the bounded recurring workers occupied indefinitely because the existing client timeout ends before response-body consumption. This change consumes and decodes the peer-rumor response within the existing deadline, then re-emits the decoded rumors after the deadline scope so local hashing and queue work are not attributed to the peer.

This is a clean organization-hosted transplant of the narrowed implementation from #1599. The commit retains @Proph151Music as its author, and the two changed files match source head `7a64ebe9b6e5866ab76d7ff9919d95f90ff28b3d`.

## Scope

- Restrict the behavior change to `queryPeerRumors`, which carries consensus declarations.
- Leave common-rumor queries, offers, and initialization unchanged.
- Add focused regressions for stalled bodies, total deadlines despite body activity, response-resource release, decode failures, and downstream processing outside the peer deadline.

No Python, Docker, workflow, configuration, consensus, reward, schema, codec, or persisted-format changes are included.

## Validation

- The current #1599 Scala check passed against the same two-file source tree.
- This internal PR is intended to run the `release/mainnet` E2E workflow with its repository secret available; the corresponding fork run could not supply `CI_PRIVATE_KEY` to two legacy transaction jobs.
- The additional local validation run was skipped at the submitter's direction so the organization CI could start immediately.

## Compatibility

This is an immediately active runtime transport change and requires no activation ordinal. Normal Mainnet binary/version rollout coordination still applies.

Supersedes #1599 for organization CI and merge purposes.

### DIFF
--- a/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/p2p/GossipClient.scala
+++ b/modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/p2p/GossipClient.scala
@@ -50,9 +50,19 @@ object GossipClient {
 
       def queryPeerRumors(request: PeerRumorInquiryRequest): PeerResponse[Stream[F, *], Signed[PeerRumorRaw]] =
         PeerResponse("rumors/peer/query", POST)(timeoutClient, session) { (req, c) =>
-          c.stream(req.withEntity(request)).flatMap { resp =>
-            resp.body.chunks.parseJsonStream[Json].evalMap(_.as[Signed[PeerRumorRaw]].liftTo[F])
-          }
+          // Bound acquisition and decoding, then release the response before emitting.
+          // Downstream hashing and queue backpressure must not consume the peer's deadline.
+          Stream
+            .eval(
+              c.stream(req.withEntity(request))
+                .flatMap { resp =>
+                  resp.body.chunks.parseJsonStream[Json].evalMap(_.as[Signed[PeerRumorRaw]].liftTo[F])
+                }
+                .compile
+                .toList
+                .timeout(gossipTimeoutsConfig.client)
+            )
+            .flatMap(Stream.emits)
         }
 
       def getInitialPeerRumors: PeerResponse[Stream[F, *], Signed[PeerRumorRaw]] =
--- a/modules/node-shared/src/test/scala/io/constellationnetwork/node/shared/infrastructure/gossip/p2p/GossipResponseDeadlineSuite.scala
+++ b/modules/node-shared/src/test/scala/io/constellationnetwork/node/shared/infrastructure/gossip/p2p/GossipResponseDeadlineSuite.scala
@@ -1,0 +1,213 @@
+package io.constellationnetwork.node.shared.infrastructure.gossip.p2p
+
+import java.util.concurrent.TimeoutException
+
+import cats.data.NonEmptySet
+import cats.effect.std.{Random, Supervisor}
+import cats.effect.testkit.TestControl
+import cats.effect.{IO, Ref, Resource}
+import cats.syntax.all._
+
+import scala.concurrent.duration._
+
+import io.constellationnetwork.node.shared.config.types.{GossipRoundConfig, GossipTimeoutsConfig}
+import io.constellationnetwork.node.shared.domain.cluster.services.Session
+import io.constellationnetwork.node.shared.domain.healthcheck.LocalHealthcheck
+import io.constellationnetwork.node.shared.http.p2p.headers.`X-Id`
+import io.constellationnetwork.node.shared.infrastructure.cluster.storage.ClusterStorage
+import io.constellationnetwork.node.shared.infrastructure.gossip.GossipRoundRunner
+import io.constellationnetwork.node.shared.infrastructure.metrics.{Metrics, NoOpMetrics}
+import io.constellationnetwork.schema.cluster._
+import io.constellationnetwork.schema.generation.Generation
+import io.constellationnetwork.schema.gossip._
+import io.constellationnetwork.schema.node.NodeState
+import io.constellationnetwork.schema.peer._
+import io.constellationnetwork.security.hash.Hash
+import io.constellationnetwork.security.hex.Hex
+import io.constellationnetwork.security.signature.Signed
+import io.constellationnetwork.security.signature.signature.{Signature, SignatureProof}
+
+import com.comcast.ip4s.{Host, Port}
+import eu.timepit.refined.auto._
+import fs2.Stream
+import io.circe.Json
+import io.circe.syntax._
+import org.http4s._
+import org.http4s.client.Client
+import weaver.SimpleIOSuite
+
+object GossipResponseDeadlineSuite extends SimpleIOSuite {
+  private val id = PeerId(Hex("1" * 128))
+  private val context = P2PContext(Host.fromString("127.0.0.1").get, Port.fromInt(9001).get, id)
+  // Transport/decode fixtures only; signature and session authentication are not under test.
+  private val session = new Session[IO] {
+    def createSession: IO[SessionToken] = IO.raiseError(new IllegalStateException("unused"))
+    def verifyToken(peer: PeerId, token: Option[SessionToken]): IO[TokenVerificationResult] = IO.pure(TokenValid)
+  }
+  private val config = GossipTimeoutsConfig(10.seconds, 5.seconds)
+  private val rumor = Signed(
+    PeerRumorRaw(id, Ordinal.MinValue, Json.fromString("test"), ContentType("test")),
+    NonEmptySet.one(SignatureProof(id.toId, Signature(Hex("1" * 128))))
+  )
+  private def query(client: Client[IO]): Stream[IO, Signed[PeerRumorRaw]] =
+    GossipClient.make(client, session, config).queryPeerRumors(PeerRumorInquiryRequest(Map.empty)).run(context)
+
+  test("acquisition and unfinished body share one deadline and release the response exactly once") {
+    TestControl.executeEmbed {
+      for {
+        acquired <- Ref.of[IO, Int](0)
+        released <- Ref.of[IO, Int](0)
+        transport = Client[IO] { _ =>
+          Resource.eval(IO.sleep(4.seconds)) >> Resource.make(
+            acquired.update(_ + 1).as(Response[IO]().putHeaders(`X-Id`(id)).withBodyStream(Stream.never[IO]))
+          )(_ => released.update(_ + 1))
+        }
+        start <- IO.monotonic
+        result <- query(transport).compile.drain.attempt
+        elapsed <- IO.monotonic.map(_ - start)
+        count <- acquired.get
+        freed <- released.get
+      } yield
+        expect(result.left.exists(_.isInstanceOf[TimeoutException])) &&
+          expect.same(elapsed, 5.seconds) && expect.same(count, 1) && expect.same(freed, 1)
+    }
+  }
+
+  test("continually trickled bytes do not reset the total deadline and the response is released") {
+    TestControl.executeEmbed {
+      for {
+        chunks <- Ref.of[IO, Int](0)
+        released <- Ref.of[IO, Int](0)
+        body = Stream.repeatEval(IO.sleep(100.millis) >> chunks.update(_ + 1).as(32.toByte))
+        transport = Client[IO] { _ =>
+          Resource.make(IO.pure(Response[IO]().putHeaders(`X-Id`(id)).withBodyStream(body)))(_ => released.update(_ + 1))
+        }
+        start <- IO.monotonic
+        result <- query(transport).compile.drain.attempt
+        elapsed <- IO.monotonic.map(_ - start)
+        received <- chunks.get
+        freed <- released.get
+      } yield
+        expect(result.left.exists(_.isInstanceOf[TimeoutException])) &&
+          expect.same(elapsed, 5.seconds) && expect(received > 1) && expect.same(freed, 1)
+    }
+  }
+
+  test("a decoded prefix is not emitted when the response body never completes") {
+    TestControl.executeEmbed {
+      for {
+        emitted <- Ref.of[IO, Int](0)
+        body = Stream.emits((rumor.asJson.noSpaces + "\n").getBytes("UTF-8")).covary[IO] ++ Stream.never[IO]
+        transport = Client[IO](_ => Resource.pure(Response[IO]().putHeaders(`X-Id`(id)).withBodyStream(body)))
+        result <- query(transport).evalTap(_ => emitted.update(_ + 1)).compile.drain.attempt
+        count <- emitted.get
+      } yield expect(result.left.exists(_.isInstanceOf[TimeoutException])) && expect.same(count, 0)
+    }
+  }
+
+  test("complete decoded response is released before downstream work exceeding five seconds") {
+    TestControl.executeEmbed {
+      for {
+        released <- Ref.of[IO, Int](0)
+        transport = Client[IO] { _ =>
+          Resource.make(IO.pure(Response[IO]().putHeaders(`X-Id`(id)).withEntity(rumor.asJson.noSpaces)))(_ => released.update(_ + 1))
+        }
+        start <- IO.monotonic
+        result <- query(transport).evalMap(r => released.get.flatMap(n => IO.sleep(6.seconds).as((r, n)))).compile.toList
+        elapsed <- IO.monotonic.map(_ - start)
+        count <- released.get
+      } yield expect.same(result, List((rumor, 1))) && expect.same(elapsed, 6.seconds) && expect.same(count, 1)
+    }
+  }
+
+  test("slow local processing completes through the real runner without a peer healthcheck") {
+    TestControl.executeEmbed {
+      implicit val metrics: Metrics[IO] = NoOpMetrics.make
+      Supervisor[IO].use { implicit supervisor =>
+        Random.scalaUtilRandomSeedInt[IO](0).flatMap { implicit random =>
+          val peer = Peer(
+            id,
+            context.ip,
+            Port.fromInt(9000).get,
+            context.port,
+            ClusterSessionToken(Generation.MinValue),
+            SessionToken(Generation.MinValue),
+            NodeState.Ready,
+            Responsive,
+            Hash.empty
+          )
+          for {
+            completed <- Ref.of[IO, Int](0)
+            healthchecks <- Ref.of[IO, Int](0)
+            cluster <- ClusterStorage.make[IO](ClusterId("8d07c061-d42f-4d9c-9efc-37e0d1ee73e7"), Map(peer.id -> peer))
+            transport = Client[IO](_ => Resource.pure(Response[IO]().putHeaders(`X-Id`(id)).withEntity(rumor.asJson.noSpaces)))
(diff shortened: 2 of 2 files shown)


ryle-ai (commented):
Cancellation semantics look right to me (the `.timeout` on the compiled `F` cancels the inner fiber and the `Stream.resource` finalizer releases the response, and the suite really does exercise a stalled body under TestControl, not the happy path). Two things I'd want addressed before this goes in, one question.

The fix isn't on develop. `git diff origin/develop <head> -- GossipClient.scala` is the same +13/-3, so develop still has the unbounded body read. If this is a mainnet-only hotfix that's fine, but the PR should say what the forward-port plan is, otherwise the next release/mainnet cut from develop reintroduces the 210s stall from #1599.

This also changes what the 5s means. Before, `gossip.timeouts.client` bounded only response acquisition (`TimeoutMiddleware` is `Resource.timeout`, which races allocation only), so a slow-but-progressing body always completed. Now transfer + JSON parse + `Signed[PeerRumorRaw]` decode all share the same 5s, and on failure `GossipRoundRunner` logs an error and spawns a `localHealthcheck.start(peer)`. `GossipRoutes` returns every rumor since the cursor for every known peer plus up to 50 additional per unseen peer, and facility declarations carry a full `Map[PeerId, Ordinal]` bound, so a burst round on a large GL0 cluster is plausibly hundreds of KB. `LocalHealthcheck` only marks the peer Unresponsive if its own probe fails, so the practical cost is noise plus extra fibers rather than eviction, but it is a new failure mode under GC pause or a lagging peer. Was 5s sized against a mainnet-scale round, or should the body budget be a separate (larger) config knob than acquisition?

Question: only `queryPeerRumors` is bounded; `getInitialPeerRumors` and `queryCommonRumors` keep the unbounded read and the new suite locks that in. I get that this was the deliberate narrowing from #1599, but `commonRound` shares the same 4-worker starvation shape, so the title claim only holds for the peer-rumor path. Worth a line in the description saying that's intentional.

Minor: the comment about "queue backpressure" is a bit misleading since the rumor queue is `Queue.unbounded`; the real downstream work is the hashing in `GossipDaemon`.

effects-ai:
The diagnosis is right: `TimeoutMiddleware.withTimeout` applies `.timeout` to a `Resource`, which races **acquisition only**, so an acquired-but-stalled body has no deadline. Confirmed against http4s 0.23.33 — ember's `withTimeout` is the *header receive* timeout, and body reads are only bounded per-read by the idle timeout.

But the chosen remedy — `compile.toList` plus a total `.timeout` at one call site — trades the stall for three new failure modes, and leaves the same stall on four of the five gossip calls.

**12 findings, most severe first. 11 are posted inline; finding 6 is in this summary because its file is not part of the diff.**

| # | Sev | Where | What |
|---|-----|-------|------|
| 1 | high | `GossipClient.scala:62` | All-or-nothing batching → cursor never advances → permanent sync livelock |
| 2 | high | `GossipClient.scala:63` | `client` timeout silently widened from acquisition to full round trip |
| 3 | high | `GossipClient.scala:62` | Unbounded heap buffering; TCP backpressure removed; ×8 concurrent rounds |
| 4 | high | `GossipClient.scala:68` | `getInitialPeerRumors` still unbounded — join path, no runner, hangs forever |
| 5 | high | `GossipClient.scala:75/80/87` | Common round still unbounded — the *same* bug, 4 workers, snapshot gossip |
| 6 | med | `TimeoutMiddleware.scala:12` | **(summary only — file not in diff)** the fix belongs in the shared middleware |
| 7 | med | `GossipClient.scala:59` | `401` from session verification parses as "zero rumors" → round reports success |
| 8 | med | `GossipClient.scala:63` | Two overlapping 5s deadlines; body budget = `5s − acquisition`, untested |
| 9 | med | `GossipClient.scala:53` | Local jawn/circe decode is still inside the peer deadline — contradicts the comment |
| 10 | low-med | `GossipClient.scala:54` | Comment's rationale is wrong: the rumor queue is `Queue.unbounded` |
| 11 | low | `GossipClient.scala:55` | Reuse: this is `Stream.evalSeq` |
| 12 | low | `GossipResponseDeadlineSuite.scala:123` | Test 5 passes on pre-change code — not a regression test |

---

### 6 (med, altitude) — the fix belongs in `TimeoutMiddleware.withTimeout`, not at one call site

**Real location: `modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/http/p2p/middlewares/TimeoutMiddleware.scala:12` — not part of this diff, so it could not be anchored inline.**

```scala
def withTimeout[F[_]: Async](client: Client[F], timeout: FiniteDuration): Client[F] =
  Client { req =>
    client.run(req).timeout(timeout)   // Temporal[Resource[F, *]] — races acquisition only
  }
```

The timer stops once the response is acquired; the body has no deadline. That single line is the root cause of every stall in this PR's description, and the same middleware is used by:

- `GossipClient.scala:46` — all **five** gossip calls
- `L0GlobalSnapshotClient.scala:28`
- `CurrencySnapshotClient.scala:25`
- `StateChannelSnapshotClient.scala:39`

This PR special-cases **one of those eight call sites** and, to do it, changes that call's streaming semantics (findings 1–3). Bounding `response.body` inside `withTimeout` — a total or idle deadline on the body stream, e.g. `resp.withBodyStream(resp.body.interruptAfter(timeout))` or an idle-timeout pipe — fixes all eight with no change to streaming behavior, no buffering, and no loss of partial progress. Special cases layered on shared infrastructure are the signal that the fix isn't deep enough.

---

### Suggested direction

1. Move the body deadline into `TimeoutMiddleware.withTimeout` (finding 6), which subsumes findings 4 and 5 and makes 1 and 3 unnecessary.
2. If the batching is kept regardless, add an element/byte cap so a peer cannot stage an unbounded response in heap (finding 3), and split the config into connect vs. total so the client budget is not smaller than the server's own `routes = 10 seconds` (findings 2 and 8).
3. Fix or drop test 5, and add a test that actually fails on the pre-change code for the stalled-worker scenario (finding 12).



---

## Inline findings

_These were prepared as inline PR comments. Each heading is the anchor location._

### `GossipClient.scala`:62

<sub>`modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/p2p/GossipClient.scala`</sub>

**1 (high) — `compile.toList` makes the peer-rumor query all-or-nothing, so a response that exceeds the deadline yields *zero* rumors and the cursor never advances.**

A node joins a ~100-peer cluster with an empty ordinal map. `GossipRoutes` `peer/query` then takes the `additionalOrdinals` branch and returns up to `peer-rumors-capacity` (50, `application.conf`) rumors for *every* known peer id — thousands of rumors. If transfer + decode takes 5.2s, `.timeout(5s)` cancels and the buffered `List` is discarded: nothing reaches `rumorQueue`, `addPeerRumorIfConsecutive` never runs, `getLastPeerOrdinals` is unchanged, and `GossipDaemon.peerRound` re-requests exactly the same oversized batch every 200ms forever.

Before this change, every rumor decoded within the window was enqueued and stored, so the cursor advanced each round and the batch shrank. The new test `"a decoded prefix is not emitted when the response body never completes"` enshrines the regression as intended behavior.

---

### `GossipClient.scala`:63

<sub>`modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/p2p/GossipClient.scala`</sub>

**2 (high) — `gossipTimeoutsConfig.client` silently changes meaning from "acquisition" to "total round trip including body transfer and JSON decode".**

`application.conf` sets `gossip.timeouts.client = 5 seconds` and `gossip.timeouts.routes = 10 seconds` — the responder is explicitly allowed 10s to produce a response the client will now discard after 5s *total*. Any healthy peer on a slow WAN link whose full response takes 5.5s now raises `TimeoutException` on every round; `GossipRoundRunner.evalRound` logs an error, skips `incrementGossipRoundSucceeded` / `recordRoundDuration`, and fires `localHealthcheck.start(peer)` — turning a working peer into a perpetual healthcheck target.

Previously the body had no deadline at all (only ember's per-read idle timeout), so these rounds succeeded. No config key, default, or doc was added to separate "connect/header" from "total".

---

### `GossipClient.scala`:62

<sub>`modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/p2p/GossipClient.scala`</sub>

**3 (high) — `compile.toList` buffers the whole decoded response with no size/count cap and removes the pull-driven TCP backpressure that previously throttled socket reads.**

With the old code, `peerRound`'s `evalMap(toHashed).enqueueUnterminated(...)` pulled one element at a time, so the socket was read only as fast as the node consumed it — resident bytes were bounded. `toList` now reads at line rate for the full 5s.

`gossip.daemon.peer-round.max-concurrent-rounds = 8`, so eight fully materialized `List[Signed[PeerRumorRaw]]` (circe `Json` ASTs, typically 5–20x wire size) can be resident at once. A malicious or misconfigured peer streaming at 100 MB/s for 5s stages ~500MB of JSON per round, x8 → OOM. There is no `Content-Length` check, element cap, or `.take(n)`.

---

### `GossipClient.scala`:68

<sub>`modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/p2p/GossipClient.scala`</sub>

**4 (high) — `getInitialPeerRumors` has the identical unbounded-body stall and sits on the node-join path with *no* round runner and no other deadline.**

*(Anchored on this context line; the affected method body is lines 68–73, unchanged by this PR.)*

`GossipDaemon.startAsRegularValidator` runs `initPeerRumorStorage(peer) >> runPeerRoundRunner >> initCommonRumorStorage >> runCommonRoundRunner >> consumeRumors` inside one supervised fiber. `initPeerRumorStorage` calls `getInitialPeerRumors`, whose only bound is the acquisition-only middleware timeout. A peer that returns `200 OK` headers then trickles one byte every 30s keeps the body alive (each read is inside ember's default idle timeout), so the join fiber never advances: the node never starts peer rounds, common rounds, or `consumeRumors` — and there is no error to observe.

---

### `GossipClient.scala`:68

<sub>`modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/p2p/GossipClient.scala`</sub>

**5 (high) — the exact bug this PR describes still exists on the common-rumor round.**

*Real location: `getCommonRumorOffer` (lines 75–78), `queryCommonRumors` (lines 80–85), `getInitialCommonRumorHashes` (lines 87–90) — outside the diff, so anchored on the nearest diff line.*

`GossipDaemon.commonRound` calls `getCommonRumorOffer.run(peer)` (`c.expect`, body consumption unbounded) then `queryCommonRumors(...).run(peer)` (stream, body consumption unbounded). `gossip.daemon.common-round.max-concurrent-rounds = 4`.

Four peers that acquire and then stall their `rumors/common/offer` or `rumors/common/query` bodies permanently occupy all four common-round workers in `GossipRoundRunner`'s `parEvalMapUnordered`. Common rumors — the snapshot gossip path — stop propagating cluster-wide, and nothing times out or healthchecks. This is precisely "an acquired peer-gossip response keeps one of the bounded recurring workers occupied indefinitely", left untouched.

---

### `GossipClient.scala`:59

<sub>`modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/p2p/GossipClient.scala`</sub>

**7 (med) — the touched function never checks `resp.status`, so the `Unauthorized` response substituted by `responseTokenVerifierMiddleware` parses as an empty rumor list and the round reports *success*.**

`PeerResponse(..., timeoutClient, session)` wraps the client in `PeerAuthMiddleware.responseTokenVerifierMiddleware`, which replaces the real response with `Response[F](status = Unauthorized)` — an empty `EmptyBody` — whenever the peer's `X-Id` header is missing or `session.verifyToken` returns anything but `TokenValid`.

`parseJsonStream` over an empty body emits nothing, `toList` yields `List()`, `peerRound` completes normally, `metrics.incrementGossipRoundSucceeded` fires, and no healthcheck starts. A node whose session token has rotated is silently cut off from that peer's consensus declarations with no log line and no metric anywhere.

Pre-existing, but this PR rewrites the function around it, and the new suite stubs `verifyToken` to always return `TokenValid`, so it stays invisible.

---

### `GossipClient.scala`:63

<sub>`modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/p2p/GossipClient.scala`</sub>

**8 (med) — two overlapping 5-second deadlines now apply to the same call, so the effective body budget is `5s − acquisition time`.**

`timeoutClient` (line 46) still applies `.timeout(5s)` to acquisition, and this new `.timeout(gossipTimeoutsConfig.client)` applies 5s to connect + headers + body + decode. A peer whose TCP connect and header response take 4.9s leaves 100ms for the entire body, so the round fails even though both individual phases are well inside the configured budget — anywhere from 5s down to ~0 depending on connect latency.

Test 1 (4s acquisition + `Stream.never` body) cannot distinguish this, because its body never completes under either interpretation.

---

### `GossipClient.scala`:53

<sub>`modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/p2p/GossipClient.scala`</sub>

**9 (med) — the change contradicts its own stated goal: jawn parsing and circe decoding of the whole batch are local CPU work and are still inside the peer deadline.**

This comment says *"Downstream hashing and queue backpressure must not consume the peer's deadline"*, but `parseJsonStream[Json].evalMap(_.as[Signed[PeerRumorRaw]].liftTo[F])` — the most expensive local step for a multi-thousand-element batch — runs inside `.timeout`.

During a stop-the-world GC pause or CPU starvation from concurrent consensus work, decoding a large batch pushes past 5s; the round fails with `TimeoutException` and `GossipRoundRunner.evalRound` blames the peer via `localHealthcheck.start(peer)` — exactly the misattribution this PR set out to eliminate.

---

### `GossipClient.scala`:54

<sub>`modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/p2p/GossipClient.scala`</sub>

**10 (low-med) — this rationale is factually wrong: there is no queue backpressure to move out of the deadline, because the rumor queue is unbounded.**

`SharedQueues.scala:14` creates it as `rumorQueue <- Queue.unbounded[F, Hashed[RumorRaw]]`, and dag-l0 / dag-l1 / currency-l0 / currency-l1 all alias `sharedQueues.rumor`. `enqueueUnterminated(rumorQueue)` in `GossipDaemon.peerRound` therefore never blocks.

A maintainer reading this comment will believe the gossip path is backpressured and replicate the `toList` pattern to "escape" a bottleneck that does not exist — while the real unbounded-growth risk in that queue goes unnoticed. The only local work that *was* inside the old deadline is `toHashed`.

---

### `GossipClient.scala`:55

<sub>`modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/p2p/GossipClient.scala`</sub>

**11 (low, cleanup) — `Stream.eval(fa).flatMap(Stream.emits)` is exactly fs2's existing `Stream.evalSeq`.**

```scala
Stream.evalSeq(
  c.stream(req.withEntity(request))
    .flatMap(resp => resp.body.chunks.parseJsonStream[Json].evalMap(_.as[Signed[PeerRumorRaw]].liftTo[F]))
    .compile
    .toList
    .timeout(gossipTimeoutsConfig.client)
)
```

Eleven lines of hand-rolled wrapping collapse to one combinator, and the batching-rather-than-streaming intent — the single most important thing a reader of this method needs to notice — becomes obvious at the call site.

---

### `GossipResponseDeadlineSuite.scala`:123

<sub>`modules/node-shared/src/test/scala/io/constellationnetwork/node/shared/infrastructure/gossip/p2p/GossipResponseDeadlineSuite.scala`</sub>

**12 (low, test quality) — this test passes identically on the pre-change code, so it guards nothing while costing the heaviest setup in the file.**

The transport serves a *complete* body via `withEntity(rumor.asJson.noSpaces)`. Under the old implementation the stream emits the single rumor immediately, `evalMap(_ => IO.sleep(6.seconds))` completes, no deadline ever applies to the body, and the round succeeds with zero healthchecks — exactly the asserted outcome. The test cannot fail if the fix is reverted.

Yet it wires up `ClusterStorage`, `Supervisor`, `Random`, `NoOpMetrics`, a fake `LocalHealthcheck` and a real `GossipRoundRunner` across ~42 lines and 7s of virtual time. Tests 1–4 and 6–7 are the ones that actually discriminate the change.

To make it a real regression test, give the transport a body that stalls partway (e.g. `Stream.emits(json) ++ Stream.never`) and assert the round still finishes and the worker is freed.

---

Proph151Music:
Thanks for the detailed review. The central criticism is correct: buffering the
whole response with `compile.toList` avoids charging downstream work to the peer,
but it also discards partial progress on timeout and removes streaming
backpressure. That is not a safe trade for this path.

I prepared a replacement as one commit directly on the current #1600 head:

https://github.com/Proph151Music/tessellation/commit/f1747e3de7de9e3a588386005e16a3dfa4dcc44e

The replacement:

- removes `compile.toList` and preserves streaming, backpressure, and cursor
  progress;
- applies response-body deadlines in the shared timeout middleware, covering all
  five gossip calls rather than only `queryPeerRumors`;
- keeps the existing 5-second value for response acquisition and body inactivity;
- adds a separate configurable 30-second total gossip-response budget, so the
  continuously trickled incomplete body from the native reproduction still
  terminates without forcing healthy transfers into a 5-second total budget;
- rejects non-success streaming responses explicitly instead of treating a 401
  as an empty successful round; and
- replaces the non-discriminating runner test with regressions for partial
  progress, continuous trickle, worker release, cancellation, and resource
  finalization.

Validation on the exact published tree:

- focused deadline suites: 13/13
- Node Shared: 282/282
- DAG L0: 93/93, with the same 2 existing ignored tests
- Currency L0: 41/41
- `scalafmtCheckAll` and `git diff --check`: pass

The 30-second default is deliberately a separate, conservative value; it is not
being presented as Mainnet-telemetry-derived. The organization should confirm or
adjust it during qualification. Also, the earlier native A/B validated the
superseded 5-second buffered candidate, so this replacement still needs its own
native A/B before merge.

Because #1600 is organization-owned, its branch owner will need to cherry-pick
`f1747e3de7de9e3a588386005e16a3dfa4dcc44e` and run the trusted CI/native
qualification.

ryle-goehausen (approved):

Proph151Music:
Quick post-merge clarification: I verified that `v3.5.33` contains the
original #1600 head (`4406291`/`d83ffbd`), not the reviewed replacement
`f1747e3` linked above.

In the released source, `TimeoutMiddleware` still bounds response acquisition
only, and only `queryPeerRumors` receives the additional
`compile.toList.timeout(5s)` treatment. This means the review findings remain
applicable: four other gossip response paths can still stall after acquisition,
streaming/backpressure and partial cursor progress are lost for peer queries,
and acquisition, transfer, and local decoding share the same five-second total
budget.

The replacement preserves streaming and partial progress, moves body deadlines
into the shared middleware, covers all five gossip calls, separates the
five-second acquisition/inactivity budget from a configurable 30-second total
response budget, and rejects unsuccessful streaming responses explicitly. Its
focused and full validation passed as documented above.

I am not claiming that every post-upgrade Mainnet incident is caused by these
remaining paths—#1607 documents a separate restart-loop trigger and amplifier. I
want to ensure the team is aware that the reviewed replacement itself is not
present in the live release.

I can rebase the replacement onto `v3.5.33` and submit it as a dedicated Mainnet
follow-up PR if that is the preferred path.

PR #1602, opened 2026-09-10
fix(mainnet): recover damaged download anchors

Proph151Music:
## Problem

Mainnet DAG L0 can remain in `WaitingForDownload` when the newest local snapshot
or snapshot-info path exists but its bytes cannot be decoded. Decoder exceptions
can escape storage, and retries can repeatedly select the same unusable anchor.

## Branch relationship

This is a Mainnet v3.5.31 backport and adaptation of the damaged-file recovery
work already merged into `develop` by #1588. The v4.1 implementation builds on
the larger crash-safe storage foundation from #1566; Mainnet does not have that
full storage stack, so this PR carries only the compatible pieces required for
safe recovery.

This PR is not intended to be forward-merged into `develop`. Mainnet deliberately
skips a mismatched persisted pair without deleting it and lets validated replay
replace it. The v4.1 implementation may delete a mismatched pair before
continuing. That difference is intentional.

The first commit is the exact organization-hosted Mainnet gossip fix from #1600.
The damaged-anchor recovery is a separate commit with no changed-file overlap;
stacking it here preserves the exact Mainnet candidate used for qualification.

## Changes

- Contain failures from both the primary JSON decoder and the historical fallback
  decoder.
- Keep expected JSON-to-Kryo fallback quiet; warn only when both decoders fail.
- Keep the highest persisted snapshot-info filename as evidence that this is
  recovery rather than a fresh-node bootstrap.
- Descend through lower candidate ordinals and validate each snapshot/info pair
  with the hasher selected for that historical ordinal.
- If persisted state existed but the bounded search finds no valid anchor, fail
  loudly for operator intervention instead of silently starting a
  multi-million-snapshot genesis replay.
- Use genesis only when no persisted snapshot-info filenames exist.
- Repair only the narrow case where an exact, readable hash-indexed snapshot
  exists and its ordinal index is missing.
- Atomically replace both persisted snapshot indexes after normal download
  validation, verify exact disk readback, and only then remove temporary data.

## Scope and safety

The existing ordinary snapshot-info writers, including live enqueue and
`persistSnapshotInfoWithCutoff`, remain unchanged and can still produce a damaged
file if a write is interrupted. Extending atomic replacement to every normal
persistence path is larger work and is intentionally outside this Mainnet
recovery hotfix. The atomic writer in this PR protects only installation of an
already validated recovery snapshot.

Downloaded snapshots still pass the existing chain, hash, and state-proof
validation. This change does not alter signatures, certificates, consensus
quorum, finality, admission, membership, or rewards. Malformed local bytes are
rejected rather than promoted.

## Verification

Review follow-up head `fdb8269a69db40353651fda8be60a69829ea7650`:

- Focused affected suites: 7/7.
- Full Shared suite: 133/133.
- Full Node Shared suite: 279/279.
- Full DAG L0 suite: 99/99, with 2 existing ignored tests.
- `scalafmtCheckAll` and `git diff --check` pass.

The original recovery commit remains separately qualified by the retained
isolated Mainnet GL0 runs:

- A damaged highest snapshot-info at ordinal 12 recovered to `Ready`, and all
  three nodes converged at ordinal 17.
- Damaged hash and ordinal snapshot indexes at ordinal 18 were rebuilt from a
  validated download; the node recovered to `Ready`, and all three nodes
  converged at ordinal 23.

The review follow-up changes logging and the exhausted-anchor failure policy; it
did not rerun those isolated network trials.

### DIFF
--- a/modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/domain/snapshot/programs/Download.scala
+++ b/modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/domain/snapshot/programs/Download.scala
@@ -56,6 +56,7 @@ object Download {
     val minBatchSizeToStartObserving: Long = 1L
     val observationOffset = NonNegLong(4L)
     val fetchSnapshotDelayBetweenTrials = 10.seconds
+    private val maxPersistedSearchAttempts: Int = 200
 
     type DownloadResult = (Signed[GlobalIncrementalSnapshot], GlobalSnapshotInfo)
     type ObservationLimit = SnapshotOrdinal
@@ -276,8 +277,11 @@ object Download {
       go(Map.empty, hash, ordinal)
     }
 
-    def isSnapshotPersistedOrReachedGenesis(hash: Hash, ordinal: SnapshotOrdinal): F[Boolean] = {
-      def isSnapshotPersisted = snapshotStorage.isPersisted(hash)
+    def isSnapshotPersistedOrReachedGenesis(hash: Hash, ordinal: SnapshotOrdinal)(
+      implicit hasherSelector: HasherSelector[F]
+    ): F[Boolean] = {
+      def isSnapshotPersisted =
+        hasherSelector.forOrdinal(ordinal)(implicit hasher => snapshotStorage.ensurePersistedAnchor(hash, ordinal))
 
       def didReachGenesis = ordinal === lastFullGlobalSnapshotOrdinal
 
@@ -374,13 +378,37 @@ object Download {
       state
         .map(_.pure[F])
         .getOrElse {
+          def findHighestValidPersisted(
+            lte: SnapshotOrdinal,
+            attempts: Int
+          ): F[Option[(Signed[GlobalIncrementalSnapshot], GlobalSnapshotInfo)]] =
+            if (attempts <= 0) none[(Signed[GlobalIncrementalSnapshot], GlobalSnapshotInfo)].pure[F]
+            else
+              snapshotStorage.getHighestSnapshotInfoOrdinal(lte).flatMap {
+                case None => none[(Signed[GlobalIncrementalSnapshot], GlobalSnapshotInfo)].pure[F]
+                case Some(ordinal) =>
+                  hasherSelector
+                    .forOrdinal(ordinal)(implicit hasher => snapshotStorage.readCombined(ordinal))
+                    .attempt
+                    .flatMap {
+                      case Right(Some(result)) => result.some.pure[F]
+                      case result =>
+                        val warning = result.swap.toOption
+                          .fold(logger.warn(s"Skipping unreadable persisted recovery anchor ordinal=${ordinal.show}"))(
+                            logger.warn(_)(s"Skipping invalid persisted recovery anchor ordinal=${ordinal.show}")
+                          )
+
+                        warning >>
+                          PartialPrevious[SnapshotOrdinal]
+                            .partialPrevious(ordinal)
+                            .map(findHighestValidPersisted(_, attempts - 1))
+                            .getOrElse(none[(Signed[GlobalIncrementalSnapshot], GlobalSnapshotInfo)].pure[F])
+                    }
+              }
+
           startingOrdinal
-            .flatTraverse(ordinal => hasherSelector.forOrdinal(ordinal)(implicit hasher => snapshotStorage.readCombined(ordinal)))
-            .flatMap {
-              _.map(_.pure[F]).getOrElse(
-                getGenesisSnapshot(tmpMap)
-              )
-            }
+            .flatTraverse(findHighestValidPersisted(_, maxPersistedSearchAttempts))
+            .flatMap(_.fold(getGenesisSnapshot(tmpMap))(_.pure[F]))
         }
         .flatMap {
           case (s, c) =>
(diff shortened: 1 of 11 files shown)


Proph151Music:
> @scasplte2, the two failing data/fee E2E jobs appear to be the same fork-secret limitation encountered on #1599. Could you use the same process as before—cherry-pick 621182e1f85fc14f81e49941224e40b0fa2720b7 onto the organization-hosted branch and rerun the trusted CI?

The recovery change is one commit directly on top of the exact #1600 commit, with authorship preserved.

For future PRs, is there a specific base branch, branch structure, commit layout, or preparation method you would prefer so submissions arrive as close as possible to the form needed for your organization-hosted branch and CI?

ryle-ai (commented):
Most of the storage-layer code here is the develop code from #1566 with the comments stripped (I diffed `CrashSafeAtomicFileWriter`, `SnapshotDownloadStorage`, `readCombined` against origin/develop; identical apart from comments and one exception message). That's fine for a mainnet backport, but the description should say it is one, since whoever forward-merges needs to know which parts are already on develop and which aren't.

Two things I'd change:

`LocalFileSystemStorage.read` now logs a WARN on every JSON decode failure before trying the fallback decoder. The old code only warned when both decoders failed. On mainnet everything at or below the last Kryo ordinal (~2.57M) is a Kryo file, so any read of a historical snapshot now emits a warning per file, and replay reads a lot of them. Logging volume on GL0 is already a real cost. Debug for the first message, or keep the single warn after the fallback also fails. Develop has the same thing, so it's a fix for both.

`findHighestValidPersisted` exhausting its 200 attempts (or `getHighestSnapshotInfoOrdinal` returning None) still falls through to `getGenesisSnapshot`, which on mainnet is a full replay from 766,717. Develop's version of the same function refuses that and raises, with a comment calling genesis re-download operationally unacceptable. This PR keeps the pre-existing mainnet fallback so it's not a regression, but a "recover damaged anchors" change that can still silently pick a multi-million-snapshot replay after 200 bad anchors leaves the worst outcome in place. I'd rather it fail loudly and let an operator decide.

Question on scope: the damaged info files come from `SnapshotStorage.enqueue` writing snapshot-info with the plain non-atomic `write`, and from `persistSnapshotInfoWithCutoff`. The PR ships the atomic writer but only uses it for recovery promotion, so torn writes will keep producing the anchors this PR then tolerates. Keeping a mainnet hotfix narrow is reasonable, just say so explicitly.

For the forward-merge: `readCombined` here raises on a mismatched pair and the search catches with `.attempt` and skips; develop's deletes the pair and returns None. The never-delete behaviour is the safer one for mainnet and the skipped ordinal self-heals when replay overwrites the info file, so I'm not asking for a change, only flagging that the two branches intentionally differ.

Proph151Music:
Thanks! These are all good catches. I updated the branch in
`fdb8269a69db40353651fda8be60a69829ea7650`.

What changed:

- Successful JSON-to-Kryo fallback is now quiet. A warning is emitted only when
  both decoders fail, so normal historical Mainnet replay will not produce one
  warning per Kryo file.
- The recovery path now distinguishes a fresh node from damaged persisted state.
  No snapshot-info filenames still means normal genesis bootstrap. If persisted
  state exists, the node searches lower anchors with a strict 200-attempt bound
  and fails loudly if none validate; it will not silently begin a
  multi-million-snapshot genesis replay.
- Added focused regressions for the persisted-state marker and the fail-loud
  selection rule.

Validation on the updated tree:

- focused affected suites: 7/7
- Shared: 133/133
- Node Shared: 279/279
- DAG L0: 99/99, with the same 2 existing ignored tests
- `scalafmtCheckAll` and `git diff --check`: pass

On lineage and scope: this is a Mainnet v3.5.31 backport/adaptation of the
damaged-file recovery already merged to `develop` in #1588, using the compatible
crash-safe pieces from the #1566 storage foundation. It is not intended to be
forward-merged as-is. Mainnet deliberately skips mismatched pairs without
deleting them; the v4.1 implementation may delete them, and that difference is
intentional.

Also agreed on the torn-write scope. Ordinary snapshot-info writes, including
enqueue and `persistSnapshotInfoWithCutoff`, remain unchanged and can still
create a damaged file if interrupted. Making every normal persistence path
atomic is larger work; this hotfix only makes recovery safe and makes validated
recovery promotion atomic.

ryle-ai (commented):
Re-read fdb8269a. All five points from my earlier pass are addressed: the single warn now fires only after both decoders fail and carries both messages, a fresh node with no snapshot-info filenames still goes to genesis while a node with persisted state raises `PersistedRecoveryAnchorsInvalid` after the 200-anchor descent, and the body now states the #1566/#1588 relationship, the out-of-scope torn-write sources, and the intentional `readCombined` difference. The raise isn't in `fetchNextSnapshot`'s retry set so it propagates out rather than looping, which is what I wanted.

Two small things. An operator who genuinely wants a genesis re-download now has to clear the snapshot_info directory to reach the `None` branch; that works, but it should be one line in the body since the old behaviour did it automatically. And the new `DownloadRecoverySelectionSuite` covers only the pure Option-to-raise helper; the genesis-vs-raise split itself has no test that walks a damaged max filename into the descent and out to the exception. Fine for a hotfix, noting it.

CI: the two red jobs (data-transactions, fee-transactions) fail at "Send transactions" with no secret source, same fork-secret limitation as #1599, so a trusted rerun from an org branch is still needed before merge.

ryle-goehausen:
@Proph151Music can you target develop with most recent changes? So we can test on newest release (can be done in separate PR or here)

Proph151Music:
Thanks. I kept the version-specific implementations separate:

- #1613 is the clean Mainnet `v3.5.33` recovery-only replacement for this PR, without the stacked gossip change.
- #1614 targets current `develop` with only the remaining cross-version fallback-logging correction. Current `develop` already contains #1588's recovery behavior and intentionally differs from Mainnet in `readCombined`.

I’m closing #1602 as superseded by #1613 and #1614. The review history remains here for reference.

PR #1604, opened 2026-09-10
fix(snapshot): remove hash sweeps and serve only indexed snapshots

scasplte2:
Port the snapshot cleanup and hash-serving fixes already applied to `release/testnet` onto `develop`. The reported testnet restart spent 89 minutes scanning 2.55 million hash files and reading 46 GB without removing any hash objects. Recovery now cleans future ordinal indexes, their directly identified hashes, indexed anchor conflicts and derived state without traversing the hash tree or consulting timestamps/link counts.

Retained hash-only objects do not establish canonical authority. Public and peer hash requests require a body matching the requested hash and an existing ordinal mapping for that same `(ordinal, hash)`; otherwise they return 404. Verification uses the ordinal-selected hasher, supports copied indexes and accepted canonical cache entries, returns the ordinal-selected snapshot, and never repairs indexes during serving. Internal reads remain available to validated recovery.

Snapshot reads contain decoder failures while propagating filesystem errors, and warn once only when both the primary JSON decoder and legacy fallback fail. Ordinary persistence verifies existing hash bodies before linking, atomically replaces malformed or mismatched bodies, and uses the existing atomic writer for new hash bodies.

## Port details

- Three cherry-picks from the applied release commits: `8a67ec76d`, `6edb1fa5f`, and `5de9d91f2`, based on `develop` at `7bb1ef6f5`.
- Preserve develop's existing GL0 staging replay through `replaceForRecovery`, which atomically replaces both indexes and supports retries after temporary-file loss. The imported poisoned-ordinal regression verifies exact repaired bytes at and below the cleanup boundary, preservation of the old torn inode, and no mutation on ordinal I/O failure. Successful repair does not require the two new indexes to share an inode.
- Preserve develop's newer download head publication and combined-response cache behavior. Omit source comment changes referring to a testnet-only watchdog heartbeat.
- Review follow-up avoids rehashing the ordinal-selected snapshot when accepted hash and ordinal caches return the same signed instance. Copied or separately decoded indexes retain a value-hash fallback check.
- The user reports successful operation after applying the work to testnet. This PR reuses the original review thread for the develop port; earlier review verdicts are historical evidence, not approval of this new head.

## Validation

Validated at `4aae2675e`:

- **144 tests pass, zero failures/errors**, across 14 suites: shared filesystem 1, nodeShared 59, GL0 70, Currency L0 14.
- Includes snapshot filesystem/torn-body tests; public/peer hash-serving and route/cache tests; GL0 validated recovery, snapshot storage, download, watchdog and terminal head publication; Currency cleanup and download.
- Whole-project `compile`, `scalafmtCheckAll`, `scalafixAll --check --rules OrganizeImports`, full-range commitlint and diff checks pass; worktree clean.
- The initial run hit the host `/tmp` inode limit. The successful rerun redirected temporary files for SBT and forked test JVMs into the workspace; no build configuration or test behavior was changed.

## Consensus compatibility

- [ ] Not consensus-adjacent
- [x] Runtime-only behavior; encodings/hash rules and historical replay are unchanged
- [ ] Replay/state-transition change with stable schema and an explicit compatibility boundary

**Classification and evidence:** This changes local file cleanup, repair and HTTP serving eligibility. It does not change serialized snapshot types, codecs, hash/signature construction, state-proof construction, snapshot acceptance or deterministic historical state reconstruction. Existing recovery anchor/context validation and develop's exact-envelope replacement remain in place. No golden wire/hash fixtures change.

**Persisted and external surfaces:** Global/Currency signed snapshot formats and ordinal/hash path layouts remain unchanged. Snapshot Streaming, Block Explorer, SDK and metagraph payload schemas remain unchanged; consumers requesting an unindexed or mismatched hash now receive 404. Indexed history and accepted canonical cache entries remain servable. No external consumer build, migration or deployment was performed or qualified in this port.

**Activation and replay:** No new gate, activation ordinal, comparator, missing/default mapping or network version change. Existing ordinal-based hasher selection is preserved and covered by the serving regression. Tests exercise recovery with retained/torn bodies and poisoned indexes at/below the cleanup boundary; no new long-range replay or external-artifact campaign is claimed.

## Remaining limitations

- Hash-only retention remains unbounded; this adds no garbage collector or serving-time repair.
- No new production restart or throughput benchmark. Atomic body writes incur fsync costs; cached hash serving performs one snapshot-value hash per request, while copied or separately decoded indexes require a fallback second value hash. Cache misses can still read snapshot bodies. Ordinal cleanup, replay and MPT reconstruction still take time.


ryle-ai (commented):
Three things, none of them about the cleanup itself (that's byte-identical to release/testnet so it's had runtime).

`getIndexedSnapshot` in `SnapshotRoutes.scala` hashes twice per by-hash request: `candidate.toHashed` and then `indexed.toHashed`. For recent ordinals `get(hash)` and `get(ordinal)` return the same cached `Signed[S]` instance (`get(ordinal)` resolves through `ordinalCache` to `get(hash)`), so the second hash is redundant work. This is the route metagraph nodes hit through `GlobalL0Service.pullGlobalSnapshot(hash)`, so a GL0 serving many metagraphs goes from a cache lookup to two full JSON serializations plus SHA per request. Comparing `indexed === candidate` (or `indexed eq candidate` short-circuit before falling back to hashing) gives the same guarantee. The description says hash serving "adds identity verification" but not that it doubles.

The ordinal leg of the same helper goes through `readFromDiskWithNegativeCache`, which remembers an ordinal as absent for 30s. Before this PR the hash route never touched the ordinal index so a stale negative entry couldn't reach it. Now a `/{ordinal}` probe that misses, followed within 30s by the snapshot landing through a path that doesn't pre-populate `ordinalCache` (recovery `setHeadForRecovery`, download `writePersisted`), returns 404 by hash for a fully indexed snapshot. Normal `enqueue` fills the positive caches first so consensus is fine; this is a recovery/download-era window. Probably worth bypassing the negative cache inside this helper rather than reasoning about which writers populate the cache.

The `read(fileName)` override in `SnapshotLocalFileSystemStorage` routes every snapshot read through `readRecoveryIndex`, which drops the two `logger.warn` lines the base class emitted on decode failure. That's listed as a known limitation, but it's the hot read path for all callers, so a corrupt body under a live ordinal is now indistinguishable in logs from a missing file. One warn after both decoders fail is small enough to land here.

Not asking for a change, just confirming I read it the same way: by-hash 404s for any body whose ordinal index is missing, and a historical ordinal with a hash-only body stays unservable until the next `write` of that ordinal or `ensurePersistedAnchor`. `pullGlobalSnapshot` returns `None` and clients retry another peer, so that's degraded rather than broken.

scasplte2:
Thanks for the careful read. I addressed the hashing and decoder-observability points in `4aae2675e`.

- `getIndexedSnapshot` now hashes only `candidate.value` for the requested-hash check. When the ordinal lookup returns the same `Signed` instance, an `eq` fast path returns it without hashing again. A different instance still gets its value hashed, preserving support for copied or separately decoded indexes without hashing proof envelopes. The regression uses a counting hasher and verifies one snapshot-value hash per public/peer request.
- `readRecoveryIndex` now warns once only after both the JSON decoder and legacy fallback fail. A successful legacy fallback stays quiet, while filesystem/read errors still propagate. The new test covers both outcomes.

I did not add a negative-cache bypass after tracing the publication paths. `setHeadForRecovery` installs the target hash and ordinal in the positive caches and invalidates the negative cache before exposing the head; exact recovery also invalidates the target. Direct `writePersisted` calls happen while download is in non-serving states, and terminal head publication performs the cache installation/invalidation before routes can serve from `Ready`/`WaitingForReady`. The existing `validated head publication exposes an externally installed suffix hidden by a cached miss` test exercises the relevant miss -> external install -> publication -> visible sequence. A bypass would therefore broaden disk reads without closing a supported post-publication window; an out-of-band file mutation with no publication remains outside that contract.

And yes, the 404 interpretation is intentional: a hash-only body is inert until an ordinal index is established. Clients can retry another peer; serving does not repair storage.

Local validation at `4aae2675e`: 144 tests across the 14 affected suites, whole-project `compile`, `scalafmtCheckAll`, `scalafixAll --check --rules OrganizeImports`, full-PR-range commitlint, and clean diff/worktree. CI is running on the new head.

ryle-ai (commented):
Re-read 1b2731fc. The double hash is fixed the right way: one value hash on the candidate, `eq` short-circuit on the ordinal leg, and a second hash only when the instances differ, which is exactly when it's needed. The decode warn is back and fires once after both decoders fail, with filesystem errors still propagating. Only side effect is that the recovery cleanup paths that intentionally treat malformed-as-absent now log twice per torn file (the new generic warn plus the existing caller-specific one). Bounded by torn-file count and recovery-only, so not worth changing.

On the negative cache, I traced your argument and it holds: `setHeadForRecovery` invalidates the negative cache before setting the head, the exact-recovery path invalidates the target ordinal, and `writePersisted` only runs while the routes are returning 503. Bypassing it would bring back the stat storm from community nodes polling pruned ordinals. Withdrawn.

Nothing new in the PR-owned delta; the docker/e2e/build.sbt hunks are the #1601/#1603 merges.

PR #1608, opened 2026-09-18
chore(deps-dev): update fast-uri and js-yaml

scasplte2:
Update the root development-tool lockfile to `fast-uri` 3.1.7 (from 3.1.5) and `js-yaml` 4.3.2 (from 4.3.1), combining the security updates from #1591 and #1605.

Only the version, registry URL and integrity fields for those two packages change, consistently in both lockfile sections. Existing local DAG4 package metadata is preserved. This supersedes #1591 and #1605.

Validation: full-project `sbt compile`, complete-range commitlint, and `git diff --check`. Structural comparison confirms the two intended package entries exactly match the Dependabot updates and every unrelated lockfile entry is unchanged; both target versions, URLs and integrity hashes match npm registry metadata.

## Consensus compatibility

- [x] Not consensus-adjacent
- [ ] Runtime-only behavior; encodings/hash rules and historical replay are unchanged
- [ ] Replay/state-transition change with stable schema and an explicit compatibility boundary

Classification and evidence: only root Node development-tool dependencies in `package-lock.json` change (`fast-uri` through AJV and `js-yaml` through cosmiconfig). Signed/persisted types, hashing, proofs, historical behavior, activation settings and Scala dependencies are unaffected. No Snapshot Streaming, Block Explorer, SDK or metagraph artifact update is required; activation/replay boundaries and external artifact versions are not applicable.


ryle-ai (commented):
Integrity hashes for `fast-uri@3.1.7` and `js-yaml@4.3.2` match the npm registry, both lockfile sections agree, nothing else in the lockfile moved. Nothing to change.

PR #1610, opened 2026-09-18
fix(gossip): prevent stalled responses from blocking timed snapshots

scasplte2:
When a peer returns response headers but stalls the gossip body, the existing HTTP client timeout no longer applies and one of the bounded gossip-round workers can remain occupied indefinitely. This ports Mainnet #1600 to `develop`: `queryPeerRumors` now acquires, consumes, and decodes the complete peer response within the configured client deadline, releases the response, and only then emits decoded rumors for local hashing and queue work.

The implementation and focused test suite are byte-identical to Mainnet commit `d83ffbd5e`. Other gossip queries retain their existing response-body lifetime. This runtime fix is separate from the BioFi and seedlist configuration port in #1609.

## Validation

- `sbt "nodeShared/testOnly io.constellationnetwork.node.shared.infrastructure.gossip.p2p.GossipResponseDeadlineSuite"` — 8 passed
- `sbt compile`
- commitlint — 0 errors (one inherited footer-spacing warning)
- `git diff --check`
- both changed file blobs match Mainnet commit `d83ffbd5e`

Source: https://github.com/Constellation-Labs/tessellation/pull/1600


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/p2p/GossipClient.scala
```diff
@@ -50,9 +50,19 @@ object GossipClient {
...
-          }
+          // Bound acquisition and decoding, then release the response before emitting.
+          // Downstream hashing and queue backpressure must not consume the peer's deadline.
+          Stream
+            .eval(
+              c.stream(req.withEntity(request))
+                .flatMap { resp =>
+                  resp.body.chunks.parseJsonStream[Json].evalMap(_.as[Signed[PeerRumorRaw]].liftTo[F])
+                }
+                .compile
```
ryle-ai, line 61: This is the hunk #1612 removes on mainnet. Because the whole decoded response is buffered before anything is emitted, a response that takes longer than `client` yields zero rumors, `lastOrdinalsR` never advances, and the next round 0.5s later re-requests the same batch. On develop that bites a joining node in a cluster of up to 100 origins (above that `limitOrdinals` randomises the subset). The streaming version in #1612 keeps per-rumor progress; I'd port that instead of this. Also the "queue backpressure" comment is wrong on develop too, `rumorQueue` is `Queue.unbounded`; the downstream cost is the hashing in `GossipDaemon`.

ryle-ai (commented):
The code is what it says: byte-identical to `d83ffbd5e`, compiles on develop (every test fixture it needs exists here with the right arity), the cancellation path releases the connection, and a `TimeoutException` goes through the normal round-failure path. My problem is with what it ports, not how.

#1600 got two reviews on mainnet that found the `compile.toList` approach deficient (all-or-nothing batching so a slow first-join response times out and re-requests the identical batch forever, unbounded heap buffering, `client` silently becoming a total budget, the other four gossip bodies still unbounded, 401 decoding as an empty successful round). The same author has #1612 open against `release/mainnet` that replaces this exact hunk with a streaming body-idle timeout in `TimeoutMiddleware` plus a separate total deadline. So merging #1610 gives develop a design mainnet is discarding, and whichever of #1610/#1612 lands second conflicts in the same lines of `GossipClient` and the suite.

I checked whether develop's config makes the livelock moot. It doesn't. `limitOrdinals` caps the inquiry at 100 random origins and sets `skipAdditionalPeers`, but only when more than 100 origins are known; at or below that the route still serves the full backlog for every unlisted origin from `Ordinal.MinValue`, which is the joining-node case. The 15s client timeout (vs 5s on mainnet) is the real mitigation, and the runner's 3-failures-in-60s cooldown turns a livelock against one peer into rotating exclusion of every large-backlog peer for a node that's behind.

I'd close this and port #1612 to develop as its own reviewed PR instead. It is not a clean cherry-pick: `GossipTimeoutsConfig.response` needs a `gossip.timeouts.response` key in develop's conf or boot fails at pureconfig load, develop's `client` is 15s so the idle:total ratio changes, and the shared-middleware change in #1612 also puts a body-idle deadline on `L0GlobalSnapshotClient`/`CurrencySnapshotClient`, where develop's `snapshot.timeouts.client` is 5s (mainnet's is 15s; the two branches have those swapped). A 5s idle deadline on full snapshot downloads is a new failure mode develop would get and mainnet wouldn't, so the port either scopes the body timeout to the 3-arg gossip overload or revisits those values.

If it merges anyway, the inline note about the comment stands, and the suite fixture uses mainnet's 5s not develop's 15s, so the elapsed assertions don't describe the shipped config.

Proph151Music:
Superseded by #1615, which ports the corrected streaming implementation to develop with develop-specific timeout configuration while preserving snapshot-download behavior. Closing this older buffered implementation to avoid merging the all-or-nothing response path.

scasplte2:
Closing in favor of #1615, which ports the streamed per-rumor body timeout from #1612. The buffered `compile.toList` approach here makes no progress at all when a response is slow, and it leaves the other four gossip bodies unbounded. Review of #1615 is posted there.

PR #1611, opened 2026-09-18
ci: check E2E runner dependencies before setup

scasplte2:
## Summary

- Check for host-side `curl` and `jq` immediately after checkout in each E2E matrix job, before Java setup and the readiness fixtures.
- Emit one GitHub error annotation per missing tool, with a runner image/bootstrap installation hint, and exit immediately.
- Leave package installation, runner provisioning, Snapshot Streaming, and the known Java-download failures unchanged.

## Why

The readiness fixture runs before `just` performs its normal dependency installation. On a runner without `jq`, suppressed parser errors appear as misleading `waiting_for=local_state` timeouts. Removing `jq` locally reproduced all 17 failing cases from [this CI job](https://github.com/Constellation-Labs/tessellation/actions/runs/35381024318/job/105722629075); all 25 cases pass with it available.

This PR improves diagnosis only. Runners must still provision both tools; the check does not install them or make an unprovisioned runner pass.

## Validation

- Executed the exact YAML shell block with isolated PATHs: both tools present, only curl, only jq, and neither. Correct exit status and complete missing-tool annotations in all four cases.
- Bash syntax and YAML parsing pass; structural comparison proves the guard is the only workflow change and precedes Java setup and readiness tests.
- Full-range commitlint and `git diff --check` pass.
- Existing readiness fixtures: 25/25 pass at the committed head.
- Exact-head full-project `sbt compile` passed (354 seconds), with no generated tracked changes.

Classification: CI-only runner-environment validation; no node runtime, consensus, replay, schema, activation, or deployment changes.


effects-ai:
Nice diagnosis — the `waiting_for=local_state` → missing-`jq` link is not an obvious one, and reproducing all 17 cases by removing it locally is convincing.

Complementary fix in #1606, so the two shouldn't collide: `afa644e0c` provisions the packages on the ephemeral Hetzner runners, which is the half your PR deliberately doesn't cover ("runners must still provision both tools"). It installs the same set the fixed pool already had in `deploy/ci-runners/fixed/terraform/templates/runner-init.tpl:23`:

```
ca-certificates curl wget gnupg jq lsb-release unzip git acl libatomic1
```

The on-demand fleet's `setup.sh` never had that list and was relying on whatever the `docker-ce` image happened to ship — `jq` isn't on it, which is what you hit.

**On sequencing:** no ordering hazard, because the fix is already applied to the live controller (~19:55 UTC today). Runners created since then have `curl` and `jq`, so #1611 can merge first without producing a window of fast-but-still-red jobs. `ci-runner-1` already had both independently (`curl 8.5.0`, `jq-1.7`), so neither runner type is missing them now.

Worth merging both regardless — the guard is the thing that makes a future regression obvious instead of another timeout hunt.

**One heads-up in the same class:** `libatomic1`. Node binaries link against `libatomic.so.1`, it's absent from the minimal Ubuntu cloud image, and without it any node — including the nvm-installed one — dies with "error while loading shared libraries". The fixed pool hit that as a live E2E failure; the ephemeral fleet has been passing only because the `docker-ce` image happens to include it. It's in the list above now and verified with `ldconfig` after install, but it's the same shape of bug: a hosted-runner assumption that holds until it doesn't.

PR #1612, opened 2026-09-18
fix(gossip): bound streamed response bodies

Proph151Music:
## Summary

This is the reviewed streaming replacement for the buffered gossip-body fix
merged in #1600 and released in `v3.5.32`/`v3.5.33`.

It is one commit directly on the exact `v3.5.33` head
(`bbde4e60fc9ada3926063852f48f5469c81094e1`). It preserves the Mainnet
seedlist and BioFi changes from #1607 and changes only the six listed
gossip/config/test files.

## Why a follow-up is needed

The released #1600 implementation bounds only `queryPeerRumors` by compiling
the full decoded response to a list under the existing five-second client
timeout. As identified in review, that makes progress all-or-nothing, removes
pull-driven backpressure, changes the five-second setting into a
transfer-plus-decode budget, and leaves the other four gossip response bodies
with acquisition-only timeout behavior.

## Changes

- Preserve response streaming, TCP backpressure, and partial cursor progress.
- Apply response-body deadlines in the shared timeout middleware so all five
  gossip calls are covered.
- Keep `gossip.timeouts.client = 5 seconds` for response acquisition and body
  inactivity.
- Add a separately configurable `gossip.timeouts.response = 30 seconds` total
  response budget so continuously trickled incomplete bodies still terminate
  without forcing healthy transfers into a five-second total budget.
- Reject unsuccessful streaming responses explicitly instead of treating an
  unauthorized empty body as a successful empty round.
- Replace the non-discriminating buffered-response expectations with
  regressions for:
  - all five gossip response paths;
  - partial progress before a later stall;
  - slow but active transfers;
  - continuously trickled incomplete bodies;
  - recurring-worker release;
  - caller cancellation and resource finalization; and
  - malformed or invalid responses.

## Validation

Validation on the exact submitted `v3.5.33`-based commit
`00ef7c5b3677c1e8ecd8aa2798f81debe4811714`:

- Focused middleware and gossip suites: 13/13 passed.
- `scalafmtCheckAll`: passed.
- `git diff --check`: passed.
- Stable patch ID exactly matches the previously reviewed and fully validated
  replacement `f1747e3de7de9e3a588386005e16a3dfa4dcc44e`.
- The earlier replacement validation also passed Node Shared 282/282, DAG L0
  93/93 with 2 existing ignored tests, and Currency L0 41/41.

## Qualification and scope

The 30-second default is a conservative configurable value, not a
Mainnet-telemetry-derived claim. Trusted CI should confirm it, and the
replacement still needs its own native Mainnet-scale A/B qualification before
merge because the earlier native A/B exercised the superseded buffered
implementation.

This follow-up does not contain the damaged-anchor/WaitingForDownload work from
#1602 and does not modify consensus, quorum, rewards, schemas, codecs, persisted
formats, or the #1607 operational changes.


modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/http/p2p/middlewares/TimeoutMiddleware.scala
```diff
@@ -1,15 +1,50 @@
...
+  ): Client[F] =
     Client { req =>
-      client.run(req).timeout(timeout)
+      client
+        .run(req)
+        .timeout(acquisitionTimeout)
+        .map(response => response.withBodyStream(transformBody(response.body)))
     }
+
+  def withTimeout[F[_]: Async](client: Client[F], timeout: FiniteDuration): Client[F] =
```
ryle-ai, line 37: This overload used to be acquisition-only and is also what `L0GlobalSnapshotClient:28`, `CurrencySnapshotClient:25` (both `snapshot.timeouts.client` = 15s on mainnet) and `StateChannelSnapshotClient:39` (5s, body only read on BadRequest) go through. After this change a `getFull` download that goes 15s without a chunk fails with `TimeoutException` where it previously hung or completed. Probably an improvement, since a 15s mid-body gap on an eagerly-encoded entity means a dead socket, but it's an undisclosed behaviour change on the join path with no test. Either say so in the description, or keep this overload acquisition-only and give the body-timeout variant its own name so the snapshot clients opt in explicitly. The second option also matters for the develop port, where `snapshot.timeouts.client` is 5s, not 15.

modules/node-shared/src/main/scala/io/constellationnetwork/node/shared/infrastructure/gossip/p2p/GossipClient.scala
```diff
@@ -43,32 +40,29 @@ object GossipClient {
...
-                }
-                .compile
-                .toList
-                .timeout(gossipTimeoutsConfig.client)
-            )
-            .flatMap(Stream.emits)
+          c.stream(req.withEntity(request)).flatMap { resp =>
+            if (resp.status.isSuccess)
+              resp.body.chunks.parseJsonStream[Json].evalMap(_.as[Signed[PeerRumorRaw]].liftTo[F])
+            else
```
ryle-ai, line 54: Worth knowing what this does downstream: `GossipRoundRunner.evalRound` catches it, logs at error, and calls `localHealthcheck.start(peer)`. If the peer's `/session` disagrees with our stored token the peer gets removed, which is the self-heal you want. If `/session` agrees, the check cancels and the next round 200ms later errors again, with no cooldown on this branch (develop has one). I couldn't construct that case for a healthy tessellation peer since both headers come from the same `SessionStorage`, so this is log-noise risk from a misbehaving endpoint, not eviction. Just noting the healthcheck coupling since the description only mentions the status change.

ryle-ai (commented):
Went through the twelve findings from the #1600 review against this. The batching, the 5s meaning change, the four unbounded bodies, the 401-as-empty-round, the nested deadlines and the non-discriminating runner test are all actually addressed, and 8 of the 13 new tests would fail on the pre-change tree (the other five pin invariants). `Pull.Timed` semantics check out against fs2 3.4.0 (`timeout` cancels the previous sleep via `switchMap`, stale tokens are filtered, so downstream work between pulls isn't charged), and `Stream#timeout` exists at that version and interrupts inside the response scope so the http4s finalizer runs. `timeoutOnPull` is not in 3.4.0, so the hand-rolled loop is the right tool here.

The thing I'd want stated, or scoped, before merge is inline on `TimeoutMiddleware`: the 2-arg `withTimeout` is also what `L0GlobalSnapshotClient`, `CurrencySnapshotClient` and `StateChannelSnapshotClient` use, so this PR gives full snapshot downloads a 15s idle-between-chunks deadline that the description ("changes only the six listed gossip/config/test files") doesn't mention.

The other one is the develop side. #1610 ports #1600's buffered version to develop byte-for-byte; this PR deletes every line #1610 adds. Whichever lands second conflicts in the same hunks of `GossipClient` and the suite, and `GossipTimeoutsConfig.response` has no default, so merging this case class into develop without adding `gossip.timeouts.response` to develop's conf fails at pureconfig load on boot. Develop's `client` is also 15s not 5s, so a straight port makes the idle timer 15s against a 30s total; someone needs to pick develop's values deliberately. I left the same note on #1610.

Two smaller things. The title says "bound" but there's no byte bound anywhere: `CirceSupportParser(None, false)` has no max value size, so one arbitrarily large JSON value still gets buffered by jawn until the idle or total timer fires. Not a regression, but the natural fix (a `Some(n)` there) is a one-liner and would make the title true. And 30s is asserted rather than sized; you say so yourself. Common-rumor queries return every missing artifact in one body, and a cold peer asking for 20 incremental snapshots at mainnet sizes could plausibly need more than 30s on a slow link. Progress is kept per decoded rumor so the retry shrinks, but a single artifact that can't transfer in 30s can never arrive via gossip.

No compile here; fs2 claims are from reading the 3.4.0 source, ember's pooled-connection behaviour after a body error I couldn't verify (no leak either way).

Proph151Music:
Addressed in `51fa24c10`.

The existing two-argument `withTimeout` remains acquisition-only, so `L0GlobalSnapshotClient`, `CurrencySnapshotClient`, and `StateChannelSnapshotClient` retain their previous behavior. Gossip now explicitly opts into `withResponseBodyTimeout`.

I added a regression test proving that an acquired response with a slow body is not timed out by the acquisition-only overload. The focused middleware and gossip suites pass 14/14.

I also changed the title to “time-bound” to clarify that this PR does not introduce an unvalidated JSON byte limit. A byte limit needs a measured safe maximum and should be handled separately.

Proph151Music:
The current head `4dc5435a3` includes the consecutive-failure cooldown and separate suspect-worker lane, along with the timeout classification and real cursor-advancement regressions.

Local validation passed:

- focused timeout, cursor, and isolation suites: 16/16;
- complete Mainnet `nodeShared` suite: 285/285;
- formatting, import, and diff checks.

GitHub’s Scala check and six functional E2E groups also passed. The fee-transaction and data-transaction jobs failed during “Send transactions” with exit code 7.

Those same two jobs failed on prior #1612 heads and on the unrelated v3.5.33-based #1613 recovery PR, which does not contain these gossip changes. That points to a shared Mainnet CI/baseline issue rather than this follow-up or the new slow-peer isolation.

Could someone with Actions log access rerun those two jobs or confirm the uploaded node-log failure?

PR #1613, opened 2026-09-18
fix(mainnet): recover damaged download anchors

Proph151Music:
## Summary

This is the reviewed Mainnet damaged-anchor recovery from #1602, restacked as
two recovery-only commits directly on the exact `v3.5.33` head
(`bbde4e60fc9ada3926063852f48f5469c81094e1`). It preserves the Mainnet gossip
change from #1600 and the seedlist/BioFi changes from #1607 without including or
reverting any of those files.

The two patches have the same stable patch identities as the reviewed recovery
and follow-up commits in #1602. The branch changes only the ten listed
recovery/storage/test files and is independent of the streaming gossip
replacement in #1612.

## Problem

Mainnet DAG L0 can remain in `WaitingForDownload` when the newest local snapshot
or snapshot-info path exists but its bytes cannot be decoded. Decoder exceptions
can escape storage, and retries can repeatedly select the same unusable anchor.

## Branch relationship

This is a Mainnet v3.5.33 backport and adaptation of the damaged-file recovery
already merged into `develop` by #1588. The v4.1 implementation builds on the
larger crash-safe storage foundation from #1566; Mainnet does not have that full
storage stack, so this branch carries only the compatible pieces required for
safe recovery.

It is not intended to be forward-merged into `develop`. Mainnet deliberately
skips a mismatched persisted pair without deleting it and lets validated replay
replace it. The v4.1 implementation may delete a mismatched pair before
continuing. That difference is intentional.

## Changes

- Contain failures from both the primary JSON decoder and the historical
  fallback decoder.
- Keep expected JSON-to-Kryo fallback quiet; warn only when both decoders fail.
- Preserve the highest snapshot-info filename as evidence of persisted state,
  even if that file is damaged.
- Descend through lower candidate ordinals and validate each snapshot/info pair
  with the hasher selected for that historical ordinal.
- If persisted state exists but the bounded 200-anchor search finds no valid
  anchor, fail loudly for operator intervention instead of silently beginning a
  multi-million-snapshot genesis replay.
- Use genesis only when no persisted snapshot-info filenames exist.
- Repair only the narrow case where an exact, readable hash-indexed snapshot
  exists and its ordinal index is missing.
- Atomically replace both persisted snapshot indexes after normal download
  validation, verify exact disk readback, and only then remove temporary data.

An operator who deliberately chooses a full genesis replay after the new
failure must first preserve diagnostics and then explicitly move or clear the
configured snapshot-info storage before restarting. The automatic destructive
fallback is intentionally removed.

## Scope and safety

The ordinary snapshot-info writers, including live enqueue and
`persistSnapshotInfoWithCutoff`, remain unchanged and can still produce a
damaged file if interrupted. Extending atomic replacement to every normal
persistence path is larger work and remains outside this Mainnet recovery
hotfix. The atomic writer here protects only installation of an already
validated recovery snapshot.

Downloaded snapshots still pass the existing chain, hash, and state-proof
validation. This change does not alter signatures, certificates, consensus
quorum, finality, admission, membership, or rewards. Malformed local bytes are
rejected rather than promoted.

## Validation

Validation on the exact `v3.5.33`-based tree:

- Focused affected suites: 13/13 passed.
- `scalafmtCheckAll`: passed.
- `git diff --check`: passed.
- No diff in gossip, Mainnet seedlist, BioFi allowance, or other #1607 files.
- Stable patch IDs exactly match #1602's reviewed recovery and follow-up
  commits.

The same two patches were previously validated with full Shared 133/133, Node
Shared 279/279, and DAG L0 99/99 (with two existing ignored tests). The original
recovery patch also passed the retained isolated damaged-info and
damaged-snapshot Mainnet GL0 recovery trials documented in #1602.


modules/dag-l0/src/main/scala/io/constellationnetwork/dag/l0/infrastructure/snapshot/SnapshotDownloadStorage.scala
```diff
@@ -48,7 +48,18 @@ object SnapshotDownloadStorage {
...
+        hashSelect.select(ordinal) match {
+          case JsonHash => snapshotInfoStorage.read(ordinal).map(_.isDefined)
+          case KryoHash => snapshotInfoKryoStorage.read(ordinal).map(_.isDefined)
+        }
+
+      def ensurePersistedAnchor(hash: Hash, ordinal: SnapshotOrdinal)(implicit hasher: Hasher[F]): F[Boolean] =
+        persistedStorage.ensureOrdinalLink(hash, ordinal).flatMap {
+          case status if status.usable => hasSnapshotInfo(ordinal)
+          case status =>
+            logger.warn(s"Rejected unusable persisted recovery anchor ordinal=${ordinal.show} reason=$status").as(false)
```
ryle-ai, line 64: This fires for `Missing` too, and `Download.go` calls `ensurePersistedAnchor` for every ordinal between the network tip and the node's own head (each of those is `Missing` by construction, the hash isn't local yet). So a restart after being down for N ordinals emits N of these, all saying `reason=Missing`, which buries the ones that matter (`HashUnreadable`, `HashContentMismatch`, `RepairIncomplete`). The base code logged nothing on this path. Develop's `ensurePersistedAnchor` returns unit for `Linked | Missing | OrdinalOccupied` with a comment that these are the normal backward-walk path; I'd match that here.

modules/dag-l0/src/test/scala/io/constellationnetwork/dag/l0/domain/snapshot/programs/DownloadRecoverySelectionSuite.scala
```diff
@@ -0,0 +1,30 @@
...
+
+import cats.effect.IO
+
+import io.constellationnetwork.schema.SnapshotOrdinal
+
+import weaver.SimpleIOSuite
+
+object DownloadRecoverySelectionSuite extends SimpleIOSuite {
+
+  test("persisted state with no valid recovery anchor fails instead of replaying from genesis") {
```
ryle-ai, line 11: Same note as on #1602: this covers the `Option.fold` helper, which can't really fail. The behaviour the PR title is about lives in `findHighestValidPersisted` (descend, 200-attempt bound, genesis only when no info filenames) and nothing drives a damaged max filename through it to `PersistedRecoveryAnchorsInvalid`. Coverage rests on the #1602 trials. Fine for a hotfix, but the suite name promises more than it tests.

ryle-ai (commented):
Patch-id matches #1602's two recovery commits exactly (diffed `fdb8269a~2..fdb8269a` against `release/mainnet..HEAD`, byte-identical), none of the ten files touch gossip, and nothing in them depends on #1600's commit, so the restack claim holds. The two things I had open on #1602 are: the operator genesis-replay instruction is now in the body, good; the descent still has no test that walks a damaged max filename down to `PersistedRecoveryAnchorsInvalid`, see inline.

One thing I'd fix before merge, inline on `SnapshotDownloadStorage.scala`: `ensurePersistedAnchor` warns on every non-usable status including `Missing`, and `Download.go` calls it once per ordinal on the walk back from the network tip, so a node that was down for N ordinals now logs N warnings on restart. Develop's version explicitly skips `Missing` and `OrdinalOccupied` for that reason.

Two things worth a line in the description rather than a code change. First, "fail loudly" here means the exception kills the `DownloadDaemon` fiber (`watchForDownload` runs `download` in an `evalTap` with no retry on mainnet), so the node sits in `WaitingForDownload` until someone restarts it, and the restart fails the same way until the snapshot_info dir is cleared. That's the intended outcome, but "the node parks, it doesn't exit" is what an operator needs to know. Second, `CrashSafeAtomicFileWriter` raises on any filesystem where `ATOMIC_MOVE` isn't supported, which turns every recovery promotion into the same parked state on a network mount. Probably no mainnet node runs that way, just flagging it.

Minor, same as develop, not asking for a change: `replaceForRecovery` derives `previousHashName` by decoding the old ordinal file, so in exactly the case this PR targets (ordinal present, undecodable) the old hash-addressed inode is orphaned under `hash/` when its hash differs from the recovered one. And `.atomic-*.tmp` files from a crash mid-write are never swept; harmless since `findAbove`/`cleanupAbove` filter on `toLongOption`.

Hasher selection for the descent checked out: `forOrdinal` and `hashSelect.select` agree at every ordinal around the 2572384 Kryo boundary, and the writers used `forOrdinal(snapshot.ordinal)` too. One pre-existing wrinkle I couldn't execute: Kryo-era info files kept by the logarithmic cutoff were written through the `GlobalSnapshotInfo` JSON storage but `readCombined` reads them as `GlobalSnapshotInfoV2`; circe should tolerate the extra fields, and it only matters after every post-boundary candidate has already failed.

No compile here, so the 13/13 claim is taken from the description.

Proph151Music:
Addressed in `30e8096d8`.

`Missing` and `OrdinalOccupied` are now treated as normal backward-walk results without warnings. Genuinely unusable anchors continue to warn. The focused recovery suites pass 6/6.

Operational clarification: if no valid persisted recovery anchor remains, the download fiber fails and the node remains in `WaitingForDownload`; the process does not exit. Restarting repeats the failure until the persisted snapshot storage is repaired using the documented operator procedure. Recovery promotion also requires filesystem support for `ATOMIC_MOVE`.

